PR #2699, opened 2024-02-02
Cleanup codec constants

abi87:
## Why this should be merged
This PR used to bring some changes to the codec that are not necessary anymore given the way I restructured dynamic fees.
Instead of fully removing the PR, I kept only some minor nits.

## How this works
Just nits

## How this was tested
CI


codec/manager.go
```diff
@@ -13,6 +13,8 @@ import (
 )
 
 const (
+	CodecVersionSize = wrappers.ShortLen
```
StephenButtolph, line 16: I think this stutters:
```suggestion
	VersionSize = wrappers.ShortLen
```

codec/reflectcodec/type_codec.go
```diff
@@ -502,7 +496,7 @@ func (c *genericCodec) marshal(
 // interface
 func (c *genericCodec) Unmarshal(bytes []byte, dest interface{}) error {
 	if dest == nil {
-		return errUnmarshalNil
+		return codec.ErrUnmarshalNil
```
StephenButtolph, line 499: nice

codec/reflectcodec/type_codec.go
```diff
@@ -84,7 +82,7 @@ func New(typer TypeCodec, tagNames []string) codec.Codec {
 
 func (c *genericCodec) Size(value interface{}) (int, error) {
 	if value == nil {
-		return 0, errMarshalNil // can't marshal nil
+		return 0, nil // can't marshal nil, we return zero size
```
StephenButtolph, line 85: I'm not exactly sure this is needed (talked offline)

StephenButtolph (approved):

codec/manager.go
```diff
@@ -13,6 +13,8 @@ import (
 )
 
 const (
+	VersionSize = wrappers.ShortLen
```
dhrubabasu, line 16: do we want to do VersionLen?

dhrubabasu (approved):

codec/manager.go
StephenButtolph: I thought about this... I think it makes sense as `Size` because it's used in the `Size` method... 🤷

PR #2700, opened 2024-02-02
X-chain - introducing fees calculators

abi87:
## Why this should be merged
This PR introduced `fees.Calculator` following closely what we have done in the `platformvm`.

## How this works

- Introduce `fee.Calculator` interface and `staticCalculator`
- Moved fee related part of avm config to fee package

## How this was tested
CI


vms/avm/txs/executor/semantic_verifier.go
```diff
@@ -146,6 +126,53 @@ func (v *SemanticVerifier) ExportTx(tx *txs.ExportTx) error {
...
+	tx *txs.BaseTx,
+	importedIns []*avax.TransferableInput,
+	exportedOuts []*avax.TransferableOutput,
+) error {
+	feeCalculator := fees.Calculator{
+		Config: v.Config,
+	}
+	if err := tx.Visit(&feeCalculator); err != nil {
+		return err
+	}
```
abi87, line 139: moved fee checks from SyntacticVerifier to SematincVerifier, where we will need it to be with dynamic fees

vms/avm/txs/executor/syntactic_verifier.go
```diff
@@ -55,17 +54,6 @@ func (v *SyntacticVerifier) BaseTx(tx *txs.BaseTx) error {
...
-		v.Config.TxFee,
-		v.FeeAssetID,
-		[][]*avax.TransferableInput{tx.Ins},
-		[][]*avax.TransferableOutput{tx.Outs},
-		v.Codec,
-	)
-	if err != nil {
-		return err
-	}
-
```
abi87, line 68: this and the following moved to SemanticVerifier

abi87:
> Does it make sense to have a function ?
> 
> ```go
> func CalculateStaticFees(c StaticConfig, tx *txs.Tx) (uint64, error)
> ```
> 
> Or maybe you have future plans no included in this PR ?

Plan is to introduce dynamic fees as a struct implementing `fee.Calculator`, similarly to what will happen with the P-chain.
I reckon static fees do not need a whole struct to be calculated. This is planning ahead a bit to handle uniformly fees pre and post fork introducing dynamic fees

PR #2702, opened 2024-02-03
[trace] upgrade opentelemetry to v1.22.0

bianyuanop:
## Why this should be merged
I'm now integrating celestia-da with hypervm created with hypersdk, where [celestia-node](https://github.com/celestiaorg/celestia-node/blob/main/go.mod#L61) has a otel dependency with version v1.22.0, which is not compatible with avalanchego‘s. To resolve this, I opened this PR. It's likely the new projects will have more probability to use more recent version of otel so this dependency conflicts will happen more often. 
## How this works
The `noOpTracer` struct change follows [this](https://github.com/open-telemetry/opentelemetry-go/issues/4714#issuecomment-1812879674) otel issue 
## How this was tested
it was tested by 
```shell
./scripts/build.sh
``` 
Then 
```shell
./scripts/tests.e2e.sh
```


trace/noop.go
```diff
@@ -7,16 +7,18 @@ import (
...
 
 	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
 
 	"github.com/ava-labs/avalanchego/utils/constants"
 )
 
 var Noop Tracer = noOpTracer{
-	t: trace.NewNoopTracerProvider().Tracer(constants.AppName),
+	t: noop.NewTracerProvider().Tracer(constants.AppName),
```
StephenButtolph, line 16: It seems like `noop.Tracer` already implements `Start` - so I think we can probably just get away with:

```golang
// noOpTracer is an implementation of Tracer that does nothing.
type noOpTracer struct {
	noop.Tracer
}

func (noOpTracer) Close() error {
	return nil
}
```

Thoughts?

go.mod
```diff
@@ -63,9 +63,9 @@ require (
 	golang.org/x/term v0.15.0
 	golang.org/x/time v0.0.0-20220922220347-f3bd1da661af
 	gonum.org/v1/gonum v0.11.0
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20230711160842-782d3b101e98
-	google.golang.org/grpc v1.58.3
-	google.golang.org/protobuf v1.31.0
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20231002182017-d307bd883b97
+	google.golang.org/grpc v1.60.1
```
StephenButtolph, line 67: Updating the `grpc` version has historically been... interesting. I think we'll probably want to hold off on merging this PR until a stable version of `v1.11.x` is released... We may also want to update to `v1.61.0` - as apparently that fixes a regression introduced in `v1.60`.

trace/noop.go
bianyuanop: That's better

go.mod
bianyuanop: yea sure

StephenButtolph:
@bianyuanop now that `v1.11.0` is out - I think this LGTM after the small nit + a rebase.

bianyuanop:
Done

StephenButtolph (dismissed):

go.mod
```diff
@@ -61,9 +61,9 @@ require (
...
```
StephenButtolph, line 65: Can we update this to `v1.62.0`? (or at least `v1.61.0`). I think we want to include this fix for a regression introduced in `v1.60.0`: https://github.com/grpc/grpc-go/pull/6922

go.sum
StephenButtolph, line 1: I think this needs a:
```
go mod tidy -compat=1.21
```

bianyuanop:
Upgraded grpc to v1.62.0 along with a `go mod tidy -compat=1.21`

trace/noop.go
```diff
@@ -4,24 +4,14 @@
 package trace
 
 import (
-	"context"
-
-	"go.opentelemetry.io/otel/trace"
-
-	"github.com/ava-labs/avalanchego/utils/constants"
+	"go.opentelemetry.io/otel/trace/noop"
 )
```
StephenButtolph, line 8: I think this needs to be:
```golang
import "go.opentelemetry.io/otel/trace/noop"
```

bianyuanop: just noticed that

StephenButtolph (approved):

PR #2704, opened 2024-02-05
Add `platform.getSubnet` API

felipemadero:
## Why this should be merged
There is no API related way to get subnet information such as transform subnet tx id, or updated owners

## How this works
It adds a new API call and client method, GetSubnet

## How this was tested
Tested on CLI branch developed to change subnet owners.


vms/platformvm/service.go
```diff
@@ -511,6 +511,83 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+// GetSubnetArgs are the arguments to GetSubnet
+type GetSubnetArgs struct {
+	// ID of the subnet to retrieve information about
+	SubnetID ids.ID `json:"subnetID"`
+}
+
+// GetSubnetResponse is the response from calling GetSubnet
+type GetSubnetResponse struct {
+	// ID of the subnet
+	SubnetID ids.ID `json:"subnetID"`
```
StephenButtolph, line 529: Why do we return the subnetID in the response? Won't the caller already know it from the request?

vms/platformvm/service.go
```diff
@@ -511,6 +511,83 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+	// whether it is permissioned or not
+	IsPermissioned bool `json:"isPermissioned"`
+	// subnet auth information for a permissioned subnet
+	ControlKeys []string    `json:"controlKeys"`
+	Threshold   json.Uint32 `json:"threshold"`
```
StephenButtolph, line 534: I think we would need to include the locktime here as well

vms/platformvm/service.go
```diff
@@ -511,6 +511,83 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+
+	response.SubnetID = args.SubnetID
+
+	if response.SubnetID == constants.PrimaryNetworkID {
+		response.IsPermissioned = false
+		response.ControlKeys = []string{}
+		response.Threshold = json.Uint32(0)
+		response.SubnetTransformationTxID = ids.Empty
+		return nil
+	}
```
StephenButtolph, line 556: I feel like we should just return an error here. The primary network isn't a subnet... (also these checks don't require the lock to be held)

vms/platformvm/service.go
```diff
@@ -511,6 +511,83 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+
+	if subnetTransformationTx, err := s.vm.state.GetSubnetTransformation(response.SubnetID); err == nil {
```
StephenButtolph, line 558: We should check for the actual error type - so that we don't report an unexpected error as incorrect information.

vms/platformvm/service.go
```diff
@@ -511,6 +511,83 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+		response.IsPermissioned = false
+		response.ControlKeys = []string{}
+		response.Threshold = json.Uint32(0)
```
StephenButtolph, line 561: We should still populate these fields for permissionless subnets

vms/platformvm/service.go
felipemadero: Yup. I had some confusion when making the rpc call with incorrect field name in the json (my mistake). The json was validated and the call returned info for platform without notice on my side. So I added the subnet ID. But it is best to just remove this and I also like returning error when requesting for platform (it is a way to validate the json param is correct).

danlaine:
Can we make a corresponding PR to update the docs with this new method?

vms/platformvm/service.go
felipemadero: what is it used for?

felipemadero:
> Can we make a corresponding PR to update the docs with this new method?

here it is:

https://github.com/ava-labs/avalanche-docs/pull/1658

felipemadero:
> > Can we make a corresponding PR to update the docs with this new method?
> 
> here it is:
> 
> [ava-labs/avalanche-docs#1658](https://github.com/ava-labs/avalanche-docs/pull/1658)

@StephenButtolph @danlaine addressed your comments

vms/platformvm/service.go
StephenButtolph: https://github.com/ava-labs/avalanchego/blob/master/vms/secp256k1fx/fx.go#L183-L184

vms/platformvm/service.go
```diff
@@ -510,6 +510,80 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+func (s *Service) GetSubnet(_ *http.Request, args *GetSubnetArgs, response *GetSubnetResponse) error {
+	s.vm.ctx.Log.Debug("API called",
+		zap.String("service", "platform"),
+		zap.String("method", "getSubnet"),
+	)
+
+	subnetID := args.SubnetID
+
+	if subnetID == constants.PrimaryNetworkID {
+		return errors.New("the primary network isn't a subnet")
```
StephenButtolph, line 546: We should define this error with the rest of the errors in this file.

vms/platformvm/service.go
```diff
@@ -510,6 +510,80 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+	Locktime    json.Uint64 `json:"locktime"`
+	// subnet transformation tx ID for a permissionless subnet
+	SubnetTransformationTxID ids.ID `json:"subnetTransformationTxID"`
+}
+
+func (s *Service) GetSubnet(_ *http.Request, args *GetSubnetArgs, response *GetSubnetResponse) error {
+	s.vm.ctx.Log.Debug("API called",
+		zap.String("service", "platform"),
+		zap.String("method", "getSubnet"),
+	)
```
StephenButtolph, line 541: Can we add `zap.Stringer("subnetID", args.SubnetID)` here?

vms/platformvm/service.go
```diff
@@ -510,6 +510,80 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+
+	subnetID := args.SubnetID
+
```
StephenButtolph, line 544: stylistically I probably would just use `args.SubnetID` in the two places below... but 🤷

vms/platformvm/service.go
```diff
@@ -510,6 +510,80 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
 	return nil
 }
 
+/*
+ *****************************************************
+ ******************* Get Subnet **********************
+ *****************************************************
+ */
+
```
StephenButtolph, line 518: I don't feel like these comments add anything anymore... (and they are really only around some of the old functions)

```suggestion
```

StephenButtolph:
just nits then lgtm

vms/platformvm/service.go
```diff
@@ -510,6 +511,73 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
...
+// GetSubnetResponse is the response from calling GetSubnet
+type GetSubnetResponse struct {
+	// whether it is permissioned or not
+	IsPermissioned bool `json:"isPermissioned"`
+	// subnet auth information for a permissioned subnet
+	ControlKeys []string    `json:"controlKeys"`
+	Threshold   json.Uint32 `json:"threshold"`
+	Locktime    json.Uint64 `json:"locktime"`
+	// subnet transformation tx ID for a permissionless subnet
+	SubnetTransformationTxID ids.ID `json:"subnetTransformationTxID"`
```
abi87, line 529: q: I wonder if we should return the whole parameters here instead of simply the txID? In such a casa I would return also Primary network config when primary network ID is specified.
I am not sure we expose this config via API currently and it'd be a more uniform handling of subnets and primary network

felipemadero: probably it is best to return formatted params instead of requiring the user to parse the tx. if @StephenButtolph agrees will make the change. also, regarding primary network params.

StephenButtolph (approved):

vms/platformvm/service.go
StephenButtolph: I'd be willing to mrege this PR as is. If we wanted to return more information about permissionless subnets I wouldn't necessarily object.

abi87 (approved):

PR #2710, opened 2024-02-07
Update minimum golang version to v1.21.7

dhrubabasu:
## Why this should be merged

go1.22 was released today. Golang only provides security patches for the prior version so we'll update to go1.21.7.

## How this works

- Update to go1.21.7
- Use the new `min` and `max` built-in functions instead of our `math.Max` and `math.Min` functions.
- Use the `clear` function instead of `maps.Clear`.
- Use the new `slices` pkg instead of `/x/exp/slices`.
- Use the new `maps` pkg instead of `/x/exp/maps`.
	- `maps.Keys` and `maps.Values` were removed in `go1.21` (https://github.com/golang/go/issues/61538) so we are still using the exp pkg in some files
- Use the new `cmp` pkg instead of `utils.Compare`
- Use `context.WithoutCancel` instead of `utils.Detach`

## How this was tested

CI


StephenButtolph:
`utils.Detach` should be replaced with `context.WithoutCancel` ref: https://go.dev/pkg/context/#WithoutCancel

StephenButtolph (approved):

PR #2711, opened 2024-02-07
Refactor chain manager subnets

joshua-kim:
## Why this should be merged

Needed as a pre-requisite to remove `vm.Initialize`. This refactors the currently running subnets into its own struct, so it can be shared with a future `chains.Factory` struct which will be responsible for creating `chain` instances (currently this is created internally in the chain manager). Ideally we'll have a chain factory to create chains and a chain manager that holds the chains that are currently running. This way, the chain manager doesn't need to know about the p-chain at all since the caller can choose arguments passed into the chain factory instead and the p-chain logic can be in the caller instead of in `chains/manager`

## How this works

Refactors currently running subnets into its own struct

## How this was tested

Added unit tests

### DIFF
--- a/chains/manager.go
+++ b/chains/manager.go
@@ -233,6 +233,8 @@ type ManagerConfig struct {
 	StateSyncBeacons []ids.NodeID
 
 	ChainDataDir string
+
+	Subnets *Subnets
 }
 
 type manager struct {
@@ -268,13 +270,12 @@ type manager struct {
 }
 
 // New returns a new Manager
-func New(config *ManagerConfig, subnets *Subnets) Manager {
+func New(config *ManagerConfig) Manager {
 	return &manager{
 		Aliaser:                ids.NewAliaser(),
 		ManagerConfig:          *config,
 		stakingSigner:          config.StakingTLSCert.PrivateKey.(crypto.Signer),
 		stakingCert:            staking.CertificateFromX509(config.StakingTLSCert.Leaf),
-		subnets:                subnets,
 		chains:                 make(map[ids.ID]handler.Handler),
 		chainsQueue:            buffer.NewUnboundedBlockingDeque[ChainParameters](initialQueueSize),
 		unblockChainCreatorCh:  make(chan struct{}),
--- a/chains/manager.go
+++ b/chains/manager.go
@@ -94,11 +94,11 @@ var (
 	// Bootstrapping prefixes for ChainVMs
 	ChainBootstrappingDBPrefix = []byte("bs")
 
-	errUnknownVMType           = errors.New("the vm should have type avalanche.DAGVM or snowman.ChainVM")
-	errCreatePlatformVM        = errors.New("attempted to create a chain running the PlatformVM")
-	errNotBootstrapped         = errors.New("subnets not bootstrapped")
-	errNoPrimaryNetworkConfig  = errors.New("no subnet config for primary network found")
-	errPartialSyncAsAValidator = errors.New("partial sync should not be configured for a validator")
+	errUnknownVMType            = errors.New("the vm should have type avalanche.DAGVM or snowman.ChainVM")
+	errCreatePlatformVM         = errors.New("attempted to create a chain running the PlatformVM")
+	errNotBootstrapped          = errors.New("subnets not bootstrapped")
+	errPrimaryNetworkNotRunning = errors.New("node is not running the primary network")
+	errPartialSyncAsAValidator  = errors.New("partial sync should not be configured for a validator")
 
 	fxs = map[ids.ID]fx.Factory{
 		secp256k1fx.ID: &secp256k1fx.Factory{},
@@ -256,10 +256,7 @@ type manager struct {
 	chainCreatorShutdownCh chan struct{}
 	chainCreatorExited     sync.WaitGroup
 
-	subnetsLock sync.RWMutex
-	// Key: Subnet's ID
-	// Value: Subnet description
-	subnets map[ids.ID]subnets.Subnet
+	subnets *Subnets
 
 	chainsLock sync.Mutex
 	// Key: Chain's ID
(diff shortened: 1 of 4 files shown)


chains/subnets.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Subnets holds the currently running subnets on this node
+type Subnets struct {
+	nodeID  ids.NodeID
+	configs map[ids.ID]subnets.Config
+
+	lock    sync.Mutex
+	subnets map[ids.ID]subnets.Subnet
+}
```
aaronbuchwald, line 42: nit: can we move the type definition above the constructor?

chains/subnets.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Add a subnet that is being run on this node
```
aaronbuchwald, line 44: nit: can we include a description of the return type in the comment?

chains/manager.go
```diff
@@ -271,13 +268,13 @@ type manager struct {
 }
 
 // New returns a new Manager
-func New(config *ManagerConfig) Manager {
+func New(config *ManagerConfig, subnets *Subnets) Manager {
```
aaronbuchwald, line 271: Should we just pass in `subnets` as part of the `ManagerConfig` in accordance with all of the other parameters?

It feels a bit weird that `subnets` would be the one exception here

chains/subnets.go
```diff
@@ -0,0 +1,87 @@
...
+func (s *Subnets) Get(subnetID ids.ID) (subnets.Subnet, bool) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	subnet, ok := s.subnets[subnetID]
+	return subnet, ok
+}
+
+// Bootstrapping returns subnets that have any chains that are still
+// bootstrapping.
```
aaronbuchwald, line 74:
```suggestion
// Bootstrapping returns the subnetIDs of any chains that are still
// bootstrapping.
```

aaronbuchwald (approved):
LGTM with optional nits

chains/manager.go
joshua-kim: I think long-term I would like to see us only store actual config values in `*Config` structs but I think it makes sense to put it in the config as per your comment for consistency for the time being

darioush (approved):

StephenButtolph (commented):
Running locally resulted in:
```
[02-10|13:01:49.389] INFO node/node.go:899 initializing chains
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1028a12c8]

goroutine 1 [running]:
github.com/ava-labs/avalanchego/chains.(*Subnets).Get(0x0?, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...})
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/chains/subnets.go:50 +0x28
github.com/ava-labs/avalanchego/chains.(*manager).StartChainCreator(0x14002a64380, {{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...})
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/chains/manager.go:1337 +0x4c
github.com/ava-labs/avalanchego/node.(*Node).initChains(0x1400024b200, {0x1400175f800, 0x1bc2, 0x3800})
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/node/node.go:910 +0x12c
github.com/ava-labs/avalanchego/node.New(0x140000d7600, {0x1039e9b18, 0x14000362380}, {0x1039f0320, 0x14000398080})
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/node/node.go:261 +0x14e4
github.com/ava-labs/avalanchego/app.New({{{0x6fc23ac00, 0x6fc23ac00, 0x6fc23ac00, 0x1bf08eb000}, {{0x0, {...}}, {0x0, 0x0}, {0x0, {...}, ...}, ...}, ...}, ...})
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/app/app.go:73 +0x3a0
main.main()
        /Users/stephen/go/src/github.com/ava-labs/avalanchego/main/main.go:48 +0x2cc
```
(which is presumably why all the E2E tests are failing).

chains/manager.go
```diff
@@ -256,10 +258,7 @@ type manager struct {
 	chainCreatorShutdownCh chan struct{}
 	chainCreatorExited     sync.WaitGroup
 
-	subnetsLock sync.RWMutex
-	// Key: Subnet's ID
-	// Value: Subnet description
-	subnets map[ids.ID]subnets.Subnet
+	subnets *Subnets
```
StephenButtolph, line 261: I think this needs to be removed (and it's usage replaced with `ManagerConfig.Subnets`

joshua-kim: Ah i fixed this on the chain manager PR but forgot to push the fix here as well. Will update.

chains/manager.go
```diff
@@ -288,25 +286,11 @@ func New(config *ManagerConfig) Manager {
...
-		}
-		sb = subnets.New(m.NodeID, sbConfig)
-		m.subnets[chainParams.SubnetID] = sb
-	}
-	addedChain := sb.AddChain(chainParams.ID)
-	m.subnetsLock.Unlock()
-
-	if !addedChain {
+	_ = m.subnets.Add(chainParams.SubnetID)
+	if sb, _ := m.subnets.Get(chainParams.SubnetID); !sb.AddChain(chainParams.ID) {
```
StephenButtolph, line 290: Might be a bit nit-picky.. but it feels pretty weird to `Add` and then immediately `Get`. Should we change `Add` to `GetOrCreate` and return the subnet?

node/node.go
```diff
@@ -304,6 +304,8 @@ type Node struct {
 	// Manages network timeouts
 	timeoutManager timeout.Manager
 
+	subnets *chains.Subnets
```
StephenButtolph, line 307: Is there a reason this is added to the struct? seems like it is only used in one function right now (perhaps used in a follow-up?)

joshua-kim: I don't think we need to have it on the struct... if we do we can fix it in the follow-up.

chains/subnets_test.go
```diff
@@ -0,0 +1,177 @@
...
+	config := map[ids.ID]subnets.Config{
+		constants.PrimaryNetworkID: {},
+	}
+
+	subnets, err := NewSubnets(ids.EmptyNodeID, config)
+	require.NoError(err)
+
+	subnet, ok := subnets.Get(constants.PrimaryNetworkID)
+	require.True(ok)
+	require.Equal(subnet.Config(), config[constants.PrimaryNetworkID])
```
StephenButtolph, line 27:
```suggestion
	require.Equal(config[constants.PrimaryNetworkID], subnet.Config())
```

chains/subnets_test.go
```diff
@@ -0,0 +1,177 @@
...
+			require := require.New(t)
+			config := map[ids.ID]subnets.Config{
+				constants.PrimaryNetworkID: {},
+			}
+			subnets, err := NewSubnets(ids.EmptyNodeID, config)
+			require.NoError(err)
+
+			for _, add := range tt.adds {
+				got := subnets.Add(add.subnetID)
+				require.Equal(got, add.want)
```
StephenButtolph, line 92:
```suggestion
				require.Equal(add.want, got)
```

StephenButtolph (approved):

PR #2714, opened 2024-02-08
Update `golangci-lint` to `v1.56.1`

dhrubabasu:
## Why this should be merged

Linters 🧹 

## How this works

- Updates version
- Enables `gci` and removes `goimports`
- Enables `spancheck`
- Enables `testifylint` and removes some of our bash linters
- Includes this patch: https://github.com/mgechev/revive/issues/965 (filed by @StephenButtolph ;))

## How this was tested

CI

### DIFF
--- a/snow/engine/avalanche/state/unique_vertex.go
+++ b/snow/engine/avalanche/state/unique_vertex.go
@@ -300,13 +300,13 @@ func (vtx *uniqueVertex) String() string {
 		len(txs),
 	))
 
-	parentFormat := fmt.Sprintf("\n    Parent[%s]: ID = %%s, Status = %%s",
+	parentFormat := fmt.Sprintf("\n    Parent[%s]: ID = %%s, Status = %%s", //nolint:perfsprint
 		formatting.IntFormat(len(parents)-1))
 	for i, parent := range parents {
 		sb.WriteString(fmt.Sprintf(parentFormat, i, parent.ID(), parent.Status()))
 	}
 
-	txFormat := fmt.Sprintf("\n    Transaction[%s]: ID = %%s, Status = %%s",
+	txFormat := fmt.Sprintf("\n    Transaction[%s]: ID = %%s, Status = %%s", //nolint:perfsprint
 		formatting.IntFormat(len(txs)-1))
 	for i, tx := range txs {
 		sb.WriteString(fmt.Sprintf(txFormat, i, tx.ID(), tx.Status()))
--- a/tests/fixture/test_data_server_test.go
+++ b/tests/fixture/test_data_server_test.go
@@ -15,19 +15,17 @@ import (
 // Check that funded test keys can be served from an http server to
 // ensure at-most-once allocation when tests are executed in parallel.
 func TestAllocatePreFundedKeys(t *testing.T) {
-	require := require.New(t)
-
 	keys := make([]*secp256k1.PrivateKey, 5)
 	for i := range keys {
 		key, err := secp256k1.NewPrivateKey()
-		require.NoError(err)
+		require.NoError(t, err)
 		keys[i] = key
 	}
 
 	uri, err := ServeTestData(TestData{
 		PreFundedKeys: keys,
 	})
-	require.NoError(err)
+	require.NoError(t, err)
 
 	testCases := []struct {
 		name              string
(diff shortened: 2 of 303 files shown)


vms/platformvm/validator_set_property_test.go
```diff
@@ -89,12 +91,12 @@ func TestGetValidatorsSetProperty(t *testing.T) {
 			// random events sequence received as test input
 			validatorsTimes, err := buildTimestampsList(events, currentTime, nodeID)
 			if err != nil {
-				return fmt.Sprintf("failed building events sequence: %s", err.Error())
+				return "failed building events sequence:" + err.Error()
```
StephenButtolph, line 94:
```suggestion
				return "failed building events sequence: " + err.Error()
```

vms/proposervm/proposer/windower_test.go
```diff
@@ -42,7 +42,7 @@ func TestWindowerNoValidators(t *testing.T) {
 
 	expectedProposer, err := w.ExpectedProposer(context.Background(), chainHeight, pChainHeight, slot)
 	require.ErrorIs(err, ErrAnyoneCanPropose)
-	require.Equal(ids.EmptyNodeID, expectedProposer)
+	require.Equal(expectedProposer, ids.EmptyNodeID)
```
StephenButtolph, line 45: Seems like a false positive

StephenButtolph (commented):
sick 🔥 🚀

StephenButtolph (approved):

snow/engine/avalanche/state/unique_vertex.go
```diff
@@ -300,13 +300,13 @@ func (vtx *uniqueVertex) String() string {
 		len(txs),
 	))
 
-	parentFormat := fmt.Sprintf("\n    Parent[%s]: ID = %%s, Status = %%s",
+	parentFormat := fmt.Sprintf("\n    Parent[%s]: ID = %%s, Status = %%s", //nolint:perfsprint
```
abi87, line 303: why are we disabling linting here? There were other places where we were able to replace  Sprintf with simpler alternative. What's the difference here?

tests/fixture/test_data_server_test.go
```diff
@@ -15,19 +15,17 @@ import (
 // Check that funded test keys can be served from an http server to
 // ensure at-most-once allocation when tests are executed in parallel.
 func TestAllocatePreFundedKeys(t *testing.T) {
-	require := require.New(t)
-
 	keys := make([]*secp256k1.PrivateKey, 5)
 	for i := range keys {
 		key, err := secp256k1.NewPrivateKey()
-		require.NoError(err)
+		require.NoError(t, err)
```
abi87, line 21: why? Is there a linter complaining about this?

trace/noop.go
```diff
@@ -21,7 +21,7 @@ type noOpTracer struct {
 }
 
 func (n noOpTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-	return n.t.Start(ctx, spanName, opts...)
+	return n.t.Start(ctx, spanName, opts...) //nolint:spancheck
```
abi87, line 24: the linter suggests there is a possible memory leak here. Should we end the spam right away for the noop tracer instead of silencing the linter?
NoOpTracer does not seem to be used in prod code currently but I think it's best to avoid any possible memory leak, even in UTs

abi87 (changes requested):
just a couple of points where we may avoid silencing the linter. All in all nice cleanup!

snow/engine/avalanche/state/unique_vertex.go
dhrubabasu: I tried changing it a few times but the code got more confusing. We don't stringify the unique vertex enough to sacrifice code clarity for performance IMO

tests/fixture/test_data_server_test.go
dhrubabasu: When we use require, we should use it with the most scoped `t` variable. Further down this test, we do `require := require.New(t)` in the sub-test. The linter was complaining that the sub-test `t *testing.T` was unused.

trace/noop.go
dhrubabasu: Removed the struct altogether- it wasn't used anywhere.

dhrubabasu: The consumer should be ending the span. The linter is complaining that the returned span is unassigned so there could be a possible memory leak

abi87 (approved):

PR #2717, opened 2024-02-08
Replace snowball/snowflake interface with single shared snow interface

aaronbuchwald:
## Why this should be merged

This PR replaces the separate snowball/snowflake interfaces for the Unary, Binary, and Nnary consensus types with a single shared snow interface.

The only difference that we change is previously `UnarySnowball` had a `RecordPollPreference` function whereas `UnarySnowflake` did not. The reason for this is that in the unary case, there is only one possible preference and we have not received enough votes to increment our confidence towards finalizing.

Therefore, in the snowball case where we maintain a confidence counter for preference, this has the impact of increasing our confidence in our current preference even though we don't get closer to finalizing.

In snowflake, there is no notion of incrementing confidence in our preference, so `RecordPollPreference` simply resets the confidence counter on a unary instance.

This also removes the slush interfaces since there is only a single implementation.

## How this works

This code change is intended solely as a refactor to reduce instances where we have an interface where there is only a single implementation.

## How this was tested

This was tested with the existing unit tests and uses the same snowball/snowflake consensus factory as the unit tests previously used since the current tests are specific to the snowball/snowflake implementation behaviors.


snow/consensus/snowball/binary_snowball.go
```diff
@@ -5,8 +5,6 @@ package snowball
 
 import "fmt"
 
-var _ BinarySnowball = (*binarySnowball)(nil)
```
aaronbuchwald, line 8: switch from removing to ensure it implements `BinarySnow`. Same for similar changes below.

snow/consensus/snowball/nnary_snowball.go
```diff
@@ -9,7 +9,21 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 )
 
-var _ NnarySnowball = (*nnarySnowball)(nil)
+var (
+	_ NnarySnow = (*nnarySnowball)(nil)
+)
```
aaronbuchwald, line 14: nit: one line

snow/consensus/snowball/nnary_snowball.go
```diff
@@ -9,7 +9,21 @@ import (
...
+
+type SnowballFactory struct{}
```
aaronbuchwald, line 16: nit: add type interface check

snow/consensus/snowball/tree.go
```diff
@@ -17,33 +17,34 @@ var (
...
 
 	return t
 }
 
-// Tree implements the snowball interface by using a modified patricia tree.
+// Tree implements the snow interface by using a modified patricia tree.
 type Tree struct {
-	// node is the root that represents the first snowball instance in the tree,
-	// and contains references to all the other snowball instances in the tree.
+	// node is the root that represents the first snow instance in the tree,
```
aaronbuchwald, line 39: Q: are we alright with using snow as the word for either snowflake or snowball

snow/consensus/snowball/unary_snowflake.go
```diff
@@ -33,6 +33,13 @@ func (sf *unarySnowflake) RecordSuccessfulPoll() {
 	sf.finalized = sf.finalized || sf.confidence >= sf.beta
 }
 
+// RecordPollPreference fails to reach an alpha threshold to increase our confidence,
+// so this is a passthrough to RecordUnsuccessfulPoll to reset our confidence counter
+// towards accepting.
```
aaronbuchwald, line 38:
```suggestion
// RecordPollPreference fails to reach an alpha threshold to increase our confidence,
// so this calls RecordUnsuccessfulPoll to reset the confidence counter.
```

snow/consensus/snowball/nnary_snowflake.go
```diff
@@ -9,7 +9,22 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 )
 
-var _ NnarySnowflake = (*nnarySnowflake)(nil)
+var (
+	_ NnarySnow   = (*nnarySnowflake)(nil)
+	_ ConsensusFactory = (snowflakeFactory{})
```
aaronbuchwald, line 14: Q: should this be exported same as snowball even though it's not used?

Probably not for now

snow/consensus/snowball/consensus.go
```diff
@@ -40,15 +40,19 @@ type Consensus interface {
...
-
-// NnarySnowflake is a snowflake instance deciding between an unbounded number
-// of values. After performing a network sample of k nodes, if you have alpha
-// votes for one of the choices, you should vote for that choice. Otherwise, you
-// should reset.
-type NnarySnowflake interface {
+// ConsensusFactory produces NnarySnowflake instances
+type ConsensusFactory interface {
+	New(params Parameters, choice ids.ID) NnarySnow
+	NewUnary(params Parameters) UnarySnow
```
joshua-kim, line 46: Can we just have a single `New` function and expect the parameters to be passed into the the implementation so that we can have a single function exposed?

snow/consensus/snowball/nnary_snowball.go
```diff
@@ -9,7 +9,22 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 )
 
-var _ NnarySnowball = (*nnarySnowball)(nil)
+var (
+	_ NnarySnow        = (*nnarySnowball)(nil)
+	_ ConsensusFactory = SnowballFactory{}
```
joshua-kim, line 14: nit: fix this i/f check for consistency with the rest of the repo

snow/consensus/snowball/nnary_snowflake.go
```diff
@@ -9,7 +9,22 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 )
 
-var _ NnarySnowflake = (*nnarySnowflake)(nil)
+var (
+	_ NnarySnow        = (*nnarySnowflake)(nil)
+	_ ConsensusFactory = (SnowflakeFactory{})
```
joshua-kim, line 14: Same comment here

snow/consensus/snowball/tree.go
```diff
@@ -17,33 +17,34 @@ var (
 	_ node      = (*binaryNode)(nil)
 )
 
-func NewTree(params Parameters, choice ids.ID) Consensus {
-	snowball := newUnarySnowball(params.BetaVirtuous)
+func NewTree(cf ConsensusFactory, params Parameters, choice ids.ID) Consensus {
```
joshua-kim, line 20: Can we just spell this out as `consensusFactory` instead of `cf`?

snow/consensus/snowball/consensus.go
```diff
@@ -40,15 +40,19 @@ type Consensus interface {
...
-// number of positive responses from a network sample.
-type NnarySnowball interface{ NnarySnowflake }
-
-// NnarySnowflake is a snowflake instance deciding between an unbounded number
-// of values. After performing a network sample of k nodes, if you have alpha
-// votes for one of the choices, you should vote for that choice. Otherwise, you
-// should reset.
-type NnarySnowflake interface {
+// ConsensusFactory produces NnarySnowflake instances
+type ConsensusFactory interface {
```
StephenButtolph, line 44: Should we just call this `Factory`? The current naming seems pretty weird because we don't actually return a `Consensus` object anywhere in this interface.

snow/consensus/snowball/nnary_snowball.go
```diff
@@ -9,7 +9,22 @@ import (
...
+
+func (SnowballFactory) New(params Parameters, choice ids.ID) NnarySnow {
+	sb := newNnarySnowball(params.BetaVirtuous, params.BetaRogue, choice)
+	return &sb
+}
+
+func (SnowballFactory) NewUnary(params Parameters) UnarySnow {
+	sb := newUnarySnowball(params.BetaVirtuous)
+	return &sb
+}
```
StephenButtolph, line 27: Should we have a factory file where we implement both `SnowballFactory` and `SnowflakeFactory`? This file seems like a bit of a weird spot (because the factory creates both `Nnary` and `Unary`)

snow/consensus/snowball/consensus.go
```diff
@@ -40,15 +40,19 @@ type Consensus interface {
...
+	New(params Parameters, choice ids.ID) NnarySnow
```
StephenButtolph, line 45: Could we be precise with both functions?
```suggestion
	NewNnary(params Parameters, choice ids.ID) NnarySnow
```

snow/consensus/snowball/consensus_performance_test.go
```diff
@@ -29,10 +29,10 @@ func TestDualAlphaOptimization(t *testing.T) {
 		source        = prng.NewMT19937()
 	)
 
-	singleAlphaNetwork := NewNetwork(params, numColors, source)
+	singleAlphaNetwork := NewNetwork(SnowballFactory{}, params, numColors, source)
```
StephenButtolph, line 32: I think this is purely stylistic. But I kind of like defining `SnowballFactory` as a global like we did with `trace.NoOp` [here](https://github.com/ava-labs/avalanchego/blob/master/trace/noop.go#L14-L16).

snow/consensus/snowball/flat.go
```diff
@@ -10,17 +10,19 @@ import (
...
-		params:        params,
+		NnarySnow: sf,
+		params:    params,
 	}
 }
 
-// Flat is a naive implementation of a multi-choice snowball instance
+// Flat is a naive implementation of a multi-choice snow instance
 type Flat struct {
 	// wraps the n-nary snowball logic
```
StephenButtolph, line 24: nit: cleanup `snowball` reference

snow/consensus/snowball/flat.go
```diff
@@ -10,17 +10,19 @@ import (
 
 var _ Consensus = (*Flat)(nil)
 
-func NewFlat(params Parameters, choice ids.ID) Consensus {
+func NewFlat(factory ConsensusFactory, params Parameters, choice ids.ID) Consensus {
+	sf := factory.New(params, choice)
```
StephenButtolph, line 14: `sf` is typically shorthand for `snowflake` - should we just call this `nnary` or something? We could also just reduce the diff by removing the local variable and calling the function during the construction of the `Flat` struct like before

snow/consensus/snowball/nnary_snowflake.go
StephenButtolph, line 14:
```suggestion
	_ ConsensusFactory = SnowflakeFactory{}
```

snow/consensus/snowball/consensus.go
StephenButtolph: These return different interfaces

snow/consensus/snowball/consensus.go
```diff
@@ -40,15 +40,19 @@ type Consensus interface {
...
+	NewUnary(params Parameters) UnarySnow
+}
+
+// NnarySnow is a snow instance deciding between an unbounded number
+// of values.
+// The caller samples k nodes and then calls
+// 1. RecordSuccessfulPoll if choice collects >= alphaConfidence votes
+// 2. RecordPollPreference if choice collects >= alphaPreference votes
+// 3. RecordUnsuccessfulPoll otherwise
+type NnarySnow interface {
```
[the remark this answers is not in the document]
aaronbuchwald: Ya that makes sense, that will also make it so that it doesn't stutter when used outside of this package

snow/consensus/snowball/consensus.go
aaronbuchwald: Ya.. I had thought of these as consensus, but that's a bit confusing because there's already a different consensus type.

I don't love making it just factory by itself, but not sure what else it would be. I'll think a bit more about this and if I don't come up with anything just make it factory.

snow/consensus/snowball/tree.go
aaronbuchwald: We should use the same word here that we decide to go with for `nnary`: https://github.com/ava-labs/avalanchego/pull/2717#discussion_r1486614751

snow/consensus/snowball/tree.go
aaronbuchwald: Yup

snow/consensus/snowball/nnary_snowball.go
aaronbuchwald: Ya, we can change this. I made it this way since it's just a value, but can definitely conform for style.

snow/consensus/snowball/nnary_snowball.go
aaronbuchwald: Ya I think it'll be cleaner that way, good idea

snow/consensus/snowball/flat.go
aaronbuchwald: Ya gonna move this into the struct creation

snow/consensus/snowball/consensus_performance_test.go
aaronbuchwald: Oh that looks way nicer, thanks 🙏

joshua-kim (approved):

snow/consensus/snowball/consensus.go
```diff
@@ -40,15 +40,19 @@ type Consensus interface {
...
-// should reset.
-type NnarySnowflake interface {
+// Factory produces Nnary and Unary decision instances
+type Factory interface {
+	NewNnary(params Parameters, choice ids.ID) Nnary
+	NewUnary(params Parameters) Unary
+}
+
+// Nnary is a snow instance deciding between an unbounded number
+// of values.
```
StephenButtolph, line 50:
```suggestion
// Nnary is a snow instance deciding between an unbounded number of values.
```

snow/consensus/snowball/unary_snowflake.go
```diff
@@ -33,6 +33,12 @@ func (sf *unarySnowflake) RecordSuccessfulPoll() {
 	sf.finalized = sf.finalized || sf.confidence >= sf.beta
 }
 
+// RecordPollPreference fails to reach an alpha threshold to increase our confidence,
+// so this calls RecordUnsuccessfulPoll to reset the confidence counter.
```
StephenButtolph, line 37:
```suggestion
// RecordPollPreference fails to reach an alpha threshold to increase our
// confidence, so this calls RecordUnsuccessfulPoll to reset the confidence
// counter.
```

StephenButtolph (approved):
small nits around comment lengths (wrapping at 80 chars)

PR #2718, opened 2024-02-08
P-chain - Tx builder cleanup

abi87:
## Why this should be merged
The upcoming work on dynamic fees will change they way we fund P-chain transactions.
This is much easier to be done and reviewed if we use txBuilder package as much as possible in UTs, instead of trying to build txs "manually".
Also we have ideas to eventually make txBuilder a test only package and use the wallet as the way to build and func transactions. This PR simplifies the upcoming changes

## How this works
Make utxo.Spend and utxo.Authorized be called only in:

1. Unit tests
2. txBuilder package

## How this was tested
CI


vms/platformvm/txs/executor/import_test.go
```diff
@@ -36,63 +38,16 @@ func TestNewImportTx(t *testing.T) {
 	sourceKey, err := secp256k1.NewPrivateKey()
 	require.NoError(t, err)
 
-	cnt := new(byte)
-
-	// Returns a shared memory where GetDatabase returns a database
-	// where [recipientKey] has a balance of [amt]
-	fundedSharedMemory := func(peerChain ids.ID, assets map[ids.ID]uint64) atomic.SharedMemory {
```
abi87, line 43: this a clean way to setup the shared memory when we want to build an importTx. Pulled the function out of the test to be reused in other tests

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -965,44 +987,23 @@ func TestDurangoDisabledTransactions(t *testing.T) {
 		{
 			name: "AddValidatorTx",
 			buildTx: func(env *environment) *txs.Tx {
-				ins, unstakedOuts, stakedOuts, signers, err := env.utxosHandler.Spend(
```
abi87, line 968: This is what I mean by building a tx "manually". The tx is supposed to pass verification, so we really should use the txBuilder to do that. This also shorten the test

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -1110,205 +1091,122 @@ func TestDurangoMemoField(t *testing.T) {
...
-					},
-				}
-				return tx, signers, onAcceptState, &tx.Memo
+				return tx, onAcceptState
 			},
 		},
 		{
 			name: "ImportTx",
-			setupTest: func(env *environment) (txs.UnsignedTx, [][]*secp256k1.PrivateKey, state.Diff, *types.JSONByteSlice) {
+			setupTest: func(env *environment, memoField []byte) (*txs.Tx, state.Diff) {
```
abi87, line 1152: This test is really the reason why I added memo fields in the txBuilder functions.
With the activation of Durango, we zeroed memo fields, so it may seem weird that I introduce them back.
However txBuilder is going to become a test package and we may want to be able to setup memos in tests.

vms/platformvm/vm_regression_test.go
```diff
@@ -1480,51 +1506,20 @@ func TestSubnetValidatorBLSKeyDiffAfterExpiry(t *testing.T) {
 	require.NoError(err)
 
 	// build primary network validator with BLS key
-	utxoHandler := utxo.NewHandler(vm.ctx, &vm.clock, vm.fx)
-	ins, unstakedOuts, stakedOuts, signers, err := utxoHandler.Spend(
-		vm.state,
-		keys,
+	primaryTx, err := vm.txBuilder.NewAddPermissionlessValidatorTx(
```
abi87, line 1509: some more "manually built" txs that we can cleanup

vms/platformvm/txs/builder/builder.go
```diff
@@ -92,6 +96,27 @@ type DecisionTxBuilder interface {
 		ownerAddrs []ids.ShortID,
 		keys []*secp256k1.PrivateKey,
 		changeAddr ids.ShortID,
+		memo []byte,
+	) (*txs.Tx, error)
+
+	NewTransformSubnetTx(
```
abi87, line 102: Added since needed in tests

vms/platformvm/txs/builder/builder.go
```diff
@@ -49,6 +50,7 @@ type AtomicTxBuilder interface {
 		to ids.ShortID,
 		keys []*secp256k1.PrivateKey,
 		changeAddr ids.ShortID,
+		memo []byte,
```
abi87, line 53: we zero memo fields upon Durango activatio, but we may still want to specify them in UTs

vms/platformvm/txs/executor/standard_tx_executor_test.go
abi87: Currently memo fields are disabled, but we may consider reintroducing them once we properly price txs.
I think generally we want to allow maximal flexibility in building txs, especially in UTs.
Note for instance that with Durango activation we won't need to specify StartTime in any of the AddValidator/Delegator txs but we still explicitly pass start time in the builder signature.
Not sure I understand what should go in the README here?

vms/platformvm/txs/builder/builder.go
abi87: I'd prefer keeping the convention of naming builders as `New + tx_name`.
The unintentional use is an interesting point but I think it's not a problem here. Generally:

- We don't build txs in the nodes usually. X-chain doesn't even have a tx builder package!
- Nodes receives txs either by gossip (no creation there, just parsing and verification) or APIs
- P-chain APIs allowing you to build txs are currently deprecated. Preferred solution is using the wallet package. We may drop the APIs with E upgrade, after which txBuilder will be only a test commodity
- Full discloruse, there is a single tx `RewardValidatorTx` that is generaled automatically by the node when it's time to evict a staker. It has its own builder, whichi we moved outside of txbuilder package recently (it's a simple tx that needs no inputs/outputs, just carries the ID of the staker to be rewarded).

vms/platformvm/txs/builder/builder.go
abi87: I think two main points here:

- We tend to have txBuilder as general as possible, allowing them to specify fields that latest fork has dropped (see start time in AddValidator/Delegator tx)
- We may reactivate memos in the future, once we price them correctly

On the other side:

- Dropping start time from AddValidator/Delegator txs builder would have been a major pain due to extensive use in UTs. That's not the case for the memo, which is never specified except in the single UT were we show it's disabled with Durango. I judged it's best to introduce them in builders (as we do in the wallet itself as well), but I think added value is definitely not a large one. I could have avoid specifiing the memo field and build "manually" the txs in that single UT. I just dislike this alternative as bulding txs with dynamic fees requires more complex code and I strived to avoid code duplication as much as possible.

vms/platformvm/txs/builder/builder.go
```diff
@@ -480,6 +521,70 @@ func (b *builder) NewCreateSubnetTx(
 	return tx, tx.SyntacticVerify(b.ctx)
 }
 
+func (b *builder) NewTransformSubnetTx(
+	subnetID ids.ID,
```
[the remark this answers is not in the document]
abi87: I followed what we do in the wallet (see https://github.com/ava-labs/avalanchego/blob/master/wallet/chain/p/builder.go#L200). This is one of the subjects where we don't have a clear line to follow across the codebase.

vms/platformvm/txs/builder/builder.go
```diff
@@ -480,6 +521,70 @@ func (b *builder) NewCreateSubnetTx(
...
+	maxStakeDuration time.Duration,
+	minDelegationFee uint32,
+	minDelegatorStake uint64,
+	maxValidatorWeightFactor byte,
+	uptimeRequirement uint32,
+	keys []*secp256k1.PrivateKey,
+	changeAddr ids.ShortID,
+	memo []byte,
+) (*txs.Tx, error) {
+	ins, unstakedOuts, _, signers, err := b.Spend(b.state, keys, 0, b.cfg.TransformSubnetTxFee, changeAddr)
```
StephenButtolph, line 543: I think we typically just call these `outs` unless we are distinguishing between staked and unstaked outputs.
```suggestion
	ins, outs, _, signers, err := b.Spend(b.state, keys, 0, b.cfg.TransformSubnetTxFee, changeAddr)
```

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -853,7 +872,8 @@ func TestBanffStandardTxExecutorAddValidator(t *testing.T) {
 			ids.ShortEmpty,
 			reward.PercentDenominator, // shares
 			[]*secp256k1.PrivateKey{preFundedKeys[0]},
-			ids.ShortEmpty, // change addr // key
+			ids.ShortEmpty, // change addr
+			nil,            // key
```
StephenButtolph, line 876:
```suggestion
			nil,
```

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -1734,7 +1530,7 @@ func TestStandardExecutorRemoveSubnetValidatorTx(t *testing.T) {
 				env := newValidRemoveSubnetValidatorTxVerifyEnv(t, ctrl)
 
 				// Set dependency expectations.
-				env.state.EXPECT().GetTimestamp().Return(env.latestForkTime)
+				env.state.EXPECT().GetTimestamp().Return(env.latestForkTime).AnyTimes().AnyTimes()
```
StephenButtolph, line 1533: Was this change intended? It seems like we should at least remove one of the `AnyTimes()` calls here.

vms/platformvm/txs/executor/import_test.go
```diff
@@ -193,3 +158,51 @@ func TestNewImportTx(t *testing.T) {
 		})
 	}
 }
+
+// Returns a shared memory where GetDatabase returns a database
+// where [recipientKey] has a balance of [amt]
+func fundedSharedMemory(t *testing.T, env *environment, sourceKey *secp256k1.PrivateKey, peerChain ids.ID, assets map[ids.ID]uint64) atomic.SharedMemory {
```
StephenButtolph, line 164:
```suggestion
func fundedSharedMemory(
	t *testing.T,
	env *environment,
	sourceKey *secp256k1.PrivateKey,
	peerChain ids.ID,
	assets map[ids.ID]uint64,
) atomic.SharedMemory {
```

StephenButtolph (approved):

vms/platformvm/txs/builder/builder.go
abi87: Done

vms/platformvm/txs/executor/standard_tx_executor_test.go
abi87: Done

vms/platformvm/txs/executor/standard_tx_executor_test.go
abi87: Mixed up manual and "scripted" changes. Dropped

vms/platformvm/txs/executor/import_test.go
abi87: Done

PR #2723, opened 2024-02-09
decouple chain creation and chain registry

joshua-kim:
## Why this should be merged

Refactors chain creation so that the p-chain can be created outside of the chain manager by passing in a vm factory to the chain factory. This is needed for us to unblock removal of `vm.Initialize`.

## How this works

Exposes chain creation into the `chains.Factory` struct

## How this was tested

Bootstrapped a node

### DIFF
--- a/chains/factory.go
+++ b/chains/factory.go
@@ -1,0 +1,150 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package chains
+
+import (
+	"crypto"
+	"crypto/tls"
+	"time"
+
+	"github.com/ava-labs/avalanchego/api/health"
+	"github.com/ava-labs/avalanchego/api/keystore"
+	"github.com/ava-labs/avalanchego/api/metrics"
+	"github.com/ava-labs/avalanchego/chains/atomic"
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/message"
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/snow/networking/router"
+	"github.com/ava-labs/avalanchego/snow/networking/sender"
+	"github.com/ava-labs/avalanchego/snow/networking/timeout"
+	"github.com/ava-labs/avalanchego/snow/networking/tracker"
+	"github.com/ava-labs/avalanchego/snow/validators"
+	"github.com/ava-labs/avalanchego/staking"
+	"github.com/ava-labs/avalanchego/subnets"
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/vms"
+)
+
+type FactoryConfig struct {
+	NetworkID                               uint32
+	PartialSyncPrimaryNetwork               bool
+	SybilProtectionEnabled                  bool
+	TracingEnabled                          bool
+	MeterVMEnabled                          bool
+	StateSyncBeacons                        []ids.NodeID
+	BootstrapMaxTimeGetAncestors            time.Duration
+	BootstrapAncestorsMaxContainersSent     int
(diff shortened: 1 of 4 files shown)


chains/manager.go
```diff
@@ -393,39 +294,47 @@ func (m *manager) createChain(chainParams ChainParameters) {
...
-func (m *manager) buildChain(chainParams ChainParameters, sb subnets.Subnet) (*chain, error) {
-	if chainParams.ID != constants.PlatformChainID && chainParams.VMID == constants.PlatformVMID {
+// New returns a new chain
+func (f *Factory) New(
+	chainID, subnetID, vmID ids.ID,
+	fxIDs []ids.ID,
+	genesis []byte,
+	beacons validators.Manager,
+	vmFactory vms.Factory,
+) (*Chain, error) {
```
StephenButtolph, line 315: Should these factory functions be defined in the factory file? Or was the hope to minimize the diff in this PR?

joshua-kim: Yeah they should be. I did this to make the diff easier to review since it otherwise the diff becomes harder to review as you need to switch back and forth both files. I was planning on doing the move into the factory file in a follow-up once this was merged.

chains/manager.go
```diff
@@ -393,39 +294,47 @@ func (m *manager) createChain(chainParams ChainParameters) {
...
+	if chain.Context.ChainID == constants.PlatformChainID {
 		if err := m.registerBootstrappedHealthChecks(); err != nil {
 			chain.Handler.StopWithError(context.TODO(), err)
 		}
 	}
 
 	// Tell the chain to start processing messages.
 	// If the X, P, or C Chain panics, do not attempt to recover
-	chain.Handler.Start(context.TODO(), !m.CriticalChains.Contains(chainParams.ID))
+	chain.Handler.Start(context.TODO(), !m.CriticalChains.Contains(chain.Context.XChainID))
```
StephenButtolph, line 305: image
```suggestion
	chain.Handler.Start(context.TODO(), !m.CriticalChains.Contains(chain.Context.ChainID))
```

joshua-kim: Yeah this is a bug

chains/manager.go
```diff
@@ -258,21 +200,16 @@ type manager struct {
...
 	return &manager{
-		Aliaser:                ids.NewAliaser(),
+		Aliaser:                config.Aliaser,
 		ManagerConfig:          *config,
-		stakingSigner:          config.StakingTLSCert.PrivateKey.(crypto.Signer),
-		stakingCert:            staking.CertificateFromX509(config.StakingTLSCert.Leaf),
 		chains:                 make(map[ids.ID]handler.Handler),
 		chainsQueue:            buffer.NewUnboundedBlockingDeque[ChainParameters](initialQueueSize),
-		unblockChainCreatorCh:  make(chan struct{}),
+		unblockChainCreatorCh:  config.UnblockChainCreatorCh,
```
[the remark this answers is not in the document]
joshua-kim: Do you mean as a parameter like `StartChainCreator(...)`? If so I think beyond this PR I would like for this channel to not even be in this struct... I feel like the caller of this code (node) should choose when this goroutine starts. I think in an ideal world we have some channel or event subscription thing where the subnet can publish a bootstrapped event and the node spins up goroutine that blocks until it sees the bootstrapped event. For now I left it as a field this struct instead of as a parameter since it keeps the current behavior.

If instead you meant just use `config.UnblockChainCreatorCh` instead of keeping this field then yeah that makes sense to me

chains/factory.go
```diff
@@ -0,0 +1,154 @@
...
+	nodeID                ids.NodeID
+	keystore              keystore.Keystore
+	msgCreator            message.Creator
+	metrics               metrics.MultiGatherer
+	router                router.Router
+	net                   sender.ExternalSender
+	blockAcceptorGroup    snow.AcceptorGroup
+	txAcceptorGroup       snow.AcceptorGroup
+	vertexAcceptorGroup   snow.AcceptorGroup
+	stakingCert           *staking.Certificate
```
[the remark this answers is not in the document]
joshua-kim: Re-ordered these to try to preserve similar ordering to what chains.Manager currently has

StephenButtolph:
@joshua-kim going to close this because it is stale. If you still want it to get merged then please re-open it.

PR #2724, opened 2024-02-10
Provide pgo file during compilation

StephenButtolph:
## Why this should be merged

> [Profile-guided optimization (PGO)](https://go.dev/doc/pgo), also known as feedback-directed optimization (FDO), is a compiler optimization technique that feeds information (a profile) from representative runs of the application back into to the compiler for the next build of the application, which uses that information to make more informed optimization decisions. For example, the compiler may decide to more aggressively inline functions which the profile indicates are called frequently.

## How this works

PGO is "ready for general use" as of go1.21. This adds a CPU profile from a validator into the default pgo location.

Also, it is expected to commit these profiles to the source tree:

> Committing profiles directly in the source repository is recommended as profiles are an input to the build important for reproducible (and performant!) builds. Storing alongside the source simplifies the build experience as there are no additional steps to get the profile beyond fetching the source.

## How this was tested

- [X] Made sure the compiler was picking up the location of the file by testing with an invalid profile.
- [X] CI


StephenButtolph:
Note: updating this file should be added as part of the version bump PR from now on.

dhrubabasu (approved):

JacobEverly (approved):

PR #2731, opened 2024-02-13
Use `BaseTx` in P-chain wallet

dhrubabasu:
## Why this should be merged

Shouldn't be merged until Durango is activated on both Mainnet and Fuji

## How this works

Uses `BaseTx` :)

## How this was tested

CI + on Fuji


StephenButtolph (approved):

abi87 (approved):

abi87:
Looking forward to merging this. In https://github.com/ava-labs/avalanchego/pull/2751 I managed to use the wallet for the deprecated P-chain APIs but the BaseTx fee is off. Mergeing this would solve it

PR #2607, opened 2024-01-13
P-chain - Memo field zeroed post Durango 

abi87:
## Why this should be merged
Post Durango we're going to enforce the Memo field is zero for any P-chain tx

## How this works
Introduced extra checks. No changes to wallets and APIs yet. We'll revisit it post Durango activation

## How this was tested
CI

PR #2732, opened 2024-02-13
Remove `Memo` field from P-chain wallet

dhrubabasu:
## Why this should be merged

The `Memo` field must be empty post-Durango (https://github.com/ava-labs/avalanchego/pull/2607). This PR removes it from the P-chain wallet.

## How this works

Don't populate the `Memo` field

## How this was tested

CI


joshua-kim (approved):

abi87 (commented):
I am favour of the cleanup, but I wonder if instead of "silentily" avoiding to include the memo field in any P-chain tx, we should explicitly err if ops.Memo() is non-empty.
It is probably not super relevant for the P-chain, but it may help wallet UX

StephenButtolph:
I'm going to close this - I think there are still plans to re-introduce the memos into this code... and I think it's probably better to error if someone thinks they're providing memos rather than silently not including them.

PR #2734, opened 2024-02-13
P-chain wallet fees UTs

abi87:
## Why this should be merged
Even before introduction of dynamic fees over the P-chain, we can really use some unit tests that make sure that P-chain wallet spends enough to pay transactions fees and no more.

## How this works
Added UTs

## How this was tested
Just added tests.


scripts/mocks.mockgen.txt
```diff
@@ -40,5 +40,7 @@ github.com/ava-labs/avalanchego/vms/proposervm=PostForkBlock=vms/proposervm/mock
 github.com/ava-labs/avalanchego/vms/registry=VMGetter=vms/registry/mock_vm_getter.go
 github.com/ava-labs/avalanchego/vms/registry=VMRegistry=vms/registry/mock_vm_registry.go
 github.com/ava-labs/avalanchego/vms=Factory,Manager=vms/mock_manager.go
+github.com/ava-labs/avalanchego/wallet/subnet/primary/common=ChainUTXOs=wallet/subnet/primary/common/mock_chain_utxos.go
```
abi87, line 43: Note: I tried to avoid introducing mocks in the wallet but currently could not.
As it is now, P-chain builder and signer testing runs into circular dependencies issues if we try and do it without mocks.
This is something we should try and improve, if we want the wallet to be the one way to build transactions, for users, in e2e tests and in `platformvm` and `avm` package UTs as well.
We'll get there in time

StephenButtolph:
Will need to rebase on top of: #2740 and #2741 when they are in master.

scripts/mocks.mockgen.txt
StephenButtolph: I removed the mocks

StephenButtolph (approved):

wallet/chain/p/builder_test.go
```diff
@@ -0,0 +1,779 @@
...
+	"github.com/ava-labs/avalanchego/utils/units"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/platformvm/reward"
+	"github.com/ava-labs/avalanchego/vms/platformvm/signer"
+	"github.com/ava-labs/avalanchego/vms/platformvm/stakeable"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+
+	stdcontext "context"
```
dhrubabasu, line 27: Does this need to be aliased?

StephenButtolph: yes

dhrubabasu (approved):

PR #2735, opened 2024-02-13
Verify BLS signature provided in Handshake messages

StephenButtolph:
## Why this should be merged

This enforces that nodes are signing their IPs with their BLS keys after the activation of Durango _on mainnet_. This enforces the mainnet upgrade time rather than the currently running network to provide Fuji operators a time period to update their nodes since the Fuji Durango upgrade has already happened.

If a node provides a BLS signature. It will be verified regardless of the Durango upgrade time.

## How this works

- Optionally parses BLS signatures from the handshake message.
- Verifies provided BLS signatures against validator BLS public keys.
- Requires BLS signatures to be provided by validators with BLS public keys after the mainnet Durango activation.

## How this was tested

- [x] Additional unit tests
- [X] CI


network/peer/peer.go
```diff
@@ -728,6 +732,59 @@ func (p *peer) sendNetworkMessages() {
...
+			zap.Stringer("peerVersion", p.version),
+			zap.Error(err),
+		)
+		return true
+	}
+
+	// Enforce that all validators that have registered a BLS key are signing
+	// their IP with it after the activation of Durango.
+	vdr, ok := p.Validators.GetValidator(constants.PrimaryNetworkID, p.id)
+	if !ok || vdr.PublicKey == nil || vdr.TxID == p.txIDOfVerifiedBLSKey {
```
patrick-ogrady, line 749: Should we trigger this verification when a peer enters the validator set?

aaronbuchwald (approved):
LGTM - agree with @patrick-ogrady 's comment https://github.com/ava-labs/avalanchego/pull/2735/files#r1488716828 on verifying the signature when a validator is added as opposed to periodically when sending ping messages.

network/peer/peer.go
StephenButtolph: Put a comment

network/peer/peer.go
```diff
@@ -728,6 +732,68 @@ func (p *peer) sendNetworkMessages() {
...
+	if postDurango && p.ip.BLSSignature == nil {
+		p.Log.Debug("disconnecting from peer",
+			zap.String("reason", "missing BLS signature"),
+			zap.Stringer("nodeID", p.id),
+		)
+		return true
+	}
+
+	// If Durango hasn't activated on mainnet yet, we don't require BLS
+	// signatures to be provided. However, if they are provided, verify that
```
patrick-ogrady, line 772: ❤️

patrick-ogrady (approved):

PR #2737, opened 2024-02-15
Add AppError to Sender interface

joshua-kim:
## Why this should be merged

Adds `SendAppError*` functions to the sender interface. Handling for this message already exists but we didn't expose a way for a VM to send these messages previously.

## How this works

Exposes a new function to send app errors

## How this was tested

Tested manually locally
```
[02-15|13:12:20.637] DEBUG <X Chain> gossip/gossip.go:172 failed gossip request {"nodeID": "NodeID-MFrZFVCXPv5iCn6M9K6XduxGTYp891xXZ", "error": "-123: random error"}
```


aaronbuchwald (approved):
LGTM - first thought was we should add tests for this, but since there are currently no tests for the sender path, I think it's better not to couple that to this PR for now.

StephenButtolph (approved):

PR #2740, opened 2024-02-15
Minimize Signer interface and document Sign

StephenButtolph:
## Why this should be merged

1. There is really no reason to have `SignUnsigned` on the interface, it just makes it more annoying to implement.
2. Wanted to document this function better because it seems like partial signing is unexpected for some users: ref: #2734

## How this works

Moves `SignUnsigned` into a static function rather than on the struct.

## How this was tested

N/A

### DIFF
--- a/wallet/chain/c/signer.go
+++ b/wallet/chain/c/signer.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/ava-labs/coreth/plugin/evm"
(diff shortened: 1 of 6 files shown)


StephenButtolph:
Wanted to document this function better because it seems like partial signing is unexpected for some users: ref: #2734

dhrubabasu (approved):

aaronbuchwald (approved):

darioush (approved):

PR #2741, opened 2024-02-15
Move UTXOs definition from primary to primary/common

StephenButtolph:
## Why this should be merged

This allows chain unit tests to utilize the standard UTXO definition without a circular dependency.

## How this works

Moves the `UTXOs` definition into the common package.

## How this was tested

N/A


aaronbuchwald (approved):
LGTM

PR #2745, opened 2024-02-16
add chainID to "consensus starting" log

darioush:
## Why this should be merged
Seems it would be helpful to have this information in the logs

## How this works
Adds a field to the logger

## How this was tested
N/A


snow/engine/snowman/transitive.go
```diff
@@ -542,6 +542,7 @@ func (t *Transitive) Start(ctx context.Context, startReqID uint32) error {
 
 	t.Ctx.Log.Info("consensus starting",
 		zap.Stringer("lastAcceptedBlock", lastAcceptedID),
+		zap.Stringer("chainID", t.Ctx.ChainID),
```
StephenButtolph, line 545: The ChainID is already included in all chain loggers implicitly

PR #2747, opened 2024-02-19
X-chain wallet fees UTs

abi87:
## Why this should be merged
Even before introduction of dynamic fees over the X-chain, we can really use some unit tests that make sure that X-chain wallet spends enough to pay transactions fees and no more.
Similar to what we did in https://github.com/ava-labs/avalanchego/pull/2734

## How this works
Added UTs

## How this was tested
Just added tests.

### DIFF
--- a/.github/workflows/build-linux-binaries.yml
+++ b/.github/workflows/build-linux-binaries.yml
@@ -82,7 +82,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v3
         with:
           go-version: ${{ env.go_version }}
           check-latest: true
--- a/.github/workflows/build-ubuntu-arm64-release.yml
+++ b/.github/workflows/build-ubuntu-arm64-release.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v3
         with:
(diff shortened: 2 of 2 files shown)


wallet/chain/x/builder_test.go
```diff
@@ -0,0 +1,583 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package x
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/slices"
```
dhrubabasu, line 10: We should be able to use the stdlib slices pkg here

abi87: Thanks for spotting this. Fixed!

wallet/chain/x/builder_test.go
```diff
@@ -0,0 +1,583 @@
...
+	utxos, err := c.ChainUTXOs.UTXOs(ctx, sourceChainID)
+	if err != nil {
+		return nil, err
+	}
+
+	slices.SortFunc(utxos, func(a, b *avax.UTXO) int {
+		return a.Compare(&b.UTXOID)
+	})
+	return utxos, nil
+}
```
StephenButtolph, line 583: Should we move these into the `common` package (or perhaps a testutil package for the wallet) rather than duplicating them?

StephenButtolph:
one comment about duplicate code, other than that lgtm

StephenButtolph (approved):

dhrubabasu (approved):

PR #2751, opened 2024-02-19
[vms/platformvm] Use `wallet` sdk in `txstest.Builder`

abi87:
## Why this should be merged
Drops txBuilders implementation in platformvm package and uses wallet logic instead.
This is the minimal change required to have a single way to generate P-chain transactions, using the recommended wallet, even in platformvm unit tests.

## How this works

- Replace txBuilders implementations with wallet backends logic.
  - This is the main responsible of the PR length. We simply drop current txBuilders logic and use wallet one instead
- ~~Minimally update unit tests to account for differences~~
- Drop utxos.Handler methods (unused)

## How this was tested
CI


vms/platformvm/txs/builder/builder.go
```diff
@@ -864,33 +677,24 @@ func (b *builder) NewTransferSubnetOwnershipTx(
 	changeAddr ids.ShortID,
 	memo []byte,
 ) (*txs.Tx, error) {
-	ins, outs, _, signers, err := b.Spend(b.state, keys, 0, b.cfg.TxFee, changeAddr)
-	if err != nil {
-		return nil, fmt.Errorf("couldn't generate tx inputs/outputs: %w", err)
+	pBuilder, pSigner := b.builders(keys)
+
+	utils.Sort(ownerAddrs) // sort control addresses
```
abi87, line 682: strictly speaking we were not sorting in `NewTransferSubnetOwnershipTx`. But we do it for `NewCreateSubnetTx, so we better be consistent

dhrubabasu:
Now that https://github.com/ava-labs/avalanchego/pull/2761 is merged, I think we should remove the `txs.Builder` struct entirely and move it to the wallet

vms/platformvm/txs/builder/builder.go
```diff
@@ -294,93 +276,24 @@ func (b *builder) NewImportTx(
 	changeAddr ids.ShortID,
 	memo []byte,
 ) (*txs.Tx, error) {
-	kc := secp256k1fx.NewKeychain(keys...)
```
abi87, line 297: This is the main change of the PR. The logic in builder is simply dropped and the wallet builder and signer are used instead.

vms/platformvm/utxo/handler.go
```diff
@@ -1,671 +0,0 @@
...
-		subnetID ids.ID,
-		keys []*secp256k1.PrivateKey,
-	) (
-		verify.Verifiable, // Input that names owners
-		[]*secp256k1.PrivateKey, // Keys that prove ownership
-		error,
-	)
-}
-
-type Verifier interface {
```
abi87, line 85: `Verifier` has been moved to `verifier.go` file

vms/platformvm/txs/builder/backend.go
```diff
@@ -26,19 +25,16 @@ var (
 )
 
 func NewBackend(
-	cfg *config.Config,
 	state state.State,
 	atomicUTXOsMan avax.AtomicUTXOManager,
 ) *Backend {
 	return &Backend{
-		cfg:            cfg,
```
abi87: I will need this for dynamic fees later on. I think I left it already to reduce diffs. It's not a big problem if we want to remove it here and re-add it later on

dhrubabasu (approved):

vms/platformvm/block/executor/proposal_block_test.go
```diff
@@ -573,6 +579,10 @@ func TestBanffProposalBlockUpdateStakers(t *testing.T) {
 						Subnet: subnetID,
 					},
 					[]*secp256k1.PrivateKey{preFundedKeys[0], preFundedKeys[1]},
+					walletcommon.WithChangeOwner(&secp256k1fx.OutputOwners{
+						Threshold: 1,
+						Addrs:     []ids.ShortID{ids.ShortEmpty},
+					}),
```
abi87, line 585: @dhrubabasu I believe we have to explicitly add the change addresses to avoid flaking here.
This UT is tricky because it tests a case where two stakers share the same end time.
The one evicted will be the one with the smallest TxID, which depends on every details of the transaction creating the staker.
If we don't declare the change address explicitly, any address from the funding keys will be picked. Since we have cases where multiple keys are specified, we have the randomness causing flaking.
The change address does not have to be empty, it can be anything fixed. I picked empty to try and signal the change address does not really matter (it's enough it does not change run by run).
I also added a comment to hopefully simplify maintenance.

vms/platformvm/txs/builder/backend.go
```diff
@@ -0,0 +1,73 @@
...
+	atomicUTXOs, _, _, err := b.atomicUTXOsMan.GetAtomicUTXOs(chainID, b.addrs, ids.ShortEmpty, ids.Empty, MaxPageSize)
+	if err != nil {
+		return nil, fmt.Errorf("problem retrieving atomic UTXOs: %w", err)
+	}
+	for _, utxo := range atomicUTXOs {
+		if utxo.InputID() == utxoID {
+			return utxo, nil
+		}
+	}
+	return nil, database.ErrNotFound
```
StephenButtolph, line 68: We shouldn't need to fetch all the UTXOs of the provided addresses here. We should just fetch the requested UTXO.

vms/platformvm/txs/builder/backend.go
```diff
@@ -0,0 +1,73 @@
...
+	state          state.State
+	atomicUTXOsMan avax.AtomicUTXOManager
+}
+
+func (b *Backend) UTXOs(_ context.Context, sourceChainID ids.ID) ([]*avax.UTXO, error) {
+	if sourceChainID == constants.PlatformChainID {
+		return avax.GetAllUTXOs(b.state, b.addrs)
+	}
+
+	atomicUTXOs, _, _, err := b.atomicUTXOsMan.GetAtomicUTXOs(sourceChainID, b.addrs, ids.ShortEmpty, ids.Empty, MaxPageSize)
```
StephenButtolph, line 50: It seems like we should be fetching all UTXOs here - not just the first `MaxPageSize` right?

vms/platformvm/txs/builder/builder.go
```diff
@@ -4,524 +4,133 @@
...
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+
+	vmsigner "github.com/ava-labs/avalanchego/vms/platformvm/signer"
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
+	walletsigner "github.com/ava-labs/avalanchego/wallet/chain/p/signer"
 )
 
 // Max number of items allowed in a page
 const MaxPageSize = 1024
```
StephenButtolph, line 28: This isn't used in this file - should it be moved into `backend.go`? (perhaps we can just fully remove it)

vms/platformvm/txs/builder/builder.go
StephenButtolph, line 1: With the exception of `MaxPageSize` - the `builder` package is only used in tests. Can we change the name of this package to something like `txstest` or something? (trying to follow the `snowtest` format).

If we specify that this is a test only package, then I think we can care a lot less about performance (like the suggested change in the `Backend` for atomic UTXO lookups).

vms/platformvm/txs/builder/builder.go
```diff
@@ -4,524 +4,133 @@
...
-	state state.State
-
-	cfg *config.Config
-	ctx *snow.Context
-	clk *mockable.Clock
-	fx  fx.Fx
+type Builder struct {
+	ctx     *snow.Context
+	cfg     *config.Config
+	backend *Backend
```
StephenButtolph, line 46: imo `backend` shouldn't be on the `Builder` struct. It should be created inside `builders()`. This will remove the weird hack we have around setting `b.backend.addrs` inside `builders()`.

wallet/chain/p/builder/builder.go
```diff
@@ -962,6 +962,13 @@ func (b *builder) spend(
 		Addrs:     []ids.ShortID{addr},
 	})
 
+	// we initialize the return values with empty slices
+	// to preserv backward compatibility of json representation
+	// for transactions with no inputs/outputs
```
StephenButtolph, line 967:
```suggestion
	// Initialize the return values with empty slices to preserve backward
	// compatibility of the json representation of transactions with no
	// inputs or outputs.
```

vms/platformvm/txs/txstest/backend.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/platformvm/fx"
+	"github.com/ava-labs/avalanchego/vms/platformvm/state"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	"github.com/ava-labs/avalanchego/wallet/chain/p/signer"
+
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
```
StephenButtolph, line 20: This doesn't seem to need to be aliased.

vms/platformvm/txs/txstest/builder.go
```diff
@@ -0,0 +1,351 @@
...
+	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/platformvm/config"
+	"github.com/ava-labs/avalanchego/vms/platformvm/state"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+
+	vmsigner "github.com/ava-labs/avalanchego/vms/platformvm/signer"
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
```
StephenButtolph, line 22: This doesn't need to be aliased afaict (feel free to leave it if we're trying to be consistent with the signer alias)

vms/platformvm/txs/txstest/context.go
```diff
@@ -0,0 +1,32 @@
...
+
+package txstest
+
+import (
+	"time"
+
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/vms/platformvm/config"
+
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
```
StephenButtolph, line 12: This doesn't need to be aliased

vms/platformvm/txs/executor/create_chain_test.go
```diff
@@ -14,12 +15,16 @@ import (
...
 	"github.com/ava-labs/avalanchego/utils/units"
-	"github.com/ava-labs/avalanchego/vms/components/avax"
 	"github.com/ava-labs/avalanchego/vms/platformvm/state"
 	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs/txstest"
 	"github.com/ava-labs/avalanchego/vms/platformvm/utxo"
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/chain/p/signer"
+
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
```
StephenButtolph, line 27: This doesn't need to be aliased

vms/platformvm/txs/executor/create_subnet_test.go
```diff
@@ -4,18 +4,22 @@
...
 	"github.com/ava-labs/avalanchego/utils/units"
-	"github.com/ava-labs/avalanchego/vms/components/avax"
 	"github.com/ava-labs/avalanchego/vms/platformvm/state"
-	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs/txstest"
 	"github.com/ava-labs/avalanchego/vms/platformvm/utxo"
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/chain/p/signer"
+
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/p/builder"
```
StephenButtolph, line 22: Doesn't need to be aliased

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -715,21 +753,23 @@ func TestApricotStandardTxExecutorAddSubnetValidator(t *testing.T) {
...
-			ids.ShortEmpty, // change addr
-			nil,
 		)
 		require.NoError(err)
 
 		// Replace a valid signature with one from keys[3]
 		sig, err := preFundedKeys[3].SignHash(hashing.ComputeHash256(tx.Unsigned.Bytes()))
 		require.NoError(err)
-		copy(tx.Creds[1].(*secp256k1fx.Credential).Sigs[0][:], sig)
+		copy(tx.Creds[0].(*secp256k1fx.Credential).Sigs[0][:], sig)
```
StephenButtolph, line 772: Why did this change?

vms/platformvm/vm_test.go
```diff
@@ -2244,7 +2341,7 @@ func TestBaseTx(t *testing.T) {
 	}
 	require.Equal(totalOutputAmt, key0OutputAmt+key1OutputAmt+changeAddrOutputAmt)
 
-	require.Equal(vm.TxFee, totalInputAmt-totalOutputAmt)
+	require.Equal(vm.Config.TxFee, totalInputAmt-totalOutputAmt) // wallet inflates baseTx fee
```
StephenButtolph, line 2344: What is this comment trying to say? afaict the wallet pays the minimum fee here... Is it trying to say something else?

dhrubabasu: reverted- this test doesn't change in this PR

vms/platformvm/txs/executor/standard_tx_executor_test.go
dhrubabasu: The old tx builder code added an extra credential that's not needed for AddSubnetValidatorTx at index 0. The new p-chain wallet code only generates the credential that's needed

StephenButtolph (approved):

PR #2753, opened 2024-02-20
Send AppErrors from p2p SDK

joshua-kim:
## Why this should be merged

Currently, if a client sends an apprequest to a peer it times out if an error condition occurs. This PR introduces the ability for the p2p SDK to send back errors to unblock the client gracefully upon error.

## How this works

Sends back an error from the app handler if a non-nil `AppError` is returned.

## How this was tested

Added UT


network/p2p/error.go
```diff
@@ -0,0 +1,33 @@
...
+	ErrNotValidator = &common.AppError{
+		Code:    -3,
+		Message: "not a validator",
+	}
+	// ErrThrottled should be used to indicate that a request failed due to the
+	// requesting peer exceeding a rate limit
+	ErrThrottled = &common.AppError{
+		Code:    -4,
+		Message: "throttled",
+	}
```
joshua-kim, line 32: Another idea instead of having a separate error code per error type might be to have an error code for a class of errors, analagous to 4xx/5xx errors:

1. Transient client side errors that a client should retry (e.g throttling).
2. Persistent client side errors that a client should not retry (e.g bad request)
3. Transient server side errors that a client should retry (e.g load shed)
4. Persistent server side errors that a client should not retry (e.g internal error)

If we went with this route we could specify the error in the error text, so an error would look like:
```go
AppError{
    Code: -2 // non-retryable client error
    Message: "this request handler is not supported"
}
```


Pros: Don't add an error code for each new error, less error codes we have to maintain.
Cons: Potentially less flexibility to the client. We also are shoe-horned into having each error fit one of these buckets which will be awkward if we run into an unforeseen error class. We can alternatively just have a new error code per error type and expect the client to determine how to handle it best.

network/p2p/error.go
```diff
@@ -0,0 +1,33 @@
...
+
+package p2p
+
+import "github.com/ava-labs/avalanchego/snow/engine/common"
+
+var (
+	// ErrUnexpected should be used to indicate that a request failed due to a
+	// generic error
+	ErrUnexpected = &common.AppError{
+		Code:    -1,
```
joshua-kim, line 12: Not sure how to handle this - we decided to reserve negative error codes for our use + positive error codes for developer use. This is a weird intersection since the p2p package is technically a vm implementation detail.

network/p2p/router.go
```diff
@@ -128,14 +128,15 @@ func (r *router) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID ui
 	start := time.Now()
 	parsedMsg, handler, handlerID, ok := r.parse(request)
 	if !ok {
-		r.log.Debug("failed to process message",
+		r.log.Debug("received message for unregistered handler",
```
StephenButtolph, line 131: On a quick look - this is a bit confusing. It feels like failing to parse a message is separable from using an unregistered handler...

On closer inspection `parse` is really just parsing the handlerID... Is it worth commenting that providing an invalid handlerID (aka, empty bytes or overflowing `uint64`) is treated the same as specifying a handler that wasn't registered?

joshua-kim: Yeah maybe we could rename this function... I can leave a comment on this case.

StephenButtolph (approved):

abi87 (approved):

network/p2p/throttler_handler.go
```diff
@@ -46,9 +42,9 @@ func (t ThrottlerHandler) AppGossip(ctx context.Context, nodeID ids.NodeID, goss
 	t.handler.AppGossip(ctx, nodeID, gossipBytes)
 }
 
-func (t ThrottlerHandler) AppRequest(ctx context.Context, nodeID ids.NodeID, deadline time.Time, requestBytes []byte) ([]byte, error) {
+func (t ThrottlerHandler) AppRequest(ctx context.Context, nodeID ids.NodeID, deadline time.Time, requestBytes []byte) ([]byte, *common.AppError) {
 	if !t.throttler.Handle(nodeID) {
-		return nil, fmt.Errorf("dropping message from %s: %w", nodeID, ErrThrottled)
+		return nil, ErrThrottled
```
tsachiherman, line 47: the nodeID would now get lost; I can't really say if that's a valuable information or not.

network/p2p/handler.go
```diff
@@ -33,13 +30,14 @@ type Handler interface {
...
-	// Returns the bytes for the response corresponding to [requestBytes]
+	// Sends a response with the response corresponding to [requestBytes] or
+	// an application-defined error.
 	AppRequest(
 		ctx context.Context,
 		nodeID ids.NodeID,
 		deadline time.Time,
 		requestBytes []byte,
-	) ([]byte, error)
+	) ([]byte, *common.AppError)
```
darioush, line 40: should we do returns an error code and error (which can be used to get the string?) so like:
```suggestion
	) ([]byte, int, error)
```
or we can make the int a named type to signal some codes (instead of the *common.AppError structs)

joshua-kim: I used the `AppError` struct since it implements the `Error` interface which is handy because the code handling the response can handle the error using `errors.Is`/other go code that handle the error interface.

network/p2p/throttler_handler.go
joshua-kim: We can add context to this error

PR #2756, opened 2024-02-21
Implement interval tree to replace bootstrapping jobs queue

StephenButtolph:
## Why this should be merged

Once this is used in snowman bootstrapping, it should _significantly_ reduce DB IO compared to the legacy job queue (that supports arbitrary DAGs).

## How this works

- Implements an interval tree backed by a database.
  - It is more performant to modify intervals by changing lower heights rather than higher heights as we fetch blocks towards genesis and execute blocks towards tip.
- Implements a function that persists blocks in addition to the interval tree
- Leaves database commitment atomicity up to the provider of `database.Database`.

## How this was tested

- [X] Unit tests
- [x] Bootstrap fuji from scratch (See #2855)


snow/engine/snowman/bootstrap/interval/blocks.go
```diff
@@ -0,0 +1,114 @@
...
+	// TODO: Add metrics
+	for it.Next() {
+		blkBytes := it.Value()
+		blk, err := parser.ParseBlock(ctx, blkBytes)
+		if err != nil {
+			return err
+		}
+
+		height := blk.Height()
+		if err := DeleteBlock(tree.db, height); err != nil {
```
patrick-ogrady, line 94: Does it make sense to delete this only after we've accepted (and persisted) the block elsewhere? Or is the idea that we'd re-fetching the missing block on restart if we exited between `DeleteBlock` and `Verify/Accept`?

snow/engine/snowman/bootstrap/interval/tree.go
```diff
@@ -0,0 +1,168 @@
...
+			UpperBound: height,
+		}
+		higher *Interval
+	)
+	t.knownHeights.AscendGreaterOrEqual(newInterval, func(item *Interval) bool {
+		higher = item
+		return false
+	})
+	if !higher.Contains(height) {
+		// height isn't in the tree
```
patrick-ogrady, line 109: Does it make sense to have a metric in these cases?

snow/engine/snowman/bootstrap/interval/state.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func PutInterval(db database.KeyValueWriter, upperBound uint64, lowerBound uint64) error {
+	rangeKey := database.PackUInt64(upperBound)
+	return database.PutUInt64(
+		db,
+		append(rangePrefix, rangeKey...),
+		lowerBound,
+	)
+}
```
StephenButtolph, line 64: We should document that the key is the `upperBound` because it is assumed that we typically fetch towards genesis. So, the `lowerBound` will be changed more frequently than the `upperBound`. This allows us to just overwrite the range rather than needing to delete the prior range and then writing the new range.

snow/engine/snowman/bootstrap/interval/state.go
```diff
@@ -0,0 +1,93 @@
...
+
+func PutBlock(db database.KeyValueWriter, height uint64, bytes []byte) error {
+	blockKey := database.PackUInt64(height)
+	return db.Put(
+		append(blockPrefix, blockKey...),
+		bytes,
+	)
+}
```
StephenButtolph, line 86: We should document that the height is encoded so that we can use a database iterator that want range from lower height to higher height.

snow/engine/snowman/bootstrap/interval/tree.go
```diff
@@ -0,0 +1,166 @@
...
+		// height is already in the tree
+		return nil
+	}
+
+	t.knownHeights.DescendLessOrEqual(newInterval, func(item *Interval) bool {
+		lower = item
+		return false
+	})
+
+	t.numKnownHeights++
```
StephenButtolph, line 62: I suppose this technically can overflow.

snow/engine/snowman/bootstrap/interval/blocks.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/utils/timer"
+)
+
+const (
+	batchWritePeriod      = 64
```
StephenButtolph, line 22: Should this be increased?

snow/engine/snowman/bootstrap/interval/blocks.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Execute all the blocks tracked by the tree. If a block is in the tree but is
+// already accepted based on the lastAcceptedHeight, it will be removed from the
+// tree but not executed.
+//
+// Execute assumes that GetMissingBlockIDs would return an empty set.
+func Execute(
+	ctx context.Context,
+	log logging.Func,
```
StephenButtolph, line 106: Taking in the log function rather than the full logger allows the caller to specify the log level.

snow/engine/snowman/bootstrap/interval/blocks.go
StephenButtolph: We'd re-fetch any missing blocks. Additionally, this is now written into a batch, so the opposite can happen (when the block is left on disk even though it was already executed).

snow/engine/snowman/bootstrap/interval/blocks.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// GetMissingBlockIDs returns the ID of the blocks that should be fetched to
+// attempt to make a single continuous range from
+// (lastAcceptedHeight, highestTrackedHeight].
+//
+// For example, if the tree currently contains heights [1, 4, 6, 7] and the
+// lastAcceptedHeight is 2, this function will return the IDs corresponding to
+// blocks [3, 5].
+func GetMissingBlockIDs(
```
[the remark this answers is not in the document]
StephenButtolph: I had forgotten about the `block.Parser` interface. Removed the redundancy.

Will think about the location of `GetMissingBlockIDs` and `Execute`.

joshua-kim (approved):

snow/engine/snowman/bootstrap/interval/blocks.go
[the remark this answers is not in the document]
StephenButtolph: moved these into the followup PR so that they could be implemented next to `process`.

snow/engine/snowman/bootstrap/interval/blocks_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/avalanchego/utils"
+)
+
+func TestAdd(t *testing.T) {
```
joshua-kim, line 16: I think it's possible for us to refactor this + remove cases into a table-style test?

snow/engine/snowman/bootstrap/interval/blocks_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+		_, err = GetBlock(db, 5)
+		require.ErrorIs(err, database.ErrNotFound)
+	})
+}
+
+func generateBlockchain(length uint64) [][]byte {
+	blocks := make([][]byte, length)
+	for i := range blocks {
+		blocks[i] = utils.RandomBytes(1024)
```
joshua-kim, line 179: I don't know what people think about this but I'm usually on team no randomness in unit tests.... feel like it drives me crazy since I sometimes jot-down the value of variables when I'm debugging them. Can we just make each block just have the `i` byte or something?

snow/engine/snowman/bootstrap/interval/tree_test.go
```diff
@@ -0,0 +1,389 @@
...
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/memdb"
+)
+
+func newTree(require *require.Assertions, db database.Database, intervals []*Interval) *Tree {
```
joshua-kim, line 16: nit: We could add a variadic `intervals Interval...` parameter at the end of `interval.NewTree` to avoid this extra testing function

StephenButtolph: I'd prefer not to introduce functionality to the production code that is only used during testing

joshua-kim (approved):

dhrubabasu (approved):

PR #2757, opened 2024-02-21
P-Chain - repackaged wallet backends

abi87:
## Why this should be merged
In the P-chain we have currently three ways to build transactions:

1. Via the wallet, the preferred way, used in e2e tests and by other downstreams
2. Via the platformvm txbuilders, used in unit tests and in some deprecated P-chain APIs
3. Laying out the whole tx "manually" in some p-chain UTs

There is really no reason why we need these difference methods. Moreover dynamic fees will deeply affect the way we build the txs and it would be a nightmare to maintain multiple ways to build txs.
This PR is the first of two steps to consolidate mode 1 (wallet) and mode 2 (txbuilder) into a single building mode.
And we'll spare some 400 LOC too, see https://github.com/ava-labs/avalanchego/pull/2751

## How this works
Step 1: just repackage wallet backends interfaces out of p package into their own backends package, along with the logic to finance and sign a transactions. This will allow us to reuse all of this logic in platformvm package by just implementing the backends (done in https://github.com/ava-labs/avalanchego/pull/2751)

## How this was tested
Just repackaging stuff, CI


wallet/chain/p/backends/backends.go
```diff
@@ -0,0 +1,35 @@
...
+	GetSubnetOwner(ctx stdcontext.Context, subnetID ids.ID) (fx.Owner, error)
+	GetUTXO(ctx stdcontext.Context, chainID, utxoID ids.ID) (*avax.UTXO, error)
+}
+
+type SignerBackend interface {
+	GetUTXO(ctx stdcontext.Context, chainID, utxoID ids.ID) (*avax.UTXO, error)
+	GetSubnetOwner(ctx stdcontext.Context, subnetID ids.ID) (fx.Owner, error)
+}
+
+type BuilderBackend interface {
```
abi87, line 31: This is the main point of the repackaging: we want to implement these backends for the platformvm txbuilder subpackage, so that we can build P-chain transactions only using the wallet. However we cannot do that if these backends are in their original p package, due to import circles.

wallet/chain/p/backends/builder.go
```diff
@@ -256,14 +253,6 @@ type Builder interface {
...
 }
 
-// BuilderBackend specifies the required information needed to build unsigned
-// P-chain transactions.
-type BuilderBackend interface {
-	Context
-	UTXOs(ctx stdcontext.Context, sourceChainID ids.ID) ([]*avax.UTXO, error)
-	GetSubnetOwner(ctx stdcontext.Context, subnetID ids.ID) (fx.Owner, error)
-}
-
```
abi87, line 266: just moved to backends.go file

wallet/chain/p/backends/signer.go
```diff
@@ -27,25 +24,20 @@ type Signer interface {
 	Sign(ctx stdcontext.Context, tx *txs.Tx) error
 }
 
-type SignerBackend interface {
-	GetUTXO(ctx stdcontext.Context, chainID, utxoID ids.ID) (*avax.UTXO, error)
-	GetSubnetOwner(ctx stdcontext.Context, subnetID ids.ID) (fx.Owner, error)
-}
-
```
abi87, line 34: just moved to backends.go file

wallet/chain/p/context.go
```diff
@@ -5,46 +5,13 @@ package p
...
-	avaxAssetID                   ids.ID
-	baseTxFee                     uint64
-	createSubnetTxFee             uint64
-	transformSubnetTxFee          uint64
-	createBlockchainTxFee         uint64
-	addPrimaryNetworkValidatorFee uint64
-	addPrimaryNetworkDelegatorFee uint64
-	addSubnetValidatorFee         uint64
-	addSubnetDelegatorFee         uint64
-}
```
abi87, line 45: moved to backends package

StephenButtolph (approved):

PR #2759, opened 2024-02-21
Remove Deprecated Auth API

StephenButtolph:
## Why this should be merged

The Auth API was deprecated as the HTTP server should not be accessible by unauthorized connections. A traditional load balancer should be used to authenticate inbound connections before being routed to a node.

## How this works

Deletes `api/auth`.

## How this was tested

- [X] CI


go.mod
```diff
@@ -17,7 +17,6 @@ require (
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.1.0
 	github.com/ethereum/go-ethereum v1.12.0
-	github.com/golang-jwt/jwt/v4 v4.3.0
```
dhrubabasu, line 20: ❤️

dhrubabasu (approved):

PR #2761, opened 2024-02-21
`vms/platformvm`: Remove all keystore APIs except `ExportKey` and `ListAddresses`

dhrubabasu:
## Why this should be merged

These functions should not be used as keys should not be stored on the node. We'll retain `ExportKey` and `ListAddresses` to enable people to export their keys if they have not done so already.

## How this works

Deletes all code relating to keystore APIs in the `platformvm` except `ExportKey` and `ListAddresses`.

## How this was tested

- [x] CI


dhrubabasu:
We should remove `txBuilder` from the `VM` struct in a follow-up. It is now only used in UTs and should be removed from production code.

vms/platformvm/service.go
StephenButtolph, line 1: `CreateAddress` should be removed

StephenButtolph (commented):
1 more - then lgtm

StephenButtolph (approved):

PR #2765, opened 2024-02-23
Refactor p-chain edge-cases out of chain manager

joshua-kim:
## Why this should be merged

Last piece of the puzzle to remove `vm.Initialize`

## How this works

Refactors all p-chain edge-cases out of the chains/manager and into the parameters of `chains.Factory#New`.

## How this was tested

CI

### DIFF
--- a/vms/platformvm/config/config.go
+++ b/vms/platformvm/config/config.go
@@ -10,10 +10,9 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/uptime"
 	"github.com/ava-labs/avalanchego/snow/validators"
-	"github.com/ava-labs/avalanchego/utils/constants"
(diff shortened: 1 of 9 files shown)


vms/platformvm/config/config.go
```diff
@@ -150,23 +153,3 @@ func (c *Config) GetCreateSubnetTxFee(timestamp time.Time) uint64 {
 	}
 	return c.CreateAssetTxFee
 }
-
-// Create the blockchain described in [tx], but only if this node is a member of
-// the subnet that validates the chain
-func (c *Config) CreateChain(chainID ids.ID, tx *txs.CreateChainTx) {
```
joshua-kim, line 156: We have to refactor this into the VM because:
1. It makes no sense that chain creation is even created from the config in the first place
2. Now the P-Chain is responsible for supplying itself into any chains it's spinning up, so this must live in the VM

PR #2771, opened 2024-02-26
P-chain - Technical docs update

abi87:
## Why this should be merged
We have some technical documentation stored in `platformvm` package, which was not visible enough, nor updated. This PR start to tackle these issues.

## How this works
Refreshed existing doc, drop outdated one and added a README at platformvm package level for visibility.

## How this was tested
a human reading the doc


vms/platformvm/docs/mempool_gossiping.md
```diff
@@ -1,16 +0,0 @@
-# Mempool Gossiping
```
abi87, line 1: gossiping has been consolidated into an sdk, shared across VMs. I feel we should document its use in the sdk.

vms/platformvm/README.md
```diff
@@ -0,0 +1,21 @@
+# Platform VM
+
+`PlatformVM` is the virtual machine implementing the [P-chain](https://support.avax.network/en/articles/4058243-what-is-the-platform-chain-p-chain) in `avalanchego`.
```
aaronbuchwald, line 3: Could we avoid linking externally? I think it would be better if AvalancheGo was a source of truth rather than referencing the support page.

Generally, I think any term that deserves its own support or medium article deserves to be defined within the internal documentation of AvalancheGo.

vms/platformvm/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+As such `PlatformVM` is responsible for:
+
+- Tracking permissioned and [elastic](https://medium.com/avalancheavax/banff-elastic-subnets-44042f41e34c) subnets, including the Primary Network subnet, which is defined in P-chain genesis.
```
aaronbuchwald, line 7: Could we capitalize Subnet and refer to the Primary Network as just the Primary Network instead of Primary Network Subnet?

vms/platformvm/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Tracking every chains associated with each subnets.
```
aaronbuchwald, line 8:
```suggestion
- Tracking every chain associated with each subnet
```

vms/platformvm/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Technical content
+
+Here is a list of technical details about the way `PlatformVM` works:
+
+- `ChainTime` helps tracking stakers' uptime and rewarding them. You can read [here](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/docs/chain_time.md) details of how P-chain mananges it.
+- Validators are versioned to support the [Warp protocol](https://medium.com/avalancheavax/avalanche-warp-messaging-awm-launches-with-the-first-native-subnet-to-subnet-message-on-avalanche-c0ceec32144a). You can read [here](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/docs/validators_versioning.md) details of how the P-chain stores validator set versions.
+- Subnets lifetime is tracked and managed in the P-chain. You can read [here](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/docs/subnets.md) details of what transactions can be used to affect a subnet.
+- The P-chain can process different kind of blocks. You can read [here](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/docs/block_formation_logic.md) details of how blocks have changed across forks.
```
aaronbuchwald, line 21: Could we use relative links here?

vms/platformvm/docs/mempool_gossiping.md
aaronbuchwald: SGTM, could we link to the gossip package (even if there is no README there at this time) from the technical content section in the README?

aaronbuchwald (commented):
Thanks for putting this up!

I wonder what the best way to document the various transaction types on the P-Chain is. It seems like this is the last section that is still distributed throughout the rest of the docs making it a bit harder to follow. Do you think it makes sense to add one more doc enumerating each of the transaction types?

vms/platformvm/README.md
abi87: Done

vms/platformvm/README.md
abi87: Done

abi87:
> Thanks for putting this up!
> 
> I wonder what the best way to document the various transaction types on the P-Chain is. It seems like this is the last section that is still distributed throughout the rest of the docs making it a bit harder to follow. Do you think it makes sense to add one more doc enumerating each of the transaction types?

We do have the list in our [officlal docs](https://github.com/ava-labs/avalanchego/pull/2771/files#:~:text=avax.network/-,reference,-/avalanchego/p%2Dchain). I have mentioned the link in the doc

meaghanfitzgerald:
@abi87 I plan on migrating all the `api.md` docs directly to the Avalanchego repo, and pulling into the docs, similar to how 
the [this page](https://docs.avax.network/build/cross-chain/awm/relayer) is pulled directly from the [AWM Relayer repo](https://github.com/ava-labs/awm-relayer?#awm-relayer). 

for example, i would move https://github.com/ava-labs/avalanche-docs/blob/master/docs/reference/avalanchego/p-chain/api.md to https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/ , and rename the file `service.md` (per @StephenButtolph 's request). 

should we do the same for the transaction format file?

PR #2773, opened 2024-02-26
Gossip cleanup suggestions

patrick-ogrady:


patrick-ogrady:
We should also be including `PushGossiper` here: https://github.com/ava-labs/avalanchego/pull/2773/files#diff-7183c682b28d9ef4699f68c9c337b2a412b6fbac74b98e5784ca32b0cc6d8203L181

PR #2774, opened 2024-02-26
[network/p2p] Integrate new gossip into P-Chain

patrick-ogrady:

### DIFF
--- a/vms/platformvm/network/network.go
+++ b/vms/platformvm/network/network.go
@@ -5,13 +5,11 @@ package network
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
-	"github.com/ava-labs/avalanchego/cache"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/network/p2p"
 	"github.com/ava-labs/avalanchego/network/p2p/gossip"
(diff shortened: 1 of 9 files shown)


vms/platformvm/network/network.go
```diff
@@ -223,35 +217,18 @@ func (n *network) AppGossip(ctx context.Context, nodeID ids.NodeID, msgBytes []b
 		)
 		return nil
 	}
-	txID := tx.ID()
-
-	if err := n.issueTx(tx); err == nil {
-		n.legacyGossipTx(ctx, txID, msgBytes)
 
-		n.txPushGossiper.Add(tx)
```
patrick-ogrady, line 231: We should not enqueue push gossip from inbound p2p.

vms/platformvm/vm.go
```diff
@@ -219,7 +219,8 @@ func (vm *VM) Initialize(
 	vm.onShutdownCtx, vm.onShutdownCtxCancel = context.WithCancel(context.Background())
 	// TODO: Wait for this goroutine to exit during Shutdown once the platformvm
 	// has better control of the context lock.
-	go vm.Network.Gossip(vm.onShutdownCtx)
+	go vm.Network.PushGossip(vm.onShutdownCtx)
```
patrick-ogrady, line 222: Seems like these should be combined? Considered adding support for this in `gossip.Every` but got weird because frequency differed by `gossiper`.

PR #2776, opened 2024-02-28
Update README.md

eino-makitalo:
Add note about missing volume in docker run

## Why this should be merged
Add some extra information

## How this works
Using docker image instead of overlay2 which is overall better is running container long time

## How this was tested
N/A


StephenButtolph:
I think if we wanted the readme to actually document the requirements of running nodes with docker, we would probably need to have a much better description than we do currently.

PR #2780, opened 2024-02-28
`merkledb` comment accuracy fixes

danlaine:
Assorted fixes to inaccurate comments in merkledb

### DIFF
--- a/x/merkledb/proof.go
+++ b/x/merkledb/proof.go
@@ -735,7 +735,7 @@ func verifyProofPath(proof []ProofNode, key maybe.Maybe[Key]) error {
 		nodeKey := currentProofNode.Key
 
 		// Because the interface only support []byte keys,
-		// a key with a partial byte should store a value
+		// a key with a partial byte may not store a value
 		if nodeKey.hasPartialByte() && proof[i].ValueOrHash.HasValue() {
 			return ErrPartialByteLengthWithValue
 		}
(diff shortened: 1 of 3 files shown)


x/merkledb/trie.go
```diff
@@ -92,11 +92,10 @@ type View interface {
 	CommitToDB(ctx context.Context) error
 }
 
-// Returns the nodes along the path to [key].
+// Calls [visitNode] on the nodes along the path to [key].
 // The first node is the root, and the last node is either the node with the
 // given [key], if it's in the trie, or the node with the largest prefix of
 // the [key] if it isn't in the trie.
-// Always returns at least the root node.
```
abi87, line 99: why was this dropped?

JacobEverly (approved):

x/merkledb/proof.go
```diff
@@ -735,7 +735,7 @@ func verifyProofPath(proof []ProofNode, key maybe.Maybe[Key]) error {
 		nodeKey := currentProofNode.Key
 
 		// Because the interface only support []byte keys,
```
darioush, line 737:
```suggestion
		// Because the interface only supports []byte keys,
```

darioush (approved):

x/merkledb/trie.go
danlaine: It used to be true, but it isn't true now. If the root is not a prefix of the key we're looking for, the root will not be visited.

PR #2785, opened 2024-02-29
Move AppGossip configs from SubnetConfig into ChainConfig

StephenButtolph:
## Why this should be merged

Currently `AppGossip` parameters are set on a per-subnet basis. This is very annoying, as different gossip mechanisms even within the same chain may desire to be configured differently.

## How this works

Moves `AppGossip` parameters from the subnet config to chain configs.

## How this was tested

- [X] CI


vms/avm/network/config.go
```diff
@@ -12,6 +12,9 @@ import (
 var DefaultConfig = Config{
 	MaxValidatorSetStaleness:                    time.Minute,
 	TargetGossipSize:                            20 * units.KiB,
+	PushGossipNumValidators:                     10,
```
patrick-ogrady, line 15: Wondering if it makes sense to just expose `PushGossipNumValidators` for now? I guess the more configs the merrier (just might get to be a lot with the regossip ones too).

patrick-ogrady: I think I like how it is but just offering a nit because everything else looks right.

StephenButtolph: Decided to remove `PushGossipNumNonValidators` because there is subtle behavior that can cause unexpected behavior when used on non-validators... In generally gossiping to non-validators should be done using the `PushGossipNumPeers` option.

I'm going to leave the `Peers` option though, in case someone wants to try to open up the mempool.

patrick-ogrady (approved):

PR #2789, opened 2024-02-29
`merkledb` proof verification refactor

danlaine:
## Why this should be merged

See comments -- there is a bug that could cause an invalid exclusion proof to pass verification.

## How this works

See comments.

## How this was tested

New unit tests.


x/merkledb/proof.go
```diff
@@ -138,31 +148,39 @@ func (proof *Proof) Verify(ctx context.Context, expectedRootID ids.ID, tokenSize
 	if len(proof.Path) == 0 {
 		return ErrEmptyProof
 	}
-
+	if proof.Key.hasPartialByte() {
```
danlaine, line 151: Enforcement of new invariant on `Key` from above.

x/merkledb/proof.go
```diff
@@ -138,31 +148,39 @@ func (proof *Proof) Verify(ctx context.Context, expectedRootID ids.ID, tokenSize
...
+		return ErrProofKeyPartialByte
+	}
 	if err := verifyProofPath(proof.Path, maybe.Some(proof.Key)); err != nil {
 		return err
 	}
 
-	// Confirm that the last proof node's value matches the claimed proof value
 	lastNode := proof.Path[len(proof.Path)-1]
+	isInclusionProof := lastNode.Key == proof.Key
```
danlaine, line 159: Usage of new invariant on `Key`.

x/merkledb/proof.go
```diff
@@ -138,31 +148,39 @@ func (proof *Proof) Verify(ctx context.Context, expectedRootID ids.ID, tokenSize
...
+			// The last node's key is a prefix of [proof.Key], but isn't equal to it,
+			// so there should be at least one unmatched token.
+			return ErrExpectedUnmatchedToken
+		}
+		nextUnmatchedToken := unmatchedKey.Token(0, tokenSize)
+		if _, ok := lastNode.Children[nextUnmatchedToken]; ok {
+			// The proof path should include all nodes which are prefixes of [proof.Key]
+			// but there's another child of the last node which is a prefix of [proof.Key].
+			// That child (or its descendant) should be the last node in the proof path.
+			return ErrUnexpectedEndOfProof
```
danlaine, line 182: This fixes a bug in the old code. Previously, it was possible for an exclusion proof to pass verification even if the key really _is_ in the trie. We were not checking that the proof path extends as far down as possible.

To generate such a proof in the old code:
1. Generate a valid inclusion proof
2. Remove the last node from `proof.Path` 
3. Set `Value` to `Nothing`

We would insert the parent of the node at the key, as well as its children IDs, and as a result verification would pass because the root ID was correct. In other words, when we're verifying an exclusion proof, we need to assert that it's impossible for the key we're proving to exist "below" the nodes in the proof.

x/merkledb/proof_test.go
```diff
@@ -21,86 +22,100 @@ import (
...
-	require.NoError(db.PutContext(ctx, []byte{0}, []byte{2}))
-
-	expectedRoot, err := db.GetMerkleRoot(ctx)
-	require.NoError(err)
-
-	proof, err := db.GetProof(ctx, []byte{})
-	require.NoError(err)
-
-	require.NoError(proof.Verify(ctx, expectedRoot, 4))
-}
```
danlaine, line 47: These are already covered in `Test_Proof_Verify`

x/merkledb/proof_test.go
```diff
@@ -21,86 +22,100 @@ import (
...
-
-func Test_Proof_Verify_Bad_Data(t *testing.T) {
```
danlaine, line 49: Renamed `Test_Proof_Verify`

x/merkledb/proof_test.go
```diff
@@ -123,9 +149,21 @@ func Test_Proof_Verify_Bad_Data(t *testing.T) {
...
+		require := require.New(t)
+
+		proof := &Proof{
+			Path: []ProofNode{{}},
+			Key:  ToKey([]byte{1}).Skip(1),
+		}
+
+		err := proof.Verify(context.Background(), ids.Empty, 4)
+		require.ErrorIs(err, ErrProofKeyPartialByte)
+	})
```
danlaine, line 163: Test new invariant on `Key`

x/merkledb/proof_test.go
```diff
@@ -123,9 +149,21 @@ func Test_Proof_Verify_Bad_Data(t *testing.T) {
...
 }
 
-func Test_Proof_ValueOrHashMatches(t *testing.T) {
+func Test_ValueOrHashMatches(t *testing.T) {
```
danlaine, line 166: Renamed because this isn't a method on `Proof`

x/merkledb/proof_test.go
```diff
@@ -1846,27 +1884,44 @@ func FuzzProofVerification(f *testing.F) {
...
-		// Delete a key-value pair so database doesn't grow unbounded
-		iter := db.NewIterator()
-		deleteKey := iter.Key()
-		iter.Release()
+		// Revert the change
+		proof.Path[changeIndex].ValueOrHash = originalValue
 
-		require.NoError(db.Delete(deleteKey))
+		// Verifying the proof against a different root should fail
+		require.Error(proof.Verify(context.Background(), ids.GenerateTestID(), db.tokenSize)) //nolint:forbidigo
```
danlaine, line 1924: Additional test coverage that would've caught the exclusion proof bug mentioned above

darioush (approved):

x/merkledb/proof.go
```diff
@@ -138,31 +148,39 @@ func (proof *Proof) Verify(ctx context.Context, expectedRootID ids.ID, tokenSize
 	if len(proof.Path) == 0 {
 		return ErrEmptyProof
 	}
-
+	if proof.Key.hasPartialByte() {
+		return ErrProofKeyPartialByte
+	}
 	if err := verifyProofPath(proof.Path, maybe.Some(proof.Key)); err != nil {
```
StephenButtolph, line 154: I know this isn't really related to this PR... But trying to verify the correctness of this function...
```golang
// Returns nil iff all the following hold:
//   - Any node with a partial byte length, should not have a value associated with it
//     since all keys with values are written in complete bytes([]byte).
//   - Each key in [proof] is a strict prefix of the following key.
//   - Each key in [proof] is a strict prefix of [keyBytes], except possibly the last.
//   - If the last element in [proof] is [Key], this is an inclusion proof.
//     Otherwise, this is an exclusion proof and [keyBytes] must not be in [proof].
func verifyProofPath(proof []ProofNode, key maybe.Maybe[Key]) error {
```

The comment is pretty weird imo. I feel like the last bullet isn't something that the function is checking. It's documenting the proof.

Should it be like this instead?
```golang
// Returns nil iff all the following hold:
//   - Any node with a partial byte length, should not have a value associated with it
//     since all keys with values are written in complete bytes([]byte).
//   - Each key in [proof] is a strict prefix of the following key.
//   - Each key in [proof] is a strict prefix of [keyBytes], except possibly the last.
//
// If the last element in [proof] is [Key], this is an inclusion proof.
// Otherwise, this is an exclusion proof and [keyBytes] must not be in [proof].
func verifyProofPath(proof []ProofNode, key maybe.Maybe[Key]) error {
```

Or am I misreading this last bullet?

I'm not really sure that I agree with the sentence as a whole either tbh

x/merkledb/proof.go
```diff
@@ -120,13 +125,18 @@ func (node *ProofNode) UnmarshalProto(pbNode *pb.ProofNode) error {
...
+
 	// This is a proof that [key] exists/doesn't exist.
+	// Must not have any partial bytes.
 	Key Key
 
-	// Nothing if [Key] isn't in the trie.
-	// Otherwise, the value corresponding to [Key].
+	// If this is an inclusion proof, this is the value associated with [key].
+	// Note this may be Nothing if the key doesn't have a value.
+	// If this is an exclusion proof, this is Nothing.
```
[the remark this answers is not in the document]
danlaine: Hmm yeah this looks weird. Removed line 138.

StephenButtolph:
This should never have been closed - as it fixes a high sev bug

rrazvan1:
fixed in #3688

PR #2790, opened 2024-02-29
Remove put gossip

StephenButtolph:
## Why this should be merged

Put gossip was replaced with periodic pull gossip in `v1.10.17`. Because nodes must update to `v1.11.x` to remain compatible, we can safely remove this legacy gossip mechanism.

This will noticeably reduce outbound bandwidth.

## How this works

Removes `Put` gossip.

## How this was tested

- [X] CI


joshua-kim (approved):

chains/subnets_test.go
```diff
@@ -117,16 +117,12 @@ func TestSubnetConfigs(t *testing.T) {
 			config: map[ids.ID]subnets.Config{
 				constants.PrimaryNetworkID: {},
 				testSubnetID: {
-					GossipConfig: subnets.GossipConfig{
-						AcceptedFrontierValidatorSize: 123456789,
-					},
+					ValidatorOnly: true,
```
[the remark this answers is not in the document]
StephenButtolph: This test wouldn't be meaningful id the subnet config was equal to the primary network config. Because we're deleting the only field that was different, we needed to introduce some other config value to be different.

message/fields.go
```diff
@@ -106,14 +105,10 @@ func GetRequestID(m any) (uint32, bool) {
...
-	// to allow a uniform handling of all messages
-	if _, ok := m.(*p2p.AppGossip); ok {
-		return constants.GossipMsgRequestID, true
-	}
-
-	return 0, false
+	// AppGossip is the only inbound message not containing a requestID. For
+	// ease of handling, imagine that it does have a requestID.
+	_, ok := m.(*p2p.AppGossip)
+	return 0, ok
```
abi87, line 111: nit: I wonder if this should be `math.MaxUint32`instead of 0? (`math.MaxUint32` is the value of `constants.GossipMsgRequestID`)

StephenButtolph: It doesn't really matter... The request ID isn't used for this message type.

PR #2795, opened 2024-03-01
[vms/platformvm] Remove `platform.getMaxStakeAmount`

dhrubabasu:
## Why this should be merged

This has been deprecated since v1.9.12. We should be able to remove it now that we are on v1.11.2.

It is also not exposed on the public RPCs.

## How this works

Remove code 🧹

## How this was tested

CI


vms/platformvm/api/static_service.go
```diff
@@ -88,10 +88,6 @@ type Staker struct {
 	EndTime   json.Uint64 `json:"endTime"`
 	Weight    json.Uint64 `json:"weight"`
 	NodeID    ids.NodeID  `json:"nodeID"`
-
-	// Deprecated: Use Weight instead
-	// TODO: remove [StakeAmount] after enough time for dependencies to update
-	StakeAmount *json.Uint64 `json:"stakeAmount,omitempty"`
```
StephenButtolph, line 94: Could we delete this in a separate PR? Seems unrelated to the removal of `platform.getMaxStakeAmount`.

StephenButtolph (approved):

abi87 (approved):

PR #2797, opened 2024-03-01
[vms/platformvm] Remove `ErrFutureStakeTime` check in `VerifyTx`

dhrubabasu:
## Why this should be merged

Post-Durango, this error will never be returned by `StandardTxExecutor` so we can remove this check.

## How this works

Delete useless code.

## How this was tested

CI


abi87 (commented):
I think we can clean up more.
Following Durango activation, we could drop `verifyStakerStartsSoon` and `MaxFutureStartTime`

StephenButtolph (approved):

JacobEverly (approved):

PR #2800, opened 2024-03-01
Remove engine type handling for everything other than GetAncestors

StephenButtolph:
## Why this should be merged

As of `v1.11.0` - the only message sent from the avalanche engine is `GetAncestors`. Every other message type can be assumed to be handled by snowman (or dropped in the case the X-chain is still fetching the DAG).

## How this works

Removes `EngineType` from all messages other than `GetAncestors`.

## How this was tested

- [X] CI
- [X] Bootstrapped Fuji


abi87 (approved):
lgtm

PR #2801, opened 2024-03-01
Remove fallback validator height indexing

StephenButtolph:
## Why this should be merged

After Durango activates, all nodes will be required to be running `v1.11.x` nodes. All `v1.11.x` nodes support the more efficient validator indexing, so we can remove the legacy indexing.

## How this works

Deletes dead code.

## How this was tested

- [X] CI
- [ ] Run a node on fuji
- [ ] Run a node on mainnet


darioush (approved):

dhrubabasu (approved):

vms/platformvm/state/state.go
```diff
@@ -1100,50 +1078,7 @@ func (s *state) ApplyValidatorWeightDiffs(
 			return err
 		}
 	}
-	if err := diffIter.Error(); err != nil {
-		return err
-	}
-
-	// TODO: Remove this once it is assumed that all subnet validators have
-	// adopted the new indexing.
-	for height := prevHeight - 1; height >= endHeight; height-- {
```
StephenButtolph, line 1109: Note: this does potentially change the values returned for heights that were processed before the new indexing existed. However - those heights are less than the current heights required by the proposervm for all chains (XPC)

abi87 (approved):

PR #2807, opened 2024-03-04
Refactor ProposerVM into `vm.Factory`

joshua-kim:
## Why this should be merged

Moves `proposervm` usage into the implementation of the VM factory as an opt-in feature.

## How this works

Removes proposervm wrapping from chain creation and instead expects vms that want to use propservm's leader election to wrap the VM in the factory.

## How this was tested

CI

### DIFF
--- a/chains/manager.go
+++ b/chains/manager.go
@@ -832,17 +832,19 @@ func (m *manager) createAvalancheChain(
 	// Create engine, bootstrapper and state-syncer in this order,
 	// to make sure start callbacks are duly initialized
 	snowmanEngineConfig := smeng.Config{
-		Ctx:                 ctx,
-		AllGetsServer:       snowGetHandler,
-		VM:                  vmWrappingProposerVM,
-		Sender:              snowmanMessageSender,
-		Validators:          vdrs,
-		ConnectedValidators: connectedValidators,
-		Params:              consensusParams,
-		Consensus:           snowmanConsensus,
+		Parameters: consensusParams,
 	}
 	var snowmanEngine common.Engine
-	snowmanEngine, err = smeng.New(snowmanEngineConfig)
+	snowmanEngine, err = smeng.New(
+		snowmanEngineConfig,
+		snowGetHandler,
+		ctx,
+		vmWrappingProposerVM,
+		snowmanMessageSender,
+		vdrs,
+		connectedValidators,
+		snowmanConsensus,
+	)
 	if err != nil {
 		return nil, fmt.Errorf("error initializing snowman engine: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


chains/manager.go
```diff
@@ -709,60 +698,29 @@ func (m *manager) createAvalancheChain(
...
-		},
-	)
-
 	if m.MeterVMEnabled {
-		vmWrappingProposerVM = metervm.NewBlockVM(vmWrappingProposerVM)
+		chainVM = metervm.NewBlockVM(chainVM)
 	}
 	if m.TracingEnabled {
-		vmWrappingProposerVM = tracedvm.NewBlockVM(vmWrappingProposerVM, "proposervm", m.Tracer)
+		chainVM = tracedvm.NewBlockVM(chainVM, "vm", m.Tracer)
```
joshua-kim, line 715: Not sure what we want this metric namespace to be

chains/manager.go
```diff
@@ -201,8 +199,7 @@ type ManagerConfig struct {
 	CriticalChains            set.Set[ids.ID] // Chains that can't exit gracefully
 	TimeoutManager            timeout.Manager // Manages request timeouts when sending messages to other validators
 	Health                    health.Registerer
-	SubnetConfigs             map[ids.ID]subnets.Config // ID -> SubnetConfig
```
ceyonur, line 204: Where did we move those?

joshua-kim: I originally thought we could remove this since we're only using the subnet configs for their proposervm configs and this PR changes the proposervm to be an opt-in feature that vm implementations wrap themselves in (for example hypersdk does not want to use the proposervm). In the next iteration we might need to keep this since the X-Chain only wants to use the proposervm once it linearizes... we can't just blindly return a wrapped AVM instance in a proposervm since proposervm is not compatible pre-linearization.

ceyonur: There are bunch of other configs in subnet configs https://github.com/ava-labs/avalanchego/pull/2807/files#diff-7a895c6b0e074cfed9bbd7a1bf256218f85323f62f5fff563b82eba35f34d58dR27-R37

AFAIK AvaCloud uses validator only config https://support.avacloud.io/will-my-network-be-private

joshua-kim: I think there might be some confusion, this PR does not remove the usage of subnet configs. The subnet configs are still used [here](https://github.com/ava-labs/avalanchego/blob/f1f88ac2775aeeb790b5463747be5850bcd78419/chains/subnets.go#L20), we just have a copy of these configs that we don't need

ceyonur: aahh sorry my bad, I wasn't aware that we already moved them. Then it's looking good 👍

PR #2808, opened 2024-03-05
P-chain:  Improve GetValidatorsSet error expressivity

abi87:
## Why this should be merged
Providing more context around GetValidatorSet errors could speed up debugging

## How this works
Just replace current `database.ErrNotFound` error, which is widely used and not very informative, with more precise error

## How this was tested
CI


vms/platformvm/validators/manager.go
```diff
@@ -247,7 +246,11 @@ func (m *manager) makePrimaryNetworkValidatorSet(
 		return nil, 0, err
 	}
 	if currentHeight < targetHeight {
-		return nil, 0, database.ErrNotFound
+		return nil, 0, fmt.Errorf("failed to fetch validator set (requested P-Chain Height: %d, SubnetID: %s, current P-chain Height: %d)",
+			targetHeight,
+			constants.PrimaryNetworkID,
+			currentHeight,
+		)
```
abi87, line 253: we use this error in one of the clients of this validator manager. Better make it available to all clients

StephenButtolph (approved):

dhrubabasu (approved):

PR #2810, opened 2024-03-05
Implement snowflake+ consensus

aaronbuchwald:
## Why this should be merged

This PR adds a new consensus implementation to the Snowball package called Snowflake+.

## How this works

Snowflake+ is a modification of Snowflake that utilizes multiple alpha threshold/beta confidence counter pairs for finalization detection.

Rather than using a single alpha/beta pair that requires beta consecutive alpha thresholds to finalize, Snowflake+ pre-calculates beta values for each possible alpha threshold in the range [k/2 + 1, k].

This provides a significant improvement in finalization detection by using more detailed information in the full voting history. For example, if Snowflake receives 10 consecutive polls K votes for blue (100% voting blue), it treats this identically to receiving AlphaConfidence votes for blue 10 times in a row, which may be less than K.

Snowflake+ implements the Consensus interface rather than the Unary/Binary/Nnary interfaces because it needs access to the mode/frequency returned by the sample. Snowflake and Snowball use either the Tree or Flat implementations to convert from an actual sample to calling a corresponding function depending on the highest threshold reached by the sample.

## How this was tested

Basic unit tests added in the PR to test finalizing after different beta values and flipping correctly between preferences.

### DIFF
--- a/snow/consensus/snowball/snowflake_plus.go
+++ b/snow/consensus/snowball/snowflake_plus.go
@@ -1,0 +1,116 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package snowball
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/bag"
+)
+
+var _ Consensus = (*snowflakePlus)(nil)
+
+// snowflakePlus implements Consensus using Snowflake with multiple
+// alpha thresholds and corresponding beta values.
+type snowflakePlus struct {
+	// alpha1 is the threshold of votes required to update the preference
+	// of the consensus instance.
+	alpha1 int
+
+	// alpha2Cutoff is the lowest alpha threshold that can be used to find
+	// the termination point.
+	alpha2Cutoff int
+
+	// betas are the number of required successful polls corresponding to
+	// each alpha threshold required to finalize.
+	// The ith element of betas is the number of consecutive polls that
+	// received >= alpha2Cutoff + i votes required to finalize.
+	betas []int
+
+	// confidence tracks the number of successful polls in a row for each
+	// corresponding alpha threshold.
+	// The ith element of confidence is the number of consecutive polls
+	// that received >= alpha2Cutoff + i votes.
+	confidence []int
+
+	finalized bool
+
+	choice ids.ID
+}
+
+func newSnowflakePlus(alpha1 int, alpha2Cutoff int, betas []int, choice ids.ID) *snowflakePlus {
+	return &snowflakePlus{
+		alpha1:       alpha1,
+		alpha2Cutoff: alpha2Cutoff,
+		betas:        betas,
+		confidence:   make([]int, len(betas)),
+		choice:       choice,
+	}
+}
+
+// Add is a no-op for snowflakePlus because it has no notion of beta rogue.
+func (s *snowflakePlus) Add(ids.ID) {}
+
+// Returns the currently preferred choice to be finalized
+func (s *snowflakePlus) Preference() ids.ID {
+	return s.choice
+}
+
+// RecordPoll records the results of a network poll. Assumes all choices
+// have been previously added.
+//
+// If the consensus instance was not previously finalized, this function
+// will return true if the poll was successful and false if the poll was
+// unsuccessful.
+//
+// If the consensus instance was previously finalized, the function may
+// return true or false.
+func (s *snowflakePlus) RecordPoll(votes bag.Bag[ids.ID]) bool {
+	choice, count := votes.Mode()
+
+	// If a new choice received an alpha1 threshold of votes,
+	// update my preference and zero out my confidence counters.
+	if choice != s.choice && count >= s.alpha1 {
+		s.choice = choice
+		s.RecordUnsuccessfulPoll()
+	}
+
+	// If my choice received an alpha2 threshold, increment the
+	// corresponding confidence counter. Otherwise, zero out
+	// the confidence counter.
+	for i := 0; i < len(s.confidence); i++ {
+		if count >= s.alpha2Cutoff+i {
+			s.confidence[i]++
+			if s.confidence[i] >= s.betas[i] {
+				s.finalized = true
+			}
+		} else {
+			s.confidence[i] = 0
+		}
+	}
+	return s.finalized
+}
+
+// RecordUnsuccessfulPoll resets the snowflake counters of this consensus
+// instance
+func (s *snowflakePlus) RecordUnsuccessfulPoll() {
+	for i := 0; i < len(s.confidence); i++ {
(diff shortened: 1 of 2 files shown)


snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+			s.confidence[i] = 0
+		}
+	}
+	return s.finalized
+}
+
+// RecordUnsuccessfulPoll resets the snowflake counters of this consensus
+// instance
+func (s *snowflakePlus) RecordUnsuccessfulPoll() {
+	for i := 0; i < len(s.confidence); i++ {
```
StephenButtolph, line 99:
```suggestion
	for i := range s.confidence {
```

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+// return true or false.
+func (s *snowflakePlus) RecordPoll(votes bag.Bag[ids.ID]) bool {
+	choice, count := votes.Mode()
+
+	// If a new choice received an alpha1 threshold of votes,
+	// update my preference and zero out my confidence counters.
+	if choice != s.choice && count >= s.alpha1 {
+		s.choice = choice
+		s.RecordUnsuccessfulPoll()
+	}
```
StephenButtolph, line 78: If the poll is unsuccessful - I think this function should be returning false.

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+//
+// If the consensus instance was previously finalized, the function may
+// return true or false.
+func (s *snowflakePlus) RecordPoll(votes bag.Bag[ids.ID]) bool {
+	choice, count := votes.Mode()
+
+	// If a new choice received an alpha1 threshold of votes,
+	// update my preference and zero out my confidence counters.
+	if choice != s.choice && count >= s.alpha1 {
+		s.choice = choice
```
StephenButtolph, line 76: I think the rest of the code would protect against updating the choice if the instance was previously finalized

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// If my choice received an alpha2 threshold, increment the
+	// corresponding confidence counter. Otherwise, zero out
+	// the confidence counter.
+	for i := 0; i < len(s.confidence); i++ {
+		if count >= s.alpha2Cutoff+i {
+			s.confidence[i]++
+			if s.confidence[i] >= s.betas[i] {
+				s.finalized = true
+			}
```
StephenButtolph, line 88: minor nit:
```suggestion
			s.finalized = s.finalized || s.confidence[i] >= s.betas[i]
```

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+	// update my preference and zero out my confidence counters.
+	if choice != s.choice && count >= s.alpha1 {
+		s.choice = choice
+		s.RecordUnsuccessfulPoll()
+	}
+
+	// If my choice received an alpha2 threshold, increment the
+	// corresponding confidence counter. Otherwise, zero out
+	// the confidence counter.
+	for i := 0; i < len(s.confidence); i++ {
```
StephenButtolph, line 83: This is a pretty big nit, but I'd personally find this more readable if we iterated from `[s.alpha2Cutoff, count]` (incrementing the beta value) and then iterating from `(max(count, s.alpha2Cutoff), k]` (setting the beta value to 0). It would probably be (insignificantly) faster as well.

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// confidence tracks the number of successful polls in a row for each
+	// corresponding alpha threshold.
+	// The ith element of confidence is the number of consecutive polls
+	// that received >= alpha2Cutoff + i votes.
+	confidence []int
+
+	finalized bool
+
+	choice ids.ID
```
StephenButtolph, line 40: `choice` makes it sound like this is final... I feel like `preference` (as used in the other implementations) is more clear.

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+var _ Consensus = (*snowflakePlus)(nil)
+
+// snowflakePlus implements Consensus using Snowflake with multiple
+// alpha thresholds and corresponding beta values.
+type snowflakePlus struct {
+	// alpha1 is the threshold of votes required to update the preference
+	// of the consensus instance.
+	alpha1 int
```
StephenButtolph, line 20:
```suggestion
	// alphaPreference is the threshold of votes required to update the preference
	// of the consensus instance.
	alphaPreference int
```

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// alpha2Cutoff is the lowest alpha threshold that can be used to find
+	// the termination point.
+	alpha2Cutoff int
```
StephenButtolph, line 24:
```suggestion
	// minAlphaConfidence is the lowest alpha threshold that can be used to find
	// the termination point.
	minAlphaConfidence int
```

snow/consensus/snowball/snowflake_plus.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return s.finalized
+}
+
+// RecordUnsuccessfulPoll resets the snowflake counters of this consensus
+// instance
+func (s *snowflakePlus) RecordUnsuccessfulPoll() {
+	for i := 0; i < len(s.confidence); i++ {
+		s.confidence[i] = 0
+	}
```
StephenButtolph, line 101:
```suggestion
	clear(s.confidence)
```

snow/consensus/snowball/snowflake_plus.go
aaronbuchwald: Did you mean should protect against updating the choice or that an upstream part of the consensus code should prevent calling `RecordPoll` on an already finalized consensus instance?

snow/consensus/snowball/snowflake_plus.go
aaronbuchwald: Setting to return count >= minAlphaConfidence as I'm defining the notion of "successful" as made progress towards termination even if that is progress towards termination on a new preference as of this round.

Looking in the `Tree` and `Flat` implementations, it looks like we will always return true if `numVotes > alphaPreference`, but this does not necessarily make progress towards termination with the dual alpha changes, so not sure if that's what we want.

Going to use `minAlphaConfidence` for now since this is just for a metric/simulation.

snow/consensus/snowball/snowflake_plus.go
aaronbuchwald: Ya that's better, thanks

PR #2812, opened 2024-03-05
Cleanup Duplicate Transitive Constructor

joshua-kim:
## Why this should be merged

We have a duplicate `New` constructor to return an opaque transitive instance by its interface. We can just let the caller choose how to use the concrete type and remove the extra constructor

## How this works

Removes extra copy of constructor

## How this was tested

CI

### DIFF
--- a/snow/engine/snowman/transitive.go
+++ b/snow/engine/snowman/transitive.go
@@ -102,7 +102,7 @@ type Transitive struct {
 	errs wrappers.Errs
 }
 
-func NewTransitive(config Config) (*Transitive, error) {
+func New(config Config) (*Transitive, error) {
 	config.Ctx.Log.Info("initializing consensus engine")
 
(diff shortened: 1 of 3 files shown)


snow/engine/snowman/transitive.go
```diff
@@ -106,7 +102,7 @@ type Transitive struct {
 	errs wrappers.Errs
 }
 
-func newTransitive(config Config) (*Transitive, error) {
+func NewTransitive(config Config) (*Transitive, error) {
```
StephenButtolph, line 105: I feel like this should just be "New" we only called it "newTransitive" to avoid shadowing "new"

dhrubabasu (approved):

aaronbuchwald (approved):

StephenButtolph (approved):

PR #2815, opened 2024-03-05
Cleanup consensus metrics

StephenButtolph:
## Why this should be merged

Currently our metrics logic in the snowman consensus engine is repetitive and makes the code more difficult to understand. Additionally, it can result in us updating the same metric multiple times in the same call... Even though we really only need to update them once at the end.

## How this works

Updates metrics once at the end of the engine modifications.

## How this was tested

N/A

### DIFF
--- a/snow/engine/snowman/transitive.go
+++ b/snow/engine/snowman/transitive.go
@@ -214,7 +214,7 @@ func (t *Transitive) Gossip(ctx context.Context) error {
 		return nil
 	}
 
-	lastAccepted, err := t.GetBlock(ctx, lastAcceptedID)
+	lastAccepted, err := t.getBlock(ctx, lastAcceptedID)
 	if err != nil {
 		t.Ctx.Log.Warn("dropping gossip request",
 			zap.String("reason", "block couldn't be loaded"),
@@ -300,7 +300,7 @@ func (t *Transitive) Put(ctx context.Context, nodeID ids.NodeID, requestID uint3
 	if _, err := t.issueFrom(ctx, nodeID, blk, issuedMetric); err != nil {
 		return err
 	}
-	return t.buildBlocks(ctx)
+	return t.executeDeferredWork(ctx)
 }
 
 func (t *Transitive) GetFailed(ctx context.Context, nodeID ids.NodeID, requestID uint32) error {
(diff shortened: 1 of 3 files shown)


snow/engine/snowman/transitive.go
```diff
@@ -553,7 +549,7 @@ func (t *Transitive) Start(ctx context.Context, startReqID uint32) error {
 		return fmt.Errorf("failed to notify VM that consensus is starting: %w",
 			err)
 	}
-	return nil
+	return t.executeDeferredWork(ctx)
```
StephenButtolph, line 552: Added here because of the call to `t.deliver` above.

snow/engine/snowman/transitive.go
```diff
@@ -584,7 +580,19 @@ func (t *Transitive) HealthCheck(ctx context.Context) (interface{}, error) {
 	return intf, fmt.Errorf("vm: %w ; consensus: %w", vmErr, consensusErr)
 }
 
-func (t *Transitive) GetBlock(ctx context.Context, blkID ids.ID) (snowman.Block, error) {
+func (t *Transitive) executeDeferredWork(ctx context.Context) error {
```
StephenButtolph, line 583: This function is going to get larger as I continue to cleanup the logic here.

snow/engine/snowman/transitive.go
```diff
@@ -584,7 +580,19 @@ func (t *Transitive) HealthCheck(ctx context.Context) (interface{}, error) {
...
+	}
+
+	t.metrics.numRequests.Set(float64(t.blkReqs.Len()))
+	t.metrics.numBlocked.Set(float64(len(t.pending)))
+	t.metrics.numBlockers.Set(float64(t.blocked.Len()))
+	t.metrics.numNonVerifieds.Set(float64(t.nonVerifieds.Len()))
+	return nil
+}
+
+func (t *Transitive) getBlock(ctx context.Context, blkID ids.ID) (snowman.Block, error) {
```
StephenButtolph, line 595: Minor unrelated cleanup - but this method didn't need to be exported anymore.

snow/engine/snowman/transitive.go
```diff
@@ -584,7 +580,19 @@ func (t *Transitive) HealthCheck(ctx context.Context) (interface{}, error) {
...
-func (t *Transitive) GetBlock(ctx context.Context, blkID ids.ID) (snowman.Block, error) {
+func (t *Transitive) executeDeferredWork(ctx context.Context) error {
+	if err := t.buildBlocks(ctx); err != nil {
+		return err
+	}
+
+	t.metrics.numRequests.Set(float64(t.blkReqs.Len()))
+	t.metrics.numBlocked.Set(float64(len(t.pending)))
+	t.metrics.numBlockers.Set(float64(t.blocked.Len()))
+	t.metrics.numNonVerifieds.Set(float64(t.nonVerifieds.Len()))
```
StephenButtolph, line 591: This is now the only location these gauges are set.

joshua-kim (approved):

abi87 (approved):

PR #2819, opened 2024-03-06
Prevent zero length values in slices and maps in codec

StephenButtolph:
## Why this should be merged

With the activation of Durango, we no longer specify the maximum slice length in the codec. However, this can cause unsuspecting users to encounter unusual performance regressions.

This explicitly errors rather than allowing slow usage.

## How this works

Prohibits the usage of the codec to encode/decode zero length values in slices and maps.

## How this was tested

- [X] Added unit test
- [X] CI

### DIFF
--- a/codec/reflectcodec/type_codec.go
+++ b/codec/reflectcodec/type_codec.go
@@ -754,6 +754,7 @@ func (c *genericCodec) unmarshal(
 			mapKey := reflect.New(mapKeyType).Elem()
 
 			keyStartOffset := p.Offset
+
 			if err := c.unmarshal(p, mapKey, typeStack); err != nil {
 				return err
 			}
--- a/codec/reflectcodec/type_codec.go
+++ b/codec/reflectcodec/type_codec.go
@@ -159,6 +159,10 @@ func (c *genericCodec) size(
 			return 0, false, err
 		}
 
+		if size == 0 {
+			return 0, false, fmt.Errorf("can't marshal slice of zero length values: %w", codec.ErrMarshalZeroLength)
+		}
+
 		// For fixed-size types we manually calculate lengths rather than
 		// processing each element separately to improve performance.
 		if constSize {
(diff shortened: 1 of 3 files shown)


codec/reflectcodec/type_codec.go
```diff
@@ -159,6 +159,10 @@ func (c *genericCodec) size(
 			return 0, false, err
 		}
 
+		if size == 0 {
```
patrick-ogrady, line 162: Do we also want to enforce this on `reflect.Array`?

patrick-ogrady: or on `reflect.Struct` for serialized fields?

patrick-ogrady: on another note, would it make more sense to check this after iterating over each `innerItem`?

StephenButtolph: Originally, this implementation banned zero length structs and zero length arrays (as everything else has a non-zero size). However, we currently use zero length structs in production: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/txs/codec.go#L115.

So, rather than preventing 0 length values. We prevent 0 length entries in dynamically allocated structures. (i.e. anything with a length encoded in the format). This ensures that processing is O(n) in the number of bytes (whereas before it could have been O(2^n).

As for if the size should be checked in each `innerItem`, we don't need to do this. If the size is 0 of an instance, it is guaranteed to always be 0 based on the type (because we don't have optional fields).

codec/reflectcodec/type_codec.go
```diff
@@ -235,6 +239,10 @@ func (c *genericCodec) size(
 			return 0, false, err
 		}
 
+		if keySize == 0 && valueSize == 0 {
```
joshua-kim, line 242: Do we not need to check this for remaining keys/value pairs?

codec/test_codec.go
```diff
@@ -36,7 +36,7 @@ var (
 		TestNilSliceSerialization,
 		TestEmptySliceSerialization,
 		TestSliceWithEmptySerialization,
-		TestSliceWithEmptySerializationOutOfMemory,
```
joshua-kim, line 39: Should we have a test case for the map case as well

codec/reflectcodec/type_codec.go
StephenButtolph: If the size is 0 of an instance, it is guaranteed to always be 0 based on the type (because we don't have optional fields).

joshua-kim (approved):

PR #2821, opened 2024-03-07
fix(INFRA:3587):  packer build fix to use hcl and version

vineetguptadev:
## Why this should be merged
This change fix the public avalanchego AMIs in AWS. 

## How this works
This fixes the packer build to use the new AMIs


.github/workflows/build-public-ami.yml
```diff
@@ -55,6 +59,8 @@ jobs:
 
       - name: Create AMI and upload to marketplace
         run: |
+          packer hcl2_upgrade .github/packer/ubuntu-jammy-x86_64-public-ami.json
```
yevhenvolchenko: add HSL file and remove json

yevhenvolchenko:
```suggestion
```

.github/workflows/update-ami.py
```diff
@@ -1,14 +1,14 @@
...
+import os
 import boto3
 import uuid
 import re
 import subprocess
 
 # Globals
 amifile = '.github/workflows/amichange.json'
-packerfile = ".github/packer/ubuntu-jammy-x86_64-public-ami.json"
+packerfile = ".github/packer/ubuntu-jammy-x86_64-public-ami.json.pkr.hcl"
```
yevhenvolchenko, line 11: comment for line 18:

remove capital `t`
skip_create_ami = os.getenv('SKIP_CREATE_AMI', "true")

.github/workflows/build-public-ami.yml
```diff
@@ -26,7 +30,7 @@ jobs:
       - name: Install aws cli
         run: |
           sudo apt update
-          sudo apt-get -y install packer python3-boto3
+          sudo apt-get -y install packer=${PACKER_VERSION} python3-boto3=${PYTHON3_BOTO3_VERSION}
```
yevhenvolchenko: comment for line 50:

remove capital F
echo "SKIP_CREATE_AMI=false" >> $GITHUB_ENV

PR #2826, opened 2024-03-07
[vms/avm] Cleanup `GetTx` + remove state pruning logic

dhrubabasu:
## Why this should be merged

Durango is live now, we can cleanup :)

## How this works

Do the cleanup

## How this was tested

CI


joshua-kim (approved):

vms/avm/state/state.go
```diff
@@ -150,9 +127,8 @@ type state struct {
 	utxoDB        database.Database
 	utxoState     avax.UTXOState
 
-	statusesPruned bool
-	statusCache    cache.Cacher[ids.ID, *choices.Status] // cache of id -> choices.Status. If the entry is nil, it is not in the database
-	statusDB       database.Database
+	statusCache cache.Cacher[ids.ID, *choices.Status] // cache of id -> choices.Status. If the entry is nil, it is not in the database
+	statusDB    database.Database
```
StephenButtolph, line 131: I think we should be able to remove this entirely right?

dhrubabasu: Removed

PR #2830, opened 2024-03-07
[network/peer] Disconnect from peers who only send legacy version field

dhrubabasu:
## Why this should be merged

Everybody should be sending the new field now that we are now on v1.11.x. We can remove support for the old field.

## How this works

🧹 

## How this was tested

CI


network/peer/peer.go
```diff
@@ -946,25 +946,19 @@ func (p *peer) handleHandshake(msg *p2p.Handshake) {
...
-	if msg.Client != nil {
-		p.version = &version.Application{
-			Name:  msg.Client.Name,
-			Major: int(msg.Client.Major),
-			Minor: int(msg.Client.Minor),
-			Patch: int(msg.Client.Patch),
-		}
-	} else {
-		// Handle legacy version field
-		peerVersion, err := version.ParseLegacyApplication(msg.MyVersion)
```
StephenButtolph, line 958: `MyVersion` should be removed from the proto

network/peer/peer.go
```diff
@@ -523,20 +523,12 @@ func (p *peer) writeMessages() {
 	}
 
 	myVersion := p.VersionCompatibility.Version()
-	legacyApplication := &version.Application{
-		Name:  version.LegacyAppName,
```
StephenButtolph, line 527: Is `LegacyAppName` dead code now?

dhrubabasu: Replaced it in tests with `Client` and removed the variable

StephenButtolph (approved):

joshua-kim (approved):

PR #2833, opened 2024-03-07
Remove legacy p2p message handling

dhrubabasu:
## Why this should be merged

Durango is active, let's cleanup.

## How this works

Removes `components/message` package since we are all-in on the new p2p sdk 🚀

## How this was tested

CI


joshua-kim (approved):

vms/avm/network/network.go
```diff
@@ -183,45 +181,7 @@ func (n *Network) PullGossip(ctx context.Context) {
 }
 
 func (n *Network) AppGossip(ctx context.Context, nodeID ids.NodeID, msgBytes []byte) error {
```
StephenButtolph, line 183: I think this function can be completely removed.

StephenButtolph (approved):

PR #2834, opened 2024-03-08
[vms/avm] Remove `snow.Context` from `Network`

dhrubabasu:
## Why this should be merged

Reduces the dependency on `snow.Context`.

## How this works

Pass in fields directly.

## How this was tested

CI


vms/avm/network/network.go
```diff
@@ -31,27 +31,30 @@ var (
 type Network struct {
 	*p2p.Network
 
+	log       logging.Logger
```
dhrubabasu, line 34: Moved this up to match the platformvm Network struct: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/network/network.go#L29-L42

joshua-kim (approved):

StephenButtolph (approved):

PR #2835, opened 2024-03-08
Allow configuring push gossip to send txs to validators by stake

StephenButtolph:
## Why this should be merged

Block production is weighted by stake. Without introspecting into the (potentially volatile) proposer list, the most likely nodes to propose a block are the nodes with the highest stake. To issue transactions as quickly as possible, an API node may want to send transactions to peers based on their stake weight.

## How this works

Adds a parameter `push-gossip-percent-stake` (which is expected be in [0, 1]) which results in sending push gossip to validators based on the inverse CDF.

## How this was tested

- [X] CI
- [x] Test on Fuji
- [x] Test on Mainnet


network/p2p/validators.go
```diff
@@ -86,28 +117,63 @@ func (v *Validators) Sample(ctx context.Context, limit int) []ids.NodeID {
 
 	v.refresh(ctx)
 
-	// TODO: Account for peer connectivity during the sampling of validators
-	// rather than filtering sampled validators.
```
StephenButtolph, line 90: This also addresses this TODO

network/p2p/gossip/gossip.go
```diff
@@ -323,9 +327,10 @@ type PushGossiper[T Gossipable] struct {
 }
 
 type BranchingFactor struct {
-	Validators    int
-	NonValidators int
-	Peers         int
+	StakePercentage float64
```
patrick-ogrady, line 330: nit: wonder if it makes sense to take the `num` strategy here like we do for warp?

patrick-ogrady: I suppose we don't need to because we don't require determinism.

network/p2p/gossip/gossip.go
```diff
@@ -323,9 +327,10 @@ type PushGossiper[T Gossipable] struct {
...
+	Validators      int
```
patrick-ogrady, line 331: nit: change to `UniformValidators` so it is clear that this is done differently rather than playing with `StakePercentage`?

StephenButtolph: I decided to document these values rather than trying to re-name them.

network/p2p/validators.go
```diff
@@ -43,23 +52,39 @@ func NewValidators(
...
+	return v.nodeID.Compare(other.nodeID)
 }
 
 func (v *Validators) refresh(ctx context.Context) {
 	if time.Since(v.lastUpdated) < v.maxValidatorSetStaleness {
 		return
 	}
 
-	v.validatorIDs.Clear()
+	v.validatorList = v.validatorList[:0]
```
patrick-ogrady, line 85: nit: you may want to note this works with `nil` arrays

network/p2p/validators.go
```diff
@@ -4,27 +4,36 @@
...
+	_ ValidatorSet     = (*Validators)(nil)
+	_ ValidatorPortion = (*Validators)(nil)
+	_ NodeSampler      = (*Validators)(nil)
 )
 
 type ValidatorSet interface {
 	Has(ctx context.Context, nodeID ids.NodeID) bool // TODO return error
 }
 
+type ValidatorPortion interface {
```
patrick-ogrady, line 33: nit: `ValidatorSubset`?

network/p2p/gossip/gossip.go
```diff
@@ -461,9 +476,11 @@ func (p *PushGossiper[T]) gossip(
 	sentCountMetric.Add(float64(len(gossip)))
 	sentBytesMetric.Add(float64(sentBytes))
 
+	validatorsByStake := p.validators.Top(ctx, gossipParams.StakePercentage)
```
patrick-ogrady, line 479: Can we add a metric for how many validators are selected here?

joshua-kim (approved):

dhrubabasu (approved):

PR #2836, opened 2024-03-08
Combine AppGossip and AppGossipSpecific

StephenButtolph:
## Why this should be merged

This PR reduces duplicate code and makes `AppGossip` more useful as we can now gossip to specific nodes in addition to randomly selected nodes.

## How this works

The only real code change is merging `Send` and `Gossip` in the networking package. This is almost as easy as performing `Send` and then `Gossip`. The only tweak is to prevent nodes from being sampled during `Gossip` that were already selected by `Send`.

## How this was tested

- [X] CI


snow/engine/common/appsender/appsender_client.go
```diff
@@ -107,35 +107,23 @@ func (c *Client) SendAppError(ctx context.Context, nodeID ids.NodeID, requestID
...
-	return err
-}
-
-func (c *Client) SendAppGossipSpecific(ctx context.Context, nodeIDs set.Set[ids.NodeID], msg []byte) error {
-	nodeIDsBytes := make([][]byte, nodeIDs.Len())
+	nodeIDs := make([][]byte, config.NodeIDs.Len())
 	i := 0
-	for nodeID := range nodeIDs {
-		nodeIDsBytes[i] = nodeID.Bytes()
+	for nodeID := range config.NodeIDs {
```
patrick-ogrady, line 115: nit: is there a reason we don't do:

```
for i, nodeID := range config.NodeIDs {
...
}
```

network/network.go
```diff
@@ -310,25 +310,40 @@ func NewNetwork(
...
+	var (
+		sampledPeers = n.samplePeers(config, subnetID, allower)
+		sentTo       = set.NewSet[ids.NodeID](len(namedPeers) + len(sampledPeers))
+		now          = n.peerConfig.Clock.Time()
+	)
+
+	// send to peer and update metrics
+	for _, peers := range [][]peer.Peer{namedPeers, sampledPeers} {
+		for _, peer := range peers {
+			if peer.Send(n.onCloseCtx, msg) {
```
patrick-ogrady, line 334: As noted below, we don't need to check `sentTo` because we avoid adding `namedPeers` to `sampledPeers`.

network/network.go
```diff
@@ -696,24 +711,19 @@ func (n *network) getPeers(
...
 func (n *network) samplePeers(
+	config common.SendConfig,
 	subnetID ids.ID,
-	numValidatorsToSample,
-	numNonValidatorsToSample int,
-	numPeersToSample int,
 	allower subnets.Allower,
 ) []peer.Peer {
 	// If there are fewer validators than [numValidatorsToSample], then only
 	// sample [numValidatorsToSample] validators.
```
patrick-ogrady, line 719: nit: we may want to note this is an optimization, not a correctness thing

patrick-ogrady (approved):
Just nits, otherwise logic changes make sense to me

network/network.go
```diff
@@ -310,25 +310,42 @@ func NewNetwork(
...
 	subnetID ids.ID,
-	numValidatorsToSend int,
-	numNonValidatorsToSend int,
-	numPeersToSend int,
 	allower subnets.Allower,
 ) set.Set[ids.NodeID] {
-	peers := n.samplePeers(subnetID, numValidatorsToSend, numNonValidatorsToSend, numPeersToSend, allower)
-	return n.send(msg, peers)
+	namedPeers := n.getPeers(config.NodeIDs, subnetID, allower)
+	n.peerConfig.Metrics.MultipleSendsFailed(
```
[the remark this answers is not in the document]
StephenButtolph: We already update the metrics in `peer.Send` for this case.

joshua-kim (approved):

PR #2839, opened 2024-03-08
[utils/compression] Remove gzip compressor

dhrubabasu:
## Why this should be merged

Durango is active, we can drop gzip compressed messages.

## How this works

🧹 Delete some code

## How this was tested

CI


utils/compression/type.go
```diff
@@ -14,16 +14,14 @@ type Type byte
 
 const (
 	TypeNone Type = iota + 1
-	TypeGzip      // Remove once v1.11.x is out.
```
joshua-kim, line 17: I don't think this value is ever sent over the wire so I _think_ it's safe to just delete instead of marking as a placeholder in this enum...

utils/compression/type.go
```diff
@@ -14,16 +14,14 @@ type Type byte
...
+	_
```
darioush, line 17: should we add a comment about this?

darioush (approved):

joshua-kim (approved):

StephenButtolph (approved):

PR #2841, opened 2024-03-09
fix(INFRA-3587): migrate packer config and fix workflow

vineetguptadev:
## Why this should be merged

This change fix the public avalanchego AMIs in AWS.

## How this works
Enables ami creation based on workflow and packer to build amis

## How this was tested
Locally tested packer validate and build


Dirrk:
going to use my branch since its working.

PR #2850, opened 2024-03-14
Fix P-chain Test

joshua-kim:
## Why this should be merged

I think this test is wrong - instead of checking for an updated preference upon accepting a block during bootstrapping we should be checking against the last accepted block because we don't participate in consensus during bootstrapping.

## How this works

Updates test

## How this was tested

Existing Test


joshua-kim:
I believe the point of this test is to make sure that once consensus starts we are starting consensus at the point after bootstrapping completes instead of at the point before it started.

PR #2855, opened 2024-03-20
Interval tree syncing integration

StephenButtolph:
## Why this should be merged

- Replaces usage of the legacy bootstrapping queue with the optimized interval tree implementation.
- Fixes an edge case that could cause bootstrapping to unexpectedly terminate early.
- Removes unnecessary calls to `GetBlock` because they are no longer expected to return a `nil` error in any case.

## How this works

^

## How this was tested

- [X] CI
- [X] Bootstrapped Fuji


snow/engine/snowman/bootstrap/bootstrapper_test.go
```diff
@@ -274,7 +270,7 @@ func TestBootstrapperSingleFrontier(t *testing.T) {
 	}
 
 	require.NoError(bs.startSyncing(context.Background(), acceptedIDs))
-	require.Equal(snow.NormalOp, config.Ctx.State.Get().State)
+	require.Equal(snow.Bootstrapping, config.Ctx.State.Get().State)
```
StephenButtolph, line 273: This test was previously incorrect (and the implementation had a minor bug). Because this iteration accepted a block and `1 < Maxuint64/2`, there should be another round of bootstrapping performed.

snow/engine/snowman/bootstrap/context.go
```diff
@@ -0,0 +1,29 @@
...
+	context.Context
+	common.Haltable
+}
+
+func (c *haltableContext) Err() error {
+	if c.Halted() {
+		return errHalted
+	}
+	return c.Context.Err()
+}
```
StephenButtolph, line 29: This is a bit of an abstraction break... This function doesn't correctly implement the `context.Context` interface because it doesn't support `Done()` correctly... But that really isn't possible to implement correctly without modifying `common.Haltable` and introducing a new goroutine... I felt like this was a reasonable implementation (seeing as `common.Haltable` should really be removed anyway).

snow/engine/snowman/bootstrap/interval/blocks.go
```diff
@@ -91,7 +91,7 @@ func Add(
 // Invariant: Execute assumes that GetMissingBlockIDs would return an empty set.
 func Execute(
 	ctx context.Context,
-	log logging.Logger,
+	log logging.Func,
```
StephenButtolph, line 94: This should be moved into the parent PR.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -683,58 +662,70 @@ func (b *Bootstrapper) process(ctx context.Context, blk snowman.Block, processin
...
 		// it
-		b.Blocked.AddMissingID(parentID)
 		if err := b.fetch(ctx, parentID); err != nil {
 			return err
 		}
 
-		if err := b.Blocked.Commit(); err != nil {
-			return err
-		}
-		return b.tryStartExecuting(ctx)
```
StephenButtolph, line 694: Calling `tryStartExecuting` multiple times in a call to the engine can cause the engine to think that it has finished fetching blocks multiple times. This means that the engine may finish multiple times - which means it may exit early than it should (may exit after only one round of bootstrapping or finish even though we were still catching up to tip).

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -434,11 +436,6 @@ func (b *Bootstrapper) fetch(ctx context.Context, blkID ids.ID) error {
 		return nil
 	}
 
-	// Make sure we don't already have this block
-	if _, err := b.VM.GetBlock(ctx, blkID); err == nil {
-		return b.tryStartExecuting(ctx)
-	}
```
StephenButtolph, line 440: Same comment as the other location - we shouldn't be calling `tryStartExecuting` in multiple places per call. Additionally, this can never really be hit anyways... because we only ever try `fetching` a block if we don't have it locally.

snow/engine/snowman/bootstrap/block_job.go
```diff
@@ -1,116 +0,0 @@
...
-		numAccepted: p.numAccepted,
-		numDropped:  p.numDropped,
-		blk:         blk,
-		vm:          p.vm,
-	}, nil
-}
-
-type blockJob struct {
-	log                     logging.Logger
-	numAccepted, numDropped prometheus.Counter
```
StephenButtolph, line 47: Need to handle the metrics correctly

snow/engine/snowman/bootstrap/metrics.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 type metrics struct {
-	numFetched, numDropped, numAccepted prometheus.Counter
```
StephenButtolph, line 13: `numDropped` was only `Inc()`ed immediately prior to `FATAL`ing - so I don't think this metric is really important to keep.

chains/manager.go
```diff
@@ -579,10 +579,6 @@ func (m *manager) createAvalancheChain(
 	if err != nil {
 		return nil, err
 	}
-	blockBlocker, err := queue.NewWithMissing(blockBootstrappingDB, "block", ctx.Registerer)
```
StephenButtolph, line 582: Do we want any of the metrics from in here?

snow/engine/snowman/bootstrap/block_job.go
```diff
@@ -1,116 +0,0 @@
...
-			zap.Stringer("blkID", blkID),
-			zap.Uint64("height", b.blk.Height()),
-			zap.Time("timestamp", b.blk.Timestamp()),
-		)
-		if err := b.blk.Accept(ctx); err != nil {
-			b.log.Debug("failed to accept block during bootstrapping",
-				zap.Stringer("blkID", blkID),
-				zap.Error(err),
-			)
-			return fmt.Errorf("failed to accept block in bootstrapping: %w", err)
```
StephenButtolph, line 108: We should keep this error information

chains/manager.go
StephenButtolph: I'm going to say no - I don't think we need the ETA metric - and we no long have a cache here (because there are no reads!)

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -149,10 +150,7 @@ func (b *Bootstrapper) Clear(context.Context) error {
 	b.Ctx.Lock.Lock()
 	defer b.Ctx.Lock.Unlock()
 
-	if err := b.Config.Blocked.Clear(); err != nil {
-		return err
-	}
-	return b.Config.Blocked.Commit()
+	return database.AtomicClear(b.DB, b.DB)
```
StephenButtolph, line 153: We use atomic clear here to ensure we don't get into a position where the database has only been half deleted.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -579,65 +576,38 @@ func (b *Bootstrapper) markUnavailable(nodeID ids.NodeID) {
...
 func (b *Bootstrapper) process(ctx context.Context, blk snowman.Block, processingBlocks map[ids.ID]snowman.Block) error {
+	lastAcceptedID, err := b.VM.LastAccepted(ctx)
+	if err != nil {
+		return fmt.Errorf("couldn't get last accepted ID: %w", err)
+	}
+	lastAccepted, err := b.VM.GetBlock(ctx, lastAcceptedID)
+	if err != nil {
+		return fmt.Errorf("couldn't get last accepted block: %w", err)
+	}
+	lastAcceptedHeight := lastAccepted.Height()
```
StephenButtolph, line 587: We could optimize this a bit by tracking the last accepted height ourselves... But I found this more obviously correct.

snow/engine/snowman/bootstrap/metrics.go
StephenButtolph: Pulled into a separate PR: #2858

snow/engine/snowman/bootstrap/acceptor.go
```diff
@@ -0,0 +1,52 @@
...
+
+type blockAcceptor struct {
+	snowman.Block
+
+	ctx         *snow.ConsensusContext
+	numAccepted prometheus.Counter
+}
+
+func (b *blockAcceptor) Accept(ctx context.Context) error {
+	b.numAccepted.Inc()
```
joshua-kim, line 47: nit: Shouldn't we increase this counter at the end of this func instead?

joshua-kim (approved):

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -566,168 +555,128 @@ func (b *Bootstrapper) markUnavailable(nodeID ids.NodeID) {
...
-		}
-
-		// If the parent is not available in processing blocks, attempt to get
-		// the block from the vm
-		parent, err = b.VM.GetBlock(ctx, parentID)
-		if err == nil {
-			blk = parent
-			continue
-		}
-		// TODO: report errors that aren't `database.ErrNotFound`
```
StephenButtolph, line 675: We technically didn't _need_ to call this here... But it avoids requesting the last accepted block over the p2p.

We now handle this in a more clean way by tracking the last accepted height.

snow/engine/snowman/bootstrap/bootstrapper_test.go
```diff
@@ -1221,62 +537,33 @@ func TestBootstrapOldBlockAfterStateSync(t *testing.T) {
...
 
+	require.NoError(bs.Ancestors(context.Background(), peerID, reqID, blocksToBytes(blks[0:1])))
 	require.Equal(snow.NormalOp, config.Ctx.State.Get().State)
-	require.Equal(choices.Processing, blk0.Status())
-	require.Equal(choices.Accepted, blk1.Status())
+	require.Equal(choices.Processing, blks[0].Status())
+	require.Equal(choices.Accepted, blks[1].Status())
 }
 
 func TestBootstrapContinueAfterHalt(t *testing.T) {
```
StephenButtolph, line 560: Honestly, it isn't particularly clear to me what this is testing.

snow/engine/snowman/bootstrap/bootstrapper_test.go
```diff
@@ -1487,62 +710,111 @@ func TestBootstrapperReceiveStaleAncestorsMessage(t *testing.T) {
...
-	require.Equal(choices.Accepted, blk1.Status())
-	require.Equal(choices.Accepted, blk2.Status())
+	requireStatusIs(require, blks, choices.Accepted)
 
-	require.NoError(bs.Ancestors(context.Background(), peerID, reqIDBlk1, [][]byte{blkBytes1}))
+	require.NoError(bs.Ancestors(context.Background(), peerID, reqIDBlk1, blocksToBytes(blks[1:2])))
 	require.Equal(snow.Bootstrapping, config.Ctx.State.Get().State)
 }
+
+func generateBlockchain(length uint64) []snowman.Block {
```
StephenButtolph, line 736: kinda wild to me that we didn't have this already... It might make sense for some of these functions to be move into a common testing package (because they seem useful in a number of places)

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -566,168 +555,128 @@ func (b *Bootstrapper) markUnavailable(nodeID ids.NodeID) {
...
 
-		// We added a new block to the queue, so track that it was fetched
-		b.numFetched.Inc()
+		height := blk.Height()
+		b.tipHeight = max(b.tipHeight, height)
 
-		// Periodically log progress
-		blocksFetchedSoFar := b.Blocked.Jobs.PendingJobs()
-		if blocksFetchedSoFar%statusUpdateFrequency == 0 {
+		if numFetched%statusUpdateFrequency == 0 {
```
StephenButtolph, line 594: Because `numFetched` is now increased in batches, this check doesn't work correctly anymore.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -566,168 +555,128 @@ func (b *Bootstrapper) markUnavailable(nodeID ids.NodeID) {
...
+	if b.restarted {
+		log = b.Ctx.Log.Debug
+	}
+
+	numToExecute := b.tree.Len()
+	err = execute(
+		&haltableContext{
+			Context:  ctx,
+			Haltable: b,
+		},
```
StephenButtolph, line 655: I'm not sure that this is safe. This context is passed into `Verify` and `Accept` - which could cause errors that were thought to be FATAL by the VMs to be treated non-fatally by the engine. It would probably be better to just pass in the haltable reference into `execute` for now.

joshua-kim (approved):

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -0,0 +1,252 @@
...
+	logPeriod             = 5 * time.Second
+)
+
+// getMissingBlockIDs returns the ID of the blocks that should be fetched to
+// attempt to make a single continuous range from
+// (lastAcceptedHeight, highestTrackedHeight].
+//
+// For example, if the tree currently contains heights [1, 4, 6, 7] and the
+// lastAcceptedHeight is 2, this function will return the IDs corresponding to
+// blocks [3, 5].
```
abi87, line 36: nit: can you clarify the example here? The tree contains intervals really. What is the tree content in the example? [{1}, {4}, {6}, {7}] or [{1,4} {6,7}]?

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -0,0 +1,252 @@
...
+	log logging.Func,
+	db database.Database,
+	parser block.Parser,
+	tree *interval.Tree,
+	lastAcceptedHeight uint64,
+) error {
+	var (
+		batch                    = db.NewBatch()
+		processedSinceBatchWrite uint
+		writeBatch               = func() error {
```
abi87, line 139: nit: not sure if it makes a difference but should we pull this function out to avoid creating writeBatch object at each execute invocation?

snow/engine/snowman/bootstrap/storage_test.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+
+			size, err := database.Count(db)
+			require.NoError(err)
+			require.Zero(size)
+		})
+	}
+}
+
+type testParser func(context.Context, []byte) (snowman.Block, error)
```
abi87, line 294: nit: add type assertion `block.Parser`

snow/engine/snowman/bootstrap/bootstrapper.go
abi87: I wonder if this comment should be in the codebase

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -378,24 +372,20 @@ func (b *Bootstrapper) startSyncing(ctx context.Context, acceptedContainerIDs []
...
+		zap.Int("numMissingBlocks", numMissingBlockIDs),
 	)
 
-	toProcess := make([]snowman.Block, 0, len(pendingContainerIDs))
-	for _, blkID := range pendingContainerIDs {
-		b.Blocked.AddMissingID(blkID)
-
+	toProcess := make([]snowman.Block, 0, numMissingBlockIDs)
+	for blkID := range b.missingBlockIDs {
 		// TODO: if `GetBlock` returns an error other than
```
abi87, line 389: q: given the rework, are we planning to solve this TODO as well? (maybe in a different PR)

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -566,168 +555,122 @@ func (b *Bootstrapper) markUnavailable(nodeID ids.NodeID) {
...
-
-		blkHeight := blk.Height()
-		if status == choices.Accepted || blkHeight <= b.startingHeight {
-			// We can stop traversing, as we have reached the accepted frontier
-			if err := b.Blocked.Commit(); err != nil {
-				return err
-			}
-			return b.tryStartExecuting(ctx)
-		}
+//   - ancestors is a set of blocks that can be used to optimisically lookup
```
abi87, line 558: typo? optimisically --> optimistically

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -0,0 +1,252 @@
...
+//
+//   - blk is a block that is assumed to have been marked as acceptable by the
+//     bootstrapping engine.
+//   - ancestors is a set of blocks that can be used to lookup blocks.
+//
+// If [blk]'s height is <= the last accepted height, then it will be removed
+// from the missingIDs set.
+//
+// Returns a newly discovered blockID that should be fetched.
+func process(
```
abi87, line 80: q: IIUC we don't check whether blocks received in an `Ancestors` messages are contiguous in `block.BatchedParseBlock. It's not up to the VM to check block contiguity and we don't really care. Instead it's `process` responsability to find what's next to download (included possible gaps in the Ancestors message). Do I understand correctly?

abi87: Also in case the ancestors blocks have multiple gaps, we stop at the first one right? We won't greedly record the other height and return multiple heights to fetch.

abi87 (changes requested):
nits and questions, still reviewing

snow/engine/snowman/bootstrap/storage.go
StephenButtolph: One thing that `process` doesn't do - which it could do... is process blocks inside of `ancestors` that are contained inside of `missingBlockIDs`. I felt like this wasn't worth implementing because:

1. It introduces more complexity
2. This will only store blocks in a couple of niche cases

Cases this could be useful:

1. If a node provides non-consecutive blocks in ancestors but during the start of bootstrapping we have added one of the provided blocks as desired.
    a. This can only happen in the byzantine node case.
2. A node provides a set of consecutive ancestors where only some of the ancestors have already been persisted.
    a. We'll have already sent a request for this ancestors, so this is only an optimization if that request were to fail (as we could then decide not to re-issue the request).

abi87 (approved):
very nice work

PR #2858, opened 2024-03-20
Remove useless bootstrapping metric

StephenButtolph:
## Why this should be merged

This metric is only incremented prior to FATALing - so we should just remove it.

## How this works

1. Delete
2. Require that bootstrapping always is able to execute every transaction. This is guaranteed to be the case because the X-chain has been linearized.

## How this was tested

- [x] CI

### DIFF
--- a/snow/engine/avalanche/bootstrap/vertex_job.go
+++ b/snow/engine/avalanche/bootstrap/vertex_job.go
@@ -20,12 +20,15 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 )
 
-var errMissingVtxDependenciesOnAccept = errors.New("attempting to execute blocked vertex")
+var (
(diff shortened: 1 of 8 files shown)


snow/engine/avalanche/bootstrap/vertex_job.go
```diff
@@ -94,15 +95,12 @@ func (v *vertexJob) Execute(ctx context.Context) error {
 	}
 	for _, tx := range txs {
 		if tx.Status() != choices.Accepted {
-			v.numDropped.Inc()
-			v.log.Warn("attempting to execute vertex with non-accepted transactions")
-			return nil
```
StephenButtolph, line 99: We now require this to never happen.

dhrubabasu (approved):

JacobEverly (approved):

PR #2859, opened 2024-03-21
`tmpnet`: Ensure nodes are properly detached from the parent process

marun:
Previously tmpnet called `Cmd.Wait` on the node process based on a misreading of the advice of the `Cmd.Start` to ensure that system resources were freed. This ensured that any error in the parent process would result in the node processes being killed.

Now, the child is properly detached by setting `Setsid` and avoiding calling Wait. This ensures that the node restarts required when setting up subnets (e.g. for subnet-evm or hypersdk) do not risk the nodes of an existing network being stopped if the test suite exits non-zero due to a test failure.

### DIFF
--- a/tests/fixture/tmpnet/defaults.go
+++ b/tests/fixture/tmpnet/defaults.go
@@ -46,7 +46,7 @@ func DefaultFlags() FlagsMap {
 		config.HealthCheckFreqKey:               "2s",
 		config.AdminAPIEnabledKey:               true,
 		config.IndexEnabledKey:                  true,
-		config.LogDisplayLevelKey:               "INFO",
+		config.LogDisplayLevelKey:               "off", // Display logging not needed since nodes run headless
 		config.LogLevelKey:                      "DEBUG",
 		config.MinStakeDurationKey:              DefaultMinStakeDuration.String(),
 	}
--- a/tests/fixture/tmpnet/defaults.go
+++ b/tests/fixture/tmpnet/defaults.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/config"
+	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/vms/platformvm/txs/executor"
 )
 
--- a/tests/fixture/tmpnet/defaults.go
+++ b/tests/fixture/tmpnet/defaults.go
@@ -46,8 +46,8 @@ func DefaultFlags() FlagsMap {
 		config.HealthCheckFreqKey:               "2s",
 		config.AdminAPIEnabledKey:               true,
 		config.IndexEnabledKey:                  true,
-		config.LogDisplayLevelKey:               "off", // Display logging not needed since nodes run headless
-		config.LogLevelKey:                      "DEBUG",
+		config.LogDisplayLevelKey:               logging.Off.String(), // Display logging not needed since nodes run headless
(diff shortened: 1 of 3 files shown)


tests/fixture/tmpnet/defaults.go
```diff
@@ -46,7 +46,7 @@ func DefaultFlags() FlagsMap {
 		config.HealthCheckFreqKey:               "2s",
 		config.AdminAPIEnabledKey:               true,
 		config.IndexEnabledKey:                  true,
-		config.LogDisplayLevelKey:               "INFO",
+		config.LogDisplayLevelKey:               "off", // Display logging not needed since nodes run headless
 		config.LogLevelKey:                      "DEBUG",
```
StephenButtolph, line 50:
```suggestion
		config.LogDisplayLevelKey:               logging.Off.String(), // Display logging not needed since nodes run headless
		config.LogLevelKey:                      logging.Debug.String(),
```

tests/fixture/tmpnet/node_process.go
```diff
@@ -112,7 +112,22 @@ func (p *NodeProcess) Start(w io.Writer) error {
...
+		Setsid: true,
+	}
+
+	// Redirect stdout and stderr error to a file since it will detach from the parent process
+	outfile, err := os.OpenFile("nohup.out", os.O_RDWR|os.O_CREATE|os.O_TRUNC, perms.ReadWrite)
+	if err != nil {
+		return fmt.Errorf("failed to create stderr/stdout output file: %w", err)
+	}
+	cmd.Stdout = outfile
+	cmd.Stderr = outfile
```
StephenButtolph, line 129: This seems slightly weird.
- Are multiple nodes going to be writing to the same file?
- Should this just drop the output rather than writing it to a file?
- Where is this file going to be placed?
  - Should it be placed in the node's folder?

tests/fixture/tmpnet/node_process.go
```diff
@@ -112,7 +112,26 @@ func (p *NodeProcess) Start(w io.Writer) error {
...
+	// the parent's stdout and stderr. Since a node is intended to run
+	// headless and should be logging everything to disk internally, its
+	// direct output can be ignored by sending to /dev/null.
+	devNull, err := os.OpenFile(os.DevNull, os.O_WRONLY, 0)
+	if err != nil {
+		return fmt.Errorf("failed to open /dev/null: %w", err)
+	}
+	defer devNull.Close()
+	cmd.Stdout = devNull
+	cmd.Stderr = devNull
```
StephenButtolph, line 133: > // Stdout and Stderr specify the process's standard output and error.
	//
	// If either is nil, Run connects the corresponding file descriptor
	// to the null device (os.DevNull).

Do we need to be setting these?

StephenButtolph: They should be `nil` afaict:

> // Command returns the Cmd struct to execute the named program with
// the given arguments.
//
// It sets only the Path and Args in the returned structure.

StephenButtolph (approved):

tests/fixture/tmpnet/node_process.go
```diff
@@ -112,7 +112,14 @@ func (p *NodeProcess) Start(w io.Writer) error {
 		return fmt.Errorf("failed to remove stale process context file: %w", err)
 	}
 
+	// All argument are provided in the flags file
```
darioush, line 115:
```suggestion
	// All arguments are provided in the flags file
```

darioush (approved):

PR #2861, opened 2024-03-21
nits

dhrubabasu:
* Removes `txBuilder` from `VM`
* Removes `atomicUtxosManager` from `VM`


vms/platformvm/service.go
```diff
@@ -381,7 +381,8 @@ func (s *Service) GetUTXOs(_ *http.Request, args *api.GetUTXOsArgs, response *ap
 			limit,
 		)
 	} else {
-		utxos, endAddr, endUTXOID, err = s.vm.atomicUtxosManager.GetAtomicUTXOs(
```
abi87, line 384: can we create this only once when we instantiace the `service` struct?

vms/platformvm/vm.go
```diff
@@ -85,8 +83,7 @@ type VM struct {
 	// Bootstrapped remembers if this chain has finished bootstrapping or not
 	bootstrapped utils.Atomic[bool]
 
-	txBuilder txbuilder.Builder
```
abi87, line 88: no client is using txBuilder, so I agree we should get rid of it as a vm member

vms/platformvm/vm.go
```diff
@@ -65,8 +64,7 @@ type VM struct {
 	*network.Network
 	validators.State
 
-	metrics            metrics.Metrics
-	atomicUtxosManager avax.AtomicUTXOManager
```
abi87, line 69: only `service` uses `atomicUtxosManager`, so `atomicUtxosManager` should be a service member

vms/platformvm/vm_regression_test.go
```diff
@@ -47,11 +47,12 @@ import (
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
 
 	blockexecutor "github.com/ava-labs/avalanchego/vms/platformvm/block/executor"
+	txbuilder "github.com/ava-labs/avalanchego/vms/platformvm/txs/builder"
 )
 
 func TestAddDelegatorTxOverDelegatedRegression(t *testing.T) {
 	require := require.New(t)
-	vm, _, _ := defaultVM(t, cortina)
+	vm, txBuilder, _, _ := defaultVM(t, cortina)
```
abi87, line 55: this is the one bit slightly annoying to me: why `defaultVM` should return the `txBuilder`?
Can we instead create the `txBuilder` only in the tests where we need it and leave defaultVM signature unchanged?

abi87 (changes requested):
I think it's a good change. I have only minor observations around handling txBuilder creation in tests.
Looking forward to merge this cleanup once we solve the comments!

vms/platformvm/vm_regression_test.go
dhrubabasu: over 50% of the tests use txBuilder (including `defaultVM` itself). I think we should keep it in there

abi87 (approved):

PR #2869, opened 2024-03-25
Reindex P-chain blocks

StephenButtolph:
## Why this should be merged

Reindexes the P-chain block mapping which was incorrectly skipped in #1719 (see: https://github.com/ava-labs/avalanchego/pull/1719#discussion_r1536714191).

## How this works

On first startup, the P-chain will iterate over the block store and update any mappings with `stateBlkBytes` to only store `blkBytes`.

## How this was tested

- [X] Added unit test

### DIFF
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -179,8 +179,8 @@ type State interface {
 	// Discard uncommitted changes to the database.
 	Abort()
 
-	// ReindexBlocks converts any block indicies using the legacy storage format
-	// to the new format. If this database has already updated the indicies,
+	// ReindexBlocks converts any block indices using the legacy storage format
+	// to the new format. If this database has already updated the indices,
 	// this function will return immediately, without iterating over the
 	// database.
 	//
@@ -2338,11 +2338,11 @@ func (s *state) ReindexBlocks(lock sync.Locker, log logging.Logger) error {
 	log.Info("starting block reindexing")
 
 	var (
-		startTime          = time.Now()
-		lastCommit         = startTime
-		nextUpdate         = startTime.Add(indexLogFrequency)
-		numIndiciesChecked = 0
-		numIndicesUpdated  = 0
+		startTime         = time.Now()
+		lastCommit        = startTime
+		nextUpdate        = startTime.Add(indexLogFrequency)
+		numIndicesChecked = 0
+		numIndicesUpdated = 0
 	)
 
 	for blockIterator.Next() {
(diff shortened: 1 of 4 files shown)


vms/platformvm/state/state_test.go
```diff
@@ -1295,10 +1297,112 @@ func requireEqualPublicKeysValidatorSet(
...
+		require.Equal(blk.ID(), gotBlk.ID())
+
+		gotBlk, isStateBlk, err = parseStoredBlock(blk.Bytes())
+		require.NoError(err)
+		require.False(isStateBlk)
+		require.Equal(blk.ID(), gotBlk.ID())
+	}
+}
+
+func TestReindexBlocks(t *testing.T) {
```
StephenButtolph, line 1326: This is the new test

vms/platformvm/vm.go
```diff
@@ -245,6 +245,15 @@ func (vm *VM) Initialize(
 	// [periodicallyPruneMempool] grabs the context lock.
 	go vm.periodicallyPruneMempool(execConfig.MempoolPruneFrequency)
 
+	go func() {
+		err := vm.state.ReindexBlocks(&vm.ctx.Lock, vm.ctx.Log)
+		if err != nil {
+			vm.ctx.Log.Warn("reindexing blocks failed",
```
StephenButtolph, line 251: This is marked as a `Warn` because shutting down the node mid-way through the indexing will report an error here. There isn't an easy way to exit this routine during `Shutdown`, because `Shutdown` has the context lock held, and this goroutine may be blocked on grabbing that lock.

vms/platformvm/state/state.go
```diff
@@ -2292,3 +2313,138 @@ func parseStoredBlock(blkBytes []byte) (block.Block, bool, error) {
...
+		blockIterator.Release()
+	}()
+
+	log.Info("starting block reindexing")
+
+	var (
+		startTime          = time.Now()
+		lastCommit         = startTime
+		nextUpdate         = startTime.Add(indexLogFrequency)
+		numIndiciesChecked = 0
```
dhrubabasu, line 2344:
```suggestion
		numIndicesChecked = 0
```

vms/platformvm/state/state.go
```diff
@@ -167,6 +179,14 @@ type State interface {
 	// Discard uncommitted changes to the database.
 	Abort()
 
+	// ReindexBlocks converts any block indicies using the legacy storage format
+	// to the new format. If this database has already updated the indicies,
+	// this function will return immediately, without iterating over the
+	// database.
```
dhrubabasu, line 185:
```suggestion
	// ReindexBlocks converts any block indices using the legacy storage format
	// to the new format. If this database has already updated the indices,
	// this function will return immediately, without iterating over the
	// database.
```

dhrubabasu (approved):
LGTM, just one small typo nit

darioush (approved):

PR #2870, opened 2024-03-26
add upgrade configs to tmpnet

ceyonur:
## Why this should be merged

We can construct e2e tests with later upgrades (precompile-activations, stateupgrades etc in Subnet-EVM) 

## How this works

Adds upgrade config to tmpnet similar to chain configs 

## How this was tested

Locally


ceyonur:
> @ceyonur Could this be tested against the cchain (coreth)? Ideally this should be testable in CI before merge.

No this can only be tested with subnet-evm (or any other VMs supporting upgrade json files)

PR #2871, opened 2024-03-26
`tmpnet`: Improve subnet configuration

marun:
## Why this should be merged

- Enable subnet validators to be a subnet of available validators. Previously tmpnet expected all validators to validate all subnets.

- Enable setting of non-default subnet configuration. Previously tmpnet did not support setting per-subnet configuration.

## How this was tested

- CI to validate that existing behavior isn't broken

## TODO

- [ ] Validate both changes with hypersdk PR

### DIFF
--- a/tests/fixture/tmpnet/network.go
+++ b/tests/fixture/tmpnet/network.go
@@ -370,6 +370,37 @@ func (n *Network) StartNode(ctx context.Context, w io.Writer, node *Node) error 
 	return nil
 }
 
+// Restart a single node.
+func (n *Network) RestartNode(ctx context.Context, w io.Writer, node *Node) error {
+	// Ensure the node reuses the same API port across restarts to ensure
+	// consistent labeling of metrics. Otherwise prometheus's automatic
+	// addition of the `instance` label (host:port) results in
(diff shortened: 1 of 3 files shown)


darioush (approved):

tests/fixture/tmpnet/network.go
```diff
@@ -501,10 +506,20 @@ func (n *Network) EnsureNodeConfig(node *Node) error {
...
 			config.GenesisFileKey:    n.getGenesisPath(),
 			config.ChainConfigDirKey: n.getChainConfigDir(),
-		})
+		}
+
+		// Only set the subnet dir if it exists or the node won't start.
+		subnetDir := n.getSubnetDir()
+		if _, err := os.Stat(subnetDir); err == nil {
+			defaultFlags[config.SubnetConfigDirKey] = subnetDir
+		} else if err != nil && !errors.Is(err, os.ErrNotExist) {
```
StephenButtolph, line 518: We're already guaranteed that `err != nil` in this check right?
```suggestion
		} else if !errors.Is(err, os.ErrNotExist) {
```

StephenButtolph (approved):

PR #2880, opened 2024-03-28
`ci`: Skip monitoring if secrets are not present

marun:
Metrics and logs can only be collected if the required secrets are present, but they will not be available to jobs executing against PRs for forks.


.github/workflows/ci.yml
```diff
@@ -68,18 +68,21 @@ jobs:
...
         run: ./scripts/build.sh -r
       - name: Start prometheus
+        if: secrets.PROMETHEUS_ID != ''
         shell: bash
         run: bash -x ./scripts/run_prometheus.sh
         env:
           PROMETHEUS_ID: ${{ secrets.PROMETHEUS_ID }}
           PROMETHEUS_PASSWORD: ${{ secrets.PROMETHEUS_PASSWORD }}
       - name: Start promtail
+        if: secrets.PROMETHEUS_ID != ''
```
StephenButtolph, line 78: Would it make more sense to check for the vars that are used by the script? I think that would be `LOKI_ID` here.

either way lgtm

StephenButtolph (approved):

joshua-kim (approved):

PR #2882, opened 2024-03-28
Optimize encodeUint

StephenButtolph:
## Why this should be merged

This code is simpler and faster.

Before:
```
Benchmark_EncodeUint/1-12  		19007980	        60.29 ns/op	      24 B/op	       1 allocs/op
Benchmark_EncodeUint/2-12  		20342215	        60.37 ns/op	      24 B/op	       1 allocs/op
Benchmark_EncodeUint/32-12 		20273323	        60.40 ns/op	      24 B/op	       1 allocs/op
Benchmark_EncodeUint/1024-12         	19789828	        60.73 ns/op	      24 B/op	       1 allocs/op
Benchmark_EncodeUint/32768-12        	19555718	        61.48 ns/op	      24 B/op	       1 allocs/op
```

After:
```
Benchmark_EncodeUint/1-12  		179371581	         6.680 ns/op	       0 B/op	       0 allocs/op
Benchmark_EncodeUint/2-12  		179073162	         7.116 ns/op	       0 B/op	       0 allocs/op
Benchmark_EncodeUint/32-12 		181321675	         6.617 ns/op	       0 B/op	       0 allocs/op
Benchmark_EncodeUint/1024-12         	180010732	         6.657 ns/op	       0 B/op	       0 allocs/op
Benchmark_EncodeUint/32768-12        	163931486	         7.240 ns/op	       0 B/op	       0 allocs/op
```

## How this works

This allows the intermediate buffer to be allocated on the stack rather than the heap. This completely removes the memory allocation and doesn't require the sync pool.

## How this was tested

- [X] Added benchmark
- [X] CI

### DIFF
--- a/x/merkledb/codec.go
+++ b/x/merkledb/codec.go
@@ -11,7 +11,6 @@ import (
 	"math"
 	"math/bits"
 	"slices"
-	"sync"
 
 	"golang.org/x/exp/maps"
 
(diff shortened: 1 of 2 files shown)


x/merkledb/codec.go
```diff
@@ -75,22 +74,12 @@ type decoder interface {
 }
 
 func newCodec() encoderDecoder {
```
StephenButtolph, line 76: I think we're going to want to completely remove this type (now that it's an empty unexported struct). But I think we can leave that to a followup.

aaronbuchwald (approved):

darioush (approved):

dhrubabasu (approved):

PR #2889, opened 2024-03-29
precompute varint lens

danlaine:
## Why this should be merged

TODO 

## How this works

TODO 

## How this was tested

TODO


StephenButtolph:
This was a great idea, but it doesn't seem to improve the `hashNode` benchmark.

PR #2899, opened 2024-04-01
Optimize key creation in hashing

StephenButtolph:
## Why this should be merged

Before:
```
Benchmark_HashChangedNodes/1-12  		 3001608	       398.0 ns/op	     160 B/op	       4 allocs/op
Benchmark_HashChangedNodes/10-12 		   54448	     23107   ns/op	     810 B/op	      34 allocs/op
Benchmark_HashChangedNodes/100-12         	   10000	    106800   ns/op	    3658 B/op	     252 allocs/op
Benchmark_HashChangedNodes/1000-12        	    1274	   1015625   ns/op	   30612 B/op	    2500 allocs/op
Benchmark_HashChangedNodes/10000-12       	     139	   8182703   ns/op	  255007 B/op	   22840 allocs/op
Benchmark_HashChangedNodes/100000-12      	      13	  82452003   ns/op	 2785910 B/op	  248196 allocs/op
```

After:
```
Benchmark_HashChangedNodes/1-12  		 3518144	       337.8 ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/10-12 		   60412	     19177   ns/op	     721 B/op	      25 allocs/op
Benchmark_HashChangedNodes/100-12         	   16171	     77663   ns/op	    2096 B/op	      83 allocs/op
Benchmark_HashChangedNodes/1000-12        	    1593	    788333   ns/op	   15686 B/op	     754 allocs/op
Benchmark_HashChangedNodes/10000-12       	     184	   6684802   ns/op	   87454 B/op	    4186 allocs/op
Benchmark_HashChangedNodes/100000-12      	      15	  71014036   ns/op	 1025369 B/op	   61043 allocs/op
```

## How this works

- Avoid performing a memory allocation in `ToToken`.
- Avoid performing a memory allocation in `Extend`.
- Avoid performing a memcopy in `Extend`.

## How this was tested

- [X] New unit tests
- [X] New benchmark
- [X] CI


x/merkledb/view.go
```diff
@@ -290,12 +290,63 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+
+	var (
+		maxBytesNeeded = bytesNeeded(maxBitLength)
+		// keyBuffer is allocated onto the heap because it is dynamically sized.
+		keyBuffer = make([]byte, maxBytesNeeded)
+		// childBuffer is allocated on the stack.
+		childBuffer = make([]byte, 1)
+		dualIndex   = dualBitIndex(v.tokenSize)
+		bytesForKey = bytesNeeded(n.key.length)
+		// We track the last byte used by the key so that we can reset the
```
danlaine, line 314: nit: slightly more clear which key we mean here
```suggestion
		// We track the last byte used of [n.key] so that we can reset the
```

x/merkledb/view.go
```diff
@@ -290,12 +290,63 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		// iterate over the children.
+		copy(keyBuffer, n.key.value)
+		lastKeyByte = keyBuffer[bytesForKey-1]
+	}
 
 	for childIndex, childEntry := range n.children {
-		childEntry := childEntry // New variable so goroutine doesn't capture loop variable.
-		childKey := n.key.Extend(ToToken(childIndex, v.tokenSize), childEntry.compressedKey)
+		childBuffer[0] = childIndex << dualIndex
+		childByteKey := Key{
```
danlaine, line 332: nit: I think this is slightly more clear but I'm ok with it as is too
```suggestion
		childIndexAsKey := Key{
```

danlaine (approved):

x/merkledb/view.go
```diff
@@ -290,12 +290,63 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		totalBitLength := n.key.length + v.tokenSize + childEntry.compressedKey.length
+		buffer := keyBuffer[:bytesNeeded(totalBitLength)]
+		// Make sure the last byte of the key is originally set correctly
+		if bytesForKey > 0 {
+			buffer[bytesForKey-1] = lastKeyByte
+		}
+		extendIntoBuffer(buffer, childByteKey, n.key.length)
+		extendIntoBuffer(buffer, childEntry.compressedKey, n.key.length+v.tokenSize)
+		childKey := Key{
+			value:  byteSliceToString(buffer),
```
danlaine, line 346: Are we violating the invariant of `byteSliceToString` since `buffer` has bytes from `keyBuffer` which is reused?

```go
// Invariant: The input []byte must not be modified.
```

danlaine: Stephen pointed out this only has to be upheld while the string is in usage, and this string is no longer used by the time we change `keyBuffer`

danlaine (approved):

x/merkledb/view.go
```diff
@@ -290,12 +290,66 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		// this byte.
+		lastKeyByte byte
+
+		// We use [wg] to wait until all descendants of [n] have been updated.
+		wg sync.WaitGroup
+	)
+
+	if bytesForKey > 0 {
+		// We can just copy this node's key once. It doesn't change as we
+		// iterate over the children.
```
dhrubabasu, line 324: nit: this wording feels a bit too informal with "just", maybe this?

```suggestion
		// We only need to copy this node's key once because it does not change as we iterate over the children.
```

dhrubabasu (approved):
one small wording nit

PR #2902, opened 2024-04-01
Reuse key buffers during hashing

StephenButtolph:
## Why this should be merged

This removes the last memory allocation inside of `HashChangedNodes` (except for the allocation performed for tracing.

Before:
```
Benchmark_HashChangedNodes/1-12  		 3299274	       361.6 ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/10-12 		  407604	      2992   ns/op	     161 B/op	       5 allocs/op
Benchmark_HashChangedNodes/100-12         	   39985	     30044   ns/op	     280 B/op	      19 allocs/op
Benchmark_HashChangedNodes/1000-12        	    3674	    332787   ns/op	    2746 B/op	     293 allocs/op
Benchmark_HashChangedNodes/10000-12       	     356	   3489978   ns/op	   14262 B/op	    1573 allocs/op
Benchmark_HashChangedNodes/100000-12      	      19	  59557794   ns/op	  355809 B/op	   37157 allocs/op
```

After:
```
Benchmark_HashChangedNodes/1-12  		 3664717	       337.9 ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/10-12 		  438346	      2842   ns/op	     160 B/op	       4 allocs/op
Benchmark_HashChangedNodes/100-12         	   42628	     28271   ns/op	     272 B/op	      11 allocs/op
Benchmark_HashChangedNodes/1000-12        	    3680	    321290   ns/op	    2464 B/op	     148 allocs/op
Benchmark_HashChangedNodes/10000-12       	     369	   3150519   ns/op	   12704 B/op	     788 allocs/op
Benchmark_HashChangedNodes/100000-12      	      22	  52984970   ns/op	  297376 B/op	   18580 allocs/op
```

Honestly it's getting pretty hard to benchmark these changes... The variability of the benchmark is higher than the improvements. This seems to give as much as another 12% reduction in runtime... But I'm not especially confident in that number. I am however confident that after this PR and #2901 are merged we'll see:

```
Benchmark_HashChangedNodes/1-12  		 3578389	       333.3 ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/10-12 		  432597	      2875   ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/100-12         	   41547	     28324   ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/1000-12        	    3766	    315036   ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/10000-12       	     345	   3263585   ns/op	     144 B/op	       3 allocs/op
Benchmark_HashChangedNodes/100000-12      	      24	  51812141   ns/op	     144 B/op	       3 allocs/op
```

## How this works

- Maintains a persistent set of keyBuffers such that `HashChangedNodes` only performs allocations during the first few calls.
- Replaces the semaphore with a byte pool that implements the buffer sharing.
- Shares the keyBuffer with recursive calls to `hashChangedNode`.
- Avoids copying `n.key.value` on each call to `hashChangedNode`.

## How this was tested

- [X] CI
- [X] Added benchmarks around the byte pool


x/merkledb/view.go
```diff
@@ -281,32 +281,27 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
-	// only allocating a single slice for all of the keys.
-	var maxBitLength int
-	for _, childEntry := range n.children {
-		totalBitLength := n.key.length + v.tokenSize + childEntry.compressedKey.length
-		maxBitLength = max(maxBitLength, totalBitLength)
-	}
-
+//
+// Invariant: [keyBuffer] must be populated with [n]'s key and have a large
+// enough capacity to contain all of [n]'s child keys.
```
danlaine, line 302:
```suggestion
// enough length to contain all of [n]'s child keys.
```

x/merkledb/view.go
```diff
@@ -352,27 +346,62 @@ func (v *view) hashChangedNode(n *node) ids.ID {
 			// This child wasn't changed.
 			continue
 		}
-		childEntry.hasValue = childNodeChange.after.hasValue()
+
+		childNode := childNodeChange.after
+		childEntry.hasValue = childNode.hasValue()
+
+		// If there are no children of the childNode, we can avoid constructing
+		// a new [innerKeyBuffer].
```
danlaine, line 354: nit: stale reference to non-existent var `innerKeyBuffer`

x/merkledb/view.go
```diff
@@ -281,32 +281,27 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		v.changes.rootID = root.calculateID(v.db.metrics)
+		return
+	}
 
-	v.changes.rootID = v.hashChangedNode(v.root.Value())
+	// Allocate [keyBuffer] and populate it with the root node's key.
+	keyBuffer := v.db.hashNodesKeyPool.Acquire()
+	keyBuffer = v.setKeyBuffer(root, keyBuffer)
+	v.changes.rootID, keyBuffer = v.hashChangedNode(root, keyBuffer)
+	v.db.hashNodesKeyPool.Release(keyBuffer)
```
danlaine, line 295: nit: should we do this in a defer? Or is it like this for performance?

danlaine (approved):
just nits

x/merkledb/view.go
StephenButtolph: It is done like this because `keyBuffer` isn't known until after `hashChangedNode` returns.

x/merkledb/bytes_pool.go
```diff
@@ -0,0 +1,67 @@
...
+		panic("release of unacquired semaphore")
+	}
+
+	p.bytesLock.Lock()
+	defer p.bytesLock.Unlock()
+
+	p.bytes = append(p.bytes, b)
+}
+
+func setLength(b []byte, size int) []byte {
```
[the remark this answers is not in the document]
StephenButtolph: It is used in `view.go` - so I think it makes sense to move it there.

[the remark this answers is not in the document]
danlaine: Woops idk why I thought this was only used in `db.go`

x/merkledb/view.go
```diff
@@ -281,47 +281,143 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+// setKeyBuffer expands [keyBuffer] to have sufficient size for any of [n]'s
+// child keys and populates [n]'s key into [keyBuffer]. If [keyBuffer] already
+// has sufficient size, this memory will not perform any memory allocations.
+func (v *view) setKeyBuffer(n *node, keyBuffer []byte) []byte {
+	keyBuffer = v.setLengthForChildren(n, keyBuffer)
+	copy(keyBuffer, n.key.value)
+	return keyBuffer
+}
+
+// setKeyBuffer expands [keyBuffer] to have sufficient size for any of [n]'s
```
darioush, line 404:
```suggestion
// setLengthForChildren expands [keyBuffer] to have sufficient size for any of [n]'s
```

x/merkledb/view.go
```diff
@@ -281,47 +281,143 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		// Because [keyBuffer] may have been modified in a prior iteration of
+		// this loop, it is not guaranteed that it's length is at least
+		// [bytesNeeded(totalBitLength)]. However, that's fine. The below
+		// slicing would only panic if the buffer didn't have sufficient
+		// capacity.
+		keyBuffer = keyBuffer[:bytesNeeded(totalBitLength)]
+		// We don't need to copy this node's key. It's assumed to already be
+		// correct; except for the last byte. We must make sure the last byte of
+		// the key is originally set correctly because extendIntoBuffer may OR
+		// the last byte of the key.
```
darioush, line 340:
```suggestion
		// the key is set correctly because extendIntoBuffer does not modify
		// the first [bytesForKey-1] bytes of [keyBuffer], however it may OR bits
		// from the extension and overwrite the last byte.
```
perhaps some more explanation seems clearer to me.

x/merkledb/view.go
```diff
@@ -281,47 +281,143 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
+		wg sync.WaitGroup
+	)
+	if bytesForKey > 0 {
+		lastKeyByte = keyBuffer[bytesForKey-1]
+	}
 
 	for childIndex, childEntry := range n.children {
-		childEntry := childEntry // New variable so goroutine doesn't capture loop variable.
-		childKey := n.key.Extend(ToToken(childIndex, v.tokenSize), childEntry.compressedKey)
+		childBuffer[0] = childIndex << dualIndex
```
darioush, line 322: could we add a comment like:
```suggestion
		// This loop is optimized to avoid allocations when calculating
		// the [childKey] by reusing [keyBuffer] and leaving the first
		// [bytesForKey-1] bytes unmodified.
		
		childBuffer[0] = childIndex << dualIndex
```

x/merkledb/view.go
```diff
@@ -281,47 +281,143 @@ func (v *view) hashChangedNodes(ctx context.Context) {
...
 }
 
 // Calculates the ID of all descendants of [n] which need to be recalculated,
 // and then calculates the ID of [n] itself.
-func (v *view) hashChangedNode(n *node) ids.ID {
-	// We use [wg] to wait until all descendants of [n] have been updated.
-	var wg sync.WaitGroup
+//
+// Invariant: [keyBuffer] must be populated with [n]'s key and have sufficient
+// length to contain any of [n]'s child keys.
```
darioush, line 302: Can we add a comment about the return value and its use?

darioush (approved):
I added some comments about the comments.

PR #2904, opened 2024-04-03
Introduce generic linked list

StephenButtolph:
## Why this should be merged

Improves the performance of `linked.Hashmap#Put` by ~50%.

Before:
```golang
Benchmark_Hashmap-12    	 6913144	       166.5 ns/op	      80 B/op	       2 allocs/op
```

After:
```golang
Benchmark_Hashmap-12    	13979869	        82.58 ns/op	       0 B/op	       0 allocs/op
```

## How this works

Previously, the `linked.Hashmap` depended on the "container/list" library. This resulted in 2 memory allocations.
1. For the creation of the `keyValue` struct that was wrapped into an `any` interface.
2. For the creation of the `list.Element` struct enforced by the `list` functions.

By implementing our own version of the linked list, we are able to use generics to avoid the memory allocation for the interface conversion.

Additionally, our version of the linked list allows allocating the `Element` externally to the list. This allows the `Hashmap` to maintain a free list (which isn't an increase in memory pressure, because the map never frees allocated space either).

## How this was tested

- [X] Added unit tests
- [X] Added `Hashmap` benchmark
- [X] CI


utils/linked/list.go
```diff
@@ -0,0 +1,178 @@
...
+	length   int
+}
+
+// NewList creates a new doubly linked list.
+func NewList[T any]() *List[T] {
+	l := &List[T]{}
+	l.sentinel.next = &l.sentinel
+	l.sentinel.prev = &l.sentinel
+	l.sentinel.list = l
+	return l
```
abi87, line 51: q: I see `containers/list` has `Init` and `lazyInit` methods. I guess this means we swear to always use the New method to build a list, right?
If so we may add this to the comment maybe?

utils/linked/list.go
```diff
@@ -0,0 +1,178 @@
...
+func (l *List[T]) Back() *ListElement[T] {
+	if l.length == 0 {
+		return nil
+	}
+	return l.sentinel.prev
+}
+
+// PushFront inserts e at the front of l.
+// If e is already in a list, l is not modified.
+func (l *List[T]) PushFront(e *ListElement[T]) {
```
abi87, line 74: I understand this is supposed to be a drop in replacement of containers/list, still let me ask a question.

- Wouldn't it be more natural to have methods to add a value (`PushFront/Back` `InsertBefore/After`) accepting a value T instead of a ListElement[T]?
- Wouldn't it be more idiomatic to signal "no next/prev element" via an error rather than a nil pointer? I guess here performance considerations may have prevailed

utils/linked/list.go
```diff
@@ -0,0 +1,178 @@
...
+//
+// This datastructure is designed to be an almost complete drop-in replacement
+// for the standard library's "container/list".
+//
+// The primary design change is to remove all memory allocations from the list
+// definition. This allows these lists to be used in performance critical paths.
+type List[T any] struct {
+	// sentinel is only used as a placeholder to avoid complex nil checks.
+	// sentinel.Value is never used.
+	sentinel ListElement[T]
```
abi87, line 41: nit: I like sentinel better than root

abi87 (approved):
just nits and questions, but LGTM

utils/linked/list.go
StephenButtolph: Thanks! Me too. What is the suggestion here?

utils/linked/list.go
StephenButtolph: All of the methods take in `ListElement` rather than `T` to allow the caller to manage all memory allocations. I'll add some helpers to expose the API with allocations.

StephenButtolph: Aside from the `ListElement` vs `T` change, I kept the signature the same as the stdlib to allow ease of switching. (Which is why we use `nil` to signal no element rather than a `bool`.

utils/linked/list.go
StephenButtolph: added

utils/linked/list.go
abi87: No action really, just comparing naming with what there is in `containers/list`, where `root` is used instead of `sentinel`

PR #2913, opened 2024-04-04
Fix MerkleDB crash recovery

StephenButtolph:
## Why this should be merged

Fixes a bug that prevented the merkledb from recovering from a process crash.

## How this works

- Calling `initializeRoot` on a corrupt DB does not error gracefully.
- This PR ensures that `initializeRoot` is never called on a corrupt DB.
- This PR guarantees that the `rootKey` on disk is correct if the DB is marked as valid.

## How this was tested

- [X] Added unit test.

### DIFF
--- a/x/merkledb/db.go
+++ b/x/merkledb/db.go
@@ -44,6 +44,19 @@ var (
 	valueNodePrefix        = []byte{1}
 	intermediateNodePrefix = []byte{2}
 
+	// cleanShutdownKey is used to flag that the database did (or did not)
+	// previously shutdown correctly.
+	//
+	// If this key has value [hadCleanShutdown] it must be true that all
+	// intermediate nodes of the trie are correctly populated on disk and that
+	// the [rootDBKey] has the correct key for the root node.
+	//
+	// If this key has value [didNotHaveCleanShutdown] the intermediate nodes of
+	// the trie may not be correct and the [rootDBKey] may not exist or point to
+	// a node that node longer exists.
+	//
+	// Regardless of the value of [cleanShutdownKey], the value nodes must
+	// always be persisted correctly.
 	cleanShutdownKey        = []byte(string(metadataPrefix) + "cleanShutdown")
 	rootDBKey               = []byte(string(metadataPrefix) + "root")
 	hadCleanShutdown        = []byte{1}


x/merkledb/db.go
```diff
@@ -1166,12 +1167,11 @@ func (db *merkleDB) invalidateChildrenExcept(exception *view) {
 // Otherwise leave [db.root] as Nothing.
 func (db *merkleDB) initializeRoot() error {
```
StephenButtolph, line 1168: The changes in this function aren't required... But I think they improve the code quality.

x/merkledb/db.go
```diff
@@ -278,9 +278,25 @@ func newDatabase(
...
+			return nil, err
+		}
+	} else {
+		if err := trieDB.initializeRoot(); err != nil {
+			return nil, err
+		}
+	}
 
 	// add current root to history (has no changes)
 	trieDB.history.record(&changeSummary{
```
StephenButtolph, line 302: Should we clear the history here? If we're recovering from a crash - we end up keeping the history of the rebuild.

x/merkledb/db.go
StephenButtolph, line 1168: The bug here was that this function assumes that the DB is in a consistent state but we called it prior to repairing the DB.

x/merkledb/db.go
```diff
@@ -460,8 +461,26 @@ func (db *merkleDB) Close() error {
...
+	// Write the root key
+	if db.root.IsNothing() {
+		err = batch.Delete(rootDBKey)
+	} else {
+		rootKey := encodeKey(db.root.Value().key)
+		err = batch.Put(rootDBKey, rootKey)
+	}
+	if err != nil {
+		return err
+	}
```
StephenButtolph, line 477: We must write the root key here to avoid the case where:
1. An empty database is corrupted.
2. Repairing an empty database doesn't update the root key
3. Closing the now (partially) repaired database updates the clean shutdown flag (but leaves the root key as invalid).

x/merkledb/db.go
```diff
@@ -978,13 +997,7 @@ func (db *merkleDB) commitChanges(ctx context.Context, trieToCommit *view) error
 	// Update root in database.
 	db.root = changes.rootChange.after
 	db.rootID = changes.rootID
-
-	if db.root.IsNothing() {
-		return db.baseDB.Delete(rootDBKey)
-	}
-
-	rootKey := encodeKey(db.root.Value().key)
-	return db.baseDB.Put(rootDBKey, rootKey)
```
StephenButtolph, line 987: Because we write the root key during shutdown, this write is no longer needed.

x/merkledb/db.go
danlaine: > Unrelated to this PR - but should we clear the history here?

That seems reasonable to me. I don't think it makes sense to have the history from the rebuild available.

x/merkledb/db_test.go
```diff
@@ -1287,3 +1287,46 @@ func TestGetChangeProofEmptyRootID(t *testing.T) {
...
+		context.Background(),
+		baseDB,
+		newDefaultConfig(),
+		&mockMetrics{},
+	)
+	require.NoError(err)
+
+	value, err := newMerkleDB.Get([]byte("is this"))
+	require.NoError(err)
+	require.Equal([]byte("hope"), value)
```
danlaine, line 1323: butterfly meme: "is this hope?"

danlaine (approved):

danlaine:
We should probably add an invariant in a comment that documents this:

```
// Invariant: If [cleanShutdownKey] has [hadCleanShutdown] the root on disk is correct.
```

dhrubabasu (approved):

PR #2914, opened 2024-04-04
Enable creation of multi-arch docker images

marun:
Currently images are only published for amd64. To enable testing and development on arm64, this PR:
 - Enables build and publication of multi-arch images that support both amd64 and arm64
    - Non-native builds should be fast on either arch thanks to cross-compilation
 - Retains the existing single-arch local image by default
 - Adds a new test job to validate both single and multi-arch image builds
   - The test checks that `/avalanchego/build/avalanchego --version` runs without error, indicating that the image and its binary are sane for the intended arch

This ended up being a lot more work than I expected - lots of ways to get multiarch builds wrong and the different options (docker build + docker manifest vs docker buildx) each having their own gotchas. The test script ended up being invaluable since at many stages the images were building without error but ending up invalid for the intended arch due to subtle problems with build configuration in the Dockerfile. 

## TODO:
 - [ ] POST-MERGE: Require the new `Image build` job


StephenButtolph (approved):
lgtm - images seem to be getting pushed to the dockerhub repo correctly 👍

PR #2915, opened 2024-04-04
Remove cancellation for Send*AppRequest messages

StephenButtolph:
## Why this should be merged

Fixes issue that caused unexpected `AppResponse` / `AppRequestFailed` messages to be reported.

## How this works

The `rpcChainVM` does not guarantee that returning an error from these methods implies that the messages was not sent to the peer.

For an error to be reported here:
- Either the `rpcChainVM` is unable to provide a reliable connection (in which case the plugin will FATAL)
- The context was cancelled

## How this was tested

- [x] Added unit test
- [X] CI

### DIFF
--- a/snow/engine/common/sender.go
+++ b/snow/engine/common/sender.go
@@ -167,11 +167,16 @@ type QuerySender interface {
 // NetworkAppSender sends VM-level messages to nodes in the network.
 type NetworkAppSender interface {
 	// Send an application-level request.
-	// A nil return value guarantees that for each nodeID in [nodeIDs],
-	// the VM corresponding to this AppSender eventually receives either:
+	//
+	// The VM corresponding to this AppSender may receive either:
 	// * An AppResponse from nodeID with ID [requestID]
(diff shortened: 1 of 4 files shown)


joshua-kim (approved):

darioush (approved):

snow/engine/common/sender.go
```diff
@@ -192,12 +197,16 @@ type CrossChainAppSender interface {
...
 	// * A CrossChainAppResponse from [chainID] with ID [requestID]
 	// * A CrossChainAppRequestFailed from [chainID] with ID [requestID]
-	// Exactly one of the above messages will eventually be received from
-	// [chainID].
+	//
+	// A nil return value guarantees that the VM corresponding to this
+	// CrossChainAppSender will eventually receive exactly one of the above
+	// messages.
+	//
+	// A non-nil return value guarantees that the VM corresponding to this
```
patrick-ogrady, line 208: Good comment 🙏

PR #2920, opened 2024-04-07
Implement `utils.BytesPool` to replace `sync.Pool` for byte slices

StephenButtolph:
## Why this should be merged

This PR improves the average workload allocation speed by `~70%`. 
This PR improves the worst case workload allocation speed by `~91%`

This PR implements a bucketed `sync.Pool` for byte slices. It has comparable performance than the (fixed) approach used in the prefixdb/merkledb for most workloads. However, the performance is workload independent. The prefixdb/merkledb approach does not handle the ascending workload efficiently. 

Performance of just allocating the requested bytes:
```
BenchmarkBytesPool_Constant/0-12    	305457604	         4.158 ns/op	         0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/8-12    	76623870	        15.83  ns/op	         8 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/16-12   	44998242	        25.46  ns/op	        16 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/32-12   	49072311	        25.67  ns/op	        32 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/64-12   	37186384	        33.22  ns/op	        64 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/256-12  	 9332716	       134.0   ns/op	       256 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/2048-12 	 2156217	       569.7   ns/op	      2048 B/op	       1 allocs/op
BenchmarkBytesPool_Descending-12    	       2	 973057916     ns/op	5309609360 B/op	  100507 allocs/op
BenchmarkBytesPool_Ascending-12     	       3	 498653917     ns/op	5309465072 B/op	  100112 allocs/op
BenchmarkBytesPool_Random-12        	     224	   5282923     ns/op	  51456611 B/op	    1001 allocs/op
```

Using the prefixdb/merkledb `sync.Pool` approach:
```
BenchmarkBytesPool_Constant/0-12    	23733140	        46.89 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/8-12    	25610904	        47.00 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/16-12   	25540450	        48.19 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/32-12   	25697008	        47.32 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/64-12   	24564198	        47.91 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/256-12  	25694967	        47.75 ns/op	      24 B/op	       1 allocs/op
BenchmarkBytesPool_Constant/2048-12 	24311798	        49.26 ns/op	      26 B/op	       1 allocs/op
BenchmarkBytesPool_Descending-12    	     235	      5059765 ns/op	 3244221 B/op	  100041 allocs/op
BenchmarkBytesPool_Ascending-12     	      79	     15944859 ns/op    203357893 B/op	  106367 allocs/op
BenchmarkBytesPool_Random-12        	   21878	        58405 ns/op	   57793 B/op	    1000 allocs/op
```

Using the prefixdb/merkledb `sync.Pool` approach (with `*[]byte`):
```
BenchmarkBytesPool_Constant/0-12    	80639740	        16.84 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/8-12    	83390275	        14.50 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/16-12   	84568580	        14.32 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/32-12   	83906548	        14.48 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/64-12   	84178773	        14.37 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/256-12  	82249292	        14.39 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/2048-12 	73301838	        14.35 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Descending-12    	     818	      1455992 ns/op	    9124 B/op	       0 allocs/op
BenchmarkBytesPool_Ascending-12     	     565	      2135166 ns/op	 9619777 B/op	     365 allocs/op
BenchmarkBytesPool_Random-12        	   76890	        15618 ns/op	     530 B/op	       0 allocs/op
```

Using the bucketed `sync.Pool` approach:
```
BenchmarkBytesPool_Constant/0-12    	85004992	        14.09 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/8-12    	84486711	        14.18 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/16-12   	86129293	        14.03 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/32-12   	89044167	        13.99 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/64-12   	87703000	        13.98 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/256-12  	88138891	        13.90 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Constant/2048-12 	85855333	        14.04 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesPool_Descending-12    	     835	      1411102 ns/op	     670 B/op	       0 allocs/op
BenchmarkBytesPool_Ascending-12     	     855	      1404976 ns/op	     349 B/op	       0 allocs/op
BenchmarkBytesPool_Random-12        	   85389	        14161 ns/op	       3 B/op	       0 allocs/op
```

## How this works

This PR implements buckets of `sync.Pool`s that return byte slices that are of size `2^n -1`. This allows us to allocate slices that are always sufficient size. This isn't too memory inefficient, it'll allocate slices that are at most 2x larger than requested.

## How this was tested

- [X] Added unit tests
- [X] Added benchmarks
- [X] CI

### DIFF
--- a/database/prefixdb/db.go
+++ b/database/prefixdb/db.go
@@ -9,11 +9,8 @@ import (
 	"sync"
 
 	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/utils"
 	"github.com/ava-labs/avalanchego/utils/hashing"
-)
-
-const (
-	defaultBufCap = 256
 )
 
 var (
@@ -26,9 +23,8 @@ var (
 // a unique value.
 type Database struct {
 	// All keys in this db begin with this byte slice
-	dbPrefix []byte
-	// Holds unused []byte
-	bufferPool sync.Pool
+	dbPrefix   []byte
+	bufferPool *utils.BytesPool
 
 	// lock needs to be held during Close to guarantee db will not be set to nil
 	// concurrently with another operation. All other operations can hold RLock.
@@ -40,13 +36,9 @@ type Database struct {
 
 func newDB(prefix []byte, db database.Database) *Database {
 	return &Database{
-		dbPrefix: prefix,
-		db:       db,
-		bufferPool: sync.Pool{
-			New: func() interface{} {
-				return make([]byte, 0, defaultBufCap)
-			},
-		},
+		dbPrefix:   prefix,
+		db:         db,
+		bufferPool: utils.NewBytesPool(),
 	}
 }
 
@@ -91,9 +83,6 @@ func PrefixKey(prefix, key []byte) []byte {
 	return prefixedKey
 }
 
-// Assumes that it is OK for the argument to db.db.Has
-// to be modified after db.db.Has returns
-// [key] may be modified after this method returns.
 func (db *Database) Has(key []byte) (bool, error) {
 	db.lock.RLock()
 	defer db.lock.RUnlock()
@@ -102,14 +91,11 @@ func (db *Database) Has(key []byte) (bool, error) {
 		return false, database.ErrClosed
 	}
 	prefixedKey := db.prefix(key)
-	has, err := db.db.Has(prefixedKey)
-	db.bufferPool.Put(prefixedKey)
-	return has, err
+	defer db.bufferPool.Put(prefixedKey)
+
+	return db.db.Has(*prefixedKey)
 }
 
-// Assumes that it is OK for the argument to db.db.Get
-// to be modified after db.db.Get returns.
-// [key] may be modified after this method returns.
 func (db *Database) Get(key []byte) ([]byte, error) {
 	db.lock.RLock()
 	defer db.lock.RUnlock()
@@ -118,15 +104,11 @@ func (db *Database) Get(key []byte) ([]byte, error) {
 		return nil, database.ErrClosed
 	}
 	prefixedKey := db.prefix(key)
-	val, err := db.db.Get(prefixedKey)
-	db.bufferPool.Put(prefixedKey)
-	return val, err
+	defer db.bufferPool.Put(prefixedKey)
+
+	return db.db.Get(*prefixedKey)
 }
 
-// Assumes that it is OK for the argument to db.db.Put
-// to be modified after db.db.Put returns.
-// [key] can be modified after this method returns.
-// [value] should not be modified.
 func (db *Database) Put(key, value []byte) error {
 	db.lock.RLock()
 	defer db.lock.RUnlock()
@@ -135,14 +117,11 @@ func (db *Database) Put(key, value []byte) error {
 		return database.ErrClosed
 	}
 	prefixedKey := db.prefix(key)
-	err := db.db.Put(prefixedKey, value)
-	db.bufferPool.Put(prefixedKey)
-	return err
+	defer db.bufferPool.Put(prefixedKey)
+
+	return db.db.Put(*prefixedKey, value)
 }
 
-// Assumes that it is OK for the argument to db.db.Delete
-// to be modified after db.db.Delete returns.
-// [key] may be modified after this method returns.
 func (db *Database) Delete(key []byte) error {
 	db.lock.RLock()
 	defer db.lock.RUnlock()
@@ -151,9 +130,9 @@ func (db *Database) Delete(key []byte) error {
 		return database.ErrClosed
 	}
 	prefixedKey := db.prefix(key)
-	err := db.db.Delete(prefixedKey)
-	db.bufferPool.Put(prefixedKey)
-	return err
+	defer db.bufferPool.Put(prefixedKey)
+
+	return db.db.Delete(*prefixedKey)
 }
 
 func (db *Database) NewBatch() database.Batch {
@@ -186,15 +165,17 @@ func (db *Database) NewIteratorWithStartAndPrefix(start, prefix []byte) database
 			Err: database.ErrClosed,
 		}
 	}
+
 	prefixedStart := db.prefix(start)
+	defer db.bufferPool.Put(prefixedStart)
+
 	prefixedPrefix := db.prefix(prefix)
-	it := &iterator{
-		Iterator: db.db.NewIteratorWithStartAndPrefix(prefixedStart, prefixedPrefix),
+	defer db.bufferPool.Put(prefixedPrefix)
+
+	return &iterator{
+		Iterator: db.db.NewIteratorWithStartAndPrefix(*prefixedStart, *prefixedPrefix),
 		db:       db,
 	}
-	db.bufferPool.Put(prefixedStart)
-	db.bufferPool.Put(prefixedPrefix)
-	return it
 }
 
 func (db *Database) Compact(start, limit []byte) error {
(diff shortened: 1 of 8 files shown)


x/merkledb/db.go
```diff
@@ -1341,33 +1337,17 @@ func (db *merkleDB) getTokenSize() int {
...
-	copy(prefixedKey[prefixLen:], key)
+	prefixedKey := bufferPool.Get(keyLen)
+	copy(*prefixedKey, prefix)
+	copy((*prefixedKey)[prefixLen:], key)
 	return prefixedKey
 }
 
-// Returns a []byte from [bufferPool] with length exactly [size].
-// The []byte is not guaranteed to be zeroed.
-func getBufferFromPool(bufferPool *sync.Pool, size int) []byte {
```
StephenButtolph, line 1357: This was so jank - super happy to remove it 🙏

x/merkledb/db.go
```diff
@@ -260,13 +260,9 @@ func newDatabase(
...
 	}
 
-	// Share a sync.Pool of []byte between the intermediateNodeDB and valueNodeDB
-	// to reduce memory allocations.
-	bufferPool := &sync.Pool{
-		New: func() interface{} {
-			return make([]byte, 0, defaultBufferLength)
-		},
-	}
+	// Share a bytes poolbetween the intermediateNodeDB and valueNodeDB to
```
dhrubabasu, line 263:
```suggestion
	// Share a bytes pool between the intermediateNodeDB and valueNodeDB to
```

dhrubabasu (approved):

utils/bytes_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"math"
+	"math/rand"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIntSize(t *testing.T) {
```
abi87, line 15: not sure I get this test. We are checking some `math` package quantities, not anything related to our code.
Is it to make sure we have something breaking on platform we do not currently support? Can you clarify?

utils/bytes.go
```diff
@@ -12,3 +16,68 @@ func RandomBytes(n int) []byte {
...
+			},
+		}
+	}
+	return &p
+}
+
+// Get returns a non-nil pointer to a slice with the requested length.
+//
+// It is not guaranteed for the returned bytes to have been zeroed.
+func (p *BytesPool) Get(length int) *[]byte {
```
abi87, line 68: q: we don't need to return a pointer here, don't we? We deserialize what p[index].Get() returns so it's guaranteed to be non-nil. I'd return a slice instead of the pointer so that the caller can avoid warring about nil checks

utils/bytes_test.go
```diff
@@ -0,0 +1,87 @@
...
+		require.Len(*bytes, i)
+		p.Put(bytes)
+	}
+}
+
+func TestBytesPoolMaxInt(t *testing.T) {
+	require := require.New(t)
+
+	p := NewBytesPool()
+	bytes := p.Get(math.MaxInt32)
```
abi87, line 42: q: does this allocate all of this memory?
Not sure we ever need this much, nor that it's the right thing to do in a UT.
Should we consider a smaller max size? Or make it configurable?

utils/bytes_test.go
```diff
@@ -0,0 +1,87 @@
...
+		b.Run(strconv.Itoa(size), func(b *testing.B) {
+			p := NewBytesPool()
+			for i := 0; i < b.N; i++ {
+				p.Put(p.Get(size))
+			}
+		})
+	}
+}
+
+func BenchmarkBytesPool_Decending(b *testing.B) {
```
abi87, line 68: type Descending

utils/bytes_test.go
```diff
@@ -0,0 +1,87 @@
...
+	p := NewBytesPool()
+	for i := b.N; i > 0; i-- {
+		p.Put(p.Get(i))
+	}
+}
+
+func BenchmarkBytesPool_Ascending(b *testing.B) {
+	p := NewBytesPool()
+	for i := 0; i < b.N; i++ {
```
abi87, line 77: Not sure it's fine to write a benchmark like this (see [here](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go, section `Traps for young players).
We should probably specify the ascending/descending values independently of `b.N`.
It may not change the conclusion we draw (I really like the change), just trying to benchmark right.

database/prefixdb/db.go
```diff
@@ -236,23 +224,12 @@ func (db *Database) HealthCheck(ctx context.Context) (interface{}, error) {
 }
 
 // Return a copy of [key], prepended with this db's prefix.
-// The returned slice should be put back in the pool
-// when it's done being used.
-func (db *Database) prefix(key []byte) []byte {
-	// Get a []byte from the pool
-	prefixedKey := db.bufferPool.Get().([]byte)
+// The returned slice should be put back in the pool when it's done being used.
+func (db *Database) prefix(key []byte) *[]byte {
```
abi87, line 228: wonder if we could get rid of the pointer here. Not sure if there is a way to return a byte slice and give back the slice internal memory to the pool one we are done using it.

abi87 (changes requested):
nits and a few questions. 
Really nice bucketing trick here!

utils/bytes_test.go
StephenButtolph: I'll add a comment

utils/bytes.go
StephenButtolph: This avoids allocating the slice struct on the heap for every `Put` call.

utils/bytes_test.go
StephenButtolph: The `BytesPool` supports allocating slices in the range `[0, math.MaxInt]` (which is the same as `make`.

I added this test when I was hardcoding `intSize` to be `32`. I'm going to remove this test as it really doesn't make sense without that modification.

database/prefixdb/db.go
StephenButtolph: using `*[]byte` avoids memory allocations

darioush (approved):

database/prefixdb/db.go
```diff
@@ -118,15 +104,11 @@ func (db *Database) Get(key []byte) ([]byte, error) {
...
-	return val, err
+	defer db.bufferPool.Put(prefixedKey)
+
+	return db.db.Get(*prefixedKey)
 }
 
-// Assumes that it is OK for the argument to db.db.Put
-// to be modified after db.db.Put returns.
-// [key] can be modified after this method returns.
-// [value] should not be modified.
```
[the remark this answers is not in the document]
StephenButtolph: These invariants are documented on the DB interface now

PR #2922, opened 2024-04-07
Remove value_node_db batch

StephenButtolph:
## Why this should be merged

We currently allocate a `valueNodeBatch.ops` map for no reason. This removes this unused map.

Seems to provide as much as a `~22%` speedup (using memdb).

Before:
```
BenchmarkCommitView/apply_and_commit_changes-12         	     561	   2142773 ns/op	  583440 B/op	    4920 allocs/op
```

After:
```
BenchmarkCommitView/apply_and_commit_changes-12         	     685	   1673874 ns/op	  438132 B/op	    4904 allocs/op
```

## How this works

Removes `valueNodeBatch` and uses the `baseDB.Batch` instead.

## How this was tested

- [X] Added benchmark
- [X] CI

### DIFF
--- a/x/merkledb/db.go
+++ b/x/merkledb/db.go
@@ -44,6 +44,19 @@ var (
 	valueNodePrefix        = []byte{1}
 	intermediateNodePrefix = []byte{2}
 
+	// cleanShutdownKey is used to flag that the database did (or did not)
+	// previously shutdown correctly.
+	//
+	// If this key has value [hadCleanShutdown] it must be true that all
+	// intermediate nodes of the trie are correctly populated on disk and that
+	// the [rootDBKey] has the correct key for the root node.
+	//
+	// If this key has value [didNotHaveCleanShutdown] the intermediate nodes of
+	// the trie may not be correct and the [rootDBKey] may not exist or point to
+	// a node that node longer exists.
+	//
+	// Regardless of the value of [cleanShutdownKey], the value nodes must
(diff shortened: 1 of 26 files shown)


darioush (approved):

JacobEverly (approved):

dboehm-avalabs (approved):

x/merkledb/db.go
```diff
@@ -1023,17 +1023,21 @@ func (db *merkleDB) applyChanges(ctx context.Context, valueNodeBatch *valueNodeB
 		}
 
 		if shouldAddValue {
-			valueNodeBatch.Put(key, nodeChange.after)
+			if err := db.valueNodeDB.Write(valueNodeBatch, key, nodeChange.after); err != nil {
+				return err
+			}
 		} else if shouldDeleteValue {
-			valueNodeBatch.Delete(key)
+			if err := db.valueNodeDB.Write(valueNodeBatch, key, nil); err != nil {
```
dboehm-avalabs, line 1030: i believe nodechange.after should be nil here so i'm not sure this needs to be split anymore.

dboehm-avalabs: Nvm, this is deletion from the value db so a node could still exist, but with no value.

StephenButtolph: `nodeChange.before` may have a value but `nodeChange.after` might not have a value

PR #2930, opened 2024-04-09
Add hashing interface to merkledb

StephenButtolph:
## Why this should be merged

Currently the `merkledb` prescribes usage of `sha256`. This PR unblocks configuring the hash function to use. It's possible that this will be useful in ZK environments (by allowing the usage of `poseidon`, for example).

This doesn't really enable configuration of the hash function yet, as it currently isn't possible to implement the `merkledb.Hasher` interface outside of the merkledb because `node` is un-exported. That will be handled in a followup.

Running the benchmarks, this doesn't seem to introduce any performance regression. 

## How this works

- Introduces a `merkledb.Hasher` interface and uses it everywhere in the `merkledb` code.

## How this was tested

- [X] CI


x/merkledb/node_test.go
```diff
@@ -19,14 +19,13 @@ func Test_Node_Marshal(t *testing.T) {
 	fullKey := ToKey([]byte("key"))
 	childNode := newNode(fullKey)
 	root.addChild(childNode, 4)
-	childNode.setValue(maybe.Some([]byte("value")))
+	childNode.setValue(SHA256Hasher, maybe.Some([]byte("value")))
 	require.NotNil(t, childNode)
 
-	childNode.calculateID(&mockMetrics{})
```
StephenButtolph, line 25: `calculateID` probably used to populate the ID in these calls... but it doesn't do that anymore... Are these tests still doing what we want them to do?

danlaine: > Are these tests still doing what we want them to do?

I think so. We're marshaling a node and making sure that when we unmarshal it, we get the same thing back.

x/merkledb/db.go
```diff
@@ -163,6 +163,11 @@ type Config struct {
 	// BranchFactor determines the number of children each node can have.
 	BranchFactor BranchFactor
 
+	// Hasher defines the hash function to use when hashing the trie.
+	//
+	// If not specified, [SHA256Hasher] will be used.
```
danlaine, line 168: NIt: we could add a `var DefaultHasher = SHA256Hasher`. I think it might be a bit cleaner than referencing `SHA256Hasher` everywhere but I don't feel strongly.

danlaine (approved):

dhrubabasu (approved):

PR #2935, opened 2024-04-10
[tmpnet] Add network reuse to e2e fixture

marun:
## Why this should be merged

Previously, using an existing network with the e2e suite required starting the network with the tmpnetctl cli. This has been replaced with direct support for network reuse within the e2e fixture to better support testing of networks with subnets and remove the need for other repos (subnet-evm and hypersdk) to use tmpnetctl.

## How this works

 - Replace the --use-existing-network flag with --reuse-network flag which will ensure that a suite-compatible network is started if not already running and reused if already running
   - network reuse is segmented by test suite by virtue of embedding the name of the network owner (e.g. avalanchego-e2e, subnet-evm-e2e) in the symlink enabling network reuse

 - Added the --stop-network flag to support stopping a network previously started by --reuse-network and exiting immediately without running tests.

tmpnetctl is left in place for now but its future is uncertain.

## How this was tested

 - Updated the tests.e2e.existing.sh script to use the new flag instead of tmpnetctl


darioush (approved):

tests/fixture/e2e/env.go
```diff
@@ -58,52 +60,83 @@ func (te *TestEnvironment) Marshal() []byte {
...
+		var networkSymlink string // If populated, prompts removal of the referenced symlink if --stop-network is specified
+		if len(networkDir) == 0 {
+			// Attempt to reuse the network at the default owner path
+			var err error
+			symlinkPath, err := tmpnet.GetReusableNetworkPathForOwner(desiredNetwork.Owner)
+			require.NoError(err)
+			_, err = os.Stat(symlinkPath)
+			if errors.Is(err, os.ErrNotExist) {
+				// New network is required
+			} else {
```
StephenButtolph, line 76:
```suggestion
			if !errors.Is(err, os.ErrNotExist) {
```

tests/fixture/e2e/env.go
```diff
@@ -58,52 +60,83 @@ func (te *TestEnvironment) Marshal() []byte {
...
-			if existing := network.GetSubnet(subnet.Name); existing != nil {
-				// Already present
-				continue
+	// Need to load the network if it is being stopped or reused
+	if flagVars.StopNetwork() || flagVars.ReuseNetwork() {
+		networkDir := flagVars.NetworkDir()
+		var networkSymlink string // If populated, prompts removal of the referenced symlink if --stop-network is specified
+		if len(networkDir) == 0 {
+			// Attempt to reuse the network at the default owner path
+			var err error
```
StephenButtolph, line 70:
```suggestion
```

StephenButtolph (approved):

PR #2940, opened 2024-04-12
Split ManuallyTrack into ManuallyTrack and ManuallyGossip

StephenButtolph:
## Why this should be merged

1. This fixes a small inaccuracy when users specify custom bootstrappers. Custom bootstrappers should not be gossiped unless they are validators.
2. This allows us to specify an initial validator set that the node is willing to connect to but not introduce persistent bandwidth to the network.

## How this works

Splits `ManuallyTrack` (which would treat provided IDs as validators) into `ManuallyTrack` (is willing to connect) and `ManuallyGossip` (is willing to connect and gossip).

## How this was tested

- [X] Added unit tests
- [X] CI

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -244,7 +244,7 @@ func NewNetwork(
 	// Track all default bootstrappers to ensure their current IPs are gossiped
 	// like validator IPs.
 	for _, bootstrapper := range genesis.GetBootstrappers(config.NetworkID) {
-		ipTracker.ManuallyTrack(bootstrapper.ID)
+		ipTracker.ManuallyGossip(bootstrapper.ID)
 	}
 
 	peerConfig := &peer.Config{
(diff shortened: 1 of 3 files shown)


network/network.go
```diff
@@ -244,7 +244,7 @@ func NewNetwork(
 	// Track all default bootstrappers to ensure their current IPs are gossiped
 	// like validator IPs.
 	for _, bootstrapper := range genesis.GetBootstrappers(config.NetworkID) {
-		ipTracker.ManuallyTrack(bootstrapper.ID)
+		ipTracker.ManuallyGossip(bootstrapper.ID)
```
aaronbuchwald, line 247: From the PR description, I would have expected this to use `ManuallyTrack` instead of `ManuallyGossip` (after dividing into two separate functions). Why do we need to gossip the bootstrappers here?

aaronbuchwald (approved):
LGTM - one question

network/network.go
StephenButtolph: These are not custom bootstrappers - these are the default bootstrappers. See: https://github.com/ava-labs/avalanchego/tree/master/network#peerlist

StephenButtolph: The reason we gossip these nodeIDs is to more gracefully handle IP changes of bootstrappers.

joshua-kim (approved):

PR #2946, opened 2024-04-16
Improve bootstrapping peer selection

StephenButtolph:
## Why this should be merged

Currently bootstrapping fetching randomly samples bootstrappers to fetch blocks.

For the X-chain and C-chain this is inefficient, because validators may provide significantly different levels of bandwidth. This makes syncing the X-chain and C-chain very slow.

For the P-chain, this is based on the bootstrappers provided by the CLI - which defaults to 5 nodes. This results in saturating these connections (causing rate limiting). This in turn makes syncing the P-chain extremely slow.

This PR changes fetching for chains to be performed based on the bandwidth that nodes are willing to provide (and opens this to all peers rather than just the bootstrappers). This significantly improves the performance of fetching the P-chain. On mainnet (on my laptop) the ETA to fetch the P-chain was larger than 5 hours. (I'd imagine this number to have risen had I cared to actually measure the full time to sync). After this PR I was able to fetch the full mainnet P-chain in 20 minutes. The bottleneck after this PR is the time spent handling the `Ancestors` messages (rather than the network latency of the `Ancestors` message requests).

This also fixes a CPU spike that would occur when fetching blocks during bootstrapping and the node disconnects from all peers. Now the node will register timeouts to re-attempt fetching rather than immediately attempting to send another request.

## How this works

- Send `GetAncestors` messages based on provided bandwidth of peers rather than randomly sampling bootstrappers.
- Uses the `*p2p.PeerTracker` used for state sync.
- Waits for timeouts to fire for `GetAncestor` messages sent to ourself.

## How this was tested

- [X] CI
- [X] Syncing fuji and mainnet P-chain


snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -427,42 +409,44 @@ func (b *Bootstrapper) fetch(ctx context.Context, blkID ids.ID) error {
 		return nil
 	}
 
-	validatorID, ok := b.fetchFrom.Peek()
+	nodeID, ok := b.PeerTracker.SelectPeer()
 	if !ok {
-		return fmt.Errorf("dropping request for %s as there are no validators", blkID)
+		// TODO: FIXME
+		nodeID = b.Ctx.NodeID
```
StephenButtolph, line 415: This is actually the same as the prior behavior... But this is pretty odd

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -513,13 +496,25 @@ func (b *Bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, request
...
 	for _, block := range blocks[1:] {
+		numBytes += len(block.Bytes())
 		blockSet[block.ID()] = block
 	}
+
+	var (
+		requestLatency = time.Since(requestTime).Seconds() + epsilon
+		bandwidth      = float64(numBytes) / requestLatency
+	)
+	b.PeerTracker.RegisterResponse(nodeID, bandwidth)
```
StephenButtolph, line 516: Should we try to calculate the "useful" bandwidth? Technically someone could provide the request block and then random ancestors and be considered good even though the ancestors were useless.

StephenButtolph: added a todo

snow/networking/sender/sender.go
```diff
@@ -665,9 +665,10 @@ func (s *sender) SendGetAncestors(ctx context.Context, nodeID ids.NodeID, reques
 		s.engineType,
 	)
 
-	// Sending a GetAncestors to myself always fails.
+	// Sending a GetAncestors to myself will fail. To avoid constantly sending
+	// myself requests when not connected to any peers, we rely on the timeout
+	// firing to deliver the GetAncestorsFailed message.
 	if nodeID == s.ctx.NodeID {
```
StephenButtolph, line 671: Could this introduce a performance regression when syncing the X-chain DAG?

StephenButtolph: The X-chain DAG is synced with the current validator set. So this is only potentially an issue if the node that is syncing is a validator... which shouldn't really be happening

JacobEverly (approved):

meaghanfitzgerald:
lgtm

snow/engine/common/tracker/peers_test.go
```diff
@@ -20,36 +20,23 @@ func TestPeers(t *testing.T) {
 
 	p := NewPeers()
 
-	require.Zero(p.TotalWeight())
```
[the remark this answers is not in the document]
StephenButtolph: The peer tracker already has its own tests. These changes here is just removing dead code that is no longer used anywhere.

Technically we could consider adding a test for how the peer tracker is populated in the chain manager -> updated in the chain handler -> used in the bootstrapping engine. But this flow felt to me like it should just be tested by the e2e tests (which would be horribly broken if any of these weren't handled correctly)

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -528,38 +528,27 @@ func (b *Bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, request
 }
 
 func (b *Bootstrapper) GetAncestorsFailed(ctx context.Context, nodeID ids.NodeID, requestID uint32) error {
```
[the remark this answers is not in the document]
StephenButtolph: These functions are documented on the interface. Ex: https://github.com/ava-labs/avalanchego/blob/master/snow/engine/common/engine.go#L239-L243

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -513,13 +499,27 @@ func (b *Bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, request
...
 			zap.Stringer("blkID", actualID),
 		)
+		b.PeerTracker.RegisterFailure(nodeID)
 		return b.fetch(ctx, wantedBlkID)
 	}
 
-	blockSet := make(map[ids.ID]snowman.Block, len(blocks))
+	var (
+		numBytes = len(requestedBlock.Bytes())
+		blockSet = make(map[ids.ID]snowman.Block, len(blocks))
```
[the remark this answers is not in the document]
StephenButtolph: `blockSet` was a horrible name. I changed it to `ancestors`. This map is supposed to include potential ancestors of `requestedBlock`. Whereas `numBytes` currently includes all the block bytes (both `requestedBlock` and its `ancestors`)

snow/engine/snowman/bootstrap/bootstrapper_test.go
```diff
@@ -351,19 +379,16 @@ func TestBootstrapperEmptyResponse(t *testing.T) {
 	require.NoError(bs.startSyncing(context.Background(), blocksToIDs(blks[1:2])))
 	require.Equal(requestedNodeID, peerID)
 
-	// add another 2 validators to the fetch set to test behavior on empty
-	// response
-	bs.fetchFrom.Add(ids.GenerateTestNodeID(), ids.GenerateTestNodeID())
+	// Add another 2 peers to the fetch set to test behavior on empty response.
```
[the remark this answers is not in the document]
StephenButtolph: added some more color to the comment (and removed one of the no-longer-needed new peer.

vms/platformvm/vm_test.go
```diff
@@ -1472,12 +1473,22 @@ func TestBootstrapPartiallyAccepted(t *testing.T) {
 	)
 	require.NoError(err)
 
+	peerTracker, err := p2p.NewPeerTracker(
```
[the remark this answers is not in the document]
StephenButtolph: There should have only been one created here... whoopsie

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -471,15 +458,12 @@ func (b *Bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, request
 			zap.Uint32("requestID", requestID),
 		)
 
-		b.markUnavailable(nodeID)
+		b.PeerTracker.RegisterFailure(nodeID)
```
joshua-kim, line 461: This marks a node as unresponsive even though it actually responded... feels like we should be using `RegisterResponse` with a bandwidth of `0` instead?

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -513,53 +499,56 @@ func (b *Bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, request
...
-		blockSet[block.ID()] = block
+		numBytes += len(block.Bytes())
+		ancestors[block.ID()] = block
 	}
-	if err := b.process(ctx, requestedBlock, blockSet); err != nil {
+
+	// TODO: Calculate bandwidth based on the blocks that were persisted to
+	// disk.
+	var (
+		requestLatency = time.Since(requestTime).Seconds() + epsilon
```
joshua-kim, line 518: If we avoid this cast to `Seconds` and track all of our bandwidth in bytes/nsec can't we avoid the usage of epsilon to guard against division by zero? I think this is also more consistent with our other timing metrics which are all in nanoseconds.

StephenButtolph: > can't we avoid the usage of epsilon to guard against division by zero?

I don't think so. That would just make it an integer div by 0 rather than a floating point div by 0 (which would panic rather than returning NaN or inf.

fwiw I don't think this should really matter... But I do know we've seen unit test flakes on windows when we expect `time.Now() - time.Now() != 0`.

As far as using nanoseconds vs seconds... I'm not sure that we have any metrics that are bytes per nanosecond... Do we? (also this isn't a metric - these values are actually used).

snow/engine/snowman/bootstrap/bootstrapper.go
StephenButtolph: I don't think we want to sample this person again - so we should mark them as unresponsive.

StephenButtolph: By returning no blocks here, the peer is saying that they don't have the requested block history, which can currently happen for state synced C-chain nodes.

joshua-kim (approved):

PR #2947, opened 2024-04-17
Cleanup avalanche bootstrapping fetching

StephenButtolph:
## Why this should be merged

I tested this by syncing the first 250k Fuji X-chain vertices. Prior to this change it took `6m 44s` to sync these vertices. After this change it took `4m 3s`.

This aligns with how chains are going to be synced.

## How this works

Selects peers by bandwidth rather than by stake. 

## How this was tested

- [X] Ran a fuji node
- [X] CI


snow/engine/avalanche/bootstrap/bootstrapper.go
```diff
@@ -125,84 +129,76 @@ func (b *bootstrapper) Clear(context.Context) error {
 // response to a GetAncestors message to [nodeID] with request ID [requestID].
 // Expects vtxs[0] to be the vertex requested in the corresponding GetAncestors.
 func (b *bootstrapper) Ancestors(ctx context.Context, nodeID ids.NodeID, requestID uint32, vtxs [][]byte) error {
```
StephenButtolph, line 131: This function was refactored to be more similar to the snowman code. The major difference here is that we no longer "optimistically" accept any vertices. If we didn't request it, we will drop it. (which should never happen anyways because the message should have been dropped at the network level now)

snow/engine/avalanche/bootstrap/bootstrapper_test.go
```diff
@@ -620,111 +636,3 @@ func TestBootstrapperIncompleteAncestors(t *testing.T) {
...
-		require.Equal(choices.Accepted, vtx1.Status())
-		return []ids.ID{vtxID1}
-	}
-
-	vm.LinearizeF = func(_ context.Context, stopVertexID ids.ID) error {
-		require.Equal(vtxID1, stopVertexID)
-		return nil
-	}
-
-	require.NoError(bs.Ancestors(context.Background(), peerID, reqID, [][]byte{vtxBytes1, vtxBytes0}))
```
StephenButtolph, line 726: This was previously testing that using the (wrong) `peerID` and `reqID` would still process `vtxBytes1`. But we no longer support that... So this test just didn't make sense anymore.

JacobEverly (approved):

joshua-kim (approved):

PR #2949, opened 2024-04-17
Consolidate beta

aaronbuchwald:
## Why this should be merged

This PR replaces the deprecated `BetaVirtuous` and `BetaRogue` parameters with a single `Beta` value.

## How this works

`BetaVirtuous` and `BetaRogue` params were already replaced to use the same value, which makes this change a no-op in production.

This change includes:

- Replace the existing flags with a single `snow-commit-threshold` parameter to set `beta`
- Update the corresponding documentation for these flags
- Remove the usage and references to `BetaVirtuous` / `BetaRogue` throughout the codebase
- Fix existing tests that previously relied on divergent `beta` values

## How this was tested

Existing unit tests and updating the current unit tests to correctly handle the change. Specifically, existing tests that assumed divergent betas where a virtuous instance was finalized after `BetaVirtuous` consecutive polls needed to be adjusted.


snow/consensus/snowman/consensus_test.go
```diff
@@ -1319,6 +1194,12 @@ func LastAcceptedTest(t *testing.T, factory Factory) {
...
 
+	require.NoError(sm.RecordPoll(context.Background(), bag.Of(block0.IDV)))
+
+	lastAcceptedID, lastAcceptedHeight = sm.LastAccepted()
+	require.Equal(GenesisID, lastAcceptedID)
+	require.Equal(GenesisHeight, lastAcceptedHeight)
+
 	require.NoError(sm.RecordPoll(context.Background(), bag.Of(block1.IDV)))
 
 	lastAcceptedID, lastAcceptedHeight = sm.LastAccepted()
```
aaronbuchwald, line 1205: block1 transitively votes for block0, and we receive 1 vote for block1 and 1 vote for block0. This means we accept block 0 and have a confidence counter of 1 for block 1 (needs one more poll to finalize).

snow/consensus/snowman/consensus_test.go
```diff
@@ -1333,6 +1214,12 @@ func LastAcceptedTest(t *testing.T, factory Factory) {
 
 	require.NoError(sm.RecordPoll(context.Background(), bag.Of(block2.IDV)))
 
+	lastAcceptedID, lastAcceptedHeight = sm.LastAccepted()
+	require.Equal(block1.IDV, lastAcceptedID)
+	require.Equal(block1.HeightV, lastAcceptedHeight)
+
+	require.NoError(sm.RecordPoll(context.Background(), bag.Of(block2.IDV)))
```
aaronbuchwald, line 1221: The poll for block2 finalizes block1, but is not enough to finalize block2. This adds a check to make sure the poll finalizes block1 correctly and then adds one more poll before checking that block2 finalizes correctly.

snow/consensus/snowman/consensus_test.go
```diff
@@ -937,110 +919,6 @@ func RecordPollTransitiveVotingTest(t *testing.T, factory Factory) {
 	require.Equal(choices.Rejected, block4.Status())
 }
 
-func RecordPollDivergedVotingTest(t *testing.T, factory Factory) {
```
aaronbuchwald, line 940: Mentioned in https://github.com/ava-labs/avalanchego/pull/2949/commits/3e9e5608d4d22faed649afbe8e0a59ba150969b4 - this test requires divergent betas so that in the binary decomposition of voting, it's possible to conditionally accept an intermediate node and then transitively apply a vote through a block that has already been pruned due to the accepted prefix.

snow/consensus/snowball/tree_test.go
```diff
@@ -464,56 +460,6 @@ func TestSnowballCloseTrinary(t *testing.T) {
 	require.False(tree.Finalized())
 }
 
-func TestSnowballAddRejected(t *testing.T) {
```
aaronbuchwald, line 467: Duplicate of `TestSnowballAddPreviouslyRejected` @StephenButtolph could you confirm this?

snow/consensus/snowball/tree_test.go
```diff
@@ -176,26 +172,21 @@ func TestSnowballAddPreviouslyRejected(t *testing.T) {
...
 	two := ids.ID{0b00000010}
-	four := ids.ID{0b00000100}
 
 	params := Parameters{
 		K:               1,
 		AlphaPreference: 1,
 		AlphaConfidence: 1,
-		BetaVirtuous:    1,
-		BetaRogue:       2,
+		Beta:            2,
```
aaronbuchwald, line 180: Re-write this test based off the name. The previous behavior relied on divergent beta values for virtuous/rogue instances.

snow/consensus/snowball/tree_test.go
```diff
@@ -541,7 +486,7 @@ func TestSnowballResetChild(t *testing.T) {
 	{
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 0
     SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 1
-        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
+        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
```
aaronbuchwald, line 489: Virtuous instance should finalize at confidence of 2 instead of 1 after choosing a single beta value of 2.

snow/consensus/snowball/tree_test.go
```diff
@@ -555,7 +500,7 @@ func TestSnowballResetChild(t *testing.T) {
 	{
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 0, Finalized = false, SL(Preference = 0))) Bit = 0
     SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 1
-        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
+        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
```
aaronbuchwald, line 503: Same comment as above.

snow/consensus/snowball/tree_test.go
```diff
@@ -568,7 +513,7 @@ func TestSnowballResetChild(t *testing.T) {
 	{
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 2, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 0
     SB(Preference = 0, PreferenceStrength[0] = 2, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 1
-        SB(PreferenceStrength = 2, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
+        SB(PreferenceStrength = 2, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
```
aaronbuchwald, line 516: Same comment as above.

snow/consensus/snowball/tree_test.go
```diff
@@ -605,7 +549,7 @@ func TestSnowballResetSibling(t *testing.T) {
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 0
     SB(Preference = 1, PreferenceStrength[0] = 0, PreferenceStrength[1] = 1, SF(Confidence = 1, Finalized = false, SL(Preference = 1))) Bit = 1
         SB(PreferenceStrength = 0, SF(Confidence = 0, Finalized = false)) Bits = [2, 256)
-        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
+        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
```
aaronbuchwald, line 552: Virtuous instance should finalize after confidence of 2 instead of 1 now

snow/consensus/snowball/tree_test.go
```diff
@@ -619,8 +563,8 @@ func TestSnowballResetSibling(t *testing.T) {
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 1, SF(Confidence = 1, Finalized = false, SL(Preference = 1))) Bit = 0
     SB(Preference = 1, PreferenceStrength[0] = 0, PreferenceStrength[1] = 1, SF(Confidence = 1, Finalized = false, SL(Preference = 1))) Bit = 1
         SB(PreferenceStrength = 0, SF(Confidence = 0, Finalized = false)) Bits = [2, 256)
-        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
-    SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [1, 256)`
+        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
+    SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [1, 256)`
```
aaronbuchwald, line 567: same comment as above

snow/consensus/snowball/tree_test.go
```diff
@@ -632,8 +576,8 @@ func TestSnowballResetSibling(t *testing.T) {
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 2, PreferenceStrength[1] = 1, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 0
     SB(Preference = 1, PreferenceStrength[0] = 0, PreferenceStrength[1] = 2, SF(Confidence = 1, Finalized = false, SL(Preference = 1))) Bit = 1
         SB(PreferenceStrength = 0, SF(Confidence = 0, Finalized = false)) Bits = [2, 256)
-        SB(PreferenceStrength = 2, SF(Confidence = 1, Finalized = true)) Bits = [2, 256)
-    SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [1, 256)`
+        SB(PreferenceStrength = 2, SF(Confidence = 1, Finalized = false)) Bits = [2, 256)
```
aaronbuchwald, line 579: same comment as above. In this case the snowball preference strength is 2, but the confidence counter towards commit is still 1.

snow/consensus/snowball/tree_test.go
```diff
@@ -632,8 +576,8 @@ func TestSnowballResetSibling(t *testing.T) {
...
+    SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [1, 256)`
```
aaronbuchwald, line 580: Same comment as above

snow/consensus/snowball/tree_test.go
```diff
@@ -742,9 +684,10 @@ func TestSnowballFineGrained(t *testing.T) {
 
 	{
 		expected := `SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 0
-    SB(Preference = 0, PreferenceStrength[0] = 1, PreferenceStrength[1] = 0, SF(Confidence = 1, Finalized = false, SL(Preference = 0))) Bit = 2
-        SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = true)) Bits = [3, 256)
-        SB(PreferenceStrength = 0, SF(Confidence = 0, Finalized = false)) Bits = [3, 256)
+    SB(PreferenceStrength = 1, SF(Confidence = 1, Finalized = false)) Bits = [1, 2)
```
aaronbuchwald, line 687: See graphs added in https://github.com/ava-labs/avalanchego/pull/2949/commits/762e2d832efdd7688894048b37de7d2ac6a81fdb for visualization of the required changes here

The name doesn't give clear indication of what it's trying to test, so I just made sure that it passes correctly after the changes.

snow/consensus/snowball/tree_test.go
```diff
@@ -875,9 +814,10 @@ func TestSnowballFilterBinaryChildren(t *testing.T) {
...
```
aaronbuchwald, line 817: See graphs added in https://github.com/ava-labs/avalanchego/pull/2949/commits/565c5b1a932e66070504b7f42722ea40f3de0789#diff-1544dc37c28e958d7c8c249b181cc9fadda8acae042a933112e3535284d1efc8 for visualization

joshua-kim (approved):

snow/consensus/snowball/nnary_snowflake.go
```diff
@@ -25,29 +24,20 @@ type nnarySnowflake struct {
...
-	// rogue tracks if this instance has multiple choices or only one
-	rogue bool
-
 	// finalized prevents the state from changing after the required number of
 	// consecutive polls has been reached
 	finalized bool
 }
 
-func (sf *nnarySnowflake) Add(choice ids.ID) {
-	sf.rogue = sf.rogue || choice != sf.preference
```
abi87, line 49: q: what was happening here and why we dropped this?

snow/consensus/snowball/parameters.go
```diff
@@ -103,16 +99,14 @@ func (p Parameters) Verify() error {
...
-	case p.BetaVirtuous <= 0:
-		return fmt.Errorf("%w: betaVirtuous = %d: fails the condition that: 0 < betaVirtuous", ErrParametersInvalid, p.BetaVirtuous)
-	case p.BetaRogue == 3 && p.BetaVirtuous == 28:
-		return fmt.Errorf("%w: betaVirtuous = %d, betaRogue = %d: fails the condition that: betaVirtuous <= betaRogue\n%s", ErrParametersInvalid, p.BetaVirtuous, p.BetaRogue, errMsg)
-	case p.BetaRogue < p.BetaVirtuous:
-		return fmt.Errorf("%w: betaVirtuous = %d, betaRogue = %d: fails the condition that: betaVirtuous <= betaRogue", ErrParametersInvalid, p.BetaVirtuous, p.BetaRogue)
+	case p.Beta <= 0:
+		return fmt.Errorf("%w: beta = %d: fails the condition that: 0 < beta", ErrParametersInvalid, p.Beta)
+	case p.AlphaConfidence == 3 && p.AlphaPreference == 28:
+		return fmt.Errorf("%w: alphaConfidence = %d, alphaPreference = %d: fails the condition that: alphaPreference <= alphaConfidence\n%s", ErrParametersInvalid, p.AlphaConfidence, p.AlphaPreference, errMsg)
```
abi87, line 105: q: we where testing `p.BetaRogue` and `p.BetaVirtuous` with. a similar condition. Why we replaced them with `p.AlphaConfidence` instead of dropping this?

abi87 (commented):
left a couple of questions

snow/consensus/snowball/nnary_snowflake.go
aaronbuchwald: Previously, `BetaVirtuous` was the number of consecutive polls required to commit when we had only seen one choice (ie a virtuous proposal without a conflict) and `BetaRogue` was the number of consecutive polls required to commit when we had seen more than one choice.

This code differentiated between a virtuous and rogue decision, so we could decide which threshold was required to finalize a decision. Now that they're consolidated (required to be the same value before this change, so this is a no-op) we can clean up this code.

snow/consensus/snowball/parameters.go
aaronbuchwald: I wanted to keep `errMsg` 😉

abi87 (approved):

snow/consensus/snowball/tree.go
StephenButtolph, line 1: `binaryNode.Add` has dead code after this change. We should never need to filter additions during a binary node addition now because the child is guaranteed not to be accepted. Right?

StephenButtolph, line 1: Running mutation tests. Can we test this case?
```diff
 func (t *Tree) Add(choice ids.ID) {
 	prefix := t.node.DecidedPrefix()
 	// Make sure that we haven't already decided against this new id
-	if ids.EqualSubset(0, prefix, t.Preference(), choice) {
+	if ids.EqualSubset(1, prefix, t.Preference(), choice) {
 		t.node = t.node.Add(choice)
 	}
 }
```

StephenButtolph: I think this same dead code exists in `Record` poll with `prunedVotes.Filter`.

StephenButtolph, line 1: I think this is dead code now, as `newChild` can only be different from `child` if `u.Finalized()` is true.
```diff
		newChild, _ := u.child.RecordPoll(votes, u.shouldReset)
		if u.Finalized() {
			// If I'm now decided, return my child
			return newChild, true
		}
-		u.child = newChild
+		_, _ = u.child, newChild
```

aaronbuchwald: Yup, I see what you mean.

A descendant node can only receive votes through its parents. This means that if B is a descendant of A, then for B to finalize in a given round, A must either already be finalized or finalize in the same round.

Previously, this was not true because A and B could hit BetaVirtuous consecutive successful polls where B is virtuous and A is rogue.

aaronbuchwald: Added `TestSnowballAddDecidedFirstBit` to cover this case

aaronbuchwald: Removed the filtering for `binaryNode` in `Add` and `RecordPoll`

snow/consensus/snowball/tree.go
```diff
@@ -557,14 +554,7 @@ func (b *binaryNode) RecordPoll(votes bag.Bag[ids.ID], reset bool) (node, bool)
...
-		// The votes are filtered to ensure that they are votes that should
-		// count for the child
-		decidedPrefix := child.DecidedPrefix()
-		filteredVotes := prunedVotes.Filter(func(id ids.ID) bool {
-			return ids.EqualSubset(b.bit+1, decidedPrefix, b.preferences[bit], id)
-		})
-
-		newChild, _ := child.RecordPoll(filteredVotes, b.shouldReset[bit])
+		newChild, _ := child.RecordPoll(prunedVotes, b.shouldReset[bit])
 		if b.snow.Finalized() {
```
StephenButtolph, line 558: It seems like the below `b.children[bit] = newChild` call is also useless code right?

StephenButtolph (approved):

PR #2952, opened 2024-04-19
Remove usage of .Status() from blocks

StephenButtolph:

### DIFF
--- a/snow/consensus/snowman/consensus_test.go
+++ b/snow/consensus/snowman/consensus_test.go
@@ -305,7 +305,6 @@ func StatusOrProcessingPreviouslyAcceptedTest(t *testing.T, factory Factory) {
 
 	require.Equal(choices.Accepted, Genesis.Status())
 	require.False(sm.Processing(Genesis.ID()))
-	require.True(sm.Decided(Genesis))
 	require.True(sm.IsPreferred(Genesis.ID()))
 
 	pref, ok := sm.PreferenceAtHeight(Genesis.Height())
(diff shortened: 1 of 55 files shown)


snow/consensus/snowman/consensus_test.go
```diff
@@ -1428,46 +1424,6 @@ func MetricsRejectedErrorTest(t *testing.T, factory Factory) {
...
-			RejectV: errTest,
-			StatusV: choices.Processing,
-		},
-		ParentV: rejectedBlock.IDV,
-		HeightV: rejectedBlock.HeightV + 1,
-	}
-
-	err := sm.Add(context.Background(), block)
-	require.ErrorIs(err, errTest)
-}
```
StephenButtolph, line 1469: This test is now a duplicate of `AddOnUnknownParentTest`.

snow/engine/snowman/transitive_test.go
```diff
@@ -1254,12 +1254,18 @@ func TestEngineBlockingChitResponse(t *testing.T) {
...
 
 	require.Len(te.blocked, 2)
 	sender.CantSendPullQuery = false
+	sender.SendPullQueryF = func(_ context.Context, inVdrs set.Set[ids.NodeID], requestID uint32, blkID ids.ID, requestedHeight uint64) {
+		*queryRequestID = requestID
+		vdrSet := set.Of(vdr)
+		require.Equal(vdrSet, inVdrs)
+		require.Equal(blockingBlk.ID(), blkID)
+		require.Equal(uint64(1), requestedHeight)
+	}
```
StephenButtolph, line 1267: This test was previously incorrect. Because `GetBlock` doesn't return `missingBlk` (even after issuance), on the old code `blockingBlk` would be dropped. Because the fallback vote was for `issuedBlk`, the preference stayed as `issuedBlk` and the sender didn't error. However, now that `blockingBlk` is issued correctly, we need to handle the repoll for our new preference.

PR #2956, opened 2024-04-19
Improve and test getProcessingAncestor

StephenButtolph:
## Why this should be merged

- This removes a call to `GetBlock` during the normal case of bubbling votes (when the voted item is provided).
- This adds a test for `getProcessingAncestor` behavior.
- This fixes a minor bug that could cause a node to issue a poll even though there were no processing blocks.

## How this works

Just moves code around slightly.

## How this was tested

- [X] Added unit test
- [X] Fixed broken unit test
- [X] CI


snow/engine/snowman/metrics.go
```diff
@@ -38,104 +38,106 @@ type metrics struct {
 	issued                                *prometheus.CounterVec
 }
 
-func (m *metrics) Initialize(namespace string, reg prometheus.Registerer) error {
+func newMetrics(namespace string, reg prometheus.Registerer) (*metrics, error) {
```
StephenButtolph, line 41: This change allows us to create the metrics in a non-jank way.

snow/engine/snowman/transitive.go
```diff
@@ -1036,6 +1040,13 @@ func (t *Transitive) deliver(
 		}
 	}
 
+	// It's possible that the blocks we just added to consensus were decided
+	// immediately by votes that were pending their issuance. If this is the
+	// case, we should not be requesting any chits.
```
StephenButtolph, line 1045: This is a (very minor) bug fix.

snow/engine/snowman/transitive.go
```diff
@@ -1103,7 +1114,59 @@ func (t *Transitive) addUnverifiedBlockToConsensus(
...
+				zap.Stringer("initialVoteID", initialVote),
+				zap.Stringer("bubbledVoteID", bubbledVote),
+			)
+			if bubbledVote != initialVote {
+				t.numProcessingAncestorFetchesSucceeded.Inc()
+			} else {
+				t.numProcessingAncestorFetchesUnneeded.Inc()
+			}
+			return bubbledVote, true
+		}
```
StephenButtolph, line 1143: Moving the `Processing` check before `getBlock` avoids calling `VM.GetBlock` when `getProcessingAncestor` is called with a processing block.

snow/engine/snowman/transitive_test.go
```diff
@@ -1045,47 +1065,115 @@ func TestEngineBlockingChitResponse(t *testing.T) {
...
+		case issuedBlk.ID():
+			return issuedBlk, nil
+		case missingBlk.ID():
+			return missingBlk, nil
+		case blockingBlk.ID():
+			return blockingBlk, nil
+		default:
+			return nil, errUnknownBlock
+		}
+	}
```
StephenButtolph, line 1164: This test was previously pretty broken because the engine assumes that blocks that have been `Verify`ed are `GetBlock`able. So once we `issue`ed `missingBlk`, the engine was in a weird state.

snow/engine/snowman/transitive.go
```diff
@@ -1103,7 +1114,59 @@ func (t *Transitive) addUnverifiedBlockToConsensus(
...
+		metrics: t.metrics,
 		tree:    t.nonVerifieds,
 	})
 }
+
+// getProcessingAncestor finds [initialVote]'s most recent ancestor that is
+// processing in consensus. If no ancestor could be found, false is returned.
+//
+// Note: If [initialVote] is processing, then [initialVote] will be returned.
+func (t *Transitive) getProcessingAncestor(ctx context.Context, initialVote ids.ID) (ids.ID, bool) {
```
StephenButtolph, line 1126: This was moved into this file because all of the fields that are used on are this struct. I was considering moving this function off the structs entirely and just passing in the fields... But there are quite a few things that can be accessed... Especially because of the `getBlock` call.

aaronbuchwald (approved):

joshua-kim (approved):

snow/engine/snowman/transitive_test.go
```diff
@@ -60,10 +64,21 @@ func BuildChain(root *snowman.TestBlock, length int) []*snowman.TestBlock {
 	return chain
 }
 
-func setup(t *testing.T) (ids.NodeID, validators.Manager, *common.SenderTest, *block.TestVM, *Transitive) {
-	require := require.New(t)
+func MakeGetBlockF(blks ...[]*snowman.TestBlock) func(context.Context, ids.ID) (snowman.Block, error) {
```
joshua-kim, line 67: Should this live in `snowtest`?

StephenButtolph: That would cause a circular import... But I think it would make sense to move `*snowman.TestBlock` into a test package and then have these utils live there.

Do you think we should do that change in this PR or a followup?

joshua-kim: Can be a followup... don't see a point blocking on this

PR #2957, opened 2024-04-21
chore: fix function names in comment

socialsister:
## Why this should be merged

 fix function names in comment

PR #2958, opened 2024-04-22
Fix mismatched method names in comments

whosehang:
## Why this should be merged

Fix mismatched method names in comments


StephenButtolph:
Going to close this as it seems to be a duplicate of #2957.

PR #2964, opened 2024-04-23
Fee Calculator Refactor

joshua-kim:
## Why this should be merged

Adds a wrapper over the existing calculator visitor to provide a more convenient API to the caller.

## How this works

Refactors code.

Currently, callers of visitors to calculate a reward fee will have to perform something like this:
```
c := NewCalculator(...)
tx  := &Tx{}
tx.Visit(c)

fee := c.Fee
```

This changes the caller's behavior to:
```
c := NewCalculator(...)
fee := c.GetFee(tx, time.Now())
```


vms/platformvm/txs/fee/calculator.go
```diff
@@ -11,105 +11,123 @@ import (
...
+		upgradeTimes: upgradeTimes,
+	}
+}
 
 type Calculator struct {
+	config       StaticConfig
+	upgradeTimes upgrade.Config
+}
+
+func (c Calculator) GetFee(tx txs.UnsignedTx, time time.Time) uint64 {
```
abi87, line 28: nit: I'd rename this `ComputeFee` to reduce forward differences.
The reason is that with dynamic fees I will need to introduce a GetFee to access calculator.fee multiple times, while a tx is being built and calculator.fee gets updated.

vms/platformvm/txs/fee/calculator_test.go
```diff
@@ -31,14 +31,11 @@ var (
...
 	type feeTests struct {
-		name          string
-		chainTime     time.Time
-		unsignedTx    func(t *testing.T) txs.UnsignedTx
-		expectedError error
-		checksF       func(*testing.T, *Calculator)
+		name       string
+		chainTime  time.Time
+		unsignedTx func(t *testing.T) txs.UnsignedTx
+		expected   uint64
```
abi87, line 38: this is fine for now and I am in favour of having an `expected` attribute instead of the pair `expectedError` and `checksF`.
Note however that in next PRs, will go back to the pair since we will want to do more chacks than just fees (complexity for instance)

PR #2970, opened 2024-04-24
Consolidate record poll

aaronbuchwald:
## Why this should be merged

This PR moves the logic that turns a consensus poll into an action from the consensus implementation in `tree.go` / `flat.go` into the nnary/binary/unary interface.

This change is made in preparation for supporting Flexible Snowflake as described in https://arxiv.org/pdf/2404.14250v1.pdf. Flexible Snowflake uses multiple alpha confidence thresholds instead of a single `AlphaPreference` and `AlphaConfidence` threshold as in the current implementation.

To support this, the nnary/binary/unary implementations need access to the actual count of votes in favor of a given value.

## How this works

This change replaces the `RecordSuccessfulPoll` and `RecordPollPreference` calls on each consensus implementation with a single `RecordPoll` function that checks what threshold was reached and calls the corresponding function internally. The previous functions are un-exported and called from the new `RecordPoll` function.

## How this was tested

Uses the existing unit tests modified so that `RecordSuccessfulPoll(...)` turns into `RecordPoll(alphaConfidence, ...)` and `RecordPollPreference(...)` turns into `RecordPoll(alphaPreference, ...)`. To make sure tests correctly differentiate between recording an alpha preference and alpha confidence majority, the alphaConfidence majority used in existing tests must be strictly greater than the alphaPreference value.

### DIFF
--- a/snow/consensus/snowball/binary_snowball.go
+++ b/snow/consensus/snowball/binary_snowball.go
@@ -40,26 +40,10 @@ func (sb *binarySnowball) Preference() int {
 }
 
 func (sb *binarySnowball) RecordPoll(count, choice int) {
-	switch {
-	case count >= sb.alphaConfidence:
-		sb.recordSuccessfulPoll(choice)
-	case count >= sb.alphaPreference:
-		sb.recordPollPreference(choice)
-	default:
-		// If the poll was unsuccessful, RecordUnsuccessfulPoll should
-		// have been called instead.
-		sb.RecordUnsuccessfulPoll()
+	if count >= sb.alphaPreference {
+		sb.increasePreferenceStrength(choice)
 	}
-}
-
-func (sb *binarySnowball) recordSuccessfulPoll(choice int) {
-	sb.increasePreferenceStrength(choice)
-	sb.binarySnowflake.recordSuccessfulPoll(choice)
-}
-
-func (sb *binarySnowball) recordPollPreference(choice int) {
-	sb.increasePreferenceStrength(choice)
-	sb.binarySnowflake.recordPollPreference(choice)
+	sb.binarySnowflake.RecordPoll(count, choice)
 }
 
 func (sb *binarySnowball) String() string {
--- a/snow/consensus/snowball/binary_snowball.go
+++ b/snow/consensus/snowball/binary_snowball.go
@@ -7,9 +7,9 @@ import "fmt"
 
 var _ Binary = (*binarySnowball)(nil)
 
-func newBinarySnowball(beta, choice int) binarySnowball {
+func newBinarySnowball(alphaPreference int, alphaConfidence int, beta int, choice int) binarySnowball {
 	return binarySnowball{
-		binarySnowflake: newBinarySnowflake(beta, choice),
+		binarySnowflake: newBinarySnowflake(alphaPreference, alphaConfidence, beta, choice),
 		preference:      choice,
 	}
 }
@@ -37,6 +37,19 @@ func (sb *binarySnowball) Preference() int {
 		return sb.binarySnowflake.Preference()
 	}
 	return sb.preference
+}
+
+func (sb *binarySnowball) RecordPoll(count, choice int) {
+	switch {
+	case count >= sb.alphaConfidence:
+		sb.RecordSuccessfulPoll(choice)
+	case count >= sb.alphaPreference:
+		sb.RecordPollPreference(choice)
+	default:
+		// If the poll was unsuccessful, RecordUnsuccessfulPoll should
+		// have been called instead.
+		sb.RecordUnsuccessfulPoll()
+	}
 }
 
 func (sb *binarySnowball) RecordSuccessfulPoll(choice int) {
--- a/snow/consensus/snowball/binary_snowball.go
+++ b/snow/consensus/snowball/binary_snowball.go
@@ -41,7 +41,10 @@ func (sb *binarySnowball) Preference() int {
 
 func (sb *binarySnowball) RecordPoll(count, choice int) {
 	if count >= sb.alphaPreference {
-		sb.increasePreferenceStrength(choice)
+		sb.preferenceStrength[choice]++
+		if sb.preferenceStrength[choice] > sb.preferenceStrength[1-choice] {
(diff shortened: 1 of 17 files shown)


snow/consensus/snowball/binary_snowflake.go
```diff
@@ -32,7 +40,24 @@ type binarySnowflake struct {
...
+	switch {
+	case count >= sf.alphaConfidence:
+		sf.recordSuccessfulPoll(choice)
+	case count >= sf.alphaPreference:
+		sf.recordPollPreference(choice)
+	default:
+		// If the poll was unsuccessful, RecordUnsuccessfulPoll should
+		// have been called instead.
+		sf.RecordUnsuccessfulPoll()
+	}
```
StephenButtolph, line 57: I feel like we should just delete `recordSuccessfulPoll` and `recordPollPreference`.
```suggestion
	if count < sf.alphaPreference {
		// If the poll was unsuccessful, RecordUnsuccessfulPoll should
		// have been called instead.
		sf.RecordUnsuccessfulPoll()
		return
	}
	if count < sf.alphaConfidence {
		sf.confidence = 0
		sf.binarySlush.RecordSuccessfulPoll(choice)
		return
	}

	if preference := sf.Preference(); preference == choice {
		sf.confidence++
	} else {
		// confidence is set to 1 because there has already been 1 successful
		// poll, namely this poll.
		sf.confidence = 1
	}
	sf.finalized = sf.confidence >= sf.beta
	sf.binarySlush.RecordSuccessfulPoll(choice)
```
This comment applies to all the other `snowflake` `RecordPoll` functions as well

aaronbuchwald: Will update - I was originally going for making this the minimal possible change

snow/consensus/snowball/binary_snowball.go
```diff
@@ -39,14 +39,11 @@ func (sb *binarySnowball) Preference() int {
...
-	sb.increasePreferenceStrength(choice)
-	sb.binarySnowflake.RecordSuccessfulPoll(choice)
-}
-
-func (sb *binarySnowball) RecordPollPreference(choice int) {
-	sb.increasePreferenceStrength(choice)
-	sb.binarySnowflake.RecordPollPreference(choice)
+func (sb *binarySnowball) RecordPoll(count, choice int) {
+	if count >= sb.alphaPreference {
+		sb.increasePreferenceStrength(choice)
```
StephenButtolph, line 44: Should we inline this function? It used to be called twice, but now it's only called here.

snow/consensus/snowball/flat.go
```diff
@@ -28,17 +28,7 @@ type Flat struct {
...
-		f.RecordSuccessfulPoll(pollMode)
-		return true
-	case numVotes >= f.params.AlphaPreference:
-		f.RecordPollPreference(pollMode)
-		return true
-	default:
-		f.RecordUnsuccessfulPoll()
-		return false
-	}
+	f.Nnary.RecordPoll(numVotes, pollMode)
```
StephenButtolph, line 31: I don't think this guarantees the invariant on `RecordPoll` that `numVotes >= AlphaPreference`. (I don't think this is a bug fwiw, I think that the invariants on `RecordPoll` about expecting at least `AlphaPreference` is not needed.

snow/consensus/snowball/nnary_snowball.go
```diff
@@ -47,14 +47,11 @@ func (sb *nnarySnowball) Preference() ids.ID {
...
-	sb.increasePreferenceStrength(choice)
-	sb.nnarySnowflake.RecordSuccessfulPoll(choice)
-}
-
-func (sb *nnarySnowball) RecordPollPreference(choice ids.ID) {
-	sb.increasePreferenceStrength(choice)
-	sb.nnarySnowflake.RecordPollPreference(choice)
+func (sb *nnarySnowball) RecordPoll(count int, choice ids.ID) {
+	if count >= sb.alphaPreference {
+		sb.increasePreferenceStrength(choice)
```
StephenButtolph, line 52: Same comment as the other instance - should we inline this?

snow/consensus/snowball/flat.go
aaronbuchwald: Ya `RecordPoll` implementations no longer actually assume this, but will update since it's a documented invariant anyways.

Could go the other way and remove `RecordUnsuccessfulPoll`, but we currently use `RecordUnsuccessfulPoll` to handle a consensus "falter" as well, which we could hardcode to `RecordPoll(0, choice)`, but it seems cleaner to have a separate function as it currently is.

StephenButtolph (approved):

snow/consensus/snowball/consensus.go
```diff
@@ -77,23 +70,17 @@ type Nnary interface {
...
 
-	// RecordSuccessfulPoll records a successful poll towards finalizing the
-	// specified choice
-	RecordSuccessfulPoll(choice int)
-
-	// RecordPollPreference records a poll that preferred the specified choice
-	// but did not contribute towards finalizing the specified choice
-	RecordPollPreference(choice int)
+	// RecordPoll records the results of a network poll
+	RecordPoll(count, choice int)
```
joshua-kim, line 83: This is pretty nit-picky but I think switching the order of the params here looks more natural (as w/ the rest of the RecordPoll signatures we change in this PR)

joshua-kim (approved):

snow/consensus/snowball/consensus.go
aaronbuchwald: Ya you're right, this also matches the return value of `bag.Mode()` https://github.com/ava-labs/avalanchego/blob/41c12018063d4ef730e78af3fa04ad6a1997db70/snow/consensus/snowball/flat.go#L30.

Will update

aaronbuchwald: Hmm further thought looking through this, I'd say I actually prefer this as is because the `Unary` interface only takes in the `count` parameter, so it seems slightly more natural to keep that as the first parameter for `Binary` and `Nnary` as well. I think this was my original rationale for choosing this order iirc, but will leave this as is.

PR #2736, opened 2024-02-14
X-chain - consolidate tx creation in unit tests

abi87:
## Why this should be merged
Reuse wallet tx builders in unit tests instead of laying out txs as we currently do.
This simplifies maintenance and introduction of dynamic fees mechanism

## How this works

1. Introduce test txBuilder based on wallet, similarly to what we have done with `platformvm`
2. Drop dependence of UT from API code soon to be deprecated and use txBuilders
3. (Following code review suggestions) consolidated UTs cleanup

## How this was tested
CI

PR #2972, opened 2024-04-26
X-chain - Repack tx builder

abi87:
## Why this should be merged

Dynamic fees will radically chance the way we build transactions. Fees won't be known ahead of time, but will be iterativelly calculated as inputs and outputs are added to the tx.
On top of this we have at least three ways to build transactions currently:

- Using wallet's builder
- With ad-hoc functions in X-chain APIs (deprecated and soon to be removed in favour of wallet)
- Manually laying out txs in unit tests

This code duplication is a maintenance nightmare as we introduce dynamic fees.
This PR simplify the maintenance burned in two ways:

- Pulls out of X-chain APIs, txs builders
- Makes txs builders use wallet builders to create transaction

In a coming PR, unit tests are update to make use of the txs builders instead of manually laying out txs, so to have a unified and maintainable way to build X-chain txs

## How this works

- Pull out of X-chain APIs, txs builders
- Make txs builders use wallet builders to create transaction
- Drop most of utxo.Spend methods (since wallet is used instead)

## How this was tested
CI


vms/avm/service.go
```diff
@@ -60,7 +60,10 @@ type FormattedAssetID struct {
 }
 
 // Service defines the base service for the asset vm
-type Service struct{ vm *VM }
```
abi87, line 63: most of the code in this files has been:

- Moved to tx_builders.go file
- replaced by wallet machinery

vms/avm/wallet_service.go
```diff
@@ -98,35 +95,6 @@ func (w *WalletService) issue(tx *txs.Tx) (ids.ID, error) {
 	return txID, nil
 }
 
-func (w *WalletService) update(utxos []*avax.UTXO) ([]*avax.UTXO, error) {
```
abi87, line 101: moved to wallet_service_backend.go

vms/avm/service_test.go
```diff
@@ -1555,27 +1555,27 @@ func TestServiceGetTxJSON_OperationTxWithPropertyFxMintOpMultiple(t *testing.T)
 func newAvaxBaseTxWithOutputs(t *testing.T, genesisBytes []byte, chainID ids.ID, fee uint64, parser txs.Parser) *txs.Tx {
 	avaxTx := getCreateTxFromGenesisTest(t, genesisBytes, "AVAX")
 	key := keys[0]
-	tx := buildBaseTx(avaxTx, chainID, fee, key)
+	tx := buildTestBaseTx(avaxTx, chainID, fee, key)
```
abi87, line 1558: these renamings here avoid conflicts with functions defined in tx_builders.go. We will further simplify this test code in next PR, https://github.com/ava-labs/avalanchego/pull/2736

vms/avm/tx_builders.go
```diff
@@ -0,0 +1,229 @@
...
+	return signer.SignUnsigned(context.Background(), pSigner, utx)
+}
+
+func mintFTs(
+	backend txBuilderBackend,
+	outputs map[ids.ID]*secp256k1fx.TransferOutput,
+	kc *secp256k1fx.Keychain,
+	changeAddr ids.ShortID,
+) (*txs.Tx, error) {
+	pBuilder, pSigner := builders(backend, kc)
```
joshua-kim, line 113: Variable naming seems to imply this is somehow P-Chain related so I suspect this a copy-paste error

vms/avm/tx_builders.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func buildCreateAssetTx(
+	backend txBuilderBackend,
+	name, symbol string,
+	denomination byte,
+	initialStates map[uint32][]verify.State,
+	kc *secp256k1fx.Keychain,
+	changeAddr ids.ShortID,
+) (*txs.Tx, ids.ShortID, error) {
```
joshua-kim, line 37: I feel like returning `ids.ShortID` is strange because the caller already knows the `changeAddr`

vms/avm/wallet_service_backend.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/ava-labs/avalanchego/utils/linked"
+	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/vms/avm/txs"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/wallet/chain/x/builder"
+)
+
+var _ txBuilderBackend = (*walletServiceBackend)(nil)
+
+func NewWalletServiceBackend(vm *VM) *walletServiceBackend {
```
joshua-kim, line 22: Can we accept an `UTXOReader` instead of the concrete `*VM` type here?

vms/avm/wallet_service.go
```diff
@@ -195,8 +163,7 @@ func (w *WalletService) SendMultiple(_ *http.Request, args *SendMultipleArgs, re
 		return err
 	}
 
-	utxos, err = w.update(utxos)
-	if err != nil {
+	if err = w.vm.walletService.update(utxos); err != nil {
```
joshua-kim, line 166: I think this call to `update` looks strange in the new code. I think we can remove this call by having `LoadUser` called in the wallet

vms/avm/wallet_service_backend.go
```diff
@@ -0,0 +1,108 @@
...
+	addrs set.Set[ids.ShortID]
+}
+
+func (b *walletServiceBackend) Context() *builder.Context {
+	return b.ctx
+}
+
+func (b *walletServiceBackend) ResetAddresses(addrs set.Set[ids.ShortID]) {
+	b.addrs = addrs
+}
```
joshua-kim, line 54: I think we should be able to remove this method + `addrs` field in this struct if we make the users a parameter in the function args instead

vms/avm/tx_builders.go
```diff
@@ -0,0 +1,229 @@
...
+	walletbuilder "github.com/ava-labs/avalanchego/wallet/chain/x/builder"
+)
+
+type txBuilderBackend interface {
+	walletbuilder.Backend
+	signer.Backend
+
+	Context() *walletbuilder.Context
+	ResetAddresses(addrs set.Set[ids.ShortID])
+}
```
joshua-kim, line 28: What do we need this interface for? Wondering if we need this

vms/avm/wallet_service_backend.go
```diff
@@ -0,0 +1,108 @@
...
+	vm         *VM
+	pendingTxs *linked.Hashmap[ids.ID, *txs.Tx]
+	utxos      []*avax.UTXO
+
+	addrs set.Set[ids.ShortID]
+}
+
+func (b *walletServiceBackend) Context() *builder.Context {
+	return b.ctx
+}
```
joshua-kim, line 50: What do we need this for?

vms/avm/wallet_service_backend.go
```diff
@@ -0,0 +1,108 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package avm
```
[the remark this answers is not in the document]
abi87: I'd go for the P-chain structure.
In `platformvm` we used to have a `txs/builder` package, which now has been renamed to `txs/txstest` since those builders are only used in tests.
Here I'd keep a `txs/builder` package as long as these builder are used in the APIs

vms/avm/wallet_service_backend.go
abi87: Dropped

vms/avm/tx_builders.go
abi87: We don't need context, but we need the whole interface to allow wallet builders and signers to retrieve UTXOs and sign transactions.
I could not find a way to avoid the ResetAddresses, as we managed in `platformvm` because we have two different backends here

vms/avm/tx_builders.go
abi87: Agree, dropped

vms/avm/wallet_service_backend.go
abi87: Dropped ResetAddresses by introducing an adapter among the AVMBuilderBackend and the wallet Builder

vms/avm/wallet_service_backend.go
abi87: the vm here is used by wallet service as well. I'd keep vm to centralize access

vms/avm/txs/builder/backends.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
abi87, line 2: the file organization of this folder is very similar to what we do in P-chain.
The difference is that in platformvm we don't have multiple backends (wallet and service backends), so we can fully define the backend in the tx builder package.

abi87:
Closing this dow since we reached agreement on not upgrading deprecated X-chain APIs to Dynamic fees.
APIs will stop working when Dynamic fees will be activated
https://github.com/ava-labs/avalanchego/pull/2736 will upgrade unit tests to use a wallet based builder, in order to simplify maintenance

PR #2976, opened 2024-04-30
[antithesis] Refactor existing job to support xsvm test setup

marun:
This PR refactors the existing antithesis test setup to simplify addition of test setups for subnets (both xsvm and subnet-evm):
 - Moved to a new path (`tests/antithesis -> tests/antithesis/avalanchego`) to allow new setups to be added
 - Replaced static docker compose configuration with generated configuration to avoid duplicating common node configuration across setups and simplify inclusion of subnet/chain IDs that will need to be determined dynamically
 - ~Collapse config+node+workload to a single image to minimize Dockerfile maintenance and image build time~
 - Added a test and CI job of image build to simplify maintenance 
 - Added doc content to aid in maintenance and the addition of new test setups

## TODO

 - [x] Merge parent #2977 
 - [x] Bring back config image


tests/antithesis/compose.go
```diff
@@ -0,0 +1,190 @@
...
+		Services: services,
+	}, nil
+}
+
+// Convert a mapping of avalanche config keys to a mapping of env vars
+func keyMapToEnvVarMap(keyMap types.Mapping) types.Mapping {
+	envVarMap := make(types.Mapping, len(keyMap))
+	for key, val := range keyMap {
+		// e.g. network-id -> AVAGO_NETWORK_ID
+		envVar := "AVAGO_" + strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
```
StephenButtolph, line 186: We should make `"AVAGO"` a constant exposed from the `config` package and use it here. Currently it's just used directly as `v.SetEnvPrefix("avago")`

StephenButtolph: We could also expose the `*strings.Replacer`

tests/fixture/tmpnet/local_network.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/config"
+	"github.com/ava-labs/avalanchego/genesis"
+	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
+)
+
+func LocalNetworkOrDie() *Network {
```
StephenButtolph, line 14:
```suggestion
func LocalNetworkOrPanic() *Network {
```

StephenButtolph (commented):
small nits then lgtm

StephenButtolph (approved):

JacobEverly (approved):

PR #2977, opened 2024-05-01
Simplify go version maintenance

marun:
Now that the minimum go version is defined in go.mod, source the version from there for everywhere that needs it rather than hard-coding. This is intended to simplify version changes.

Also added repo-local actions (in .github/actions) to simplify configuring setup-go with the project go version.

As part of this change, support for running protobuf with docker has been removed. It wasn't commonly used and its removal avoids having to maintain the golang version in the Dockerfile.


StephenButtolph (approved):

dhrubabasu (commented):
one small comment around go version being listed in markdown files

dhrubabasu (approved):

arturrez:
there should be default versions specified as currently `docker build .` is broken

PR #2981, opened 2024-05-02
Mark nodes as connected to the P-chain networking stack

StephenButtolph:
## Why this should be merged

Currently the P-chain does not correctly mark nodes as connected. This impacts peer connectivity checks during gossip.

## How this works

Marks nodes as connected.

## How this was tested

- [X] CI
- [X] Running a local network node shows that pull queries are sent.


dhrubabasu (approved):
good catch ❤️

joshua-kim (approved):

abi87 (approved):

JacobEverly (approved):

PR #2982, opened 2024-05-02
[antithesis] Add test setup for xsvm

marun:
Add an antithesis test setup targeting xsvm to provide an example of subnet use with antithesis and increase coverage for consensus.

## TODO

 - [x] Merge parent #2976
 - [x] Figure out why the docker compose network fails to restart

### DIFF
--- a/tests/antithesis/compose.go
+++ b/tests/antithesis/compose.go
@@ -20,6 +20,8 @@ import (
 	"github.com/ava-labs/avalanchego/utils/perms"
 )
 
+const bootstrapIndex = 0
+
 // Initialize the given path with the docker-compose configuration (compose file and
 // volumes) needed for an Antithesis test setup.
 func GenerateComposeConfig(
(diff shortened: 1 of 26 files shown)


vms/example/xsvm/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# The version is supplied as a build argument rather than hard-coded
```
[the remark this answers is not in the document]
StephenButtolph: Makes sense to me.

tests/antithesis/compose.go
```diff
@@ -188,3 +228,14 @@ func keyMapToEnvVarMap(keyMap types.Mapping) types.Mapping {
...
+
+// Retrieve the service name for a node at the given index. Common to
+// GenerateComposeConfig and InitDBVolumes to ensure consistency
+// between db volumes configuration and volume paths.
+func getServiceName(index int) string {
+	baseName := "avalanche"
+	if index == 0 {
+		return baseName + "-bootstrap-node"
+	}
+	return fmt.Sprintf("%s-node-%d", baseName, index+1)
```
StephenButtolph, line 240: Why do we do `index+1`? Seems odd that we would have `avalanche-node-2` but not `avalanche-node-1`

tests/antithesis/node_health.go
```diff
@@ -0,0 +1,50 @@
...
+			log.Printf("node reported healthy at %s", uri)
+			return nil
+		default:
+			log.Printf("node reported unhealthy at %s", uri)
+		}
+
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			return fmt.Errorf("node health check cancelled at %s", uri)
```
StephenButtolph, line 47: We should probably wrap `ctx.Err()` here

StephenButtolph (approved):
lgtm

abi87 (approved):

tests/fixture/subnet/xsvm.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	if key == nil {
+		var err error
+		key, err = secp256k1.NewPrivateKey()
+		if err != nil {
+			panic(err)
+		}
+	}
+
```
joshua-kim, line 28: I feel like it looks weird to pass in `nil` in the caller function and override it with a generated key. Why don't we just force the caller to generate the key instead?

joshua-kim (approved):

PR #2986, opened 2024-05-03
network/p2p/gossip: refactor Set.Add to accept multiple elements

lebdron:
Allows potential optimized implementations for adding multiple elements to sets of Gossipables.

## Why this should be merged

Coreth currently has to pass every received transaction to the mempool in a separate Add call, requiring exclusive mutex acquire and release on every call.

https://github.com/ava-labs/coreth/blob/8d22112c7033658eb83b64e8b31d2f9c69676e28/plugin/evm/gossip.go#L192-L194

https://github.com/ava-labs/coreth/blob/8d22112c7033658eb83b64e8b31d2f9c69676e28/core/txpool/legacypool/legacypool.go#L1108-L1110

This interface will allow an optimized implementation to avoid individual Add calls for Coreth.


network/p2p/gossip/gossipable.go
```diff
@@ -29,3 +29,20 @@ type Set[T Gossipable] interface {
 	// corresponding salt.
 	GetFilter() (bloom []byte, salt []byte)
 }
+
+type AddManySet[T Gossipable] interface {
```
StephenButtolph, line 33: I know this would increase the scope of this change (especially because changing interfaces that are depended on by Coreth creates annoy circular dependencies)... But should we just be updating the `Set` interface to require this rather than adding this optional interface?

It seems like the X-chain and the P-chain could also try and avoid grabbing their locks multiple times.

lebdron: Sure, that was exactly my concern, trying to add this feature in a way without modifying the existing interface. But if it's OK, then I can just refactor `Add` to take a slice instead of a single element.

StephenButtolph: I think we should probably push for just having one `Add`. Let me know if you have any questions on how to handle this weird circular import.

joshua-kim: +1. Having an interface that has both `Add` and the `AddMany` signature is going to be an awkward API IMO. I would recommend an API that uses a variadic parameter like:

```
Add(gossipables ... Gossipable) []error
```

vms/platformvm/network/gossip.go
```diff
@@ -90,49 +90,54 @@ type gossipMempool struct {
...
+			continue
+		}
+
+		if errs[i] = g.Mempool.Add(tx); errs[i] != nil {
+			g.Mempool.MarkDropped(txID, errs[i])
+			continue
+		}
+
+		g.lock.Lock()
+		defer g.lock.Unlock()
```
StephenButtolph, line 121: I don't think this is going to work as expected if multiple transactions are provided.

lebdron: Whoops, overlooked that, thanks! Regarding X/P-Chain mempools, I'd rather put all the checks and tx verification inside the mempool not to acquire RLock for every lookup. Plus, `GetDropReason` acquires RW lock on mutex inside LRU cache, which would also be possible to rearrange if we do the verification inside Add. It's possible to pass txVerifier as a dependency to mempool I think.

vms/platformvm/network/gossip.go
```diff
@@ -92,49 +92,63 @@ type gossipMempool struct {
...
+		}
+
+		g.lock.Lock()
+
+		g.bloom.Add(tx)
+		var reset bool
+		reset, errs[i] = gossip.ResetBloomFilterIfNeeded(g.bloom, g.Mempool.Len()*bloomChurnMultiplier)
+		if errs[i] != nil {
+			hasErr = true
+			continue
```
StephenButtolph, line 134: We should add tests for multiple txs - I think this can continue to the next iteration with the lock held.

Would it make sense to keep the single tx addition as a method so that we can keep using `defer g.lock.Unlock()`?

joshua-kim:
Do you happen to have any benchmarks/profile that indicate that this is giving us any measurable change in performance? I think unless there is some non-trivial benchmark that shows that this is an improvement we should close this because returning `[]error` is going to complicate a lot of the upstream code (ref: diff upstream of `p2p`) + I think it's a somewhat non-idiomatic way of handling errors in Go.

joshua-kim:
Closing for the above reason, feel free to re-open.

PR #3000, opened 2024-05-06
prefixdb: fix bug with Compact nil limit

a1k0n:
## Why this should be merged

Compact(nil, nil) is a no-op in prefixdb, but should actually work according to the docs:
	// A nil start is treated as a key before all keys in the DB.
	// And a nil limit is treated as a key after all keys in the DB.
	// Therefore if both are nil then it will compact entire DB.

## How this was tested
It wasn't. We need a way to test Compact in db_test, but maybe I can just test the suffix() function.

### DIFF
--- a/database/prefixdb/db.go
+++ b/database/prefixdb/db.go
@@ -189,7 +189,12 @@ func (db *Database) Compact(start, limit []byte) error {
 	prefixedStart := db.prefix(start)
 	defer db.bufferPool.Put(prefixedStart)
 
-	prefixedLimit := db.prefix(limit)
+	var prefixedLimit *[]byte
+	if limit == nil {
+		prefixedLimit = db.suffix()
+	} else {
+		prefixedLimit = db.prefix(limit)
+	}
 	defer db.bufferPool.Put(prefixedLimit)
 
 	return db.db.Compact(*prefixedStart, *prefixedLimit)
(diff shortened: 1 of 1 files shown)


database/prefixdb/db.go
```diff
@@ -233,6 +238,21 @@ func (db *Database) prefix(key []byte) *[]byte {
...
+	n := len(db.dbPrefix)
+	// lexically increment prefix to find the suffix of this db range
+	for i := n - 1; i >= 0; i-- {
+		(*prefixedKey)[i]++
+		if (*prefixedKey)[i] != 0 {
+			break
+		}
+	}
+	return prefixedKey
+}
```
StephenButtolph, line 254: Would it make sense to define `dbPrefix` as a constant? Similar to `dbPrefix`? I don't think it can ever change... So we probably don't need to allocate a new slice for it whenever we need it.

a1k0n: You mean dbSuffix? And as a member of the struct, not a constant?

It's only ever used by Compact(), as there are no iterators that define a limit in the API. So, maybe?

database/prefixdb/db_test.go
```diff
@@ -25,6 +26,19 @@ func TestInterface(t *testing.T) {
...
 
+func TestSuffix(t *testing.T) {
+	testString := []string{"hello", "world", "a\xff", "\x01\xff\xff\xff\xff"}
+	expected := []string{"hellp", "worle", "b\x00", "\x02\x00\x00\x00\x00"}
+	for i, str := range testString {
+		db := newDB([]byte(str), nil)
+		suf := db.suffix()
+		if !bytes.Equal(*suf, []byte(expected[i])) {
+			t.Fatalf("Expected %s but got %s", expected[i], string(*suf))
+		}
```
StephenButtolph, line 37: nit: can we use the `github.com/stretchr/testify/require` library here to align with the other tests?

database/prefixdb/db.go
a1k0n: ok sure, pushing

StephenButtolph:
> random thought: `suffix` is a misleading name here, it should really be `dbLimit`, not `dbSuffix`. say the word and I'll push the change.

dbLimit makes sense to me

database/prefixdb/db_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/stretchr/testify/require"
```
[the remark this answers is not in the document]
a1k0n: yep.

StephenButtolph (approved):
lgtm

dhrubabasu (approved):

PR #3002, opened 2024-05-07
[tmpnet] Ensure tmpnet compatibility with windows

marun:
## Why this should be merged

This avoids breaking unit test discovery on windows for any package that imports tmpnet.

## How this works

Separate os-specific process configuration into files that are conditionally compiled.

## How this was tested

CI should be sufficient


scripts/build_test.sh
```diff
@@ -9,12 +9,6 @@ source "$AVALANCHE_PATH"/scripts/constants.sh
 
 EXCLUDED_TARGETS="| grep -v /mocks | grep -v proto | grep -v tests/e2e | grep -v tests/upgrade"
 
-GOOS=$(go env GOOS)
-if [[ "$GOOS" == "windows" ]]; then
-  # tmpnet is not compatible with windows
-  EXCLUDED_TARGETS="${EXCLUDED_TARGETS} | grep -v tests/fixture"
-fi
-
```
abi87, line 17: I guess now tmpnet is windows compatible

joshua-kim (approved):

tests/fixture/tmpnet/detached_process_windows.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+//go:build windows
+
+package tmpnet
+
+import "os/exec"
+
+func configureDetachedProcess(cmd *exec.Cmd) {
```
StephenButtolph, line 10: I think this probably wasn't caught because we don't run the linter on different OSes but normally this would fail linting:
```suggestion
func configureDetachedProcess(*exec.Cmd) {
```

tests/fixture/tmpnet/node_process.go
```diff
@@ -114,11 +114,8 @@ func (p *NodeProcess) Start(w io.Writer) error {
 
 	// All arguments are provided in the flags file
 	cmd := exec.Command(p.node.RuntimeConfig.AvalancheGoPath, "--config-file", p.node.getFlagsPath()) // #nosec G204
-
 	// Ensure process is detached from the parent process so that an error in the parent will not affect the child
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Setsid: true,
-	}
+	configureDetachedProcess(cmd)
```
StephenButtolph, line 118: To clarify - this PR is to make it so that this code can _compile_ on windows, but there is no expectation for it to _run_ on windows?

PR #3003, opened 2024-05-08
[tmpnet] Enable single node networks

marun:
## Why this should be merged

Single-node networks can be useful for development.

## How this works

Allow the number of nodes to be configured with `--node-count` to the e2e suite and tmpnetctl and ensure `--sybil-protection-enabled=false` is configured when a network is started with a single node.

## How this was tested

CI


tests/fixture/tmpnet/node.go
```diff
@@ -104,16 +104,16 @@ func NewEphemeralNode(flags FlagsMap) *Node {
 }
 
 // Initializes the specified number of nodes.
-func NewNodes(count int) ([]*Node, error) {
```
joshua-kim, line 107: nit: very perl-esque naming... I think something like `OrPanic` or `OrFatal` is probably more in the style of Go

joshua-kim (approved):

abi87 (approved):

StephenButtolph (approved):

PR #3005, opened 2024-05-09
[tmpnet] Bootstrap subnets with a single node

marun:
## Why this should be merged

Speeds up subnet creation for a network of multiple nodes.

A 10 node network with 2 xsvm subnets takes 
 - old: ~110s
 - new: ~20s 

## How this works

 - start a single node with sybil protection disabled
 - create subnets and chains
 - restart the node with sybil protection enabled
 - start the remaining nodes

## How this was tested

CI

### DIFF
--- a/tests/e2e/x/transfer/virtuous.go
+++ b/tests/e2e/x/transfer/virtuous.go
@@ -55,7 +55,7 @@ var _ = e2e.DescribeXChainSerial("[Virtuous Transfer Tx AVAX]", func() {
 				require.NoError(err)
 
 				for _, metrics := range allNodeMetrics {
-					xBlksProcessing, ok := tests.GetFirstMetricValue(metrics, xBlksProcessingMetric)
+					xBlksProcessing, ok := tests.GetMetricValue(metrics, xBlksProcessingMetric, nil)
 					if !ok || xBlksProcessing > 0 {
 						return false
 					}
@@ -248,13 +248,13 @@ RECEIVER  NEW BALANCE (AFTER) : %21d AVAX
 
 					// +0 since X-chain tx must have been processed and accepted
 					// by now
-					currentXBlksProcessing, _ := tests.GetFirstMetricValue(mm, xBlksProcessingMetric)
-					previousXBlksProcessing, _ := tests.GetFirstMetricValue(prev, xBlksProcessingMetric)
+					currentXBlksProcessing, _ := tests.GetMetricValue(mm, xBlksProcessingMetric, nil)
+					previousXBlksProcessing, _ := tests.GetMetricValue(prev, xBlksProcessingMetric, nil)
 					require.Equal(currentXBlksProcessing, previousXBlksProcessing)
 
 					// +1 since X-chain tx must have been accepted by now
-					currentXBlksAccepted, _ := tests.GetFirstMetricValue(mm, xBlksAcceptedMetric)
-					previousXBlksAccepted, _ := tests.GetFirstMetricValue(prev, xBlksAcceptedMetric)
+					currentXBlksAccepted, _ := tests.GetMetricValue(mm, xBlksAcceptedMetric, nil)
+					previousXBlksAccepted, _ := tests.GetMetricValue(prev, xBlksAcceptedMetric, nil)
 					require.Equal(currentXBlksAccepted, previousXBlksAccepted+1)
 
 					metricsBeforeTx[u] = mm
--- a/tests/metrics.go
+++ b/tests/metrics.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/ava-labs/avalanchego/api/metrics"
 
 	dto "github.com/prometheus/client_model/go"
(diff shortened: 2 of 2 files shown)


StephenButtolph:
I'm pretty surprised this is significantly faster... Do we know why?

tests/fixture/e2e/flags.go
```diff
@@ -82,11 +87,17 @@ func RegisterFlags() *FlagVars {
...
+		false,
+		"[optional] whether to delay network shutdown to allow a final metrics scrape. The duration is set with --network-shutdown-delay.",
+	)
 	flag.DurationVar(
 		&vars.networkShutdownDelay,
 		"network-shutdown-delay",
 		12*time.Second, // Make sure this value takes into account the scrape_interval defined in scripts/run_prometheus.sh
-		"[optional] the duration to wait before shutting down the test network at the end of the test run. A value greater than the scrape interval is suggested. 0 avoids waiting for shutdown.",
+		"[optional] the duration to wait before shutting down the test network at the end of the test run. A value greater than the scrape interval is suggested.",
 	)
```
StephenButtolph, line 101: Why do we have 2 values for this? Seems like it would make more sense to just have the `--network-shutdown-delay` flag that defaults to `0`.

tests/fixture/tmpnet/network.go
```diff
@@ -358,6 +362,48 @@ func (n *Network) Start(ctx context.Context, w io.Writer) error {
...
+		return err
+	}
+	if err := n.CreateSubnets(ctx, w); err != nil {
+		return err
+	}
+
+	if _, err := fmt.Fprintf(w, "Restarting %s with sybil protection enabled to serve as bootstrap node\n", bootstrapNode.NodeID); err != nil {
+		return err
+	}
+	delete(bootstrapNode.Flags, config.SybilProtectionEnabledKey)
```
StephenButtolph, line 392: I suppose we don't think it would make sense for the network to actually want sybil protection disabled in tests... I think that makes sense (it's really a flag intended for quick local testing... not for programatic testing

tests/fixture/tmpnet/network.go
```diff
@@ -781,6 +796,41 @@ func (n *Network) getBootstrapIPsAndIDs(skippedNode *Node) ([]string, []string,
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("failed to see all nodes healthy before timeout: %w", ctx.Err())
+		case <-ticker.C:
+		}
+	}
+	return nil
+}
```
StephenButtolph, line 832: The only actual change here is that this this reduces the time this function takes to return by `networkHealthCheckInterval`.

I also felt it was a bit cleaner to not iterate over nodes that were already healthy... but 🤷

 
```suggestion
// Waits until the provided nodes are healthy.
func waitForHealthy(ctx context.Context, w io.Writer, nodes []*Node) error {
	ticker := time.NewTicker(networkHealthCheckInterval)
	defer ticker.Stop()

	unhealthyNodes := set.Of(nodes...)
	for {
		for node := range unhealthyNodes {
			healthy, err := node.IsHealthy(ctx)
			if err != nil && !errors.Is(err, ErrNotRunning) {
				return err
			}
			if !healthy {
				continue
			}

			unhealthyNodes.Remove(node)
			if _, err := fmt.Fprintf(w, "%s is healthy @ %s\n", node.NodeID, node.URI); err != nil {
				return err
			}
		}

		if unhealthyNodes.Len() == 0 {
			return nil
		}

		select {
		case <-ctx.Done():
			return fmt.Errorf("failed to see all nodes healthy before timeout: %w", ctx.Err())
		case <-ticker.C:
		}
	}
}
```

tests/fixture/tmpnet/node.go
```diff
@@ -369,6 +369,10 @@ func (n *Node) EnsureNodeID() error {
 // labeling of metrics.
 func (n *Node) SaveAPIPort() error {
 	hostPort := strings.TrimPrefix(n.URI, "http://")
+	if len(hostPort) == 0 {
+		// Without a staking address there is nothing to save
```
StephenButtolph, line 373: This isn't the staking address is it? This is the API address right?

tests/fixture/e2e/flags.go
StephenButtolph: If we're adding `--delay-network-shutdown` so that the user doesn't need to know the metrics scrape interval, is there ever a time we expect the user to specify the shutdown delay?

I'm wondering if we could remove `--network-shutdown-delay` if we don't want to specify `--network-shutdown-delay=12s` in the CI runs.

tests/fixture/tmpnet/network.go
```diff
@@ -325,26 +322,33 @@ func (n *Network) Create(rootDir string) error {
...
-	if _, err := fmt.Fprintf(w, "Starting network %s (UUID: %s)\n", n.Dir, n.UUID); err != nil {
-		return err
+// Starts the specified nodes
+func (n *Network) Start(ctx context.Context, w io.Writer, nodesToStart ...*Node) error {
+	nodesToWaitFor := nodesToStart
+	if len(nodesToStart) < len(n.Nodes) && nodesToStart[0] != n.Nodes[0] {
+		// Wait for all nodes to ensure health is logged for the bootstrap node
+		nodesToWaitFor = n.Nodes
+	} else {
+		// Simplify output by only logging network start for the first node or when all nodes are starting at once
```
StephenButtolph, line 332: If all the nodes are starting at once, we're also starting the first node right?

tests/fixture/tmpnet/network.go
```diff
@@ -718,22 +774,20 @@ func (n *Network) CreateSubnets(ctx context.Context, w io.Writer) error {
...
+
+		// Restart nodes to allow configuration for the new chains to take effect
+		for _, node := range n.Nodes {
+			if !validatorsToRestart.Contains(node.NodeID) {
+				continue
+			}
+			if err := n.RestartNode(ctx, w, node); err != nil {
+				return err
+			}
+		}
```
StephenButtolph, line 790: Is there a reason we re-indented this chunk of code? It seems like we could have just modified the single if check to exit the function previously.

I think we generally try to avoid nesting to improve readability.

StephenButtolph (approved):

tsachiherman (approved):

PR #3007, opened 2024-05-09
Generalize validator sampling

StephenButtolph:

### DIFF
--- a/network/p2p/validators.go
+++ b/network/p2p/validators.go
@@ -35,14 +35,12 @@ type ValidatorSubset interface {
 }
 
 func NewValidators(
-	peers *Peers,
 	log logging.Logger,
 	subnetID ids.ID,
 	validators validators.State,
 	maxValidatorSetStaleness time.Duration,
 ) *Validators {
 	return &Validators{
-		peers:                    peers,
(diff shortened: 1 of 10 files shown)


network/p2p/validators.go
```diff
@@ -131,7 +133,7 @@ func (v *Validators) Sample(ctx context.Context, limit int) []ids.NodeID {
 		}
 
 		nodeID := v.validatorList[i].nodeID
-		if !v.peers.has(nodeID) {
+		if !canReturn(nodeID) {
```
StephenButtolph, line 136: Could we refactor this to reduce duplication?

utils/set/slice.go
```diff
@@ -0,0 +1,67 @@
...
+// See the file LICENSE for licensing terms.
+
+package set
+
+import (
+	"github.com/ava-labs/avalanchego/utils"
+)
+
+// Slice is a set of elements that supports indexing.
+type Slice[T comparable] struct {
```
StephenButtolph, line 9: Should we remove the `SampleableSet` implementation in favor of this?

StephenButtolph:
I don't remember why we wanted this.

PR #3011, opened 2024-05-10
[e2e] Simplify pre-funded key usage

marun:
## Why this should be merged

As part of refactoring the e2e tests for reuse in antithesis and load testing, this change simplifies key usage so that tests can be supplied a specific pre-funded key rather than retrieving their own key at runtime. Fund usage is also minimized . It should be possible to run most tests repeatedly for extended periods of time without running out of keys or funds.

## How this works

Previously, pre-funded keys were distributed via an http server to ensure a given key was used by at most a single test. This was intended to ensure compatibility with parallel test execution by guaranteeing that no pre-funded key would ever be used concurrently.

Since ginkgo uses process-based parallization, it's sufficient to ensure that each process is allocated a unique key. The tests assigned to run in a given test process are executed sequentially so a key won't be in danger of concurrent usage.

As part of this change, the virtuous test was updated to use the one funded key to fund a new set of keys. Previously 10 pre-funded keys were required.

## How this was tested

CI, local parallel execution


abi87 (approved):
lgtm

tests/e2e/x/transfer/virtuous.go
```diff
@@ -76,7 +79,39 @@ var _ = e2e.DescribeXChainSerial("[Virtuous Transfer Tx AVAX]", func() {
...
+								Addrs: []ids.ShortID{
+									key.Address(),
+								},
+							},
+						},
+					}},
+					tc.WithDefaultContext(),
+				)
+				require.NoError(err)
+			}
```
StephenButtolph, line 114: Would it make more sense to do this in a single transaction? It seems like this funding step might take a little while as it is currently coded (as each `IssueBaseTx` will issue the tx and then block on its acceptance).

tests/e2e/x/transfer/virtuous.go
```diff
@@ -76,7 +79,39 @@ var _ = e2e.DescribeXChainSerial("[Virtuous Transfer Tx AVAX]", func() {
...
 
 			runFunc := func(round int) {
```
StephenButtolph, line 116: Note to myself: This test looks... Interesting... Should look to see what the use actually is.

StephenButtolph (approved):

dhrubabasu (approved):

PR #3014, opened 2024-05-10
Suggested nits for generalize peer sampling

joshua-kim:
## Why this should be merged

Suggestions to make the p2p api a bit cleaner for https://github.com/ava-labs/avalanchego/pull/3007

## How this works

1. Attaches the `Validators` struct to `Network` to be consistent w/ `Peers`
2. Gets rid of the `canReturn` function being passed in from the callers with filters provided at the time of client creation

## How this was tested

Unit tests


network/p2p/sampler.go
```diff
@@ -0,0 +1,59 @@
...
+	for len(sampled) < limit {
+		i, err := uniform.Next()
+		if err != nil {
+			break
+		}
+
+		nodeID := p.peers.set.Elements[i]
+		for _, filter := range p.filters {
+			if !filter(ctx, nodeID) {
+				continue
```
StephenButtolph, line 51: This `continue` is going to apply to the inner `for` loop - not the outer `for` loop.

network/p2p/sampler.go
```diff
@@ -0,0 +1,59 @@
...
+
+type SamplingFilter func(ctx context.Context, nodeID ids.NodeID) bool
+
+// NewPeerValidatorFilter returns a filter that only samples other peer
+// validators
+func NewPeerValidatorFilter(self ids.NodeID, validators ValidatorSet) SamplingFilter {
+	return func(ctx context.Context, nodeID ids.NodeID) bool {
+		return nodeID != self && validators.Has(ctx, nodeID)
+	}
+}
```
StephenButtolph, line 18: I think this should be 2 different filters. One for filtering a set of peers. Another for filtering for validators.

network/p2p/network.go
```diff
@@ -153,24 +151,22 @@ func (n *Network) Disconnected(_ context.Context, nodeID ids.NodeID) error {
...
 		handlerID:     handlerID,
 		handlerIDStr:  strconv.FormatUint(handlerID, 10),
 		handlerPrefix: ProtocolPrefix(handlerID),
 		sender:        n.sender,
 		router:        n.router,
-		options: &clientOptions{
-			nodeSampler: n.Peers,
-		},
+		options:       clientOptions,
+		sampler:       newPeerSampler(n.Peers, clientOptions.samplingFilters...),
```
StephenButtolph, line 168: While we don't currently use it, it feels like it may behoove us to support sampling by stake weight as well (this only really supports sampling uniformly).

network/p2p/network.go
```diff
@@ -105,7 +113,8 @@ func NewNetwork(
 // Network exposes networking state and supports building p2p application
 // protocols
 type Network struct {
-	Peers *Peers
+	Peers      *Peers
+	Validators *Validators
```
joshua-kim, line 117: This is technically an unrelated change so I can pull this out of the diff if we feel like this makes the review better but it's sort of in the vein of making the sampling more generic since this allows us to in the future configure sampling by stake if we had something like:

```
n := NewNetwork(...)
client := n.NewClient(0x0, WithPeerSampler(NewStakeWeightedPeerSampler(n.Validators)))
```

Technically we don't need to bind this onto the `Network` struct to do the above change but I feel like it is more consistent with the fact that `Peers` is bound to the `Network` struct + I think it looks cleaner to have it in one spot instead of the VM managing all of these + it lets us use validator stake sampling as a default if we want to go that route.

network/p2p/network_test.go
```diff
@@ -480,141 +560,58 @@ func TestAppRequestDuplicateRequestIDs(t *testing.T) {
 	require.ErrorIs(err, ErrRequestPending)
 }
 
-// Sample should always return up to [limit] peers, and less if fewer than
-// [limit] peers are available.
-func TestPeersSample(t *testing.T) {
```
joshua-kim, line 485: Sample is no longer exported out of the package so we're testing this through `AppRequestAny` now

network/p2p/network_test.go
```diff
@@ -629,122 +626,30 @@ func TestAppRequestAnyNodeSelection(t *testing.T) {
...
+			client := n.NewClient(1, WithSamplingFilters(tt.filters...))
 
 			err = client.AppRequestAny(context.Background(), []byte("foobar"), nil)
 			require.ErrorIs(err, tt.expected)
 			require.Subset(tt.expectedPeers, sent.List())
-		})
-	}
-}
 
-func TestNodeSamplerClientOption(t *testing.T) {
```
joshua-kim, line 647: This is now replaced w/ the `AppRequestAny` test + tests on the filters.

network/p2p/validators.go
```diff
@@ -108,43 +106,7 @@ func (v *Validators) refresh(ctx context.Context) {
...
-
-		sampled = append(sampled, nodeID)
-	}
-
-	return sampled
-}
-
-// Top returns the top [percentage] of validators, regardless of if they are
-// connected or not.
+// Top returns the top [percentage] of validators by stake
```
joshua-kim, line 109: Changed this comment because this struct doesn't care about connectivity now

network/p2p/validators.go
```diff
@@ -171,7 +133,7 @@ func (v *Validators) Top(ctx context.Context, percentage float64) []ids.NodeID {
 	return top
 }
 
-// Has returns if nodeID is a connected validator
+// Has returns if nodeID is a validator
```
joshua-kim, line 136: This is an invariant change. I believe this doesn't matter because:
1. In the `ValidatorGossiper` we don't care if we are connected to ourselves or not
2. In our `ValidatorHandler` if we are receiving a message over p2p we are connected to the peer

I'm not actually sure why we cared that this peer was connected to begin with... so either I overlooked this in the original implementation or I am forgetting something and this change is introducing a bug.

network/p2p/network.go
StephenButtolph: Hm - `Peers` makes sense to live on the `Network` struct because it is modified by calls to `Connected`/`Disconnected`. It feels pretty odd to me to pass `Validators` around on the `Network` struct even though it is never directly used.

network/p2p/client.go
```diff
@@ -42,11 +42,11 @@ type CrossChainAppResponseCallback func(
 )
 
 type Client struct {
-	self          ids.NodeID
 	handlerID     uint64
 	handlerIDStr  string
 	handlerPrefix []byte
 	router        *router
+	peers         *Peers
```
StephenButtolph, line 49: I was trying to avoid directly referencing the `Peers` struct in the client to support custom samplers (like sampling validators by stake). I feel like defining an interface we're happy with here is how we should be enabling different sampling mechanisms.

How would this look if we were enabling sampling by stake?

joshua-kim: Added an interface that exposes the peer slice

network/p2p/sampler.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/ava-labs/avalanchego/utils/sampler"
+)
+
+var _ Sampler = (*UniformSampler)(nil)
+
+// Sampler samples connected peers
+type Sampler interface {
+	// Sample returns a nodeID from nodeIDs
+	// invariant: nodeIDs must not be modified
+	Sample(ctx context.Context, nodeIDs []ids.NodeID) (ids.NodeID, bool)
```
joshua-kim, line 19: Signature was changed to only return a single node since the intended use case for this was to choose a node to send a request to if the caller didn't provide one (i.e `AppRequestAny`). This also gets rid of edge-cases like what we have to do if the sampler implementation gives us more than the old `limit` parameter's amount of nodes.

StephenButtolph:
Parent PR is closed.

PR #3016, opened 2024-05-13
[antithesis] Ensure images with a prefix are pushed

marun:
https://github.com/ava-labs/avalanchego/pull/2976 accidentally disabled image push.


StephenButtolph:
Should we change the name of the action to `build and push`? Having just `build` without a push step makes it seem like we aren't pushing the image

darioush (approved):

abi87 (approved):

PR #3017, opened 2024-05-13
Standardize peer logs

StephenButtolph:
## Why this should be merged

Uses constants to show standardized logs. Using standardized logs is nice to simplify centralized logging queries.

Standardizes some edge case behavior around closing connections if messages fail to be built (which should essentially never happen)

## How this works

Defines some constants and moves a bit of code around.

## How this was tested

- [X] CI


network/peer/peer.go
```diff
@@ -635,21 +646,18 @@ func (p *peer) sendNetworkMessages() {
...
 			msg, err := p.Config.MessageCreator.GetPeerList(knownPeersFilter, knownPeersSalt)
 			if err != nil {
-				p.Log.Error("failed to create get peer list message",
+				p.Log.Error(failedToCreateMessageLog,
 					zap.Stringer("nodeID", p.id),
+					zap.Stringer("messageOp", message.GetPeerListOp),
 					zap.Error(err),
 				)
-				continue
+				return
```
StephenButtolph, line 654: In most places we closed the connection after failing to build a message... This PR standardizes that behavior.

network/peer/peer.go
```diff
@@ -635,21 +646,18 @@ func (p *peer) sendNetworkMessages() {
...
 			}
 
-			if !p.Send(p.onClosingCtx, msg) {
-				p.Log.Debug("failed to send get peer list",
-					zap.Stringer("nodeID", p.id),
-				)
-			}
```
StephenButtolph, line 649: This is already logged internally.

network/peer/peer.go
```diff
@@ -774,17 +781,67 @@ func (p *peer) handle(msg message.InboundMessage) {
 }
 
 func (p *peer) handlePing(msg *p2p.Ping) {
-	p.observeUptimes(msg.Uptime, msg.SubnetUptimes)
```
StephenButtolph, line 777: Previously this function could have started closing the connection, but then we would have still attempted to send a pong message... Which is pretty odd... We no longer handle the uptimes in `Pong` messages, so this function was just inlined.

network/peer/peer.go
```diff
@@ -774,17 +781,67 @@ func (p *peer) handle(msg message.InboundMessage) {
...
 	pongMessage, err := p.MessageCreator.Pong()
 	if err != nil {
-		p.Log.Error("failed to create message",
-			zap.Stringer("messageOp", message.PongOp),
+		p.Log.Error(failedToCreateMessageLog,
 			zap.Stringer("nodeID", p.id),
+			zap.Stringer("messageOp", message.PongOp),
 			zap.Error(err),
 		)
+		p.StartClose()
```
StephenButtolph, line 841: The peer may close us anyways if we'd dropping the ping message... so we can just close now and maintain the pattern that failing to build a message will close the connection.

network/peer/peer.go
```diff
@@ -885,44 +896,46 @@ func (p *peer) observeUptime(subnetID ids.ID, uptime uint32) {
 
 func (p *peer) handleHandshake(msg *p2p.Handshake) {
 	if p.gotHandshake.Get() {
-		// TODO: this should never happen, should we close the connection here?
```
StephenButtolph, line 888: We now close the connection.

network/peer/peer.go
```diff
@@ -635,21 +646,18 @@ func (p *peer) sendNetworkMessages() {
...
-				continue
+				return
 			}
 
-			if !p.Send(p.onClosingCtx, msg) {
-				p.Log.Debug("failed to send get peer list",
-					zap.Stringer("nodeID", p.id),
-				)
-			}
+			p.Send(p.onClosingCtx, msg)
```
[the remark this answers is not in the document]
StephenButtolph: We only require that for returned errors.

joshua-kim (approved):

PR #3019, opened 2024-05-14
Repackaged upgrades times into upgrade package

abi87:

### DIFF
--- a/vms/platformvm/upgrade/config.go
+++ b/vms/platformvm/upgrade/config.go
@@ -1,0 +1,50 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package upgrade
+
+import "time"
+
+type Config struct {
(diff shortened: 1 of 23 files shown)


vms/platformvm/upgrade/config.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package upgrade
+
+import "time"
+
+type Config struct {
```
abi87, line 8: We repack this config to avoid circular dependencies down the line.
As per @joshua-kim's suggestion, we'll repack fee related config into a fee package (along with dynamic fees calculators).
Since fee may depend on the specific fork active the fee package requires upgrades config and is required by vm config package `config`.
Repackaging upgrades breaks the circular dependency

joshua-kim (approved):

StephenButtolph (approved):

PR #3023, opened 2024-05-16
[vms/avm] fix linter error in benchmark : Use of weak random number generator

tsachiherman:
## Why this should be merged

The benchmark `BenchmarkGetUTXOs` was using `rand.Uint32()` as part of it's implementation.
While it's technically correct, it could lead to inconsistent results if there were some underlying issue triggered by specific values. Moreover, this would be expressed as a random failure which is (might) be very hard to reproduce.

Three additional linter-related changes:
1. removal of `//nolint:goconst`, and implement changes ( intuitive )
2. removal of `// #nosec G404`, which wan't needed ( maybe it was needed historically ? )
3. relocating `HTTPWriteTimeoutKey` to avoid "G101: Potential hardcoded credentials"

## How this works

Instead of using `rand.Uint32()`, I've used a deterministic random source. This would provide the test with pseudo-random value that won't change across iterations.

## How this was tested

I ran the benchmark manually and verified it works as expected.


vms/avm/vm_benchmark_test.go
```diff
@@ -81,7 +81,7 @@ func GetAllUTXOsBenchmark(b *testing.B, utxoCount int) {
 		utxo := &avax.UTXO{
 			UTXOID: avax.UTXOID{
 				TxID:        ids.GenerateTestID(),
-				OutputIndex: rand.Uint32(),
+				OutputIndex: uint32(randSrc.Int63()),
```
StephenButtolph, line 84: Do we still need the `// #nosec G404` above this? It doesn't seem like we do... I would have thought the `nolintlint`er would catch this... But maybe it doesn't check `#nosec` commands...

tsachiherman: I will look into that and remove that if it's no longer needed.

tsachiherman: yes, that's true. I've removed that.

StephenButtolph (approved):

PR #3024, opened 2024-05-17
[build] Update linter version

tsachiherman:
## Why this should be merged
This PR updates the golangci-lint tool to it's latest release (1.58.1).
In addition it resolves the newly detected issues.

## How this works
Nothing really notable here.
1. The linter configuration file is using `exclude-dirs-use-default` instead of the deprecated `skip-dirs-use-default`.
2. The new gosec requires that the suppress annotation would be on the same line.
3. The fundedSharedMemory used to use a random number. While it's logic is correct, it might randomly fail due to this random number generator. I've switch that to use a fixed-random-source to address this.
4. I've updated the linter version.

## How this was tested

Tested manually as well as using github actions.

### DIFF
--- a/network/peer/tls_config.go
+++ b/network/peer/tls_config.go
@@ -23,7 +23,7 @@ func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
 		//
 		// During our security audit by Quantstamp, this was investigated
 		// and confirmed to be safe and correct.
-		InsecureSkipVerify: true, // #nosec G402
+		InsecureSkipVerify: true, //#nosec G402
 		MinVersion:         tls.VersionTLS13,
 		KeyLogWriter:       keyLogWriter,
 	}
--- a/network/peer/tls_config.go
+++ b/network/peer/tls_config.go
@@ -14,7 +14,6 @@ import (
 // It is safe, and typically expected, for [keyLogWriter] to be [nil].
 // [keyLogWriter] should only be enabled for debugging.
 func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
-	// #nosec G402
 	return &tls.Config{
 		Certificates: []tls.Certificate{cert},
 		ClientAuth:   tls.RequireAnyClientCert,
@@ -24,7 +23,7 @@ func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
 		//
 		// During our security audit by Quantstamp, this was investigated
 		// and confirmed to be safe and correct.
-		InsecureSkipVerify: true,
+		InsecureSkipVerify: true, // #nosec G402
 		MinVersion:         tls.VersionTLS13,
 		KeyLogWriter:       keyLogWriter,
 	}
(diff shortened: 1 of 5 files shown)


scripts/lint.sh
```diff
@@ -32,7 +32,9 @@ fi
 TESTS=${TESTS:-"golangci_lint license_header require_error_is_no_funcs_as_params single_import interface_compliance_nil require_no_error_inline_func"}
 
 function test_golangci_lint {
-  go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.1
+  # We use the commit hash (dc2815312726b4f374f5c69a1c183a85990917a7) to reference the specific version (v1.58.1) used in our build process.
+  # This guarantees a known-good and verifiable version. Unlike releases, commits cannot be altered or deleted after they are made, providing a more reliable reference point for our build system.
```
StephenButtolph, line 36: Does this actually protect against anything? Google maintains protections against changing the checksum of the provided code against a given tag already... If we assumed this protection didn't exist, couldn't the maintainer of the repo just delete the old commit and tag a new commit with the hash of the old commit?

Unless I'm missing something this feels like some additional complexity that we probably don't need to maintain.

vms/platformvm/txs/executor/import_test.go
```diff
@@ -180,7 +186,7 @@ func fundedSharedMemory(
 		utxo := &avax.UTXO{
```
StephenButtolph, line 186: Should we remove the `// #nosec G404` here?

network/peer/tls_config.go
```diff
@@ -24,7 +23,7 @@ func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
 		//
 		// During our security audit by Quantstamp, this was investigated
 		// and confirmed to be safe and correct.
-		InsecureSkipVerify: true,
+		InsecureSkipVerify: true, // #nosec G402
```
StephenButtolph, line 26: I know we weren't doing this correctly before... But I think it's standard to have comments for automated systems not to start with a space.
```suggestion
		InsecureSkipVerify: true, //#nosec G402
```

tsachiherman: I wasn't aware of that standard, but don't have any reason not to adhere to that.

scripts/lint.sh
tsachiherman: afaik, the protection against changing the hash is only in go.mod/go.sum. Not for "random" tools being installed.
Let's hash that in private.

vms/platformvm/txs/executor/import_test.go
tsachiherman: removed.

network/peer/tls_config.go
StephenButtolph: I think I first saw this here: https://golangci-lint.run/usage/false-positives/

> Use `//nolint` instead of `// nolint` because machine-readable comments should have no space by Go convention.

There is probably a reference in a more "standard" golang place... but I couldn't quickly find it.

scripts/lint.sh
tsachiherman: I've rollback this change so that we can de-couple this one from this PR. I will keep working on validating this in the background.

StephenButtolph (approved):

darioush (approved):

PR #3025, opened 2024-05-17
Change default staking key from RSA 4096 to secp256r1

StephenButtolph:
## Why this should be merged

`RSA 4096` is very slow compared to `secp256r1` for a comparable level of security. In the future we want to move toward using `ED25519`, but using `ED25519` would require a network upgrade. We can change the default value to `secp256r1` today without a network upgrade.

[Benchamarks](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat):
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/staking
                      │      old.txt      │               new.txt                │
                      │      sec/op       │    sec/op     vs base                │
Parse-12                     449.6n ±  5%   469.0n ±  2%   +4.33% (p=0.015 n=10)
NewCertAndKeyBytes-12   1205790.64µ ± 28%   98.23µ ± 10%  -99.99% (p=0.000 n=10)
Sign-12                    5477.39µ ±  4%   17.22µ ±  3%  -99.69% (p=0.000 n=10)
Verify-12                   136.76µ ±  2%   49.48µ ±  3%  -63.82% (p=0.000 n=10)
geomean                      798.3µ         14.08µ        -98.24%
```

## How this works

When generating a TLS certificate, this now generates an `ECDSA 256` key rather than an `RSA 4096` key.

## How this was tested

- [X] CI
- [X] Synced Fuji
- [X] Added Benchamrk


StephenButtolph:
Before:
```
ok  	github.com/ava-labs/avalanchego/network/peer	55.400s	coverage: 56.0% of statements
```

After:
```
ok  	github.com/ava-labs/avalanchego/network/peer	1.546s	coverage: 56.5% of statements
```

JacobEverly (approved):

PR #3026, opened 2024-05-17
Simplify sampler interface

StephenButtolph:
## Why this should be merged

Currently the sampler interface returns an `error` even though it only means one thing `had enough weight` or `didn't have enough weight`.

## How this works

This PR replaces the `error` with a `bool`.

## How this was tested

- [X] CI

### DIFF
--- a/snow/consensus/snowman/bootstrapper/sampler.go
+++ b/snow/consensus/snowman/bootstrapper/sampler.go
@@ -4,10 +4,14 @@
 package bootstrapper
 
 import (
+	"errors"
+
 	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/utils/sampler"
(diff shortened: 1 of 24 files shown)


snow/consensus/snowman/bootstrapper/sampler.go
```diff
@@ -36,9 +40,9 @@ func Sample[T comparable](elements map[T]uint64, maxSize int) (set.Set[T], error
 	}
 
 	maxSize = int(min(uint64(maxSize), totalWeight))
-	indices, err := sampler.Sample(maxSize)
-	if err != nil {
-		return nil, err
+	indices, ok := sampler.Sample(maxSize)
+	if !ok {
+		return nil, errUnexpectedSamplerFailure
```
StephenButtolph, line 45: This should never happen because of the `maxSize` calculation above.

darioush (approved):

PR #3027, opened 2024-05-19
fix: broken links regarding tooling

cocoyeal:
Note: in this PR, i only fixed the broken links regarding dir tooling. but there are more broken links regarding other dir like nodes and reference. i will draft other PRs to fix them one by one since the number of files being affected will be big.

## Why this should be merged
broken links being fixed.

## How this works
by replacing the broken links with the right links.

## How this was tested
by manually checking if the replaced links works.

### DIFF
--- a/api/admin/service.md
+++ b/api/admin/service.md
@@ -13,7 +13,7 @@ This API can be used for measuring node health and debugging.
 The Admin API is disabled by default for security reasons. To run a node with the Admin API
 enabled, use [config flag `--api-admin-enabled=true`](/nodes/configure/avalanchego-config-flags.md#--api-admin-enabled-boolean).
 
-This API set is for a specific node, it is unavailable on the [public server](/tooling/rpc-providers.md).
+This API set is for a specific node, it is unavailable on the [public server](https://docs.avax.network/tooling/rpc-providers).
 
 :::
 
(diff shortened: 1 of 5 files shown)


api/admin/service.md
```diff
@@ -13,7 +13,7 @@ This API can be used for measuring node health and debugging.
 The Admin API is disabled by default for security reasons. To run a node with the Admin API
 enabled, use [config flag `--api-admin-enabled=true`](/nodes/configure/avalanchego-config-flags.md#--api-admin-enabled-boolean).
 
-This API set is for a specific node, it is unavailable on the [public server](/tooling/rpc-providers.md).
+This API set is for a specific node, it is unavailable on the [public server](https://docs.avax.network/tooling/rpc-providers).
```
StephenButtolph, line 16: For example, this page is rendered here: https://docs.avax.network/reference/avalanchego/admin-api and the link seems to work as expected.

StephenButtolph (commented):
The `service.md` files are actually used to generate the https://docs.avax.network/ website. These links are intentionally set to local paths for the website.

I checked the website and it seems like these links are working as expected there. If you find any broken links on the site please make a new PR to fix those links here!

cocoyeal:
Hi @StephenButtolph , thanks for the info. i do find one broken link, will make a new PR to fix it.

PR #3032, opened 2024-05-20
P-Chain: Block-level fee Calculator

abi87:
## Why this should be merged
Instantiate `fee.Calculator` only once per block, instead of doing it once per each block transaction.

## How this works
`fee.Calculator` is instantianted once in each block. UTs need to be modified to instantiate the calculator.

## How this was tested
CI


vms/platformvm/txs/fee/calculator.go
```diff
@@ -13,24 +13,26 @@ import (
 
 var _ txs.Visitor = (*calculator)(nil)
 
-func NewStaticCalculator(config StaticConfig, upgradeTimes upgrade.Config) Calculator {
-	return Calculator{
+func NewStaticCalculator(config StaticConfig, upgradeTimes upgrade.Config, chainTime time.Time) *Calculator {
```
abi87, line 16: Pass to the constructor all quantities that won't change.for different transactions in a single block and only those.
Pass to CalculateFee all quantities that will change for different transactions in a single block, and only those.

vms/platformvm/txs/fee/calculator.go
```diff
@@ -13,24 +13,26 @@ import (
...
 
 type Calculator struct {
 	config       StaticConfig
 	upgradeTimes upgrade.Config
+	time         time.Time
 }
 
 // [CalculateFee] returns the minimal fee needed to accept [tx], at chain time [time]
-func (c *Calculator) CalculateFee(tx txs.UnsignedTx, time time.Time) uint64 {
+func (c *Calculator) CalculateFee(tx txs.UnsignedTx) uint64 {
```
abi87, line 31: in next PR we will pass tx credentials here as well

vms/platformvm/txs/fee/calculator.go
```diff
@@ -13,24 +13,26 @@ import (
...
 }
 
 type Calculator struct {
 	config       StaticConfig
 	upgradeTimes upgrade.Config
+	time         time.Time
 }
 
 // [CalculateFee] returns the minimal fee needed to accept [tx], at chain time [time]
-func (c *Calculator) CalculateFee(tx txs.UnsignedTx, time time.Time) uint64 {
```
joshua-kim, line 29: Why are we moving this parameter into the constructor? I think if we keep this parameter here we can reuse the same calculator instance for all blocks.

abi87: Apologies for the delayed answer. I cleaned up the PRs following this one, following all of your suggestions but this one.
You're right, `staticCalculator` could be instantiated once and reused per each block if we move `chainTime` among `CalculateFee` parameters, as it is now (even if `chainTime` does not change across block transactions).
The problem is that for `dynamicCalculator` (soon to come) is not that simple to be made reusable for all blocks.
`DynamicCalculator` has a `commonfee.Calculator` core that needs to be update block by block. `commonfee.Calculator` tracks `gasPrice`, `gasCap` and the gas consumed by each block, which need to be updated/reset block by block.
So to instantiate DynamicCalculator once for each block I see two alternatives:

- Add `commonfee.Calculator` as `CalculateFee` parameter, along with `chainTime`.
- Instantiate `fee.Calculator` once, but introduce an `Update` method that takes care of updating/resetting all block by block.

I don't like the first alternative and I find the second alternative not much different that what we have now.

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -98,10 +99,12 @@ func TestStandardTxExecutorAddValidatorTxEmptyID(t *testing.T) {
 		stateDiff, err := state.NewDiff(lastAcceptedID, env)
 		require.NoError(err)
 
+		feeCalculator := fee.NewStaticCalculator(env.config.StaticFeeConfig, env.config.UpgradeConfig, stateDiff.GetTimestamp())
```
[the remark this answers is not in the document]
abi87: Backported from next PRs a function to create a calculator. Will use it later on to decide whether a static or dynamic calculator should be created.

vms/platformvm/state/chain_time_helpers.go
```diff
@@ -68,3 +70,10 @@ func GetNextStakerChangeTime(state Chain) (time.Time, error) {
 		return time.Time{}, database.ErrNotFound
 	}
 }
+
+// [PickFeeCalculator] creates either a static or a dynamic fee calculator, depending on the active upgrade
+// [PickFeeCalculator] does not modify [state]
+func PickFeeCalculator(cfg *config.Config, state Chain) (fee.Calculator, error) {
```
StephenButtolph: I assume this takes in the full state because it will be used in the future to access the excess gas values.

StephenButtolph (approved):

vms/platformvm/state/chain_time_helpers.go
abi87: Correct. Eventually we'll also need parentBlockTimestamp

PR #3038, opened 2024-05-21
Pchain: Multidimensional Static Fees, no wallet

abi87:
## How this works

- Introduce a `fee.Manager` in charge of tracking each block tx's complexity and calculating each block tx fees. This is a component that will be shared among P-chain and X-chain.
- Introduce `platformvm` dynamic fee calculator, in charge of metering P-chain txs complexities
- Update `platformvm` txs verification to select static or dynamic fee calculator, depending on the active fork


vms/platformvm/txs/fee/dynamic_config.go
```diff
@@ -0,0 +1,57 @@
...
+var (
+	eUpgradeDynamicFeesConfig = commonfees.DynamicFeesConfig{
+		FeeRate: commonfees.Dimensions{
+			1 * units.NanoAvax,
+			2 * units.NanoAvax,
+			3 * units.NanoAvax,
+			4 * units.NanoAvax,
+		},
+
+		BlockMaxComplexity: commonfees.Max,
```
abi87, line 25: these values don't really matter.  We'll change these parameters when introducing the dynamic multifee and then we'll tuno those parameters (which will actually see production)

vms/platformvm/block/builder/helpers_test.go
```diff
@@ -275,6 +275,7 @@ func addSubnet(t *testing.T, env *environment) {
 
 	stateDiff.AddTx(testSubnet1, status.Committed)
 	require.NoError(stateDiff.Apply(env.state))
+	require.NoError(env.state.Commit())
```
abi87, line 278: this is a pre-existing bug which will show up when fees won't be zero anymore.

abi87:
Will go for a different split

PR #3040, opened 2024-05-22
Manually compact during bootstrapping iteration

StephenButtolph:
## Why this should be merged

When introducing the interval tree into bootstrapping, the database was designed to support iteration rather than random reads to improve querying performance. However, it appears that holding the the iterator prevents compaction from occurring in a timely manner. Over time (after executing > 1M blocks) the iteration starts to slow down significantly. By manually compacting the database periodically, we manage to avoid this slowdown.

## How this works

Periodically calls `Compact` every 128 iterator releases.

## How this was tested

- [X] CI
- [X] Syncing mainnet P-chain with this code finished execution in ~`6h`. Running master executed `9.1M` of the blocks (out of `13.7M`) in `30h 22m`

### DIFF
--- a/snow/engine/snowman/bootstrap/storage.go
+++ b/snow/engine/snowman/bootstrap/storage.go
@@ -23,7 +23,7 @@ import (
 
 const (
 	batchWritePeriod      = 64
-	iteratorReleasePeriod = 25_000
+	iteratorReleasePeriod = 50_000
 	logPeriod             = 5 * time.Second
 	minBlocksToCompact    = 5000
 )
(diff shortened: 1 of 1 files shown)


dhrubabasu (approved):

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -206,8 +210,22 @@ func execute(
 				return err
 			}
 
+			lastProcessedKey := iterator.Key()
```
abi87, line 213: nit: update comment on lines 200/201 to mention that we don't just reduce memory pressure but we also compact?

abi87 (approved):

StephenButtolph:
Closing in favor of #3045

PR #3045, opened 2024-05-22
Grab iterator at previously executed height

StephenButtolph:
## Why this should be merged

Internally the disk may not have deleted the blocks yet when we grab a new iterator. If that happens, we would previously iterate over all the previously deleted blocks on the first call to `.Next()`.

## How this works

By specifying the start key, we can avoid iterating over deleted (but not garbage collected) entries.

## How this was tested

- [X] CI
- [X] Syncing mainnet P-chain with this code finished execution in ~`6h`. Running master executed `9.1M` of the blocks (out of `13.7M`) in `30h 22m`

### DIFF
--- a/snow/engine/snowman/bootstrap/storage.go
+++ b/snow/engine/snowman/bootstrap/storage.go
@@ -170,19 +170,22 @@ func execute(
 	defer func() {
 		iterator.Release()
 
-		log("compacting database after executing blocks...")
-		if err := db.Compact(nil, nil); err != nil {
-			// Not a fatal error, log and move on.
-			log("failed to compact bootstrap database after executing blocks",
-				zap.Error(err),
-			)
+		halted := haltable.Halted()
+		if !halted {
+			log("compacting database after executing blocks...")
+			if err := db.Compact(nil, nil); err != nil {
+				// Not a fatal error, log and move on.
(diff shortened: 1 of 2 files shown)


snow/engine/snowman/bootstrap/storage.go
```diff
@@ -208,7 +226,7 @@ func execute(
 
 			processedSinceIteratorRelease = 0
 			iterator.Release()
-			iterator = interval.GetBlockIterator(db)
+			iterator = interval.GetBlockIteratorWithStart(db, height)
```
patrick-ogrady, line 229: nit: leave a comment that doing this is a very explicit decision.

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -167,6 +169,22 @@ func execute(
 	)
 	defer func() {
 		iterator.Release()
+
+		log("compacting database after executing blocks...")
+		if err := db.Compact(nil, nil); err != nil {
```
patrick-ogrady, line 174: Do we want this to run if we are exiting to shutdown?

abi87 (approved):

patrick-ogrady (approved):

PR #3048, opened 2024-05-23
add a metrics gauge for built block slot

tsachiherman:
## Why this should be merged

This PR adds a guage to the proposerVM, enabling the reporting of the slot at which the proposed block was built for.

## How this works

The change is quite straight forward. In `SetPreference` and `verifyPostDurangoBlockDelay`, report the selected slot.

In addition, in `postForkCommonComponents` we're keeping track of the selected slot in `verifyPostDurangoBlockDelay`, and during `postForkOption`'s  `Accept`, we report the accepted slot to the metric's histogram maintained by the `VM`.

## How this was tested
The testing was done manually by running a local node against Fuji, and monitoring the accumulated slots.


vms/proposervm/vm.go
```diff
@@ -727,3 +768,9 @@ func (vm *VM) cacheInnerBlock(outerBlkID ids.ID, innerBlk snowman.Block) {
 		vm.innerBlkCache.Put(outerBlkID, innerBlk)
 	}
 }
+
+// writeAcceptedSlotMetrics use the previously stored slot index and add that to the
+// metrics.
+func (vm *VM) writeAcceptedSlotMetrics(slot uint64) {
+	vm.acceptedBlocksSlotHistogram.Observe(float64(slot))
+}
```
StephenButtolph, line 776: Since this is only called in one place (and it's a one-liner), it feels like this could just be in-lined.

vms/proposervm/block.go
```diff
@@ -366,15 +375,17 @@ func (p *postForkCommonComponents) verifyPostDurangoBlockDelay(
 	var (
 		blkTimestamp = blk.Timestamp()
 		blkHeight    = blk.Height()
-		currentSlot  = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
 		proposerID   = blk.Proposer()
 	)
+	// populate the slot for the block.
+	p.slot = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
```
StephenButtolph, line 381: `p *postForkCommonComponents` is actually the parent block here - I think we should probably move the `slot` definition to the `postForkBlock` type. Then we'd have:
```suggestion
	// populate the slot for the block.
	blk.slot = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
```

vms/proposervm/vm.go
```diff
@@ -220,6 +227,27 @@ func (vm *VM) Initialize(
...
+
+	if err = registerer.Register(vm.proposerBuildSlotGauge); err != nil {
+		return err
+	}
+
+	vm.acceptedBlocksSlotHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "accepted_blocks_slot_histogram",
+			Help:    "the post-durango slot in which the block was accepted at",
+			Buckets: []float64{1.0, 2.0, 3.0},
```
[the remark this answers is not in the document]
tsachiherman: I don't think its more "clear", but it would definitely work around the floating point precision issue with comparing elements.

vms/proposervm/block.go
tsachiherman: thanks!

vms/proposervm/vm.go
```diff
@@ -305,14 +341,16 @@ func (vm *VM) SetPreference(ctx context.Context, preferred ids.ID) error {
...
 		currentTime := vm.Clock.Time().Truncate(time.Second)
+		proposalSlot = proposer.TimeToSlot(parentTimestamp, currentTime)
 		nextStartTime, err = vm.getPostDurangoSlotTime(
 			ctx,
 			childBlockHeight,
 			pChainHeight,
-			proposer.TimeToSlot(parentTimestamp, currentTime),
+			proposalSlot,
 			parentTimestamp,
 		)
```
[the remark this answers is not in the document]
tsachiherman: The reason I updated that below is that I wanted to keep the `err` checking before adding this.

vms/proposervm/vm.go
```diff
@@ -334,6 +372,13 @@ func (vm *VM) SetPreference(ctx context.Context, preferred ids.ID) error {
 		// until the P-chain's height has advanced.
 		return nil
 	}
+
+	// if we're done bootstrapping, and we have a proposal window slot, update
+	// the metrics accordingly.
+	if (proposalSlot != unspecifiedSlotIndex) && (vm.consensusState == snow.NormalOp) {
```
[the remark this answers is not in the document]
tsachiherman: I wasn't really sure if we want to have this added during the initial synchronization. I think that the block times *should* be ok, but wasn't sure if it's really needed or not. Definitely open to remove this if not needed.

vms/proposervm/vm.go
```diff
@@ -305,14 +341,16 @@ func (vm *VM) SetPreference(ctx context.Context, preferred ids.ID) error {
 		childBlockHeight = blk.Height() + 1
 		parentTimestamp  = blk.Timestamp()
 		nextStartTime    time.Time
+		proposalSlot     = unspecifiedSlotIndex
 	)
 	if vm.IsDurangoActivated(parentTimestamp) {
 		currentTime := vm.Clock.Time().Truncate(time.Second)
+		proposalSlot = proposer.TimeToSlot(parentTimestamp, currentTime)
```
StephenButtolph, line 348: This is the "first proposal slot to start looking for which proposal slot will be ours." Not our proposal slot. I think we actually want to set the metric to `proposer.TimeToSlot(parentTimestamp, nextStartTime)`. Right?

vms/proposervm/vm.go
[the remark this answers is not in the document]
StephenButtolph: `SetPreference` is only called immediately before consensus starts or after consensus has started, so this code won't be hit during the initial synchronization anyways

vms/proposervm/block.go
```diff
@@ -452,6 +460,11 @@ func (p *postForkCommonComponents) shouldBuildSignedBlockPostDurango(
 		)
 		return false, err
 	}
+
+	// report the build slot to the metrics.
+	p.vm.proposerBuildSlotGauge.Set(float64(currentSlot + 1))
```
StephenButtolph, line 465: Shouldn't this be:
```suggestion
	p.vm.proposerBuildSlotGauge.Set(float64(proposer.TimeToSlot(parentTimestamp, nextStartTime)))
```
?

vms/proposervm/block.go
```diff
@@ -24,6 +24,11 @@ import (
 const (
 	// allowable block issuance in the future
 	maxSkew = 10 * time.Second
+
+	// unspecifiedSlotIndex defines the default value being used for a proposal slot index, prior of being populated
+	// by the actual slot index. Given that the slot index is zero base, we would want to use an "invalid" value as
+	// non-zero.
+	unspecifiedSlotIndex = ^uint64(0)
```
StephenButtolph, line 31: Optional: Might be a bit more clear on what this value is (and why we never expect a slot to be equal to this)
```suggestion
	unspecifiedSlotIndex = math.MaxUint64
```

vms/proposervm/vm.go
```diff
@@ -99,6 +99,13 @@ type VM struct {
 
 	// lastAcceptedHeight is set to the last accepted PostForkBlock's height.
 	lastAcceptedHeight uint64
+
+	// proposerBuildSlotGauge is the metric gauge used when reporting the current slot block build built.
```
StephenButtolph, line 103:
```suggestion
	// proposerBuildSlotGauge reports the slot index when this node may attempt
	// to build a block.
```

vms/proposervm/vm.go
```diff
@@ -99,6 +99,13 @@ type VM struct {
...
+	proposerBuildSlotGauge prometheus.Gauge
+
+	// acceptedBlocksSlotHistogram is the metric histogram that is being updated once a block gets accepted with it's corresponding slot.
+	// note that this histogram is being updated by postForkCommonComponents.writeAcceptedSlotMetrics.
```
StephenButtolph, line 107:
```suggestion
	// acceptedBlocksSlotHistogram reports the slots that accepted blocks were
	// proposed in.
```

vms/proposervm/vm.go
```diff
@@ -220,6 +227,34 @@ func (vm *VM) Initialize(
...
+			Help: "the post-durango slot in which the block was built",
+		})
+
+	if err = registerer.Register(vm.proposerBuildSlotGauge); err != nil {
+		return err
+	}
+
+	vm.acceptedBlocksSlotHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "accepted_blocks_slot_histogram",
```
StephenButtolph, line 243: I think the type of the metric already implies that it is a histogram.
```suggestion
			Name: "accepted_blocks_slot",
```

vms/proposervm/vm.go
```diff
@@ -220,6 +227,34 @@ func (vm *VM) Initialize(
 	default:
 		return err
 	}
+
+	vm.proposerBuildSlotGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "block_building_slot",
+			Help: "the post-durango slot in which the block was built",
```
StephenButtolph, line 234:
```suggestion
			Help: "the slot that this node may attempt to build a block",
```

vms/proposervm/vm.go
```diff
@@ -220,6 +227,34 @@ func (vm *VM) Initialize(
...
+		})
+
+	if err = registerer.Register(vm.proposerBuildSlotGauge); err != nil {
+		return err
+	}
+
+	vm.acceptedBlocksSlotHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "accepted_blocks_slot_histogram",
+			Help: "the post-durango slot in which the block was accepted at",
```
StephenButtolph, line 244:
```suggestion
			Help: "the slot accepted blocks were proposed in",
```

vms/proposervm/vm.go
```diff
@@ -220,6 +227,34 @@ func (vm *VM) Initialize(
...
+			// (2, inf)
+			// the usage of ".5" before was to ensure we work around the limitation
+			// of comparing floating point of the same numerical value.
+			Buckets: []float64{0.5, 1.5, 2.5},
+		},
+	)
+	if err = registerer.Register(vm.acceptedBlocksSlotHistogram); err != nil {
+		return err
+	}
 	return nil
```
StephenButtolph, line 258: We typically use `utils.Err` for checks like this to improve readability.
```suggestion
	vm.proposerBuildSlotGauge = prometheus.NewGauge(prometheus.GaugeOpts{
		Name: "block_building_slot",
		Help: "the slot that this node may attempt to build a block",
	})
	vm.acceptedBlocksSlotHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
		Name: "accepted_blocks_slot",
		Help: "the slot accepted blocks were proposed in",
		// define the following ranges:
		// (-inf, 0]
		// (0, 1]
		// (1, 2]
		// (2, inf)
		// the usage of ".5" before was to ensure we work around the limitation
		// of comparing floating point of the same numerical value.
		Buckets: []float64{0.5, 1.5, 2.5},
	})
	return utils.Err(
		registerer.Register(vm.proposerBuildSlotGauge),
		registerer.Register(vm.acceptedBlocksSlotHistogram),
	)
```

vms/proposervm/block.go
```diff
@@ -24,6 +24,11 @@ import (
 const (
 	// allowable block issuance in the future
 	maxSkew = 10 * time.Second
+
+	// unspecifiedSlotIndex defines the default value being used for a proposal slot index, prior of being populated
+	// by the actual slot index. Given that the slot index is zero base, we would want to use an "invalid" value as
+	// non-zero.
```
StephenButtolph, line 30: Optional: I think this reads a bit better.
```suggestion
	// unspecifiedSlotIndex is the default value used for a proposal slot index,
	// prior of being populated by the actual slot index.
	// Because the slot index is zero based, we use a non-zero value to signal
	// that the value hasn't been populated.
```

vms/proposervm/post_fork_block.go
```diff
@@ -17,6 +17,10 @@ var _ PostForkBlock = (*postForkBlock)(nil)
 type postForkBlock struct {
 	block.SignedBlock
 	postForkCommonComponents
+
+	// slot is used to store the selected slot for this block. It's being used during Accept, and populated
+	// during verifyPostDurangoBlockDelay.
```
StephenButtolph, line 22: Optional: I think this reads a bit better.
```suggestion
	// slot of the proposer that produced this block.
	// It is populated in verifyPostDurangoBlockDelay.
	// It is used to report metrics during Accept.
```

StephenButtolph (commented):
I put a number of comments/wording/style suggestions. Once we align on those I think this is ready to merge.

vms/proposervm/vm.go
StephenButtolph: To clarify, everything is post-durango at this point. So it probably isn't meaningful to specify it in ongoing metrics.

vms/proposervm/vm.go
tsachiherman: that one is really nice. I haven't seen that design pattern used before.. and I like it.

StephenButtolph (approved):

vms/proposervm/block.go
```diff
@@ -270,6 +277,7 @@ func (p *postForkCommonComponents) buildChild(
 			innerBlk: innerBlock,
 			status:   choices.Processing,
 		},
+		slot: unspecifiedSlotIndex,
```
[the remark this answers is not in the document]
darioush: alternatively, we could make the metric report a 1 based number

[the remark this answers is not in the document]
tsachiherman: The idea of using a `*uint64` is an interesting one. it eliminate the need of initializing this field, and having the same downsides as the current comparison. I'll add that.

vms/proposervm/block.go
```diff
@@ -366,15 +366,18 @@ func (p *postForkCommonComponents) verifyPostDurangoBlockDelay(
 	var (
 		blkTimestamp = blk.Timestamp()
 		blkHeight    = blk.Height()
-		currentSlot  = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
 		proposerID   = blk.Proposer()
 	)
+	// populate the slot for the block.
+	blk.slot = new(uint64)
+	*blk.slot = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
```
StephenButtolph, line 373: I feel like this would be cleaner:
```golang
var (
		blkTimestamp = blk.Timestamp()
		blkHeight    = blk.Height()
		currentSlot  = proposer.TimeToSlot(parentTimestamp, blkTimestamp)
		proposerID   = blk.Proposer()
	)
	// populate the slot for the block.
	blk.slot = &currentSlot
```
And then we could just use `currentSlot` below like before.

PR #3049, opened 2024-05-23
Expose canonical warp formatting function

StephenButtolph:
## Why this should be merged

This exposes the canonical ordering of validators without needing to re-implement the `GetCanonicalValidatorSet` function.

## How this works

Exposes a helper function.

## How this was tested

- [X] CI

### DIFF
--- a/vms/platformvm/warp/validator.go
+++ b/vms/platformvm/warp/validator.go
@@ -58,9 +58,17 @@ func GetCanonicalValidatorSet(
 		return nil, 0, err
 	}
 
+	// Convert the validator set into the canonical ordering.
+	return FlattenValidatorSet(vdrSet)
+}
+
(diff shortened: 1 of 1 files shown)


dhrubabasu (approved):

darioush (approved):

abi87 (approved):

tsachiherman (approved):
The change looks fine, however, exposing this endpoint might suggest that if a new validator decide to use another's public key, one of the validators would "disappear" from the output ( randomly ).

StephenButtolph:
> The change looks fine, however, exposing this endpoint might suggest that if a new validator decide to use another's public key, one of the validators would "disappear" from the output ( randomly ).

The BLS keys are de-duplicated (and the weights are aggregated). So the validators don't "disappear".. They are merged. (notice that the returned `Validator` struct has a `[]ids.NodeID` - not a single `ids.NodeID`.

PR #3050, opened 2024-05-24
[antithesis] Enable longer vm init handshake duration

marun:
## Why this should be merged

Antithesis restarts nodes in a resource-constrained environment such that the duration of VM init can exceed the default handshake timeout of 5s. Enabling this timeout to be configured to a longer duration may increase the reliability of VM init in an antithesis test run. 

## How this works

- add new flag (`--rpcchainvm-handshake-timeout`) and update rpcchainvm manager to use it 
- configure a longer duration (30s) for antithesis testing 

## How this was tested

CI will ensure against regression but only an antithesis test run can verify that the longer timeout solves the problem.

### DIFF
--- a/vms/platformvm/warp/validator.go
+++ b/vms/platformvm/warp/validator.go
@@ -58,9 +58,17 @@ func GetCanonicalValidatorSet(
 		return nil, 0, err
 	}
 
+	// Convert the validator set into the canonical ordering.
+	return FlattenValidatorSet(vdrSet)
+}
+
+// FlattenValidatorSet converts the provided [vdrSet] into a canonical ordering.
+// Also returns the total weight of the validator set.
+func FlattenValidatorSet(vdrSet map[ids.NodeID]*validators.GetValidatorOutput) ([]*Validator, uint64, error) {
 	var (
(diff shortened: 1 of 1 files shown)


config/flags.go
```diff
@@ -361,6 +361,9 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.StringToString(TracingHeadersKey, map[string]string{}, "The headers to provide the trace indexer")
 
 	fs.String(ProcessContextFileKey, defaultProcessContextPath, "The path to write process context to (including PID, API URI, and staking address).")
+
+	// RPC Chain VMs
+	fs.Duration(RPCVMHandshakeTimeoutKey, constants.DefaultRPCVMHandshakeTimeout, "Maximum duration to wait for an RPC VM to complete its initialization handshake.")
```
joshua-kim, line 366: nit: it feels weird that the vm is called the `rpcchainvm` but we're referencing it as `rpc vm` here. My guess is that you wanted to avoid this since the name is long/awkward... I feel like something like `rpcchainvm-handshake-timeout` or `vm-grpc-handshake-timeout` might make sense but I don't have anything concretely in mind...

joshua-kim (approved):

PR #3052, opened 2024-05-24
Remove optional gatherer

StephenButtolph:
## Why this should be merged

Removes one of the custom gatherers and simplifies the registration of metrics.

## How this works

Exposes `metrics.MultiGatherer` in `snow.Context` rather than `metrics.OptionalGatherer`.

## How this was tested

- [X] CI
- [X] Ran a Fuji node and verified that none of the metrics names changed.


api/metrics/multi_gatherer_test.go
```diff
@@ -4,13 +4,17 @@
...
+	"errors"
 	"testing"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 
 	dto "github.com/prometheus/client_model/go"
 )
 
+var errTest = errors.New("non-nil error")
```
tsachiherman, line 16: unused local variable.

tsachiherman (approved):
looks good. just one comment regarding newly local variable that's not in use.

api/metrics/multi_gatherer_test.go
StephenButtolph: Hmmm - this variable is used here: https://github.com/ava-labs/avalanchego/blob/1b4bb45e3b4aaba3f8fd7860b6835bd2de3fc5f8/api/metrics/multi_gatherer_test.go#L42-L56 But it probably makes sense to change this from a package var to a function var... I'll update that

dhrubabasu (approved):

Issue #1581, opened 2023-06-05
X-Chain FATAL on bootstrap


jpop32:
**Describe the bug**
When doing a clean bootstrap (deleted database, restarted node), P and C finished with no issues. X-Chain ended up with a FATAL error and exited. Relevant bit of log:
```
Jun 05 13:55:27 node-main avalanchego[2309622]: [06-05|13:55:27.297] INFO <X Chain> queue/jobs.go:203 executing operations {"numExecuted": 4092719, "numToExecute": 4098651, "eta": "7s"}
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.402] INFO <X Chain> queue/jobs.go:224 executed operations {"numExecuted": 4098651}
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.405] INFO <X Chain> bootstrap/bootstrapper.go:115 starting bootstrapper
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.428] INFO <X Chain> bootstrap/bootstrapper.go:575 executing transactions
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.428] INFO <X Chain> queue/jobs.go:224 executed operations {"numExecuted": 0}
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.428] INFO <X Chain> bootstrap/bootstrapper.go:592 executing vertices
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.428] INFO <X Chain> queue/jobs.go:224 executed operations {"numExecuted": 0}
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.428] FATAL <X Chain> handler/handler.go:344 shutting down chain {"reason": "received an unexpected error", "error": "duplicated register while processing sync message: {%!s(*message.inboundMessage=&{[129 130 201 232 153 92 130 67 7 139 58 175 15 206 144 51 209 180 85 29] 19 0xc10043d6e0 {0 9223372036854775807 0x34eabe0} 0xc15dc0 10198}) ENGINE_TYPE_AVALANCHE}"}
Jun 05 13:55:35 node-main avalanchego[2309622]: [06-05|13:55:35.429] INFO <X Chain> bootstrap/bootstrapper.go:297 shutting down bootstrapper
[...]

PR #3053, opened 2024-05-28
Replace all chain namespaces with labels

StephenButtolph:
## Why this should be merged

Avalanchego exposes numerous metrics of the format:
```
avalanche_C_...
avalanche_P_...
avalanche_X_...
```

However, many of these metrics are standard across all chains. For example:
```
avalanche_C_blks_processing
avalanche_P_blks_processing
avalanche_X_blks_processing
```

This is not how metrics are typically reported. And this can introduce significant complexities into handling these metrics. A more standard mechanism, which is implemented by this PR, would be:
```
avalanche_snowman_blks_processing{chain:"C"}
avalanche_snowman_blks_processing{chain:"P"}
avalanche_snowman_blks_processing{chain:"X"}
```

1. This will reduce the cardinality of metrics exposed by avalanchego.
2. This will significantly reduce the total cardinality of metrics for subnet services which sync many chains. 

## How this works

1. Uses `metrics.labelGatherer`s to populate chain labels.
2. Removes custom `snow.Context#Metrics` overrides by providing a registry for wrapping VMs to register metrics.
3. Changes various metrics namespaces.

## How this was tested

- [X] CI
- [X] Running a Fuji node

See differences in metrics outputs:
[old.txt](https://github.com/user-attachments/files/15598624/old.txt)
[new.txt](https://github.com/user-attachments/files/15598625/new.txt)


api/metrics/label_gatherer.go
```diff
@@ -0,0 +1,93 @@
...
+				continue
+			}
+
+			gatheredMetric.Label = append(gatheredMetric.Label, &dto.LabelPair{
+				Name:  &g.labelName,
+				Value: &g.labelValue,
+			})
+			slices.SortFunc(gatheredMetric.Label, func(i, j *dto.LabelPair) int {
+				return cmp.Compare(i.GetName(), j.GetName())
+			})
```
StephenButtolph, line 87: Technically we could use `slices.Insert` here after binary searching into the labels... I think this is fine.

api/metrics/multi_gatherer.go
```diff
@@ -4,92 +4,48 @@
...
+	Register(name string, gatherer prometheus.Gatherer) error
 }
 
-type multiGatherer struct {
-	lock      sync.RWMutex
-	gatherers map[string]prometheus.Gatherer
+// Deprecated: Use NewPrefixGatherer instead.
+//
+// TODO: Remove once coreth is updated.
+func NewMultiGatherer() MultiGatherer {
```
StephenButtolph, line 28: This is currently used by coreth... Figured keeping it around for now was easier than updating the dependency for now.

api/metrics/prefix_gatherer.go
```diff
@@ -0,0 +1,96 @@
...
+
+	for _, existingPrefix := range g.names {
+		if eitherIsPrefix(prefix, existingPrefix) {
+			return fmt.Errorf("%w: %q conflicts with %q",
+				errOverlappingNamespaces,
+				prefix,
+				existingPrefix,
+			)
+		}
+	}
```
StephenButtolph, line 45: This is a change from the prior code.

Previously we explicitly allowed potential conflicts for metrics backwards compatibility. However, this change explicitly doesn't maintain backwards compatibility... So, I felt like it was a good time to properly restrict this. 

Previously, it would have been possible to register things like `""` and `"vm"`. If the `""` registerer added a metric that started with `"vm"`, there would have been a conflict. Now, `eitherIsPrefix` checks for this case and disallows the registration.

chains/linearizable_vm.go
```diff
@@ -42,7 +40,6 @@ type initializeOnLinearizeVM struct {
 
 func (vm *initializeOnLinearizeVM) Linearize(ctx context.Context, stopVertexID ids.ID) error {
 	vm.vmToLinearize.stopVertexID = stopVertexID
-	vm.ctx.Metrics = vm.registerer
```
StephenButtolph, line 45: This likely resolves #1581.

chains/manager.go
```diff
@@ -1385,3 +1557,43 @@ func (m *manager) getChainConfig(id ids.ID) (ChainConfig, error) {
...
+			vmIDStr = "avm"
+		case constants.CorethID:
+			vmIDStr = "coreth"
+		case constants.SubnetEVMID:
+			vmIDStr = "subnetevm"
+		case xsvm.ID:
+			vmIDStr = "xsvm"
+		default:
+			vmIDStr = vmID.String()
+		}
```
StephenButtolph, line 1581: Are we okay with this for now? Or should we try to clean this up a bit?

snow/networking/timeout/manager.go
```diff
@@ -71,27 +71,33 @@ type Manager interface {
 func NewManager(
 	timeoutConfig *timer.AdaptiveTimeoutConfig,
 	benchlistMgr benchlist.Manager,
-	metricsNamespace string,
-	metricsRegister prometheus.Registerer,
+	requestReg prometheus.Registerer,
+	responseReg prometheus.Registerer,
```
StephenButtolph, line 75: Should we try to merge these namespaces?

utils/constants/vm_ids.go
```diff
@@ -8,5 +8,6 @@ import "github.com/ava-labs/avalanchego/ids"
 var (
 	PlatformVMID = ids.ID{'p', 'l', 'a', 't', 'f', 'o', 'r', 'm', 'v', 'm'}
 	AVMID        = ids.ID{'a', 'v', 'm'}
-	EVMID        = ids.ID{'e', 'v', 'm'}
+	CorethID     = ids.ID{'e', 'v', 'm'}
+	SubnetEVMID  = ids.ID{'s', 'u', 'b', 'n', 'e', 't', 'e', 'v', 'm'}
```
StephenButtolph, line 12: This was added so that subnet-evm metrics don't have the vmID in them. I'm okay with removing it... But it seemed nice to have... I'm not willing to import subnet-evm to grab this ID directly though.

vms/proposervm/vm.go
```diff
@@ -130,21 +128,9 @@ func (vm *VM) Initialize(
 	fxs []*common.Fx,
 	appSender common.AppSender,
 ) error {
-	// TODO: Add a helper for this metrics override, it is performed in multiple
-	//       places.
```
StephenButtolph, line 134: We never do this anymore 🥳

vms/rpcchainvm/vm_server.go
```diff
@@ -567,22 +586,8 @@ func (vm *VMServer) AppGossip(ctx context.Context, req *vmpb.AppGossipMsg) (*emp
...
-		return nil, err
-	}
-
-	// Gather metrics registered by rpcchainvm server Gatherer. These
-	// metrics are collected for each Go plugin process.
-	pluginMetrics, err := vm.processMetrics.Gather()
-	if err != nil {
-		return nil, err
-	}
-	mfs = append(mfs, pluginMetrics...)
```
StephenButtolph, line 583: This was just wrong. This isn't the correct way to merge metric families.

snow/networking/timeout/manager.go
StephenButtolph: I mainly just found it weird that we now have to pass in multiple registerers... The implications could be either changing some metrics or changing where metrics are specified... I didn't make the change here because I wasn't very happy with the alternatives

dhrubabasu (approved):

PR #3054, opened 2024-05-28
[vms/platformvm] Return the correct owner in `platform.GetSubnets` after transfer

### LINKED ISSUE #3012: API returns wrong control keys data after subnet ownership transfer
**Describe the bug**
When subnet ownership has been transferred over, `platform.getSubnets()` API call shows the old data unless selecting for the subent specifically using the `ids` param.

**To Reproduce**
Curl-ing with a payload of:
```json
{
    "jsonrpc": "2.0",
    "method": "platform.getSubnets",
    "params": {},
    "id": 1
}
```
gives the list of all the subnets, among others it has:
```json
            {
                "id": "HDHyjxQi3fnMDmuEGZzEx33rbGe6Yz1dz2WGgEit6jEuTbxdj",
                "controlKeys": [
                    "P-fuji1ywzvrftfqexh5g6qa9zyrytj6pqdfetza2hqln"
                ],
                "threshold": "1"
            },
```
But calling with
```json
{
    "jsonrpc": "2.0",
    "method": "platform.getSubnets",
    "params": {"ids":["HDHyjxQi3fnMDmuEGZzEx33rbGe6Yz1dz2WGgEit6jEuTbxdj"]},
    "id": 1
}
```
returns the correct control key:
```json
            {
                "id": "HDHyjxQi3fnMDmuEGZzEx33rbGe6Yz1dz2WGgEit6jEuTbxdj",
                "controlKeys": [
                    "P-fuji1mlhaxu92c5p57ce56trsc7eha6cflh20cfy4gn"
                ],
                "threshold": "1"
            }
```
**Expected behavior**
The API call should return the correct data in both cases.

**Operating System**
Node running on Ubuntu. Version info:
```json
{
    "jsonrpc": "2.0",
    "result": {
        "version": "avalanchego/1.11.4",
        "databaseVersion": "v1.4.5",
        "rpcProtocolVersion": "35",
        "gitCommit": "e8904ae11aa144ae601bb9795682d11c3fb178ee",
        "vmVersions": {
            "avm": "v1.11.4",
            "evm": "v0.13.2",
            "platform": "v1.11.4"
        }
    },
    "id": 1
}
```

dhrubabasu:
## Why this should be merged

Fixes #3012 

## How this works

Uses `GetSubnetOwner` like we do when a single subnetID is passed in

## How this was tested

Added a UT that fails on master

### DIFF
--- a/vms/platformvm/service.go
+++ b/vms/platformvm/service.go
@@ -545,9 +545,6 @@ func (s *Service) GetSubnets(_ *http.Request, args *GetSubnetsArgs, response *Ge
 			}
 
 			subnetOwner, err := s.vm.state.GetSubnetOwner(subnetID)
-			if err == database.ErrNotFound {
-				continue
-			}
 			if err != nil {
 				return err
 			}
--- a/vms/platformvm/service.go
+++ b/vms/platformvm/service.go
@@ -544,15 +544,26 @@ func (s *Service) GetSubnets(_ *http.Request, args *GetSubnetsArgs, response *Ge
 				continue
 			}
 
-			unsignedTx := subnet.Unsigned.(*txs.CreateSubnetTx)
-			owner := unsignedTx.Owner.(*secp256k1fx.OutputOwners)
-			controlAddrs := []string{}
-			for _, controlKeyID := range owner.Addrs {
+			subnetOwner, err := s.vm.state.GetSubnetOwner(subnetID)
+			if err == database.ErrNotFound {
+				continue
+			}
+			if err != nil {
+				return err
+			}
+
+			owner, ok := subnetOwner.(*secp256k1fx.OutputOwners)
+			if !ok {
+				return fmt.Errorf("expected *secp256k1fx.OutputOwners but got %T", subnetOwner)
+			}
+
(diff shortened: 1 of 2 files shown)


vms/platformvm/service.go
```diff
@@ -544,15 +544,26 @@ func (s *Service) GetSubnets(_ *http.Request, args *GetSubnetsArgs, response *Ge
 				continue
 			}
 
-			unsignedTx := subnet.Unsigned.(*txs.CreateSubnetTx)
-			owner := unsignedTx.Owner.(*secp256k1fx.OutputOwners)
-			controlAddrs := []string{}
-			for _, controlKeyID := range owner.Addrs {
+			subnetOwner, err := s.vm.state.GetSubnetOwner(subnetID)
+			if err == database.ErrNotFound {
```
tsachiherman, line 548: That would leave "holes" in the "response.Subnets". Was that your intent ?

dhrubabasu: We shouldn't be leaving "holes" here since we're returning all the subnets. Removed the `database.ErrNotFound` check, good catch!

abi87 (approved):

PR #3055, opened 2024-05-28
[vms/platformvm] Replace `GetSubnets` with `GetSubnetIDs` in `State`

dhrubabasu:
## Why this should be merged

We don't use the full subnet txs in most places. Let's reduce the amount we store in memory 🧹

## How this works

Modifies `GetSubnets` in the `State` struct to be `GetSubnetIDs`

## How this was tested

CI

### DIFF
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -104,7 +104,7 @@ type Chain interface {
 
 	AddRewardUTXO(txID ids.ID, utxo *avax.UTXO)
 
-	AddSubnet(createSubnetTx *txs.Tx)
+	AddSubnet(createSubnetTxID ids.ID)
 
 	GetSubnetOwner(subnetID ids.ID) (fx.Owner, error)
 	SetSubnetOwner(subnetID ids.ID, owner fx.Owner)
@@ -753,9 +753,7 @@ func (s *state) GetSubnetIDs() ([]ids.ID, error) {
 	return subnetIDs, nil
 }
 
-func (s *state) AddSubnet(createSubnetTx *txs.Tx) {
-	createSubnetTxID := createSubnetTx.ID()
-
+func (s *state) AddSubnet(createSubnetTxID ids.ID) {
 	s.addedSubnetIDs = append(s.addedSubnetIDs, createSubnetTxID)
 	if s.cachedSubnetIDs != nil {
 		s.cachedSubnetIDs = append(s.cachedSubnetIDs, createSubnetTxID)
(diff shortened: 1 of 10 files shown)


vms/platformvm/state/state.go
```diff
@@ -728,39 +728,39 @@ func (s *state) doneInit() error {
...
 
-	txs := []*txs.Tx(nil)
+	subnetIDs := []ids.ID{}
 	for subnetDBIt.Next() {
 		subnetIDBytes := subnetDBIt.Key()
 		subnetID, err := ids.ToID(subnetIDBytes)
 		if err != nil {
 			return nil, err
 		}
-		subnetTx, _, err := s.GetTx(subnetID)
```
dhrubabasu, line 746: Removing this `GetTx` call from every invocation is the main benefit of this PR

JacobEverly (approved):

vms/platformvm/vm_test.go
```diff
@@ -958,12 +958,12 @@ func TestCreateSubnet(t *testing.T) {
...
 
-	subnets, err := vm.state.GetSubnets()
+	subnetIDs, err := vm.state.GetSubnetIDs()
 	require.NoError(err)
 
 	found := false
-	for _, subnet := range subnets {
-		if subnet.ID() == createSubnetTx.ID() {
+	for _, subnetID := range subnetIDs {
+		if subnetID == createSubnetTx.ID() {
```
StephenButtolph, line 966: I think this can be:
```golang
require.Contains(subnetIDs, createSubnetTx.ID())
```

vms/platformvm/state/state.go
```diff
@@ -728,39 +728,37 @@ func (s *state) doneInit() error {
...
 	}
-	txs = append(txs, s.addedSubnets...)
-	s.cachedSubnets = txs
-	return txs, nil
+	subnetIDs = append(subnetIDs, s.addedSubnetIDs...)
+	s.cachedSubnetIDs = subnetIDs
+	return subnetIDs, nil
 }
 
 func (s *state) AddSubnet(createSubnetTx *txs.Tx) {
```
StephenButtolph, line 756: Should this take in just the ID?

vms/platformvm/state/state.go
```diff
@@ -728,39 +728,37 @@ func (s *state) doneInit() error {
...
+func (s *state) GetSubnetIDs() ([]ids.ID, error) {
+	if s.cachedSubnetIDs != nil {
+		return s.cachedSubnetIDs, nil
 	}
 
 	subnetDBIt := s.subnetDB.NewIterator()
 	defer subnetDBIt.Release()
 
-	txs := []*txs.Tx(nil)
+	subnetIDs := []ids.ID{}
```
StephenButtolph, line 739: This is a super small change in behavior, but I think it's a good change. In the case that there are no subnets (which isn't realistic in most cases outside of testing), previously we wouldn't cache the lookup. Now we will cache the lookup as there being no subnets.

StephenButtolph (approved):

PR #3057, opened 2024-05-29
Add metrics client

StephenButtolph:
## Why this should be merged

Currently our metrics parsing in the e2e tests is ad-hoc and incomplete. This implements a metrics client for anyone to use and uses it in the e2e tests.

## How this works

^

## How this was tested

- [x] CI

### DIFF
--- a/api/metrics/client.go
+++ b/api/metrics/client.go
@@ -1,0 +1,68 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package metrics
+
+import (
+	"bytes"
+	"context"
+	"fmt"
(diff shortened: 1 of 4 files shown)


api/metrics/client.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/prometheus/common/expfmt"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Client for requesting metrics from a remote AvalancheGo instance Info API Client
```
dhrubabasu, line 18:
```suggestion
// Client for requesting metrics from a remote AvalancheGo instance
```

dhrubabasu (approved):

PR #3058, opened 2024-05-30
Implement `constants.VMName`

StephenButtolph:
## Why this should be merged

This is factored out of #3053. This function will be used to improve the readability of VM specific metrics.

## How this works

Implements a map from `vmID` to `vmName`.

## How this was tested

N/A

### DIFF
--- a/utils/constants/vm_ids.go
+++ b/utils/constants/vm_ids.go
@@ -5,8 +5,37 @@ package constants
 
 import "github.com/ava-labs/avalanchego/ids"
 
+const (
+	PlatformVMName = "platformvm"
+	AVMName        = "avm"
+	EVMName        = "evm"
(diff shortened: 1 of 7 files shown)


dhrubabasu (approved):

utils/constants/vm_ids.go
```diff
@@ -5,8 +5,37 @@ package constants
 
 import "github.com/ava-labs/avalanchego/ids"
 
+const (
+	PlatformVMName = "platformvm"
+	AVMName        = "avm"
+	EVMName        = "evm"
+	SubnetEVMName  = "subnetevm"
+	XSVMName       = "xsvm"
```
abi87, line 13: not sure I like having non-Primary-Network VMs here, but I guess there is no alternative?

abi87 (approved):

PR #3059, opened 2024-05-30
[antithesis] Enable workload instrumentation

marun:
## Why this should be merged

The Antithesis test setups currently use regex matches on log output to trigger fault injection (by matching on output indicating that test setup was completed) and record desired test properties (e.g. by matching on output indicating that an X-Chain transaction was verified). The configuration of this matching is currently stored on Antithesis infrastructure and only accessible to Antithesis personal.

This changeset uses the Antithesis SDK to explicitly signal events so that the use of potentially unreliable regex matching can be avoided and so that the workload for a test setup can have more control over test execution and analysis. The use of the SDK's assertions by the workload requires the workload binary to be optionally instrumented as per the example of the node binary.

## How this works

- Enable optional instrumentation of the workload images. This is required for the workloads to use assertions supported by the Antithesis SDK. 
- Update the workloads to emit lifecycle events and add assertions

## How this was tested

- CI already includes a job that validates antithesis image build and execution


abi87 (approved):
lgtm

StephenButtolph (commented):
Is there a reason why we need to update the docker images in this PR to support using the antithesis sdk in the workloads? Or are those separate changes

tests/antithesis/xsvm/main.go
```diff
@@ -157,12 +163,19 @@ func (w *workload) run(ctx context.Context) {
...
 
 func (w *workload) confirmTransferTx(ctx context.Context, tx *status.TxIssuance) {
+	confirmed := true
 	for _, uri := range w.uris {
 		client := api.NewClient(uri, w.chainID.String())
 		if err := api.WaitForAcceptance(ctx, client, w.key.Address(), tx.Nonce); err != nil {
 			log.Printf("worker %d failed to confirm transaction %s on %s: %s", w.id, tx.TxID, uri, err)
-			return
+			confirmed = false
+			break
```
StephenButtolph, line 172: Should we just `assert.Unreachable` here and `assert.Reachable` at the bottom and then remove the control flow change + `confirmed` bool?

tests/antithesis/xsvm/main.go
```diff
@@ -92,6 +96,8 @@ func main() {
 		}
 	}
 
+	lifecycle.SetupComplete(fmt.Sprintf("Initialized %d workers", NumKeys))
```
StephenButtolph, line 99: Seems like this supports structure, should this be
```suggestion
	lifecycle.SetupComplete(map[string]any{
	    "msg": "initialized workers"
	    "numWorkers": NumKeys,
	})
```

StephenButtolph:
I made a PR to add a number of additional assertions: https://github.com/ava-labs/avalanchego/pull/3064

StephenButtolph (approved):

PR #3063, opened 2024-05-30
[testing] Remove superfluous gomega dep

marun:
Our only use of gomega as a dependency is to register a fail handler for it. Turns out that does nothing at all.


darioush (approved):
nice

abi87 (approved):

PR #3064, opened 2024-05-30
Add more Antithesis assertions

StephenButtolph:
## Why this should be merged

This should provide full coverage for the tests if we were to remove the "failed" regex.

## How this works

Adds various assertions.

## How this was tested

N/A


tests/antithesis/avalanchego/main.go
```diff
@@ -220,6 +225,10 @@ func (w *workload) issueXChainBaseTx(ctx context.Context) {
 	balances, err := xBuilder.GetFTBalance()
 	if err != nil {
 		log.Printf("failed to fetch X-chain balances: %s", err)
+		assert.Unreachable("failed to fetch X-chain balances", map[string]any{
+			"worker": w.id,
+			"err":    err,
+		})
```
StephenButtolph, line 231: I was very close to making this a helper function... But I'm not sure how easily that would work with the library because the `Unreachable` call would then be in a single place. (and the library seems to care about the location of the function call)

PR #3065, opened 2024-05-30
Only compact after executing a large number of blocks

StephenButtolph:
## Why this should be merged

After a chain has finished bootstrapping (like the P-chain) it will re-bootstrap every 10 seconds to keep up-to-tip until the full subnet has finished bootstrapping.

Calling `db.Compact` every 10 seconds (even with a small range of impacted values on disk) significantly reduces the performance of other DB operations.

This can cause the X-chain and C-chain syncing to be significantly slower than before this compaction was added.

## How this works

Only compact the bootstrapping database if the number of blocks executed was significant.

## How this was tested

- [X] Bootstrapped a mainnet node on `v1.11.6` (C-chain state-sync and X-chain DAG fetching are still ongoing after >48 hours).
- [X] Bootstrapped a mainnet node with this change (C-chain state-sync and X-chain DAG fetching have finished in <24 hours).
- [X] CI

### DIFF
--- a/snow/engine/snowman/bootstrap/storage.go
+++ b/snow/engine/snowman/bootstrap/storage.go
@@ -135,7 +135,7 @@ func execute(
 	lastAcceptedHeight uint64,
 ) error {
 	totalNumberToProcess := tree.Len()
-	if totalNumberToProcess > minBlocksToCompact {
+	if totalNumberToProcess >= minBlocksToCompact {
 		log("compacting database before executing blocks...")
 		if err := db.Compact(nil, nil); err != nil {
 			// Not a fatal error, log and move on.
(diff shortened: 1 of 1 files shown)


tsachiherman (approved):
I don't see any inherent issue here, but I'm wondering why wouldn't we compact the database inline with the block processing, i.e. within `for !haltable.Halted() && iterator.Next() ` loop.

StephenButtolph:
> I'm wondering why wouldn't we compact the database inline with the block processing, i.e. within `for !haltable.Halted() && iterator.Next() ` loop.

We tested that and it was less performant than this code. (for executing a single chain).

darioush (approved):

PR #3066, opened 2024-05-30
Remove api namespace

StephenButtolph:
## Why this should be merged

Pulled out of https://github.com/ava-labs/avalanchego/pull/3053.

## How this works

Replaces the passing of a namespace around with registering an additional registerer into the multigatherer.

## How this was tested

- [X] CI
- [x] Verified that metrics were unchanged

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -66,6 +66,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/ips"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/math/meter"
+	"github.com/ava-labs/avalanchego/utils/metric"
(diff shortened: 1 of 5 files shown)


node/node.go
```diff
@@ -89,6 +90,8 @@ const (
 	httpPortName    = constants.AppName + "-http"
 
 	ipResolutionTimeout = 30 * time.Second
+
+	apiNamespace = constants.PlatformName + metric.NamespaceSeparator + "api"
```
darioush, line 94: here, I don't understand why we add `constants.PlatformName` to the metric.

darioush (approved):
otherwise, lgtm

node/node.go
StephenButtolph: all of the metrics exported by avalanchego start with the `avalanche_` prefix. This is where we're adding that prefix.

PR #3067, opened 2024-05-31
Remove network namespace

StephenButtolph:
## Why this should be merged

Pulled out of https://github.com/ava-labs/avalanchego/pull/3053.

## How this works

Replaces the passing of a namespace around with registering an additional registerer into the multigatherer.

## How this was tested

- [X] CI
- [ ] Verified that metrics were unchanged


network/throttling/bandwidth_throttler.go
```diff
@@ -69,16 +68,15 @@ func newBandwidthThrottler(
 		limiters:                 make(map[ids.NodeID]*rate.Limiter),
 		metrics: bandwidthThrottlerMetrics{
 			acquireLatency: metric.NewAveragerWithErrs(
-				namespace,
+				"",
```
darioush, line 71: Will this also be removed soon?

darioush (approved):

network/throttling/bandwidth_throttler.go
StephenButtolph: yes

PR #3073, opened 2024-05-31
Select metric by label in e2e tests

StephenButtolph:
## Why this should be merged

Factored out of #3053.

## How this works

This supports filtering metrics by labels. This will be needed after the metrics refactor.

## How this was tested

- [X] CI

### DIFF
--- a/tests/metrics.go
+++ b/tests/metrics.go
@@ -39,6 +39,13 @@ func GetNodesMetrics(ctx context.Context, nodeURIs []string) (NodesMetrics, erro
 	return metrics, nil
 }
 
+// GetMetricValue returns the value of the specified metric which has the
+// required labels.
+//
+// If multiple metrics match the provided labels, the first metric found is
+// returned.
+//
+// Only Counter and Gauge metrics are supported.
 func GetMetricValue(metrics NodeMetrics, name string, labels prometheus.Labels) (float64, bool) {
 	metricFamily, ok := metrics[name]
 	if !ok {
--- a/tests/metrics.go
+++ b/tests/metrics.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


tests/metrics.go
```diff
@@ -37,19 +39,45 @@ func GetNodesMetrics(ctx context.Context, nodeURIs []string) (NodesMetrics, erro
...
+	}
+	return 0, false
+}
+
+func labelsMatch(metric *dto.Metric, labels prometheus.Labels) bool {
+	var found int
+	for _, label := range metric.Label {
+		expectedValue, ok := labels[label.GetName()]
+		if !ok {
+			continue
```
tsachiherman, line 75: no point in `continue` here; you can just 
```golang
return false
```

tsachiherman: also, comparing `len(metric.Label) == len(labels)` before the loop could save you the need to track `found`.

StephenButtolph: This wasn't quite what I had intended for this function. Perhaps we should improve the documentation. If we have metrics like:
```
{a:"0", b:"1"}
{a:"2", b:"3"}
```
I intended to be able to call this function with `{a:"0"}` and have it return the first metric (even though the specified labels didn't mention anything about `b`).

StephenButtolph: We could also just expect the labels to be fully specified... 🤷

tsachiherman: ohh - that's create even bigger issue ;-(
the `prometheus.Labels` is a map, so iteration on that would not have a consistent ordering..

StephenButtolph: I'm not sure I follow. This function isn't iterating over a map.

tsachiherman: oh - sorry. yes, it was very confusing, but you're correct. it should be ok.

tsachiherman (approved):

PR #3084, opened 2024-06-04
Remove rejection from `consensus.Add`

StephenButtolph:
## Why this should be merged

Currently `consensus.Add` attempts to handle the case that a block is issued into consensus that should not have been (either because it was previously issued or because it's parent isn't known by the consensus instance).

This should never happen because the engine already checks for these cases:
1. This is where the engine ensures the block is not already processing or decided: https://github.com/ava-labs/avalanchego/blob/2cf7bd6a053029a808c6f0a6fd244dd50330f166/snow/engine/snowman/transitive.go#L956-L963
2. This is where the engine ensures that the parent block is either the last accepted block or is processing: https://github.com/ava-labs/avalanchego/blob/2cf7bd6a053029a808c6f0a6fd244dd50330f166/snow/engine/snowman/transitive.go#L965-L975
  a. Note that this does ensure that the parent block isn't an older accepted block because the initial check ensures that the block isn't accepted with the `blk.Height() <= ts.lastAcceptedHeight` check inside of `Decided`.
  b. Specifically, if `blk.Height() > ts.lastAcceptedHeight`, and `parent.Status() == choices.Accepted` we know that `parentID == lastAcceptedID`

## How this works

Simplifies block addition to consensus.

## How this was tested

- [X] Modified unit tests

### DIFF
--- a/snow/consensus/snowman/consensus_test.go
+++ b/snow/consensus/snowman/consensus_test.go
@@ -32,7 +32,7 @@ var (
 		NumProcessingTest,
 		AddToTailTest,
 		AddToNonTailTest,
-		AddToUnknownTest,
+		AddOnUnknownParentTest,
 		StatusOrProcessingPreviouslyAcceptedTest,
 		StatusOrProcessingPreviouslyRejectedTest,
 		StatusOrProcessingUnissuedTest,
@@ -51,7 +51,6 @@ var (
 		MetricsProcessingErrorTest,
 		MetricsAcceptedErrorTest,
 		MetricsRejectedErrorTest,
-		ErrorOnInitialRejectionTest,
 		ErrorOnAcceptTest,
 		ErrorOnRejectSiblingTest,
 		ErrorOnTransitiveRejectionTest,
@@ -139,7 +138,7 @@ func NumProcessingTest(t *testing.T, factory Factory) {
 	require.Zero(sm.NumProcessing())
 
 	// Adding to the previous preference will update the preference
-	require.NoError(sm.Add(context.Background(), block))
+	require.NoError(sm.Add(block))
 	require.Equal(1, sm.NumProcessing())
 
 	votes := bag.Of(block.ID())
@@ -176,7 +175,7 @@ func AddToTailTest(t *testing.T, factory Factory) {
 	block := snowmantest.BuildChild(snowmantest.Genesis)
 
 	// Adding to the previous preference will update the preference
-	require.NoError(sm.Add(context.Background(), block))
+	require.NoError(sm.Add(block))
 	require.Equal(block.ID(), sm.Preference())
 	require.True(sm.IsPreferred(block))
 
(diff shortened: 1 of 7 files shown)


snow/consensus/snowman/traced_consensus.go
```diff
@@ -29,16 +29,6 @@ func Trace(consensus Consensus, tracer trace.Tracer) Consensus {
...
 
-func (c *tracedConsensus) Add(ctx context.Context, blk Block) error {
-	ctx, span := c.tracer.Start(ctx, "tracedConsensus.Add", oteltrace.WithAttributes(
-		attribute.Stringer("blkID", blk.ID()),
-		attribute.Int64("height", int64(blk.Height())),
-	))
-	defer span.End()
-
-	return c.Consensus.Add(ctx, blk)
-}
```
aaronbuchwald, line 40: Why are we removing the context and tracing from `Add` here?

snow/consensus/snowman/consensus_test.go
```diff
@@ -260,11 +259,9 @@ func AddToUnknownTest(t *testing.T, factory Factory) {
 		HeightV: snowmantest.GenesisHeight + 2,
 	}
 
-	// Adding a block with an unknown parent means the parent must have already
-	// been rejected. Therefore the block should be immediately rejected
-	require.NoError(sm.Add(context.Background(), block))
-	require.Equal(snowmantest.GenesisID, sm.Preference())
-	require.Equal(choices.Rejected, block.Status())
```
aaronbuchwald, line 267: Why is the preference check removed? I assume the rejected check is removed because consensus now only returns an error, but does not mark the failed to add block as rejected.

aaronbuchwald (approved):

joshua-kim (approved):

snow/consensus/snowman/consensus_test.go
joshua-kim: I think we don't have to since we don't add to consensus anymore. Previously we were adding and immediately deciding rejected, now we're just dropping it from consensus entirely and not making a decision at all.

aaronbuchwald: Previously, this was testing that the preference stayed the same as it was previously (ie. preference is still genesis after adding the new block fails) though since it dropped and rejected the block in the same place.

I agree we don't need the check in the test though.

StephenButtolph: It just didn't seem like a useful check anymore. I can revert that if we feel like it was important to test.

snow/consensus/snowman/traced_consensus.go
StephenButtolph: The only reason `ctx` was added to this function was to pass it through into `Reject`. When we added tracing we only added tracing for methods that had `ctx` anyway... We could re-introduce it and just drop the context from the implementation of `Add` if we think the tracing is important here.

PR #3086, opened 2024-06-05
Remove deadcode

dhrubabasu:
## Why this should be merged

s/o @darioush for pointing out this tool

```
go install golang.org/x/tools/cmd/deadcode@latest
deadcode -test ./...
```

## How this works

Removes some deadcode. There may be more from that tool but these are the ones I feel comfortable with removing.

## How this was tested

CI


utils/sampler/weighted_without_replacement.go
```diff
@@ -26,13 +26,3 @@ func NewWeightedWithoutReplacement() WeightedWithoutReplacement {
 		w: NewWeighted(),
 	}
 }
-
-// NewBestWeightedWithoutReplacement returns a new sampler
-func NewBestWeightedWithoutReplacement(
```
joshua-kim, line 31: I don't think we ever use this outside of tests. Should we delete the entire struct too since we're doing the same for `UniformBest` (or keep both)?

dhrubabasu: Broke out the utils/sampler change here: https://github.com/ava-labs/avalanchego/pull/3219

dhrubabasu:
> Maybe add a new deadcode linter?

There's a decent amount of false positives, this wouldn't really belong in CI IMO

StephenButtolph (approved):

PR #3089, opened 2024-06-06
Fix race in test

StephenButtolph:
## Why this should be merged

This test has a race: https://github.com/ava-labs/avalanchego/actions/runs/9402917779/job/25898166965?pr=3053#step:5:348

## How this works

Correctly locks state before using it.

## How this was tested

- [X] CI

### DIFF
--- a/vms/platformvm/service_test.go
+++ b/vms/platformvm/service_test.go
@@ -1085,10 +1085,13 @@ func TestServiceGetSubnets(t *testing.T) {
 	newOwnerIDStr := "P-testing1t73fa4p4dypa4s3kgufuvr6hmprjclw66mgqgm"
 	newOwnerID, err := service.addrManager.ParseLocalAddress(newOwnerIDStr)
 	require.NoError(err)
+
+	service.vm.ctx.Lock.Lock()
 	service.vm.state.SetSubnetOwner(testSubnet1ID, &secp256k1fx.OutputOwners{
 		Addrs:     []ids.ShortID{newOwnerID},
 		Threshold: 1,
 	})
+	service.vm.ctx.Lock.Unlock()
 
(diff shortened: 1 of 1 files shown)


abi87 (approved):

vms/platformvm/service_test.go
```diff
@@ -1085,10 +1085,13 @@ func TestServiceGetSubnets(t *testing.T) {
 	newOwnerIDStr := "P-testing1t73fa4p4dypa4s3kgufuvr6hmprjclw66mgqgm"
 	newOwnerID, err := service.addrManager.ParseLocalAddress(newOwnerIDStr)
 	require.NoError(err)
+
+	service.vm.ctx.Lock.Lock()
 	service.vm.state.SetSubnetOwner(testSubnet1ID, &secp256k1fx.OutputOwners{
 		Addrs:     []ids.ShortID{newOwnerID},
 		Threshold: 1,
 	})
+	service.vm.ctx.Lock.Unlock()
```
tsachiherman, line 1094: this would definitally fix the race error - but it might fail ( in the future ) for other errors -
the subnet owners might change between the SetSubnetOwner and GetSubnets below.
would it make sense to have a lock-free `service.getSubnetsInternal`, and wrap the lock here around both calls ?

StephenButtolph: I think there is a long conversation to be had around how we current test services and VMs in unit tests... But I'm not really sure increasing the scope of the locking is the direction we'd want to go.

My hope is that when we switch to use the merkledb as a backend for our state, that we'll be able to essentially remove the usage of the context lock entirely and use the merkledb's internal locking to provide consistent reads + writes.

This test is correct now that the locking is added here because the test fully owns the VM (and it isn't concurrently issuing any transactions that could modify the subnet owner).

PR #3091, opened 2024-06-06
[ci] Ensure focal arm64 builds all have their required dependencies

marun:


StephenButtolph (approved):

darioush (approved):

tsachiherman (approved):
looks good to me. As you've already suggested, it would be better for us to find an image that already has that dependency.

even smaller note : maybe the `uses: ./.github/actions/install-focal-deps` should take place before the `actions/checkout` so that it would be the same "flow" as if it was already included in the image.

PR #3104, opened 2024-06-11
[antithesis] Remove assertions incompatible with fault injection

marun:
## Why this should be merged

The antithesis test runs are currently reporting errors due to the recent addition of assertions that cannot be guaranteed to succeed in the face of injected faults. 

Future effort will ensure the addition of more reliable assertions and/or add retries to account for the possibility of temporary error conditions induced by injected faults.

## How this works

As per @StephenButtolph:

Remove assertions for network calls: 

- client calls
- wallet.Issue calls
- primary.AddAllUTXOs calls

Retain assertions for local calls:

 - utxos.GetUTXO
- status checks
- balance checks

## How this was tested
 
- [x] Manually triggered publication of images from this branch: 
  - https://github.com/ava-labs/avalanchego/actions/runs/9467893563
- [x] Manually triggered a run for avalanchego images
   - [x] Passed?
- [x] Manually triggered a run for xsvm images
   - [x] Passed?

### DIFF
--- a/tests/antithesis/avalanchego/main.go
+++ b/tests/antithesis/avalanchego/main.go
@@ -263,10 +263,6 @@ func (w *workload) issueXChainBaseTx(ctx context.Context) {
 	)
 	if err != nil {
 		log.Printf("failed to issue X-chain baseTx: %s", err)
-		assert.Unreachable("failed to issue X-chain baseTx", map[string]any{
-			"worker": w.id,
-			"err":    err,
-		})
 		return
 	}
 	log.Printf("issued new X-chain baseTx %s in %s", baseTx.ID(), time.Since(baseStartTime))
@@ -320,10 +316,6 @@ func (w *workload) issueXChainCreateAssetTx(ctx context.Context) {
 	)
 	if err != nil {
 		log.Printf("failed to issue X-chain create asset transaction: %s", err)
-		assert.Unreachable("failed to issue X-chain create asset transaction", map[string]any{
-			"worker": w.id,
-			"err":    err,
-		})
 		return
 	}
 	log.Printf("created new X-chain asset %s in %s", createAssetTx.ID(), time.Since(createAssetStartTime))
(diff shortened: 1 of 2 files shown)


tests/antithesis/avalanchego/main.go
```diff
@@ -655,19 +599,13 @@ func (w *workload) verifyXChainTxConsumedUTXOs(ctx context.Context, tx *xtxs.Tx)
...
-				"err":    err,
-			})
 			return
 		}
 
 		inputs := tx.Unsigned.InputIDs()
 		for input := range inputs {
 			_, err := utxos.GetUTXO(ctx, chainID, chainID, input)
-			if err != database.ErrNotFound {
+			if err != nil && err != database.ErrNotFound {
```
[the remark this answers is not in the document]
darioush: I think this changes the semantics, `failed to verify that X-chain UTXO %s was deleted`

aaronbuchwald (approved):
LGTM - my understanding here is the retained assertions do NOT require any API call to the node since the wallet tracks the UTXO set locally.

tests/antithesis/avalanchego/main.go
```diff
@@ -584,22 +552,10 @@ func (w *workload) confirmXChainTx(ctx context.Context, tx *xtxs.Tx) {
...
 			return
 		}
 		if status != choices.Accepted {
 			log.Printf("failed to confirm X-chain transaction %s on %s: status == %s", txID, uri, status)
-			assert.Unreachable("failed to confirm an X-chain transaction", map[string]any{
-				"worker": w.id,
-				"txID":   txID,
-				"uri":    uri,
-				"status": status,
-			})
```
StephenButtolph, line 602: It's my understanding that this assertion could remain. (I'd be interested if it couldn't stay)

tests/antithesis/avalanchego/main.go
```diff
@@ -614,22 +570,10 @@ func (w *workload) confirmPChainTx(ctx context.Context, tx *ptxs.Tx) {
...
 			return
 		}
 		if s.Status != status.Committed {
 			log.Printf("failed to confirm P-chain transaction %s on %s: status == %s", txID, uri, s.Status)
-			assert.Unreachable("failed to confirm a P-chain transaction", map[string]any{
-				"worker": w.id,
-				"txID":   txID,
-				"uri":    uri,
-				"status": s.Status,
-			})
```
StephenButtolph, line 632: It's my understanding that this assertion could remain. (I'd be interested if it couldn't stay)

tests/antithesis/avalanchego/main.go
aaronbuchwald: As I understand it, if this hits an error here: https://github.com/ava-labs/avalanchego/blob/master/wallet/chain/x/wallet.go#L303 when issuing the transaction while the node is down due to an injected failure then the test will fail incorrectly

StephenButtolph (approved):

PR #3105, opened 2024-06-11
Add P2P test utility

joshua-kim:
## Why this should be merged

Currently testing p2p apis is painful and requires mocking the underlying app sender calls to redirect messages to a VM's network handler. This will be used in a future PR when we require testing for the `x/sync` package's migration to use the `p2p` package.

## How this works

Adds a test utility to remove the boiler plate code and provide a cleaner ux.

## How this was tested

Added ut


aaronbuchwald (approved):
LGTM

PR #3106, opened 2024-06-12
Migrate `x/sync` to `p2p`

joshua-kim:
## Why this should be merged

Eliminates our last usage of the blocking clients

## How this works

Replaces usage of blocking client with `p2p` async clients

## How this was tested

Added unit tests


x/sync/client.go
```diff
@@ -31,177 +25,25 @@ const (
...
 	errInvalidChangeProof            = errors.New("failed to verify change proof")
 	errTooManyKeys                   = errors.New("response contains more than requested keys")
 	errTooManyBytes                  = errors.New("response contains more than requested bytes")
 	errUnexpectedChangeProofResponse = errors.New("unexpected response type")
 )
 
-// Client synchronously fetches data from the network
-// to fulfill state sync requests.
-// Repeatedly retries failed requests until the context is canceled.
 type Client interface {
```
joshua-kim, line 35: Pretty annoying - we need to have this interface still for testing purposes. This is because we need to be able to mock timeouts to test our retries because the p2p package does not actually send app errors yet. This can be removed once https://github.com/ava-labs/avalanchego/pull/2753 is merged.

x/sync/metrics.go
```diff
@@ -19,7 +19,7 @@ var (
 type SyncMetrics interface {
 	RequestFailed()
 	RequestMade()
-	RequestSucceeded()
+	RequestSucceeded() //TODO only need one of succeeded/failed
```
joshua-kim, line 22: Think we don't need both failed + success metric since we can infer one off the other if we know the `n` of requests + one of either the success/failed count

x/sync/manager.go
```diff
@@ -234,38 +274,47 @@ func (m *Manager) close() {
...
-	}()
+	m.processingWorkItems--
+	m.unprocessedWorkCond.Signal()
+}
+
+// Processes [item] by fetching a change or range proof.
+func (m *Manager) doWork(ctx context.Context, work *workItem) {
+	// Backoff for failed requests accounting for time this job has already
+	// spent waiting in the unprocessed queue
+	backoff := calculateBackoff(work.attempt)
```
joshua-kim, line 289: Not optimal since this worker goroutine is going to be waiting until this timeout passes and ideally we would prioritize requests that are always executable, but felt like something more complex would make the diff hard to review since the scope of this PR is just trying to get rid of the blocking networking calls

x/sync/client_test.go
```diff
@@ -39,739 +34,29 @@ func newDefaultDBConfig() merkledb.Config {
...
-			serverResponseChan <- responseBytes
-
-			return nil
-		},
-	).AnyTimes()
-
-	return client.GetRangeProof(ctx, request)
-}
-
-func TestGetRangeProof(t *testing.T) {
```
joshua-kim, line 154: All of these tests were moved into `Test_Sync_Result_Correct_Root`

x/sync/manager.go
```diff
@@ -354,46 +384,267 @@ func (m *Manager) getAndApplyRangeProof(ctx context.Context, work *workItem) {
...
+
+	// Get the next nodeID to query using the [nodeIdx] offset.
+	// If we're out of nodes, loop back to 0.
+	// We do this try to query a different node each time if possible.
+	nodeIdx := atomic.AddUint32(&m.stateSyncNodeIdx, 1)
+	nodeID := m.config.StateSyncNodes[nodeIdx%uint32(len(m.config.StateSyncNodes))]
+	return client.AppRequest(ctx, set.Of(nodeID), requestBytes, onResponse)
+}
+
+func (m *Manager) retryWork(work *workItem) {
```
joshua-kim, line 440: Another idea is to have a separate failed requests data structure. Feel like this will introduce more complexity into this PR though so for now we are just retrying by putting stuff in the work heap w/ it's own priority.

x/sync/manager.go
```diff
@@ -354,46 +384,267 @@ func (m *Manager) getAndApplyRangeProof(ctx context.Context, work *workItem) {
...
+		if err := m.config.DB.CommitRangeProof(ctx, work.start, work.end, rangeProof); err != nil {
+			m.setError(err)
+			return nil
+		}
+		largestHandledKey = maybe.Some(rangeProof.KeyValues[len(rangeProof.KeyValues)-1].Key)
 	}
 
-	m.completeWorkItem(ctx, work, largestHandledKey, targetRootID, proof.EndProof)
+	m.completeWorkItem(ctx, work, largestHandledKey, targetRootID, rangeProof.EndProof)
+	return nil
```
StephenButtolph, line 647: What is the purpose of:
```golang
var changeOrRangeProof *merkledb.ChangeOrRangeProof
```
I feel like the entire `merkledb.ChangeOrRangeProof` should be removed and the usage of that variable should be replaced with moving the logic into the switch statement.

x/sync/network_server.go
```diff
@@ -50,146 +48,70 @@ var (
...
+	_ p2p.Handler = (*SyncGetRangeProofHandler)(nil)
+)
 
-func NewNetworkServer(appSender common.AppSender, db DB, log logging.Logger) *NetworkServer {
-	return &NetworkServer{
-		appSender: appSender,
-		db:        db,
-		log:       log,
+func maybeBytesToMaybe(mb *pb.MaybeBytes) maybe.Maybe[[]byte] {
+	if mb != nil && !mb.IsNothing {
```
StephenButtolph, line 57: I think `mb.IsNothing` could just be removed right? Because `mb == nil` already implies `Nothing`. Should we remove this field from the type? Or is this needed for some other language?

To clarify - the `mb.IsNothing` check is needed if it's on the type... But I think we could get rid of it entirely.

x/sync/network_client.go
```diff
@@ -1,368 +0,0 @@
...
-	lock sync.Mutex
-	log  logging.Logger
-	// requestID counter used to track outbound requests
-	requestID uint32
-	// requestID => handler for the response/failure
-	outstandingRequestHandlers map[uint32]ResponseHandler
-	// controls maximum number of active outbound requests
-	activeRequests *semaphore.Weighted
-	// tracking of peers & bandwidth usage
-	peers *p2p.PeerTracker
```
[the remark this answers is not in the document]
joshua-kim: Yeah we'll need to merge in some form of https://github.com/ava-labs/avalanchego/pull/3007

x/sync/client.go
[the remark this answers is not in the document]
joshua-kim: Yeah. Originally did this to minimize the diff but agree this looks weird now.

x/sync/manager.go
joshua-kim: It's not clear to me... this was previously existing code. As far as I can tell it's not needed though. We can clean it up as part of this PR.

x/sync/network_server.go
joshua-kim: This is code that existed previously but will get rid of the `IsNothing` check

x/sync/manager.go
danlaine: iirc this exists because the response to a change proof request can be either a range proof or a change proof, so we need to be able to parse either response

x/sync/network_server.go
joshua-kim: I think this is showing up as a diff because of some code that got moved around it but this is previously existing code

StephenButtolph: A future improvement then haha

x/sync/manager.go
StephenButtolph: > iirc this exists because the response to a change proof request can be either a range proof or a change proof, so we need to be able to parse either response

Both responses are already handled. We end up wrapping the different responses into another custom type for (afaict) no reason

x/sync/manager.go
```diff
@@ -234,38 +283,54 @@ func (m *Manager) close() {
...
+	deadline, ok := ctx.Deadline()
+	if ok && now.Add(waitTime).After(deadline) {
+		m.finishWorkItem()
+		return
+	}
 
-		m.processingWorkItems--
-		m.unprocessedWorkCond.Signal()
-	}()
+	<-time.After(waitTime)
```
StephenButtolph, line 308: Previously, the context would be honored here and we wouldn't block this thread during shutdown. I think we should still be cleaning this up. Right?

x/sync/manager.go
```diff
@@ -354,46 +400,252 @@ func (m *Manager) getAndApplyRangeProof(ctx context.Context, work *workItem) {
...
+}
+
+func (m *Manager) retryWork(work *workItem) {
+	work.priority = retryPriority
+	work.queueTime = time.Now()
+	work.requestFailed()
+
+	m.workLock.Lock()
+	m.unprocessedWork.Insert(work)
+	m.workLock.Unlock()
```
StephenButtolph, line 463: Shouldn't we be calling `unprocessedWorkCond.Signal()` here?

x/sync/manager.go
```diff
@@ -354,46 +400,252 @@ func (m *Manager) getAndApplyRangeProof(ctx context.Context, work *workItem) {
...
+}
+
+// Returns an error if we should drop the response
+func (m *Manager) handleResponse(
+	bytesLimit uint32,
+	responseBytes []byte,
+	err error,
+) error {
```
StephenButtolph, line 471: Can we make a better name for this function? This function doesn't actually handle the response... It returns an error if the caller shouldn't handle the response.

x/sync/network_server.go
```diff
@@ -50,128 +48,46 @@ var (
...
-	// TODO danlaine: Do we need this? Why?
-	if time.Until(bufferedDeadline) < minRequestHandlingDuration {
-		// Drop the request if we already missed the deadline to respond.
-		s.log.Info(
-			"deadline to process AppRequest has expired, skipping",
-			zap.Stringer("nodeID", nodeID),
-			zap.Uint32("requestID", requestID),
-		)
-		return nil
+func NewSyncGetChangeProofHandler(log logging.Logger, db DB) *SyncGetChangeProofHandler {
```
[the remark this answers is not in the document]
joshua-kim: Yeah looks like this is all in the diff anyways, I can rename these.

x/sync/network_server.go
```diff
@@ -198,12 +120,18 @@ func (s *NetworkServer) HandleChangeProofRequest(
 			if !errors.Is(err, merkledb.ErrInsufficientHistory) {
 				// We should only fail to get a change proof if we have insufficient history.
 				// Other errors are unexpected.
-				return err
+				return nil, &common.AppError{
+					Code:    p2p.ErrUnexpected.Code,
+					Message: fmt.Sprintf("failed to get change proof: %s", err),
+				}
```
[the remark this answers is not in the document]
joshua-kim: Added a TODO

x/sync/sync_test.go
```diff
@@ -79,20 +39,18 @@ func Test_Creation(t *testing.T) {
 
 	syncer, err := NewManager(ManagerConfig{
 		DB:                    db,
-		Client:                NewMockClient(ctrl),
-		TargetRoot:            ids.Empty,
+		RangeProofClient:      &p2p.Client{},
+		ChangeProofClient:     &p2p.Client{},
```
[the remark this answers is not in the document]
joshua-kim: Yeah these panic if you use them. I have been usually initializing the bare-minimum to make the test pass... but maybe this is not very test-driven approach...

x/sync/manager.go
joshua-kim: Yeah it's an invariant we're suppsoed to honor to signal after adding to the workheap...

x/sync/sync_test.go
```diff
@@ -793,68 +754,257 @@ func TestFindNextKeyRandom(t *testing.T) {
 	}
 }
 
+// Tests that we are able to sync to the correct root while the server is
+// updating
 func Test_Sync_Result_Correct_Root(t *testing.T) {
```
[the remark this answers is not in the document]
joshua-kim: Yeah server-side-flakes/client-side-error handling weren't really tested previously so I just ended up re-using this test for it... updated the naming here

x/sync/manager.go
```diff
@@ -277,74 +347,55 @@ func (m *Manager) getAndApplyChangeProof(ctx context.Context, work *workItem) {
...
-		largestHandledKey := work.end
-		// if the proof wasn't empty, apply changes to the sync DB
-		if len(changeProof.KeyChanges) > 0 {
-			if err := m.config.DB.CommitChangeProof(ctx, changeProof); err != nil {
-				m.setError(err)
-				return
-			}
-			largestHandledKey = maybe.Some(changeProof.KeyChanges[len(changeProof.KeyChanges)-1].Key)
+		if err := m.handleChangeProofResponse(ctx, targetRootID, work, request, responseBytes, err); err != nil {
+			m.config.Log.Debug("dropping response", zap.Error(err))
```
darioush, line 376: should we log some information about the request/response?

x/sync/manager.go
```diff
@@ -234,38 +283,59 @@ func (m *Manager) close() {
...
+	waitTime := max(0, calculateBackoff(work.attempt)-now.Sub(work.queueTime))
+
+	// Check if we can start this work item before the context deadline
+	deadline, ok := ctx.Deadline()
+	if ok && now.Add(waitTime).After(deadline) {
+		m.finishWorkItem()
+		return
+	}
+
+	<-time.After(waitTime)
```
StephenButtolph, line 313: I think we are still blocking here if the context is cancelled. I think we should remove the above check for `ctx.Err()` and use a `select`

x/sync/network_server.go
```diff
@@ -184,33 +100,46 @@ func (s *NetworkServer) HandleChangeProofRequest(
...
 	for keyLimit > 0 {
-		changeProof, err := s.db.GetChangeProof(ctx, startRoot, endRoot, start, end, int(keyLimit))
+		changeProof, err := g.db.GetChangeProof(ctx, startRoot, endRoot, start, end, int(keyLimit))
 		if err != nil {
 			if !errors.Is(err, merkledb.ErrInsufficientHistory) {
 				// We should only fail to get a change proof if we have insufficient history.
 				// Other errors are unexpected.
-				return err
+				// TODO define custom errors
+				return nil, &common.AppError{
```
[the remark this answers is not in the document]
joshua-kim: This is because previously the network server used `AppRequest`, implementing the VM `AppHandler`. Errors are fatal so we would swallow any non-fatal errors and return nil, and return an err if the error was fatal.

In `p2p`, errors are never fatal. We return `[]byte, error` where `[]byte` is the app response, and `error` is the app error. Which ever one is set is sent back to the requesting node id.

x/sync/network_server_test.go
```diff
@@ -27,116 +28,115 @@ func Test_Server_GetRangeProof(t *testing.T) {
...
 			request: &pb.SyncGetRangeProofRequest{
 				RootHash:   smallTrieRoot[:],
 				KeyLimit:   2 * defaultRequestKeyLimit,
 				BytesLimit: defaultRequestByteSizeLimit,
 			},
 			expectedResponseLen: defaultRequestKeyLimit,
 		},
-		"bytes limit too large": {
+		{
+			name: "response bounded by byte limit",
```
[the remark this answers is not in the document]
joshua-kim: it's the same test but we renamed it because the previous name makes it sound like it's supposed to error when the expected behavior is to truncate/cap the response from the server.

x/sync/network_server_test.go
```diff
@@ -155,77 +155,117 @@ func Test_Server_GetChangeProof(t *testing.T) {
...
-		key := make([]byte, r.Intn(100))
-		_, err = r.Read(key)
-		require.NoError(t, err)
-
-		val := make([]byte, r.Intn(100))
-		_, err = r.Read(val)
-		require.NoError(t, err)
-
-		ops = append(ops, database.BatchOp{Key: key, Value: val})
+	for x := 0; x < defaultRequestKeyLimit/2; x++ {
```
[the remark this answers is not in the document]
joshua-kim: This is from here: https://github.com/ava-labs/avalanchego/blob/ef3b7b5aed9878a6a7f0b48d6448a4124f985608/x/sync/client_test.go#L504, which is heavily duplicated with this unit test, so I consolidated them.

x/sync/sync_test.go
```diff
@@ -793,68 +759,257 @@ func TestFindNextKeyRandom(t *testing.T) {
...
+				handler := newFlakyChangeProofHandler(t, db, func(response *merkledb.ChangeProof) {
+					response.StartProof = nil
+					response.EndProof = nil
+				})
+
+				return p2ptest.NewClient(t, context.Background(), handler, ids.GenerateTestNodeID(), ids.GenerateTestNodeID())
+			},
+		},
+		{
+			name: "flaky change proof client",
```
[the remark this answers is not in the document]
joshua-kim: Renamed for consistency

x/sync/sync_test.go
```diff
@@ -1196,10 +1279,20 @@ func generateTrieWithMinKeyLen(t *testing.T, r *rand.Rand, count int, minKeyLen
...
+			return db, err
 		}
 		i++
 	}
-	slices.SortFunc(allKeys, bytes.Compare)
-	return db, allKeys, batch.Write()
+	return db, batch.Write()
+}
+
+type testHandler struct {
```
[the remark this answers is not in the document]
joshua-kim: Moved+renamed

x/sync/manager.go
```diff
@@ -277,74 +347,55 @@ func (m *Manager) getAndApplyChangeProof(ctx context.Context, work *workItem) {
...
+		KeyLimit:   defaultRequestKeyLimit,
+		BytesLimit: defaultRequestByteSizeLimit,
+	}
+
+	requestBytes, err := proto.Marshal(request)
 	if err != nil {
+		m.finishWorkItem()
 		m.setError(err)
 		return
 	}
```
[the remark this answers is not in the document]
joshua-kim: It's not obvious to me that it does... but the reasoning for the ordering in this PR is that it's consistent with the existing code which is always to call `finishWorkItem` -> `setError`.

x/sync/manager.go
joshua-kim: https://github.com/ava-labs/avalanchego/blob/5c070f8ccefcf8ef5b7c77d17da88bccd05298fc/x/sync/client.go#L316

Previously we would make an outgoing request and if it failed, we would just retry. Since this was called synchronously it meant that the the blocking thread would basically just keep trying a request until it succeeded.

In the async style we're just failing and re-inserting it back onto the pending request heap with the highest priority. I think there's probably better ways to handle the retry strategy but I am trying to change it in this PR because this PR is already unmanageably big.

x/sync/manager.go
```diff
@@ -354,46 +405,253 @@ func (m *Manager) getAndApplyRangeProof(ctx context.Context, work *workItem) {
...
+
 	largestHandledKey := work.end
 
 	// Replace all the key-value pairs in the DB from start to end with values from the response.
-	if err := m.config.DB.CommitRangeProof(ctx, work.start, work.end, proof); err != nil {
+	if err := m.config.DB.CommitRangeProof(ctx, work.start, work.end, &rangeProof); err != nil {
 		m.setError(err)
-		return
+		return nil
 	}
```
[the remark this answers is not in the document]
joshua-kim: We only return errors in this function when we want the caller to retry. Commiting to the db i don't think is something we can fix with a retry so we're fataling here by setting the error + returning nil to not retry in the caller.

x/sync/manager.go
joshua-kim: logging the response will make this diff bigger and we don't already do it, but it's trivial to add logging the request.

x/sync/sync_test.go
```diff
@@ -793,68 +759,257 @@ func TestFindNextKeyRandom(t *testing.T) {
...
+			}, prometheus.NewRegistry())
+
+			require.NoError(err)
+			require.NotNil(syncer)
+
+			// Start syncing from the server
+			require.NoError(syncer.Start(ctx))
+
+			// Simulate writes on the server
+			// TODO more than a single write when API is less flaky
```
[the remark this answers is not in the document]
joshua-kim: Yea good callout - I actually had to re-inspect this test to figure out what the race condition was. Added a comment here, but basically `UpdateSyncTarget` errors if it has already finished syncing (i.e it the syncer closes once it hits the target it's trying to hit). I noticed on larger `N` of writes that we were erroring out sometimes because we would finish a previous sync target before we updated it.

x/sync/sync_test.go
```diff
@@ -917,71 +1075,8 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
 	require.Equal(syncRoot, newRoot)
 }
 
-func Test_Sync_Error_During_Sync(t *testing.T) {
```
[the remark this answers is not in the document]
joshua-kim: We actually don't exhaust retries (previous code didn't either so I didn't add it into this PR). Retry behavior is tested in `Test_Sync_Result_Correct_Root`.

x/sync/sync_test.go
```diff
@@ -793,68 +759,257 @@ func TestFindNextKeyRandom(t *testing.T) {
...
+					response.KeyChanges = append(response.KeyChanges, make([]merkledb.KeyChange, defaultRequestKeyLimit)...)
+				})
+
+				return p2ptest.NewClient(t, context.Background(), handler, ids.GenerateTestNodeID(), ids.GenerateTestNodeID())
+			},
+		},
+		{
+			name: "change proof bad response - removed first key in response",
+			changeProofClient: func(db merkledb.MerkleDB) *p2p.Client {
+				handler := newFlakyChangeProofHandler(t, db, func(response *merkledb.ChangeProof) {
```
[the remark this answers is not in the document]
joshua-kim: I guess that would be one way of writing the test, but I would argue that we don't care how many retries happened as long as we get the correct resulting root (in-spite of server-side flakes) anyways. I feel like this more closely resembles the way that a real-usage of the caller would behave, as the caller would have no way of introspecting into the retires the `Manager` performs.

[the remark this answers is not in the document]
darioush: I think it would give more confidence to assert the modification code was called.

x/sync/network_server_test.go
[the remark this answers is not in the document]
darioush: We could have a case where the bytes limit requested by the client is valid (less than the limit enforced by the server) and is the limiting factor in the response size.

x/sync/manager.go
darioush: I think logging the request is good enough for now

x/sync/client_test.go
```diff
@@ -40,739 +38,110 @@ func newDefaultDBConfig() merkledb.Config {
...
-				KeyLimit:      defaultRequestKeyLimit,
-				BytesLimit:    defaultRequestByteSizeLimit,
-			},
-			modifyChangeProofResponse: func(response *merkledb.ChangeProof) {
-				response.StartProof = nil
-				response.EndProof = nil
-			},
-			expectedErr: merkledb.ErrInvalidProof,
-		},
-		"range proof response; remove first key": {
```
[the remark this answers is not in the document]
darioush: These are slightly different tests, one through the path where a range proof is requested and one through the path where a change proof is requested but a range proof is returned.
I suppose we can rely on the range proof test to verify failure cases for range proofs

darioush (approved):

StephenButtolph (approved):

PR #3110, opened 2024-06-12
Standardize wallet tx acceptance polling

StephenButtolph:
## Why this should be merged

This standardizes the names and behavior of `avm.Client#ConfirmTx`, `platformvm.Client#AwaitTxDecided`, `xsvmapi.WaitForAcceptance` as: `avm.AwaitTxAccepted`, `platformvm.AwaitTxAccepted`, and `xsvmapi.AwaitTxAccepted`.

This also implements `awaitTxAccepted` for the C-chain wallet.

## How this works

The changes are to ensure that `AwaitTxAccepted` now returns if:
1. The context is cancelled
2. The tx is accepted
3. The tx is rejected
4. An API error is returned

## How this was tested

- [X] CI


vms/avm/client.go
```diff
@@ -766,3 +745,32 @@ func (c *client) Export(
...
+	ctx context.Context,
+	txID ids.ID,
+	freq time.Duration,
+	options ...rpc.Option,
+) error {
+	ticker := time.NewTicker(freq)
+	defer ticker.Stop()
+
+	for {
+		status, err := c.GetTxStatus(ctx, txID, options...)
```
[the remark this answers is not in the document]
StephenButtolph: This PR now reports client errors to the caller.

dhrubabasu (approved):

vms/example/xsvm/api/client.go
```diff
@@ -245,20 +245,23 @@ func (c *client) Message(
 	return resp.Message, resp.Signature, resp.Message.Initialize()
 }
 
-func WaitForAcceptance(
+func AwaitTxAccepted(
```
tsachiherman, line 248: this is a bit uncommon go function signature. I would expect it to look like
```golang
func (c Client) AwaitTxAccepted(ctx, address, nonce, freq, options)
```

tsachiherman (approved):

vms/example/xsvm/api/client.go
StephenButtolph: Client is an interface

PR #3123, opened 2024-06-17
Remove `Decided` from the `Consensus` interface

StephenButtolph:
## Why this should be merged

This PR simplifies the `snowman.Consensus` interface to not expose the `Decided` method.

## How this works

The `Decided` method can be implemented directly on top of the existing `snowman.Consensus` interface.

## How this was tested

- [X] CI

### DIFF
--- a/snow/consensus/snowman/topological.go
+++ b/snow/consensus/snowman/topological.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow"
-	"github.com/ava-labs/avalanchego/snow/choices"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowball"
 	"github.com/ava-labs/avalanchego/utils/bag"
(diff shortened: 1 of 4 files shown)


snow/consensus/snowman/topological.go
```diff
@@ -182,16 +181,6 @@ func (ts *Topological) Add(blk Block) error {
...
 }
 
-func (ts *Topological) Decided(blk Block) bool {
-	// If the block is decided, then it must have been previously issued.
-	if blk.Status().Decided() {
-		return true
-	}
-	// If the block is marked as fetched, we can check if it has been
-	// transitively rejected.
-	return blk.Status() == choices.Processing && blk.Height() <= ts.lastAcceptedHeight
```
StephenButtolph, line 192: The `choices.Processing` check was previously needed because block's used to have a `Parent` method that returned a `snowman.Block` which could potentially have a status of `choices.Unknown` (if the parent wasn't known). But we no longer have this weird mechanism.

darioush (approved):

PR #3124, opened 2024-06-17
Remove .Status() from .Accepted()

StephenButtolph:
## Why this should be merged

Removes the usage of `.Status()` from the consensus code.

## How this works

The blocks map includes both the last accepted block and all the processing blocks. This PR changes how we inspect the last accepted block to use heights rather than checking the status.

We update the last accepted variables immediately after calling `Accept`, so these values are always kept up-to-date.

## How this was tested

- [X] CI

### DIFF
--- a/snow/consensus/snowman/topological.go
+++ b/snow/consensus/snowman/topological.go
@@ -127,7 +127,7 @@ func (ts *Topological) Initialize(
 	ts.lastAcceptedID = lastAcceptedID
 	ts.lastAcceptedHeight = lastAcceptedHeight
 	ts.blocks = map[ids.ID]*snowmanBlock{
-		lastAcceptedID: {params: ts.params},
+		lastAcceptedID: {t: ts},
 	}
 	ts.preferredHeights = make(map[uint64]ids.ID)
 	ts.preference = lastAcceptedID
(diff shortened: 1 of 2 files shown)


snow/consensus/snowman/topological.go
```diff
@@ -289,7 +289,7 @@ func (ts *Topological) RecordPoll(ctx context.Context, voteBag bag.Bag[ids.ID])
 
 	// Runtime = |live set| ; Space = Constant
 	// Traverse from the preferred ID to the last accepted ancestor.
-	for block := startBlock; !block.Accepted(); {
+	for block := startBlock; !block.Decided(); {
```
aaronbuchwald, line 292: Could you add a comment for future reference why it's alright to switch from `Accepted()` to `Decided()` here?

aaronbuchwald (approved):

snow/consensus/snowman/topological.go
aaronbuchwald: My understanding: this code iterates from the preference to the last accepted ancestor, which guarantees that the blocks can only be Processing or Accepted (when it hits the last accepted block). It should never hit a rejected block, which makes this change a no-op.

StephenButtolph: added a comment

PR #3125, opened 2024-06-18
Refactor `event.Blocker` into `job.Scheduler`

StephenButtolph:
## Why this should be merged

Currently the `snow/event` package exposes a sub-optimal API. Because errors are not returned by the `event.Blocker`, we maintain an accumulated `error`. Additionally, each job must track outstanding dependencies to know when the jobs have been either fulfilled or abandoned.

## How this works

This PR replaces the `event.Blocker` and `event.Blockable` interfaces with a `job.Job` interface and `job.Scheduler` struct. These functions all returns errors so the error accumulator is removed.

Additionally all dependency tracking is removed from the `issuer` and `voter` structs.

Finally this implementation is moved into the `snow/engine/snowman` package as it is no longer used by the `avalanche` engine.

## How this was tested

- [x] Added test where voting must wait for both dependencies to be resolved.
- [X] CI


snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+	dependencies set.Set[T]
+	// If true, the job should be cancelled.
+	shouldCancel bool
+	// If nil, the job has already been executed or cancelled.
+	job Job
+}
+
+type Queue[T comparable] struct {
+	// jobs maps a dependency to the jobs that depend on it.
+	jobs map[T][]*job[T]
```
ARR4N, line 28: I think `dependents` is a clearer name than `jobs` (which I originally assumed to mean `dependencies`).

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+	delete(q.jobs, dependency)
+
+	for _, job := range jobs {
+		// Removing the dependency keeps the queue in a consistent state.
+		// However, it isn't strictly needed.
+		job.dependencies.Remove(dependency)
+		job.shouldCancel = shouldCancel || job.shouldCancel
+
+		userJob := job.job
+		if userJob == nil || job.dependencies.Len() != 0 {
```
ARR4N, line 97: The comment on `Abandon()` states that if _any_ dependencies are abandoned then the job will be cancelled. Why wait until all dependencies of that job are resolved? Short-circuiting would remove the need for the `job.shouldCancel` field (but may introduce other issues I haven't considered).

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+			continue
+		}
+
+		// Mark the job as cancelled so that any reentrant calls do not interact
```
ARR4N, line 101: (nit) `Mark the job as executed / cancelled so that...`

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package event
```
ARR4N, line 4: (style) I think this is a new package so it requires a package-level comment.

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+		cancel: func(context.Context) error {
+			return errUnexpectedInvocation
+		},
+	}
+
+	tests := []struct {
+		name          string
+		queue         *Queue[int]
+		dependencies  []int
+		shouldExecute bool
```
ARR4N, line 66: (style,nit) `wantExecute`; "should" implies that it's a parameter for the test loop whereas "want" is in keeping with idiomatic `got` and `want` (in place of `actual` and `expected`).

If it were me I'd also changed `expected*` to `want*` but I don't think it's as important since `expected` still conveys intent unambiguously.

ARR4N: Reading through the tests I now see that it's always false. I think you can remove it entirely, but my original comment re the prefix applies to the other tests so I'll leave it.

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			// Reset the variable between tests
+			calledExecute = false
+
+			require.NoError(test.queue.Register(context.Background(), userJob, test.dependencies...))
+			require.Equal(test.expectedLen, test.queue.NumDependencies())
```
ARR4N, line 157: `require.X()` will terminate early and create a whack-a-mole testing experience, so `assert.X()` is more appropriate. It makes sense to still use `require` for `NoError()` though, as that is guaranteed to invalidate all future tests.

This applies to all tests.

[See style guide](https://google.github.io/styleguide/go/decisions#keep-going)

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+							dependencies: set.Of(1, 2),
+							job:          userJob,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "additional dependency",
+			queue: &Queue[int]{
```
ARR4N, line 120: In the other test you use `newQueueWithJob(userJob)`; is that appropriate here too?

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+	}{
+		{
+			name:          "no jobs",
+			queue:         NewQueue[int](),
+			shouldExecute: false,
+			expectedQueue: NewQueue[int](),
+		},
+		{
+			name:          "single dependency",
+			queue:         newQueueWithJob(t, userJob, false, 1),
```
ARR4N, line 193: The use of `1` vs `2` is important but not obvious until later. I think you should introduce constants before the test table as it currently requires reading the test loop to be understood:

```go
const (
  depToFulfill = iota
  depToNeglect
)

...

test.queue.Fulfill(..., depToFulfil)
```

Same applies to the `Abandon()` test.

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+			calledCancel = true
+			return nil
+		},
+	}
+
+	tests := []struct {
+		name          string
+		queue         *Queue[int]
+		shouldCancel  bool
```
ARR4N, line 248: (style) `wantCancelled` with same rationale as earlier.

snow/engine/snowman/issuer.go
```diff
@@ -10,45 +10,27 @@ import (
...
-// Abandon the attempt to issue [i.block]
-func (i *issuer) Abandon(ctx context.Context, _ ids.ID) {
-	if !i.abandoned {
-		blkID := i.blk.ID()
-		i.t.removeFromPending(i.blk)
-		i.t.addToNonVerifieds(i.blk)
-		i.t.blocked.Abandon(ctx, blkID)
-	}
-	i.abandoned = true
+func (i *issuer) Execute(ctx context.Context) error {
```
ARR4N, line 27: (fyi) I don't have enough context to properly review this, but noting that to me it looks like a correct refactoring.

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+type Job interface {
```
ARR4N, line 12: (style) This and `Queue` require comments as top-level, exported identifiers.

[See style guide](https://google.github.io/styleguide/go/decisions#assert:~:text=All%20top%2Dlevel%20exported%20names%20must%20have%20doc%20comments)

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+type job[T comparable] struct {
+	// If empty, the job is ready to be executed.
+	dependencies set.Set[T]
+	// If true, the job should be cancelled.
+	shouldCancel bool
+	// If nil, the job has already been executed or cancelled.
+	job Job
+}
+
+type Queue[T comparable] struct {
```
ARR4N, line 26: (pedantry) It's not really a queue, which might give the wrong impression to someone who wants to use it in the future. The same could probably be argued about the package name too: `event.Queue` is misleading. That said, I'm not exactly sure what it is! Perhaps rename both and this becomes a `job.Pool`?

snow/engine/snowman/transitive_test.go
```diff
@@ -170,7 +170,7 @@ func TestEngineDropsAttemptToIssueBlockAfterFailedRequest(t *testing.T) {
 	// job blocked on [parent]'s issuance.
 	require.NoError(engine.Put(context.Background(), peerID, 0, child.Bytes()))
 	require.NotNil(request)
-	require.Len(engine.blocked, 1)
+	require.Equal(1, engine.blocked.NumDependencies())
```
ARR4N, line 173: Since you're already updating it, consider `assert.Equal` (see rationale in earlier comment).

snow/engine/snowman/voter.go
```diff
@@ -9,39 +9,21 @@ import (
...
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/event"
 	"github.com/ava-labs/avalanchego/utils/bag"
-	"github.com/ava-labs/avalanchego/utils/set"
 )
 
+var _ event.Job = (*voter)(nil)
+
 // Voter records chits received from [vdr] once its dependencies are met.
```
ARR4N, line 18: This needs to change to `...received from [nodeID] once...`.

snow/engine/snowman/voter.go
```diff
@@ -75,24 +57,24 @@ func (v *voter) Update(ctx context.Context) {
...
+		return nil
 	}
 
 	v.t.Ctx.Log.Debug("Snowman engine can't quiesce")
 	v.t.repoll(ctx)
+	return nil
+}
+
+func (v *voter) Cancel(ctx context.Context) error {
+	return v.Execute(ctx)
```
ARR4N, line 79: Why is cancellation considered equivalent to execution? I don't know much about this part of the code, but I'd assume that a call to `Cancel()` should never happen? If so, I'd return an error.

ARR4N (changes requested):
I don't know enough to know if this can be approved or not, sorry.

snow/engine/snowman/event/queue.go
StephenButtolph: Hm - perhaps a job.Scheduler?

snow/engine/snowman/transitive.go
```diff
@@ -1026,23 +1024,29 @@ func (t *Transitive) deliver(
 		t.sendQuery(ctx, blkID, blk.Bytes(), push)
 	}
 
-	t.blocked.Fulfill(ctx, blkID)
+	if err := t.blocked.Fulfill(ctx, blkID); err != nil {
```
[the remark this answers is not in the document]
StephenButtolph: Any encountered error is considered FATAL - so immediately returning seems more reasonable than what was done previously.

snow/engine/snowman/event/queue_test.go
StephenButtolph: `shouldExecute` is set to `true` in the first case

snow/engine/snowman/event/queue_test.go
StephenButtolph: I had avoided using it in this test because it internally calls `Register`... Which is what this function is testing... But I think it it's fine to use for the setup (not the expected).

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+			for _, j := range jobs {
+				j.shouldCancel = true
+			}
+		}
+	}
+	return q
+}
+
+func TestQueue_Register(t *testing.T) {
+	var calledExecute bool
```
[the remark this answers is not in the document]
StephenButtolph: > In the process, maybe avoid reuse across test cases to allow use of calledExecute as a struct property instead of reuse/reset across tests.

I managed to simplify the flow, but we use the `userJob` to specify the structure of the tests... And it wasn't immediately clear to me how best to avoid that re-use with how we are handling the `wantScheduler`.

snow/engine/snowman/event/queue_test.go
```diff
@@ -0,0 +1,294 @@
...
+		name          string
+		queue         *Queue[int]
+		shouldExecute bool
+		expectedQueue *Queue[int]
+	}{
+		{
+			name:          "no jobs",
+			queue:         NewQueue[int](),
+			shouldExecute: false,
+			expectedQueue: NewQueue[int](),
```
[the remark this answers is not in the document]
StephenButtolph: We need to instantiate the empty queue for the `queue` variable (as it is modified)... It feels a tad odd to me to use `NewQueue` in some places and then have an `emptyQueue` constant for other places...

snow/engine/snowman/event/queue.go
StephenButtolph: Talked with @aaronbuchwald offline about this... going to make a sub-PR with an alternative design here to make the requirements a bit more explicit.

snow/engine/snowman/event/queue.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Register a job that should be executed once all of its dependencies are
+// fulfilled. In order to prevent a memory leak, all dependencies must
+// eventually either be fulfilled or abandoned.
+//
+// While registering a job with duplicate dependencies is discouraged, it is
+// allowed and treated similarly to registering the job with the dependencies
+// de-duplicated.
+func (q *Queue[T]) Register(ctx context.Context, userJob Job, dependencies ...T) error {
+	if len(dependencies) == 0 {
```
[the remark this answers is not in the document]
StephenButtolph: @marun How about `Schedule` rather than `Register`?

snow/engine/snowman/transitive_test.go
```diff
@@ -2857,6 +2857,137 @@ func TestEngineVoteStallRegression(t *testing.T) {
 	require.Equal(choices.Rejected, rejectedChain[0].Status())
 }
 
+// When a voter is registered with multiple dependencies, the engine must not
+// execute the voter until all of the dependencies have been resolved; even if
+// one of the dependencies has been abandoned.
+func TestEngineEarlyTerminateVoterRegression(t *testing.T) {
```
StephenButtolph, line 2863: This is a regression testing the original pass of this PR in commit 1c1501dd60691d37d23d88f7ae8dc2b296b4b6fc. This regression was fixed inside this PR with commit e99eee0b3d832aec7f33b21bae67a1281351902a.

This bug never existed on the master branch, but it was a subtle enough of an issue that I felt it deserved an explicit test. (And clearly it is a mistake people could make... I made it.)

There is going to be a pass soon to cleanup all of these engine tests... But the first step of that is simplifying the engine and having better abstractions to test against.

snow/engine/snowman/event/queue.go
StephenButtolph: Ended up merging that PR in. Let me know what you think

snow/engine/snowman/voter.go
StephenButtolph: I refactored this a bit and added a comment - does that additional context answer your question?

snow/engine/snowman/event/queue_test.go
StephenButtolph: Currently we actually explicitly ban the usage of `assert` in favor of `require`: https://github.com/ava-labs/avalanchego/blob/2e72c7c29c498e7ffa5bb31ff18495e14e2cdfb7/.golangci.yml#L93-L94

In the past we had the issue (especially in some of our more, _ahem_, "legacy" tests) where assertions were used and caused annoying to debug errors (because the original error was lost in the weeds).

I think it probably makes sense to pull this conversation out of this PR and then we can discuss:
1. Unbanning `assert`.
2. Doing a pass to use `assert` where it makes sense.

snow/engine/snowman/event/queue_test.go
ARR4N: I agree with marun re consistency of `expect` over `want`.

> shouldExecute is set to true in the first case

Whoops!

snow/engine/snowman/event/queue.go
ARR4N: I like it

snow/engine/snowman/voter.go
ARR4N: Yup, makes more sense now

ARR4N (approved):
Your refactor is much easier to understand

snow/engine/snowman/job/scheduler.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// Schedule a job to be executed once all of its dependencies are resolved.
+//
+// In order to prevent a memory leak, all dependencies must eventually either be
+// fulfilled or abandoned.
+//
+// While registering a job with duplicate dependencies is discouraged, it is
+// allowed.
```
aaronbuchwald, line 44: optional
```suggestion
// allowed.
// If a job is scheduled with no dependencies, it's executed immediately.
```

aaronbuchwald (approved):
LGTM

PR #3126, opened 2024-06-18
[vms/platformvm] Add tracking of a Subnet manager

dhrubabasu:
## Why this should be merged

[ACP-77](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/77-reinventing-subnets/README.md) introduces a notion of a "Subnet manager" defined as a `sourceChainID` and `sourceAddress` pair. This PR adds tracking of this pair to the P-Chain state.

## How this works

Adds support for storing and retrieving a `chainID` and `addr` for each Subnet

## How this was tested

CI + added UTs


vms/platformvm/state/diff.go
```diff
@@ -294,6 +296,27 @@ func (d *diff) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
 	d.subnetOwners[subnetID] = owner
 }
 
+func (d *diff) GetSubnetManager(subnetID ids.ID) (ids.ID, []byte, error) {
+	manager, exists := d.subnetManagers[subnetID]
+	if exists {
+		return manager.chainID, manager.addr, nil
+	}
+
+	// If the subnet owner was not assigned in this diff, ask the parent state.
```
dhrubabasu, line 305:
```suggestion
	// If the subnet manager was not assigned in this diff, ask the parent state.
```

vms/platformvm/state/state.go
```diff
@@ -409,6 +418,11 @@ type fxOwnerAndSize struct {
 	size  int
 }
 
+type chainIDAndAddr struct {
```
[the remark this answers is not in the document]
dhrubabasu: I thought about this when drafting the PR but decided to punt the decision till later in the ACP-77 development process. Right now, it's only used as an intermediate struct only for the codec marshalling. The PlatformVM `state` pkg is not expected for external use either so it wouldn't live here if we do export it.

Good call out!

vms/platformvm/state/state_test.go
```diff
@@ -1401,6 +1401,28 @@ func TestStateSubnetOwner(t *testing.T) {
 	require.Equal(owner2, owner)
 }
 
+func TestStateSubnetManager(t *testing.T) {
```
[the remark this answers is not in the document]
dhrubabasu: I think testing the database stuff is a bit difficult to do, it's been a long-standing TODO on the `state` pkg. Will punt for now just for the sake of progressing on ACP-77

vms/platformvm/config/execution_config.go
```diff
@@ -36,6 +36,7 @@ type ExecutionConfig struct {
 	ChainDBCacheSize             int            `json:"chain-db-cache-size"`
 	BlockIDCacheSize             int            `json:"block-id-cache-size"`
 	FxOwnerCacheSize             int            `json:"fx-owner-cache-size"`
+	ChainIDAndAddrCacheSize      int            `json:"chain-id-and-addr-cache-size"`
```
StephenButtolph, line 39: Do we not have a more concise name for this? `SubnetManager` or something? `ChainIDAndAddr` feels pretty amorphous.

vms/platformvm/state/diff.go
```diff
@@ -294,6 +296,26 @@ func (d *diff) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
...
+	}
+	return parentState.GetSubnetManager(subnetID)
+}
+
+func (d *diff) SetSubnetManager(subnetID ids.ID, chainID ids.ID, addr []byte) {
+	d.subnetManagers[subnetID] = chainIDAndAddr{
+		ChainID: chainID,
+		Addr:    addr,
+	}
+}
```
StephenButtolph, line 317: Don't we need to use `subnetManagers` during `Apply`? Can we add a test for this?

vms/platformvm/state/diff.go
```diff
@@ -294,6 +296,26 @@ func (d *diff) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
...
+	// If the subnet manager was not assigned in this diff, ask the parent state.
+	parentState, ok := d.stateVersions.GetState(d.parentID)
+	if !ok {
+		return ids.Empty, nil, ErrMissingParentState
+	}
+	return parentState.GetSubnetManager(subnetID)
+}
+
+func (d *diff) SetSubnetManager(subnetID ids.ID, chainID ids.ID, addr []byte) {
+	d.subnetManagers[subnetID] = chainIDAndAddr{
```
StephenButtolph, line 313: Where is `subnetManagers` initialized? Test?

vms/platformvm/state/state.go
```diff
@@ -808,6 +837,42 @@ func (s *state) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
...
+		return chainIDAndAddr.ChainID, chainIDAndAddr.Addr, nil
+	}
+
+	if chainIDAndAddr, cached := s.subnetManagerCache.Get(subnetID); cached {
+		return chainIDAndAddr.ChainID, chainIDAndAddr.Addr, nil
+	}
+
+	chainIDAndAddrBytes, err := s.subnetManagerDB.Get(subnetID[:])
+	if err == database.ErrNotFound {
+		s.SetSubnetManager(subnetID, ids.Empty, []byte{})
```
StephenButtolph, line 851: Why are we calling this here?

vms/platformvm/state/state.go
```diff
@@ -808,6 +837,42 @@ func (s *state) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
...
+		return ids.Empty, []byte{}, nil
```
StephenButtolph, line 852: Returning `nil` here seems unexpected to me... Why are these values returned?

vms/platformvm/state/state.go
```diff
@@ -808,6 +837,42 @@ func (s *state) SetSubnetOwner(subnetID ids.ID, owner fx.Owner) {
...
+	}
+
+	var manager chainIDAndAddr
+	if _, err := block.GenesisCodec.Unmarshal(chainIDAndAddrBytes, &manager); err != nil {
+		return ids.Empty, nil, err
+	}
+	s.subnetManagerCache.Put(subnetID, chainIDAndAddr{
+		ChainID: manager.ChainID,
+		Addr:    manager.Addr,
+	})
```
StephenButtolph, line 865:
```suggestion
	s.subnetManagerCache.Put(subnetID, manager)
```

vms/platformvm/state/state.go
```diff
@@ -2212,6 +2278,29 @@ func (s *state) writeSubnetOwners() error {
...
+
+		managerBytes, err := block.GenesisCodec.Marshal(block.CodecVersion, &manager)
+		if err != nil {
+			return fmt.Errorf("failed to marshal subnet manager: %w", err)
+		}
+
+		s.subnetManagerCache.Put(subnetID, chainIDAndAddr{
+			ChainID: manager.ChainID,
+			Addr:    manager.Addr,
+		})
```
StephenButtolph, line 2295:
```suggestion
		s.subnetManagerCache.Put(subnetID, manager)
```

vms/platformvm/config/execution_config.go
dhrubabasu: was mostly taking the queue from `fx-owner-cache-size`, updated to `subnet-manager-cache-size`

vms/platformvm/state/diff.go
dhrubabasu: It wasn't 😓... fixed

vms/platformvm/state/diff.go
dhrubabasu: Yes, added a UT

vms/platformvm/state/state.go
dhrubabasu: My thought was to cache the `{ids.Empty, []byte{}}` as the subnet manager when we first do `GetSubnetManager` on a newly created Subnet

vms/platformvm/state/state.go
dhrubabasu: If we call `GetSubnetManager` on existing Subnets, we'll get an `ErrNotFound`. It's not that these are errors, it's that these Subnets don't have managers. Returning `{ids.Empty, []byte{}}` is the empty Subnet manager which is expected until `SetSubnetValidatorManagerTx` is called.

vms/platformvm/state/state.go
StephenButtolph: This writes to the database though - no?

vms/platformvm/state/state.go
StephenButtolph: I think that this is an unusual place to specify that "No manager is treated as `{ids.Empty, []byte{}}`" is defined in the state package rather than the execution logic.

dhrubabasu: Fair- will handle in a followup PR for the execution logic

vms/platformvm/state/state_test.go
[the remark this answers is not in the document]
joshua-kim: i agree w/ @marun 's comment - it should be trivial to add test tables for this new test we're adding (we don't have to do this on the whole package). Also I don't think we're testing diff application + whether or not the writes are actually occurring to state + the missing parent state case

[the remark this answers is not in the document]
dhrubabasu: The diff application is tested in diff_test.go (`TestDiffSubnetManager`)

vms/platformvm/state/state.go
```diff
@@ -339,6 +343,10 @@ type state struct {
 	subnetOwnerCache cache.Cacher[ids.ID, fxOwnerAndSize] // cache of subnetID -> owner; if the entry is nil, it is not in the database
 	subnetOwnerDB    database.Database
 
+	subnetManagers     map[ids.ID]chainIDAndAddr            // map of subnetID -> manager of the subnet
+	subnetManagerCache cache.Cacher[ids.ID, chainIDAndAddr] // cache of subnetID -> manager; if the entry is nil, it is not in the database
```
StephenButtolph, line 347: The entry can't be `nil` as it is a struct - no?

vms/platformvm/state/state.go
```diff
@@ -547,6 +560,18 @@ func newState(
 		return nil, err
 	}
 
+	subnetManagerDB := prefixdb.New(SubnetManagerPrefix, baseDB)
+	subnetManagerCache, err := metercacher.New[ids.ID, chainIDAndAddr](
+		"subnet_manager_cache",
+		metricsReg,
+		cache.NewSizedLRU[ids.ID, chainIDAndAddr](execCfg.SubnetManagerCacheSize, func(_ ids.ID, f chainIDAndAddr) int {
+			return ids.IDLen + len(f.Addr)
```
StephenButtolph, line 568: We typically account for both the key size and the value size.
```suggestion
			return 2*ids.IDLen + len(f.Addr)
```

vms/platformvm/state/state.go
dhrubabasu: Good catch, fixed the comment

vms/platformvm/state/state.go
```diff
@@ -339,6 +343,10 @@ type state struct {
 	subnetOwnerCache cache.Cacher[ids.ID, fxOwnerAndSize] // cache of subnetID -> owner; if the entry is nil, it is not in the database
 	subnetOwnerDB    database.Database
 
+	subnetManagers     map[ids.ID]chainIDAndAddr            // map of subnetID -> manager of the subnet
+	subnetManagerCache cache.Cacher[ids.ID, chainIDAndAddr] // cache of subnetID -> manager; if the entry is {ids.Empty, nil}, it is not in the database
```
StephenButtolph, line 347: We don't cache the entry if it is not in the database.

vms/platformvm/state/state_test.go
StephenButtolph, line 1404: 1. One thing we could test here is the cache usage.
2. Another interesting edge case to consider is: "what is the maximum Address size?"

dhrubabasu: Added a cache test.

For the maximum address size, I think bounding it at `4096` is more than enough. [ML-DSA-87](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.ipd.pdf) (previously known as DILITHIUM) has a public key size of 2592. I don't think this should be bounded on the `state` interface and should instead be bounded on the `ConvertSubnetTx` transaction.

vms/platformvm/state/diff_test.go
```diff
@@ -580,6 +580,55 @@ func TestDiffSubnetOwner(t *testing.T) {
 	require.Equal(owner2, owner)
 }
 
+func TestDiffSubnetManager(t *testing.T) {
```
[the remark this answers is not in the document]
dhrubabasu: This test is testing one thing: is the diff and state correct before the diff is applied and after

table tests can be used for the state test. did it here: cda1001

vms/platformvm/state/state.go
```diff
@@ -338,6 +342,10 @@ type state struct {
 	subnetOwnerCache cache.Cacher[ids.ID, fxOwnerAndSize] // cache of subnetID -> owner; if the entry is nil, it is not in the database
 	subnetOwnerDB    database.Database
 
+	subnetManagers     map[ids.ID]chainIDAndAddr            // map of subnetID -> manager of the subnet
+	subnetManagerCache cache.Cacher[ids.ID, chainIDAndAddr] // cache of subnetID -> manager
```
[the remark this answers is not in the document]
dhrubabasu: The `map` is populated with any new subnet managers that are added during node operation. At shutdown, this in-memory map is flushed to the database. At startup, the `map` is not populated with all the database entries.

This is where the cache comes in. If `GetSubnetManager` is called, it does a db read. If it's a hit, it is stored in the `subnetManagerCache` to hopefully save on db reads in the future.

vms/platformvm/config/execution_config.go
```diff
@@ -21,6 +21,7 @@ var DefaultExecutionConfig = ExecutionConfig{
 	ChainDBCacheSize:             2048,
 	BlockIDCacheSize:             8192,
 	FxOwnerCacheSize:             4 * units.MiB,
+	SubnetManagerCacheSize:       2048,
```
StephenButtolph, line 24: This is pretty confusing. All of the bytes caches use `units.MiB` to be clear that it is representing bytes. All of the other values are denominating the number of entries.

This entry is interpreted as bytes - but is set as if it is in the number of entries.

I think we should either be specifying `2 * units.KiB` or we should be setting this to a different value (if the assumption was that this is denominated in entries rather than bytes)

vms/platformvm/state/state_test.go
```diff
@@ -1401,6 +1401,56 @@ func TestStateSubnetOwner(t *testing.T) {
...
+	}{
+		{
+			name: "in-memory",
+			setup: func(s State, subnetID ids.ID, chainID ids.ID, addr []byte) {
+				s.SetSubnetManager(subnetID, chainID, addr)
+			},
+		},
+		{
+			name: "cache",
+			setup: func(s State, subnetID ids.ID, chainID ids.ID, addr []byte) {
```
StephenButtolph, line 1417: Should this pass in the `*testing.T` or the `*require.Assertions`? Right now this function is executed by the inner test but uses the outer testing struct.

StephenButtolph (approved):

PR #3129, opened 2024-06-19
Emit version in JSON format for --json-version

marun:
## Why this should be merged

Simplifies determination of the version of avalanchego by providing machine-readable version output when `--version-json` is specified. This is intended to allow a test controller to easily determine the versions for a given image published with the `latest` tag. 

## How this works

 - Create a JSON equivalent to the version string
 - Output JSON version string in response to `--version-json`

Alternate additive-only implementation: https://github.com/ava-labs/avalanchego/compare/json-version-additive?expand=1

## How this was tested

Locally confirmed output consistency:

`--version`
```
avalanchego/1.11.8 [database=v1.4.5, rpcchainvm=35, commit=eff4dc65a9073554000edace407a181df40c2857, go=1.21.11]
```

`--version-json`
```json
{
  "app": "avalanchego/1.11.8",
  "commit": "eff4dc65a9073554000edace407a181df40c2857",
  "database": "v1.4.5",
  "go": "1.21.11",
  "rpcchainvm": 35
}
```

### DIFF
--- a/version/string.go
+++ b/version/string.go
@@ -16,8 +16,8 @@ var GitCommit string
 
 // Versions contains the versions relevant to a build of avalanchego. In
 // addition to supporting construction of the strings displayed by
-// --version and --json-version, it can be used to unmarshal the output
-// of --json-version.
+// --version and --version-json, it can be used to unmarshal the output
+// of --version-json.
 type Versions struct {
 	Application string `json:"application"`
 	Database    string `json:"database"`
--- a/version/string.go
+++ b/version/string.go
@@ -6,7 +6,6 @@ package version
 import (
 	"fmt"
 	"runtime"
-	"strconv"
 	"strings"
 )
 
@@ -20,7 +19,7 @@ var GitCommit string
 type Versions struct {
 	Application string `json:"application"`
 	Database    string `json:"database"`
-	RPCChainVM  string `json:"rpcchainvm"`
+	RPCChainVM  uint64 `json:"rpcchainvm"`
 	// Commit may be empty if GitCommit was not set at compile time
 	Commit string `json:"commit"`
 	Go     string `json:"go"`
@@ -30,7 +29,7 @@ func GetVersions() *Versions {
 	versions := &Versions{
 		Application: CurrentApp.String(),
 		Database:    CurrentDatabase.String(),
-		RPCChainVM:  strconv.FormatUint(uint64(RPCChainVMProtocol), 10),
+		RPCChainVM:  uint64(RPCChainVMProtocol),
 		Go:          strings.TrimPrefix(runtime.Version(), "go"),
 	}
 	if GitCommit != "" {
@@ -41,7 +40,7 @@ func GetVersions() *Versions {
 
 func (v *Versions) String() string {
 	// This format maintains consistency with previous --version output
-	versionString := fmt.Sprintf("%s [database=%s, rpcchainvm=%s, ", v.Application, v.Database, v.RPCChainVM)
+	versionString := fmt.Sprintf("%s [database=%s, rpcchainvm=%d, ", v.Application, v.Database, v.RPCChainVM)
 	if len(v.Commit) > 0 {
 		versionString += fmt.Sprintf("commit=%s, ", v.Commit)
 	}
--- a/version/string.go
+++ b/version/string.go
@@ -22,8 +22,9 @@ type Versions struct {
 	Application string `json:"application"`
 	Database    string `json:"database"`
 	RPCChainVM  string `json:"rpcchainvm"`
-	Commit      string `json:"commit"`
-	Go          string `json:"go"`
+	// Commit may be empty if GitCommit was not set at compile time
+	Commit string `json:"commit"`
+	Go     string `json:"go"`
 }
 
 func GetVersions() *Versions {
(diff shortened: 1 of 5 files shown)


version/string.go
```diff
@@ -4,37 +4,68 @@
...
-	args := []interface{}{
-		CurrentApp,
-		CurrentDatabase,
-		RPCChainVMProtocol,
+	// Define the ordered set of versions that are common to regular and JSON
+	// output. The order is maintained to ensure consistency with previous
+	// --version output.
+	versions := []namedVersion{
+		{name: "database", version: CurrentDatabase.String()},
+		{name: "rpcchainvm", version: strconv.FormatUint(uint64(RPCChainVMProtocol), 10)},
```
ARR4N, line 36: (nit,readability) `fmt.Sprint(RPCChainVMProtocol)` is a little clearer.

aaronbuchwald (dismissed):
LGTM

version/string.go
```diff
@@ -4,37 +4,68 @@
...
 	String string
 
+	// String is displayed when CLI arg --json-version is used
+	JSONString string
+
 	// GitCommit is set in the build script at compile time
 	GitCommit string
 )
 
+type namedVersion struct {
```
ARR4N, line 25: Given that the motivation behind this PR is to allow for programmatic access to the version details, I think it's worth exporting this type (and its fields) so other code can unmarshal the JSON into it.

If you do export it, I think it might need a different name so it's not `version.NamedVersion` when used elsewhere. Even just `version.Named`? Mandatory [style guide reference](https://google.github.io/styleguide/go/decisions#repetition) :rofl:

Alternatively, an explicit version struct would really help in such a scenario, but that undoes the nice pattern you have for text marshalling.

Consider the above as an exploration of trade-offs for you to consider, not explicit requests to change anything.

ARR4N (commented):
How catastrophic would a bug in the version string be? If so I think it might be worth having a test that (a) creates `String` with just a `fmt.Sprintf()` and explicit format + (b) unmarshals the JSON. Both of these make it really easy to spot mistakes IMO.

version/string.go
```diff
@@ -4,37 +4,54 @@
...
 
-	format += "]\n"
-	String = fmt.Sprintf(format, args...)
+func (v *Versions) JSON() string {
+	jsonBytes, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		panic(err)
+	}
+	return string(jsonBytes)
 }
```
StephenButtolph, line 58: This seems a bit non-standard. I feel like the caller should call `json.MarshalIndent` on their own terms and handle the error however they want.

version/string.go
```diff
@@ -4,37 +4,54 @@
...
+var GitCommit string
 
-	// GitCommit is set in the build script at compile time
-	GitCommit string
-)
+// Versions contains the versions relevant to a build of avalanchego. In
+// addition to supporting construction of the strings displayed by
+// --version and --json-version, it can be used to unmarshal the output
+// of --json-version.
+type Versions struct {
```
StephenButtolph, line 21: I don't really love this name... Would `version.Description` be better than `version.Versions`? I don't really love that either tbh.

version/string.go
```diff
@@ -4,37 +4,54 @@
...
+	"encoding/json"
 	"fmt"
 	"runtime"
+	"strconv"
 	"strings"
 )
 
-var (
-	// String is displayed when CLI arg --version is used
-	String string
```
StephenButtolph, line 14: Do we still want to keep this `String`? We could just export the struct as well rather than implementing a function to grab it

version/string.go
```diff
@@ -9,32 +9,40 @@ import (
...
+	versions := &Versions{
+		Application: CurrentApp.String(),
+		Database:    CurrentDatabase.String(),
+		RPCChainVM:  uint64(RPCChainVMProtocol),
+		Go:          strings.TrimPrefix(runtime.Version(), "go"),
 	}
 	if GitCommit != "" {
-		format += ", commit=%s"
-		args = append(args, GitCommit)
+		versions.Commit = GitCommit
```
StephenButtolph, line 36: The default value of `versions.Commit` is `""`. So I think we can just directly set this in the construction of the struct rather than checking if it is populated here.

version/string_test.go
```diff
@@ -0,0 +1,39 @@
...
+		RPCChainVM:  3,
+		Go:          "5",
+	}
+	expectedVersions := fmt.Sprintf(
+		"%s [database=%s, rpcchainvm=%d, go=%s]",
+		versions.Application,
+		versions.Database,
+		versions.RPCChainVM,
+		versions.Go,
+	)
```
StephenButtolph, line 26: Would this be more readable if we just specified the string rather than using `fmt.Sprintf` to create it?

version/string_test.go
```diff
@@ -0,0 +1,39 @@
...
+package version
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVersionsGetString(t *testing.T) {
```
StephenButtolph, line 13: A bit surprised this isn't done with a table test... But with only two cases... I leave it up to you on if you feel like it's worthwhile.

StephenButtolph (approved):

darioush (approved):

PR #3133, opened 2024-06-20
Rename `GetPeerList` and `PeerList` messages

### LINKED ISSUE #3128: Rename `GetPeerList` and `PeerList` messages
`get_peerlist` and `peerlist` messages often treat `peerlist` like a single word. Whereas all the other messages `get_accepted`, `pull_query`, ect, are named consistently.

We should rename `get_peerlist` to `get_peers`.
We should rename `peerlist` to `peers`.

This will also impact some CLI flags, as `peer-list` options should become `peers` options.

MukulKolpe:
Resolves #3128 

## Why this should be merged

This change ensures consistency in message naming conventions, aligning `get_peerlist` and `peerlist` with other message names such as `get_accepted` and `pull_query`. It improves readability and maintains uniformity across the codebase.

## How this works

- `get_peerlist` is renamed to `get_peers`.
- `peerlist` is renamed to `peers`.
- Corresponding CLI flags are updated from `peer-list` options to `peers` options.


message/ops.go
```diff
@@ -212,9 +212,9 @@ func (op Op) String() string {
 	case HandshakeOp:
 		return "handshake"
 	case GetPeerListOp:
-		return "get_peerlist"
+		return "get_peers"
 	case PeerListOp:
```
StephenButtolph, line 216: I think there are still a number of places in the code that reference `PeerList`s. These op types are an example, but I think a number of places in the `network` package need to be renamed as well as the `p2p` proto

StephenButtolph: @MukulKolpe were you still planning on working on this?

MukulKolpe: Yes, I'm on it.

StephenButtolph:
Closed due to stale-ness.

PR #3134, opened 2024-06-20
Repackaged NextBlockTime and GetNextStakerChangeTime

abi87:
## Why this should be merged
With dynamic fees, a bunch of `platformvm` packages will need to calculate block time in order to pick the right gas price.
In anticipation for it, this PR moves  `NextBlockTime` and its dependencies to `state` package, in order to avoid import cycles.

## How this works
Just move `NextBlockTime` and `GetNextStakerChangeTime`  from `txexecutor` package to `state` package, where they won't cause import cycles.

## How this was tested
CI (it's pure refactoring

### DIFF
--- a/vms/platformvm/txs/executor/state_changes.go
+++ b/vms/platformvm/txs/executor/state_changes.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/constants"
-	"github.com/ava-labs/avalanchego/utils/timer/mockable"
 	"github.com/ava-labs/avalanchego/vms/platformvm/reward"
 	"github.com/ava-labs/avalanchego/vms/platformvm/state"
 	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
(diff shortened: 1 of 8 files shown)


vms/platformvm/txs/executor/state_changes.go
```diff
@@ -58,30 +57,6 @@ func VerifyNewChainTime(
 	return nil
 }
 
-func NextBlockTime(state state.Chain, clk *mockable.Clock) (time.Time, bool, error) {
```
abi87, line 61: `NextBlockTime` is the function we need to reuse e.g. in txBuilder package, when dynamic fees are activated.
We move it to state package to avoid circular dependencies and unlock compilation

aaronbuchwald (approved):
Refactor LGTM.

I don't see any similar code currently in the state package, so this may be a new outlier. Not sure if there's any other place to move it.

abi87:
> Refactor LGTM.
> 
> I don't see any similar code currently in the state package, so this may be a new outlier. Not sure if there's any other place to move it.

I agree with you. An alternative would be to introduce a `chainTime` package containing anything related to chain time (not just the functions I repackaged). I picked this change since it's minimal, but definitely open to revisit the choice down the line

dhrubabasu (approved):

PR #3135, opened 2024-06-20
[chains/atomic] Remove a nested if statement

dhrubabasu:
## Why this should be merged

Readability

## How this works

Move nested if down

## How this was tested

Existing UTs

### DIFF
--- a/chains/atomic/state.go
+++ b/chains/atomic/state.go
@@ -147,12 +147,7 @@ func (s *state) SetValue(e *Element) error {
 // current engine state.
 func (s *state) RemoveValue(key []byte) error {
 	value, err := s.loadValue(key)
-	if err != nil {
-		if err != database.ErrNotFound {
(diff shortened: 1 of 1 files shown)


chains/atomic/state.go
```diff
@@ -161,6 +156,10 @@ func (s *state) RemoveValue(key []byte) error {
 		}
 		return s.valueDB.Put(key, valueBytes)
 	}
+	if err != nil {
+		// An unexpected error occurred, so we should propagate that error
```
ARR4N, line 160: I think the comment is now redundant.

ARR4N (approved):

chains/atomic/state.go
dhrubabasu, line 160:
```suggestion
```

StephenButtolph (approved):

PR #3137, opened 2024-06-20
[vms/platformvm] Cleanup execution config tests

dhrubabasu:
## Why this should be merged

@marun had a good suggestion that this test should just do a round-trip for maintainability and clarity

## How this works

struct1 -> marshal to json -> unmarshall to struct2, call Equal on the 2 structs

## How this was tested

CI

### DIFF
--- a/vms/platformvm/config/execution_config_test.go
+++ b/vms/platformvm/config/execution_config_test.go
@@ -4,6 +4,10 @@
 package config
 
 import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


vms/platformvm/config/execution_config_test.go
```diff
@@ -105,64 +102,14 @@ func TestExecutionConfigUnmarshal(t *testing.T) {
 			ChecksumsEnabled:             true,
 			MempoolPruneFrequency:        time.Minute,
 		}
-		require.Equal(expected, ec)
-	})
+		require.NoError(VerifyInitializedStruct(*expected))
+		require.NoError(VerifyInitializedStruct(expected.Network))
 
-	t.Run("default values applied correctly", func(t *testing.T) {
```
dhrubabasu, line 111: This test is not really doing anything more than the "mix default and extracted values from json" test earlier

vms/platformvm/config/execution_config_test.go
```diff
@@ -12,6 +16,30 @@ import (
 	"github.com/ava-labs/avalanchego/vms/platformvm/network"
 )
 
+// Errors if all values in a struct are not initialized
+func VerifyInitializedStruct(s interface{}) error {
```
ARR4N, line 20: (optional) Since this is in a test, you can accept `testing.TB` as a first parameter and just call `tb.Fatalf()` instead of returning errors. If you call `tb.Helper()` at the beginning of the function then error reports will have the line of the call site not of the `Fatal()`.

I marked as optional because the gain here is limited, but if you find yourself regularly wrapping the call in `require.NoError()` then it cleans all of those up.

```go
func VerifyInitializedStruct(tb testing.TB, s interface{}) {
    tb.Helper()
```

dhrubabasu: I totally blanked on this pattern when putting together the PR. Thanks for the reminder! Updated

vms/platformvm/config/execution_config_test.go
```diff
@@ -12,6 +14,23 @@ import (
 	"github.com/ava-labs/avalanchego/vms/platformvm/network"
 )
 
+// Errors if all values in a struct are not initialized
```
StephenButtolph, line 17:
```suggestion
// Requires all values in a struct to be initialized
```

StephenButtolph (approved):

ARR4N (approved):

PR #3139, opened 2024-06-20
Simplify dependency registration

StephenButtolph:
## Why this should be merged

This PR removes complex handling to try to track whether a block was issued into consensus to determine if we can register a dependency on the block or not. To replace this handling, we simply look to see if the block was either requested or is pending to determine if the block can be a dependency.

## How this works

Dependencies are resolved when:
1. There was a request over the network for the dependency and that either succeeded or failed
2. There was an issuer job with this dependency that was itself blocking on a different dependency.

## How this was tested

- [x] Ran on Fuji
- [x] Verified that this approach avoids the error introduced by #3131 
- [X] Existing CI

### DIFF
--- a/snow/engine/snowman/transitive.go
+++ b/snow/engine/snowman/transitive.go
@@ -734,15 +734,14 @@ func (t *Transitive) issueFrom(
 	// issue [blk] and its ancestors to consensus.
 	blkID := blk.ID()
 	for !t.wasIssued(blk) {
-		if err := t.issue(ctx, nodeID, blk, false, issuedMetric); err != nil {
+		err := t.issue(ctx, nodeID, blk, false, issuedMetric)
+		if err != nil {
 			return err
 		}
 
+		// If we don't have this ancestor, request it from [nodeID]
 		blkID = blk.Parent()
-		var err error
 		blk, err = t.getBlock(ctx, blkID)
-
-		// If we don't have this ancestor, request it from [vdr]
 		if err != nil {
 			t.sendRequest(ctx, nodeID, blkID, issuedMetric)
 			return nil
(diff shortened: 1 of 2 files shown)


snow/engine/snowman/transitive.go
```diff
@@ -1063,10 +1045,6 @@ func (t *Transitive) pendingContains(blkID ids.ID) bool {
 	return ok
 }
 
-func (t *Transitive) removeFromPending(blk snowman.Block) {
-	delete(t.pending, blk.ID())
-}
```
StephenButtolph, line 1068: 1 line functions are annoying imo. Also I have no clue why this ever took in the full block rather than just the ID.

dhrubabasu (approved):

abi87 (approved):

snow/engine/snowman/transitive.go
```diff
@@ -1176,6 +1158,13 @@ func (t *Transitive) getProcessingAncestor(ctx context.Context, initialVote ids.
 	}
 }
 
+// canDependOn reports true if it is guaranteed for the provided block ID to
+// eventually either be fulfilled or abandoned.
+func (t *Transitive) canDependOn(blkID ids.ID) bool {
```
patrick-ogrady, line 1163: This is probably a very dumb question but can you expand on why we shouldn't return that we can depend on blkIDs that are decided?

patrick-ogrady: If there is an invariant that it must be checked prior to calling this function that the block should not be decided, it may be worth adding a comment.

StephenButtolph: > // canDependOn reports true if it is guaranteed for the provided block ID to
// eventually either be fulfilled or abandoned.

Decided blocks are not guaranteed to be either fulfilled or abandoned. So this function returns `false`. It is definitely possible that decided blockIDs are passed into this function.

PR #3141, opened 2024-06-20
Disallow rejected blocks from `GetBlock`

StephenButtolph:


StephenButtolph:
This PR is going to be too complicated.

PR #3143, opened 2024-06-22
ACP-113 : Provable Virtual Machine Randomness

tsachiherman:
## Why this should be merged

This is the implementatinon of ACP-113, as described [here](https://github.com/avalanche-foundation/ACPs/pull/113).

## How this works

(taken from the design doc)
Existing avalanche protocol breaks the block building into two parts : external and internal. The external block is the SnowMan++ block, whereas the internal block is the actual virtual machine block.

To support randomness, a BLS based VRF implementation is used, that would be recursively signing its own signatures as its message. Since the BLS signatures are deterministic, they provide a great way to construct a reliable VRF.

For proposers that do not have a BLS key associated with their node, the hash of the signature from the previous round.

In order to bootstrap the signatures chain, a missing signature would be replaced with a byte slice that is the hash product of a verifiable and trustable seed.

The changes proposed here would affect the way a block is being validated. Therefore, when this change gets implemented, it needs to be deployed as a mandatory upgrade.

## How this was tested

1. Existing tests were extended to cover new code.
2. Manual testing against fuji.
3. Passed all existing unit tests.


vms/proposervm/block.go
```diff
@@ -221,10 +226,27 @@ func (p *postForkCommonComponents) buildChild(
...
+		// in this case, we can't sign with BLS key, since we're not going to include the Certificate, which is required
+		// for the signature validation. Instead, we'll just hash the previous
+		childBlockVrfSig = block.NextHashBlockSignature(parentBlockSig)
+	}
+
 	var innerBlock snowman.Block
 	if p.vm.blockBuilderVM != nil {
 		innerBlock, err = p.vm.blockBuilderVM.BuildBlockWithContext(ctx, &smblock.Context{
 			PChainHeight: parentPChainHeight,
+			// todo : add block.CalculateVRFOut(childBlockVrfSig)
```
tsachiherman, line 249: @StephenButtolph am I on the right track here ?

ARR4N (changes requested):
I think this needs some integration tests, specifically some with failing verification.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,140 @@ package block
...
+	// +-------------------------+----------+------------+
+	// |  vrfSig :               | [48]byte |  48 bytes  |
+	// +-------------------------+----------+------------+
+	if len(vrfSig) != 48 {
+		return nil
+	}
+	buffer := make([]byte, 8+48)
+	copy(buffer, "rng-derv")
+	copy(buffer[8:], vrfSig[:])
+	outHash := hashing.Hash256(buffer)
```
StephenButtolph, line 57: This is attempting to cast `buffer` into the `hashing.Hash256` type (`[32]byte`) and then slicing the result. This isn't actually computing a hash.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,140 @@ package block
...
 
 	return block, block.initialize(bytes)
 }
 
+func CalculateVRFOut(vrfSig []byte) []byte {
+	// build the hash of the following struct:
+	// +-------------------------+----------+------------+
+	// |  prefix :               | [8]byte  | "rng-derv" |
+	// +-------------------------+----------+------------+
+	// |  vrfSig :               | [48]byte |  48 bytes  |
```
StephenButtolph, line 49: Compressed BLS signatures are `96` bytes for us.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,140 @@ package block
...
+	// +-------------------------+----------+------------+
+	if len(vrfSig) != 48 {
```
StephenButtolph, line 51:
```suggestion
	if len(vrfSig) != bls.SignatureLen {
```

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,140 @@ package block
...
+		return nil
+	}
+	buffer := make([]byte, 8+48)
+	copy(buffer, "rng-derv")
```
StephenButtolph, line 55: Should probably make this a constant

vms/proposervm/block/build.go
```diff
@@ -6,75 +6,169 @@ package block
...
+}
+
+func nextHashBlockSignature(parentBlockSig []byte) []byte {
+	if len(parentBlockSig) == 0 {
+		return nil
+	}
+	// previous block had a valid signature, hash that signature.
+	sigParentBlockSig := hashing.ComputeHash256(parentBlockSig)
+
+	// as long as the signature length is too short, generate additional hashes.
```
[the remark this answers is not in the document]
StephenButtolph: I'm a bit confused as to why we even need to pad this at all. Wouldn't it make more sense to just have it be the 32-byte slice? We already encode the length of the signature in the block, I'm not sure why we care to ensure it is always the same length.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,140 @@ package block
...
+			// no parent block signature.
+			return []byte{}
+		}
+
+		return NextHashBlockSignature(parentBlockSig)
+	}
+
+	// we have bls key
+	var signMsg []byte
+	if parentBlockSig == nil {
```
StephenButtolph, line 133: I don't like comparing `[]byte` types with `nil`. IMO `[]byte{}` and `[]byte(nil)` should be handled the same unless we have a very good reason not to.

vms/proposervm/block/build.go
```diff
@@ -55,26 +157,27 @@ func Build(
...
-	// signature, which is prefixed by a uint32. Because we are marshalling the
-	// block with an empty signature, we only need to strip off the length
-	// prefix to get the unsigned bytes.
-	lenUnsignedBytes := len(unsignedBytesWithEmptySignature) - wrappers.IntLen
-	unsignedBytes := unsignedBytesWithEmptySignature[:lenUnsignedBytes]
-	block.id = hashing.ComputeHash256Array(unsignedBytes)
+	// calculate the block ID.
+	if err = block.initializeID(); err != nil {
+		return nil, err
+	}
```
StephenButtolph, line 178: If this is the whole reason why we created this `initializeID` method, I think it would make more sense as a standalone function that is `calculateID` and takes in the `bytes []byte` directly.

vms/proposervm/block/block.go
```diff
@@ -4,29 +4,36 @@
...
+	errFailedToParseVRFSignature  = errors.New("failed to parse VRF signature")
 )
 
 type Block interface {
 	ID() ids.ID
 	ParentID() ids.ID
 	Block() []byte
 	Bytes() []byte
+	// VerifySignature validates the correctness of the VRF signature.
+	VerifySignature(*bls.PublicKey, []byte, ids.ID, uint32) bool
```
StephenButtolph, line 34: Does it make sense to actually add this to the block interface? Or would it make more sense to just use the VRF signature on the `SignedBlock` interface and perform the verification then.

vms/proposervm/config.go
```diff
@@ -37,8 +38,18 @@ type Config struct {
 
 	// Registerer for prometheus metrics
 	Registerer prometheus.Registerer
+
+	// The BLS key associated with the node; used for signing chained signature in order to provide provable randomness.
+	StakingBLSKey *bls.SecretKey
+
+	// VRFSig activation time
+	VRFSigTime time.Time
```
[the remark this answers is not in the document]
StephenButtolph: The config is specified externally (in the node package)

vms/proposervm/post_fork_block.go
```diff
@@ -155,12 +196,14 @@ func (b *postForkBlock) verifyPostForkOption(ctx context.Context, child *postFor
 }
 
 // Return the child (a *postForkBlock) of this block
-func (b *postForkBlock) buildChild(ctx context.Context) (Block, error) {
+func (b *postForkBlock) buildChild(ctx context.Context, blsSignKey *bls.SecretKey) (Block, error) {
```
StephenButtolph, line 199: Isn't the bls key already accessible through the postForkBlock's reference to the VM?

vms/proposervm/post_fork_option.go
```diff
@@ -122,6 +123,8 @@ func (b *postForkOption) buildChild(ctx context.Context) (Block, error) {
 		parentID,
 		b.Timestamp(),
 		parentPChainHeight,
+		[]byte{}, // TODO : verify me.
```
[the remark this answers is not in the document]
tsachiherman: sorry for not being more clear here. @StephenButtolph - do we need to have a VRF Sig on this block type ?

vms/proposervm/post_fork_block.go
```diff
@@ -139,6 +145,41 @@ func (b *postForkBlock) verifyPostForkChild(ctx context.Context, child *postFork
...
+		}
+		// verify that the VRFSig was generated correctly.
+		// ( this works for both signed and unsigned blocks ).
+		if !child.SignedBlock.VerifySignature(proposerPublicKey, b.SignedBlock.VRFSig(), b.vm.ctx.ChainID, b.vm.ctx.NetworkID) {
+			return errInvalidVRFSignature
+		}
+	} else if len(child.SignedBlock.VRFSig()) != 0 {
+		// !b.vm.Config.IsVRFSigActivated(child.Timestamp())
+		// if the VRFSig has yet to be activated, verify that the VRF signature isn't there.
+		return errInvalidVRFSignature
```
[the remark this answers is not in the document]
tsachiherman: yes, I agree. it's more of a "malicious actor detected".

vms/proposervm/block/block.go
tsachiherman: given that it's being used only on signed blocks, I'll move it to the `SignedBlock` interface. thanks!

vms/proposervm/block/block.go
```diff
@@ -77,15 +88,31 @@ func (b *statelessBlock) Bytes() []byte {
...
 
+func (b *statelessBlock) initializeID() error {
+	var unsignedBytes []byte
 	// The serialized form of the block is the unsignedBytes followed by the
 	// signature, which is prefixed by a uint32. So, we need to strip off the
 	// signature as well as it's length prefix to get the unsigned bytes.
-	lenUnsignedBytes := len(bytes) - wrappers.IntLen - len(b.Signature)
-	unsignedBytes := bytes[:lenUnsignedBytes]
+	lenUnsignedBytes := len(b.bytes) - wrappers.IntLen - len(b.Signature)
+	if lenUnsignedBytes <= 0 {
```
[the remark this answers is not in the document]
tsachiherman: exactly. just handling the missing ( theoretical ) codepath.

vms/proposervm/block/build.go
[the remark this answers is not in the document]
tsachiherman: I was thinking about that too. The question that I tried to answer is whether an observer should have the ability to distinguish between a hash pseudo signature block vs a genuine signature.
Just padding the 32 bytes so it would fit into a 96 byte slice, doesn't give us any advantage.

vms/proposervm/block/build.go
tsachiherman: good point. I'll use `len(..)` to cover all basis.

vms/proposervm/block/block.go
```diff
@@ -48,69 +69,168 @@ type statelessUnsignedBlock struct {
...
 	if err != nil {
 		return fmt.Errorf("%w: %w", errInvalidCertificate, err)
 	}
 
 	b.proposer = ids.NodeIDFromCert(b.cert)
+
+	if vrfBytes := b.VRFSig(); len(vrfBytes) > 0 {
+		b.vrfSig, err = bls.SignatureFromBytes(vrfBytes)
+		if err != nil {
+			return fmt.Errorf("%w: %w", errFailedToParseVRFSignature, err)
```
ARR4N, line 218: `b.vrfSig` may have been corrupted when the error occurred. Perhaps I'm being pedantic but I think it's worth setting it to `nil` in the error branch.

vms/proposervm/block/build.go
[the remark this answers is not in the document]
ARR4N: An observer will presumably have [access to all other relevant information](https://en.wikipedia.org/wiki/Kerckhoffs%27s_principle), including the fact that the proposer doesn't have a BLS key.

vms/proposervm/block/block.go
tsachiherman: I don't think that you're pedantic; I looked into that myself and confirmed that is already being guaranteed by `SignatureFromBytes`.
I really like when struct function like this are atomic ( i.e. function doesn't change any of the struct field in case of an error). In this case, however, I don't believe it would have a notable value.
If you'd like to see this as an atomic function, please let me know and I will update it accordingly.

vms/proposervm/block/block.go
```diff
@@ -48,69 +69,168 @@ type statelessUnsignedBlock struct {
 	PChainHeight uint64 `serialize:"true"`
 	Certificate  []byte `serialize:"true"`
 	Block        []byte `serialize:"true"`
+	VRFSig       []byte `serialize:"true"`
+}
+
+type unsignedBlockTypes interface {
+	interface{ statelessUnsignedBlockV0 } | interface{ statelessUnsignedBlock }
```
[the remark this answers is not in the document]
tsachiherman: good point. I thought that the "|" would only work for aggregarion of interfaces, but it seems to be working for struct we well. thanks!

vms/proposervm/config.go
```diff
@@ -37,8 +38,18 @@ type Config struct {
...
+
+	// VRFSig activation time
+	VRFSigActivationTime time.Time
 }
 
 func (c *Config) IsDurangoActivated(timestamp time.Time) bool {
 	return !timestamp.Before(c.DurangoTime)
 }
+
+func (c *Config) IsVRFSigActivated(timestamp time.Time) bool {
```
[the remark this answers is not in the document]
tsachiherman: I think that it's going to change anyway to align with the official codename for this release. I've heard "Etna" coming up, but unsure if that's going to stick or not.

vms/proposervm/post_fork_block.go
```diff
@@ -8,9 +8,11 @@ import (
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
 	"github.com/ava-labs/avalanchego/vms/proposervm/block"
 )
 
+// compile time check to ensure that postForkBlock implements PostForkBlock
```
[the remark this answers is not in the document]
tsachiherman: When I read that code, it wasn't obvious to me.. which is why I wrote it once I figured it out. But I don't have any strong feeling about these.

vms/proposervm/block/block_test.go
```diff
@@ -39,8 +39,9 @@ func TestBlockSizeLimit(t *testing.T) {
 	timestamp := time.Unix(123, 0)
```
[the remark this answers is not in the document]
tsachiherman: In the context of this change, I don't believe it would add value to explicitly tests both block types with varying inner block sizes.

chains/manager.go
```diff
@@ -794,14 +794,16 @@ func (m *manager) createAvalancheChain(
...
+			ActivationTime:         m.ApricotPhase4Time,
+			DurangoTime:            version.GetDurangoTime(m.NetworkID),
+			MinimumPChainHeight:    m.ApricotPhase4MinPChainHeight,
+			MinBlkDelay:            minBlockDelay,
+			NumHistoricalBlocks:    numHistoricalBlocks,
+			StakingLeafSigner:      m.StakingTLSSigner,
+			StakingCertLeaf:        m.StakingTLSCert,
+			Registerer:             proposervmReg,
+			StakingBLSKey:          m.StakingBLSKey,
+			EUpgradeActivationTime: version.GetEUpgradeTime(m.NetworkID),
```
dhrubabasu, line 806: nit:
```suggestion
			EUpgradeTime: version.GetEUpgradeTime(m.NetworkID),
```

vms/proposervm/block_test.go
```diff
@@ -92,12 +92,15 @@ func TestPostForkCommonComponents_buildChild(t *testing.T) {
...
 
+	emptyParentBlockSig := []byte{}
+
 	// Should call BuildBlockWithContext since proposervm is activated
 	gotChild, err := blk.buildChild(
 		context.Background(),
 		parentID,
 		parentTimestamp,
 		pChainHeight-1,
+		emptyParentBlockSig,
```
dhrubabasu, line 103: style (optional): we do this pattern elsewhere in the codebase for values we hardcode into function params:

```suggestion
	// Should call BuildBlockWithContext since proposervm is activated
	gotChild, err := blk.buildChild(
		context.Background(),
		parentID,
		parentTimestamp,
		pChainHeight-1,
		[]byte{} /*=parentBlockVRFSig*/,
```

vms/proposervm/block/block.go
```diff
@@ -40,6 +45,20 @@ type SignedBlock interface {
...
+	VerifySignature(*bls.PublicKey, []byte, ids.ID, uint32) bool
+}
+
+// statelessUnsignedBlockV0 is the "old" version of statelessUnsignedBlock, which doesn't contain the VRF signature.
+type statelessUnsignedBlockV0 struct {
+	ParentID     ids.ID `serialize:"true"`
+	Timestamp    int64  `serialize:"true"`
+	PChainHeight uint64 `serialize:"true"`
+	Certificate  []byte `serialize:"true"`
+	Block        []byte `serialize:"true"`
```
dhrubabasu, line 61: I think a better pattern would be to bump the codec version instead of two separate structs

We do this in the validator metadata here: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/metadata_codec.go

and how it is used is here: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/metadata_validator.go#L31-L40

where the `StakerStartTime` is only serialized if codec version 1 is used

dhrubabasu: This would also remove the `panic` calls later on in the generic version

vms/proposervm/block/block.go
```diff
@@ -48,69 +67,171 @@ type statelessUnsignedBlock struct {
...
 
-func (b *statelessBlock) Bytes() []byte {
+func (b *statelessBlock[T]) innerTimestamp() int64 {
+	switch sb := any(b.StatelessBlock).(type) {
+	case statelessUnsignedBlock:
+		return sb.Timestamp
+	case statelessUnsignedBlockV0:
+		return sb.Timestamp
+	default:
+		panic(errUnexpectedStatelessUnsignedBlock)
```
abi87, line 151: q: this function seems to be called only in functions returning error. Why not returning the error instead of panicking?
I understand you need to panic in exported methods, but for the unexported ones we may go a different route.
Same for `innerCertificate`

vms/proposervm/vm_test.go
```diff
@@ -478,15 +478,21 @@ func TestCoreBlockFailureCauseProposerBlockParseFailure(t *testing.T) {
 		return nil, errMarshallingFailed
 	}
 
+	emptyParentBlockSig := []byte{}
```
abi87, line 481: nit: should this be a package level variable, to avoid reinstantiating it in different tests?

vms/rpcchainvm/vm_server.go
```diff
@@ -387,9 +388,19 @@ func (vm *VMServer) BuildBlock(ctx context.Context, req *vmpb.BuildBlockRequest)
...
-		})
+		}
+		switch len(req.RandomnessSeed) {
+		case 0:
+			blkContext.RandomnessSeed = [hashing.HashLen]byte{}
+		case hashing.HashLen:
+			blkContext.RandomnessSeed = [hashing.HashLen]byte(req.RandomnessSeed)
+		default:
+			// invalid input.
+			return nil, errors.New("build block request contained invalid randomness seed length")
```
abi87, line 401: nit: usually we define a package level error instead of using rogue errors.New. It would help testing, as it allow asserting a specific error

vms/proposervm/post_fork_block.go
```diff
@@ -128,6 +136,41 @@ func (b *postForkBlock) verifyPostForkChild(ctx context.Context, child *postFork
...
+	// if there is a known proposer. ( i.e. block was signed )
+	if childProposer := child.SignedBlock.Proposer(); childProposer != ids.EmptyNodeID {
+		// get the validators set.
+		validatorSet, err := child.vm.ctx.ValidatorState.GetValidatorSet(ctx, parentPChainHeight, b.vm.ctx.SubnetID)
+		if err != nil {
+			return err
+		}
+
+		// get the validator for this proposer
+		childValidator := validatorSet[childProposer]
```
abi87, line 159: nit: this is fine, but I believe usually we use the boolean to check whether a key is present in a map, rather than checking whether the value is not nil

vms/proposervm/block/block.go
```diff
@@ -40,6 +45,20 @@ type SignedBlock interface {
 	// Proposer returns the ID of the node that proposed this block. If no node
 	// signed this block, [ids.EmptyNodeID] will be returned.
 	Proposer() ids.NodeID
+
+	VRFSig() []byte
+
+	// VerifySignature validates the correctness of the VRF signature.
+	VerifySignature(*bls.PublicKey, []byte, ids.ID, uint32) bool
```
abi87, line 52: q: should we consider passing `ctx *snow.Context` rather then chainID and networkID separately? Just to slim down the signature.
I don't have a strong opinion about this, just raising the point

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,152 @@ package block
...
+	copy(buffer, vrfOutPrefix)
+	copy(buffer[len(vrfOutPrefix):], vrfSig)
+	return hashing.ComputeHash256Array(buffer)
+}
+
+func initializeID(bytes []byte, signature []byte) (ids.ID, error) {
+	var unsignedBytes []byte
+	// The serialized form of the block is the unsignedBytes followed by the
+	// signature, which is prefixed by a uint32. So, we need to strip off the
+	// signature as well as it's length prefix to get the unsigned bytes.
```
abi87, line 72: typo: its rather than it's?

abi87 (commented):
small nits. Will do a second pass to look deeper at the whole logic. Nice work so far!

chains/manager.go
tsachiherman: done.

vms/proposervm/block/block.go
tsachiherman: The `snow` package is currently not a dependency of the `block` package.
Do you forsee any future parameters that would be derived off `snow.Context` ? 
I can see that the key, networkID and chainID are pretty much "constants", whereas the `signature` is varying. maybe we should bundle these three up and store them in the VM ? 
```golang
type signatureVerificationContext struct {
    blsKey *bls.PublicKey
    chainID ids.ID
    networkID uint32
}
```

vms/proposervm/block/block.go
tsachiherman: thats very interesting. I wasn't aware of the conditional serialization. I will take another attempt with that.

vms/proposervm/block/block.go
tsachiherman: these panics would never be invoked ( which is why I used panic.. ).
The underlying datatype must be either `statelessUnsignedBlock` or `statelessUnsignedBlockV0`, and this is validated during compile time by `unsignedBlockTypes`.
but it might not be needed if @dhrubabasu 's approach would work.

vms/proposervm/block/build.go
tsachiherman: fixed.

vms/proposervm/block_test.go
tsachiherman: If you'll scroll above, you'll see @marun 's comment on doing exactly the opposite:

> (No action required) Maybe use a constant (e.g. emptyParentBlockSig) instead of an empty value with a comment to declare your intent?

I'm quite neutral on which one to pick. Ideally, I'd like to be consistent.

vms/proposervm/vm_test.go
tsachiherman: done.

vms/rpcchainvm/vm_server.go
tsachiherman: done

vms/proposervm/post_fork_block.go
tsachiherman: let me do both, just to be on the safe side ?

vms/proposervm/block/block.go
danlaine, line 52: nit: might be good to name these parameters so we don't need to look at an implementation to see what they are.

i.e.
```
(pk *bls.PublicKey, parentVRFSig []byte, chainID ids.ID, networkID uint32)
```

vms/proposervm/block.go
```diff
@@ -38,6 +38,8 @@ var (
 	errProposerMismatch         = errors.New("proposer mismatch")
 	errProposersNotActivated    = errors.New("proposers haven't been activated yet")
 	errPChainHeightTooLow       = errors.New("block P-chain height is too low")
+	errInvalidVRFSignature      = errors.New("invalid VRF signature")
+	errVRFSignaturePresents     = errors.New("unexpected VRF signature presents")
```
danlaine, line 42: wording nit. I think just changing `presents` to `present` would be ok too.
```suggestion
	errUnexpectedVRFSignature     = errors.New("unexpected VRF signature present")
```

vms/proposervm/block/block.go
```diff
@@ -48,69 +67,171 @@ type statelessUnsignedBlock struct {
 	PChainHeight uint64 `serialize:"true"`
 	Certificate  []byte `serialize:"true"`
 	Block        []byte `serialize:"true"`
+	VRFSig       []byte `serialize:"true"`
+}
+
+type unsignedBlockTypes interface {
```
danlaine, line 73: See Dhruba's point above about the codec's versioning. Also, I think we could instead declare this interface as:

```
type unsignedBlockTypes interface {
	ParentID() ids.ID
	Block() []byte
	VRFSig() []byte
	PChainHeight() uint64
	innerCertificate() []byte
	innerTimestamp() int64
}
```

And then implement that interface for `statelessUnsignedBlock` and `statelessUnsignedBlockV0` so that below, we don't need to `switch` on the type and we can do, for example:

```go
func (b *statelessBlock[T]) ParentID() ids.ID {
	return b.StatelessBlock.ParentID()
}
```
without using a switch.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,152 @@ package block
...
+	// build the hash of the following struct:
+	// +-----------------------+----------+------------+
+	// |  prefix :             | [8]byte  | "rng-root" |
+	// +-----------------------+----------+------------+
+	// |  chainID :            | [32]byte |  32 bytes  |
+	// +-----------------------+----------+------------+
+	// |  networkID:           | uint32   |  4 bytes   |
+	// +-----------------------+----------+------------+
+
+	buffer := make([]byte, len(vrfRngRootPrefix)+ids.IDLen+4)
```
danlaine, line 116: nit: could use `wrappers.IntLen` instead of `4` to remove use of a "magic number" even though it's fairly obvious from the comments here

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,152 @@ package block
...
+	copy(buffer, vrfRngRootPrefix)
+	copy(buffer[len(vrfRngRootPrefix):], chainID[:])
+	binary.LittleEndian.PutUint32(buffer[len(vrfRngRootPrefix)+ids.IDLen:], networkID)
```
danlaine, line 119: Any reason in particular we use `LittleEndian` here? We use `BigEndian` everywhere else in the codecase, I think.

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,152 @@ package block
...
+	return hashing.ComputeHash256(parentBlockSig)
+}
+
+func NextBlockVRFSig(parentBlockVRFSig []byte, blsSignKey *bls.SecretKey, chainID ids.ID, networkID uint32) []byte {
+	if blsSignKey == nil {
+		// if we need to build a block without having a BLS key, we'll be hashing the previous
+		// signature only if it presents. Otherwise, we'll keep it empty.
+		if len(parentBlockVRFSig) == 0 {
+			// no parent block signature.
+			return []byte{}
```
danlaine, line 137: Nit: do we need to return `[]byte{}` instead of `nil`? Typically I think we return the latter, unless `[]byte{}` has a special meaning distinct from `nil`.

vms/proposervm/block/build.go
```diff
@@ -47,34 +160,37 @@ func Build(
...
 		cert:      cert,
 		proposer:  ids.NodeIDFromCert(cert),
 	}
-	var blockIntf SignedBlock = block
 
-	unsignedBytesWithEmptySignature, err := Codec.Marshal(CodecVersion, &blockIntf)
-	if err != nil {
+	var err error
+
+	var blkBytes []byte
```
danlaine, line 181: nit: is there a reason we declare these here instead of with `:=` like this?

```go
blkBytes, err := marshalBlock(block)
```

vms/proposervm/block/block.go
tsachiherman: np

vms/proposervm/block.go
tsachiherman: done

vms/proposervm/block/block.go
tsachiherman: the versioned codec eliminates the needs for that..

vms/proposervm/block/build.go
tsachiherman: done.

vms/proposervm/block/build.go
tsachiherman: done.

vms/proposervm/block/build.go
tsachiherman: done

vms/proposervm/block/build.go
tsachiherman: done

abi87 (approved):

vms/proposervm/block/block.go
```diff
@@ -73,32 +80,82 @@ func (b *statelessBlock) Block() []byte {
...
 	if err != nil {
 		return fmt.Errorf("%w: %w", errInvalidCertificate, err)
 	}
 
 	b.proposer = ids.NodeIDFromCert(b.cert)
+
+	// The following ensures that we would initialize the vrfSig member only when
+	// the provided signature is 96 bytes long. That supports both statelessUnsignedBlockV0 & statelessUnsignedBlock
+	// variations, as well as optional 32-byte hashes stored in the VRFSig.
+	if len(b.StatelessBlock.VRFSig) == bls.SignatureLen {
```
danlaine, line 152: Should `len(b.StatelessBlock.VRFSig)` always be either `0` or `bls.SignatureLen`? If so, should we assert that here?

vms/proposervm/block/block.go
```diff
@@ -73,32 +80,82 @@ func (b *statelessBlock) Block() []byte {
...
+
+func (b *statelessBlock) PChainHeight() uint64 {
+	return b.StatelessBlock.PChainHeight
+}
+
+func (b *statelessBlock) Timestamp() time.Time {
+	return b.timestamp
+}
+
+func (b *statelessBlock) Proposer() ids.NodeID {
```
danlaine, line 95: Nit: we could reduce the diff in this PR by keeping these methods in their original place but idc if they move

danlaine (approved):

vms/proposervm/block/block.go
tsachiherman: when a proposer doesn't have a BLS key, they would hash the previous signature instead of signing it. In that case, the VRFSig would be 32 byte long.

I think that adding another case here that validates that the signature length is in [0,32,96] makes sense here. I'll add that.

tsachiherman: done

vms/proposervm/block/block.go
tsachiherman: done.

proto/vm/vm.proto
```diff
@@ -151,6 +151,7 @@ message Handler {
 
 message BuildBlockRequest {
   optional uint64 p_chain_height = 1;
+  optional bytes randomness_seed = 2;
```
StephenButtolph, line 154: Does specifying optional do anything here? Or is this just for consistency with `p_chain_height`?

vms/avm/vm.go
```diff
@@ -395,7 +395,7 @@ func (vm *VM) GetBlockIDAtHeight(_ context.Context, height uint64) (ids.ID, erro
  */
 
 func (vm *VM) Linearize(ctx context.Context, stopVertexID ids.ID, toEngine chan<- common.Message) error {
-	time := vm.Config.Upgrades.CortinaTime
+	time := vm.Upgrades.CortinaTime
```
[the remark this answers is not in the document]
tsachiherman: done.

vms/proposervm/block.go
```diff
@@ -218,10 +222,23 @@ func (p *postForkCommonComponents) buildChild(
 		return nil, err
 	}
 
+	var childBlockVrfSig []byte
+	// do we have the VRFSig activated ? if so, figure out the child block vrf signature.
+	if p.vm.Upgrades.IsEtnaActivated(newTimestamp) {
+		if shouldBuildSignedBlock {
+			childBlockVrfSig = block.NextBlockVRFSig(parentBlockVRFSig, p.vm.Config.StakingBLSKey, p.vm.ctx.ChainID, p.vm.ctx.NetworkID)
```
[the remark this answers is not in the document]
tsachiherman: I see. I believe that my code already handles the validation part correctly ( since it gets the BLS public key from the `ValidatorState.GetValidatorSet`).
For the block building, would it work if I were to retrieve the validator set from `ValidatorState.GetValidatorSet` and compare the public registered BLS key there with the block builder's public key ?

vms/proposervm/block/build.go
```diff
@@ -6,39 +6,143 @@ package block
...
 
+func CalculateVRFOut(vrfSig []byte) [32]byte {
+	// build the hash of the following struct:
+	// +-------------------------+----------+------------+
+	// |  prefix :               | [8]byte  | "rng-derv" |
+	// +-------------------------+----------+------------+
+	// |  vrfSig :               | [96]byte |  96 bytes  |
+	// +-------------------------+----------+------------+
+	if len(vrfSig) != bls.SignatureLen {
+		return [32]byte{0}
```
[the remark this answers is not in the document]
tsachiherman: done

vms/proposervm/block/block.go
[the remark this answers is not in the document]
tsachiherman: I'll try to extract the changes to `proposervm/block` onto it's own PR.

proto/vm/vm.proto
tsachiherman: I used "optional" so that the serialized requests would remain backward compatible. Isn't that the case ?

vms/proposervm/block/build.go
```diff
@@ -86,7 +193,17 @@ func Build(
...
+
+	// The following ensures that we would initialize the vrfSig member only when
+	// the provided signature is 96 bytes long. That supports both statelessUnsignedBlockV0 & statelessUnsignedBlock
+	// variations, as well as optional 32-byte hashes stored in the VRFSig.
+	if len(blockVrfSig) == bls.SignatureLen {
+		block.vrfSig, err = bls.SignatureFromBytes(blockVrfSig)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", errFailedToParseVRFSignature, err)
+		}
+	}
```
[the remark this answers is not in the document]
tsachiherman: sure. done.

proto/vm/vm.proto
StephenButtolph: It isn't clear to me that `optional bytes` is different at all from just `bytes`. See the proto spec: https://protobuf.dev/programming-guides/proto3/#field-labels

StephenButtolph:
This ACP was marked as stale.

PR #3147, opened 2024-06-25
Restrict `Owner` usage after a Subnet manager is set

dhrubabasu:
## Why this should be merged

[ACP-77](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/77-reinventing-subnets/README.md) calls for disabling all transactions that uses `SubnetAuth` other than `TransferSubnetOwnershipTx` when a Subnet sets a manager.

## How this works

There are 5 transactions that have `SubnetAuth` as a field:
- AddSubnetValidatorTx
- CreateChainTx
- RemoveSubnetValidatorTx
- TransferSubnetOwnershipTx
- TransformSubnetTx

This PR considers the first three invalid if a Subnet has a manager that is set.

The execution of TransformSubnetTx does not need to be modified as it is already disabled post-Etna: https://github.com/ava-labs/avalanchego/pull/3152

## How this was tested

added UTs


vms/platformvm/txs/executor/staker_tx_verification.go
```diff
@@ -204,6 +206,13 @@ func verifyAddSubnetValidatorTx(
 		return err
 	}
 
+	if backend.Config.UpgradeConfig.IsEActivated(currentTimestamp) {
```
dhrubabasu, line 209: This `E` check isn't strictly required since there are no `SetSubnetManager` calls until `E` is activate. However, doing a timestamp check avoids a DB read that is performed in `GetSubnetManager`.

vms/platformvm/txs/executor/staker_tx_verification.go
```diff
@@ -301,6 +310,13 @@ func verifyRemoveSubnetValidatorTx(
 		return nil, false, err
 	}
 
+	if backend.Config.UpgradeConfig.IsEActivated(currentTimestamp) {
```
dhrubabasu, line 313: This `E` check isn't strictly required since there are no `SetSubnetManager` calls until `E` is activate. However, doing a timestamp check avoids a DB read that is performed in `GetSubnetManager`.

vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -1969,6 +2007,27 @@ func TestStandardExecutorRemoveSubnetValidatorTx(t *testing.T) {
 			},
 			expectedErr: ErrFlowCheckFailed,
 		},
+		{
+			name: "attempted to remove subnet validator after subnet manager is set",
```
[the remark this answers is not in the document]
dhrubabasu: I'd break out this test and do the series of events that can lead to the `expectedErr`. The main reason (afaict) it's pretty annoying to avoid mocks is because the `state` pkg is so unwieldy. I imagine a lot of this will be streamlined once we remove historical execution.

vms/platformvm/txs/executor/staker_tx_verification.go
```diff
@@ -207,6 +209,13 @@ func verifyAddSubnetValidatorTx(
 		return err
 	}
 
+	if backend.Config.UpgradeConfig.IsEtnaActivated(currentTimestamp) {
+		_, _, err := chainState.GetSubnetManager(tx.Subnet)
```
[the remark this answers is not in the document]
dhrubabasu: `database.ErrNotFound` means that there is definitively no subnet manager. `nil` or any other error means there is a subnet manager. The details of the error aren't required since the only other possible error is a codec unmarshal error which could only happen if the db is corrupted.

vms/platformvm/txs/executor/subnet_tx_verification.go
```diff
@@ -43,6 +43,14 @@ func verifyPoASubnetAuthorization(
...
 	}
 
+	_, _, err = chainState.GetSubnetManager(subnetID)
+	if err == nil {
+		return nil, fmt.Errorf("%q %w", subnetID, errIsImmutable)
+	}
+	if err != database.ErrNotFound {
+		return nil, err
+	}
+
```
dhrubabasu, line 53: I think this is probably the most obvious way to do this change versus how I had it originally (cc @marun)

vms/platformvm/txs/executor/staker_tx_verification.go
```diff
@@ -306,6 +307,13 @@ func verifyRemoveSubnetValidatorTx(
 		return nil, false, err
 	}
 
+	if backend.Config.UpgradeConfig.IsEtnaActivated(currentTimestamp) {
+		_, _, err := chainState.GetSubnetManager(tx.Subnet)
+		if err != database.ErrNotFound {
+			return nil, false, ErrRemoveValidatorManagedSubnet
+		}
```
StephenButtolph, line 314: Would it make sense to use the same check pattern in both places?
```suggestion
		_, _, err := chainState.GetSubnetManager(tx.Subnet)
		if err == nil {
			return nil, fmt.Errorf("%w: %q", ErrRemoveValidatorManagedSubnet, tx.Subnet)
		}
		if err != database.ErrNotFound {
			return nil, err
		}
```

StephenButtolph (commented):
lgtm

StephenButtolph (approved):

Issue #3109, opened 2024-06-12
Unexpected `failed to connect to bootstrap nodes` warning


StephenButtolph:
Avalanchego will incorrectly report that the beacon nodes are unreachable when running a node with reduced performance.

https://github.com/ava-labs/avalanchego/blob/41e46d1178116eb7421fe2adbf2eaa7ef97b0f8b/node/node.go#L622-L625

This issue is able to be replicated consistently on nodes with high performance by extending the time to initialize the `platformvm`.

For example, modifying https://github.com/ava-labs/avalanchego/blob/41e46d1178116eb7421fe2adbf2eaa7ef97b0f8b/vms/platformvm/vm.go#L108. To:
```diff
+	time.Sleep(time.Minute)
+
	chainCtx.Log.Verbo("initializing platform chain")
```

PR #3156, opened 2024-06-28
Fix bootstrapping warn log

joshua-kim:
## Why this should be merged

Fixes https://github.com/ava-labs/avalanchego/issues/3109

## How this works

Starts timeout goroutine before starting the networking stack, instead of during network initialization where other long-running processes may falsely call this to time out.

## How this was tested

Tested manually

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -411,11 +411,6 @@ type Node struct {
  ******************************************************************************
  */
 
-func (n *Node) getRequiredConns() int {
-	numBootstrappers := n.bootstrappers.Count(constants.PrimaryNetworkID)
-	return (3*numBootstrappers + 3) / 4
-}
-
 // Initialize the networking layer.
 // Assumes [n.vdrs], [n.CPUTracker], and [n.CPUTargeter] have been initialized.
 func (n *Node) initNetworking(reg prometheus.Registerer) error {
@@ -605,7 +600,10 @@ func (n *Node) initNetworking(reg prometheus.Registerer) error {
 	}
 
 	n.onSufficientlyConnected = make(chan struct{})
-	if requiredConns := n.getRequiredConns(); requiredConns > 0 {
+	numBootstrappers := n.bootstrappers.Count(constants.PrimaryNetworkID)
+	requiredConns := (3*numBootstrappers + 3) / 4
+
+	if requiredConns > 0 {
 		consensusRouter = &beaconManager{
 			Router:                  consensusRouter,
 			beacons:                 n.bootstrappers,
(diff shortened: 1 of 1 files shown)


darioush (approved):

node/node.go
```diff
@@ -400,6 +400,9 @@ type Node struct {
 	// Specifies how much disk usage each peer can cause before
 	// we rate-limit them.
 	diskTargeter tracker.Targeter
+
+	// Fired when a sufficient amount of bootstrapper nodes are connected to
```
StephenButtolph, line 404:
```suggestion
	// Closed when a sufficient amount of bootstrap nodes are connected to
```

node/node.go
```diff
@@ -725,6 +709,25 @@ func (n *Node) Dispatch() error {
 		n.Net.ManuallyTrack(bootstrapper.ID, bootstrapper.IP)
 	}
 
+	// Log a warning if we aren't able to connect to a sufficient portion of
+	// nodes.
+	go func() {
```
StephenButtolph, line 714: I think that this would now incorrectly fire whenever `if requiredConns <= 0 {`.

StephenButtolph, line 714: Super small nit, but I think it would make more sense to move this prior to calling `n.Net.ManuallyTrack`. The comment below here stating `// Start P2P connections` is a bit misleading. Calling `n.Net.Dispatch()` starts handling _inbound_ connections. Calling `ManuallyTrack` can start creating _outbound_ connections.

patrick-ogrady (approved):

node/node.go
StephenButtolph: When running a 1-node local network this fired unexpectedly:
```
[06-28|15:19:45.577] WARN node/node.go:723 failed to connect to bootstrap nodes {"bootstrappers": "Validator Manager: (Size = 1)\n    Subnet[11111111111111111111111111111111LpoYY]: Validator Set: (Size = 0, Weight = 0)", "duration": "1m0s"}
```

node/node.go
```diff
@@ -596,36 +604,14 @@ func (n *Node) initNetworking(reg prometheus.Registerer) error {
 		}
 	}
 
-	numBootstrappers := n.bootstrappers.Count(constants.PrimaryNetworkID)
-	requiredConns := (3*numBootstrappers + 3) / 4
-
-	if requiredConns > 0 {
-		onSufficientlyConnected := make(chan struct{})
+	if requiredConns := n.getRequiredConns(); requiredConns > 0 {
```
StephenButtolph, line 607: This is probably just mean being paranoid... But the initialization flow is pretty complex (and I worry that we might introduce a weird bug here in the future by modifying `n.bootstrappers` between here and `Start`)... 

Would it be weird to just create `onSufficientlyConnected` regardless of the value of `requiredConns`, and close it if `requiredConns <= 0`?

StephenButtolph (approved):

PR #3157, opened 2024-06-29
check router is closing in requests

### LINKED ISSUE #3155: Unexpected ERROR during Shutdown
**Describe the bug**

Observed:
```
[06-25|16:06:30.953] ERROR router/chain_router.go:721 trying to issue InternalConnectedSubnet message, but platform chain is not registered {"nodeID": "NodeID-MFrZFVCXPv5iCn6M9K6XduxGTYp891xXZ", "subnetID": "29uVeLPJB1eQJkzRemU8g8wZDw5uJRqpab5U2mX9euieVwiEbL"}
```
during shutdown.

The node was running an XSVM subnet.

**To Reproduce**

This was run in a custom environment with faults being injected into the system.

**Expected behavior**

There shouldn't be an `ERROR` during shutdown.

ceyonur:
## Why this should be merged

Fixes #3155 

## How this works

This pull request primarily focuses on introducing a closing mechanism to the `ChainRouter` structure in the `snow/networking/router/chain_router.go` file. This mechanism prevents the router from processing requests and messages once it's in the process of shutting down. The most significant changes include the addition of the `closing` boolean field to the `ChainRouter` struct, a new error type `errClosing` to signal the router is closing, and modifications to several methods to check the `closing` field before proceeding with their operations.

### Added closing mechanism:

* [`snow/networking/router/chain_router.go`](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R67): Added `closing` boolean field to `ChainRouter` struct to indicate whether the router is in the process of shutting down.
* [`snow/networking/router/chain_router.go`](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R34): Introduced `errClosing` error type to signal that the router is closing.

### Modified methods to check `closing` field:

* [`snow/networking/router/chain_router.go`](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R159-R170): Modified `RegisterRequest`, `HandleInbound`, `Shutdown`, `AddChain`, `Connected`, `Disconnected`, `Benched`, and `Unbenched` methods to check the `closing` field before proceeding. If the router is closing, these methods log a debug message and return early. [[1]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R159-R170) [[2]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R261-R271) [[3]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R384) [[4]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R417-R423) [[5]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R482-R489) [[6]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R537-R544) [[7]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R574-R582) [[8]](diffhunk://#diff-c2531a5dcd3ad1ff091a235c9b0a4f60064ab41e8f1317466ce0eb4ad2a798e4R615-R623)

## How this was tested

Added unit test

### DIFF
--- a/snow/networking/router/chain_router.go
+++ b/snow/networking/router/chain_router.go
@@ -31,6 +31,7 @@ import (
 var (
 	errUnknownChain  = errors.New("received message for unknown chain")
 	errUnallowedNode = errors.New("received message from non-allowed node")
+	errClosing       = errors.New("router is closing")
 
 	_ Router              = (*ChainRouter)(nil)
 	_ benchlist.Benchable = (*ChainRouter)(nil)
@@ -63,6 +64,7 @@ type ChainRouter struct {
 	clock         mockable.Clock
 	log           logging.Logger
 	lock          sync.Mutex
+	closing       bool
 	chainHandlers map[ids.ID]handler.Handler
 
 	// It is only safe to call [RegisterResponse] with the router lock held. Any
@@ -154,6 +156,18 @@ func (cr *ChainRouter) RegisterRequest(
 	engineType p2p.EngineType,
 ) {
 	cr.lock.Lock()
+	if cr.closing {
+		cr.log.Debug("dropping request",
+			zap.Stringer("nodeID", nodeID),
+			zap.Stringer("requestingChainID", requestingChainID),
+			zap.Stringer("respondingChainID", respondingChainID),
+			zap.Uint32("requestID", requestID),
+			zap.Stringer("messageOp", op),
+			zap.Error(errClosing),
+		)
+		cr.lock.Unlock()
+		return
+	}
 	// When we receive a response message type (Chits, Put, Accepted, etc.)
 	// we validate that we actually sent the corresponding request.
 	// Give this request a unique ID so we can do that validation.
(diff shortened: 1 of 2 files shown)


snow/networking/router/chain_router.go
```diff
@@ -63,6 +64,7 @@ type ChainRouter struct {
 	clock         mockable.Clock
 	log           logging.Logger
 	lock          sync.Mutex
+	closing       bool
```
StephenButtolph, line 67: Do we care that the P-chain encountering a FATAL error could still cause the unexpected ERROR log? Since we are already FATALing... I don't think it would be too big of an issue... but just wanted to point out that it is still possible.

ceyonur: I think if P-Chain fatals chain router won't be able to know it right, because `Push` won't return anything? Unless we talk about `SetOnStopped`/`removeChain` in this case it will be shutdown through node's shutdown path (through router's `onFatal` callback). I don't know if it violates any invariant if we shutdown if [here](https://github.com/ava-labs/avalanchego/blob/master/snow/networking/router/chain_router.go#L739-L741).

StephenButtolph: Yeah... I think this is fine, if the node is FATALing we have bigger issues than this logging.

dhrubabasu (approved):

PR #3158, opened 2024-06-30
Remove `block.Status`

StephenButtolph:
## Why this should be merged

This PR is the culmination of:
1. #3111
2. #3123
3. #3124
4. #3125
5. #3130
6. #3139
7. #3131
8. #3132
9. #3140

By removing `Status` from the `snowman.Block` interface, VMs no longer need to implement this method; which has historically been the source of many VM bugs.

## How this works

By removing the last usage of `snowman.Block#Status` in the bootstrapping engine, the function can be removed from the interface and all VM implementations.

## How this was tested

- [x] CI
- [x] Fuji Sync
  - Reduced the number of DB reads when bootstrapping the P-chain by 10% (from `1,403,721` to `1,268,418`)
- [x] Fuji Probe
- [x] Mainnet Probe
- [x] Fuji Canary
- [x] Mainnet Canary


vms/example/xsvm/chain/block.go
```diff
@@ -178,12 +167,12 @@ func (b *block) VerifyWithContext(ctx context.Context, blockContext *smblock.Con
...
 
 func (b *block) State() (database.Database, error) {
-	if b.id == b.chain.lastAccepted {
+	if b.id == b.chain.lastAcceptedID {
 		return b.chain.acceptedState, nil
 	}
 
 	// States of accepted blocks other than the lastAccepted are undefined.
-	if b.Status() == choices.Accepted {
+	if b.Height() <= b.chain.lastAcceptedHeight {
```
StephenButtolph, line 175: This feels a bit dubious... should look into this tomorrow

vms/proposervm/post_fork_option.go
```diff
@@ -25,7 +24,7 @@ type postForkOption struct {
 }
 
 func (b *postForkOption) Timestamp() time.Time {
-	if b.Status() == choices.Accepted {
+	if b.Height() <= b.vm.lastAcceptedHeight {
```
StephenButtolph, line 27: We shouldn't care about any rejected timestamps - so this should be the same

vms/proposervm/vm.go
```diff
@@ -532,42 +531,30 @@ func (vm *VM) parsePostForkBlock(ctx context.Context, b []byte) (PostForkBlock,
 		return nil, err
 	}
 
-	// if the block already exists, then make sure the status is set correctly
 	blkID := statelessBlock.ID()
-	blk, err := vm.getPostForkBlock(ctx, blkID)
```
StephenButtolph, line 537: Do we need to handle verified blocks better here?

vms/example/xsvm/chain/block.go
StephenButtolph: All of this code is just defense in depth, but I cleaned it up a bit.

snow/snowtest/status.go
```diff
@@ -0,0 +1,28 @@
...
+func (s Status) String() string {
+	switch s {
+	case Undecided:
+		return "Undecided"
+	case Accepted:
+		return "Accepted"
+	case Rejected:
+		return "Rejected"
+	default:
+		return "Unknown"
```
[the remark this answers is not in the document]
StephenButtolph: I think this is fine.

snow/consensus/snowman/snowmantest/engine.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/ava-labs/avalanchego/snow/snowtest"
+)
+
+func MakeLastAcceptedBlockF(blks ...[]*Block) func(context.Context) (ids.ID, error) {
+	return func(context.Context) (ids.ID, error) {
+		var (
+			highestHeight uint64
+			highestID     ids.ID
+		)
+		for _, blks := range blks {
```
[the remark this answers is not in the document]
StephenButtolph: It was... But probably makes sense to avoid the shadowing. (updated)

snow/engine/snowman/getter/getter.go
```diff
@@ -148,10 +147,34 @@ func (gh *getter) GetAcceptedFrontier(ctx context.Context, nodeID ids.NodeID, re
...
+		return err
+	}
+	lastAcceptedHeight := lastAccepted.Height()
+
 	acceptedIDs := make([]ids.ID, 0, containerIDs.Len())
 	for blkID := range containerIDs {
 		blk, err := gh.vm.GetBlock(ctx, blkID)
-		if err == nil && blk.Status() == choices.Accepted {
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
StephenButtolph: It is expected for the last accepted block to be locally accessible. However, it is not guaranteed that the (potentially malicious peer) is only including block IDs that you have locally.

vms/example/xsvm/chain/block.go
```diff
@@ -12,11 +12,9 @@ import (
 	"github.com/ava-labs/avalanchego/database/versiondb"
```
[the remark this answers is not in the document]
StephenButtolph: Makes it easier to keep it up-to-date, I think we already have enough VMs in-tree to have that feedback.

vms/proposervm/vm_test.go
```diff
@@ -2012,101 +1986,6 @@ func TestVMInnerBlkCache(t *testing.T) {
 	require.False(ok)
 }
 
-func TestVMInnerBlkCacheDeduplicationRegression(t *testing.T) {
```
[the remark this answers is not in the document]
StephenButtolph: These tests were regression tests specifically for incorrect `.Status()` behavior. We've removed the `.Status()` function entirely, so there is nothing left to test here.

vms/components/chain/state_test.go
```diff
@@ -153,45 +146,35 @@ func checkDecidedBlock(t *testing.T, s *State, blk snowman.Block, expectedStatus
...
 	testBlks = append(testBlks, blk3)
 
-	getBlock, parseBlock, getCanonicalBlockID := createInternalBlockFuncs(testBlks)
+	getBlock, parseBlock := createInternalBlockFuncs(testBlks)
 	chainState := NewState(&Config{
-		DecidedCacheSize:    2,
-		MissingCacheSize:    2,
-		UnverifiedCacheSize: 2,
-		BytesToIDCacheSize:  2,
+		DecidedCacheSize:    256,
```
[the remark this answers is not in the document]
StephenButtolph: I'd prefer not to redesign these tests in this PR (I doesn't seem like a super straightforward cut..). But I changed the sizes to use a shared var.

proto/vm/vm.proto
```diff
@@ -177,7 +170,6 @@ message ParseBlockRequest {
 message ParseBlockResponse {
   bytes id = 1;
   bytes parent_id = 2;
-  Status status = 3;
```
dhrubabasu, line 180: should we add `reserved 3;` here?

proto/vm/vm.proto
```diff
@@ -190,7 +182,6 @@ message GetBlockRequest {
 message GetBlockResponse {
   bytes parent_id = 1;
   bytes bytes = 2;
-  Status status = 3;
```
dhrubabasu, line 193: should we add `reserved 3;` here?

proto/vm/vm.proto
StephenButtolph: We don't typically do this for the VM interface, because this is gated by the rpcchainvm proto version (which will need to be increased in this release).

proto/vm/vm.proto
StephenButtolph: See above comment

joshua-kim (approved):

dhrubabasu (approved):

vms/components/chain/state.go
```diff
@@ -471,32 +413,25 @@ func (s *State) deduplicate(ctx context.Context, blk snowman.Block) (snowman.Blo
...
+	return s.addBlockOutsideConsensus(blk)
 }
 
 // addBlockOutsideConsensus adds [blk] to the correct cache and returns
 // a wrapped version of [blk]
 // assumes [blk] is a known, non-wrapped block that is not currently
 // in consensus. [blk] could be either decided or a block that has not yet
 // been verified and added to consensus.
-func (s *State) addBlockOutsideConsensus(ctx context.Context, blk snowman.Block) (snowman.Block, error) {
+func (s *State) addBlockOutsideConsensus(blk snowman.Block) (snowman.Block, error) {
```
tsachiherman, line 424: congrats! you've just eliminated all the error path in this method. mind removing the `error` return value ?

tsachiherman (approved):
looks good. just one comment above.

vms/components/chain/state.go
StephenButtolph: I thought we had a linter for this 🤔

PR #3160, opened 2024-07-01
[ci] Add actionlint job

marun:
Closes #3113

PR #3161, opened 2024-07-01
[ci] Simplify ci monitoring with custom actions

marun:
Previously, enabling monitoring of temporary networks for a CI job required duplicating complicated action steps. Encapsulating those steps in a custom action is intended to simplify maintenance and enable direct reuse by other repos (e.g. subnet-evm, hypersdk).  

## TODO

- [x] Verify metrics collection
- [x] Merge #3160


StephenButtolph (approved):

joshua-kim (approved):

tsachiherman (approved):
looks good. Note that if you get to the point where you'd be using the same action across multiple github repositories, you could create a "common" repository that would host these shared actions.
That's particulary useful when you have a username/password that is stored as part of the repository secrets, and you want to make sure there is only a single "instance" of it.

PR #3165, opened 2024-07-02
Use `ids.GenerateTestID` helper in tests

dhrubabasu:
## Why this should be merged

We hardcode random byte slices in a lot of the tests. Let's use the helper we have already instead :)

## How this works

Replace hard-coded byte slices with `ids.GenerateTestID()`

## How this was tested

CI

### DIFF
--- a/snow/engine/snowman/syncer/state_syncer_test.go
+++ b/snow/engine/snowman/syncer/state_syncer_test.go
@@ -1262,7 +1262,7 @@ func TestStateSyncIsStoppedIfEnoughVotesAreCastedWithNoClearMajority(t *testing.
 				context.Background(),
 				voterID,
 				reqID,
-				set.Of(ids.ID{'u', 'n', 'k', 'n', 'o', 'w', 'n', 'I', 'D'}),
+				set.Of(ids.GenerateTestID()),
 			))
 			votingWeightStake += beacons.GetWeight(ctx.SubnetID, voterID)
 		}
--- a/vms/platformvm/block/serialization_test.go
+++ b/vms/platformvm/block/serialization_test.go
@@ -6,6 +6,7 @@ package block
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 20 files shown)


snow/engine/snowman/syncer/state_syncer_test.go
```diff
@@ -1262,7 +1262,7 @@ func TestStateSyncIsStoppedIfEnoughVotesAreCastedWithNoClearMajority(t *testing.
 				context.Background(),
 				voterID,
 				reqID,
-				set.Of(ids.ID{'u', 'n', 'k', 'n', 'o', 'w', 'n', 'I', 'D'}),
+				set.Of(ids.GenerateTestID()),
```
ARR4N, line 1265: Was it ever useful in debugging to have these kinds of "labelled" ids? If so, you could add `ids.GenerateLabelledTestID(tb testing.TB, lbl string)` that then calls `tb.Logf("test ID %q: %v", lbl, id)` before returning the value.

The only problem I see with this would be introducing the `testing` package into prod code.

vms/platformvm/block/serialization_test.go
```diff
@@ -142,20 +145,23 @@ func TestBanffProposalBlockJSON(t *testing.T) {
 	simpleBanffProposalBlockBytes, err := json.MarshalIndent(simpleBanffProposalBlock, "", "\t")
 	require.NoError(err)
 
-	require.Equal(`{
+	expectedSimpleBanffProposalBlockString := `{
```
ARR4N, line 148: This would be clearer if it used `fmt.Sprintf()` instead of `strings.Replace()`.

(optional, subjective) You could use [explicit arg indices](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes) if it makes it more readable, but I'm not sure that it will.

vms/platformvm/block/serialization_test.go
```diff
@@ -186,7 +192,7 @@ func TestBanffProposalBlockJSON(t *testing.T) {
 	complexBanffProposalBlockBytes, err := json.MarshalIndent(complexBanffProposalBlock, "", "\t")
 	require.NoError(err)
 
-	require.Equal(`{
+	expectedComplexBanffProposalBlockString := `{
```
ARR4N, line 195: As above re `fmt.Sprintf()`.

ARR4N (approved):
Approving with requested changes (use `fmt.Sprintf()` instead of `strings.Replace()`) as I don't need to re-review them, but please address before merging.

snow/engine/snowman/syncer/state_syncer_test.go
ARR4N: This is completely optional and dependent on the answer to my opening question.

vms/platformvm/block/serialization_test.go
StephenButtolph: TIL about explicit arg indicies

StephenButtolph (commented):
Are there any cases here where this introduces a weird dependency on the sorted-ness of IDs returned by `ids.GenerateTestID()`?

dhrubabasu:
Converting to draft for now, there seems to be dependencies in the codebase already that rely on the sorted-ness of the IDs returned. Will investigate further.

PR #3166, opened 2024-07-02
Use `ids.Empty` instead of `ids.ID{}`

dhrubabasu:
## Why this should be merged

We should be consistent on which one we use

## How this works

Aligns on `ids.Empty`

## How this was tested

CI

### DIFF
--- a/chains/test_manager.go
+++ b/chains/test_manager.go
@@ -42,7 +42,7 @@ func (testManager) StartChainCreator(ChainParameters) error {
 }
 
 func (testManager) SubnetID(ids.ID) (ids.ID, error) {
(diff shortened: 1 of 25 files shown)


darioush (approved):

meaghanfitzgerald:
Did a global search and it says this variable exists in 36 files, this PR changes 25. 
i.e. https://github.com/ava-labs/avalanchego/blob/master/vms/components/avax/utxo_fetching.go

image

dhrubabasu:
The remaining 11 hits are `[]ids.ID{}`, not `ids.ID{}`

StephenButtolph (approved):

PR #3167, opened 2024-07-02
Replace usage of utils.Err with errors.Join

joshua-kim:
## Why this should be merged

Remove helper we maintain and use the stdlib instead

## How this works

Removes `utils.Err`

## How this was tested

CI


utils/error.go
```diff
@@ -1,13 +0,0 @@
-// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
-// See the file LICENSE for licensing terms.
-
-package utils
-
-func Err(errors ...error) error {
-	for _, err := range errors {
-		if err != nil {
-			return err
```
dhrubabasu, line 9: A subtle change between `errors.Join` and `utils.Err` is that `utils.Err` returns the _first_ error whereas `errors.Join` returns a combined error (https://go.dev/play/p/n4_zaEd4ODl):

```golang
import (
	"errors"
	"fmt"
)

func main() {
	err1 := errors.New("err1")
	err2 := errors.New("err2")
	err := errors.Join(err1, err2)
	fmt.Println(err)
	if errors.Is(err, err1) {
		fmt.Println("err is err1")
	}
	if errors.Is(err, err2) {
		fmt.Println("err is err2")
	}
}
```

```text
err1
err2
err is err1
err is err2
```

dhrubabasu: I don't think this matters for our use-cases but thought I'd point it out

cache/metercacher/metrics.go
```diff
@@ -78,7 +78,7 @@ func newMetrics(
 			Help:      "fraction of cache filled",
 		}),
 	}
-	return m, utils.Err(
```
dhrubabasu, line 81: (style) can we assign a new `err` variable before the return statement like we do in the other metrics functions? I know this wasn't changed in this PR

database/meterdb/db.go
```diff
@@ -125,7 +125,7 @@ func New(
 			methodLabels,
 		),
 	}
-	return meterDB, utils.Err(
+	return meterDB, errors.Join(
```
dhrubabasu, line 128: (style) can we assign a new `err` variable before the return statement like we do in the other metrics functions? I know this wasn't changed in this PR

cache/metercacher/metrics.go
```diff
@@ -78,7 +78,7 @@ func newMetrics(
...
+	return m, errors.Join(
```
dhrubabasu, line 81: (style) for these metrics ones, we either do `errors.Join` inline in the return statement or assign a new `err` variable before the return statement. can we align them for consistency? I know this wasn't changed in this PR.

I think I would prefer assigning a new `err` variable before but don't have a strong opinion either way

StephenButtolph (approved):

cache/metercacher/metrics.go
joshua-kim: Going to leave this as-is because it's not something this PR changed. If we want we can do a follow-up to get this consistent.

dhrubabasu (approved):

PR #3172, opened 2024-07-04
[testing] Add scheduled jobs to test bootstrap from mainnet and testnet

marun:
## Why this should be merged

Periodic bootstrap testing is intended to insure that bootstrapping issues are detected prior to releasing to testnet and mainnet.

## How this works

- adds a new test binary and image to initialize and monitor a bootstrapping node
- adds an e2e test job to validate the image

## How this was tested

- [ ] The new test job is passing in CI


.github/workflows/check-bootstrap-testnet-full-sync.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  # TODO(marun) Add a schedule
+  workflow_dispatch:
+
+jobs:
+  check_bootstrap_testnet_full_sync:
+    name: Check Bootstrap (testnet,full-sync)
+    runs-on: avalanche-avalanchego
+    timeout-minutes: 7200 # 5 days, maximum allowed for jobs running on a self-hosted runner
```
abi87, line 11: q: I guess testnet is much faster to sync than mainnet. Should we try a smaller timeout here, just to see this job failing earlier than mainnet one if anything weird happens?

tests/bootstrap/main.go
```diff
@@ -0,0 +1,154 @@
...
+	// Avoid checking too often to prevent log spam
+	healthCheckInterval := 1 * time.Minute
+
+	ctx, cancel = context.WithTimeout(context.Background(), maxDuration)
+	defer cancel()
+	if err := tmpnet.WaitForHealthyWithInterval(ctx, node, healthCheckInterval); err != nil {
+		return fmt.Errorf("node failed to become healthy before timeout: %w", err)
+	}
+
+	log.Print("Bootstrap completed successfully!\n")
```
abi87, line 140: q: thinking alond here,  I wonder if we should also test that the node can keep up with traffic once it goes out of bootstrap? Like run it some time after bootstrap completes and check that it does not fall behind too much.

abi87 (approved):
lgtm once the conflict and lint issue is solved

tests/bootstrap/main.go
StephenButtolph: That seems out of scope

PR #3173, opened 2024-07-04
build(tests): require `//go:build test` tag if importing test packages outside of `_test.go` files

ARR4N:
## Why this should be merged

Including test-only code in non-test files risks leaking said functionality into production.

## How this works

A new linter finds all `*.go` files that (i) aren't `_test.go`; and (ii) import test packages[^1]. If any of these files don't include a build-constraint tag of `test` then the linter fails. To the best of my knowledge, there's no native tag that's set when running tests, so all runs of `go test` and `ginkgo` now need to include `-tags test`.

[^1]: Currently detects `testing` and `stretchr/testify/*` packages. As `mockgen` doesn't add build tags, detection of mocks is disabled pending a [PR](https://github.com/uber-go/mock/pull/191) to them.

## How this was tested

Manually: I wrote the linter first as a way to find the offending files.

## Future work

See #3174 for files flagged by the linter that have been temporarily ignored.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,11 @@
+{
+    "gopls": {
+        "build.buildFlags": [
+            // Context: https://github.com/ava-labs/avalanchego/pull/3173
+            // Without this tag, the language server won't build the test-only
+            // code in non-_test.go files.
+            "--tags='test'",
+        ],
+    },
+    "go.testTags": "test",
+}
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,8 @@ tmp/
 
 *logs/
 
+.vscode*
+
 *.pb*
 
 db*
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,6 @@ tmp/
 
 *logs/
 
-.vscode*
-
 *.pb*
 
 db*
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,11 @@ run:
   # By default, it isn't set.
   modules-download-mode: readonly
 
+  # Include non-test files tagged as test-only.
+  # Context: https://github.com/ava-labs/avalanchego/pull/3173
+  build-tags:
+    - test
+
 output:
   # Make issues output unique by line.
   # Default: true
--- a/cache/test_cacher.go
+++ b/cache/test_cacher.go
@@ -1,6 +1,8 @@
 // Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
+//go:build test
+
 package cache
 
 import (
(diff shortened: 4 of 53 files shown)


darioush (approved):

StephenButtolph (dismissed):
This change LGTM - will wait to hear if @marun has any thoughts w.r.t. the `tests/` directory

StephenButtolph:
This change seems to break the default behavior for running tests in VS code.. ex:
image
image

Running the test results in:
```bash
go test -timeout 30s -coverprofile=/var/folders/5l/z00z8zbd0dbd6lf8djdgl_000000gn/T/vscode-godMWCa1/go-code-cover -run ^TestGossiperGossip$ github.com/ava-labs/avalanchego/network/p2p/gossip -coverpkg=all

# github.com/ava-labs/avalanchego/network/p2p/gossip [github.com/ava-labs/avalanchego/network/p2p/gossip.test]
/Users/stephen/go/src/github.com/ava-labs/avalanchego/network/p2p/gossip/gossip_test.go:107:30: undefined: common.FakeSender
/Users/stephen/go/src/github.com/ava-labs/avalanchego/network/p2p/gossip/gossip_test.go:136:29: undefined: common.FakeSender
/Users/stephen/go/src/github.com/ava-labs/avalanchego/network/p2p/gossip/gossip_test.go:512:22: undefined: common.FakeSender
FAIL	github.com/ava-labs/avalanchego/network/p2p/gossip [build failed]
FAIL
```

Should we add a `.vscode/settings.json` file? It would specify:
```json
{
    "go.buildTags": "test"
}
```

If we do this then we end up with:
```bash
go test -timeout 30s -tags test -coverprofile=/var/folders/5l/z00z8zbd0dbd6lf8djdgl_000000gn/T/vscode-go731twr/go-code-cover -run ^TestGossiperGossip$ github.com/ava-labs/avalanchego/network/p2p/gossip -coverpkg=all

ok  	github.com/ava-labs/avalanchego/network/p2p/gossip	0.876s	coverage: 9.8% of statements in all
```

which works as expected.

StephenButtolph:
This is also a `go.testTags` option, but it seems like that doesn't fix the errors presented by the IDE 😞

ARR4N:
> There is also a `go.testTags` option, but it seems like that doesn't fix the errors presented by the IDE 😞 (although it would fix the running of the tests)

There's a `gopls` build-flag setting that fixes the IDE complaints:

```json
{
    "gopls": {
        "build.buildFlags": [
            "--tags='test'",
        ],
    },
    "go.testTags": "test",
}
```

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+    "gopls": {
+        "build.buildFlags": [
+            // Context: https://github.com/ava-labs/avalanchego/pull/3173
```
tsachiherman, line 4: I'm surprised to see that this works, but wouldn't touch it if it does.

StephenButtolph: Is is documented somewhere that this is supported? This also surprises me.

StephenButtolph: It is documented: https://code.visualstudio.com/docs/languages/json#_json-with-comments. TIL

StephenButtolph (approved):

StephenButtolph:
Maru is OOO - going to merge.

PR #3177, opened 2024-07-06
[vms/avm] Replace `strings.Replace` with `fmt.Sprintf` in tests

dhrubabasu:
## Why this should be merged

@ARR4N pointed out that we could use `fmt.Sprintf` with explicit arg indices for better readability during review of another PR (https://github.com/ava-labs/avalanchego/pull/3165#discussion_r1664012222)

## How this works

Replaces `strings.Replace` with `fmt.Sprintf` in the avm tests (only place we do this pattern currently)

## How this was tested

CI


StephenButtolph (approved):

vms/avm/service_test.go
```diff
@@ -538,10 +538,13 @@ func TestServiceGetTxJSON_BaseTx(t *testing.T) {
...
 
-	expectedReplyTxString := `{
+	sigStr, err := formatting.Encode(formatting.HexNC, newTx.Creds[0].Credential.(*secp256k1fx.Credential).Sigs[0][:])
+	require.NoError(err)
+
+	expectedReplyTxString := fmt.Sprintf(`{
 	"unsignedTx": {
 		"networkID": 10,
-		"blockchainID": "PLACEHOLDER_BLOCKCHAIN_ID",
+		"blockchainID": "%s",
```
ARR4N, line 547: `"%s"` can (and usually should) be replaced with `%q`, which will also escape inputs as necessary. This applies throughout the PR.

vms/avm/service_test.go
```diff
@@ -1101,10 +1080,13 @@ func TestServiceGetTxJSON_OperationTxWithMultipleNftxMintOp(t *testing.T) {
...
 
-	expectedReplyTxString := `{
+	sigStr, err := formatting.Encode(formatting.HexNC, mintNFTTx.Creds[1].Credential.(*nftfx.Credential).Sigs[0][:])
+	require.NoError(err)
+
+	expectedReplyTxString := fmt.Sprintf(`{
 	"unsignedTx": {
 		"networkID": 10,
-		"blockchainID": "PLACEHOLDER_BLOCKCHAIN_ID",
+		"blockchainID": "%[1]s",
```
ARR4N, line 1089: (optional, subjective) With such a long format string I'd consider defining the argument slice just before. My hypothesis is that it reduces cognitive load in reading the code because you don't have to repeatedly jump down and back up.

(fyi) Another option for very long blocks like this is to use a `text/template.Template`, which effectively gives you named placeholders instead of argument indices. If you were to do this then I'd abstract a helper function to avoid clogging up the test definitions:

```go
func fmtTemplate(t *testing.T, tmpl string, data any) {
  t.Helper()
  tt, err := template.New("").Parse(tmpl)
  require.NoError(t, err)
  
  var buf bytes.Buffer
  require.NoError(t, tt.Execute(&buf, data))
  return buf.String()
}
...
  expectedReplyTxString := fmtTemplate(t,
    `some long string {{.Foo}} over many {{.Bar}} lines`,
    map[string]any{
      "Foo": "value",
      "Bar": 42,
    } // could also be a struct{Foo string; Bar int}
  )
```

ARR4N (approved):
Approving with a requested change that doesn't need re-review.

tsachiherman:
I think it's much better with these changes; next step, I'd suggest moving these large json structure into an external file and embed these doing buildtime. ( I wouldn't make this suggestion for few lines of json.. but these jsons are quite massive, and look like they might be changing regardless of the testing code ).

vms/avm/service_test.go
dhrubabasu: Good call, replaced.

vms/avm/service_test.go
dhrubabasu: The JSON isn't _really_ meant to be read by us. These tests only exist to ensure we don't accidentally change the JSON of any transaction unknowingly.

Moved the argument slice above.

dhrubabasu:
> I think it's much better with these changes; next step, I'd suggest moving these large json structure into an external file and embed these doing buildtime. ( I wouldn't make this suggestion for few lines of json.. but these jsons are quite massive, and look like they might be changing regardless of the testing code ).

These won't be changing at all. The reason the JSON is in the test code is to guarantee we don't make a change that accidentally changes a JSON serialization (which would impact downstream consumers, particularly in javascript land)

dhrubabasu:
> Assuming use of `json.Marshal/Unmarshal` in the code under test, would it make sense to switch to composing JSON test data in the same manner rather than via string interpolation?

I like the idea of keeping the JSON itself since Marshal versus Unmarshal could lead to an undesired change (serializing `null` instead of `[]` for example)

PR #3178, opened 2024-07-07
Include VM path in plugin version error

StephenButtolph:
## Why this should be merged

Adds the explicit file that should be updated to the error message.

## How this works

Passes the file path into the initializer.

## How this was tested

- [X] Manually:
```
[07-07|11:26:13.794] ERROR node/node.go:1287 failed to register VM {"vmID": "srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy", "error": "handshake failed: RPCChainVM protocol version mismatch between AvalancheGo and Virtual Machine plugin. AvalancheGo version v1.11.9 implements RPCChainVM protocol version 35. The VM located at /Users/stephen/.avalanchego/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy implements RPCChainVM protocol version 33. Please make sure that there is an exact match of the protocol versions. This can be achieved by updating your VM or running an older/newer version of AvalancheGo. Please be advised that some virtual machines may not yet support the latest RPCChainVM protocol version"}
```

### DIFF
--- a/vms/rpcchainvm/runtime/subprocess/initializer.go
+++ b/vms/rpcchainvm/runtime/subprocess/initializer.go
@@ -16,6 +16,8 @@ var _ runtime.Initializer = (*initializer)(nil)
 
 // Subprocess VM Runtime intializer.
 type initializer struct {
+	path string
+
 	once sync.Once
 	// Address of the RPC Chain VM server
 	vmAddr string
@@ -25,8 +27,9 @@ type initializer struct {
 	initialized chan struct{}
 }
 
-func newInitializer() *initializer {
+func newInitializer(path string) *initializer {
 	return &initializer{
+		path:        path,
 		initialized: make(chan struct{}),
 	}
 }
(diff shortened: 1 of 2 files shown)


dhrubabasu (approved):

tsachiherman (approved):
change look good. do you mind if I'll add a unit test for this ?

StephenButtolph:
> change look good. do you mind if I'll add a unit test for this ?

@tsachiherman would you want to add the unit test on top of this PR? Or in this PR itself? I'm all for adding more tests.

tsachiherman:
> > change look good. do you mind if I'll add a unit test for this ?
> 
> @tsachiherman would you want to add the unit test on top of this PR? Or in this PR itself? I'm all for adding more tests.

this PR looks ready.. I'll do it in a separate PR based on this one. ( that way, you could merge this one right away )

PR #3182, opened 2024-07-09
Rename GetPeerList and PeerList messages

won-js:
## Why this should be merged
Resolve  [#3128](https://github.com/ava-labs/avalanchego/issues/3128)

**Discussion and alternatives**
I have not changed the "peerLists" in network/network.go.
Should I change it to peers?

## How this works

- peerList -> peers
- get_peerList -> peers
- Cli flags and comments (peer-list && peerlist -> peers)


config/config.go
```diff
@@ -353,10 +353,10 @@ func getNetworkConfig(
 			ReadHandshakeTimeout: v.GetDuration(NetworkReadHandshakeTimeoutKey),
 		},
 
-		PeerListGossipConfig: network.PeerListGossipConfig{
-			PeerListNumValidatorIPs: v.GetUint32(NetworkPeerListNumValidatorIPsKey),
-			PeerListPullGossipFreq:  v.GetDuration(NetworkPeerListPullGossipFreqKey),
-			PeerListBloomResetFreq:  v.GetDuration(NetworkPeerListBloomResetFreqKey),
+		PeersGossipConfig: network.PeersGossipConfig{
```
joshua-kim, line 356: Applicable to the similar spots in this PR but we should name this `PeerGossipConfig` since `PeersGossipConfig` reads awkwardly.

joshua-kim (changes requested):

config/config.go
won-js: Thank you for reviewing. 
I am not a native English speaker, so I'm not very familiar with it. 
I have made the changes.

StephenButtolph:
Closing due to staleness. Fwiw, I think we probably still want to make this improvement, but we'll want to be holistic throughout the repo when we do rename them.

PR #3184, opened 2024-07-11
[antithesis] Refactor compose config generation to simplify reuse

marun:
## Why this should be merged

Refactor generation of compose configuration for antithesis test setups to simplify reuse by subnet-evm. 

## How this works

Creates a new function (`GenerateComposeConfig`) that encapsulates generation of docker-compose.yml and associated initial db state (if testing a VM) for any antithesis test setup.

## How this was tested

CI

### DIFF
--- a/tests/antithesis/compose.go
+++ b/tests/antithesis/compose.go
@@ -23,6 +23,13 @@ import (
 
 const bootstrapIndex = 0
 
+var (
+	errTargetPathEnvVarNotSet = errors.New("TARGET_PATH environment variable not set")
+	errImageTagEnvVarNotSet   = errors.New("IMAGE_TAG environment variable not set")
+	errAvalancheGoEvVarNotSet = errors.New("AVALANCHEGO_PATH environment variable not set")
+	errPluginDirEnvVarNotSet  = errors.New("AVALANCHEGO_PLUGIN_DIR environment variable not set")
+)
+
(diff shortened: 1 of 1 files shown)


joshua-kim (approved):

tests/antithesis/compose.go
```diff
@@ -22,9 +23,56 @@ import (
...
+	if len(network.Subnets) > 0 {
+		avalancheGoPath := os.Getenv("AVALANCHEGO_PATH")
+		if len(avalancheGoPath) == 0 {
+			return errors.New("AVALANCHEGO_PATH environment variable not set")
+		}
+
+		pluginDir := os.Getenv("AVALANCHEGO_PLUGIN_DIR")
+		if len(pluginDir) == 0 {
+			return errors.New("AVALANCHEGO_PLUGIN_DIR environment variable not set")
+		}
```
StephenButtolph, line 51: Rather than using `errors.New` directly in the `return` - can we define them as variables? There is a longstanding goal I've had to add `goerr113` as a linter... But there have been a few blockers (but this would be prevented by that linter)

StephenButtolph: In this case it isn't super important (more of a consistency style thing). But in other locations it's important to be able to test the different error conditions more naturally (with `errors.Is`)

StephenButtolph (approved):

PR #3188, opened 2024-07-15
P-chain: introducing dynamic fee calculator

abi87:
## Why this should be merged
Splitting introduction of dynamic fees into 4 PRs:

- Introduce new dynamic fee calculator
- [Update wallet to build transactions with dynamic fees](https://github.com/ava-labs/avalanchego/pull/3189)
- [Wire in new code into blocks/txs verification and building](https://github.com/ava-labs/avalanchego/pull/2703)
- [Update gas prices](https://github.com/ava-labs/avalanchego/pull/2682)

## How this works

- Introduce code to meter P-chain transactions along all fee dimensions
- Introduce fee dimensions conversion to gas
- Introduce gas cap

## How this was tested
Extra UTs. All new code is not wired in yet. Will be done in a sequent PR

PR #2703, opened 2024-02-04
wip: meter P-chain txs

abi87:

PR #2682, opened 2024-01-31
P-chain - update unit fees

abi87:

PR #3189, opened 2024-07-15
P-chain: updating wallet to use dynamic fee calculator

abi87:
## Why this should be merged
Splitting introduction of dynamic fees into 4 PRs:

- [Introduce new dynamic fee calculator](https://github.com/ava-labs/avalanchego/pull/3188)
- Update wallet to build transactions with dynamic fees
- [Wire in new code into blocks/txs verification and building](https://github.com/ava-labs/avalanchego/pull/2703)
- [Update gas prices](https://github.com/ava-labs/avalanchego/pull/2682)

## How this works

- Updated platforvm API to serve dynamic fees related stuff (config and gas cap)
- Updated wallet to use dynamic fee calculator when needed
- Updated UTs to use the new wallet interface (explicitly pass fee calculator)

## How this was tested
Extra UTs. All new code is not wired in yet. Will be done in a sequent PR

### DIFF
--- a/vms/platformvm/block/executor/helpers_test.go
+++ b/vms/platformvm/block/executor/helpers_test.go
@@ -163,6 +163,7 @@ func newEnvironment(t *testing.T, ctrl *gomock.Controller, f fork) *environment 
 		res.factory = txstest.NewWalletFactory(
 			res.ctx,
 			res.config,
+			res.clk,
 			res.state,
 		)
 	} else {
@@ -173,10 +174,12 @@ func newEnvironment(t *testing.T, ctrl *gomock.Controller, f fork) *environment 
 		res.factory = txstest.NewWalletFactory(
 			res.ctx,
 			res.config,
+			res.clk,
 			res.mockedState,
 		)
 
 		// setup expectations strictly needed for environment creation
+		res.mockedState.EXPECT().GetTimestamp().Return(time.Time{}).AnyTimes() // to initialize createSubnet/BlockchainTx fee
 		res.mockedState.EXPECT().GetLastAccepted().Return(genesisBlkID).Times(1)
 	}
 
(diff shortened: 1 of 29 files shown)


vms/platformvm/block/executor/helpers_test.go
```diff
@@ -252,7 +255,11 @@ func newEnvironment(t *testing.T, ctrl *gomock.Controller, f fork) *environment
 }
 
 func addSubnet(env *environment) {
-	builder, signer := env.factory.NewWallet(preFundedKeys[0])
+	builder, signer, feeCalc, err := env.factory.NewWallet(preFundedKeys[0])
```
tsachiherman, line 258: maybe add the `t *testing.T` to the `addSubnet` instead of panic'ing ?

vms/platformvm/service.go
```diff
@@ -1828,6 +1830,67 @@ func (s *Service) GetBlockByHeight(_ *http.Request, args *api.GetBlockByHeightAr
 	return err
 }
 
+// DynamicFeesConfigReply is the response from GetDynamicFeeConfig
+type DynamicFeesConfigReply struct {
+	commonfee.DynamicFeesConfig `json:"nextGasPrice"`
+}
+
+// GetNextFeeRates returns the next fee rates that a transaction must pay to be accepted now
+func (s *Service) GetDynamicFeeConfig(_ *http.Request, _ *struct{}, reply *DynamicFeesConfigReply) error {
```
tsachiherman, line 1839: The method, as implemented, makes me feel a bit uncomfortable. The `reply` is being used in some cases, but not in others, and the caller is expected to deduce that by examining the error code. In addition, the caller is expected to provide a valid `reply` pointer, or the method might (sometimes) panic.

I know that this aligns with the existing calling convension in this file, and was wondering if there is a good reason for that. In particular, I would that written it as 
```golang
func (s *Service) GetDynamicFeeConfig(_ *http.Request, _ *struct{}) (reply *DynamicFeesConfigReply, err error) 
```
since that would have the convention that the `err` and the `reply` are mutually exclusive.

tsachiherman (commented):
overall looks good. few small comments.

StephenButtolph:
This has been refactored into #3232 and #3286

PR #3193, opened 2024-07-16
[ci] Ensure monitoring action compatibility for other repos

marun:
The path to the notification script needs to be constructed with `github.action_path` to be compatible with other repos.


abi87 (approved):

ARR4N (approved):

tsachiherman (approved):
looks good. I was grepping ".github/action" and found quite few more instances.
would these also be good candidate for that ?

PR #3197, opened 2024-07-17
[vms/proposervm] Set build block time correctly  when anyone can propose

dhrubabasu:
## Why this should be merged

When anyone can propose, the next block build time isn't correctly set since we wrongly return that error. This can cause delays in block building. Block verification is unaffected because this is strictly the build path.

## How this works

Don't return `ErrAnyoneCanPropose`

## How this was tested

Adds a regression test

### DIFF
--- a/vms/proposervm/vm.go
+++ b/vms/proposervm/vm.go
@@ -407,7 +407,7 @@ func (vm *VM) getPostDurangoSlotTime(
 		delay = max(delay, vm.MinBlkDelay)
 		return parentTimestamp.Add(delay), nil
 	case errors.Is(err, proposer.ErrAnyoneCanPropose):
-		return parentTimestamp.Add(vm.MinBlkDelay), err
+		return parentTimestamp.Add(vm.MinBlkDelay), nil
 	default:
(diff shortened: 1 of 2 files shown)


StephenButtolph (approved):

tsachiherman (approved):
maybe I'm not reading this correctly, but wouldn't that means that all honest "AnyoneCanPropose" would never propose a block ?

StephenButtolph:
> maybe I'm not reading this correctly, but wouldn't that means that all honest "AnyoneCanPropose" would never propose a block ?

That is the bug that is being resolved by this PR yes.

tsachiherman:
> This can cause delays in block building

sounds to me like a "This can cause delays in block building" would be an understatement.

StephenButtolph:
> sounds to me like a "This can cause delays in block building" would be an understatement.

Sorry I actually mis-read your prior comment.

This bug would cause errors in resetting the block timer. So the previous block build time would always be used. Essentially, block producers could block for up to 1 hour before they would attempt to build the block.

They would still eventually build blocks.

PR #3199, opened 2024-07-17
feat: autogenerate flag documentation

ARR4N:
## Why this should be merged

It will keep Stephen happy.

## How this works

Duct tape and glue.

## How this was tested

It wasn't.


StephenButtolph:
Closing as I think this was a PoC - if we wanted to actually do this, I think we would need to make the readme a bit more searchable/linkable like the current docs.

PR #3205, opened 2024-07-19
Make math.Add64 and math.Mul64 generic

StephenButtolph:
## Why this should be merged

Resolves the asymmetry between `math.Sub` and `math.Add64`/`math.Mul64`.

## How this works

`^uint(0) == maxUint`

## How this was tested

- [X] Added test of wrap-around behavior for all unsigned int types.
- [X] CI

### DIFF
--- a/utils/math/safe_math.go
+++ b/utils/math/safe_math.go
@@ -7,8 +7,6 @@ import (
 	"errors"
 
 	"golang.org/x/exp/constraints"
-
-	"github.com/ava-labs/avalanchego/utils"
 )
 
 var (
@@ -26,9 +24,7 @@ var (
 func MaxUint[T constraints.Unsigned]() T {
 	// Unsigned integers will underflow to their max value.
 	// Ref: https://go.dev/ref/spec#Arithmetic_operators
-	var maxValue T
-	maxValue--
-	return maxValue
+	return T(0) - 1
 }
 
 // Add returns:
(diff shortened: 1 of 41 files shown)


utils/math/safe_math.go
```diff
@@ -15,17 +14,28 @@ import (
 var (
 	ErrOverflow  = errors.New("overflow")
 	ErrUnderflow = errors.New("underflow")
+
+	// Deprecated: Add64 is deprecated. Use Add[uint64] instead.
+	Add64 = Add[uint64]
+
+	// Deprecated: Mul64 is deprecated. Use Mul[uint64] instead.
+	Mul64 = Mul[uint64]
```
StephenButtolph, line 22: These are used by coreth. I was too lazy to make a PR to update the dependency with this. We should remove these once coreth has removed their usage.

abi87 (approved):

utils/math/safe_math.go
```diff
@@ -5,27 +5,33 @@ package math
...
 )
 
-// Add64 returns:
+// MaxUint returns the maximum value of an unsigned integer of type T.
+func MaxUint[T constraints.Unsigned]() T {
```
ARR4N, line 24:
```go
func MaxUint[T constraints.Unsigned]() T {
	return ^T(0)
}
```

It has the same outcome but isn't dependent on knowing the spec.

dhrubabasu (approved):

utils/math/safe_math_test.go
```diff
@@ -12,28 +12,39 @@ import (
 
 const maxUint64 uint64 = math.MaxUint64
 
-func TestAdd64(t *testing.T) {
+func TestMaxUint(t *testing.T) {
 	require := require.New(t)
 
-	sum, err := Add64(0, maxUint64)
+	require.Equal(uint(math.MaxUint), MaxUint[uint]())
```
ARR4N, line 18: `assert`s throughout. And the bike shed should be purple.

PR #3207, opened 2024-07-20
Add network upgrade config

aaronbuchwald:
This PR replaces the constants/per-networkID getter functions for network upgrade timestamps with a network upgrade schedule struct and a single function to retrieve the corresponding upgrade schedule.

- [x] Update Coreth dependency
- [x] Support configuring upgrade schedule via CLI flags


config/config.go
```diff
@@ -779,6 +780,47 @@ func getTxFeeConfig(v *viper.Viper, networkID uint32) fee.StaticConfig {
 	return genesis.GetTxFeeConfig(networkID)
 }
 
+func getUpgradeConfig(v *viper.Viper, networkID uint32) (upgrade.Config, error) {
```
aaronbuchwald, line 783: Decided this was better implemented here than broken into the `upgrade` package like for genesis.

chains/manager.go
```diff
@@ -235,6 +234,9 @@ type ManagerConfig struct {
 
 	ApricotPhase4Time            time.Time
 	ApricotPhase4MinPChainHeight uint64
+	CortinaTime                  time.Time
+	CortinaXChainStopVertexID    ids.ID
+	DurangoTime                  time.Time
```
aaronbuchwald, line 239: May be nicer to pass the whole upgrade struct through, but kept this for now since it followed the previous style.

aaronbuchwald:
This is failing due to https://github.com/ava-labs/coreth/pull/613/files#r1694251921

aaronbuchwald:
I think that 1) passing the full upgrade struct around as needed instead of individual upgrade times including adding it to `*snow.Context` for each VM and 2) updating coreth to align the default local network fork activation times with AvalancheGo should be follow up PRs.

upgrade/upgrade.go
ceyonur, line 1: how do we pass this new config to VMs? For instance Subnet-EVM also has a overridable config that depends on avago's values for networks. I'm not sure if this is strictly required to be correctly supported since this is only for custom networks (I assume for testing).

ceyonur: I think we can even disable this option if we're running any subnet.

config/config.go
```diff
@@ -796,6 +797,47 @@ func getTxFeeConfig(v *viper.Viper, networkID uint32) genesis.TxFeeConfig {
...
+		return upgrade.GetConfig(networkID), nil
+	}
+
+	switch networkID {
+	case constants.MainnetID:
+		return upgrade.Config{}, fmt.Errorf("cannot configure upgrades for mainnet networkID: %d", networkID)
+	case constants.FujiID:
+		return upgrade.Config{}, fmt.Errorf("cannot configure upgrades for fuji networkID: %d", networkID)
+	case constants.LocalID:
+		return upgrade.Config{}, fmt.Errorf("cannot configure upgrades for local networkID: %d", networkID)
```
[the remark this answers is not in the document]
StephenButtolph: We don't allow configuration of the local network genesis - so restricting this seems to align.

upgrade/upgrade.go
```diff
@@ -0,0 +1,211 @@
...
+	ApricotPhase4MinPChainHeight uint64    `json:"apricotPhase4MinPChainHeight"`
+	ApricotPhase5Time            time.Time `json:"apricotPhase5Time"`
+	ApricotPhasePre6Time         time.Time `json:"apricotPhasePre6Time"`
+	ApricotPhase6Time            time.Time `json:"apricotPhase6Time"`
+	ApricotPhasePost6Time        time.Time `json:"apricotPhasePost6Time"`
+	BanffTime                    time.Time `json:"banffTime"`
+	CortinaTime                  time.Time `json:"cortinaTime"`
+	CortinaXChainStopVertexID    ids.ID    `json:"cortinaXChainStopVertexID"`
+	DurangoTime                  time.Time `json:"durangoTime"`
+	EtnaTime                     time.Time `json:"etnaTime"`
```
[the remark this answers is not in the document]
StephenButtolph: Updated the rest of the repo to refer to `Etna`.

upgrade/upgrade.go
StephenButtolph: This will be added to the `snow.Context` in a followup.

upgrade/upgrade.go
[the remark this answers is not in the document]
StephenButtolph: I don't think it makes sense to try to provide compatibility to a placeholder.

StephenButtolph (approved):

dhrubabasu (approved):

PR #3214, opened 2024-07-22
[ci] Lint on non-test code importing packages from /tests

marun:
This avoids having to redundantly tag test code in `/tests` with `go:build test` or supply `--tags test` on every ginkgo invocation.

## How was this tested

- Negative case validated in CI
- Positive case manually validated by adding a reference to a package under `/tests` to non-test code and witnessing a failure:

```bash
Non-test Go files importing test-only packages MUST have '//go:build test' tag:
/home/me/src/avalanchego/master/main/main.go
```

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -92,8 +92,8 @@ function test_import_testing_only_in_tests {
 
   IMPORT_TESTING=$( echo "${NON_TEST_GO_FILES}" | xargs grep -lP '^\s*(import\s+)?"testing"');
   IMPORT_TESTIFY=$( echo "${NON_TEST_GO_FILES}" | xargs grep -l '"github.com/stretchr/testify');
-  IMPORT_FROM_TESTS=$( echo "${NON_TEST_GO_FILES}" | xargs grep -l '"github.com/ava-labs/tests/');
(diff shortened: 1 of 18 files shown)


joshua-kim (approved):

scripts/lint.sh
```diff
@@ -88,13 +88,14 @@ function test_interface_compliance_nil {
 
 function test_import_testing_only_in_tests {
   ROOT=$( git rev-parse --show-toplevel )
-  NON_TEST_GO_FILES=$( find "${ROOT}" -iname '*.go' ! -iname '*_test.go');
+  NON_TEST_GO_FILES=$( find "${ROOT}" -iname '*.go' ! -iname '*_test.go' ! -path "${ROOT}/tests/"'*' );
```
ARR4N, line 91: Please add a comment explaining the exclusion to anyone who reads this outside of this PR. I think it's easier than finding a better variable name than `NON_TEST_GO_FILES`, which now doesn't make sense.

ARR4N (approved):
- [ ] Approving with one change requested

This is much cleaner. Thanks for the addition!

abi87 (approved):

StephenButtolph (approved):

PR #3215, opened 2024-07-23
[antithesis] Add tmpnet support to workloads to simplify development

marun:
## Why this should be merged

Iterating on the workload for the avalanchego test setup is simple - just launch a local node. Iterating on the tests for any VM is more complicated due to needing to deploy a subnet and chain. This PR adds tmpnet integration to make iterating on an antithesis workload as simple as iterating on an e2e test.

## How this works

 - added support for starting a tmpnet network for a workload when URIs of a running network are not provided
 - implemented non-ginkgo version of TestContext to support reuse of tmpnet test helpers
 - replaced the use of `log.Fatalf` with `require` and a panic-based `require.TestingT` to ensure deferred cleanup functions can execute to enable reliable teardown of tmpnet networks
   - `log.Fatalf` calls `os.Exit` which precludes the execution of deferred functions 
 - replaces the use of `context.Background()` with a context created with `signal.NotifyContext` to support orderly shutdown when `SIGINT` (i.e. Ctrl-C) and `SIGTERM` signals are received

## How this was tested

Existing CI, plus added a new sanity check for workloads running with tmpnet

## TODO

 - [ ] Set new CI job required post-merge

### DIFF
--- a/tests/other_test_context.go
+++ b/tests/other_test_context.go
@@ -49,30 +49,33 @@ func (tc *OtherTestContext) Cleanup() {
 	}
 	tc.cleanupCalled = true
 
-	// Only exit non-zero if a panic caused cleanup or cleanup caused a panic
-	exitNonZero := false
-
+	var panicData any
 	if r := recover(); r != nil {
-		exitNonZero = true
-		if r.(string) != failNowMessage {
-			// Only print panic messages not originating from this context
-			fmt.Println(r)
+		errorString, ok := r.(string)
+		if !ok || errorString != failNowMessage {
+			// Retain the panic data to raise after cleanup
+			panicData = r
 		}
 	}
 
+	// Only exit non-zero if a cleanup caused a panic
+	exitNonZero := false
 	for _, cleanupFunc := range tc.cleanupFuncs {
 		func() {
 			// Ensure a failed cleanup doesn't prevent subsequent cleanup functions from running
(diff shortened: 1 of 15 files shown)


tests/other_test_context.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/onsi/ginkgo/v2/formatter"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+)
+
+const failNowMessage = "OtherTestContext.FailNow called"
+
+// TODO(marun) Consider a more descriptive name for this type
+type OtherTestContext struct {
```
aaronbuchwald, line 23: optional - seems like renaming this would be worthwhile in this PR

aaronbuchwald (approved):

tests/other_test_context.go
aaronbuchwald: `DefaultTestContext` ?

StephenButtolph: Also not a huge fan of `Other`... `Simple`? `Fallback`? `Default`?

StephenButtolph (approved):
lgtm - would also like to rename `OtherTestContext`... But I don't love any of my ideas

StephenButtolph (approved):

PR #3218, opened 2024-07-23
ACP 118 reference implementation

cam-schultz:
## Why this should be merged
- Adds proto definitions for `SignatureRequest` and `SignatureResponse` `AppRequest`/`AppResponse` message types as defined in [ACP-118](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/118-warp-signature-request)
- ~~Defines the canonical handler ID for `SignatureRequests` as `1`, as specified in https://github.com/avalanche-foundation/ACPs/pull/129~~
- Defines the canonical handler ID for `SignatureRequests` as `2`, as specified in https://github.com/avalanche-foundation/ACPs/pull/132
- Moves the existing definition of the handler ID for Tx Gossip messages to the `p2p` package

Does **not** go so far as to add a new package under `network/p2p` with message parsing, message marshalling, default handlers, etc. ala the [gossip](https://github.com/ava-labs/avalanchego/tree/master/network/p2p/gossip) package.

## How this works
See above

## How this was tested
CI


michaelkaplan13 (approved):
Makes sense to me, would defer to those with more context of the codebase to comment on where the definitions live within the files/packages.

tsachiherman (approved):
looks good to me, but would be happy to have others review this as well.

network/p2p/handler.go
```diff
@@ -21,6 +21,11 @@ var (
 	_ Handler = (*ValidatorHandler)(nil)
 )
 
+const (
+	TxGossipHandlerID = iota
+	SignatureRequestHandlerID
+)
```
StephenButtolph, line 27: Super nit - but typically we define `const` blocks before `var` blocks

network/p2p/handler.go
```diff
@@ -21,6 +21,11 @@ var (
 	_ Handler = (*ValidatorHandler)(nil)
 )
 
+const (
+	TxGossipHandlerID = iota
+	SignatureRequestHandlerID
```
StephenButtolph, line 26: This actually conflicts with the existing `atomicTxGossipHandlerID` used by `coreth`. We should probably update this and the ACP accordingly.

cam-schultz: Bumped to `2` by adding `AtomicTxGossipHandlerID` here. Corresponding ACP change: https://github.com/avalanche-foundation/ACPs/pull/132

proto/sdk/sdk.proto
```diff
@@ -16,3 +16,12 @@ message PullGossipResponse {
...
 }
+
+message SignatureRequest {
+  bytes message = 1;
+  bytes justification = 2;
+}
+
+message SignatureResponse {
+  bytes signature = 1;
+}
```
StephenButtolph, line 27: Can document these with links out to the ACP?

network/p2p/handler.go
```diff
@@ -15,6 +15,12 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
+const (
+	TxGossipHandlerID = iota
+	AtomicTxGossipHandlerID
+	SignatureRequestHandlerID
```
StephenButtolph, line 21: Can we document these as standards and link out to the ACP in a comment?

StephenButtolph:
After adding the couple requested comments - lgtm

StephenButtolph (approved):

PR #3222, opened 2024-07-24
Add benchmarks for add and sub fee dimensions

abi87:
## Why this should be merged
Dimensions are structured as value objects meaning that Add and Sub methods have a value receiver.
This causes receiver to be copied, but the receiver itself can be reused as the returned value.
On the contrary, if we used a pointer received, we'd avoid copying the receiver but we'd have to allocate the return value.
This PR introduces a couple of small benchmarks to understand what's the fastest alternative.
There seems not to be a significant different among the two alternative

Value receiver:
```
Benchmark_Dimensions_Add-12    	47075594	        25.24 ns/op	       0 B/op	       0 allocs/op
Benchmark_Dimensions_Sub-12    	55160683	        21.76 ns/op	       0 B/op	       0 allocs/op
```

Pointer receiver:
```
Benchmark_Dimensions_Add-12    	46841310	        25.22 ns/op	       0 B/op	       0 allocs/op
Benchmark_Dimensions_Sub-12    	55187002	        21.95 ns/op	       0 B/op	       0 allocs/op
```

## How this works
Just added a couple of benchmarks

## How this was tested
Benchmarks

### DIFF
--- a/vms/components/fee/dimensions_test.go
+++ b/vms/components/fee/dimensions_test.go
@@ -439,10 +439,17 @@ func Benchmark_Dimensions_Add(b *testing.B) {
 		{10_000, 10_000, 10_000, 10_000},
 	}
 
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
-		_, _ = lhs.Add(rhs[0], rhs[1], rhs[2], rhs[3], rhs[4], rhs[5], rhs[6])
-	}
+	b.Run("single", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = lhs.Add(rhs[0])
+		}
+	})
+
+	b.Run("multiple", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = lhs.Add(rhs[0], rhs[1], rhs[2], rhs[3], rhs[4], rhs[5], rhs[6])
+		}
+	})
 }
 
 func Benchmark_Dimensions_Sub(b *testing.B) {
@@ -456,8 +463,15 @@ func Benchmark_Dimensions_Sub(b *testing.B) {
 		{1_000, 1_000, 1_000, 1_000},
 	}
 
-	b.ResetTimer()
-	for i := 0; i < b.N; i++ {
-		_, _ = lhs.Sub(rhs[0], rhs[1], rhs[2], rhs[3], rhs[4], rhs[5])
-	}
+	b.Run("single", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = lhs.Sub(rhs[0])
+		}
+	})
+
+	b.Run("multiple", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, _ = lhs.Sub(rhs[0], rhs[1], rhs[2], rhs[3], rhs[4], rhs[5])
+		}
+	})
 }
--- a/vms/components/fee/dimensions_test.go
+++ b/vms/components/fee/dimensions_test.go
@@ -16,7 +16,7 @@ func Test_Dimensions_Add(t *testing.T) {
 	tests := []struct {
 		name        string
 		lhs         Dimensions
-		rhs         []Dimensions
+		rhs         []*Dimensions
 		expected    Dimensions
 		expectedErr error
 	}{
@@ -28,7 +28,7 @@ func Test_Dimensions_Add(t *testing.T) {
 				DBWrite:   3,
 				Compute:   4,
 			},
-			rhs: []Dimensions{
+			rhs: []*Dimensions{
 				{
 					Bandwidth: 10,
 					DBRead:    20,
@@ -52,7 +52,7 @@ func Test_Dimensions_Add(t *testing.T) {
 				DBWrite:   3,
 				Compute:   4,
 			},
-			rhs: []Dimensions{
+			rhs: []*Dimensions{
 				{
 					Bandwidth: 10,
 					DBRead:    20,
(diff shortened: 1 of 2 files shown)


vms/components/fee/dimensions_test.go
```diff
@@ -426,3 +426,46 @@ func Test_Dimensions_ToGas(t *testing.T) {
...
+		{1_000, 1_000, 1_000, 1_000},
+		{10_000, 10_000, 10_000, 10_000},
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := lhs.Add(rhs...)
+		require.NoError(err)
+	}
+	b.StopTimer()
```
ARR4N, line 449: I don't think you need to explicitly call `StopTimer()` at the end. It's for pausing (and later resuming) in the middle.

vms/components/fee/dimensions_test.go
```diff
@@ -426,3 +426,46 @@ func Test_Dimensions_ToGas(t *testing.T) {
...
+		{200, 200, 200, 200},
+		{500, 500, 500, 500},
+		{1_000, 1_000, 1_000, 1_000},
+		{10_000, 10_000, 10_000, 10_000},
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := lhs.Add(rhs...)
+		require.NoError(err)
```
ARR4N, line 447: This does a few things under the hood that may add non-negligible time on top of the benchmarked function. Consider a raw `if err != nil { b.Error...`.

ARR4N (approved):
Approving with comments that I don't need to re-review if implemented.

vms/components/fee/dimensions_test.go
abi87: Done

vms/components/fee/dimensions_test.go
abi87: Done

StephenButtolph:
> Benchmarks report zero allocations, which I understand means the compiler is able to optimize out copies.

That is not what the allocations are reporting in the benchmarks. Allocations are reporting heap allocations. Not stack copies.

vms/components/fee/dimensions_test.go
StephenButtolph: Based on the above benchmarks diffs, it seems like we are essentially benchmarking `require.NoError` rather than benchmarking `Add`.

imo - we should drop the error handling entirely. We have correctness checking in the tests for these methods already.

vms/components/fee/dimensions_test.go
```diff
@@ -426,3 +426,44 @@ func Test_Dimensions_ToGas(t *testing.T) {
...
+		{100, 100, 100, 100},
+		{200, 200, 200, 200},
+		{500, 500, 500, 500},
+		{1_000, 1_000, 1_000, 1_000},
+		{10_000, 10_000, 10_000, 10_000},
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := lhs.Add(rhs...)
```
StephenButtolph, line 446: By passing in varargs with a slice expansion, you are not copying the dimensions. The slice is passed in directly without copying.

StephenButtolph:
Was the purpose of this PR to benchmark whether to:

- use a value receiver or a pointer receiver?
- pass in pointers for the rhs?

If we change how the rhs is passed into the benchmark, I'd expect the pointer values to be faster

StephenButtolph (approved):
Definitely not going to impact performance, but might as well save ~6%.
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/vms/components/fee
                            │   old.txt   │              new.txt               │
                            │   sec/op    │   sec/op     vs base               │
_Dimensions_Add/single-12     4.469n ± 1%   4.146n ± 3%  -7.22% (p=0.000 n=10)
_Dimensions_Add/multiple-12   28.12n ± 1%   26.57n ± 5%  -5.49% (p=0.001 n=10)
_Dimensions_Sub/single-12     4.465n ± 0%   4.130n ± 3%  -7.48% (p=0.001 n=10)
_Dimensions_Sub/multiple-12   25.47n ± 1%   23.03n ± 6%  -9.60% (p=0.000 n=10)
geomean                       10.93n        10.12n       -7.46%

                            │   old.txt    │               new.txt               │
                            │     B/op     │    B/op     vs base                 │
_Dimensions_Add/single-12     0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Add/multiple-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Sub/single-12     0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Sub/multiple-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                  ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                            │   old.txt    │               new.txt               │
                            │  allocs/op   │ allocs/op   vs base                 │
_Dimensions_Add/single-12     0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Add/multiple-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Sub/single-12     0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
_Dimensions_Sub/multiple-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                  ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

PR #3226, opened 2024-07-24
deprecate uptime apis

ceyonur:
We're going to remove subnet uptime tracking and recording in P-chain state. This PR puts an deprecation message for required APIs and places.


darioush (approved):

api/info/service.go
```diff
@@ -312,11 +312,20 @@ type UptimeRequest struct {
 	SubnetID ids.ID `json:"subnetID"`
 }
 
+// Deprecated: Uptime is deprecated for Subnet Validators.
+// It will be available only for Primary Network Validators.
```
StephenButtolph, line 316: I think this applies to a number of locations throughout this PR:

This marks the whole method as deprecated. I think we should only be deprecating the `SubnetID` field inside of the `UptimeRequest` struct.

vms/platformvm/client_permissionless_validator.go
```diff
@@ -44,9 +44,13 @@ type ClientPermissionlessValidator struct {
...
 	AccruedDelegateeReward *uint64
 	DelegationFee          float32
-	Uptime                 *float32
-	Connected              *bool
-	Signer                 *signer.ProofOfPossession
+	// Deprecated: Uptime is deprecated for Subnet Validators.
+	//						 It will be available only for Primary Network Validators.
+	Uptime *float32
+	// Deprecated: Connected is deprecated for Subnet Validators.
+	//						 It will be available only for Primary Network Validators.
```
StephenButtolph, line 51: These comments look pretty weird. Was this intended?

ceyonur:
> If this is going to happen following a fork activation, I'd explicitly mention it in the deprecation message

These will be removed prior to E-Upgrade

vms/platformvm/client_permissionless_validator.go
ceyonur: nope, seems it used tabs instead of spaces

api/info/service.go
ceyonur: I don't know if there is a way to deprecate only a single param for functions like `NodeUptime(subnetID ids.ID)`. I think it's fine to omit functions and deprecate APIs. What do you say?

vms/platformvm/api/static_service.go
```diff
@@ -115,15 +115,19 @@ type PermissionlessValidator struct {
...
+	DelegationRewardOwner  *Owner       `json:"delegationRewardOwner,omitempty"`
+	PotentialReward        *json.Uint64 `json:"potentialReward,omitempty"`
+	AccruedDelegateeReward *json.Uint64 `json:"accruedDelegateeReward,omitempty"`
+	DelegationFee          json.Float32 `json:"delegationFee"`
+	ExactDelegationFee     *json.Uint32 `json:"exactDelegationFee,omitempty"`
+	// Deprecated: Uptime is deprecated for Subnet Validators.
+	//             It will be available only for Primary Network Validators.
+	Uptime *json.Float32 `json:"uptime,omitempty"`
+	// Deprecated: Connected is deprecated for Subnet Validators.
+	//             It will be available only for Primary Network Validators.
```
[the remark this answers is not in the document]
ceyonur: hmm I thought it would only do that for function annotations. Removed them.

vms/platformvm/client_permissionless_validator.go
```diff
@@ -44,9 +44,13 @@ type ClientPermissionlessValidator struct {
...
 	AccruedDelegateeReward *uint64
 	DelegationFee          float32
-	Uptime                 *float32
-	Connected              *bool
-	Signer                 *signer.ProofOfPossession
+	// Deprecated: Uptime is deprecated for Subnet Validators.
+	//             It will be available only for Primary Network Validators.
+	Uptime *float32
+	// Deprecated: Connected is deprecated for Subnet Validators.
+	//             It will be available only for Primary Network Validators.
```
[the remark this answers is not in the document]
ceyonur: done

StephenButtolph (approved):

PR #3228, opened 2024-07-24
Refactoring test implementations of NNary

yacovm:
The NNary snowflake and snowball implementations are only initialized in tests, but they were residing in production code.

I renamed the files (to preserve git blame metadata) and did some refactoring so now it's more clear which files are used for testing and which are used for production.

## Why this should be merged

Moving test code to reside under test files, reduces confusion for new contributors as it is easier for them to spot which code is used in prod and which isn't.

## How this works

This is a refactoring change, no functionality change.

## How this was tested

If the tests pass, we should be good to go.


aaronbuchwald:
Seems like there was a typo on the rename of binary slush?

yacovm:
> Could we remove the `-s` suffix from the renames?

But the file name without the `s` suffix already exists. I was trying to find something I can rename them to that doesn't conflict with an existing file name.

snow/consensus/snowball/flats_test.go
```diff
@@ -10,9 +10,10 @@ import (
 
 var _ Consensus = (*Flat)(nil)
 
-func NewFlat(factory Factory, params Parameters, choice ids.ID) Consensus {
+func NewFlat(_ Factory, params Parameters, choice ids.ID) Consensus {
```
aaronbuchwald, line 13: Could we either honor the param and keep passing in `SnowballFactory` in the one location it's used or remove it and this will just be a `NewSnowballFlat` function?

I'd lean towards honoring the param

aaronbuchwald (changes requested):

yacovm:
> Nice! I appreciate your wanting to ensure better separation between test and non-test code.
> 
> But can we hold off on this PR for now? @ARR4N is actively working on isolating test-only code into separate packages which I think is preferable to renaming the files in this PR to names of the form `*_test.go` as proposed here.
> 
> I think we would ideally limit files named `*_test.go` to those containing tests intended to be executed, and put supporting test-only code in files with regular names (to allow import) in separate packages (to simplify linting for the import of test code by non-test code).

Sure, will wait for his work to finish.

yacovm:
No problem I can always re-open it if it is still needed.

PR #3231, opened 2024-07-24
Merge unlocked stake outputs

StephenButtolph:
## Why this should be merged

- After dynamic fees, this will reduce the cost to issue the same transaction.
- This can reduce low value UTXOs in the P-chain state.

## How this works

- Previously it was possible that we would include multiple staked outputs with the exact same assetID + locktime + owner pairs. This combines those outputs into a single output.

## How this was tested

- [X] Added check that fails on master


wallet/chain/p/builder/builder.go
```diff
@@ -1103,24 +1104,15 @@ func (b *builder) spend(
 		)
 		amountsToBurn[assetID] -= amountToBurn
 
-		amountAvalibleToStake := out.Amt - amountToBurn
+		amountAvailableToStake := out.Amt - amountToBurn
```
StephenButtolph, line 1107: typo fix - got in before the sybils!)

abi87 (approved):

PR #3232, opened 2024-07-24
Implement dynamic fee builder

StephenButtolph:
## Why this should be merged

Refactored out of #3189.

## How this works

This PR adds complexity logic to the P-chain tx builder.

- The wallet supports building (pre-Etna) transactions with static fees (by setting either the complexity weights or the gas price to 0).
- The wallet supports building (post-Etna) transactions with dynamic fees (by setting the static fees to 0 and populating the complexity weights and gas price).

## How this was tested

- [x] Unit tests


wallet/chain/p/builder/builder.go
```diff
@@ -1194,3 +1493,173 @@ func (b *builder) initCtx(tx txs.UnsignedTx) error {
...
+	s.amountsToStake[assetID] -= amountToStake
+	return amountAvailableToStake - amountToStake
+}
+
+func (s *spendHelper) calculateFee() (uint64, error) {
+	gas, err := s.complexity.ToGas(s.weights)
+	if err != nil {
+		return 0, err
+	}
+	return math.Mul(uint64(s.gasPrice), uint64(gas))
```
abi87, line 1591: I really think we should reuse the proper fee Calculator here. I think this is a dangerour code duplication in case we change the way we map complexity to gas

wallet/chain/p/builder/builder.go
```diff
@@ -677,14 +810,59 @@ func (b *builder) NewImportTx(
...
+	)
+	if importedAVAX := importedAmounts[avaxAssetID]; importedAVAX > txFee {
+		toBurn = map[ids.ID]uint64{}
+		excessFee = importedAVAX - txFee
+	} else {
+		toBurn = map[ids.ID]uint64{
+			avaxAssetID: txFee - importedAVAX,
+		}
+		excessFee = 0
+	}
```
abi87, line 850: I am not sure this is correct. `txFee` is set to `b.context.BaseTxFee` but this is true only pre-Etna right?
I would expect excessFee to depend on importedAvax complexity post Etna

abi87: Same with ExportTx

abi87 (commented):
I have two main concerns:

1. import/exports seems to be using baseFee instead of duly metering importedInputs/Outputs and derive excessfee from them
2. there is some small but I believe dangerous code duplication around s.calculateFee method.

The helper is maybe made up of methods that are too fine, but I can live with it and it is definitely and improvement wrt to current state.

wallet/chain/p/builder/builder.go
```diff
@@ -962,40 +1210,32 @@ func (b *builder) spend(
...
-		lockedOut, ok := outIntf.(*stakeable.LockOut)
-		if !ok {
-			// This output isn't locked, so it will be handled during the next
-			// iteration of the UTXO set
-			continue
-		}
-		if minIssuanceTime >= lockedOut.Locktime {
-			// This output isn't locked, so it will be handled during the next
-			// iteration of the UTXO set
+	lockedUTXOs, unlockedUTXOs := splitLockedStakeableUTXOs(utxos, minIssuanceTime)
```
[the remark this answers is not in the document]
StephenButtolph: Yeah - I didn't feel like any UTXO selection optimizations needed to be in this PR. Was considering that future work.

wallet/chain/p/builder/builder.go
StephenButtolph: I modified this method so that we only convert `gas` and `gasPrice` to `nAVAX` in one location. lmk if that addresses your concern.

wallet/chain/p/builder/builder.go
abi87: I got confused into thinking that we were specifying both a fixed fee to be burned and the fee related to complexity.
Instead I believe you're levering the fact that either b.context.Fees or b.context.Weights/GasPrice are non-zero, but not both. This way the code will work both pre and post Etna.
Do I understand correctly?
If this is the case, code change is fine with me. I'd consider adding a validator of context that enforces that quantities about are not all non-zero

StephenButtolph: I don't believe that this code makes any assumptions around requiring that static fees and dynamic fees are not specified at the same time.

wallet/chain/p/builder/builder.go
```diff
@@ -677,14 +810,59 @@ func (b *builder) NewImportTx(
...
+		toBurn = map[ids.ID]uint64{}
+		excessFee = importedAVAX - txFee
+	} else {
+		toBurn = map[ids.ID]uint64{
+			avaxAssetID: txFee - importedAVAX,
+		}
+		excessFee = 0
+	}
+
+	inputs, changeOutputs, _, err := b.spend(toBurn, toStake, excessFee, complexity, ops)
```
StephenButtolph, line 852: Bug: the change address is not guaranteed to be the `to` address.

wallet/chain/p/builder/builder.go
```diff
@@ -324,7 +328,26 @@ func (b *builder) NewBaseTx(
 	toStake := map[ids.ID]uint64{}
 
 	ops := common.NewOptions(options)
-	inputs, changeOutputs, _, err := b.spend(toBurn, toStake, ops)
+	memo := ops.Memo()
+	memoComplexity := feecomponent.Dimensions{
```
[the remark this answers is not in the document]
StephenButtolph: The builder needs to account for accrued complexity while building the transaction so that it can guarantee that sufficient fees are paid.

So, this complexity calculation is being performed here rather than using `fee.TxComplexity` because the tx hasn't been created yet (this function is creating the tx). This is why things like `OutputComplexity` ect are exported.

wallet/chain/p/builder/builder.go
```diff
@@ -1175,26 +1409,37 @@ func (b *builder) spend(
...
+		}
 	}
 
 	for _, utxo := range utxosByAVAXAssetID.requested {
+		requiredFee, err := s.calculateFee()
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
 		// If we have consumed enough of the asset, then we have no need burn
```
dhrubabasu, line 1440:
```suggestion
		// If we have consumed enough of the asset, then we have no need to burn
```

dhrubabasu (approved):

wallet/chain/p/builder_test.go
```diff
@@ -48,26 +50,50 @@ var (
...
+		ComplexityWeights: feecomponent.Dimensions{
+			feecomponent.Bandwidth: 1,
+			feecomponent.DBRead:    10,
+			feecomponent.DBWrite:   100,
+			feecomponent.Compute:   1000,
+		},
+		GasPrice: 1,
+	}
+
+	tests = []struct {
```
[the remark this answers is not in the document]
StephenButtolph: Renamed to `testEnvironment`

wallet/chain/p/builder_test.go
```diff
@@ -81,35 +107,45 @@ func TestBaseTx(t *testing.T) {
...
+			var (
+				require    = require.New(t)
+				chainUTXOs = utxotest.NewDeterministicChainUTXOs(t, map[ids.ID][]*avax.UTXO{
+					constants.PlatformChainID: utxos,
+				})
+				backend = NewBackend(test.context, chainUTXOs, nil)
+				builder = builder.New(set.Of(utxoAddr), test.context, backend)
+			)
+
+			utx, err := builder.NewBaseTx([]*avax.TransferableOutput{outputToMove})
```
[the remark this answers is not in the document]
StephenButtolph: I did a pass to slim these down. It isn't super clear to me how we would use test vectors across all the transactions in a clean way... But I felt like what I ended up doing was a nice middle ground

dhrubabasu (approved):

PR #3235, opened 2024-07-25
Remove subnetIDs from uptime manager

ceyonur:
## Why this should be merged

With ACP-77, P-chain will not track uptimes; instead every VM should be tracking their own uptimes for their subnets. This removes any subnetID related uptime tracking logic from AvalancheGo/P-chain.

Should be merged after #3226 gets in a release

## How this works

* Removes subnetID params from uptime manager and related interfaces
* Removes handling of subnet uptimes in APIs and metrics
* Removes `ConnectedSubnet` internal message
* Removes `SubnetUptimes` and it's handling in p2p ping message

## How this was tested

- [x] Modified UTs
- [x] Local Testing


api/info/service.go
```diff
@@ -307,18 +307,13 @@ type UptimeResponse struct {
 	WeightedAveragePercentage json.Float64 `json:"weightedAveragePercentage"`
 }
 
-type UptimeRequest struct {
-	// if omitted, defaults to primary network
-	SubnetID ids.ID `json:"subnetID"`
-}
-
-func (i *Info) Uptime(_ *http.Request, args *UptimeRequest, reply *UptimeResponse) error {
+func (i *Info) Uptime(_ *http.Request, _ *struct{}, reply *UptimeResponse) error {
```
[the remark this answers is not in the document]
ceyonur: i guess because golang does not allow mixed named and unnamed parameters

vms/platformvm/service.go
```diff
@@ -830,13 +830,20 @@ func (s *Service) GetCurrentValidators(_ *http.Request, args *GetCurrentValidato
 
 			shares := attr.shares
 			delegationFee := avajson.Float32(100 * float32(shares) / float32(reward.PercentDenominator))
-
-			uptime, err := s.getAPIUptime(currentStaker)
-			if err != nil {
-				return err
+			var uptime *avajson.Float32
```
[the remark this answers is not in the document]
ceyonur: yes, in fact no uptimes were being tested for this API. Added them.

vms/platformvm/service.go
```diff
@@ -1828,20 +1828,21 @@ func (s *Service) GetBlockByHeight(_ *http.Request, args *api.GetBlockByHeightAr
...
-	if constants.PrimaryNetworkID != staker.SubnetID && !s.vm.TrackedSubnets.Contains(staker.SubnetID) {
-		return nil, nil
-	}
-
-	rawUptime, err := s.vm.uptimeManager.CalculateUptimePercentFrom(staker.NodeID, staker.SubnetID, staker.StartTime)
+// Returns:
+// 1) the uptime of a validator in the API format
+// 2) whether the validator is currently connected
+// 3) an error if one occurred
+func (s *Service) getAPIUptime(staker *state.Staker) (avajson.Float32, bool, error) {
```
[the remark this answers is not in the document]
ceyonur: we were using them twice but that code got removed, so I think we don't need even this to be in a function

vms/platformvm/state/state_test.go
```diff
@@ -106,9 +106,6 @@ func TestPersistStakers(t *testing.T) {
 		// with the right weight and showing the BLS key
 		checkValidatorsSet func(*require.Assertions, *state, *Staker)
 
-		// Check that node duly track stakers uptimes
-		checkValidatorUptimes func(*require.Assertions, *state, *Staker)
```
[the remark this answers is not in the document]
ceyonur: I was under impression that it was, but looking at tests we also have been explicitly using primary network ID. So readded it back

api/info/service.go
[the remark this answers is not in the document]
ceyonur: yes https://pkg.go.dev/github.com/gorilla/rpc#Server.RegisterService

snow/uptime/manager.go
```diff
@@ -105,41 +111,34 @@ func (m *manager) StopTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
...
 func (m *manager) Disconnect(nodeID ids.NodeID) error {
-	// Update every subnet that this node was connected to
-	for subnetID := range m.connections[nodeID] {
-		if err := m.updateSubnetUptime(nodeID, subnetID); err != nil {
-			return err
-		}
+	if err := m.updateUptime(nodeID); err != nil {
+		return err
 	}
+	// TODO: shall we delete the connection regardless of the error?
```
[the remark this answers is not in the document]
ceyonur: currently `database.NotFound` is the only error recovarable (returns nil in `updateUptime`). All other errors will be non-recovarable (if I'm not missing anything). but again if program does recover at some point, the node will be stay marked as connected. I don't think failing setting the uptime should keep it marked as connected.

proto/p2p/p2p.proto
```diff
@@ -64,16 +64,7 @@ message Message {
...
-  repeated SubnetUptime subnet_uptimes = 2;
-}
-
-// SubnetUptime is a descriptor for a peer's perceived uptime on a subnet.
-message SubnetUptime {
-  // Subnet the peer is validating
-  bytes subnet_id = 1;
-  // Uptime percentage on the subnet [0, 100]
-  uint32 uptime = 2;
+  reserved 2; // Until E upgrade is activated.
```
darioush, line 67: We should test a node with this ping behavior against existing nodes (esp on a node that tracks some subnets).

snow/uptime/manager.go
```diff
@@ -68,33 +69,38 @@ func (m *manager) StartTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
...
 }
 
-func (m *manager) StopTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
+func (m *manager) StopTracking(nodeIDs []ids.NodeID) error {
+	// TODO: this was not here before, should we add it?
+	if !m.startedTracking {
+		return nil
+	}
+	defer func() {
+		m.startedTracking = false
```
darioush, line 86: it seems that `startedTracking` replaces `trackedSubnets` but only for the primary network.
but I don't see any code in the previous revision for `m.trackesSubnets.Remove`

vms/platformvm/service.go
```diff
@@ -830,13 +830,27 @@ func (s *Service) GetCurrentValidators(_ *http.Request, args *GetCurrentValidato
...
 			delegationFee := avajson.Float32(100 * float32(shares) / float32(reward.PercentDenominator))
-
-			uptime, err := s.getAPIUptime(currentStaker)
-			if err != nil {
-				return err
+			var uptime *avajson.Float32
+			var connected *bool
+			// Only calculate uptime for primary network validators
+			// TODO: decide whether we want to keep connected for subnet validators
+			// it should be available at this point
```
darioush, line 837: what is the downside of keeping it?

message/ops.go
```diff
@@ -120,7 +119,6 @@ var (
 		CrossChainAppErrorOp,
 		CrossChainAppResponseOp,
 		ConnectedOp,
-		ConnectedSubnetOp,
```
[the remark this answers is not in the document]
ceyonur: yes

snow/uptime/manager.go
```diff
@@ -20,42 +19,44 @@ type Manager interface {
...
 
 type manager struct {
 	// Used to get time. Useful for faking time during tests.
 	clock *mockable.Clock
 
-	state          State
-	connections    map[ids.NodeID]map[ids.ID]time.Time // nodeID -> subnetID -> time
-	trackedSubnets set.Set[ids.ID]
+	state       State
+	connections map[ids.NodeID]time.Time // nodeID  -> time
```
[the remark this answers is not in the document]
ceyonur: updated the comment

proto/p2p/p2p.proto
ceyonur: I did run a local network with an old node + new nodes and I also bootstrapped fuji with this new code.

snow/uptime/manager.go
ceyonur: I think this was an issue previously. we typically use this function after shutdown, so we never restart tracking. but I think the code itself should support restarting in any case.

vms/platformvm/service.go
ceyonur: we won't be able to show this info. but after removal of connectedsubnet, we won't have this info at all. removed the comment.

darioush (approved):

snow/uptime/manager.go
```diff
@@ -68,33 +69,36 @@ func (m *manager) StartTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
...
-	m.trackedSubnets.Add(subnetID)
+	m.startedTracking = true
 	return nil
 }
 
-func (m *manager) StopTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
+func (m *manager) StopTracking(nodeIDs []ids.NodeID) error {
+	// TODO: this was not here before, should we add it?
+	if !m.startedTracking {
+		return nil
```
StephenButtolph, line 83: If we are adding this handling - it seems like we should probably error here no? (Although I'm not sure the current code is correct if we were to do that)

snow/uptime/manager.go
```diff
@@ -105,41 +109,38 @@ func (m *manager) StopTracking(nodeIDs []ids.NodeID, subnetID ids.ID) error {
...
 func (m *manager) Disconnect(nodeID ids.NodeID) error {
-	// Update every subnet that this node was connected to
-	for subnetID := range m.connections[nodeID] {
-		if err := m.updateSubnetUptime(nodeID, subnetID); err != nil {
-			return err
-		}
+	return m.disconnect(nodeID)
+}
+
+func (m *manager) disconnect(nodeID ids.NodeID) error {
```
StephenButtolph, line 133:
```suggestion
```

snow/uptime/manager.go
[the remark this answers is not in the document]
StephenButtolph: If this function returns an error - the node is going to exit with a FATAL.

vms/platformvm/block/executor/options.go
```diff
@@ -175,10 +175,8 @@ func (o *options) prefersCommit(tx *txs.Tx) (bool, error) {
 		expectedUptimePercentage = float64(transformSubnet.UptimeRequirement) / reward.PercentDenominator
 	}
```
StephenButtolph, line 176: I think we can just use the primary network uptime requirement here now as well.

vms/platformvm/service.go
```diff
@@ -847,13 +847,24 @@ func (s *Service) GetCurrentValidators(_ *http.Request, args *GetCurrentValidato
 
 			shares := attr.shares
 			delegationFee := avajson.Float32(100 * float32(shares) / float32(reward.PercentDenominator))
-
-			uptime, err := s.getAPIUptime(currentStaker)
-			if err != nil {
-				return err
+			var uptime *avajson.Float32
+			var connected *bool
```
StephenButtolph, line 851: nit: can we use a var block here?

vms/platformvm/block/executor/options.go
StephenButtolph: Thinking about this again - I think that change would make more sense in a different PR.

snow/uptime/manager.go
ceyonur: I added errors and checks with `StartedTracking` so we don't recall them twice. in case bootstrapping/shutdown is called multiple times + some test helpers.

StephenButtolph (approved):

PR #3239, opened 2024-07-26
Refactor cache implementations

StephenButtolph:
## Why this should be merged

The `cache` package is currently a bit odd - as there is a `metercacher` sub-package (similar to the `database` implementations). But `lru` is implemented in the top level `cache` package.

This moves the `lru` implementations out of the `cache` package into their own package (to align with the database structure)

## How this works

- Moves code around
- Deprecates some code
- Unexports as much code as possible

## How this was tested

Existing tests


cache/cachetest/cacher.go
```diff
@@ -23,11 +23,11 @@ var Tests = []struct {
 	Size int
 	Func func(t *testing.T, c cache.Cacher[ids.ID, int64])
 }{
-	{Size: 1, Func: TestBasic},
-	{Size: 2, Func: TestEviction},
+	{Size: 1, Func: Basic},
+	{Size: 2, Func: Eviction},
 }
 
-func TestBasic(t *testing.T, cache cache.Cacher[ids.ID, int64]) {
```
joshua-kim, line 30: Why are these renamed? Is this because we felt the `cachetest.Test*` is stuttering?

StephenButtolph: Yes

alarso16 (approved):
lgtm

PR #3240, opened 2024-07-27
Simplify static fee calculations

StephenButtolph:
## Why this should be merged

The main goal of this PR is to simplify the `fee.NewStaticCalculator` signature. Previously the signature was:
```golang
func NewStaticCalculator(
	config StaticConfig,
	upgradeTimes upgrade.Config,
	chainTime time.Time,
) Calculator
```
Now it is:
```golang
func NewStaticCalculator(config StaticConfig) Calculator
```

## How this works

This is done by removing `CreateAssetFee` from the `StaticConfig` and expecting `CreateSubnetTxFee` and `CreateBlockchainTxFee` to be set by the caller.

## How this was tested

- [X] CI
- [x] Bootstrapped P-chain


vms/platformvm/txs/fee/static_calculator.go
```diff
@@ -70,30 +63,12 @@ func (c *staticVisitor) AddDelegatorTx(*txs.AddDelegatorTx) error {
...
-func (c *staticVisitor) AdvanceTimeTx(*txs.AdvanceTimeTx) error {
-	c.fee = 0 // no fees
-	return nil
-}
-
-func (c *staticVisitor) RewardValidatorTx(*txs.RewardValidatorTx) error {
-	c.fee = 0 // no fees
+	c.fee = c.config.CreateSubnetTxFee
 	return nil
 }
```
StephenButtolph, line 73: These are replaced with more explicit errors rather than returning no fees.

wallet/chain/p/builder/context.go
```diff
@@ -12,21 +12,15 @@ import (
...
-	CreateSubnetTxFee             uint64
-	TransformSubnetTxFee          uint64
-	CreateBlockchainTxFee         uint64
-	AddPrimaryNetworkValidatorFee uint64
-	AddPrimaryNetworkDelegatorFee uint64
-	AddSubnetValidatorFee         uint64
-	AddSubnetDelegatorFee         uint64
+	NetworkID       uint32
+	AVAXAssetID     ids.ID
+	StaticFeeConfig fee.StaticConfig
```
StephenButtolph, line 23: Now that the config doesn't have the `CreateAssetFee` - we can just use the config here directly 🥳

vms/platformvm/txs/fee/calculator_test.go
```diff
@@ -3,251 +3,127 @@
...
-	// chain times needed to have specific upgrades active
-	preEUpgradeTime := upgrades.EUpgradeTime.Add(-1 * time.Second)
-	preApricotPhase3Time := upgrades.ApricotPhase3Time.Add(-1 * time.Second)
-
-	tests := []struct {
-		name       string
-		chainTime  time.Time
-		unsignedTx func() txs.UnsignedTx
-		expected   uint64
+	txTests = []struct {
```
StephenButtolph, line 20: These tests will be re-used for the dynamic fees tests.

vms/platformvm/state/chain_time_helpers.go
```diff
@@ -77,5 +77,16 @@ func GetNextStakerChangeTime(state Chain) (time.Time, error) {
...
+// NewStaticFeeCalculator creates a static fee calculator, with the config set
+// to either the pre-AP3 or post-AP3 config.
+func NewStaticFeeCalculator(cfg *config.Config, timestamp time.Time) fee.Calculator {
+	config := cfg.StaticFeeConfig
+	if !cfg.UpgradeConfig.IsApricotPhase3Activated(timestamp) {
+		config.CreateSubnetTxFee = cfg.CreateAssetTxFee
+		config.CreateBlockchainTxFee = cfg.CreateAssetTxFee
+	}
+	return fee.NewStaticCalculator(config)
 }
```
StephenButtolph, line 92: I'm not convinced that this is the right place for these methods... But will leave that decision to a later PR.

vms/platformvm/txs/fee/calculator_test.go
```diff
@@ -3,251 +3,127 @@
...
+			expectedStaticFee:    0,
+			expectedStaticFeeErr: ErrUnsupportedTx,
 		},
 		{
-			name:       "CreateChainTx pre ApricotPhase3",
-			chainTime:  preApricotPhase3Time,
-			unsignedTx: createChainTx,
-			expected:   feeTestsDefaultCfg.CreateAssetTxFee,
+			name:                 "RewardValidatorTx",
+			tx:                   "0000000000143d0ad12b8ee8928edf248ca91ca55600fb383f07c32bff1d6dec472b25cf59a700000000",
```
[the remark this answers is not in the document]
StephenButtolph: Do you think it would be better to move these transactions to separate files (and embed them?). These are really more intended to be compliance tests that should never change... I feel like if we included them in their non-binary forms... Their definition would dominate the tests... When really I think of them as amorphous tests to prevent unexpected changes.

dhrubabasu (approved):

PR #3243, opened 2024-07-29
Cleanup predurango proposervm

tsachiherman:
## Why this should be merged

This PR removes from the proposerVM all the pre-durango block building support.
We're far since the pre-durango era, and our codebase have no reason to create pre-durango blocks, for any reason anymore. This cleanup eliminates the code that is no longer is use.

( i.e. the code was heavily in use within the unit testing.. but not being used in production ).

## How this works
The proposerVM itself is now preventing setting it's state to NormalOp unless we've already passed the Durango activation time. That ensures that all subsequent building / validation would have the "freedom" of acting as if that point in time have already surpassed.

## How this was tested

Tested using unit tests, e2e tests


StephenButtolph:
Closing as this is stale/not passing CI

PR #3244, opened 2024-07-29
Rename Transitive snowman to Engine snowman

yacovm:
## Why this should be merged

In the package snow/engine/snowman, the engine is called Transitive, and resides in transitive.go. while in the package snow/engine/avalanche it's respectively Engine and engine.go.

Moreover, in the tests of snow/engine/snowman/, the bespoken object is referred to as engine.

I therefore rename the transitive to engine to better fit the rest of the code.

## How this works

Just a file and struct renaming

## How this was tested

If the unit tests pass, this should be good to go.

### DIFF
--- a/snow/engine/snowman/engine.go
+++ b/snow/engine/snowman/engine.go
@@ -33,15 +33,15 @@ import (
 
 const nonVerifiedCacheSize = 64 * units.MiB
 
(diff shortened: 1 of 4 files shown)


StephenButtolph (approved):

aaronbuchwald (approved):
Huge

PR #3248, opened 2024-07-30
Add fee.State to P-chain state

StephenButtolph:
## Why this should be merged

Dynamic fees needs to track the capacity and excess usage of the chain for rate limiting and fee pricing.

## How this works

Adds `GetFeeState` and `SetFeeState` to the `state.Chain` interface.

## How this was tested

- [X] Added unit tests


vms/platformvm/state/state.go
```diff
@@ -2440,3 +2468,27 @@ func (s *state) ReindexBlocks(lock sync.Locker, log logging.Logger) error {
...
+	if err != nil {
+		return err
+	}
+	return db.Put(FeeStateKey, feeStateBytes)
+}
+
+func getFeeState(db database.KeyValueReader) (fee.State, error) {
+	feeStateBytes, err := db.Get(FeeStateKey)
+	if err == database.ErrNotFound {
+		return fee.State{}, nil
```
abi87, line 2483: q: shouldn't the capacity be max here? I believe this case happens only before etna is activated. In such a casa shouldn't capacity be max instead of zero?

abi87 (commented):
I am not sure we init capacity in the right way when Etna is not activated yet.

vms/platformvm/state/state.go
StephenButtolph: The ACP doesn't specify this... I think that it makes more sense to have it be set to 0 so that the fork activation doesn't have a complexity spike (although I doubt this would really matter).

I'll make a PR to the ACP to clarify this.

StephenButtolph: Added this to the spec: https://github.com/avalanche-foundation/ACPs/pull/136. Feel free to comment on that PR if you think it should be a different value.

abi87 (approved):

PR #3236, opened 2024-07-26
[vms/platformvm] Add `VerifyWithContext` to `Block`s

dhrubabasu:
## Why this should be merged

For ACP-77, we will need the ProposerVM's `PChainHeight` to verify warp signatures. This takes the first step towards being able to use it by implementing the `block.WithVerifyContext` interface on P-Chain blocks.

## How this works

Implement `block.WithVerifyContext`

## How this was tested

CI

PR #3252, opened 2024-07-30
Inline `verifier` struct creation

dhrubabasu:
## Why this should be merged

Required for https://github.com/ava-labs/avalanchego/pull/3236

We'll need the proposerVMCtx to be passed into `verifier` which is nicer to construct on the fly.

## How this works

Inline `verifier` struct creation

## How this was tested

CI

### DIFF
--- a/vms/platformvm/block/executor/block.go
+++ b/vms/platformvm/block/executor/block.go
@@ -29,7 +29,10 @@ func (b *Block) Verify(context.Context) error {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


tsachiherman (commented):
the only downside I can see here is the fact that if we were to have a dozen of places where `verifier` is being constructed in that way ***and*** we add another member to `verifier`, than we might miss adding that field to all the instances.

StephenButtolph (approved):

PR #3256, opened 2024-08-01
Fix comparison comment in snowflake algorithms

yacovm:
## Why this should be merged

In the snowflake algorithms, there is the following comment:

```
// terminationConditions[i].alphaConfidence < terminationConditions[i+1].alphaConfidence 
// terminationConditions[i].beta <= terminationConditions[i+1].beta
```
However, the two lines contradict, as if alpha confidence grows with i, then beta should decrease and not increase. The reason is that the higher our confidence, the less consecutive polls we need to finalize.

## How this works

Just a comment fix

## How this was tested

No need to test, it's a comment fix.

### DIFF
--- a/snow/consensus/snowball/binary_snowflake.go
+++ b/snow/consensus/snowball/binary_snowflake.go
@@ -20,7 +20,7 @@ func newBinarySnowflake(alphaPreference int, terminationConditions []termination
 // Invariant:
 // len(terminationConditions) == len(confidence)
 // terminationConditions[i].alphaConfidence < terminationConditions[i+1].alphaConfidence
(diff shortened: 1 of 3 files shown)


StephenButtolph (approved):

aaronbuchwald (approved):
LGTM - nice catch

StephenButtolph:
🤔 I wonder if we broke the merge queue for PRs coming from forks

yacovm:
> 🤔 I wonder if we broke the merge queue for PRs coming from forks

it's not my first PR from a fork, maybe there could be another explanation?

PR #3258, opened 2024-08-01
Remove trackedSubnet check for explicitly named peers in network.Send()

iansuvak:
## Why this should be merged

This would close ava-labs/awm-relayer#400 . This removes the requirement for an explicitly named peer in `common.SendConfig` to be actively tracking the relevant subnet. This doesn't affect the allower functionality or any sampled peers.  

The particular case this would solve is let nodes respond to `SignatureRequest` `AppRequests` from nodes that haven't explicitly declared that they are tracking the relevant subnet which seems like a valid usecase. 

The function comment was also updated since it seemed to be out of date, but the update isn't relevant to the functional change since filtering out peers based on their tracked subnets was undocumented functionality. 

## How this works

It removes the trackedSubnet check when checking explicitly named peers supplied to `getPeers` function which is called from `network.Send`

## How this was tested
CI + confirmed that it allows signature aggregation on the awm-relayer E2E tests (both relayer and signature aggregator)


network/network.go
```diff
@@ -698,10 +698,6 @@ func (n *network) getPeers(
 			continue
 		}
 
-		if trackedSubnets := peer.TrackedSubnets(); !trackedSubnets.Contains(subnetID) {
-			continue
-		}
```
StephenButtolph, line 703: After going through the code - the one potential downside to this PR is:

If a subnet validator forgets to `--track-subnet` the subnet that it is validating, the nodes querying them will now need to timeout the request rather than immediately marking the request as failed.

I think that is sufficiently niche that it isn't a concern to merge this PR... But it could be something to keep in mind as a future optimization. This is not a change in any security assumptions, as byzantine nodes could always report that they are tracking the subnet but then just never respond to the request.

StephenButtolph (approved):

network/network.go
```diff
@@ -678,10 +678,10 @@ func (n *network) track(ip *ips.ClaimedIPPort) error {
...
 //   - [nodeIDs] the IDs of the peers that should be returned if they are
 //     connected.
-//   - [subnetID] the subnetID whose membership should be considered if
-//     [validatorOnly] is set to true.
-//   - [validatorOnly] is the flag to drop any nodes from [nodeIDs] that are not
-//     validators in [subnetID].
+//   - [subnetID] the subnetID whose membership should be considered to
+//     determine if the node is a validator.
+//   - [allower] interface that determines if a node is allowed to connect to
+//     the subnet based on it's validator status.
```
dhrubabasu, line 684:
```suggestion
//     the subnet based on its validator status.
```

dhrubabasu (approved):

PR #3259, opened 2024-08-02
[ci] Switch to v2 of docker compose plugin

marun:
## Why this should be merged

As per https://github.com/actions/runner-images/issues/9692, the default runner image no longer includes the docker compose v1 plugin (`docker-compose` command). The v2 plugin (`docker compose` command) is provided, so this change updates the antithesis image build test to use that instead.

## How this was tested

CI


StephenButtolph (approved):
Thanks for looking into this so quickly!

darioush (approved):

aaronbuchwald (approved):

PR #3263, opened 2024-08-02
Minor code quality cleanup for snow/consensus

yacovm:
## Why this should be merged

- `snowflakeFactory` was un-used, renamed to `_`

- `SnowballFactory` was stuttering (is in snowball package), renamed to `FactoryInstance`

- Fixed grammar mistake in `snow/consensus/snowman/poll/set.go`

## How this works

Just a minor code cleanup

## How this was tested

CI


snow/consensus/snowball/factory.go
```diff
@@ -6,8 +6,8 @@ package snowball
 import "github.com/ava-labs/avalanchego/ids"
 
 var (
-	SnowballFactory  Factory = snowballFactory{}
-	SnowflakeFactory Factory = snowflakeFactory{}
+	FactoryInstance Factory = snowballFactory{}
+	_               Factory = snowflakeFactory{}
```
StephenButtolph, line 10: @aaronbuchwald Do we use `SnowflakeFactory` in some other repo?

I think we added this to test out the `Tree` using just `Snowflake` rather than using `Snowball`... Does this change break anything?

meaghanfitzgerald: > Do we use `SnowflakeFactory` in some other repo?

afaict according to [this search](https://github.com/search?q=org%3Aava-labs+SnowflakeFactory&type=code) across ava-labs repos, it is only used in avalanchego

aaronbuchwald: iirc I was using this in the simulator repo, but am not continuing that work, so it's fine to remove imo

snow/consensus/snowball/consensus_performance_test.go
```diff
@@ -28,10 +28,10 @@ func TestDualAlphaOptimization(t *testing.T) {
 		source        = prng.NewMT19937()
 	)
 
-	singleAlphaNetwork := NewNetwork(SnowballFactory, params, numColors, source)
```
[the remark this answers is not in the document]
yacovm: Factory is already [taken](https://github.com/ava-labs/avalanchego/blob/9a89b5e802bd740fce00bfaa1de88875ba301440/snow/consensus/snowball/consensus.go#L44).

PR #3266, opened 2024-08-02
Expose upgrade config in the info API

StephenButtolph:
## Why this should be merged

This will be used in the E2E testing to determine which ruleset should be used when executing tests.

## How this works

Exposes the `upgrade.Config` through a new `info.upgrades` API.

## How this was tested

- Manually

### DIFF
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -14,7 +14,7 @@ import (
 
 var (
 	InitiallyActiveTime       = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
-	UnscheduledActivationTime = time.Date(10000, time.December, 1, 0, 0, 0, 0, time.UTC)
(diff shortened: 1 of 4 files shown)


upgrade/upgrade.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	InitiallyActiveTime       = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
-	UnscheduledActivationTime = time.Date(10000, time.December, 1, 0, 0, 0, 0, time.UTC)
+	UnscheduledActivationTime = time.Date(9999, time.December, 1, 0, 0, 0, 0, time.UTC)
```
StephenButtolph, line 17: JSON marshaling actually doesn't support timestamps with years `>9999` because the format is fixed width.

dhrubabasu (approved):

PR #3268, opened 2024-08-04
Separate e2e tests by etna activation

StephenButtolph:
## Why this should be merged

This converts the single e2e test network (which currently runs with the Etna upgrade activated) into two e2e test networks:
1. e2e tests with Etna activated
2. e2e tests without Etna activated
All the tests run on each network - if a test is upgrade dependent, then it should check the network setup internally.

## How this works

- Adds `--activate-etna` to the tmpnet flags.
- Adds a new test with `--activate-etna` provided.

## How this was tested

- [X] CI
  - https://github.com/ava-labs/avalanchego/actions/runs/10239550091/job/28325281174?pr=3268#step:5:605
  - https://github.com/ava-labs/avalanchego/actions/runs/10239550091/job/28325280991?pr=3268#step:5:2982

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       - name: fuzz_test
         shell: bash
         run: ./scripts/build_fuzz.sh 10 # Run each fuzz test 10 seconds
-  e2e:
+  e2e_pre_etna:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


StephenButtolph:
> Is this test intended to exist long-term, or is it temporarily validating configuration of etna enablement in advance of future work

It was primarily added here to serve as an example of how to detect+skip if Etna is active (for the future e2e tests we are going to build out).

I also felt like it may be useful as we add additional Etna functionality to expand on this test... but if we don't think this test is useful in its current form, we can remove it.

PR #3280, opened 2024-08-08
Add StoppedTimer helper

marun:
As per @StephenButtolph's [recent comment](https://github.com/ava-labs/subnet-evm/pull/1166#discussion_r1688145358).

### DIFF
--- a/network/throttling/inbound_conn_upgrade_throttler.go
+++ b/network/throttling/inbound_conn_upgrade_throttler.go
@@ -12,7 +12,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/utils/timer/mockable"
 
-	utils_timer "github.com/ava-labs/avalanchego/utils/timer"
+	timerpkg "github.com/ava-labs/avalanchego/utils/timer"
 )
 
 var (
@@ -133,7 +133,7 @@ func (n *inboundConnUpgradeThrottler) ShouldUpgrade(addrPort netip.AddrPort) boo
 }
 
 func (n *inboundConnUpgradeThrottler) Dispatch() {
-	timer := utils_timer.StoppedTimer()
+	timer := timerpkg.StoppedTimer()
 
 	defer timer.Stop()
 	for {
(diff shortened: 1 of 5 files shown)


aaronbuchwald (approved):

network/throttling/inbound_conn_upgrade_throttler.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/utils/timer/mockable"
+
+	utils_timer "github.com/ava-labs/avalanchego/utils/timer"
```
ARR4N, line 15: (optional, subjective) When a package and a local variable clash, [one recommended approach](https://google.github.io/styleguide/go/decisions#:~:text=preferred%20way%20to%20do%20so%20is%20with%20the%20pkg%20suffix%20(e.g.%20urlpkg).) is to use the `pkg` suffix so this would be `timerpkg`.

The underscore in the name is what jumped out at me as out of the ordinary.

ARR4N (approved):
Approving with minor requests that don't require re-review.

utils/timer/stopped_timer.go
```diff
@@ -0,0 +1,24 @@
...
+// the C channel (and there isn't one scheduled to be added).
+//
+// This means that after calling Reset there will be no events on the
+// channel until the timer fires (at which point there will be exactly
+// one event sent to the channel).
+
+// It enables re-using the timer across loop iterations without
+// needing to have the first loop iteration perform any == nil checks
+// to initialize the first invocation.
+func StoppedTimer() *time.Timer {
```
StephenButtolph, line 18: I think this should also be used here:
https://github.com/ava-labs/avalanchego/blob/6626d2b7c1d5a108b12504a0d491f6cebd9dcb57/network/throttling/inbound_resource_throttler.go#L109-L114

StephenButtolph (commented):
lgtm after the few nits are addressed

StephenButtolph (approved):

PR #3282, opened 2024-08-08
Refactor P-chain Builder

StephenButtolph:
## Why this should be merged

It seems like #3232 is daunting to review - this PR further reduces that diff by refactoring the builder without adding complexity in yet.

## How this works

- This PR refactors `ImportTx` to always call `spend`.
- This PR refactors all the transactions to call `spend` last.
- This PR introduces a `spendHelper` to encapsulate common operations performed in `spend`
- This PR introduces a few helper functions that separate UTXOs into different groups.

## How this was tested

- [X] This PR is a pure refactor and passes existing CI
- [X] Additional helpers are explicitly tested

### DIFF
--- a/utils/slice.go
+++ b/utils/slice.go
@@ -3,7 +3,7 @@
 
 package utils
 
-// join the provided slices into a single slice.
+// Join merges the provided slices into a single slice.
 //
 // TODO: Use slices.Concat once the minimum go version is 1.22.
 func Join[T any](slices ...[]T) []T {
--- a/utils/slice.go
+++ b/utils/slice.go
@@ -1,0 +1,19 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package utils
+
+// join the provided slices into a single slice.
+//
+// TODO: Use slices.Concat once the minimum go version is 1.22.
+func Join[T any](slices ...[]T) []T {
+	size := 0
+	for _, s := range slices {
+		size += len(s)
+	}
+	newSlice := make([]T, 0, size)
+	for _, s := range slices {
+		newSlice = append(newSlice, s...)
+	}
+	return newSlice
+}
--- a/wallet/chain/p/builder/builder.go
+++ b/wallet/chain/p/builder/builder.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/components/verify"
 	"github.com/ava-labs/avalanchego/vms/platformvm/fx"
 	"github.com/ava-labs/avalanchego/vms/platformvm/signer"
 	"github.com/ava-labs/avalanchego/vms/platformvm/stakeable"
@@ -324,7 +325,13 @@ func (b *builder) NewBaseTx(
 	toStake := map[ids.ID]uint64{}
 
 	ops := common.NewOptions(options)
-	inputs, changeOutputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, changeOutputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -355,7 +362,13 @@ func (b *builder) NewAddValidatorTx(
 		avaxAssetID: vdr.Wght,
 	}
 	ops := common.NewOptions(options)
-	inputs, baseOutputs, stakeOutputs, err := b.spend(toBurn, toStake, ops)
+	inputs, baseOutputs, stakeOutputs, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -386,7 +399,13 @@ func (b *builder) NewAddSubnetValidatorTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -420,7 +439,13 @@ func (b *builder) NewRemoveSubnetValidatorTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -458,7 +483,13 @@ func (b *builder) NewAddDelegatorTx(
 		avaxAssetID: vdr.Wght,
 	}
 	ops := common.NewOptions(options)
-	inputs, baseOutputs, stakeOutputs, err := b.spend(toBurn, toStake, ops)
+	inputs, baseOutputs, stakeOutputs, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -492,7 +523,13 @@ func (b *builder) NewCreateChainTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -530,7 +567,13 @@ func (b *builder) NewCreateSubnetTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -559,7 +602,13 @@ func (b *builder) NewTransferSubnetOwnershipTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -645,29 +694,12 @@ func (b *builder) NewImportTx(
 		)
 	}
 
-	var (
-		inputs       []*avax.TransferableInput
-		outputs      = make([]*avax.TransferableOutput, 0, len(importedAmounts))
-		importedAVAX = importedAmounts[avaxAssetID]
-	)
-	if importedAVAX > txFee {
-		importedAmounts[avaxAssetID] -= txFee
-	} else {
-		if importedAVAX < txFee { // imported amount goes toward paying tx fee
-			toBurn := map[ids.ID]uint64{
-				avaxAssetID: txFee - importedAVAX,
-			}
-			toStake := map[ids.ID]uint64{}
-			var err error
-			inputs, outputs, _, err = b.spend(toBurn, toStake, ops)
-			if err != nil {
-				return nil, fmt.Errorf("couldn't generate tx inputs/outputs: %w", err)
-			}
-		}
-		delete(importedAmounts, avaxAssetID)
-	}
-
+	outputs := make([]*avax.TransferableOutput, 0, len(importedAmounts))
 	for assetID, amount := range importedAmounts {
+		if assetID == avaxAssetID {
+			continue
+		}
+
 		outputs = append(outputs, &avax.TransferableOutput{
 			Asset: avax.Asset{ID: assetID},
 			Out: &secp256k1fx.TransferOutput{
@@ -676,6 +708,33 @@ func (b *builder) NewImportTx(
 			},
 		})
 	}
+
+	var (
+		toBurn    map[ids.ID]uint64
+		toStake   = map[ids.ID]uint64{}
+		excessFee uint64
+	)
+	if importedAVAX := importedAmounts[avaxAssetID]; importedAVAX > txFee {
+		toBurn = map[ids.ID]uint64{}
+		excessFee = importedAVAX - txFee
+	} else {
+		toBurn = map[ids.ID]uint64{
+			avaxAssetID: txFee - importedAVAX,
+		}
+		excessFee = 0
+	}
+
+	inputs, changeOutputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		excessFee,
+		to,
+		ops,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't generate tx inputs/outputs: %w", err)
+	}
+	outputs = append(outputs, changeOutputs...)
 
 	avax.SortTransferableOutputs(outputs, txs.Codec) // sort imported outputs
 	tx := &txs.ImportTx{
@@ -711,7 +770,13 @@ func (b *builder) NewExportTx(
 
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, changeOutputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, changeOutputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -754,7 +819,13 @@ func (b *builder) NewTransformSubnetTx(
 	}
 	toStake := map[ids.ID]uint64{}
 	ops := common.NewOptions(options)
-	inputs, outputs, _, err := b.spend(toBurn, toStake, ops)
+	inputs, outputs, _, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -811,7 +882,13 @@ func (b *builder) NewAddPermissionlessValidatorTx(
 		assetID: vdr.Wght,
 	}
 	ops := common.NewOptions(options)
-	inputs, baseOutputs, stakeOutputs, err := b.spend(toBurn, toStake, ops)
+	inputs, baseOutputs, stakeOutputs, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -854,7 +931,13 @@ func (b *builder) NewAddPermissionlessDelegatorTx(
 		assetID: vdr.Wght,
 	}
 	ops := common.NewOptions(options)
-	inputs, baseOutputs, stakeOutputs, err := b.spend(toBurn, toStake, ops)
+	inputs, baseOutputs, stakeOutputs, err := b.spend(
+		toBurn,
+		toStake,
+		0,
+		nil,
+		ops,
+	)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


dhrubabasu (approved):

wallet/chain/p/builder/builder.go
```diff
@@ -1200,3 +1293,136 @@ func (b *builder) initCtx(tx txs.UnsignedTx) error {
...
+
+	// Burn any value that should be burned
+	amountToBurn := min(
+		remainingAmountToBurn, // Amount we still need to burn
+		amount,                // Amount available to burn
+	)
+	s.amountsToBurn[assetID] -= amountToBurn
+
+	// Stake any remaining value that should be staked
+	return s.consumeLockedAsset(assetID, amount-amountToBurn)
```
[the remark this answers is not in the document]
StephenButtolph: Since the logic works as expected when `amount == amountToBurn`, it seems like an unnecessary check to me

wallet/chain/p/builder/builder.go
```diff
@@ -677,6 +713,33 @@ func (b *builder) NewImportTx(
 		})
 	}
 
+	var (
+		toBurn    map[ids.ID]uint64
+		toStake   = map[ids.ID]uint64{}
+		excessFee uint64
+	)
+	if importedAVAX := importedAmounts[avaxAssetID]; importedAVAX > txFee {
+		toBurn = map[ids.ID]uint64{}
```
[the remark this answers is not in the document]
StephenButtolph: So, the fee that this transaction must pay is `txFee = b.context.StaticFeeConfig.TxFee`. Transactions pay fees by consuming that amount of AVAX in the inputs and _not_ producing that amount in the outputs (aka burning that amount of AVAX).

So, this code is looking at the amount of AVAX currently consumed, `importedAVAX`, if the builder needs to burn more AVAX (which happens if `importedAVAX < txFee`) then `toBurn[avax] = txFee - importedAVAX`, as that is the amount of additional burn that is required.

If the builder has already burned more AVAX than needed `importedAVAX > txFee` we want to make sure that the wallet returns the extra AVAX (and not pay a fee higher than required). This is handled by setting the `excessAVAX = importedAVAX - txFee`

wallet/chain/p/builder/builder.go
```diff
@@ -677,6 +713,33 @@ func (b *builder) NewImportTx(
 		})
 	}
 
+	var (
+		toBurn    map[ids.ID]uint64
+		toStake   = map[ids.ID]uint64{}
+		excessFee uint64
```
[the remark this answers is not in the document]
StephenButtolph: Changed to `excessAVAX`

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+	return utxos
+}
+
+func TestSplitUTXOsByLocktime(t *testing.T) {
+	var (
+		require = require.New(t)
+
+		expectedUnlockedUTXOsPart0 = generateUTXOs(ids.GenerateTestID(), 0)
```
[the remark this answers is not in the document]
StephenButtolph: I added a variable to hopefully make the code more clear here

utils/slice.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package utils
+
+// Join merges the provided slices into a single slice.
+//
+// TODO: Use slices.Concat once the minimum go version is 1.22.
```
ARR4N, line 8: (no action required) Their signature is interesting:

```go
func Concat[S ~[]E, E any](slices ...S) S {
```

Without looking too deeply I guess it's to play nicely with other functions in that package.

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+//nolint:gosec // This file does not need cryptographically strong randomness
```
ARR4N, line 4: Can this be limited to `G404`?

wallet/chain/p/builder/builder.go
```diff
@@ -935,15 +1023,22 @@ func (b *builder) getBalance(
...
+	ownerOverride *secp256k1fx.OutputOwners,
 	options *common.Options,
 ) (
-	inputs []*avax.TransferableInput,
-	changeOutputs []*avax.TransferableOutput,
-	stakeOutputs []*avax.TransferableOutput,
-	err error,
+	[]*avax.TransferableInput,
+	[]*avax.TransferableOutput,
+	[]*avax.TransferableOutput,
```
ARR4N, line 1040: I think it's worth still differentiating between the two outputs as part of the signature. Input and error can just be `_`.

wallet/chain/p/builder/builder.go
```diff
@@ -645,29 +698,12 @@ func (b *builder) NewImportTx(
...
-			toStake := map[ids.ID]uint64{}
-			var err error
-			inputs, outputs, _, err = b.spend(toBurn, toStake, ops)
-			if err != nil {
-				return nil, fmt.Errorf("couldn't generate tx inputs/outputs: %w", err)
-			}
+	outputs := make([]*avax.TransferableOutput, 0, len(importedAmounts))
+	for assetID, amount := range importedAmounts {
+		if assetID == avaxAssetID {
+			continue
```
ARR4N, line 704: This `continue` has the same effect as the `delete(importedAmounts, avaxAssetID)` before the loop in the old code, but that was only if `importedAVAX <= txFee` (the `else`). Is that deliberate, or should this `if` be updated?

wallet/chain/p/builder/builder.go
```diff
@@ -677,6 +713,33 @@ func (b *builder) NewImportTx(
 		})
 	}
 
+	var (
+		toBurn     map[ids.ID]uint64
```
ARR4N, line 717: This section can be simplified further.

```go
toBurn := map[ids.ID]uint64{}
toStake := map[ids.ID]uint64{}
var excessAVAX uint64

if avax := importedAmounts[avaxAssetID]; avax > txFee {
    excessAVAX = avax - txFee
} else {
    toBurn[avaxAssetID] = txFee - avax
}
```

wallet/chain/p/builder/builder.go
```diff
@@ -935,15 +1023,22 @@ func (b *builder) getBalance(
 //     place into the staked outputs. First locked UTXOs are attempted to be
 //     used for these funds, and then unlocked UTXOs will be attempted to be
 //     used. There is no preferential ordering on the unlock times.
+//   - [excessAVAX] contains the amount of extra AVAX that spend can produce in
```
ARR4N, line 1026: 👀 

(This is my way of remembering where I'm up to in a review)

ARR4N (commented):
Partial review before dinner. I've marked where I'm up to with 👀 .

> * This PR refactors `ImportTx` to always call `spend`.
> * This PR refactors all the transactions to call `spend` last.

Why?

wallet/chain/p/builder/builder_test.go
StephenButtolph: Is there a way to do this without littering `// #nosec G404` throughout the file?

wallet/chain/p/builder/builder.go
StephenButtolph: Setting the `excessAVAX` to a non-zero value is the equivalent of not deleting the entry from `importedAmounts`.

Essentially, what this is doing is:
1. Populate outputs for all the imported non-AVAX assets (with the value that was imported)
2. Remove the static fees from the imported AVAX
3. If there was more AVAX imported than needed to pay the static fee, set the excess to that amount (and spend will add the output for it)
4. If there was not enough AVAX imported to pay the static fee, set the amount to burn to that amount (and spend will consume additional inputs to pay the fee)

wallet/chain/p/builder/builder.go
StephenButtolph: Just removed the diff entirely.

wallet/chain/p/builder/builder.go
ARR4N: Sorry, your [explanation](https://github.com/ava-labs/avalanchego/pull/3282#discussion_r1711870623) made me realise that it should be `avax >= txFee`.

wallet/chain/p/builder/builder_test.go
ARR4N: [4yo open issue](https://github.com/golangci/golangci-lint/issues/1165) requesting the feature :(

wallet/chain/p/builder/builder.go
```diff
@@ -1200,3 +1289,140 @@ func (b *builder) initCtx(tx txs.UnsignedTx) error {
 	tx.InitCtx(ctx)
 	return nil
 }
+
+type spendHelper struct {
+	amountsToBurn  map[ids.ID]uint64
```
ARR4N, line 1294: Are the prefixes necessary? I think the `amounts` (implied by `uint64`) can be removed as it's too generic a word. `s.toBurn` and `s.amountsToBurn` carry equal signal for the reader.

Similarly for use of `amounts` in method variables:

```go
func (s *spendHelper) consumeLockedAsset(assetID ids.ID, amount uint64) uint64 {
	// Stake any value that should be staked
	toStake := min(
		s.toStake[assetID], // Amount we still need to stake
		amount,             // Amount available to stake
	)
	s.toStake[assetID] -= toStake
	return amount - toStake
}
```

wallet/chain/p/builder/builder.go
```diff
@@ -1068,31 +1146,61 @@ func (b *builder) spend(
 		})
 	}
 
-	// Iterate over the unlocked UTXOs
-	for _, utxo := range utxos {
+	// AVAX is handled last to account for fees.
+	avaxUTXOs, nonAVAXUTXOs := splitUTXOsByAssetID(unlockedUTXOs, b.context.AVAXAssetID)
```
ARR4N, line 1150: Is the logic here similar enough to the locked/unlocked case that it's worth abstracting the commonalities further?

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/components/verify"
+	"github.com/ava-labs/avalanchego/vms/platformvm/stakeable"
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+)
+
+func generateUTXOs(assetID ids.ID, locktime uint64) []*avax.UTXO {
+	utxos := make([]*avax.UTXO, rand.Intn(10))
+	for i := range utxos {
+		var output avax.TransferableOut = &secp256k1fx.TransferOutput{
```
ARR4N, line 24: (optional, nit)

```go
var output avax.TransferableOut
if lockTime == 0 {
    output = &secp...
} else {
    output = &stakeable...
}
```

I was a little confused as to why you were making it an interface type at the beginning and it only became apparent later that you might overwrite it. The above pattern communicates that.

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/components/verify"
+	"github.com/ava-labs/avalanchego/vms/platformvm/stakeable"
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+)
+
+func generateUTXOs(assetID ids.ID, locktime uint64) []*avax.UTXO {
+	utxos := make([]*avax.UTXO, rand.Intn(10))
```
ARR4N, line 22: Use `rand.New(rand.NewSource(seed))` to allow for deterministic testing. Failures may otherwise be flaky and very difficult to debug.

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+			expectedUnlockedUTXOsPart1,
+			expectedUnlockedUTXOsPart2,
+		)
+		expectedLockedUTXOs = utils.Join(
+			expectedLockedUTXOsPart0,
+			expectedLockedUTXOsPart1,
+		)
+	)
+
+	utxosWithAssetID, utxosWithOtherAssetID := splitUTXOsByLocktime(utxos, unlockedTime)
```
ARR4N, line 81: I promise I hadn't seen this when I suggested returning structs!

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+	require.Equal(expectedUnlockedUTXOs, utxosWithAssetID)
+	require.Equal(expectedLockedUTXOs, utxosWithOtherAssetID)
+}
+
+func TestSplitUTXOsByAssetID(t *testing.T) {
+	var (
+		require = require.New(t)
+
+		assetID                            = ids.GenerateTestID()
+		expectedUTXOsWithAssetIDPart0      = generateUTXOs(assetID, rand.Uint64())
```
ARR4N, line 91: Use `rand.New(rand.NewSource(...))` as above.

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+			expectedUnlockedUTXOsPart2,
+		)
+		expectedLockedUTXOs = utils.Join(
+			expectedLockedUTXOsPart0,
+			expectedLockedUTXOsPart1,
+		)
+	)
+
+	utxosWithAssetID, utxosWithOtherAssetID := splitUTXOsByLocktime(utxos, unlockedTime)
+	require.Equal(expectedUnlockedUTXOs, utxosWithAssetID)
```
ARR4N, line 82: This is an example of where I strongly believe that `assert` should be used over `require`. The failure of this comparison doesn't necessarily corrupt the next one, and it may very well be useful to know if/how they fail simultaneously.

Same applies to the next test.

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return utxos
+}
+
+func TestSplitUTXOsByLocktime(t *testing.T) {
+	var (
+		require = require.New(t)
+
+		unlockedTime               uint64 = 100
+		expectedUnlockedUTXOsPart0        = generateUTXOs(ids.GenerateTestID(), 0)
```
ARR4N, line 57: I think this can be made clearer if you remove the `UTXOs` and `Part\d` from all variable names. The former is implied from context and RHS and the latter is unnecessary because you don't need them split. If you do still want them separate then I think it's better to have a `[][]*avax.UTXO` than multiple variables.

```
expectUnlocked := utils.Join(
  generateUTXOs(...),
  generateUTXOs(...),
  generateUTXOs(...),
)

expectLocked := utils.Join(...)

utxos := utils.Join(expectUnlocked, expectLocked)
```

wallet/chain/p/builder/builder_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			output, locktime, err := unwrapOutput(test.output)
+			require.Equal(test.expectedOutput, output)
+			require.Equal(test.expectedLocktime, locktime)
+			require.ErrorIs(err, test.expectedErr)
```
ARR4N, line 180: If there's an error it almost certainly guarantees that the other values are invalid, so check it first to avoid red herrings in test failures.

I also think the `output` and `lockTime` checks should use `assert` because simultaneous failure could provide more information, but not a hill I'll die on.

ARR4N (approved):
Approving with comments because I don't need to re-review the implementations (unless you want me to).

wallet/chain/p/builder/builder.go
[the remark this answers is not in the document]
StephenButtolph: > What circumstances would result in more AVAX being burned than needed?
An `ImportTx` is attempting to move all the available UTXOs from shared memory onto the P-chain. If the available UTXOs contain more AVAX than the fee required to pay for the transaction, there was an excess amount burned.

Similarly, if there is only a single UTXO of, say, 5 AVAX and the fee is 1 AVAX. The wallet will spend the 5 AVAX UTXO and then return 4 AVAX in a "change" output.

wallet/chain/p/builder/builder_test.go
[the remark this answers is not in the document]
StephenButtolph: `unlockedTime`

`OutputOwners.Locktime` is when the UTXO can be spent.
`Lockout.Locktime` is when the UTXO can be transferred to a new owner.

So, if `OutputOwners.Locktime <= now < Lockout.Locktime`, then the UTXO can be spent. However, it can not be transferred to a new owner.

wallet/chain/p/builder/builder.go
StephenButtolph: I think the code actually works regardless of `avax >= txFee` or `avax > txFee` (as setting `toBurn[avaxAssetID] = 0` is fine... Although slightly inefficient I suppose.

wallet/chain/p/builder/builder.go
```diff
@@ -961,41 +1052,32 @@ func (b *builder) spend(
...
+	s := spendHelper{
+		amountsToBurn:  amountsToBurn,
+		amountsToStake: amountsToStake,
 
-		// If we have staked enough of the asset, then we have no need burn
-		// more.
-		if remainingAmountToStake == 0 {
-			continue
-		}
+		// Initialize the return values with empty slices to preserve backward
```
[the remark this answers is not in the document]
StephenButtolph: Didn't know that this was a proposal... That would be awesome.

wallet/chain/p/builder/builder_test.go
StephenButtolph: I don't think this is really do-able.

```golang
if locktime != 0 {
	output = &stakeable.LockOut{
		Locktime:        locktime,
		TransferableOut: output,
	}
}
```

references the original `output` (which is set in the body of your `if` statement). So `if` always needs to execute and the `else` only needs to execute sometimes...

Unless the thought would be to have a `secpOutput` variable and then setting `output` to either that or a wrapped version of it... That seems clunkier to me than the current code.

wallet/chain/p/builder/builder_test.go
StephenButtolph: Replaced this code with `random.Shuffle` and `require.ElementsMatch`.

wallet/chain/p/builder/builder.go
StephenButtolph: I had considered implementing a single `filter` function that took in a slice and a function... I felt like it was a bit overkill for this... But if you feel like that would be better... I could be convinced.

wallet/chain/p/builder/builder_test.go
StephenButtolph: 🦅 👁️

wallet/chain/p/builder/builder.go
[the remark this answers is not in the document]
StephenButtolph: Ah - I think this was a miscommunication.

> If the builder has already burned more AVAX than needed `importedAVAX > txFee` we want to make sure that the wallet returns the extra AVAX (and not pay a fee higher than required). This is handled by setting the `excessAVAX = importedAVAX - txFee`

Should have been
```
If the builder has already consumed more AVAX than needed importedAVAX > txFee we want to make sure that the wallet returns the extra AVAX (and not pay a fee higher than required). This is handled by setting the excessAVAX = importedAVAX - txFee
```

wallet/chain/p/builder/builder.go
ARR4N: Agreed, it should be functionally identical. Just feels "more correct" to leave `excessAVAX == 0` instead of also adding a zero burn.

wallet/chain/p/builder/builder_test.go
ARR4N: Fuzz all the things

wallet/chain/p/builder/builder.go
```diff
@@ -1200,3 +1289,140 @@ func (b *builder) initCtx(tx txs.UnsignedTx) error {
...
+		if utxo.AssetID() == assetID {
+			requestedUTXOs = append(requestedUTXOs, utxo)
+		} else {
+			otherUTXOs = append(otherUTXOs, utxo)
+		}
+	}
+	return requestedUTXOs, otherUTXOs
+}
+
+func unwrapOutput(output verify.State) (*secp256k1fx.TransferOutput, uint64, error) {
```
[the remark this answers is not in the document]
ARR4N: Given that it's only used over 5 lines it could even just be `unwrapped` but I know that I prefer succinct variables more than most people.

wallet/chain/p/builder/builder_test.go
ARR4N: Ahh I missed the reference to the original. Sorry.

PR #3283, opened 2024-08-09
add network upgrades to chain ctx

ceyonur:
## Why this should be merged

Network upgrade configs are dynamic for non-production (mainnet, fuji) + non-local chains. Chains should be aware that any changed network upgrade and configure themselves accordingly. 

Related Coreth PR: https://github.com/ava-labs/coreth/pull/630/files

## How this works

This adds the specified (or default) network upgrade configs to chain context.

## How this was tested

Locally (with coreth + subnet-evm)


proto/vm/vm.proto
```diff
@@ -116,6 +116,8 @@ message InitializeRequest {
   // the messenger, keystore, shared memory, blockchain alias,
   // subnet alias, and appSender services
   string server_addr = 14;
+  // network_upgrades_bytes is the json encoded network upgrades
+  bytes network_upgrades_bytes = 15;
```
darioush, line 120: does this need an RPC version bump?

darioush (approved):

proto/vm/vm.proto
ceyonur: yes

StephenButtolph, line 120: It seems like an anti-pattern to encode a json blob into the proto. Can we create an upgrade message that specifies the actual upgrade times?

ceyonur: added

vms/rpcchainvm/vm_server.go
```diff
@@ -843,3 +851,76 @@ func (vm *VMServer) StateSummaryAccept(
 		Err:  errorToErrEnum[err],
 	}, errorToRPCError(err)
 }
+
+func convertNetworkUpgrades(pbUpgrades *vmpb.NetworkUpgrades) (upgrade.Config, error) {
```
StephenButtolph, line 855: `pbUpgrades` may be `nil` here - Can we error rather than panic in that case?

StephenButtolph (commented):
1 nit then lgtm

vms/rpcchainvm/vm_server.go
ceyonur: added 👍

StephenButtolph (approved):

PR #3286, opened 2024-08-10
Add P-chain fee APIs

StephenButtolph:
## Why this should be merged

This PR:
1. Exposes the dynamic fee configuration after Etna is activated.
2. Exposes the dynamic fee state of the chain.

## How this works

Adds APIs to the service + client.

## How this was tested

- [X] Manual testing. This will be e2e tested when it is incorporated into the wallet.

### DIFF
--- a/vms/platformvm/service.go
+++ b/vms/platformvm/service.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/ava-labs/avalanchego/vms/components/fee"
 	"github.com/ava-labs/avalanchego/vms/components/keystore"
 	"github.com/ava-labs/avalanchego/vms/platformvm/fx"
 	"github.com/ava-labs/avalanchego/vms/platformvm/reward"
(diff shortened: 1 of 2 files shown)


vms/platformvm/service.go
dhrubabasu, line 1: Should we add tests for these new functions in `service_test.go`?

StephenButtolph: tbh - not sure how useful the tests are going to be... But I added them

vms/platformvm/service.go
```diff
@@ -1828,6 +1829,48 @@ func (s *Service) GetBlockByHeight(_ *http.Request, args *api.GetBlockByHeightAr
...
+	s.vm.ctx.Log.Debug("API called",
+		zap.String("service", "platform"),
+		zap.String("method", "getFeeState"),
+	)
+
+	s.vm.ctx.Lock.Lock()
+	defer s.vm.ctx.Lock.Unlock()
+
+	reply.State = s.vm.state.GetFeeState()
+	reply.Price = s.vm.DynamicFeeConfig.MinGasPrice.MulExp(
```
ARR4N, line 1866: Is there shared code that performs this calculation? I'm guessing the price is used somewhere internally, and this is at risk of diverging.

vms/platformvm/service_test.go
```diff
@@ -1107,3 +1108,65 @@ func TestServiceGetSubnets(t *testing.T) {
...
+		})
+	}
+}
+
+func TestGetFeeState(t *testing.T) {
+	require := require.New(t)
+
+	seed := time.Now().UnixNano()
+	t.Logf("Seed: %d", seed)
+	random := rand.New(rand.NewSource(seed)) // #nosec G404
```
ARR4N, line 1148: This suggests that this test should be `func FuzzGetFeeState(f *testing.F)` with the internal test func remaining the same as below.

ARR4N (approved):
Changes don't require my re-review.

vms/platformvm/service.go
```diff
@@ -1828,6 +1829,48 @@ func (s *Service) GetBlockByHeight(_ *http.Request, args *api.GetBlockByHeightAr
...
+	if !s.vm.Config.UpgradeConfig.IsEtnaActivated(now) {
+		return nil
+	}
+
+	*reply = s.vm.DynamicFeeConfig
+	return nil
+}
+
+type GetFeeStateReply struct {
+	fee.State
```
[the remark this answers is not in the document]
StephenButtolph: yes

vms/platformvm/service.go
```diff
@@ -1828,6 +1829,48 @@ func (s *Service) GetBlockByHeight(_ *http.Request, args *api.GetBlockByHeightAr
 	return err
 }
 
+// GetFeeConfig returns the dynamic fee config of the chain.
+func (s *Service) GetFeeConfig(_ *http.Request, _ *struct{}, reply *fee.Config) error {
```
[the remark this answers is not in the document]
StephenButtolph: `Service` is provided to `"github.com/gorilla/rpc/v2".Server.RegisterService` which expects the functions to match this definition.

vms/platformvm/service.go
StephenButtolph: I'm not sure how better to share this... This function takes in 3 params to calculate the price... Any helper would also take in 3 params and just call this function... No?

vms/platformvm/service_test.go
StephenButtolph: ok

PR #3287, opened 2024-08-11
Implement SoV data struct

dhrubabasu:

### DIFF
--- a/vms/components/fee/validator.go
+++ b/vms/components/fee/validator.go
@@ -1,0 +1,90 @@
+package fee
+
+import (
+	"fmt"
+
+	"github.com/holiman/uint256"
+)
+
+type ValidatorState struct {
+	Current  Gas
+	Target   Gas
+	Capacity Gas
+	Excess   Gas
+	MinFee   GasPrice
+	K        Gas
+}
+
+func (v ValidatorState) CurrentFeeRate() GasPrice {
+	return v.MinFee.MulExp(v.Excess, v.K)
+}
+
+func (v ValidatorState) CalculateContinuousFee(seconds uint64) uint64 {
+	if v.Current == v.Target {
+		return uint64(v.MinFee.MulExp(v.Excess, v.K)) * seconds
+	}
+
+	uint256.NewInt(uint64(v.Excess))
+
+	var totalFee uint64
+	if v.Current < v.Target {
+		secondsTillExcessIsZero := uint64(v.Excess / (v.Target - v.Current))
+
+		if secondsTillExcessIsZero < seconds {
+			totalFee += uint64(v.MinFee) * (seconds - secondsTillExcessIsZero)
+			seconds = secondsTillExcessIsZero
+		}
+	}
+
+	x := v.Excess
+	for i := uint64(1); i <= seconds; i++ {
+		if v.Current < v.Target {
+			x = x.SubPerSecond(v.Target-v.Current, 1)
+		} else {
+			x = x.AddPerSecond(v.Current-v.Target, 1)
+		}
+
+		if x == 0 {
+			totalFee += uint64(v.MinFee)
+			continue
+		}
+
+		totalFee += uint64(v.MinFee.MulExp(x, v.K))
+	}
+
+	return totalFee
+}
+
+// Returns the first number n where CalculateContinuousFee(n) >= balance
+func (v ValidatorState) CalculateTimeTillContinuousFee(balance uint64) (uint64, uint64) {
+	// Lower bound can be derived from [MinFee].
+	n := balance / uint64(v.MinFee)
+	interval := n
+
+	numIters := 0
+	for {
+		fmt.Printf("n=%d", n)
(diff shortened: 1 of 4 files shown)


vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+			n -= interval
+		}
+
+		if feeAtN < balance {
+			n += interval
+		}
+
+		numIters++
+	}
+}
```
StephenButtolph, line 93: A couple comments:
1. We shouldn't re-implement binary search. [`sort.Search`](https://pkg.go.dev/sort#Search) should be used for something like this.
2. `CalculateContinuousFee` iterates from the current time to the specified time... So this implementation ends up being `O(n log n)` in the number of seconds, when just performing a linear scan would be `O(n)`.

vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+func (v ValidatorState) CurrentFeeRate() GasPrice {
+	return v.MinFee.MulExp(v.Excess, v.K)
+}
+
+func (v ValidatorState) CalculateContinuousFee(seconds uint64) uint64 {
+	if v.Current == v.Target {
+		return uint64(v.MinFee.MulExp(v.Excess, v.K)) * seconds
+	}
+
+	uint256.NewInt(uint64(v.Excess))
```
StephenButtolph, line 30: dead code?

vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+
+	var totalFee uint64
+	if v.Current < v.Target {
+		secondsTillExcessIsZero := uint64(v.Excess / (v.Target - v.Current))
```
StephenButtolph, line 34: If `v.Excess = 5` and `v.Target - v.Current = 2`, then `secondsTillExcessIsZero = 2`. But that isn't correct, as in `2` seconds `v.Excess` will equal `1`.

vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+		if v.Current < v.Target {
+			x = x.SubPerSecond(v.Target-v.Current, 1)
+		} else {
+			x = x.AddPerSecond(v.Current-v.Target, 1)
+		}
+
+		if x == 0 {
+			totalFee += uint64(v.MinFee)
+			continue
+		}
```
StephenButtolph, line 53: This is not needed.

vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/holiman/uint256"
+)
+
+type ValidatorState struct {
+	Current  Gas
+	Target   Gas
+	Capacity Gas
+	Excess   Gas
+	MinFee   GasPrice
+	K        Gas
```
StephenButtolph, line 18: Can we call this `ExcessConversionConstant` like in the other places? `K` isn't very descriptive.

vms/components/fee/validator.go
```diff
@@ -0,0 +1,93 @@
...
+		secondsTillExcessIsZero := uint64(v.Excess / (v.Target - v.Current))
+
+		if secondsTillExcessIsZero < seconds {
+			totalFee += uint64(v.MinFee) * (seconds - secondsTillExcessIsZero)
+			seconds = secondsTillExcessIsZero
+		}
+	}
+
+	x := v.Excess
+	for i := uint64(1); i <= seconds; i++ {
```
StephenButtolph, line 43: This is more standard (because `i` can never overflow)
```suggestion
	for i := uint64(0); i < seconds; i++ {
```

vms/platformvm/state/subnet_only_validators.go
```diff
@@ -0,0 +1,383 @@
...
+
+var (
+	_ SubnetOnlyValidators = (*subnetOnlyValidators)(nil)
+
+	ErrAlreadyValidator  = errors.New("already a validator")
+	ErrValidatorNotFound = errors.New("validator not found")
+	ErrNonZeroWeight     = errors.New("weight must be 0 when minNonce is MaxUint64")
+)
+
+type SubnetOnlyValidators interface {
```
StephenButtolph, line 30: 1. Where are we exposing the "Next staker evicted time"?
2. Have we started looking into how the state diff will be implemented here?

vms/components/fee/validator.go
dhrubabasu: `v.Excess` corresponds to the excess _after_ a block timestamp has elapsed. If `v.Excess` is currently 5, we have already charged the fee for when it is 5.

That means the excess we have not yet charged will be: `[3, 1, 0, 0, ...]` if `v.Target - v.Current = 2`. This then gives us `secondsTillExcessIsZero = 2` and we can optimize the calculation by simply adding `MinFee * (seconds - 2)`.

vms/components/fee/validator.go
dhrubabasu: converted to a linear scan, I don't think there's a way to make it more efficient than `O(n)`

PR #3288, opened 2024-08-12
Add scaffolding for warp verifier

dhrubabasu:
## Why this should be merged

Preparatory work for when we verify warp messages as specified in ACP-77. Currently, it does nothing since no txs with warp messages have been introduced into the codebase.

## How this works

pretty straightforward

## How this was tested

existing UTs. Will add UTs for the verifier itself once we add txs.

### DIFF
--- a/vms/platformvm/block/executor/block.go
+++ b/vms/platformvm/block/executor/block.go
@@ -38,7 +38,7 @@ func (b *Block) VerifyWithContext(_ context.Context, ctx *smblock.Context) error
 	blkID := b.ID()
 	if blkState, ok := b.manager.blkIDToState[blkID]; ok {
 		if !blkState.verifiedHeights.Contains(pChainHeight) {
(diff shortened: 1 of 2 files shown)


dhrubabasu:
> Why would it be desirable to merge this in advance of transactions that would exercise it and allow for testing?

Makes it less error-prone when we do add transactions since this will have to be updated to comply with txs.Visitor

not necessarily required now but in the interest of reducing diffs

dhrubabasu:
> > > Why would it be desirable to merge this in advance of transactions that would exercise it and allow for testing?
> > 
> > 
> > Makes it less error-prone when we do add transactions since this will have to be updated to comply with txs.Visitor
> > not necessarily required now but mostly in the interest of reducing diffs down the line
> 
> Not my call, but in general I think adding code that is actually exercised is preferable to minimizing diffs.

That's fair, we can merge it with the first warp message verification via `RegisterSubnetValidatorTx`

PR #3293, opened 2024-08-14
Remove local chain config from e2e test

ceyonur:
## Why this should be merged

Coreth now sets network upgrades with snow.Context. A local chain config should not be used with local network ID.

## How this works

Sets only chainID for the config

## How this was tested

e2e test should cover this


tests/upgrade/upgrade_test.go
```diff
@@ -61,12 +63,20 @@ var _ = ginkgo.Describe("[Upgrade]", func() {
 			cChainGenesisStr := network.Genesis.CChainGenesis
 			require.NoError(json.Unmarshal([]byte(cChainGenesisStr), cChainGenesis))
 			unscheduledActivationTime := uint64(upgrade.UnscheduledActivationTime.Unix())
-			cChainGenesis.Config.EtnaTime = &unscheduledActivationTime
+			cChainGenesis.Config.EtnaTimestamp = &unscheduledActivationTime
 			cChainGenesisBytes, err := json.Marshal(cChainGenesis)
 			require.NoError(err)
 			network.Genesis.CChainGenesis = string(cChainGenesisBytes)
```
darioush, line 69: when I was looking at the issue with the upgrade test, seems these lines 
https://github.com/ava-labs/coreth/pull/635/files#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912R469-R471
overwrite it with the chainCtx, where is the chainCtx being set for this test? Should we apply the desired behavior there?

tests/upgrade/upgrade_test.go
```diff
@@ -61,12 +63,20 @@ var _ = ginkgo.Describe("[Upgrade]", func() {
...
 		}
 
+		// Previous version does not have unactivated upgrades.
+		// The default upgrade config usually sets the latest upgrade to be activated
+		// Use unscheduled
+		upgradeJSON, err := json.Marshal(upgrade.LatestUnscheduled)
+		require.NoError(err)
+		upgradeBase64 := base64.StdEncoding.EncodeToString(upgradeJSON)
+		network.DefaultFlags[config.UpgradeFileContentKey] = upgradeBase64
```
darioush, line 78: I'm not sure why the upgrade json is needed (it is not parsed in coreth, afaik)

tests/upgrade/upgrade_test.go
ceyonur: this is set by the avalanchego flag `upgrade-file-content-key`. It's applied with a flag and then the node passes it to Coreth. We don't actually create chainCtx here.

tests/upgrade/upgrade_test.go
ceyonur: This is the new network upgrades flag (not the `upgrade.json` under chain-config-dir`). I agree this is slightly confusing.

tests/fixture/tmpnet/genesis.go
```diff
@@ -147,11 +147,12 @@ func NewTestGenesis(
 		)
 	}
 
+	chainID := big.NewInt(int64(networkID))
 	// Define C-Chain genesis
 	cChainGenesis := &core.Genesis{
-		Config:     params.AvalancheLocalChainConfig,
-		Difficulty: big.NewInt(0),                              // Difficulty is a mandatory field
-		Timestamp:  uint64(upgrade.InitiallyActiveTime.Unix()), // This time enables Avalanche upgrades by default
+		Config:     params.GetChainConfig(upgrade.Default, chainID), // upgrade will be again set by VM according to the snow.Context
```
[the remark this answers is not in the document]
ceyonur: yes, the older version of Coreth won't be able to set this from ctx, so it won't enable upgrades at genesis.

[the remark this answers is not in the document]
darioush: could we document this and add a TODO to remove it after Etna?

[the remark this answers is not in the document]
ceyonur: done

darioush (approved):

tests/upgrade/upgrade_test.go
```diff
@@ -54,19 +55,18 @@ var _ = ginkgo.Describe("[Upgrade]", func() {
...
 		}
 
+		// Previous version does not have unactivated upgrades.
+		// The default upgrade config usually sets the latest upgrade to be activated
+		// Use unscheduled
+
+		upgradeJSON, err := json.Marshal(upgrade.LatestUnscheduled)
+		require.NoError(err)
+		upgradeBase64 := base64.StdEncoding.EncodeToString(upgradeJSON)
+		network.DefaultFlags = tmpnet.FlagsMap{}
```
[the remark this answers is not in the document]
ceyonur: it was supposed to be a review comment, forgot to remove it.

scripts/tests.upgrade.sh
```diff
@@ -17,7 +17,8 @@ fi
 # version with the latest code.
 #
 # v1.11.0 activates Durango.
```
StephenButtolph, line 19:
```suggestion
```

upgrade/upgrade.go
```diff
@@ -73,6 +73,26 @@ var (
...
+		ApricotPhase5Time:            InitiallyActiveTime,
+		ApricotPhasePre6Time:         InitiallyActiveTime,
+		ApricotPhase6Time:            InitiallyActiveTime,
+		ApricotPhasePost6Time:        InitiallyActiveTime,
+		BanffTime:                    InitiallyActiveTime,
+		CortinaTime:                  InitiallyActiveTime,
+		CortinaXChainStopVertexID:    ids.Empty,
+		DurangoTime:                  InitiallyActiveTime,
+		EtnaTime:                     UnscheduledActivationTime,
+	}
```
StephenButtolph, line 94: Since this is only used in tests - I feel like using `upgradetest` would be preferred rather than adding this here. Also we could then use `GetConfig(Latest - 1)` rather than redefining this whole struct again.

ceyonur: hmm I think that code was not there before. and I thought this can be used in elsewhere and can provide a reliable config after every new upgrade definition.

ceyonur: changed it now with `GetConfig` from `upgradetest`

tests/upgrade/upgrade_test.go
```diff
@@ -54,19 +55,17 @@ var _ = ginkgo.Describe("[Upgrade]", func() {
...
-			cChainGenesisStr := network.Genesis.CChainGenesis
-			require.NoError(json.Unmarshal([]byte(cChainGenesisStr), cChainGenesis))
-			unscheduledActivationTime := uint64(upgrade.UnscheduledActivationTime.Unix())
-			cChainGenesis.Config.EtnaTime = &unscheduledActivationTime
-			cChainGenesisBytes, err := json.Marshal(cChainGenesis)
-			require.NoError(err)
-			network.Genesis.CChainGenesis = string(cChainGenesisBytes)
 		}
 
+		latestUnscheduled := upgradetest.GetConfig(upgradetest.Latest - 1)
```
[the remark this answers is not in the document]
ceyonur: removed the block and added comment

StephenButtolph (approved):

PR #3294, opened 2024-08-14
Add support for VM HTTP2 handlers

joshua-kim:
## Why this should be merged

Adds native support for GRPC services being registered over the avalanchego http server. This adds support for streaming via grpc as an alternative to websockets. 

## How this works

1. Uses `h2c`'s http handler to support plaintext http2
2. Makes a change to rpcchainvm to fix a bug where headers that are updated in the plugin process are not reflected on avalanchego's side by returning a response at the end of the `Handle` rpc that includes headers
3. Introduces a new rpc for the http request `request.Body` so that the plugin process can stream the body instead of the current behavior which uses `io.ReadAll` which may not be possible in the case of a streaming rpc.

## How this was tested

Added a e2e test for unary + streaming rpcs over rpcchainvm

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/StephenButtolph/canoto v0.15.0
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/coreth v0.15.1-rc.0.0.20250606020240-be987ae92bf9
+	github.com/ava-labs/coreth v0.15.2-rc.0
 	github.com/ava-labs/ledger-avalanche/go v0.0.0-20241009183145-e6f90a8a1a60
 	github.com/ava-labs/libevm v1.13.14-0.2.0.release
 	github.com/btcsuite/btcd/btcutil v1.1.3
--- a/go.sum
+++ b/go.sum
@@ -68,8 +68,8 @@ github.com/antithesishq/antithesis-sdk-go v0.3.8/go.mod h1:IUpT2DPAKh6i/YhSbt6Gl
 github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6/go.mod h1:grANhF5doyWs3UAsr3K4I6qtAmlQcZDesFNEHPZAzj8=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 h1:0CwZNZbxp69SHPdPJAN/hZIm0C4OItdklCFmMRWYpio=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5/go.mod h1:wHh0iHkYZB8zMSxRWpUBQtwG5a7fFgvEO+odwuTv2gs=
-github.com/ava-labs/coreth v0.15.1-rc.0.0.20250606020240-be987ae92bf9 h1:RHgQBDUifaEJ8xplpyEZS41lR5U2nMGC0Fxn+l4dQ5w=
-github.com/ava-labs/coreth v0.15.1-rc.0.0.20250606020240-be987ae92bf9/go.mod h1:Tl6TfpOwSq0bXvN7DTENXB4As3hml8ma9OQEWS9DH9I=
+github.com/ava-labs/coreth v0.15.2-rc.0 h1:RnCH59A/WMEdMyuqugaPIk1xrQQuo9vEgMVOjk9ZjWQ=
+github.com/ava-labs/coreth v0.15.2-rc.0/go.mod h1:Tl6TfpOwSq0bXvN7DTENXB4As3hml8ma9OQEWS9DH9I=
 github.com/ava-labs/ledger-avalanche/go v0.0.0-20241009183145-e6f90a8a1a60 h1:EL66gtXOAwR/4KYBjOV03LTWgkEXvLePribLlJNu4g0=
 github.com/ava-labs/ledger-avalanche/go v0.0.0-20241009183145-e6f90a8a1a60/go.mod h1:/7qKobTfbzBu7eSTVaXMTr56yTYk4j2Px6/8G+idxHo=
 github.com/ava-labs/libevm v1.13.14-0.2.0.release h1:uKGCc5/ceeBbfAPRVtBUxbQt50WzB2pEDb8Uy93ePgQ=


proto/http/http.proto
```diff
@@ -149,6 +147,12 @@ message HTTPRequest {
   Request request = 2;
 }
 
+message HttpResponse {
+  // header is sent to synchronize any header modifications from the plugin
+  // process
+  repeated Element header = 1;
```
joshua-kim, line 153: grpc requires a header to indicate the end of an rpc, which is directly set onto the map in the plugin process ([ref](https://github.com/grpc/grpc-go/blob/cd05c9e58fb9d6563e1c7d895f55faec763fc20e/internal/transport/handler_server.go#L244)).

proto/platform/platform.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+package platform;
+
+option go_package = "github.com/ava-labs/avalanchego/proto/pb/platform";
+
+// Platform defines the P-Chain APIs
+service Platform {
```
joshua-kim, line 8: This doesn't need to be included as part of this PR (as we already have tests through xsvm)... I can unstage this if we want.

vms/rpcchainvm/ghttp/http_client.go
```diff
@@ -160,7 +159,20 @@ func (c *Client) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	// Copy headers from the plugin process
+	header := w.Header()
+	clear(header)
+
+	for _, h := range reply.Header {
+		header[h.Key] = h.Values
+	}
```
joshua-kim, line 174: I need to think about if this is safe... upon closer inspection all of the header syncing in rpcchainvm feels weird.

proto/platform/platform.proto
joshua-kim: Removed

api/server/grpc_router.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+type grpcRouter struct {
+	lock     sync.Mutex
+	handlers map[string]http.Handler
+}
+
+func (g *grpcRouter) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	g.lock.Lock()
```
yacovm, line 32: maybe use a RWMutex instead of a regular Mutex?

joshua-kim:
> Would it be possible to elaborate as to the motivation for this addition? The current 'why this should be merged' in the description seems more like 'what'.

Added

joshua-kim:
> The upgrade failure is legit - I'm guessing you might need to rebase?
I think my coreth branch fell behind master. Gonna rebase my coreth branch again.

vms/rpcchainvm/vm_client.go
```diff
@@ -392,6 +392,25 @@ func (vm *VMClient) CreateHandlers(ctx context.Context) (map[string]http.Handler
 	return handlers, nil
 }
 
+func (vm *VMClient) CreateGRPCService(ctx context.Context) (string, http.Handler, error) {
+	resp, err := vm.client.CreateGRPCService(ctx, &emptypb.Empty{})
+	if err != nil {
+		return "", nil, err
+	}
+
+	if resp.ServiceName == "" && resp.ServerAddr == "" {
```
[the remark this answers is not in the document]
joshua-kim: We're doing this instead of comparing against the zero-value of the response because proto responses auto-embed `protoimpl.MessageState` which cannot be compared.

If one of these was empty and the other wasn't, it would be because the vm explicitly tried to register a grpc service with an empty service name which I think should be invalid (since grpc calls take the form of service + method).

vms/rpcchainvm/vm_server.go
```diff
@@ -363,6 +363,30 @@ func (vm *VMServer) CreateHandlers(ctx context.Context, _ *emptypb.Empty) (*vmpb
 	return resp, nil
 }
 
+func (vm *VMServer) CreateGRPCService(ctx context.Context, _ *emptypb.Empty) (*vmpb.CreateGRPCHandlerResponse, error) {
+	serviceName, handler, err := vm.vm.CreateGRPCService(ctx)
```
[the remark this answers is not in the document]
joshua-kim: That looks like a bug. I think it's not caught in CI because the only vm that uses rpcchainvm in our testing as part of this PR does register a grpc service.

vms/rpcchainvm/vm_client.go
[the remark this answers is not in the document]
joshua-kim: Using or now

api/server/server.go
```diff
@@ -303,3 +325,19 @@ func (a readPathAdder) AddRoute(handler http.Handler, base, endpoint string) err
...
+	allowedHostsHandler := filterInvalidHosts(handler, allowedHosts)
+	corsHandler := cors.New(cors.Options{
+		AllowedOrigins:   allowedOrigins,
+		AllowCredentials: true,
+	}).Handler(allowedHostsHandler)
+	gzipHandler := gziphandler.GzipHandler(corsHandler)
+	return http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			// Attach this node's ID as a header
+			w.Header().Set("node-id", nodeID.String())
```
joshua-kim, line 339: It's not clear to me why we attach this header

api/server/server.go
```diff
@@ -199,6 +198,26 @@ func (s *server) RegisterChain(chainName string, ctx *snow.ConsensusContext, vm
...
+		)
+		return
+	}
+
+	if serviceName == "" && grpcHandler == nil {
+		return
+	}
+
+	grpcHandler = s.wrapMiddleware(chainName, grpcHandler, ctx)
+	if !s.grpcRouter.Add(ctx.ChainID.String(), serviceName, grpcHandler) {
```
joshua-kim: Should we use chain alias here and fall back to chain id if it's not available?

tests/e2e/vms/xsvm.go
```diff
@@ -161,6 +164,65 @@ var _ = ginkgo.Describe("[XSVM]", func() {
...
+				require.NoError(err)
+
+				tc.Outf("pong: %s\n", reply.Message)
+				require.Equal(fmt.Sprintf("ping-%d", i), reply.Message)
+			}
+		}()
+
+		wg.Wait()
+
+		require.NoError(stream.CloseSend())
```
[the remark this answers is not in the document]
joshua-kim: It needs to be explicitly called. Updated to use `DeferCleanup`

tests/e2e/vms/xsvm.go
```diff
@@ -178,6 +184,86 @@ var _ = ginkgo.Describe("[XSVM]", func() {
 
 		_ = e2e.CheckBootstrapIsPossible(tc, network)
 	})
+
+	ginkgo.It("should serve grpc api requests", func() {
+		if e2e.GetEnv(tc).GetNetwork().DefaultRuntimeConfig.Kube != nil {
+			ginkgo.Skip("h2c is not currently supported in kube")
```
joshua-kim, line 191: Follow-up https://github.com/ava-labs/avalanchego/pull/3294/files

vms/rpcchainvm/ghttp/gresponsewriter/writer_server.go
```diff
@@ -69,7 +69,6 @@ func (s *Server) WriteHeader(
 	for _, header := range req.Headers {
 		headers[header.Key] = header.Values
 	}
-	s.writer.WriteHeader(grpcutils.EnsureValidResponseCode(int(req.StatusCode)))
```
joshua-kim: Remove this from diff

snow/engine/common/vm.go
```diff
@@ -77,4 +77,8 @@ type VM interface {
 	// it have an extension called `accounts`, where clients could get
 	// information about their accounts.
 	CreateHandlers(context.Context) (map[string]http.Handler, error)
+
+	// CreateGRPCService returns the gRPC service to register into the
+	// avalanchego api server.
+	CreateGRPCService(ctx context.Context) (string, http.Handler, error)
```
joshua-kim: Also wondering if we should support a map like `CreateHandlers` if we want more than one grpc service on the VM.

tests/e2e/vms/xsvm.go
```diff
@@ -178,6 +184,86 @@ var _ = ginkgo.Describe("[XSVM]", func() {
 
 		_ = e2e.CheckBootstrapIsPossible(tc, network)
 	})
+
+	ginkgo.It("should serve grpc api requests", func() {
+		if e2e.GetEnv(tc).GetNetwork().DefaultRuntimeConfig.Kube != nil {
```
maru-ava: GetNetwork is loading the configuration from disk every time. Prefer loading the network once in the context of a given test rather than at every point of call.

tests/e2e/vms/xsvm.go
```diff
@@ -178,6 +184,86 @@ var _ = ginkgo.Describe("[XSVM]", func() {
...
+				ginkgo.GinkgoRecover()
+				wg.Done()
+			}()
+
+			for i := 0; i < n; i++ {
+				msg := fmt.Sprintf("ping-%d", i)
+				require.NoError(stream.Send(&xsvm.StreamPingRequest{
+					Message: msg,
+				}))
+				tc.Log().Info("sent message", zap.String("msg", msg))
```
maru-ava: (No action required) Not strictly required, but it may make sense to get a reference to the log globally for the test rather than on each usage.

tests/e2e/vms/xsvm.go
joshua-kim: Updated to load once

tests/e2e/vms/xsvm.go
joshua-kim: Updated xsvm tests to just hold a log reference

tests/e2e/vms/xsvm.go
```diff
@@ -178,6 +185,86 @@ var _ = ginkgo.Describe("[XSVM]", func() {
...
+				grpcclient.PrefixChainIDUnaryClientInterceptor(chainID),
+			),
+			grpc.WithStreamInterceptor(
+				grpcclient.PrefixChainIDStreamClientInterceptor(chainID),
+			),
+		)
+		require.NoError(err)
+
+		tc.By("serving unary rpc")
+		client := xsvm.NewPingClient(conn)
```
StephenButtolph: Should this be a util somewhere? I'd imagine every time we create a new connection passing in:

1. insecure creds
2. unary interceptor
3. stream interceptor

Would be a bit annoying.

Also, since the connection is scoped to the `xsvm` (because of the chainID) it feels like we could also wrap connection in the client seamlessly as well.

tests/e2e/vms/xsvm.go
```diff
@@ -178,6 +185,86 @@ var _ = ginkgo.Describe("[XSVM]", func() {
...
+		}
+
+		tc.By("establishing connection")
+		nodeID := network.GetSubnet(subnetAName).ValidatorIDs[0]
+		node, err := network.GetNode(nodeID)
+		require.NoError(err)
+
+		uri := strings.TrimPrefix(node.URI, "http://")
+		chainID := network.GetSubnet(subnetAName).Chains[0].ChainID
+		conn, err := grpc.NewClient(
```
StephenButtolph: Should we be closing this connection?

vms/example/xsvm/grpc_service.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/ava-labs/avalanchego/proto/pb/xsvm"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+var _ xsvm.PingServer = (*grpcService)(nil)
+
+type grpcService struct {
+	Log logging.Logger
+	xsvm.UnimplementedPingServer
```
StephenButtolph: We typically use `Unsafe...Server` rather than `Unimplemented...Server` so that it enforces that we update our golang code when the proto file changes.
```suggestion
	xsvm.UnsafePingServer

	Log logging.Logger
```

snow/engine/common/vm.go
StephenButtolph: As far as I can see, we do not (need to) use the service name in the api server.
```suggestion
	// CreateGRPCHandler returns the gRPC server to register into the
	// avalanchego api server.
	CreateGRPCHandler(ctx context.Context) (http.Handler, error)
```
(If this can be made generic to `HTTP2` that would be sick and we could rename it `CreateHTTP2Handler`)

api/server/grpc_router.go
```diff
@@ -0,0 +1,72 @@
...
+		return false
+	}
+
+	g.handlers[prefixedService] = handler
+	return true
+}
+
+func prefixChainGRPCService(chainID string, service string) string {
+	return chainID + "/" + service
+}
```
StephenButtolph: Why do we include the `service` here? Isn't the `chainID` sufficient?

api/server/server.go
```diff
@@ -115,33 +118,28 @@ func New(
...
+		Handler: h2c.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.ProtoMajor == 2 && strings.Contains(r.Header.Get("Content-Type"), "application/grpc") {
+				grpcHandler.ServeHTTP(w, r)
+				return
+			}
+
+			handler.ServeHTTP(w, r)
+		}), &http2.Server{
+			MaxConcurrentStreams: maxConcurrentStreams,
+		}),
```
StephenButtolph: Optional: I just find the previous indentation hard to read.
```suggestion
		Handler: h2c.NewHandler(
			http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
				if r.ProtoMajor == 2 && strings.Contains(r.Header.Get("Content-Type"), "application/grpc") {
					grpcHandler.ServeHTTP(w, r)
					return
				}

				handler.ServeHTTP(w, r)
			}),
			&http2.Server{
				MaxConcurrentStreams: maxConcurrentStreams,
			},
		),
```

Alternatively, we could make the handler _outside_ the `http.Server{` definition (like it was before).

api/server/server.go
```diff
@@ -115,33 +118,28 @@ func New(
...
+	handler := wrapHandler(router, nodeID, allowedOrigins, allowedHosts, true)
+
+	grpcRouter := newGRPCRouter()
+	// Do not use gzip middleware because it breaks the grpc spec
+	grpcHandler := wrapHandler(grpcRouter, nodeID, allowedOrigins, allowedHosts, false)
 
 	httpServer := &http.Server{
-		Handler:           handler,
+		Handler: h2c.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.ProtoMajor == 2 && strings.Contains(r.Header.Get("Content-Type"), "application/grpc") {
```
StephenButtolph: Does this need to be grpc specific?

proto/xsvm/service.proto
```diff
@@ -0,0 +1,26 @@
...
+  string message = 1;
+}
+
+message StreamPingRequest {
+  string message = 1;
+}
+
+message StreamPingReply {
+  string message = 1;
+}
```
StephenButtolph, line 26: If it was done this way to conform to gRPC standards... Feel free to keep the code as is... But I feel like this is reasonable.
```suggestion
service Ping {
  rpc Ping(PingRequest) returns (PingReply);
  rpc StreamPing(stream PingRequest) returns (stream PingReply);
}

message PingRequest {
  string message = 1;
}

message PingReply {
  string message = 1;
}
```

vms/example/xsvm/grpc_service.go
```diff
@@ -0,0 +1,50 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to receive message: %w", err)
+		}
+
+		g.Log.Debug("stream ping", zap.String("message", request.Message))
+		if err := server.Send(&xsvm.StreamPingReply{
+			Message: request.Message,
+		}); err != nil {
+			return fmt.Errorf("failed to send message: %w", err)
+		}
```
StephenButtolph: Optional nit again, I dislike the multi-line `if` statements.
```suggestion
		err = server.Send(&xsvm.StreamPingReply{
			Message: request.Message,
		})
		if err != nil {
			return fmt.Errorf("failed to send message: %w", err)
		}
```

proto/xsvm/service.proto
joshua-kim: I usually create separate types for each api so they can evolve independently but in retrospect it was overkill for a dummy testing api

tests/e2e/vms/xsvm.go
joshua-kim: Yeah for some reason I closed the stream but not the connection...

proto/xsvm/service.proto
joshua-kim: Ah, this actually fails the profobuf linter. Since it's just a test api, we could make an exception and ignore the linter here. Are you fine with us just keeping it as-is to avoid the linter override?
```
Running protobuf lint check...
xsvm/service.proto:8:3:"xsvm.PingReply" is used as the request or response type for multiple RPCs.
xsvm/service.proto:8:3:"xsvm.PingRequest" is used as the request or response type for multiple RPCs.
xsvm/service.proto:9:3:"xsvm.PingReply" is used as the request or response type for multiple RPCs.
xsvm/service.proto:9:3:"xsvm.PingRequest" is used as the request or response type for multiple RPCs.
ERROR: protobuf linter failed
```

api/server/grpc_router.go
joshua-kim: Removed

api/server/server.go
joshua-kim: This suggestion shows up as outdated so I had to manually apply the diff.

api/server/server.go
joshua-kim: Updated to just check for http2

snow/engine/common/vm.go
joshua-kim: Updated + same w/ all other grpc specific comments.

tests/e2e/vms/xsvm.go
joshua-kim: Added some new helpers for this in xsvm + grpcclient.

proto/xsvm/service.proto
StephenButtolph: Yeah we can just follow the linter here

vms/example/xsvm/api/ping.go
```diff
@@ -0,0 +1,69 @@
...
+
+// NewPingClient returns a client for PingService
+func NewPingClient(
+	uri string,
+	chainID ids.ID,
+	opts ...grpc.DialOption,
+) (xsvm.PingClient, *grpc.ClientConn, error) {
+	conn, err := grpcclient.NewChainClient(uri, chainID, opts...)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to initialize grpc client: %w", err)
```
StephenButtolph: I think the returned error here is going to look like:

`failed to initialize grpc client: failed to initialize grpc client: SOME_ERROR`

Should we remove one of the wrappings?

api/server/http2_router.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func (h *http2Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+
+	// The :path pseudo-header takes the form of /Prefix/Path
+	parsed := strings.Split(r.URL.Path, "/")
+	if len(parsed) < 3 {
```
StephenButtolph: We only actually care about `[1]`, so why do we enforce that this has length >=3?
```suggestion
	if len(parsed) < 2 {
```

api/server/http2_router.go
```diff
@@ -0,0 +1,66 @@
...
+func (h *http2Router) Add(chainID ids.ID, handler http.Handler) bool {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	if _, ok := h.handlers[chainID.String()]; ok {
+		return false
+	}
+
+	h.handlers[chainID.String()] = handler
+	return true
```
StephenButtolph:
```suggestion
	chainIDStr := chainID.String()
	if _, ok := h.handlers[chainIDStr]; ok {
		return false
	}

	h.handlers[chainIDStr] = handler
	return true
```

api/server/http2_router.go
```diff
@@ -0,0 +1,66 @@
...
+
+func newHTTP2Router() *http2Router {
+	return &http2Router{
+		handlers: make(map[string]http.Handler),
+	}
+}
+
+func (h *http2Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
```
StephenButtolph: Do we want this lock held during `handler.ServeHTTP`? It seems like we might be better off only holding the lock when touching the `handlers` map.

api/server/http2_router_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+func TestGRPCRouterAdd(t *testing.T) {
```
StephenButtolph:
```suggestion
func TestHTTP2RouterAdd(t *testing.T) {
```

api/server/http2_router_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require := require.New(t)
+	h := newHTTP2Router()
+	handler := http.HandlerFunc(func(http.ResponseWriter, *http.Request) {})
+
+	require.True(h.Add(ids.Empty, handler))
+	require.False(h.Add(ids.Empty, handler))
+}
+
+func TestGRPCRouterServeHTTP(t *testing.T) {
```
StephenButtolph:
```suggestion
func TestHTTP2RouterServeHTTP(t *testing.T) {
```

api/server/server.go
```diff
@@ -115,33 +117,30 @@ func New(
...
-		func(w http.ResponseWriter, r *http.Request) {
-			// Attach this node's ID as a header
-			w.Header().Set("node-id", nodeID.String())
-			gzipHandler.ServeHTTP(w, r)
-		},
-	)
+	handler := wrapHandler(router, nodeID, allowedOrigins, allowedHosts, true)
+
+	http2Router := newHTTP2Router()
+	// Do not use gzip middleware because it breaks the grpc spec
```
StephenButtolph, line 123: I don't know if this is actually a correct suggestion, but should this be h2 rather than grpc?
```suggestion
	// Do not use gzip middleware because it breaks the http2 spec
```

api/server/server.go
```diff
@@ -199,6 +199,26 @@ func (s *server) RegisterChain(chainName string, ctx *snow.ConsensusContext, vm
...
+		return
+	}
+
+	if http2Handler == nil {
+		return
+	}
+
+	http2Handler = s.wrapMiddleware(chainName, http2Handler, ctx)
+	if !s.http2Router.Add(ctx.ChainID, http2Handler) {
+		s.log.Error("failed to add route to http2 handler")
```
StephenButtolph: can we include the chainID and/or chainName here?

vms/example/xsvm/vm.go
```diff
@@ -152,6 +154,13 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 	}, server.RegisterService(api, constants.XSVMName)
 }
 
+func (vm *VM) CreateHTTP2Handler(context.Context) (http.Handler, error) {
+	server := grpc.NewServer()
+	server.RegisterService(&xsvm.Ping_ServiceDesc, &api.PingService{Log: vm.chainContext.Log})
+
```
StephenButtolph: Just seems like a weird space to me.
```suggestion
```

vms/rpcchainvm/vm_server.go
```diff
@@ -359,6 +359,34 @@ func (vm *VMServer) CreateHandlers(ctx context.Context, _ *emptypb.Empty) (*vmpb
 	return resp, nil
 }
 
+func (vm *VMServer) CreateHTTP2Handler(ctx context.Context, _ *emptypb.Empty) (*vmpb.CreateHTTP2HandlerResponse, error) {
+	handler, err := vm.vm.CreateHTTP2Handler(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// The vm does not expose a gRPC service
```
StephenButtolph:
```suggestion
	// The vm does not expose an HTTP2 handler
```

api/server/http2_router.go
joshua-kim: Ah, this is a good catch. This was because in grpc the path was `/chainID/service/method` but I forgot to update this.

vms/example/xsvm/api/ping.go
joshua-kim: updated https://github.com/ava-labs/avalanchego/pull/3294/commits/2b3a8e3cb08e21d7cce2529c86fc72fce9e5f848

api/server/http2_router.go
joshua-kim: updated: https://github.com/ava-labs/avalanchego/pull/3294/commits/d18fe88baa2a8c9a8678691b172b7aca9d55e2b8

api/server/server.go
joshua-kim: It specifically breaks the grpc spec, but http2 is fine with gzip compression. I assigned Michael to https://github.com/ava-labs/avalanchego/issues/3992 to remove gzip compression entirely from our middleware

api/server/server.go
joshua-kim: Added https://github.com/ava-labs/avalanchego/pull/3294/commits/5878beee696636d034676ac4bb5d9f959211d2f4

StephenButtolph (approved):

joshua-kim:
Related coreth pr https://github.com/ava-labs/coreth/pull/1006

PR #3295, opened 2024-08-14
Refactor e2e tests for P-chain tests

StephenButtolph:
## Why this should be merged

- Improves the readability of some of the E2E tests.

## How this works

- Introduces `e2e.NewPrivateKey` to avoid annoying error checks.

## How this was tested

- [X] CI


tests/e2e/p/staking_rewards.go
```diff
@@ -68,68 +75,50 @@ var _ = ginkgo.Describe("[Staking Rewards]", func() {
...
-			betaDelegationRewardKey,
-			gammaDelegationRewardKey,
-			deltaDelegationRewardKey,
-		}
+			keychain = env.NewKeychain(1)
+			nodeURI  = tmpnet.NodeURI{
+				NodeID: alphaNodeID,
+				URI:    alphaNode.URI,
+			}
+			baseWallet = e2e.NewWallet(tc, keychain, nodeURI)
```
StephenButtolph, line 101: This is the actual change. The keychain being passed into the wallet here only has the single funded key.

This doesn't change any behavior for the current test, because we create new wallets below for checking balances. However, this change is important with dynamic fees because the transactions that are issued later in the file will produce a change output. The default change output is selected from any of the provided keys (which can break the balance checks).

This isn't a problem in the current test because the static fee for all the Primary Network staking is `0` - which means we never include a change output. This is because the wallet consumes locked UTXOs first for staking operations, and the remaining tokens from those transactions must use the same owner that it was consumed from.

tests/e2e/p/workflow.go
```diff
@@ -25,177 +24,163 @@ import (
...
-				require.NoError(err)
-			})
+				},
+				tc.WithDefaultContext(),
+			)
+			require.NoError(err)
 
-			// check balances post export
-			pBalances, err = pWallet.Builder().GetBalance()
+			exportFee, err := pFeeCalculator.CalculateFee(exportTx.Unsigned)
```
StephenButtolph, line 151: This is the functional change for this test. We now use the `pFeeCalculator` rather than referencing the hardcoded static fee.

tests/fixture/e2e/helpers.go
```diff
@@ -49,6 +52,13 @@ const (
 	PrivateNetworksDirName = "private_networks"
 )
 
+// Create a new private key.
+func NewPrivateKey(tc tests.TestContext) *secp256k1.PrivateKey {
+	key, err := secp256k1.NewPrivateKey()
+	require.NoError(tc, err)
+	return key
+}
```
StephenButtolph, line 60: I could pull this out into a separate PR to avoid mixing functional + style changes... But I ended up cleaning up the tests so that I could better understand what they were doing (so that I could fix them for dynamic fees) and ended up making this change as part of it... Will defer to reviewer opinion here.

tests/e2e/p/workflow.go
```diff
@@ -25,177 +24,163 @@ import (
...
 
 // PChainWorkflow is an integration test for normal P-Chain operations
-// - Issues an Add Validator and an Add Delegator using the funding address
-// - Exports AVAX from the P-Chain funding address to the X-Chain created address
-// - Exports AVAX from the X-Chain created address to the P-Chain created address
-// - Checks the expected value of the funding address
+// - Issues an AddPermissionlessValidatorTx
+// - Issues an AddPermissionlessDelegatorTx
+// - Issues an ExportTx on the P-chain and verifies the expected balances
+// - Issues an ImportTx on the X-chain and verifies the expected balances
```
StephenButtolph, line 30: This description was previously misleading... The functionality in this test isn't changed.

aaronbuchwald (approved):

tests/e2e/p/staking_rewards.go
```diff
@@ -229,84 +225,104 @@ var _ = ginkgo.Describe("[Staking Rewards]", func() {
 			require.NoError(err)
 		})
 
-		tc.By("stopping beta node to prevent it and its delegator from receiving a validation reward")
-		require.NoError(betaNode.Stop(tc.DefaultContext()))
+		tc.By("stopping beta node to prevent it and its delegator from receiving a validation reward", func() {
```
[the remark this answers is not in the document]
StephenButtolph: I was under the impression that the reason that `By` took in a function like this was to allow for that function to specify the logic that the `By` text was describing.

Is that not the case? If that isn't the case, what is the purpose of the function? If it's just as a scope delimiter, why wouldn't we just explicitly use scope delimiters (`{` and `}`)?

PR #3296, opened 2024-08-15
Fix e2e tests to support dynamic fees

StephenButtolph:
## Why this should be merged

- Introduces support for dynamic fee txs in the `p/staking_rewards` and `p/workflow` tests.
## How this works

- Introduces `e2e.NewPChainFeeCalculatorFromContext` to allow for calculating the expected fee of a transaction (either dynamic or static).

- Uses `e2e.NewPChainFeeCalculatorFromContext` to fix `p/workflow`.

We now use the `pFeeCalculator` rather than referencing the hardcoded static fee.

- Uses less keys in the wallet to fix `p/staking_rewards`.

The keychain being passed into the wallet for tx issuance now only has the single funded key.

This doesn't change any behavior for the current test, because we create new wallets for checking balances. However, this change is important with dynamic fees because the transactions that are issued later in the file will produce a change output. The default change output is selected from any of the provided keys (which can break the balance checks).

This isn't a problem in the current test because the static fee for all the Primary Network staking is 0 - which means we never include a change output. This is because the wallet consumes locked UTXOs first for staking operations, and the remaining tokens from those transactions must use the same owner that it was consumed from. 

## How this was tested

- [X] CI


StephenButtolph:
Going to merge once CI passes as Aaron approved this change in the prior PR (before it was separated into a different PR)

PR #3305, opened 2024-08-16
[antithesis] Enable custom plugin dir for subnet-evm

marun:
## Why this should be merged

subnet-evm's docker image uses a non-default plugin dir and it needs to be possible to set it for nodes that will run under the antithesis docker compose project.

## How this works

- support setting the plugin dir for the compose project while still using the default plugin dir when initializing the db 

## How this was tested

- CI workflows testing image build in this repo and the workflow testing image build in https://github.com/ava-labs/subnet-evm/pull/1166

### DIFF
--- a/tests/antithesis/compose.go
+++ b/tests/antithesis/compose.go
@@ -61,8 +61,26 @@ func GenerateComposeConfig(network *tmpnet.Network, baseImageName string) error 
 			return fmt.Errorf("failed to get bootstrap volume path: %w", err)
 		}
 
+		// Save the plugin dir for use with compose configuration and remove it from flags so pluginDir can
+		// be used for the db-initialization bootstrap.
+		//
+		// TODO(marun) Separate bootstrap configuration from runtime configuration to avoid having to do this
+		pluginDirForCompose, err := network.GetPluginDir()
+		if err != nil {
+			return fmt.Errorf("failed to get plugin dir: %w", err)
+		}
+		delete(network.DefaultFlags, config.PluginDirKey)
+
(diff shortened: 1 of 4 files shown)


tests/antithesis/compose.go
```diff
@@ -61,9 +61,27 @@ func GenerateComposeConfig(network *tmpnet.Network, baseImageName string) error
...
 		if err := initBootstrapDB(network, avalancheGoPath, pluginDir, bootstrapVolumePath); err != nil {
 			return fmt.Errorf("failed to initialize db volumes: %w", err)
 		}
+
+		if len(pluginDirForCompose) > 0 {
+			// Restore the plugin dir
+			network.DefaultFlags[config.PluginDirKey] = pluginDirForCompose
+		} else {
+			// Ensure the plugin dir is not provided so that the default is used
+			delete(network.DefaultFlags, config.PluginDirKey)
```
ARR4N, line 83: You've already deleted it above, on line 72.

tests/antithesis/compose.go
```diff
@@ -61,9 +61,27 @@ func GenerateComposeConfig(network *tmpnet.Network, baseImageName string) error
 			return fmt.Errorf("failed to get bootstrap volume path: %w", err)
 		}
 
+		// Save the plugin dir for use with compose configuration and remove it from flags so pluginDir can
```
ARR4N, line 64: How does this relate to the `pluginDir` variable set earlier from the environment variable? Why does the flag need to be removed if `pluginDir` is passed as the third argument to `initBootstrapDB()`?

ARR4N (approved):

StephenButtolph (approved):

PR #3309, opened 2024-08-19
Remove crosschain leftovers

ceyonur:
## Why this should be merged

Removes unused fields & functions, left-overs after cross-chain removal. 

## How this works

Removes `sourceChainID` and renames `destinationChainID` to `chainID`

## How this was tested

existing tests should cover this.


darioush (approved):
not sure whether it makes sense to leave some of the comments that refer to "Destination Chain", this is kind of nit picking so no action needed imo

StephenButtolph (approved):

PR #3313, opened 2024-08-20
Introduce ShortNodeID

tsachiherman:
## Why this should be merged

This PR is the first step toward [ACP-20](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/20-ed25519-p2p).
The existing `NodeID` was renamed into `ShortNodeID`, and a new, backward compatible `NodeID` was added.
Given that the ACP-20 changes are expected to be massive, I'd like to merge this backward-compatible change
before moving with the rest of the changes.

## How this works
For most of the places, the usages of `NodeID` would remain as is.
Places that need to retain the classic `NodeID` were modified to be using the `ShortNodeID`.
The change is expected to be in a single direction only. i.e. `ShortNodeID` can be converted into `NodeID` by calling it's `NodeID()` function. No conversion from `NodeID` to `ShortNodeID` should be needed.

## How this was tested

Tested against existing unit tests.


ids/node_id.go
```diff
@@ -21,65 +13,15 @@ const (
...
-
-	var err error
-	*id, err = NodeIDFromString(str[1:lastIndex])
-	return err
-}
-
-func (id *NodeID) UnmarshalText(text []byte) error {
-	return id.UnmarshalJSON(text)
+type NodeID struct {
+	ShortNodeID `serialize:"true"`
```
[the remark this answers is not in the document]
StephenButtolph: `NodeID` is going to be _either_ a `32`-byte or `20`-byte ID once everything is in and implemented (fwiw)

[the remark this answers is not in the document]
tsachiherman: yes, exactly as @StephenButtolph said. This is a "small step forward" in order to reduce the number of files in the subsequent(s) PR, and make the reviewing process easier.

snow/engine/common/appsender/appsender_server.go
StephenButtolph, line 1: Why does this file use `ids.ShortNodeID` rather than keeping it generic as `ids.NodeID`?

snow/validators/gvalidators/validator_state_client.go
StephenButtolph, line 1: Why does this file use `ids.ShortNodeID` rather than keeping it generic as `ids.NodeID`?

vms/platformvm/state/disk_staker_diff_iterator.go
StephenButtolph, line 1: `marshalDiffKey` and `unmarshalDiffKey` seem to make the assumption that `ids.NodeID` has the same length as `ids.ShortNodeID`. This doesn't seem safe in the long term

vms/rpcchainvm/vm_server.go
StephenButtolph, line 1: Why does this file use `ids.ShortNodeID` rather than keeping it generic as `ids.NodeID`?

ids/short_node_id.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/ava-labs/avalanchego/staking"
+	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/utils/hashing"
+)
+
+const ShortNodeIDLen = ShortIDLen
+
+var (
+	EmptyShortNodeID = ShortNodeID{}
```
StephenButtolph, line 19: I've kind of grown to hate this `Empty...` pattern. Feels like we should probably just use the zero value.

Not asking for anything to change in this PR to be clear... Just felt like voicing that.

ids/test_generator.go
StephenButtolph, line 1: Not for this PR - but we should move this file out into an `idstest` package.

vms/platformvm/genesis/genesistest/genesis.go
```diff
@@ -69,7 +69,9 @@ type Config struct {
 
 func New(t testing.TB, c Config) *platformvmgenesis.Genesis {
 	if len(c.NodeIDs) == 0 {
-		c.NodeIDs = DefaultNodeIDs
+		for _, shortNodeID := range DefaultNodeIDs {
+			c.NodeIDs = append(c.NodeIDs, shortNodeID.NodeID())
+		}
```
StephenButtolph, line 74: I don't think we should be transforming the types here. I think the users of this util should be specifying the `NodeID` type they expect to have used.

vms/platformvm/service_test.go
```diff
@@ -623,7 +623,7 @@ func TestGetCurrentValidators(t *testing.T) {
 		found := false
 		for i := 0; i < len(response.Validators); i++ {
 			gotVdr := response.Validators[i].(pchainapi.PermissionlessValidator)
-			if gotVdr.NodeID != nodeID {
+			if gotVdr.NodeID.Compare(nodeID) != 0 {
```
StephenButtolph, line 626: Why change from using `!=` to `.Compare` (here and below)?

ids/node_id.go
StephenButtolph, line 20: If this diff is "forwards safe" I think we should be able to add an additional field (or somehow else change the serialized format of this struct) without breaking a ton of tests.

utils/ips/claimed_ip_port.go
```diff
@@ -50,13 +50,13 @@ func NewClaimedIPPort(
...
+		NodeID:    ids.ShortNodeIDFromCert(cert).NodeID(),
 	}
 
 	packer := wrappers.Packer{
 		Bytes: make([]byte, preimageLen),
 	}
-	packer.PackFixedBytes(ip.NodeID[:])
+	packer.PackFixedBytes(ip.NodeID.Bytes())
 	packer.PackLong(timestamp)
 	ip.GossipID = hashing.ComputeHash256Array(packer.Bytes)
```
StephenButtolph, line 61: This looks dubious. If `ip.NodeID.Bytes()` started returning `32` bytes, it seems like this may behave unexpectedly.

I need to look into this a bit more carefully.

StephenButtolph (commented):
- I think we should be removing `ids.NodeIDLen` entirely. Every usage of it right now seems like it will break in the future.
- `"vms/platformvm/txs".Validator#NodeID` should be a `ids.ShortNodeID` because it is serialized.
  - `"vms/platformvm/txs/fee.intrinsicValidatorBandwidth"` should be using `ids.ShortNodeIDLen`
  - `"vms/platformvm/txs/fee.IntrinsicRemoveSubnetValidatorTxComplexities"` should be using `ids.ShortNodeIDLen`
- `"vms/platformvm/txs.RemoveSubnetValidatorTx".NodeID` should be a `ids.ShortNodeID` because it is serialized.

tests/e2e/p/workflow.go
```diff
@@ -100,12 +100,12 @@ var _ = e2e.DescribePChain("[Workflow]", func() {
 
 		// Use a random node ID to ensure that repeated test runs will succeed
 		// against a network that persists across runs.
-		validatorID, err := ids.ToNodeID(utils.RandomBytes(ids.NodeIDLen))
+		validatorID, err := ids.ToShortNodeID(utils.RandomBytes(ids.NodeIDLen))
```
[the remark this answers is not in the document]
tsachiherman: done.

snow/engine/common/appsender/appsender_server.go
tsachiherman: done

snow/validators/gvalidators/validator_state_client.go
tsachiherman: done

ids/test_generator.go
tsachiherman: created https://github.com/ava-labs/avalanchego/issues/3343 to track this.

vms/platformvm/service_test.go
tsachiherman: given that the `NodeID` is now a struct, using a comparison method allow us to ensure we can retain the comparison functionality even if we were to add, let's say, a non-comparable variable to the struct ( i.e. pointer ).
I figured that since we already having a comparison method, we should use it instead of relying on a mix of comparison techniques.

utils/ips/claimed_ip_port.go
tsachiherman: in the context of this PR, I believe that this change is safe. The `ip.NodeID` is populated few lines above as `NodeID:    ids.ShortNodeIDFromCert(cert).NodeID()`. Hence, guarantee that it's a 20 bytes slice.
Future PRs would need to update this logic, naturally.

vms/platformvm/state/disk_staker_diff_iterator.go
tsachiherman: good point. let me address that.

tsachiherman: done

vms/rpcchainvm/vm_server.go
tsachiherman: done.

StephenButtolph:
Closing due to staleness.

PR #3321, opened 2024-08-21
Rename `components/fee` pkg to `components/gas`

dhrubabasu:
## Why this should be merged

Removes some aliasing

## How this works

Rename + update

## How this was tested

existing UTs


vms/components/gas/config.go
```diff
@@ -1,9 +1,9 @@
 // Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-// The fee package implements dynamic gas pricing specified in ACP-103:
+// The gas package implements dynamic gas pricing specified in ACP-103:
 // https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/103-dynamic-fees
-package fee
+package gas
 
 type Config struct {
```
StephenButtolph, line 8: Should this be:
```golang
type Config struct {
	// Weights to merge fee dimensions into a single gas value.
	Weights Dimensions `json:"weights"`
	// Maximum amount of gas the chain is allowed to store for future use.
	MaxCapacity Gas `json:"maxCapacity"`
	// Maximum amount of gas the chain is allowed to consume per second.
	MaxPerSecond Gas `json:"maxPerSecond"`
	// Target amount of gas the chain should consume per second to keep the fees
	// stable.
	TargetPerSecond Gas `json:"targetPerSecond"`
	// Minimum price per unit of gas.
	MinPrice Price `json:"minPrice"`
	// Constant used to convert excess gas to a gas price.
	ExcessConversionConstant Gas `json:"excessConversionConstant"`
}
```
now?

I think this update would also apply to `AdvanceTime`

vms/platformvm/state/state.go
```diff
@@ -102,8 +102,8 @@ type Chain interface {
 	GetTimestamp() time.Time
 	SetTimestamp(tm time.Time)
 
-	GetFeeState() fee.State
-	SetFeeState(f fee.State)
+	GetFeeState() gas.State
+	SetFeeState(f gas.State)
```
StephenButtolph, line 106: `GasState`?

StephenButtolph (approved):

PR #3322, opened 2024-08-22
Remove mocks for `Staker` and `ScheduledStaker`

dhrubabasu:
## Why this should be merged

These were only used in one place. We should be using the transactions themselves.

## How this works

pretty straightforward

## How this was tested

CI


vms/platformvm/state/staker_test.go
```diff
@@ -133,88 +134,89 @@ func TestStakerLess(t *testing.T) {
...
+		Priority:        stakerTx.CurrentPriority(),
+	}, staker)
+
+	ctrl := gomock.NewController(t)
+	signer := signer.NewMockSigner(ctrl)
+	signer.EXPECT().Verify().Return(errCustom)
+	stakerTx.Signer = signer
 
 	_, err = NewCurrentStaker(txID, stakerTx, startTime, potentialReward)
 	require.ErrorIs(err, errCustom)
```
dhrubabasu, line 167: Needed to add a signer mock to make sure this is tested correctly :(

vms/platformvm/state/staker_test.go
```diff
@@ -133,88 +134,89 @@ func TestStakerLess(t *testing.T) {
...
-	publicKey := bls.PublicFromSecretKey(sk)
-	subnetID := ids.GenerateTestID()
-	weight := uint64(12345)
-	startTime := time.Now()
-	endTime := startTime.Add(time.Hour)
-	potentialReward := uint64(54321)
-	currentPriority := txs.SubnetPermissionedValidatorCurrentPriority
-
-	stakerTx := txs.NewMockStaker(ctrl)
-	stakerTx.EXPECT().EndTime().Return(endTime)
```
ARR4N, line 151: 🤮 

`<rant>`
Expecting a call to a test-double, such that the test fails if it doesn't occur, is too regularly a tautological test of implementation instead of behaviour. Such tests are brittle (and often useless) because changing the code requires changing the test and then there's no way to know if the new implementation is correct. Conversely, if the lack of a call doesn't result in the test failing then the method shouldn't be called `EXPECT()`.
`</rant>`

vms/platformvm/state/staker_test.go
```diff
@@ -133,88 +134,89 @@ func TestStakerLess(t *testing.T) {
...
-	require.Equal(endTime, staker.EndTime)
-	require.Equal(potentialReward, staker.PotentialReward)
-	require.Equal(endTime, staker.NextTime)
-	require.Equal(currentPriority, staker.Priority)
-
-	stakerTx.EXPECT().PublicKey().Return(nil, false, errCustom)
+	publicKey, isNil, err := stakerTx.PublicKey()
+	require.NoError(err)
+	require.True(isNil)
+	require.Equal(&Staker{
```
ARR4N, line 148: 😍

ARR4N (approved):
Thank you for doing this!

StephenButtolph (approved):

PR #3325, opened 2024-08-23
Add example of P-chain partial signing

StephenButtolph:


wallet/subnet/primary/examples/p-partial-signing/main.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	signedTx := &txs.Tx{Unsigned: unsignedTx}
+	for _, signer := range signers {
+		if err := signer.Sign(ctx, signedTx); err != nil {
+			log.Fatal(err)
+		}
+	}
```
StephenButtolph, line 92: Since these signers aren't related to each other, this transaction could be passed around to untrusting parties to sign (the keys do not need to be co-located).

PR #3334, opened 2024-08-26
Add tx complexity helper

StephenButtolph:
## Why this should be merged

This wasn't originally added because it wasn't expected to be useful for block execution. However, it is useful for tooling around block execution. Also, this pattern aligns with existing behavior of the functions exposed by this file.

## How this works

Allows varargs to be passed into `TxComplexity`. The visitor is re-used to avoid performing additional memory allocations.

## How this was tested

- [X] Added benchmark+unit test.

### DIFF
--- a/vms/platformvm/txs/fee/complexity_test.go
+++ b/vms/platformvm/txs/fee/complexity_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
 )
 
-func TestTxComplexity(t *testing.T) {
+func TestTxComplexity_Individual(t *testing.T) {
 	for _, test := range txTests {
 		t.Run(test.name, func(t *testing.T) {
 			require := require.New(t)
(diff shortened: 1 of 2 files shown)


vms/platformvm/txs/fee/complexity_test.go
```diff
@@ -71,6 +101,30 @@ func BenchmarkTxComplexity(b *testing.B) {
...
+
+		tx, err := txs.Parse(txs.Codec, txBytes)
+		require.NoError(err)
+
+		unsignedTxs = append(unsignedTxs, tx.Unsigned)
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, _ = TxComplexity(unsignedTxs...)
```
darioush, line 124: do we need to assign the outputs to global variables so they are not optimized out?

darioush (approved):

vms/platformvm/txs/fee/complexity_test.go
StephenButtolph: afaict the benchmark is running as expected

dhrubabasu (approved):

PR #3335, opened 2024-08-27
UnmarshalFrom support in codec.Codec

containerman17:
## Why this should be merged
This should be merged because it makes sense for HyperSDK to use avalanchego's default codec, but the current implementation is not compatible.

## How this works
This adds an `UnmarshalFrom` function in the `codec.Codec` interface, allowing unpacking directly from `wrappers.Packer` instead of `[]byte`.

## How this was tested
In [codectest.go](https://github.com/ava-labs/avalanchego/blob/b11b3c6c43ab2406d9f109216cc4cd034fa63ff3/codec/codectest/codectest.go#L1158-L1180)

It was also indirectly tested in [this HyperSDK PR](https://github.com/ava-labs/hypersdk/pull/1459).


codec/reflectcodec/type_codec.go
```diff
@@ -523,6 +523,20 @@ func (c *genericCodec) Unmarshal(bytes []byte, dest interface{}) error {
 	return nil
 }
 
+// UnmarshalFrom unmarshals [p.Bytes] into [dest], where [dest] must be a pointer or
+// interface
+func (c *genericCodec) UnmarshalFrom(p *wrappers.Packer, dest interface{}) error {
```
aaronbuchwald, line 528: This looks to be identical to `Unmarshal` with the exception of not enforcing the offset == len(bytes) at the end of unmarshalling.

One small improvement would be to re-use this code in `Unmarshal` to avoid duplicating the same functionality.

I think that would just be `UnmarshalFrom(wrappers.Packer{bytes}, dest)` and then enforce the offset == length check that exists in `Unmarshal` but not `UnmarshalFrom`

aaronbuchwald (approved):
Approved with one optional nit

aaronbuchwald (approved):

darioush (approved):

codec/codec.go
```diff
@@ -23,6 +23,7 @@ var (
 type Codec interface {
 	MarshalInto(interface{}, *wrappers.Packer) error
 	Unmarshal([]byte, interface{}) error
```
StephenButtolph, line 25: Would it make sense to remove `Unmarshal` entirely and have the codec manager do the exhaustion check?

This interface feels weird. (Not really having to do with your PR... but the existing code feels weird)

StephenButtolph: If we'd prefer to leave that as a potential followup - I think that's reasonable too.

aaronbuchwald: I think that should be a follow up instead of being part of this PR

codec/codectest/codectest.go
```diff
@@ -1153,3 +1156,26 @@ func FuzzStructUnmarshal(codec codecpkg.GeneralCodec, f *testing.F) {
...
+
+	bytesLen, err := codec.Size(mySlice)
+	require.NoError(err)
+	require.Equal(3+bytesLen+3, p.Offset)
+
+	p = wrappers.Packer{Bytes: p.Bytes, MaxSize: p.MaxSize, Offset: 3}
+
+	var sliceUnmarshaled []bool
+	require.NoError(codec.UnmarshalFrom(&p, &sliceUnmarshaled))
+	require.Equal(mySlice, sliceUnmarshaled)
```
StephenButtolph, line 1180: Can we test the state of `p` after calling `UnmarshalFrom`?
```suggestion
	var sliceUnmarshaled []bool
	require.NoError(codec.UnmarshalFrom(&p, &sliceUnmarshaled))
	require.Equal(mySlice, sliceUnmarshaled)
	require.Equal(
		wrappers.Packer{
			Bytes:   p.Bytes,
			MaxSize: p.MaxSize,
			Offset:  11,
		},
		p,
	)
```
or something

StephenButtolph: after this lgtm

aaronbuchwald (approved):

StephenButtolph (approved):

PR #3340, opened 2024-08-27
[e2e] Emit a duration-scoped grafana link for each test

marun:
## Why this should be merged

When an e2e failure occurs in CI, it would be useful to start with a metrics link scoped to the duration of execution of the failing test. 

## How this works

 - Defaults to emitting a metrics link at the end of spec execution
   - A test can disable this behavior by setting `e2e.EmitMetricsLink` to `false`
 - Composes a metrics link in a global ginkgo `AfterEach` handler for the shared network UUID and the start/end times of the current spec
   -  The end time is extended by the network shutdown delay to account for the network scrape interval

## How this was tested

CI

### DIFF
--- a/tests/fixture/e2e/flags.go
+++ b/tests/fixture/e2e/flags.go
@@ -12,6 +12,10 @@ import (
 	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 )
 
+// Ensure that this value takes into account the scrape_interval
+// defined in scripts/run_prometheus.sh.
+const networkShutdownDelay = 12 * time.Second
+
 type FlagVars struct {
 	avalancheGoExecPath  string
 	pluginDir            string
@@ -52,9 +56,8 @@ func (v *FlagVars) RestartNetwork() bool {
 
 func (v *FlagVars) NetworkShutdownDelay() time.Duration {
 	if v.delayNetworkShutdown {
-		// Only return a non-zero value if the delay is enabled.  Make sure this value takes
-		// into account the scrape_interval defined in scripts/run_prometheus.sh.
-		return 12 * time.Second
+		// Only return a non-zero value if the delay is enabled.
+		return networkShutdownDelay
 	}
 	return 0
 }
--- a/tests/fixture/e2e/metrics_link.go
+++ b/tests/fixture/e2e/metrics_link.go
@@ -5,6 +5,7 @@ package e2e
 
 import (
 	"strconv"
+	"time"
 
 	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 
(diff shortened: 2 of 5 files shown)


tests/fixture/e2e/metrics_link.go
```diff
@@ -0,0 +1,55 @@
...
+
+	ginkgo "github.com/onsi/ginkgo/v2"
+)
+
+// The ginkgo event handlers defined in this file will be automatically
+// applied to all ginkgo suites importing this package.
+
+// Whether a spec-scoped metrics link should be emitted after the current
+// spec finishes executing.
+var EmitMetricsLink bool
```
StephenButtolph, line 20: Is there a reason this is defined as a global rather than being on the `TestEnvironment`?

tests/fixture/tmpnet/network.go
```diff
@@ -373,7 +373,7 @@ func (n *Network) StartNodes(ctx context.Context, w io.Writer, nodesToStart ...*
 		return err
 	}
 	// Provide a link to the main dashboard filtered by the uuid and showing results from now till whenever the link is viewed
-	if _, err := fmt.Fprintf(w, "\nMetrics: https://grafana-poc.avax-dev.network/d/kBQpRdWnk/avalanche-main-dashboard?&var-filter=network_uuid%%7C%%3D%%7C%s&var-filter=is_ephemeral_node%%7C%%3D%%7Cfalse&from=%d&to=now\n", n.UUID, startTime.UnixMilli()); err != nil {
+	if _, err := fmt.Fprintf(w, "\nMetrics: %s\n", MetricsLinkForNetwork(n.UUID, strconv.FormatInt(startTime.UnixMilli(), 10), "")); err != nil {
```
StephenButtolph, line 376: I feel like there is a lot happening in this line - might make sense to break it out
```suggestion
	startTimeStr := strconv.FormatInt(startTime.UnixMilli(), 10)
	metricsURL := MetricsLinkForNetwork(n.UUID, startTimeStr, "")
	if _, err := fmt.Fprintf(w, "\nMetrics: %s\n", metricsURL); err != nil {
```

tests/fixture/e2e/metrics_link.go
StephenButtolph: I guess the global `TestEnvironment` isn't actually specialized for individual tests.

StephenButtolph: I always feel strongly - but I don't think we need to do it in this PR haha

StephenButtolph (approved):

ceyonur (approved):

tests/fixture/e2e/flags.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 )
 
+// Ensure that this value takes into account the scrape_interval
+// defined in scripts/run_prometheus.sh.
```
aaronbuchwald, line 16: Was this meant as a TODO?

tests/fixture/e2e/metrics_link.go
```diff
@@ -0,0 +1,55 @@
...
+// This event handler attempts to emit a metrics link scoped to the duration
+// of the current spec.
+//
+// TODO(marun) Make this conditional on metrics collection being enabled
+var _ = ginkgo.AfterEach(func() {
+	tc := NewTestContext()
+	env := GetEnv(tc)
+	if env == nil || !EmitMetricsLink {
+		return
+	}
```
aaronbuchwald, line 41: Where is handling `env == nil` necessary?

aaronbuchwald (approved):

PR #3344, opened 2024-08-28
Include disconnected validators in health message

ceyonur:
## Why this should be merged

Adds disconnected validators to the networking health message details. NodeIDs are sorted.

 Example:
```json
 "networking": {
    "disconnectedValidators": [
        "NodeID-7Xhw2mDxuDS44j42TCB6U5579esbSt3Lg",
        "NodeID-GWPcbFJZFfZreETSoWjPimr846mXEKCtu",
        "NodeID-MFrZFVCXPv5iCn6M9K6XduxGTYp891xXZ",
        "NodeID-P7oB2McjBGgW2NXXWVYjV8JEDFoW9xDE5"
    ],
    "percentConnected": 0.2
}
```

## How this works

adds a new field to the health check details and expand the peer Tracker interface.

## How this was tested

* added unit test
* tested in local network

### DIFF
--- a/snow/engine/common/tracker/peers.go
+++ b/snow/engine/common/tracker/peers.go
@@ -38,7 +38,7 @@ type Peers interface {
 	// GetValidators returns the set of all validators
 	// known to this peer manager
 	GetValidators() set.Set[ids.NodeID]
-	// GetConnectedValidators returns the set of all validators
+	// ConnectedValidators returns the set of all validators
 	// that are currently connected
 	ConnectedValidators() set.Set[ids.NodeID]
 }
--- a/snow/engine/common/tracker/peers.go
+++ b/snow/engine/common/tracker/peers.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/exp/maps"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/validators"
@@ -34,6 +35,12 @@ type Peers interface {
 	// SampleValidator returns a randomly selected connected validator. If there
 	// are no currently connected validators then it will return false.
 	SampleValidator() (ids.NodeID, bool)
+	// GetValidators returns the set of all validators
+	// known to this peer manager
+	GetValidators() set.Set[ids.NodeID]
+	// GetConnectedValidators returns the set of all validators
+	// that are currently connected
+	ConnectedValidators() set.Set[ids.NodeID]
 }
 
 type lockedPeers struct {
@@ -103,6 +110,20 @@ func (p *lockedPeers) SampleValidator() (ids.NodeID, bool) {
 	defer p.lock.RUnlock()
 
 	return p.peers.SampleValidator()
+}
+
+func (p *lockedPeers) GetValidators() set.Set[ids.NodeID] {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	return p.peers.GetValidators()
+}
+
+func (p *lockedPeers) ConnectedValidators() set.Set[ids.NodeID] {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	return p.peers.ConnectedValidators()
 }
 
 type meteredPeers struct {
@@ -249,4 +270,15 @@ func (p *peerData) ConnectedPercent() float64 {
 
 func (p *peerData) SampleValidator() (ids.NodeID, bool) {
 	return p.connectedValidators.Peek()
+}
+
+func (p *peerData) GetValidators() set.Set[ids.NodeID] {
+	vdrs := set.Of(maps.Keys(p.validators)...)
+	return vdrs
+}
+
+func (p *peerData) ConnectedValidators() set.Set[ids.NodeID] {
+	// copy in case
+	vdrs := set.Of(p.connectedValidators.List()...)
+	return vdrs
 }
--- a/snow/engine/common/tracker/peers.go
+++ b/snow/engine/common/tracker/peers.go
@@ -273,8 +273,7 @@ func (p *peerData) SampleValidator() (ids.NodeID, bool) {
 }
 
 func (p *peerData) GetValidators() set.Set[ids.NodeID] {
-	vdrs := set.Of(maps.Keys(p.validators)...)
-	return vdrs
+	return set.Of(maps.Keys(p.validators)...)
 }
 
 func (p *peerData) ConnectedValidators() set.Set[ids.NodeID] {
--- a/snow/networking/handler/health.go
+++ b/snow/networking/handler/health.go
@@ -7,6 +7,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
 )
 
 var ErrNotConnectedEnoughStake = errors.New("not connected to enough stake")
@@ -39,7 +42,7 @@ func (h *handler) HealthCheck(ctx context.Context) (interface{}, error) {
 
 func (h *handler) networkHealthCheck() (interface{}, error) {
 	percentConnected := h.peerTracker.ConnectedPercent()
-	details := map[string]float64{
+	details := map[string]interface{}{
 		"percentConnected": percentConnected,
 	}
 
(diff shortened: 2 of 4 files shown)


snow/networking/handler/health.go
```diff
@@ -54,5 +57,24 @@ func (h *handler) networkHealthCheck() (interface{}, error) {
 		)
 	}
 
+	details["disconnectedValidators"] = h.getDisconnectedValidators()
```
ceyonur, line 60: this will always return disconnectedValidators in the health details message. if we want we can move it to above and return only when it's under the minPercentConnected

snow/networking/handler/health.go
```diff
@@ -54,5 +57,24 @@ func (h *handler) networkHealthCheck() (interface{}, error) {
...
+
 	return details, err
 }
+
+func (h *handler) getDisconnectedValidators() set.Set[ids.NodeID] {
```
ceyonur, line 65: set.Set is sorted when marshalled into JSON.

snow/networking/handler/health.go
```diff
@@ -39,7 +42,7 @@ func (h *handler) HealthCheck(ctx context.Context) (interface{}, error) {
 
 func (h *handler) networkHealthCheck() (interface{}, error) {
```
ceyonur, line 43: this is being run for every chain, so we might end up having same disconnected validator IDs for each chain.

snow/engine/common/tracker/peers.go
```diff
@@ -34,6 +35,12 @@ type Peers interface {
 	// SampleValidator returns a randomly selected connected validator. If there
 	// are no currently connected validators then it will return false.
 	SampleValidator() (ids.NodeID, bool)
+	// GetValidators returns the set of all validators
+	// known to this peer manager
+	GetValidators() set.Set[ids.NodeID]
+	// GetConnectedValidators returns the set of all validators
```
darioush, line 41:
```suggestion
	// ConnectedValidators returns the set of all validators
```

darioush (approved):

snow/networking/handler/health.go
```diff
@@ -54,5 +57,24 @@ func (h *handler) networkHealthCheck() (interface{}, error) {
...
 
+	details["disconnectedValidators"] = h.getDisconnectedValidators()
+
 	return details, err
 }
+
+func (h *handler) getDisconnectedValidators() set.Set[ids.NodeID] {
+	allVdrs := h.peerTracker.GetValidators()
+	if allVdrs.Len() == 0 {
+		return set.NewSet[ids.NodeID](0)
```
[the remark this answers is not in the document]
ceyonur: yea makes sense

snow/networking/handler/health.go
ceyonur: Sorry I did not understand this, I directly set it to details without requiring any check for minimum connected. 

At first I intended to move this inside the above if clause, so we only show it when it errors. Currently it shows regardless of if minConnectedPercent>connectedPercent or not (to be a bit more verbose). If we think this can be too noisy, I can move it.

snow/networking/handler/health.go
ceyonur: I think this is only used here rather than being a shared utility function, so IMHO it should be fine. I added cases to the existing UT to cover this.

snow/engine/common/tracker/peers.go
```diff
@@ -250,3 +271,14 @@ func (p *peerData) ConnectedPercent() float64 {
...
+func (p *peerData) GetValidators() set.Set[ids.NodeID] {
+	vdrs := set.Of(maps.Keys(p.validators)...)
+	return vdrs
+}
+
+func (p *peerData) ConnectedValidators() set.Set[ids.NodeID] {
+	// copy in case
+	vdrs := set.Of(p.connectedValidators.List()...)
+	return vdrs
+}
```
StephenButtolph, line 284: `.List()` allocates a slice and then `set.Of` populates the map from the slice, `.Union` will just directly populate the map, without allocating the slice. So I think that would be a bit nicer.

Also I think the comment can be improved a bit.

```suggestion
func (p *peerData) ConnectedValidators() set.Set[ids.NodeID] {
	// The set is copied to avoid future changes from being reflected in the
	// returned set.
	copied := set.NewSet[ids.NodeID](len(p.connectedValidators))
	copied.Union(p.connectedValidators)
	return copied
}
```

snow/engine/common/tracker/peers.go
```diff
@@ -250,3 +271,14 @@ func (p *peerData) ConnectedPercent() float64 {
 func (p *peerData) SampleValidator() (ids.NodeID, bool) {
 	return p.connectedValidators.Peek()
 }
+
+func (p *peerData) GetValidators() set.Set[ids.NodeID] {
+	vdrs := set.Of(maps.Keys(p.validators)...)
+	return vdrs
```
StephenButtolph, line 277:
```suggestion
	return set.Of(maps.Keys(p.validators)...)
```

snow/networking/handler/health.go
```diff
@@ -54,5 +57,27 @@ func (h *handler) networkHealthCheck() (interface{}, error) {
...
+	connectedVdrs := h.peerTracker.ConnectedValidators()
+	if connectedVdrs.Len() == 0 {
+		return allVdrs
+	} else if connectedVdrs.Len() == allVdrs.Len() {
+		return nil
+	}
+	// difference is disconnected validator
+	allVdrs.Difference(connectedVdrs)
+
+	return allVdrs
```
StephenButtolph, line 82: I think the only one of these optimizations that could actually reduce iterations is the `if connectedVdrs.Len() == allVdrs.Len()` check, everything else wouldn't actually have any iterations anyways...

I'd imagine for any large network that there will almost always be a disconnected validator... So I feel like just having more readable code here would be best.


```suggestion
	vdrs := h.peerTracker.GetValidators()
	connectedVdrs := h.peerTracker.ConnectedValidators()
	// vdrs - connectedVdrs is equal to the disconnectedVdrs
	vdrs.Difference(connectedVdrs)
	return vdrs
```

snow/networking/handler/health.go
```diff
@@ -54,5 +57,27 @@ func (h *handler) networkHealthCheck() (interface{}, error) {
 		)
 	}
 
+	disconnectedVdrs := h.getDisconnectedValidators()
+	if disconnectedVdrs.Len() > 0 {
+		details["disconnectedValidators"] = disconnectedVdrs
+	}
```
StephenButtolph, line 63: Is there a reason that we don't populate this field when everyone is connected? Feels like we can just populate `"disconnectedValidators": h..getDisconnectedValidators()` in the map declaration.

snow/networking/handler/health_test.go
```diff
@@ -154,11 +154,16 @@ func TestHealthCheckSubnet(t *testing.T) {
...
+				networkingDetails, ok := networkingMap.(map[string]interface{})
 				require.True(ok)
 				percentConnected, ok := networkingDetails["percentConnected"]
 				require.True(ok)
 				require.Equal(expectedPercentConnected, percentConnected)
+				disconnectedValidators, ok := networkingDetails["disconnectedValidators"]
+				require.True(ok)
+				vdrSet, ok := disconnectedValidators.(set.Set[ids.NodeID])
+				require.True(ok)
+				require.Equal(vdrSet.Len(), testVdrCount-index-1)
```
StephenButtolph, line 166: If we assign `vdrIDs.List()` to `vdrIDsList` then I think this would be much more readable:
```golang
detailsMap, ok := details.(map[string]interface{})
require.True(ok)
require.Equal(
	map[string]interface{}{
		"percentConnected":       expectedPercentConnected,
		"disconnectedValidators": set.Of(vdrIDsList[index+1:]...),
	},
	detailsMap["networking"],
)
```

snow/engine/common/tracker/peers.go
```diff
@@ -34,6 +35,12 @@ type Peers interface {
 	// SampleValidator returns a randomly selected connected validator. If there
 	// are no currently connected validators then it will return false.
 	SampleValidator() (ids.NodeID, bool)
+	// GetValidators returns the set of all validators
+	// known to this peer manager
+	GetValidators() set.Set[ids.NodeID]
+	// ConnectedValidators returns the set of all validators
+	// that are currently connected
+	ConnectedValidators() set.Set[ids.NodeID]
```
StephenButtolph, line 43: It is a bit temping to just have a `DisconnectedValidators` here - but I think this interface is fine

ceyonur: yea if we care about that performance gain here we can add it.

snow/networking/handler/health.go
ceyonur: thought it would be weird to have it for an empty list.

StephenButtolph (approved):

StephenButtolph (approved):

PR #3345, opened 2024-08-28
Abort query if not enough stake is connected

yacovm:
This pull request makes the snowman engine to abort early and avoid sending a query if the node considers the connected stake as insufficient to achieve a successful poll.
  
The rational for this, is to prevent the node to try to agree on blocks when it is in a partition.

The intent here is twofold: (1) To prevent needless resource expenditure, and (2) to prevent snowball from increasing its preference strength while the protocol has no way to make progress.

The higher preference strength in a partition, the longer it takes to recover and achieve consensus once the partition heals.

## Why this should be merged

Without this change, nodes in a partition consistently try to poll each other and depending on the configuration, it may be that all nodes in a partition manage to increase the preference strength over time in snowball (but fail to finalize blocks, because they cannot consistently achieve the required confidence thresholds).
Once the nodes have increased the preference strength of the next block they prefer, even if the partition heals, it will take a long time for the nodes to change their preference, and the consensus will stall for a long time or until a restart of the nodes.

With this change, the nodes will avoid sending polls once they detect they are in a partition, and thus the preference strength of snowball won't increase needlessly. 

## How this works

Before sending a query, a node checks if it senses enough stake is reachable, and aborts early if it is not the case.

## How this was tested

I ran a Fuji node with verbose logging level, and blocked its communication to 33% of the stake via `iptables` rules.
Then, without this change, the node still tries to query other nodes.
However, with this change, the node aborts and does not query anymore:
Screenshot 2024-08-30 at 01 34 41

Instead, it prints that it failed fast and aborted the query:
Screenshot 2024-08-30 at 00 13 35

I also did a more comprehensive test: 

I deployed a 4 node [subnet](https://subnets-test.avax.network/subnets/xmPw1LwL3dGAnKyWRJgopd6MvZkowQn41utA6BRaQHiDFPqdC) on Fuji (testnet) and then setup a network partition where two nodes are on one side and the other two nodes are on the other.
Then I submitted transactions to some nodes on each side of the partition, and monitored a custom metric I added which measures the preference strength of snowball. Afterwards, I let the partition heal and attempted to submit more transactions to the network. 

I repeated the experiment above for both a build from master and a build based on this fix.

In the build based on master, i observed the snowball preference strength go wild, and the network was not functional after the partition healed:

image (1)

In the build based on this fix, the metric didn't change as the queries are not being sent in the first place.
After the partition healed, conflicting blocks were rejected as consensus was re-established:

image

### DIFF
--- a/snow/engine/snowman/engine.go
+++ b/snow/engine/snowman/engine.go
@@ -31,7 +31,10 @@ import (
 	"github.com/ava-labs/avalanchego/utils/units"
 )
 
-const nonVerifiedCacheSize = 64 * units.MiB
+const (
+	nonVerifiedCacheSize = 64 * units.MiB
+	errInsufficientStake = "insufficient connected stake"
+)
 
 var _ common.Engine = (*Engine)(nil)
 
@@ -871,6 +874,10 @@ func (e *Engine) sendQuery(
 	blkBytes []byte,
 	push bool,
 ) {
+	if e.abortDueToInsufficientConnectedStake(blkID) {
+		return
+	}
+
 	e.Ctx.Log.Verbo("sampling from validators",
 		zap.Stringer("validators", e.Validators),
 	)
@@ -914,6 +921,21 @@ func (e *Engine) sendQuery(
 	} else {
 		e.Sender.SendPullQuery(ctx, vdrSet, e.requestID, blkID, nextHeightToAccept)
 	}
+}
+
+func (e *Engine) abortDueToInsufficientConnectedStake(blkID ids.ID) bool {
+	stakeConnectedRatio := e.Config.ConnectedValidators.ConnectedPercent()
+	minConnectedStakeToQuery := float64(e.Params.AlphaConfidence) / float64(e.Params.K)
+
+	if stakeConnectedRatio < minConnectedStakeToQuery {
+		e.Ctx.Log.Debug("dropped query for block",
+			zap.String("reason", errInsufficientStake),
+			zap.Stringer("blkID", blkID),
+			zap.Float64("ratio", stakeConnectedRatio),
+		)
+		return true
+	}
+	return false
 }
 
 // issue [blk] to consensus
--- a/snow/engine/snowman/engine_test.go
+++ b/snow/engine/snowman/engine_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman/snowmantest"
 	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/snow/engine/common/tracker"
 	"github.com/ava-labs/avalanchego/snow/engine/enginetest"
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/ancestor"
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/block/blocktest"
@@ -29,6 +30,7 @@ import (
 	"github.com/ava-labs/avalanchego/snow/snowtest"
 	"github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/version"
 )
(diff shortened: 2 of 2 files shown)


snow/engine/snowman/engine_test.go
```diff
@@ -3182,3 +3185,35 @@ func TestShouldIssueBlock(t *testing.T) {
...
+	conf.Ctx.Log = logging.NewLogger("", logging.NewWrappedCore(logging.Info, buff, logging.Plain.ConsoleEncoder()))
+	conf.Params = snowball.DefaultParameters
+	conf.ConnectedValidators = &mockConnVDR{percent: 0.7, Peers: conf.ConnectedValidators}
+
+	_, _, _, _, engine := setup(t, conf)
+
+	// Gossip will cause a pull query if enough stake is connected
+	require.NoError(engine.Gossip(context.Background()))
+
+	shouldSay := gbytes.Say("WARN snowman/engine.go:935 dropped query for block {\"reason\": \"insufficient stake connected\", \"blkID\": \"j7sgHpXKn6y22DQdBNr5bzidt345dor9L8rL3Mxo9bt91k4ta\", \"ratio\": 0.7}")
```
aaronbuchwald, line 3215: If we move the line where this log is emitted from, will it cause this test to fail?

Maybe we could match the log for just the error string?

Another alternative would be to use the `TestSender` and fail the test if `Gossip` triggers a query unexpectedly.

yacovm: > If we move the line where this log is emitted from, will it cause this test to fail?

Yeah good point. I was in a hurry (it was really late here :-) ) so I just copy-pasted everything.

snow/engine/snowman/engine_test.go
```diff
@@ -3182,3 +3185,35 @@ func TestShouldIssueBlock(t *testing.T) {
...
+	conf.Ctx.Log = logging.NewLogger("", logging.NewWrappedCore(logging.Info, buff, logging.Plain.ConsoleEncoder()))
+	conf.Params = snowball.DefaultParameters
+	conf.ConnectedValidators = &mockConnVDR{percent: 0.7, Peers: conf.ConnectedValidators}
+
+	_, _, _, _, engine := setup(t, conf)
+
+	// Gossip will cause a pull query if enough stake is connected
+	require.NoError(engine.Gossip(context.Background()))
+
+	shouldSay := gbytes.Say("dropped query for block {\"reason\": \"insufficient stake connected\", \"blkID\": \"j7sgHpXKn6y22DQdBNr5bzidt345dor9L8rL3Mxo9bt91k4ta\", \"ratio\": 0.7}")
```
aaronbuchwald, line 3215: Could we go one step further to make sure that this test passes as long as nothing has broken? For example, could we create an error type that will be included in the log when this condition is triggered and then match the logs on only that string? I like to avoid the case where changing an arbitrary string like a log message or the string inside of an error will cause a test to fail.

yacovm: Fair point, done.

snow/engine/snowman/engine_test.go
```diff
@@ -3182,3 +3185,35 @@ func TestShouldIssueBlock(t *testing.T) {
...
+
+func (m *mockConnVDR) ConnectedPercent() float64 {
+	return m.percent
+}
+
+func TestEngineAbortQueryWhenInPartition(t *testing.T) {
+	require := require.New(t)
+
+	// Buffer to record the log entries
+	buff := gbytes.NewBuffer()
```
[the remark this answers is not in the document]
yacovm: Yeah I did notice I added gomega as a dependency. I have mixed feelings about this.
On the one hand I hate adding dependencies for ad-hoc changes.
On the other hand, I didn't want to add more boilerplate code just for that, and I think that gomega is a useful framework for testing. It allows you to [control processes](https://onsi.github.io/gomega/#codegexeccode-testing-external-processes) and also to [poll for changes](https://pkg.go.dev/github.com/onsi/gomega).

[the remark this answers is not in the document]
yacovm: Done.

snow/engine/snowman/engine.go
```diff
@@ -875,6 +885,10 @@ func (e *Engine) sendQuery(
 		zap.Stringer("validators", e.Validators),
 	)
 
+	if e.abortDueToInsufficientConnectedStake(blkID) {
+		return
+	}
```
StephenButtolph, line 890: Should we move this check to be done before logging `sampling from validators`? It feels weird to log that we are doing something and then potentially exiting _before_ doing that thing.

snow/engine/snowman/engine.go
```diff
@@ -190,12 +193,19 @@ func (e *Engine) Gossip(ctx context.Context) error {
 		return nil
 	}
 
+	blkID := e.Consensus.Preference()
+
+	if e.abortDueToInsufficientConnectedStake(blkID) {
+		return nil
+	}
```
StephenButtolph, line 200: I'm not sure that this change is really needed / relevant. The `PullQuery` sent here is used just to learn about potential new blocks (it can't impact the snowball preference strength).

Was it intended to disable fetching new blocks when not connected to sufficient stake as well?

snow/engine/snowman/engine_test.go
```diff
@@ -3182,3 +3184,40 @@ func TestShouldIssueBlock(t *testing.T) {
...
+	return m.percent
+}
+
+type logBuffer struct {
+	bytes.Buffer
+}
+
+func (logBuffer) Close() error {
+	return nil
+}
```
StephenButtolph, line 3203: I'm shocked that there is an `io.NopCloser` for an `io.Reader` but not for an `io.Writer`... pain.

snow/engine/snowman/engine.go
yacovm: yeah good point

snow/engine/snowman/engine.go
yacovm: But doesn't handling of the PullQuery cause chits to be [sent](https://github.com/ava-labs/avalanchego/blob/a922d0f5f03821dc5c032dbd66006a7f46259c8c/snow/engine/snowman/engine.go#L312), which in turn causes snowman [voting](https://github.com/ava-labs/avalanchego/blob/a922d0f5f03821dc5c032dbd66006a7f46259c8c/snow/engine/snowman/engine.go#L400)?

StephenButtolph: The chit gets dropped here: https://github.com/ava-labs/avalanchego/blob/a922d0f5f03821dc5c032dbd66006a7f46259c8c/snow/consensus/snowman/poll/set.go#L121 because we never register a poll for this requestID here

yacovm: oh, I see.

Even though a node that has sent a pull query is responded to with chits, it is only if the pull query originated from [sendQuery](https://github.com/ava-labs/avalanchego/blob/a922d0f5f03821dc5c032dbd66006a7f46259c8c/snow/engine/snowman/engine.go#L916) that it can actually get into snowman.

StephenButtolph (approved):

PR #3352, opened 2024-09-01
[ci] Add test binary and image to support bootstrap testing 

marun:
## Why this should be merged

Periodic bootstrap testing is intended to ensure that bootstrapping issues are detected prior to releasing to testnet and mainnet.

## How this works

- adds a new test binary and image to initialize and monitor a bootstrapping node
- adds an e2e test job to validate the image

## How this was tested

CI + bootstrap testing on eks cluster

## TODO
- [x] Verify that new test job is passing in CI
- [x] Refactor changes rooted at `/tests/bootstrap`
- [x] Record start time on test start and report duration and storage used upon test completion
   - Also include network, whether state sync is enabled, and avalanchego version details in start/resume/completion messages
- [x] Finalize README

### DIFF
--- a/tests/fixture/bootstrapmonitor/e2e/e2e_test.go
+++ b/tests/fixture/bootstrapmonitor/e2e/e2e_test.go
@@ -252,7 +252,7 @@ func buildImage(tc tests.TestContext, imageName string, forceNewHash bool, scrip
 	)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		require.FailNow("Image build failed: %s\nWith output: %s", err, output)
+		require.FailNow("Image build failed: %v\nWith output: %s", err, output)
 	}
 }
 


tests/fixture/bootstrapmonitor/common.go
```diff
@@ -0,0 +1,294 @@
...
+			pod, ok := event.Object.(*corev1.Pod)
+			if !ok {
+				continue
+			}
+
+			if acceptable(&pod.Status) {
+				return nil
+			}
+		case <-ctx.Done():
+			return errors.New("timeout waiting for pod readiness")
```
tsachiherman, line 85: not sure if this would be meaningful, but generally speaking, when you have an expired context, you want to return an error that is derived from context.ContextCanceled or similar.

tests/fixture/bootstrapmonitor/common.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil, fmt.Errorf("failed to wait for pod %s to terminate: %w", qualifiedPodName, err)
+	}
+
+	terminatedPod, err := clientset.CoreV1().Pods(namespace).Get(ctx, createdPod.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve terminated pod %s: %w", qualifiedPodName, err)
+	}
+
+	// Get the image id for the avalanchego image
+	imageID := ""
```
tsachiherman, line 228: I've seen that you have been using this pattern quite frequently within this PR. maybe consider to refactor in some way ? ( future PR )

tests/fixture/bootstrapmonitor/wait.go
```diff
@@ -0,0 +1,194 @@
...
+		return fmt.Errorf("failed to wait for new image to test: %w", err)
+	}
+
+	// Avoid exiting immediately to avoid container restart before the pod is recreated with the new image
+	time.Sleep(5 * time.Minute)
+	return nil
+}
+
+// Determines the current disk usage for the specified directory
+func getDiskUsage(log logging.Logger, dir string) string {
```
tsachiherman, line 152: I'm sure that it would work.. but it's also very error prone.
I'd opt into something like :
```golang
func DirSize(path string) (int64, error) {
    var size int64
    err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
        if err != nil {
            return err
        }
        if !info.IsDir() {
            size += info.Size()
        }
        return err
    })
    return size, err
}
```

tests/fixture/tmpnet/utils.go
```diff
@@ -17,6 +20,29 @@ const (
 	DefaultNodeTickerInterval = 50 * time.Millisecond
 )
 
+func CheckNodeHealth(ctx context.Context, uri string) (*health.APIReply, error) {
```
tsachiherman, line 23: the most common error that I'd expect here is "context timeout". Would that be recoverable ?

tsachiherman (approved):
This is quite a large PR, and I don't have the full ability to analyze it all.
I added few comments and I'm going to approve it so it won't be blocked.

tests/fixture/bootstrapmonitor/common.go
tsachiherman: sure. I'm referring to the pattern you've used here:

tests/fixture/bootstrapmonitor/bootstrap_test_config_test.go
```diff
@@ -0,0 +1,285 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			config, err := bootstrapTestConfigForPod(test.pod, nodeContainerName)
+			if test.expectedErr != nil {
+				require.ErrorIs(err, test.expectedErr)
+				return
+			}
+			require.NoError(err)
+			require.Equal(test.expectedConfig, config)
```
joshua-kim, line 154: I don't think we technically need this if-statement (applicable to other tests added)? We can explicitly set `expectedErr` and `expectedConfig` to `nil/wantConfigHere` and `errHere/nil` right?

tests/fixture/bootstrapmonitor/e2e/e2e_test.go
```diff
@@ -0,0 +1,610 @@
...
+	require.NoError(err)
+}
+
+// waitForLogOutput streams the logs from the specified pod container until the desired output is found or the context times out.
+func waitForLogOutput(tc tests.TestContext, clientset *kubernetes.Clientset, namespace string, podName string, containerName string, desiredOutput string) {
+	// TODO(marun) Figure out why log output is randomly truncated (not flushed?)
+
+	outputLogLine(tc, fmt.Sprintf(
+		"Logs from: %q container of pod %s.%s (may not be complete)", containerName, namespace, podName))
+	outputLogLine(tc,
```
joshua-kim, line 585: Maybe use `strings.Repeat`? Maybe there's a clever way to do this by fetching the terminal width from `x/term` as well.

tests/fixture/bootstrapmonitor/wait.go
```diff
@@ -0,0 +1,194 @@
...
+				zap.String("stderr", stderr.String()),
+			)
+			return ""
+		default:
+			log.Error("du command failed for dir",
+				zap.Int("exitCode", exitError.ExitCode()),
+				zap.String("dir", dir),
+			)
+			return ""
+		}
```
joshua-kim, line 183: Should we just append the original `exitError` to these logs for these cases anyways?

tests/fixture/bootstrapmonitor/common.go
tsachiherman: I've experimented with several permutation for the above; I didn't really like the Index here, since it's not the index that is being looked for.
```golang
        func SelectSliceElement[K any, V any](slc []K, fnc func(e *K) (V, bool), miss func() error) (V, error) {
	        for _, e := range slc {
		        if r, found := fnc(&e); found {
			        return r, nil
		        }
	        }
	        var v V
	        return v, miss()
        }

...
	// Get the image id for the avalanchego image
	imageID, err := SelectSliceElement[corev1.ContainerStatus, string](terminatedPod.Status.ContainerStatuses,
		func(e *corev1.ContainerStatus) (string, bool) {
			return e.ImageID, (e.Name == containerName) && len(containerName) > 0
		},
		func() error {
			return fmt.Errorf("failed to get image id for pod %s", qualifiedPodName)
		})
	if err != nil {
		return nil, err
	}
```

but that doesn't seem to be right. It's definitely more structured, but not necessarily simpler or cleaner.

What I ended up with was this:
```golang
        func FirstContainerStatusByName(slc []corev1.ContainerStatus, name string) corev1.ContainerStatus {
	        for _, status := range slc {
		        if status.Name == name {
			        return status
		        }
	        }
	        return corev1.ContainerStatus{}
        }

...

	// Get the image id for the avalanchego image
	imageID := FirstContainerStatusByName(terminatedPod.Status.ContainerStatuses, containerName).ImageID
	if len(imageID) == 0 {
		return nil, fmt.Errorf("failed to get image id for pod %s", qualifiedPodName)
	}
```

I think that it's a cleaner solution, at the expense of writing the `FirstContainerStatusByName`.
( but please don't let this block your PR )

tests/fixture/bootstrapmonitor/wait.go
tsachiherman: I generally dislike writing linux/darwin specific code when there are equivalent solutions in the GO ecosystem.

I'm going to resolve this one, since I don't want this to block you.

tests/fixture/bootstrapmonitor/bootstrap_test_config.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/ava-labs/avalanchego/chains"
+	"github.com/ava-labs/avalanchego/config"
+	"github.com/ava-labs/avalanchego/version"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const VersionsAnnotationKey = "avalanche.avax.network/avalanchego-versions"
```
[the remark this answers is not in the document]
StephenButtolph: No preference

StephenButtolph (approved):

PR #3354, opened 2024-09-03
Support unmarshalling of json byte slices

StephenButtolph:
## Why this should be merged

`JSONByteSlice` currently doesn't implement `UnmarshalJSON`. This is pretty unusual (and is something that is desired for #3217.

## How this works

1. Correctly supports `nil`
2. Implements `UnmarshalJSON`

## How this was tested

- [X] Added new unit tests
- [X] Fixed tests that relied on marshaling `nil` as `"0x"`

### DIFF
--- a/vms/types/blob_data.go
+++ b/vms/types/blob_data.go
@@ -9,13 +9,33 @@ import (
 	"github.com/ava-labs/avalanchego/utils/formatting"
 )
 
+const nullStr = "null"
+
 // JSONByteSlice represents [[]byte] that is json marshalled to hex
 type JSONByteSlice []byte
 
 func (b JSONByteSlice) MarshalJSON() ([]byte, error) {
+	if b == nil {
+		return []byte(nullStr), nil
+	}
+
 	hexData, err := formatting.Encode(formatting.HexNC, b)
 	if err != nil {
 		return nil, err
 	}
 	return json.Marshal(hexData)
+}
+
+func (b *JSONByteSlice) UnmarshalJSON(jsonBytes []byte) error {
+	if string(jsonBytes) == nullStr {
+		return nil
+	}
+
+	var hexData string
+	err := json.Unmarshal(jsonBytes, &hexData)
+	if err != nil {
+		return err
+	}
+	*b, err = formatting.Decode(formatting.HexNC, hexData)
+	return err
 }
--- a/vms/types/blob_data.go
+++ b/vms/types/blob_data.go
@@ -32,10 +32,13 @@ func (b *JSONByteSlice) UnmarshalJSON(jsonBytes []byte) error {
 	}
 
 	var hexData string
-	err := json.Unmarshal(jsonBytes, &hexData)
+	if err := json.Unmarshal(jsonBytes, &hexData); err != nil {
+		return err
+	}
+	v, err := formatting.Decode(formatting.HexNC, hexData)
 	if err != nil {
 		return err
 	}
-	*b, err = formatting.Decode(formatting.HexNC, hexData)
-	return err
+	*b = v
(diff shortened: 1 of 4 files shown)


vms/types/blob_data.go
```diff
@@ -9,13 +9,33 @@ import (
...
 	hexData, err := formatting.Encode(formatting.HexNC, b)
 	if err != nil {
 		return nil, err
 	}
 	return json.Marshal(hexData)
 }
+
+func (b *JSONByteSlice) UnmarshalJSON(jsonBytes []byte) error {
+	if string(jsonBytes) == nullStr {
+		return nil
```
ARR4N, line 31: Should an explicit `null` empty `b` instead of being a no-op? This would be [in keeping](https://pkg.go.dev/encoding/json#Unmarshal:~:text=Unmarshal%20first%20handles%20the%20case%20of%20the%20JSON%20being%20the%20JSON%20literal%20null.%20In%20that%20case%2C%20Unmarshal%20sets%20the%20pointer%20to%20nil.) with `json.Unmarshal()`. It's also the inverse of the change you made to `JSONByteSlice.MarshalJSON()`.

vms/types/blob_data.go
```diff
@@ -9,13 +9,33 @@ import (
...
+	}
+
+	var hexData string
+	err := json.Unmarshal(jsonBytes, &hexData)
```
ARR4N, line 35: (style) Any reason to not use `if err := ...; err != nil {`?

vms/types/blob_data.go
```diff
@@ -9,13 +9,33 @@ import (
...
+	if err != nil {
+		return err
+	}
+	*b, err = formatting.Decode(formatting.HexNC, hexData)
```
ARR4N, line 39: (optional) I typically try to ensure that error paths don't leak other values, but appreciate that the behaviour is undefined so there's no clear-cut "correct" answer:

```go
v, err := formatting.Decode(...)
if err != nil {
    return err
}
*b = v
return nil
```

vms/types/blob_data_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			jsonBytes, err := json.Marshal(test.value)
+			require.NoError(err)
+			require.Equal(test.expectedJSON, string(jsonBytes))
+
+			var unmarshaled JSONByteSlice
```
ARR4N, line 43: Does this need to support overwriting existing data (related to my comment about unmarshalling `"null"`)? If so, set this to something non-empty to begin with.

ARR4N (approved):
No need for me to re-review the changes so approving.

samliok (approved):

vms/types/blob_data.go
StephenButtolph: This implementation is keeping with the best-practices of `UnmarshalJSON` afaik.

> By convention, to approximate the behavior of [Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) itself, Unmarshalers implement UnmarshalJSON([]byte("null")) as a no-op.

vms/types/blob_data_test.go
StephenButtolph: I added an explicit test for the `null` behavior

vms/types/blob_data.go
ARR4N: Looks like that comment on `Unmarshal()` is incorrect (or I misunderstood it). [Confirmed in playground](https://go.dev/play/p/Szzlkzrkerv).

PR #3355, opened 2024-09-03
Add explicit p-chain current validator check in AddValidator

StephenButtolph:
## Why this should be merged

Adding this check enforces the [invariant](https://github.com/ava-labs/avalanchego/blob/7d2e631b63317b788399aff2486ee5ca94e9da1d/vms/platformvm/state/stakers.go#L278) documented in the `state.diff` implementation without making assumptions on the interaction pattern between the execution and the state packages.

## How this works

Adds an explicit check to see if the validator was previously removed in the diff.

## How this was tested

- [X] Added unit test

### DIFF
--- a/vms/platformvm/vm_regression_test.go
+++ b/vms/platformvm/vm_regression_test.go
@@ -1402,6 +1402,10 @@ func TestAddValidatorDuringRemoval(t *testing.T) {
 	// Accept firstAddSubnetValidatorTx
 	require.NoError(buildAndAcceptStandardBlock(vm))
 
+	// Verify that the validator was added
+	_, err = vm.state.GetCurrentValidator(subnetID, nodeID)
+	require.NoError(err)
+
 	secondEndTime := firstEndTime.Add(duration)
 	secondSubnetValidatorTx, err := wallet.IssueAddSubnetValidatorTx(&txs.SubnetValidator{
 		Validator: txs.Validator{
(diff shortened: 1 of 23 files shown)


vms/platformvm/vm_regression_test.go
```diff
@@ -1366,6 +1367,62 @@ func TestRemovePermissionedValidatorDuringPendingToCurrentTransitionTracked(t *t
...
+	require.NoError(err)
+
+	vm.clock.Set(firstEndTime)
+	vm.ctx.Lock.Unlock()
+	err = vm.issueTxFromRPC(secondSubnetValidatorTx)
+	vm.ctx.Lock.Lock()
+	require.ErrorIs(err, state.ErrAddingStakerAfterDeletion)
+
+	// Remove the first subnet validator
+	require.NoError(buildAndAcceptStandardBlock(vm))
```
ceyonur, line 1423: should we add a check if that the validator indeed got removed?

vms/platformvm/vm_regression_test.go
```diff
@@ -1366,6 +1367,62 @@ func TestRemovePermissionedValidatorDuringPendingToCurrentTransitionTracked(t *t
...
+			End:    uint64(secondEndTime.Unix()),
+			Wght:   1,
+		},
+		Subnet: subnetID,
+	})
+	require.NoError(err)
+
+	vm.clock.Set(firstEndTime)
+	vm.ctx.Lock.Unlock()
+	err = vm.issueTxFromRPC(secondSubnetValidatorTx)
```
ceyonur, line 1418: similar to below: should we add a check to confirm the validator is still there?

tsachiherman (approved):

michaelkaplan13 (approved):

vms/platformvm/vm_regression_test.go
StephenButtolph: sure

vms/platformvm/vm_regression_test.go
StephenButtolph: I added some additional tests that I think are reasonable

ceyonur (approved):

PR #3356, opened 2024-09-03
[tmpnet] Avoid calling Stat in advance of Read/Open

marun:
## Why this should be merged

Checking for a file/dir in advance of attempting to open it is redundant and invites race conditions where error handling is not duplicated:

https://github.com/ava-labs/avalanchego/actions/runs/10688505984/job/29628507363

## How this works

Removes calls to Stat in tmpnet code and ensures error handling for ErrNotExist in the calls to read a file or dir. 

## How this was tested

CI

### DIFF
--- a/tests/fixture/tmpnet/subnet.go
+++ b/tests/fixture/tmpnet/subnet.go
@@ -339,14 +339,10 @@ func WaitForActiveValidators(
 
 // Reads subnets from [network dir]/subnets/[subnet name].json
 func readSubnets(subnetDir string) ([]*Subnet, error) {
-	if _, err := os.Stat(subnetDir); os.IsNotExist(err) {
+	entries, err := os.ReadDir(subnetDir)
+	if os.IsNotExist(err) {
 		return nil, nil
 	} else if err != nil {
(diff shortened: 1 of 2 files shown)


tests/fixture/tmpnet/subnet.go
```diff
@@ -339,14 +339,10 @@ func WaitForActiveValidators(
 
 // Reads subnets from [network dir]/subnets/[subnet name].json
 func readSubnets(subnetDir string) ([]*Subnet, error) {
-	if _, err := os.Stat(subnetDir); os.IsNotExist(err) {
+	entries, err := os.ReadDir(subnetDir)
+	if os.IsNotExist(err) {
```
StephenButtolph, line 343: I think the doc on `os.IsNotExist` says:

> // This function predates errors.Is. It only supports errors returned by
// the os package. New code should use errors.Is(err, fs.ErrNotExist).

Should we align the error checks to this new format (which is what we currently use in `readState`)

StephenButtolph (approved):

darioush (approved):

PR #3359, opened 2024-09-04
Align codec interface to exclusively use packers

StephenButtolph:
## Why this should be merged

Because the interface doesn't include a `Marshal` function, it feels very weird to have an `Unmarshal` function.

## How this works

Removes the `Unmarshal` function from the interface.

## How this was tested

- [X] Existing CI

### DIFF
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -14,7 +14,6 @@ var (
 	ErrMaxSliceLenExceeded       = errors.New("max slice length exceeded")
 	ErrDoesNotImplementInterface = errors.New("does not implement interface")
 	ErrUnexportedField           = errors.New("unexported field")
-	ErrExtraSpace                = errors.New("trailing buffer space")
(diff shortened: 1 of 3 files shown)


tsachiherman (approved):
it would impact non-genericCodec unmarshalers as well. I don't know if we have any.
generally speaking, this would prevent serializing a series of objects into the same byte array in a sequence. This might be ok if we don't have any such usage.

darioush (approved):

StephenButtolph:
> it would impact non-genericCodec unmarshalers as well. I don't know if we have any. generally speaking, this would prevent serializing a series of objects into the same byte array in a sequence. This might be ok if we don't have any such usage.

Good point, we don't have any such usage. I think going forward this PR (and it's parent) standardize this a bit. After these PRs the manager enforces a codec version with no trailing data, the internal codecs don't know about the codec version and support reading from a sequence.

PR #3361, opened 2024-09-04
Avoid including 'failed' in debug log messages

marun:
When using logs to diagnose a problem, 'failed' would ideally be reserved for a serious problem rather than something that occurs commonly.

### DIFF
--- a/snow/engine/snowman/engine.go
+++ b/snow/engine/snowman/engine.go
@@ -587,7 +587,7 @@ func (e *Engine) sendChits(ctx context.Context, nodeID ids.NodeID, requestID uin
 			// Because we only return accepted state here, it's fairly likely
 			// that the requested height is higher than the last accepted block.
 			// That means that this code path is actually quite common.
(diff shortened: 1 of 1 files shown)


snow/engine/snowman/engine.go
```diff
@@ -630,7 +630,7 @@ func (e *Engine) sendChits(ctx context.Context, nodeID ids.NodeID, requestID uin
 		var ok bool
 		preferenceAtHeight, ok = e.Consensus.PreferenceAtHeight(requestedHeight)
 		if !ok {
-			e.Ctx.Log.Debug("failed fetching processing block",
+			e.Ctx.Log.Debug("processing block not found",
```
StephenButtolph, line 633: I think the above `failed fetching...` logs in this function should be updated as well.

StephenButtolph (approved):

aaronbuchwald (approved):

yacovm (approved):

PR #3364, opened 2024-09-05
Disable Etna by default on local networks

michaelkaplan13:
## Why this should be merged
When running a local network, the Etna changes should not be activated by default until they are more stable (i.e. activated on Fuji). Otherwise, the local network may break when updating to future AvalancheGo versions.

## How this works
Self explanatory

## How this was tested
Built and ran `/build/avalanchego --network-id=12345`. Confirmed activation timestamps using
```
curl --location 'http://localhost:9650/ext/info' \
--data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"info.upgrades"
}'
```

### DIFF
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -70,7 +70,9 @@ var (
 		CortinaTime:                  InitiallyActiveTime,
 		CortinaXChainStopVertexID:    ids.Empty,
 		DurangoTime:                  InitiallyActiveTime,
-		EtnaTime:                     InitiallyActiveTime,
+		// Etna is left unactivated by default on local networks. It can be configured to
+		// activate by overriding the activation time in the upgrade file.
+		EtnaTime: UnscheduledActivationTime,
 	}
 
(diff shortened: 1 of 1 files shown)


StephenButtolph (commented):
lgtm... @ceyonur is this a problematic usage of `upgrade.Default` here? https://github.com/ava-labs/avalanchego/blob/75eb89fc6ef57e62bcfbd94169d7610c3129c907/tests/fixture/tmpnet/genesis.go#L153-L154

ceyonur:
> lgtm... @ceyonur is this a problematic usage of `upgrade.Default` here?
> 
> https://github.com/ava-labs/avalanchego/blob/75eb89fc6ef57e62bcfbd94169d7610c3129c907/tests/fixture/tmpnet/genesis.go#L153-L154

That config doesn't really matter, we can only set chainID there.

StephenButtolph (approved):

ceyonur (approved):

PR #3367, opened 2024-09-06
Implement ACP-77 fee calculations

StephenButtolph:
## Why this should be merged

The fees logic is factored out of #3287.

## How this works

ACP-77 charges validators fees over time. While it may be possible to solve for these fees directly using integrals, the approximation function makes performing this optimization difficult.

Therefore, this PR does not attempt to calculate the fees in constant time (w.r.t. the duration). Rather, it permits falling back to a linear iteration (based on the duration of the fee charge). This is fine for our performance needs, as the duration is limited to the number of seconds between blocks on the chain and calculating weeks of fees takes milliseconds.

There are some special cases that are further optimized.

There is also a derivation of the `excessConversionConstant` based on the `capacity`, `target`, and `minimumTimeToDouble`.

## How this was tested

- [X] New unit tests
- [X] New benchmarks
- [X] Both `CalculateCost` and `CalculateDuration` have "naive" implementations that were used to fuzz the optimized implementations.


vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+	duration uint64,
+) gas.Gas {
+	if current < target {
+		return excess.SubPerSecond(target-current, duration)
+	}
+	return excess.AddPerSecond(current-target, duration)
+}
+
+// CalculateCost calculates the how much to charge based on the dynamic fee
+// mechanism for [duration].
```
aaronbuchwald, line 29:
```suggestion
// CalculateCost calculates how much to charge based on the dynamic fee
// mechanism for [duration].
```

aaronbuchwald (approved):

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -0,0 +1,372 @@
...
+	maxExcessIncreasePerSecond    = capacity - target
+	doubleEvery                   = day
+	excessIncreasePerDoubling     = maxExcessIncreasePerSecond * doubleEvery
+	ln2                           = .6931471805599453094172321214581765680755001343602552541206800094
+	excessConversionConstantFloat = excessIncreasePerDoubling / ln2
+)
+
+var (
+	_excessConversionConstantFloat = excessConversionConstantFloat
+	excessConversionConstant       = gas.Gas(_excessConversionConstantFloat)
```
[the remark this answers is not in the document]
StephenButtolph: `gas.Gas(excessConversionConstantFloat)` doesn't compile so we need to assign the constant to a temporary var... cc: @ARR4N on if you have any suggestions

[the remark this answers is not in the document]
ARR4N: All [instances I can find in the SDK](https://github.com/search?q=repo%3Agolang%2Fgo+language%3AGo+%2F%5Cs_%5Ba-z%5D%2F&type=code&p=2) appear to use an underscore prefix if (1) they want to adhere to some non-Go style like `O_RDONLY` but make it internal; (2) they appear to be using a variable from a canonical implementation (e.g. `sha256`); and/or (3) they are so old that they probably pre-date any style guide.

I know @StephenButtolph won't like it but I lean to an identity function here:

```go
// ftoi truncates f to an int; `int(f)` is preferred and ftoi should only be used if its argument is a `const`.
func ftoi(f float64) int { return int(f) }

const theBestSequence = 0.142857
var foo = ftoi(theBestSequence)
```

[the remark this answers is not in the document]
StephenButtolph: twisting my arm here (I added a `floatToGas` function).

StephenButtolph:
TODO: Add dhruba as co-author.

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+		return cost
+	}
+
+	var (
+		cost uint64
+		err  error
+	)
+	for i := uint64(0); i < duration; i++ {
+		excess = CalculateExcess(target, current, excess, 1)
+		// If the excess is 0, the price will remain constant for all future
```
ARR4N, line 54: This took a long time for me to grok (including asking you in a DM) so I think it's ok to be verbose in the explanation. Here's a first draft of my understanding that you can copy and tweak:

_The change in excess is positively correlated with the difference between current and target. It is also strictly bounded in [0,2^64) with (under/over)flows being clipped. As we know that current != target, excess would be strictly increasing or strictly decreasing if it weren't for clipping. However, when coupled with clipping, this guarantees that a zero excess will remain zero for all future iterations._

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		return cost
+	}
+
+	var (
+		cost uint64
+		err  error
+	)
+	for i := uint64(0); i < duration; i++ {
+		excess = CalculateExcess(target, current, excess, 1)
```
ARR4N, line 53: (optional) Following on from comment above, this could be `feeCtx.UpdateExcess(1)`. On first pass I missed the lack of `:=` so didn't appreciate that it was an update.

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+	current gas.Gas,
+	minPrice gas.Price,
+	excessConversionConstant gas.Gas,
+	excess gas.Gas,
+	maxDuration uint64,
+	targetCost uint64,
+) uint64 {
+	// Because this function can divide by prices, we need to sanity check the
+	// parameters to avoid division by 0.
+	if minPrice == 0 {
```
ARR4N, line 93: Why not return an error?

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		duration++
+		price := gas.CalculatePrice(minPrice, excess, excessConversionConstant)
+		cost, err = safemath.Add(cost, uint64(price))
+		if err != nil {
+			return duration
+		}
+	}
+	return duration
```
ARR4N, line 141: This will always be `maxDuration` so I think it should be explicit.

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+			)
+
+			duration, err = safemath.Add(duration, durationWithZeroExcess)
+			if err != nil || duration >= maxDuration {
+				return maxDuration
+			}
+			return duration
+		}
+
+		duration++
```
ARR4N, line 134: I think this should be in the loop. As far as I can tell, the only branch it would affect would be the error on `safemath.Add(cost...)`.

```go
for ; cost < targetCost && duration < maxDuration; duration++ {
```

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+		price := gas.CalculatePrice(minPrice, excess, excessConversionConstant)
+		cost, err = safemath.Add(cost, uint64(price))
+		if err != nil {
+			return duration
```
ARR4N, line 138: Why `duration` and not `maxDuration`?

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -0,0 +1,402 @@
...
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+const (
+	second = 1
+	minute = 60 * second
+	hour   = 60 * minute
+	day    = 24 * hour
+	week   = 7 * day
+	year   = 365 * day
```
ARR4N, line 23: Not 525,600 minutes?

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+	minPrice = 2_048
+
+	capacity                      = 20_000
+	target                        = 10_000
+	maxExcessIncreasePerSecond    = capacity - target
+	doubleEvery                   = day
+	excessIncreasePerDoubling     = maxExcessIncreasePerSecond * doubleEvery
+	ln2                           = .6931471805599453094172321214581765680755001343602552541206800094
```
ARR4N, line 32: Any reason to not use `math.Ln2`?

ARR4N (commented):
I need to concentrate on the tests and there's currently construction outside my window so I'm sending the rest of the review early.

vms/platformvm/validators/fee/fee.go
StephenButtolph: I felt like handling this edge case was cleaner than adding an error to the return value.

vms/platformvm/validators/fee/fee.go
StephenButtolph: This is not always `maxDuration`. (The tests will fail if this is changed to `maxDuration`). The loop exits if `duration == maxDuration` _or_ if `cost >= targetCost`

vms/platformvm/validators/fee/fee.go
StephenButtolph: Having:
```golang
if err != nil {
	return duration + 1
}
```
feels pretty weird to me... Personally I find the code as-is more readable... but if you have conviction then I'm fine moving the increment around.

vms/platformvm/validators/fee/fee.go
StephenButtolph: We are looking for the duration where `cost >= targetCost`. `cost` would be pinned to `MaxUint64` here - so we know that `cost` is `>=` any possible `targetCost` at this `duration`.

vms/platformvm/validators/fee/fee_test.go
StephenButtolph: I'm not sure if this comment is implying that `year` should be set to `52 * week` rather than `365 * day` or not haha... But the rest of the P-chain uses this definition of years.

The chain doesn't really need to care about any leap seconds or leap years... We can set our own time cadence... We use this definition of a year for our maximum staking duration (among other things)

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+	excessConversionConstant gas.Gas,
+	excess gas.Gas,
+	duration uint64,
+) uint64 {
+	// If the current and target are the same, the price is constant.
+	if current == target {
+		price := gas.CalculatePrice(minPrice, excess, excessConversionConstant)
+		cost, err := safemath.Mul(duration, uint64(price))
+		if err != nil {
+			return math.MaxUint64
```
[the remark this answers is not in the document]
StephenButtolph: This is actually a _discount_!  (because this is just capping the cost at `MaxUint64` if it would have overflowed)

vms/platformvm/validators/fee/fee.go
```diff
@@ -0,0 +1,161 @@
...
+	"math"
+
+	"github.com/ava-labs/avalanchego/vms/components/gas"
+
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+// CalculateExcess updates the excess value based on the target and current
+// usage over the provided duration.
+func CalculateExcess(
```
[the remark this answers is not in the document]
StephenButtolph: The thing that is a bit weird is: `target`, `minPrice`, and `excessConversionConstant` are all essentially constants. `current` and `excess` are dynamic (and persisted to state). `duration`, `maxDuration`, and `targetCost` are the "real" arguments for these functions.

I had considered having:
```golang
type Config struct {
	Target                   gas.Gas
	MinPrice                 gas.Price
	ExcessConversionConstant gas.Gas
}

type State struct {
	Current gas.Gas
	Excess  gas.Gas
}
```

But then we'd need to be passing both of these into `CalculateExcess`... I'll try to play around with some things to see if I can find something I don't dislike

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -0,0 +1,402 @@
...
+)
+
+var (
+	excessConversionConstant = floatToGas(excessConversionConstantFloat)
+
+	tests = []struct {
+		name                string
+		target              gas.Gas
+		current             gas.Gas
+		excess              gas.Gas
```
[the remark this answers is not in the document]
StephenButtolph: Been folded into `state`.

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -0,0 +1,402 @@
...
+		cost                uint64
+		excessAfterDuration gas.Gas
+	}{
+		{
+			name:                "excess=0, current<target, minute",
+			target:              10_000,
+			current:             10,
+			excess:              0,
+			duration:            minute,
+			cost:                122_880,
```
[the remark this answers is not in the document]
StephenButtolph: The "canonical source" is the unoptimized implementations.

vms/platformvm/validators/fee/fee.go
StephenButtolph: I didn't want `AdvanceTime` to mutate the `State` because `State` is going to be passed around in the execution and we don't want to run into a situation where we accidentally mutate it.

vms/platformvm/validators/fee/fee_test.go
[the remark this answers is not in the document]
StephenButtolph: See: https://github.com/avalanche-foundation/ACPs/pull/119

vms/platformvm/validators/fee/fee.go
```diff
@@ -11,34 +11,44 @@ import (
...
+// If Excess would underflow, it is set to 0.
+// If Excess would overflow, it is set to MaxUint64.
+func (s State) AdvanceTime(target gas.Gas, seconds uint64) State {
+	excess := s.Excess
+	if s.Current < target {
+		excess = excess.SubPerSecond(target-s.Current, seconds)
+	} else if s.Current > target {
+		excess = excess.AddPerSecond(s.Current-target, seconds)
+	}
+	return State{
```
ARR4N, line 39: Is this because we were talking about functional languages? 🤣

vms/platformvm/validators/fee/fee.go
```diff
@@ -76,86 +89,77 @@ func CalculateCost(
 	return cost
 }
 
-// CalculateDuration calculates the duration that it would take to charge at
+// SecondsUntil calculates the number of seconds that it would take to charge at
```
ARR4N, line 92: Nice naming! I love code that reads like prose.

ARR4N (approved):
LGTM!

vms/platformvm/validators/fee/fee.go
StephenButtolph: It's actually because we might be passing these through different block states... And we definitely don't want them to be mutable there

PR #3370, opened 2024-09-07
Add SoV state struct

StephenButtolph:
## Why this should be merged

Factored out of #3287. This PR does not include the optional `EndTime` because I'm not convinced that the current definition of `RecoverSubnetTx` is comprehensive. 

## How this works

This adds the state accessors, and definition, for SoV structs.

## How this was tested

- [X] Adds unit tests for `Less`, `put*`, `get*`, and `delete*`


StephenButtolph:
TODO: Add dhruba as co-author.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/google/btree"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/platformvm/block"
+)
+
+var _ btree.LessFunc[*SubnetOnlyValidator] = (*SubnetOnlyValidator).Less
+
+type SubnetOnlyValidator struct {
```
ceyonur, line 18: Should we also add StartTime as a field here?

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -0,0 +1,92 @@
...
+	NodeID   ids.NodeID `serialize:"true"`
+
+	// PublicKey is the uncompressed BLS public key of the validator. It is
+	// guaranteed to be populated.
+	PublicKey []byte `serialize:"true"`
+
+	// Weight of this validator. It can be updated when the MinNonce is
+	// increased. If the weight is being set to 0, the validator is being
+	// removed.
+	Weight uint64 `serialize:"true"`
```
ceyonur, line 33: I wonder Weight will be used in a formula to determine if SoV is an active/deactive one. I think using Weight solely could result into a misinterpration if the underlying SoV is being deleted rather than deactivating.

StephenButtolph: Weight is not related to the status of the validator

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: Yeah we'll need that for the WARP message + uptimes 👍

vms/platformvm/state/subnet_only_validator.go
ceyonur: how/where should we put/deduce that info? as a field here (assuming we'd want to persist that info)?

aaronbuchwald (approved):

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: There will be different partitions in disk to include this information. This is needed because the active validator set will be iterated on startup - but the inactive validators will only live on disk (and we won't iterate over them on startup)

ceyonur (approved):

tsachiherman (approved):

PR #3371, opened 2024-09-07
Add AccruedFees to state.Chain interface

StephenButtolph:
## Why this should be merged

In ACP-77 the chain must track the accrued fees for SoV validators.

## How this works

Adds a new singleton data field: `AccruedFees` and enables the usage of this field through the `state.Chain` interface.

## How this was tested

- [X] Added unit tests


aaronbuchwald (approved):
LGTM

PR #3372, opened 2024-09-08
[merkledb benchmark] implement simple write profile benchmark

tsachiherman:


x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,372 @@
...
+)
+
+var (
+	databaseEntries = pflag.Uint64("n", defaultDatabaseEntries, "number of database entries")
+	httpMetricPort  = pflag.Uint64("p", defaultMetricsPort, "default metrics port")
+	verbose         = pflag.Bool("v", false, "verbose")
+)
+
+func getMerkleDBConfig(promRegistry prometheus.Registerer) merkledb.Config {
+	const defaultHistoryLength = 300
```
rkuris, line 42: Let's use the same number as firewood of 120: https://github.com/ava-labs/firewood/blob/main/firewood/src/manager.rs#L21

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,372 @@
...
+	return merkledb.Config{
+		BranchFactor:                merkledb.BranchFactor16,
+		Hasher:                      merkledb.DefaultHasher,
+		RootGenConcurrency:          0,
+		HistoryLength:               defaultHistoryLength,
+		ValueNodeCacheSize:          units.MiB,
```
rkuris, line 48: These seem really small. If I am reading this correctly, there is about 2Mb of total cache?

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	mdb, err := merkledb.New(context.Background(), levelDB, getMerkleDBConfig(nil))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to create merkleDB database : %v\n", err)
+		return err
+	}
+
+	fmt.Printf("Initializing database.")
+	ticksCh := make(chan interface{})
```
rkuris, line 112: nit: do you really need this ticker? Might be easier to report every 100k rows or something.

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,372 @@
...
+	"github.com/ava-labs/avalanchego/database/leveldb"
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/units"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+)
+
+const (
+	defaultDatabaseEntries    = 2000000
+	databaseCreationBatchSize = 1000000
```
rkuris, line 29: Batch size is supposed to be 10k. This is 1M.

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,372 @@
...
+		startDeleteTime := time.Now()
+		for keyToDeleteIdx := low; keyToDeleteIdx < low+databaseRunningBatchSize; keyToDeleteIdx++ {
+			entryHash := calculateIndexEncoding(keyToDeleteIdx)
+			err = batch.Delete(entryHash)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "unable to delete merkleDB entry : %v\n", err)
+				return err
+			}
+		}
+		deleteDuration = time.Since(startDeleteTime)
```
rkuris, line 297: You can avoid all this math and just report the raw number of deletes. Grafana can convert this to a rate for you.

rkuris (changes requested):

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,404 @@
...
+	insertRate = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "merkledb_bench",
+		Name:      "entry_insert_rate",
+		Help:      "The rate at which elements are inserted",
+	})
+	batchWriteRate = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "merkledb_bench",
+		Name:      "batch_write_rate",
+		Help:      "The rate at which the batch was written",
+	})
```
StephenButtolph, line 79: We should not be calculating the rates in the benchmark. The prometheus server should do this based on the counts.
```suggestion
```

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,404 @@
...
+	err = mdb.Close()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to close levelDB database : %v\n", err)
+		return err
+	}
+	err = levelDB.Close()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to close merkleDB database : %v\n", err)
+		return err
+	}
```
StephenButtolph, line 223: The logs seem inverted here

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,404 @@
...
+
+func getMerkleDBConfig(promRegistry prometheus.Registerer) merkledb.Config {
+	const defaultHistoryLength = 120
+	return merkledb.Config{
+		BranchFactor:                merkledb.BranchFactor16,
+		Hasher:                      merkledb.DefaultHasher,
+		RootGenConcurrency:          0,
+		HistoryLength:               defaultHistoryLength,
+		ValueNodeCacheSize:          units.MiB,
+		IntermediateNodeCacheSize:   1024 * units.MiB,
```
StephenButtolph, line 91: How much memory are we using? Feels like we could probably increase these

x/merkledb/benchmarks/benchmark.go
tsachiherman: done.

x/merkledb/benchmarks/benchmark.go
tsachiherman: I attempted to tweak this value, but it had no performance impact.

x/merkledb/benchmarks/benchmark.go
tsachiherman: reporting every 100k rows doesn't work nicely because of the batch writing ( which blocks for a long time ).

x/merkledb/benchmarks/benchmark.go
tsachiherman: done.

x/merkledb/benchmarks/benchmark.go
tsachiherman: I've added both. I believe that my calculation would be more accurate, but let's have both for the time being.

x/merkledb/benchmarks/benchmark.go
tsachiherman: I believe that it won't generate accurate results, since we're mixing batch writing and put in the same sequence.
I've included both the counter and the rate metrics so that we can get both numbers in the grafana.

x/merkledb/benchmarks/benchmark.go
tsachiherman: fixed.

x/merkledb/benchmarks/benchmark.go
tsachiherman: I've attempted to tweak these, but haven't seen any concrete gains. Adjusting the leveldb config was helpful, though.

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+		addDuration = time.Since(startInsertTime)
+		insertRate.Set(float64(databaseRunningBatchSize) * float64(time.Second) / float64(addDuration))
+		insertsCounter.Add(databaseRunningBatchSize)
+
+		// update middle 5k entries
+		startUpdateTime := time.Now()
+		for keyToUpdateIdx := low + ((*databaseEntries) / 2); keyToUpdateIdx < low+((*databaseEntries)/2)+databaseRunningUpdateSize; keyToUpdateIdx++ {
+			updateEntryKey := calculateIndexEncoding(keyToUpdateIdx)
+			updateEntryValue := calculateIndexEncoding(keyToUpdateIdx - ((*databaseEntries) / 2))
```
rkuris, line 315: This is incorrect, should be:
```suggestion
			updateEntryValue := calculateIndexEncoding(low)
```

x/merkledb/benchmarks/benchmark.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to create merkleDB database : %v\n", err)
+		return err
+	}
+	defer func() {
+		mdb.Close()
+		levelDB.Close()
+	}()
+
+	low := uint64(0)
```
rkuris, line 277: `low` never changes, and should be increased by 2.5k each pass

rkuris (changes requested):

x/merkledb/benchmarks/benchmark.go
tsachiherman: good catch; fixed.

x/merkledb/benchmarks/benchmark.go
tsachiherman: hmm.. I think that it's ok to use the `low` as you suggested, although using the above would yield different and unique values ( i.e. [low..low+5k] ).

StephenButtolph:
This is no longer used for the benchmarks.

PR #3377, opened 2024-09-10
feat(GitHub): migrate configure-aws-credentials steps to IAM role assumption

adutchak:
## Why this should be merged
This improves the security and we can get rid of using static AWS credentials

## How this works
The configure-aws-credentials now assumes pre-configured IAM role

## How this was tested
We successfully used this approach for another repos

### DIFF
--- a/.github/workflows/build-linux-binaries.yml
+++ b/.github/workflows/build-linux-binaries.yml
@@ -32,8 +32,8 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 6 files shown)


StephenButtolph (dismissed):

StephenButtolph (commented):
Should we be updating these here as well? https://github.com/ava-labs/avalanchego/blob/8974ed8700b367c8b6bbbc325e5e14d7a606e482/.github/workflows/build-public-ami.yml#L50-L55

It does use a different access key tho.. so might be intended

adutchak:
> Should we be updating these here as well?
> 
> https://github.com/ava-labs/avalanchego/blob/8974ed8700b367c8b6bbbc325e5e14d7a606e482/.github/workflows/build-public-ami.yml#L50-L55
> 
> It does use a different access key tho.. so might be intended

Updated

StephenButtolph (approved):

PR #3379, opened 2024-09-10
Add Expiry Replay Protection

StephenButtolph:
## Why this should be merged

This PR introduces the ACP-77 validation expiry state. It supports maintaining a set of `ExpiryEntry`s and iterating over them from lower to higher timestamps.

- Iteration will be used to remove expired entries when advancing the chain time.
  - During this `DeleteExpiry` will be used
- When verifying a `RegisterValidatorTx` - `HasExpiry` will be used to very that the tx isn't being replayed.
- When executing a `RegisterValidatorTx` - `PutExpiry` will be used to prevent replaying in a following transaction.

## How this works

There is one interesting oddity. Specifically, I wanted the interface of the `state.Chain` to make no assumptions on the usage pattern. Even though we should never call `PutExpiry` when the `ExpiryEntry` already exists, this implementation handles that gracefully by deduplicating values between the `added` tree and parent/base `expiry` tree.

## How this was tested

- [X] Added unit tests for the new Diff logic
- [X] Added unit tests for the new State writing and loding logic


joshua-kim (approved):

vms/platformvm/state/expiry.go
```diff
@@ -22,6 +24,13 @@ var (
 	_ btree.LessFunc[ExpiryEntry] = ExpiryEntry.Less
 )
 
+type Expiry interface {
+	GetExpiryIterator() (iterator.Iterator[ExpiryEntry], error)
+	HasExpiry(ExpiryEntry) (bool, error)
+	PutExpiry(ExpiryEntry)
+	DeleteExpiry(ExpiryEntry)
```
StephenButtolph, line 31: These should be documented

vms/platformvm/state/diff_test.go
```diff
@@ -112,6 +113,154 @@ func TestDiffCurrentSupply(t *testing.T) {
...
+		otherExpiries := set.Of(test.initialExpiries...)
+		for _, op := range test.ops {
+			if op.put {
+				d.PutExpiry(op.entry)
+			} else {
+				d.DeleteExpiry(op.entry)
+			}
+			otherExpiries.Add(op.entry)
+		}
+		otherExpiries.Remove(test.expectedExpiries...)
```
aaronbuchwald, line 235: What is other expiries doing here? This test flow seems a bit odd to me

StephenButtolph: It is the expiries that shouldn't be in state

aaronbuchwald: Could you add a comment explaining this into the code? The test setup of:

1. add the initial expiries
2. add all of the expiry ops (both put/remove)
3. remove all of the expected values

does not seem clear just from the variable names that this results in the expiries not in state

StephenButtolph: 1. I changed the name from `otherExpiries` to `unexpectedExpiries`.
2. I removed `expectedExpiries` from the test struct and just calculated them explicitly in the test body.

aaronbuchwald (approved):

PR #3381, opened 2024-09-11
Implement iterator deduplicator

StephenButtolph:
## Why this should be merged

Factored out of #3379.

## How this works

While looking for the `Next` value - if the value was already seen, skip it. If the value wasn't seen, mark it as seen and stop iterating.

## How this was tested

- [X] Added unit test


utils/iterator/deduplicate.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package iterator
+
+import "github.com/ava-labs/avalanchego/utils/set"
+
+var _ Iterator[any] = (*deduplicator[any])(nil)
+
+type deduplicator[T comparable] struct {
```
yacovm, line 10:
```
type deduplicator[T comparable] struct {
	Iterator[T]
	seen set.Set[T]
}
```

Wondering if it makes sense to just embed the iterator and drop Value and Release

yacovm: I works locally for me: 

```
index b36c1e0a1..4ba769c12 100644
--- a/utils/iterator/deduplicate.go
+++ b/utils/iterator/deduplicate.go
@@ -8,7 +8,7 @@ import "github.com/ava-labs/avalanchego/utils/set"
 var _ Iterator[any] = (*deduplicator[any])(nil)
 
 type deduplicator[T comparable] struct {
-       it   Iterator[T]
+       Iterator[T]
        seen set.Set[T]
 }
 
@@ -16,13 +16,13 @@ type deduplicator[T comparable] struct {
 // been returned from [it].
 func Deduplicate[T comparable](it Iterator[T]) Iterator[T] {
        return &deduplicator[T]{
-               it: it,
+               Iterator: it,
        }
 }
 
 func (i *deduplicator[_]) Next() bool {
-       for i.it.Next() {
-               element := i.it.Value()
+       for i.Iterator.Next() {
+               element := i.Iterator.Value()
                if !i.seen.Contains(element) {
                        i.seen.Add(element)
                        return true
@@ -30,11 +30,3 @@ func (i *deduplicator[_]) Next() bool {
        }
        return false
 }
-
-func (i *deduplicator[T]) Value() T {
-       return i.it.Value()
-}
-
-func (i *deduplicator[_]) Release() {
-       i.it.Release()
-}
```

StephenButtolph: I removed this custom iterator implementation and reused `Filter`.

StephenButtolph: Embedding does work - but it has introduced subtle bugs in the past (because changing the signature of `Next` to `NewNext` or something wouldn't cause the code to fail to compile, but would cause the implementation to be wrong)...

This is why we use `Unsafe.*Server` rather than `Unimplemented.*Server` in our grpc services.

yacovm (approved):

utils/iterator/deduplicate.go
yacovm: Reminds me of `InsecureSkipVerify`

PR #3385, opened 2024-09-13
remove state commit on disconnect

ceyonur:
## Why this should be merged

Committing state adds an overhead on peer disconnects

## How this works

Removes state commit from disconnect. Block acceptance/Shutdown should be sufficient for writing disk to the state.

## How this was tested

existing UTs should cover this

### DIFF
--- a/vms/platformvm/vm.go
+++ b/vms/platformvm/vm.go
@@ -487,9 +487,6 @@ func (vm *VM) Disconnected(ctx context.Context, nodeID ids.NodeID) error {
 	if err := vm.uptimeManager.Disconnect(nodeID); err != nil {
 		return err
 	}
-	if err := vm.state.Commit(); err != nil {
-		return err
-	}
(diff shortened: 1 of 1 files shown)


vms/platformvm/vm.go
```diff
@@ -487,9 +487,6 @@ func (vm *VM) Disconnected(ctx context.Context, nodeID ids.NodeID) error {
 	if err := vm.uptimeManager.Disconnect(nodeID); err != nil {
 		return err
 	}
-	if err := vm.state.Commit(); err != nil {
-		return err
-	}
```
ceyonur, line 492: if we want to we can move this to `SetPreference` (or some periodic go routine like mempool pruning etc). I'd imagine block acceptance should be sufficient enough though.

ceyonur: yea good idea we could do that instead. but I slightly think that state should be commited with respect to chain progress.

PR #3388, opened 2024-09-13
ACP-77: Implement validator state

StephenButtolph:
## Why this should be merged

ACP-77 introduces a new type of validator: the `SubnetOnlyValidator`.

This PR introduces this new validator type to the P-chain state.

### `SubnetOnlyValidator`

For the new `SubnetOnlyValidator` type, the validator can either be `active` (`EndAccumulatedFee != 0`) or `inactive` (`EndAccumulatedFee == 0`).

`active` validators are handled similarly to other validator types. They can be sampled during consensus, can contribute to block production, can contribute to ICM (Warp) messages, and are held in memory to efficiently perform these operations. To limit the memory pressure introduced by `active` validators, the number of `active` validators is limited.

Unlike `active` validators, `inactive` validators are not sampled during consensus, can not contribute to block production, and can not contribute to ICM (Warp) messages. This allows `inactive` validators to be removed from memory and only kept on disk. Each `L1` will have an entry in their validator set representing the cumulative amount of inactive stake.

Removing a `SubnetOnlyValidator` is performed simply by setting their `Weight` to be `0`.

### State Modifications

The P-chain state is broken into 2 implementations:
1. The on-disk representation `state`
2. Potential changes to the on-disk representation `diff`

The typical flow for performing a change is:
1. Create a `diff` on top of either the `state` or another `diff`.
2. Apply the `diff` to the parent, either the `state` or another `diff`.
3. Atomically commit changes made to `state` to disk.

Contrary to prior validator implementations within the state package, this PR attempts to provide validation of the requested changes.

Returned errors from `PutSubnetOnlyValidator` should result in no changes to the state.

`PutSubnetOnlyValidator` succeeding should result in a valid state representation, regardless of the expected API behavior.

The only exception to this is documented with:
```golang
// If an SoV with the same validationID as a previously removed SoV is
// added, the behavior is undefined.
```
It is difficult to handle this case within the state without introducing non-deterministic error reporting because validationIDs are deleted from disk when the state is deleted. This means that the state code would need to handle all of the fields of an SoV to change on demand. However, that would require additional changes to the validator manager and other parts of the code (which make assumptions around `ValidationID` implying a static `subnetID + nodeID` pair).

The in-memory validator manager and historical validator set diffs are only updated during the atomic disk commitment changes.

## How this was tested

Added unit tests.

Added P-chain configs:
- [x] 	`l1-weights-cache-size`
- [x] 	`l1-inactive-validators-cache-size`
- [x] 	`l1-subnet-id-node-id-cache-size`


vms/platformvm/state/diff.go
```diff
@@ -184,6 +188,70 @@ func (d *diff) DeleteExpiry(entry ExpiryEntry) {
...
+	}
+
+	return d.sovDiff.getActiveSubnetOnlyValidatorsIterator(parentIterator), nil
+}
+
+func (d *diff) NumActiveSubnetOnlyValidators() int {
+	return d.parentActiveSOVs + d.sovDiff.numAddedActive
+}
+
+func (d *diff) NumSubnetOnlyValidators(subnetID ids.ID) (int, error) {
```
StephenButtolph, line 209: This should return the weight rather than the count.

vms/platformvm/state/state.go
```diff
@@ -1712,9 +1830,53 @@ func (s *state) loadPendingValidators() error {
...
+		if err := s.validators.AddStaker(sov.SubnetID, sov.NodeID, pk, validationID, sov.Weight); err != nil {
+			return err
+		}
+	}
+
+	// Load inactive weights
+	it := s.weightsDB.NewIterator()
+	defer it.Release()
+
+	for it.Next() {
```
StephenButtolph, line 1848: We should only be loading subnets with at least one active validator.

vms/platformvm/state/state.go
```diff
@@ -2004,6 +2172,303 @@ func (s *state) writeExpiry() error {
 	return nil
 }
 
+// TODO: Update validator sets
+// TODO: Add caching
+func (s *state) writeSubnetOnlyValidators(updateValidators bool, height uint64) error {
```
StephenButtolph, line 2177: Note: this only works if the nodeIDs written here are unrelated to legacy subnet validators. This is fairly reasonable because we're going to be introducing 32-byte nodeIDs for SoVs.

vms/platformvm/validators/manager.go
StephenButtolph, line 1: These changes need to be guarded by a network upgrade flag.

StephenButtolph, line 1: These changes need to be guarded by a network upgrade flag.

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
+	v := new(V)
+	m[k] = v
+	return v
+}
+
+func (s *state) makeSubnetOnlyValidatorHistoricalDiffs() (map[subnetIDNodeID]*validatorChanges, error) {
+	changes := make(map[subnetIDNodeID]*validatorChanges, len(s.sovDiff.modified))
+
+	// Perform deletions:
+	for validationID := range s.sovDiff.modified {
```
ceyonur, line 2446: how do we determine that this is deletion?

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
+			diff = getOrDefault(changes, subnetIDNodeID)
+		}
+
+		if err := diff.weightDiff.Add(true, priorSOV.Weight); err != nil {
+			return nil, err
+		}
+	}
+
+	// Perform additions:
+	for _, sov := range s.sovDiff.modified {
```
ceyonur, line 2476: seems like we're reiterating over deletions as well?

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
+	if v, ok := m[k]; ok {
+		return v
+	}
+
+	v := new(V)
+	m[k] = v
+	return v
+}
+
+func (s *state) makeSubnetOnlyValidatorHistoricalDiffs() (map[subnetIDNodeID]*validatorChanges, error) {
```
ceyonur, line 2442: in general I wonder if this function is necessary. everything here seems can be handled in `writeSubnetOnlyValidators`.

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
+			subnetIDNodeID.nodeID = ids.EmptyNodeID
+			diff = getOrDefault(changes, subnetIDNodeID)
+		}
+
+		if err := diff.weightDiff.Add(true, priorSOV.Weight); err != nil {
+			return nil, err
+		}
+	}
+
+	// Perform additions:
```
ceyonur, line 2475: nit: this also includes modifications right?

ceyonur (commented):
mostly questions

vms/platformvm/state/diff.go
```diff
@@ -184,6 +198,70 @@ func (d *diff) DeleteExpiry(entry ExpiryEntry) {
...
+	if !ok {
+		return 0, fmt.Errorf("%w: %s", ErrMissingParentState, d.parentID)
+	}
+
+	return parentState.WeightOfSubnetOnlyValidators(subnetID)
+}
+
+func (d *diff) GetSubnetOnlyValidator(validationID ids.ID) (SubnetOnlyValidator, error) {
+	if sov, modified := d.sovDiff.modified[validationID]; modified {
+		if sov.Weight == 0 {
```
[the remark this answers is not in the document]
StephenButtolph: `Weight=0` means that the SoV is being removed. Inactive validators do not have `Weight=0`.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -46,44 +99,63 @@ type SubnetOnlyValidator struct {
...
 	default:
-		return v.ValidationID.Compare(o.ValidationID) == -1
+		return v.ValidationID.Compare(o.ValidationID)
 	}
 }
 
-func getSubnetOnlyValidator(db database.KeyValueReader, validationID ids.ID) (*SubnetOnlyValidator, error) {
+// validateConstants returns true if the constants of this validator have not
+// been modified.
+func (v SubnetOnlyValidator) validateConstants(o SubnetOnlyValidator) bool {
```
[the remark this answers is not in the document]
StephenButtolph: I don't think equals would be a good name for this function. It does not return that the SoVs are equal, it verifies that fields that are not expected to be modified by the execution logic were not modified. So it does allow `Weight`, `MinNonce` and `EndAccumulatedFee` to be modified.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -94,3 +166,129 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr *SubnetOnlyValidator
...
+		modifiedHasNodeIDs:  make(map[subnetIDNodeID]bool),
+		active:              btree.NewG(defaultTreeDegree, SubnetOnlyValidator.Less),
+	}
+}
+
+func (d *subnetOnlyValidatorsDiff) getActiveSubnetOnlyValidatorsIterator(parentIterator iterator.Iterator[SubnetOnlyValidator]) iterator.Iterator[SubnetOnlyValidator] {
+	return iterator.Merge(
+		SubnetOnlyValidator.Less,
+		iterator.Filter(parentIterator, func(sov SubnetOnlyValidator) bool {
+			_, ok := d.modified[sov.ValidationID]
```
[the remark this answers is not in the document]
StephenButtolph: This:

1. Starts with the parent state's active validators
2. Removes any validators that were modified by this state diff
3. Adds any active validators that were modified by this state diff

Which gives us all of the active validators after this state diff.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -94,3 +166,129 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr *SubnetOnlyValidator
...
+
+	if prevSOV, ok := d.modified[sov.ValidationID]; ok {
+		prevSubnetIDNodeID := subnetIDNodeID{
+			subnetID: prevSOV.SubnetID,
+			nodeID:   prevSOV.NodeID,
+		}
+		d.modifiedHasNodeIDs[prevSubnetIDNodeID] = false
+		d.active.Delete(prevSOV)
+	}
+	d.modified[sov.ValidationID] = sov
```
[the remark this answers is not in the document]
StephenButtolph: It might be true that:
```golang
prevSubnetIDNodeID := subnetIDNodeID{
	subnetID: prevSOV.SubnetID,
	nodeID:   prevSOV.NodeID,
}
d.modifiedHasNodeIDs[prevSubnetIDNodeID] = false
```
is dead code... The `d.active.Delete(prevSOV)` is needed because the tree is going to delete the entry based on the `EndAccumulatedFee` which may be different between `prevSoV` and `sov`.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -46,44 +99,63 @@ type SubnetOnlyValidator struct {
...
+	return v.SubnetID == o.SubnetID &&
+		v.NodeID == o.NodeID &&
+		bytes.Equal(v.PublicKey, o.PublicKey) &&
+		bytes.Equal(v.RemainingBalanceOwner, o.RemainingBalanceOwner) &&
+		bytes.Equal(v.DeactivationOwner, o.DeactivationOwner) &&
+		v.StartTime == o.StartTime
+}
+
+func (v SubnetOnlyValidator) isActive() bool {
+	return v.Weight != 0 && v.EndAccumulatedFee != 0
```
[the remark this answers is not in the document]
StephenButtolph: `Weight = 0` -> Deleted
`EndAccumulatedFee = 0` -> Disabled

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -94,3 +166,129 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr *SubnetOnlyValidator
...
+		d.active.Delete(prevSOV)
+	}
+	d.modified[sov.ValidationID] = sov
+
+	subnetIDNodeID := subnetIDNodeID{
+		subnetID: sov.SubnetID,
+		nodeID:   sov.NodeID,
+	}
+	isDeleted := sov.Weight == 0
+	d.modifiedHasNodeIDs[subnetIDNodeID] = !isDeleted
```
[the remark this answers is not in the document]
StephenButtolph: `modifiedHasNodeIDs` can be `true` for disabled validators.

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
+		copy(subnetIDNodeIDKey, sov.SubnetID[:])
+		copy(subnetIDNodeIDKey[len(sov.SubnetID):], sov.NodeID[:])
+		if err := s.subnetIDNodeIDDB.Delete(subnetIDNodeIDKey); err != nil {
+			return err
+		}
 
-				if err = validatorDB.Put(staker.TxID[:], metadataBytes); err != nil {
-					return fmt.Errorf("failed to write current validator to list: %w", err)
-				}
+		if priorSOV.isActive() {
```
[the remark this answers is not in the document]
StephenButtolph: `priorSOV` is known not to be deleted at this point, so this is only going to be checking if it was active or inactive... The naming here can probably be improved

vms/platformvm/state/state.go
```diff
@@ -2004,149 +2204,337 @@ func (s *state) writeExpiry() error {
...
-					err := s.validatorPublicKeyDiffsDB.Put(
-						marshalDiffKey(constants.PrimaryNetworkID, height, nodeID),
-						nil,
-					)
-					if err != nil {
-						return err
-					}
-				}
+		// The next loops shouldn't consider this change.
+		delete(sovChanges, validationID)
```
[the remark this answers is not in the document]
StephenButtolph: I had tried that originally, but the code became too complex for me to be able to understand it clearly.

vms/platformvm/state/state.go
StephenButtolph: This function works by:
- Iterate over all the validators that were modified and delete their prior value from the validator set.
- Iterate over all the validators that were modified and add their new value to the validator set.

At the end we now have what the end diff is.

vms/platformvm/state/state.go
StephenButtolph: We immediately bail from this function if `sov.Weight == 0` (which is the deletion)

vms/platformvm/state/state.go
StephenButtolph: Again, that was my original pass, but the code was getting very complex.

vms/platformvm/state/state.go
StephenButtolph: We immediately bail from this function if `sov.Weight == 0` (which is the deletion)

vms/platformvm/state/state.go
```diff
@@ -1712,22 +1853,75 @@ func (s *state) loadPendingValidators() error {
...
+		if err != nil {
+			// This should never happen, as the total weight should always be at
+			// least the sum of the active weights.
+			return err
+		}
+		if inactiveWeight == 0 {
+			continue
+		}
+
+		if err := s.validators.AddStaker(subnetID, ids.EmptyNodeID, nil, ids.Empty, inactiveWeight); err != nil {
```
ceyonur, line 1897: shall we use a special ID (like 111, 001 etc) for inactive validators (rather than empty nodeID)

vms/platformvm/state/subnet_only_validator.go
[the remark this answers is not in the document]
ceyonur: oh ok I thought filter is there to eliminate inactives. Can we add few comments to clarify this?

vms/platformvm/state/subnet_only_validator.go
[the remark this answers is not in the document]
ceyonur: keeping this comment as a reminder for the deadcode. otherwise we can resolve this.

vms/platformvm/state/state.go
```diff
@@ -2365,19 +2924,19 @@ func (s *state) writeSubnetOwners() error {
 }
 
 func (s *state) writeSubnetManagers() error {
```
ceyonur, line 2926: nit: writeSubnetConversions

StephenButtolph: Addressed in the factored out PR #3481

vms/platformvm/state/state.go
StephenButtolph: Because only the `EmptyNodeID` is prevented from being issued to the network, I don't think we have any other options than using the `EmptyNodeID` here. Unless we modified the validator manager to know about inactive weight explicitly.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -128,3 +176,119 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr SubnetOnlyValidator)
...
+			return ErrConflictingSubnetOnlyValidator
+		}
+		if err != database.ErrNotFound {
+			return err
+		}
+
+		has, err := state.HasSubnetOnlyValidator(sov.SubnetID, sov.NodeID)
+		if err != nil {
+			return err
+		}
```
ceyonur, line 247: are we handling the `database.NotFound` err in the `HasSubnetOnlyValidator` (and returning false)?

vms/platformvm/state/state.go
ceyonur: not sure I follow, doesn't the validator manager already know abut total inactive weights (by `EmptyNodeID`)?

vms/platformvm/state/state.go
```diff
@@ -582,12 +583,12 @@ func New(
...
-	subnetManagerCache, err := metercacher.New[ids.ID, subnetConversion](
-		"subnet_manager_cache",
+	subnetConversionDB := prefixdb.New(SubnetConversionPrefix, baseDB)
+	subnetConversionCache, err := metercacher.New[ids.ID, SubnetConversion](
+		"subnet_conversion_cache",
 		metricsReg,
-		cache.NewSizedLRU[ids.ID, subnetConversion](execCfg.SubnetManagerCacheSize, func(_ ids.ID, f subnetConversion) int {
-			return 2*ids.IDLen + len(f.Addr)
+		cache.NewSizedLRU[ids.ID, SubnetConversion](execCfg.SubnetConversionCacheSize, func(_ ids.ID, c SubnetConversion) int {
+			return 3*ids.IDLen + len(c.Addr)
```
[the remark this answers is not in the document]
StephenButtolph: This was changed because previously we didn't persist the `ConversionID`.

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: This aligns with `Get` and `Has` behave differently on the database interface. `Get` returns `ErrNotFound` if the entry is missing. `Has` explicitly returns `true` if the entry exists and `false` if the entry does not exist.

`Has` never returns `ErrNotFound`.

vms/platformvm/state/state.go
StephenButtolph: The validator manager does not treat `ids.EmptyNodeID` differently than other `NodeID`s.

ceyonur: My only concern and question is would it be safe against the proposer vm's special casing of emptyNodeID? I guess we still need to explicitly handle it  in proposer VM even if we decide to use another ID

StephenButtolph: > My only concern and question is would it be safe against the proposer vm's special casing of emptyNodeID

I have a local branch working to remove that special casing. Not quite ready to push up yet though

vms/platformvm/state/state.go
StephenButtolph: Added a TODO

vms/platformvm/state/state.go
```diff
@@ -2315,92 +2382,41 @@ func (s *state) updateValidatorManager(updateValidators bool) error {
...
+					err = s.validators.RemoveWeight(sov.SubnetID, nodeID, priorSOV.Weight-sov.Weight)
+				}
 			} else {
-				err = s.validators.AddWeight(sov.SubnetID, ids.EmptyNodeID, sov.Weight)
+				// This validator's active status is changing.
+				err = errors.Join(
+					s.validators.RemoveWeight(sov.SubnetID, priorSOV.effectiveNodeID(), priorSOV.Weight),
+					addSoVToValidatorManager(s.validators, sov),
+				)
 			}
```
ceyonur, line 2416: noicer

vms/platformvm/state/state.go
```diff
@@ -2467,61 +2483,40 @@ func (s *state) calculateValidatorDiffs() (map[subnetIDNodeID]*validatorDiff, er
...
-		)
-		if priorSOV.isActive() {
-			subnetIDNodeID.nodeID = priorSOV.NodeID
-			diff = getOrDefault(changes, subnetIDNodeID)
-			diff.prevPublicKey = priorSOV.PublicKey
-		} else {
-			subnetIDNodeID.nodeID = ids.EmptyNodeID
-			diff = getOrDefault(changes, subnetIDNodeID)
+			diff := getOrDefault(changes, subnetIDNodeID)
+			if err := diff.weightDiff.Add(true, priorSOV.Weight); err != nil {
```
ceyonur, line 2496: not directly related with this PR but can we use `weightDiff.Remove(amount)` rather then `weightDiff.Add(true, amount)`?

vms/platformvm/state/state.go
```diff
@@ -2727,82 +2722,82 @@ func writePendingDiff(
...
+	}
+
+	// The SoV diff application is split into two loops to ensure that all
+	// deletions to the subnetIDNodeIDDB happen prior to any additions.
+	// Otherwise replacing an SoV by deleting it and then re-adding it with a
+	// different validationID could result in an inconsistent state.
+	for validationID, sov := range s.sovDiff.modified {
+		// Delete the prior validator if it exists
+		var err error
+		if s.activeSOVs.delete(validationID) {
```
ceyonur, line 2747: can we use get first and then delete  from activeSovs if it's indeed active or delete from inactive? I think `delete` does a bit too much work here and also wrapping `get`.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -292,3 +326,68 @@ func (d *subnetOnlyValidatorsDiff) putSubnetOnlyValidator(state Chain, sov Subne
...
+func (a *activeSubnetOnlyValidators) put(sov SubnetOnlyValidator) {
+	a.lookup[sov.ValidationID] = sov
+	a.tree.ReplaceOrInsert(sov)
+}
+
+func (a *activeSubnetOnlyValidators) delete(validationID ids.ID) bool {
+	sov, ok := a.lookup[validationID]
+	if !ok {
+		return false
+	}
```
ceyonur, line 356: If we want to keep it as a safeguard for deletions it's fine but IMO `delete` should not be used instead of `get`.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -292,3 +326,68 @@ func (d *subnetOnlyValidatorsDiff) putSubnetOnlyValidator(state Chain, sov Subne
...
+
+func (a *activeSubnetOnlyValidators) len() int {
+	return len(a.lookup)
+}
+
+func (a *activeSubnetOnlyValidators) newIterator() iterator.Iterator[SubnetOnlyValidator] {
+	return iterator.FromTree(a.tree)
+}
+
+func (a *activeSubnetOnlyValidators) addStakers(vdrs validators.Manager) error {
```
ceyonur, line 371: nit: addStakersToValidatorManager?

vms/platformvm/state/state.go
```diff
@@ -330,13 +339,15 @@ type state struct {
...
 	expiryDB   database.Database
 
-	activeSOVLookup        map[ids.ID]SubnetOnlyValidator
-	activeSOVs             *btree.BTreeG[SubnetOnlyValidator]
+	activeSOVs             *activeSubnetOnlyValidators
 	sovDiff                *subnetOnlyValidatorsDiff
 	subnetOnlyValidatorsDB database.Database
+	weightsCache           cache.Cacher[ids.ID, uint64] // subnetID -> total SoV weight
 	weightsDB              database.Database
+	subnetIDNodeIDCache    cache.Cacher[subnetIDNodeID, bool] // subnetID+nodeID -> is validator
```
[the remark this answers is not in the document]
StephenButtolph: I'm not sure that it actually helps for our planned usage of the `validationID`. We would need to be performing `Get`s on the `subnetIDNodeIDDB`, but I think right now we are planning on iterating over a `subnetID` prefix. The iteration provides the values along with the keys, so I don't think we will actually ever be able to use the cached `validationID`s.

If it would be useful later, then we can definitely change it then... But for now I felt like it didn't make sense.

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: Unless `delete` took in the full `SubnetOnlyValidator` struct, we need this check. This is because deleting from the `tree` requires the full struct, not just the ID.

vms/platformvm/state/state.go
```diff
@@ -2727,82 +2722,82 @@ func writePendingDiff(
...
+	for validationID, sov := range s.sovDiff.modified {
+		// Delete the prior validator if it exists
+		var err error
+		if s.activeSOVs.delete(validationID) {
 			err = deleteSubnetOnlyValidator(s.activeDB, validationID)
 		} else {
-			// It is technically possible for the validator not to exist on disk
-			// here, but that's fine as deleting an entry that doesn't exist is
-			// a noop.
+			s.inactiveCache.Put(validationID, maybe.Nothing[SubnetOnlyValidator]())
```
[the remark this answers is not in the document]
StephenButtolph: sgtm

vms/platformvm/state/state.go
StephenButtolph: I actually feel like it's pretty normal for `delete` to return whether or not it deleted something. For example, this is the behavior of `tree.Delete`

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -12,17 +12,63 @@ import (
...
+	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/vms/platformvm/block"
 )
 
 var (
 	_ btree.LessFunc[SubnetOnlyValidator] = SubnetOnlyValidator.Less
 	_ utils.Sortable[SubnetOnlyValidator] = SubnetOnlyValidator{}
 
-	ErrMutatedSubnetOnlyValidator = errors.New("subnet only validator contains mutated constant fields")
+	ErrMutatedSubnetOnlyValidator     = errors.New("subnet only validator contains mutated constant fields")
```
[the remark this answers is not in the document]
StephenButtolph: I don't think this is actually a huge change, updated the few instances I found

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -128,3 +210,184 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr SubnetOnlyValidator)
...
+func (d *subnetOnlyValidatorsDiff) hasSubnetOnlyValidator(subnetID ids.ID, nodeID ids.NodeID) (bool, bool) {
+	subnetIDNodeID := subnetIDNodeID{
+		subnetID: subnetID,
+		nodeID:   nodeID,
+	}
+	has, modified := d.modifiedHasNodeIDs[subnetIDNodeID]
+	return has, modified
+}
+
+func (d *subnetOnlyValidatorsDiff) putSubnetOnlyValidator(state Chain, sov SubnetOnlyValidator) error {
```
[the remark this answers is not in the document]
StephenButtolph: `putSubnetOnlyValidator` is essentially directly called by `state.PutSubnetOnlyValidator` and `diff.PutSubnetOnlyValidator`. So I suppose the question is why those functions are tested rather than this one directly?

It feels much more complex to me to need to implement a test-double for `Chain` rather than using either `Diff` or `State`, which is what the existing tests are already doing.

(I also generally prefer testing exported functions over un-exported functions, as there is a more clear separation between the caller and the callee).

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -128,3 +210,184 @@ func putSubnetOnlyValidator(db database.KeyValueWriter, vdr SubnetOnlyValidator)
...
+		if err := vdrs.AddStaker(sov.SubnetID, sov.NodeID, pk, validationID, sov.Weight); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func addSoVToValidatorManager(vdrs validators.Manager, sov SubnetOnlyValidator) error {
+	nodeID := sov.effectiveNodeID()
+	if vdrs.GetWeight(sov.SubnetID, nodeID) != 0 {
```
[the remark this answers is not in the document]
StephenButtolph: If the weight of a staker goes to 0 in the validator manager, it is removed.

If `AddStaker`, `AddWeight`, or `RemoveWeight` are provided `weight=0` an error is returned.

vms/platformvm/state/state.go
StephenButtolph: Refactored into #3514

vms/platformvm/state/state.go
```diff
@@ -658,13 +715,15 @@ func New(
...
+		activeSOVs:             newActiveSubnetOnlyValidators(),
 		sovDiff:                newSubnetOnlyValidatorsDiff(),
 		subnetOnlyValidatorsDB: subnetOnlyValidatorsDB,
+		weightsCache:           weightsCache,
 		weightsDB:              prefixdb.New(WeightsPrefix, subnetOnlyValidatorsDB),
+		subnetIDNodeIDCache:    subnetIDNodeIDCache,
 		subnetIDNodeIDDB:       prefixdb.New(SubnetIDNodeIDPrefix, subnetOnlyValidatorsDB),
 		activeDB:               prefixdb.New(ActivePrefix, subnetOnlyValidatorsDB),
+		inactiveCache:          inactiveSOVsCache,
 		inactiveDB:             prefixdb.New(InactivePrefix, subnetOnlyValidatorsDB),
```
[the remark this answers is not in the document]
StephenButtolph: Yeah I mean... The state package is unreal at this point

vms/platformvm/state/diff.go
```diff
@@ -504,6 +572,26 @@ func (d *diff) Apply(baseState Chain) error {
 			baseState.DeleteExpiry(entry)
 		}
 	}
+	// Ensure that all sov deletions happen before any sov additions. This
+	// ensures that a subnetID+nodeID pair that was deleted and then re-added in
+	// a single diff can't get reordered into the addition happening first;
+	// which would return an error.
```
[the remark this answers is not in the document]
StephenButtolph: > If a SoV is modified and then deleted in the same diff, is it still safe to put the deletion first and modification second here?

If there are multiple modifications to the same `ValidationID` in a single diff, then only the last modification is tracked.

So, if an SoV (with the same `ValidationID`) is `Added` ->? `Modified` ->* `Deleted`, then the map will only track that the SoV was `Deleted`.

It is typically guaranteed that only one SoV can exist for a given `subnetID` + `nodeID`. However, this is not guaranteed with deletions. Specifically, there could be any number of deleted SoVs with the same `subnetID` + `nodeID` along with at most one SoV with the same `subnetID` + `nodeID` that isn't deleted.

In this case, the, not deleted, SoV with the same `subnetID` + `nodeID` should be written last.

vms/platformvm/state/state.go
ceyonur: I just found it a bit weird that we try to immediately delete it from activeSOVs before we check if it's there. But I think it's mostly a preference so should be fine.

vms/platformvm/state/state.go
```diff
@@ -743,6 +831,71 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+func (s *state) WeightOfSubnetOnlyValidators(subnetID ids.ID) (uint64, error) {
+	if weight, modified := s.sovDiff.modifiedTotalWeight[subnetID]; modified {
+		return weight, nil
+	}
+
+	if weight, ok := s.weightsCache.Get(subnetID); ok {
+		return weight, nil
+	}
+
+	return database.WithDefault(database.GetUInt64, s.weightsDB, subnetID[:], 0)
```
StephenButtolph, line 853: Should cache result

vms/platformvm/state/state.go
```diff
@@ -743,6 +831,71 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+	subnetIDNodeID := subnetIDNodeID{
+		subnetID: subnetID,
+		nodeID:   nodeID,
+	}
+	if has, ok := s.subnetIDNodeIDCache.Get(subnetIDNodeID); ok {
+		return has, nil
+	}
+
+	key := subnetIDNodeID.Marshal()
+	return s.subnetIDNodeIDDB.Has(key)
```
StephenButtolph, line 892: Should cache result

vms/platformvm/state/state.go
```diff
@@ -536,9 +565,57 @@ func New(
...
+		metricsReg,
+		cache.NewSizedLRU[ids.ID, maybe.Maybe[SubnetOnlyValidator]](
+			execCfg.L1InactiveValidatorsCacheSize,
+			func(_ ids.ID, maybeSOV maybe.Maybe[SubnetOnlyValidator]) int {
+				const (
+					sovOverhead      = ids.IDLen + ids.NodeIDLen + 4*wrappers.LongLen + 3*constants.PointerOverhead
+					maybeSOVOverhead = wrappers.BoolLen + sovOverhead
+					entryOverhead    = ids.IDLen + maybeSOVOverhead
+				)
+				if maybeSOV.IsNothing() {
```
yacovm, line 595: under what circumstance is it nothing?

vms/platformvm/state/diff.go
```diff
@@ -194,6 +198,70 @@ func (d *diff) DeleteExpiry(entry ExpiryEntry) {
...
+	if has, modified := d.sovDiff.hasSubnetOnlyValidator(subnetID, nodeID); modified {
+		return has, nil
+	}
+
+	parentState, ok := d.stateVersions.GetState(d.parentID)
+	if !ok {
+		return false, fmt.Errorf("%w: %s", ErrMissingParentState, d.parentID)
+	}
+
+	return parentState.HasSubnetOnlyValidator(subnetID, nodeID)
```
yacovm, line 258: this might seem as a weird question, but - how deep is the expected recursion here?

StephenButtolph: The recursion ends at the lastAcceptedState and starts at the most recently verified block, so typically the recursion is very shallow.

vms/platformvm/state/state.go
StephenButtolph: This cache stores hits as `maybe.Some` and misses as `maybe.Nothing`. So, if someone performs a `delete` or a `get` which results in not found, `maybe.Nothing` would be cached.

vms/platformvm/state/diff.go
```diff
@@ -194,6 +198,70 @@ func (d *diff) DeleteExpiry(entry ExpiryEntry) {
...
+func (d *diff) NumActiveSubnetOnlyValidators() int {
+	return d.parentActiveSOVs + d.sovDiff.numAddedActive
+}
+
+func (d *diff) WeightOfSubnetOnlyValidators(subnetID ids.ID) (uint64, error) {
+	if weight, modified := d.sovDiff.modifiedTotalWeight[subnetID]; modified {
+		return weight, nil
+	}
+
+	parentState, ok := d.stateVersions.GetState(d.parentID)
```
yacovm, line 224: can you explain the high level idea of doing this backwards traversal in search of the subnet, both here and in the next function? 

We basically don't have a full snapshot of the validators in each state? Why is that?

vms/platformvm/state/state.go
```diff
@@ -743,6 +831,83 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
 }
 
+func (s *state) GetActiveSubnetOnlyValidatorsIterator() (iterator.Iterator[SubnetOnlyValidator], error) {
+	return s.sovDiff.getActiveSubnetOnlyValidatorsIterator(
+		s.activeSOVs.newIterator(),
+	), nil
+}
+
+func (s *state) NumActiveSubnetOnlyValidators() int {
+	return s.activeSOVs.len() + s.sovDiff.numAddedActive
```
michaelkaplan13, line 841: Nit: Would `numAddedActive` make more sense to name `numNetAddedActive` or `netAddedActive`? Implementation looks correct, but wasn't sure based on the name only

vms/platformvm/state/state.go
```diff
@@ -743,6 +831,83 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+		}
+		return sov, nil
+	}
+
+	return s.getPersistedSubnetOnlyValidator(validationID)
+}
+
+// getPersistedSubnetOnlyValidator returns the currently persisted
+// SubnetOnlyValidator with the given validationID. It is guaranteed that any
+// returned validator is either active or inactive.
```
michaelkaplan13, line 875: (No action needed) Aren't all validators always either active or inactive? Not sure I understand the purpose of the comment

vms/platformvm/state/diff.go
```diff
@@ -504,6 +572,26 @@ func (d *diff) Apply(baseState Chain) error {
 			baseState.DeleteExpiry(entry)
 		}
 	}
+	// Ensure that all sov deletions happen before any sov additions. This
+	// ensures that a subnetID+nodeID pair that was deleted and then re-added in
+	// a single diff can't get reordered into the addition happening first;
+	// which would return an error.
+	for _, sov := range d.sovDiff.modified {
```
yacovm, line 579: Playing devil's advocate here, we're iterating here over a map, and calling `baseState.PutSubnetOnlyValidator` which may return with different errors.  This means that across different nodes, we may fail differently, no? 

Is that a concern?

vms/platformvm/state/state.go
```diff
@@ -1501,6 +1667,33 @@ func (s *state) loadExpiry() error {
 	return nil
 }
 
+func (s *state) loadActiveSubnetOnlyValidators() error {
+	it := s.activeDB.NewIterator()
+	defer it.Release()
+
+	for it.Next() {
```
michaelkaplan13, line 1674: Super nit: Inconsistent newline between `defer it.Release()` and the for loop here compared to `loadCurrentValidators()` below.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -22,9 +26,51 @@ var (
 	_ btree.LessFunc[SubnetOnlyValidator] = SubnetOnlyValidator.Less
 	_ utils.Sortable[SubnetOnlyValidator] = SubnetOnlyValidator{}
 
-	ErrMutatedSubnetOnlyValidator = errors.New("subnet only validator contains mutated constant fields")
+	ErrMutatedSubnetOnlyValidator     = errors.New("subnet-only validator contains mutated constant fields")
+	ErrConflictingSubnetOnlyValidator = errors.New("subnet-only validator contains conflicting subnetID + nodeID pair")
+	ErrDuplicateSubnetOnlyValidator   = errors.New("subnet-only validator contains duplicate subnetID + nodeID pair")
 )
 
+type SubnetOnlyValidators interface {
```
michaelkaplan13, line 34: Can we add compile time checks that `diff` and `state` implement `SubnetOnlyValidators`?

vms/platformvm/state/state.go
StephenButtolph: This was supposed to signify that no deleted SoVs will be returned from this function (`isDeleted()` will always return false). Specifically, this is to ensure that `!isActive()` means `inactive` (and not `deleted`)

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: 1. There already exists a compile time check that `diff` implements `Diff`. `Diff` extends `Chain` which extends `SubnetOnlyValidators`.
2. There already exists a compile time check that `state` implements `State`. `Chain` extends `Chain` which extends `SubnetOnlyValidators`. 

So the checks _do_ exist... But could be more explicit. Do you want them added anyway?

vms/platformvm/state/state.go
```diff
@@ -2132,6 +2379,65 @@ func (s *state) updateValidatorManager(updateValidators bool) error {
...
+		case nil:
+			// Modifying an existing validator
+			if priorSOV.isActive() == sov.isActive() {
+				// This validator's active status isn't changing. This means
+				// the effectiveNodeIDs are equal.
+				nodeID := sov.effectiveNodeID()
+				if priorSOV.Weight < sov.Weight {
+					err = s.validators.AddWeight(sov.SubnetID, nodeID, sov.Weight-priorSOV.Weight)
+				} else if priorSOV.Weight > sov.Weight {
+					err = s.validators.RemoveWeight(sov.SubnetID, nodeID, priorSOV.Weight-sov.Weight)
```
yacovm, line 2422: is `priorSOV.Weight > sov.Weight` legal? what does it mean?

vms/platformvm/state/diff.go
StephenButtolph: It is generally unexpected for `Apply` to ever error.

`Apply` is called during `Accept`. If `Apply` were to return an error, that error would be propagated through `Accept` into the consensus engine (causing a FATAL).

vms/platformvm/state/state.go
StephenButtolph: Yeah, `net` seems like a good indicator of a possibly negative value.

vms/platformvm/state/diff.go
StephenButtolph: Performing a full copy of the validator states for each diff would be pretty expensive. The number of changes per block is typically small, but the number total entries being tracked can be quite large.

vms/platformvm/state/state.go
StephenButtolph: If the `priorSOV.Weight` is greater than `sov.Weight`, then the weight of the validator is being decreased by `priorSOV.Weight - sov.Weight` units

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -22,9 +26,51 @@ var (
...
+	// subnet-only validators.
+	NumActiveSubnetOnlyValidators() int
+
+	// WeightOfSubnetOnlyValidators returns the total active and inactive weight
+	// of subnet-only validators on [subnetID].
+	WeightOfSubnetOnlyValidators(subnetID ids.ID) (uint64, error)
+
+	// GetSubnetOnlyValidator returns the validator with [validationID] if it
+	// exists. If the validator does not exist, [err] will equal
+	// [database.ErrNotFound].
```
cam-schultz, line 49: It's a bit odd to call out the error type here when that's not enforceable by the interface

vms/platformvm/state/state.go
```diff
@@ -2230,6 +2571,16 @@ func (s *state) writeValidatorDiffs(height uint64) error {
 	return nil
 }
 
+func getOrDefault[K comparable, V any](m map[K]*V, k K) *V {
```
cam-schultz, line 2574: A comment explaining that the default value is written to the map would be helpful. Or consider renaming to `getOrAddDefault`

vms/platformvm/state/state.go
```diff
@@ -2384,6 +2735,92 @@ func writePendingDiff(
 	return nil
 }
 
+func (s *state) writeSubnetOnlyValidators() error {
+	// Write modified weights
+	for subnetID, weight := range s.sovDiff.modifiedTotalWeight {
+		var err error
+		if weight == 0 {
+			err = s.weightsDB.Delete(subnetID[:])
```
cam-schultz, line 2743: What's the expected behavior of `WeightOfSubnetOnlyValidators` after the entry is deleted from the DB and the cache entry expires? If it's to error, perhaps we should not write to the cache so that `WeightOfSubnetOnlyValidators` deterministically errors for any subnets with weight 0.

vms/platformvm/state/diff.go
```diff
@@ -35,15 +35,17 @@ type diff struct {
...
 
-	timestamp   time.Time
-	feeState    gas.State
-	sovExcess   gas.Gas
-	accruedFees uint64
+	timestamp        time.Time
+	feeState         gas.State
+	sovExcess        gas.Gas
+	accruedFees      uint64
+	parentActiveSOVs int
```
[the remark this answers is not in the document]
richardpringle: Or `parentActiveSOVCount` because a "count" is more descriptive than a "number" 😉

darioush (approved):

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -166,3 +248,184 @@ func deleteSubnetOnlyValidator(
...
+
+func (a *activeSubnetOnlyValidators) newIterator() iterator.Iterator[SubnetOnlyValidator] {
+	return iterator.FromTree(a.tree)
+}
+
+func (a *activeSubnetOnlyValidators) addStakersToValidatorManager(vdrs validators.Manager) error {
+	for validationID, sov := range a.lookup {
+		pk := bls.PublicKeyFromValidUncompressedBytes(sov.PublicKey)
+		if err := vdrs.AddStaker(sov.SubnetID, sov.NodeID, pk, validationID, sov.Weight); err != nil {
+			return err
```
darioush, line 413: Is it possible that sov.Weight be 0 here and if so should we add a test?

vms/platformvm/state/state.go
yacovm: no but i asked what it means when they're equal, I edited shortly after, probably github didn't refresh my comment

vms/platformvm/state/subnet_only_validator.go
michaelkaplan13: Nope, as long as it's there 🫡

michaelkaplan13: thanks for clarifying

vms/platformvm/state/state.go
StephenButtolph: Ah. It is possible for them to be equal. In that case the weight does not need to be modified.

vms/platformvm/state/diff.go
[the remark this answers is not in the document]
StephenButtolph: I ended up going with `parentNumActiveSOVs`.

Because it aligns most with the actual source of this: `parentState.NumActiveSubnetOnlyValidators()`

vms/platformvm/state/diff.go
```diff
@@ -194,6 +198,70 @@ func (d *diff) DeleteExpiry(entry ExpiryEntry) {
...
+
+	parentState, ok := d.stateVersions.GetState(d.parentID)
+	if !ok {
+		return 0, fmt.Errorf("%w: %s", ErrMissingParentState, d.parentID)
+	}
+
+	return parentState.WeightOfSubnetOnlyValidators(subnetID)
+}
+
+func (d *diff) GetSubnetOnlyValidator(validationID ids.ID) (SubnetOnlyValidator, error) {
```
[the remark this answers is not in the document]
StephenButtolph: Diffs are created for a few reasons. But the reason the diffs are designed this way is because each diff (can) represent the changes that are performed by a block. When a block is verified a diff is created. When a block is accepted a diff is applied to the disk state.

Diffs are updated in-place during the execution (verification) of a block - but we wouldn't want a child block to modify the state that will be committed when the parent block is accepted (as the child could be rejected for a different child)

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: I don't think this is particularly unusual... This is done in the standard lib as well. For an example, see `SetDeadline` here: https://pkg.go.dev/net#Conn

vms/platformvm/state/state.go
StephenButtolph: did both

vms/platformvm/state/state.go
StephenButtolph: `WeightOfSubnetOnlyValidators` calls `database.WithDefault(database.GetUInt64, s.weightsDB, subnetID[:], 0)` which will return `0` if the value is not on disk. So, `WeightOfSubnetOnlyValidators` deterministically returns `0` for any subnets with weight `0`.

vms/platformvm/state/subnet_only_validator.go
```diff
@@ -166,3 +248,184 @@ func deleteSubnetOnlyValidator(
...
+		}
+	}
+	return nil
+}
+
+func addSoVToValidatorManager(vdrs validators.Manager, sov SubnetOnlyValidator) error {
+	nodeID := sov.effectiveNodeID()
+	if vdrs.GetWeight(sov.SubnetID, nodeID) != 0 {
+		return vdrs.AddWeight(sov.SubnetID, nodeID, sov.Weight)
```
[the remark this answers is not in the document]
StephenButtolph: This can occur if the `effectiveNodeID` is the `ids.EmptyNodeID` (it does not always happen if that is the case, but it can.

This function is only called if there is no prior version of this SoV.

I added a test that fails if the weight is not increased as expected.

vms/platformvm/state/state.go
```diff
@@ -1501,6 +1667,32 @@ func (s *state) loadExpiry() error {
...
+			return fmt.Errorf("failed to unmarshal ValidationID during load: %w", err)
+		}
+
+		var (
+			value = it.Value()
+			sov   = SubnetOnlyValidator{
+				ValidationID: validationID,
+			}
+		)
+		if _, err := block.GenesisCodec.Unmarshal(value, &sov); err != nil {
```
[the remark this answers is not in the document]
StephenButtolph: This type is not registered. The code only requires types to be registered that are used as values for interfaces (there are no interfaces being passed around here)

vms/platformvm/state/state.go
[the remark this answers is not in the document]
StephenButtolph: Used cam's suggestion of `getOrAddDefault` and added a comment.

vms/platformvm/state/subnet_only_validator.go
StephenButtolph: `activeSubnetOnlyValidators` should only ever contain `sov`s that return `.isActive` as true. Is should be impossible for `sov.Weight == 0` here.

michaelkaplan13 (approved):

PR #3394, opened 2024-09-17
Implement ACP-118 Aggregator

joshua-kim:
## Why this should be merged

Implements  `p2p` client + server logic for signature request handling as described in `acp-118` ([ref](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/118-warp-signature-request)).

## How this works

Client:
* Introduces an aggregator type which will make signature requests in batches block until a caller-provided threshold of stake signs the provided warp message
* Introduces a server handler that servers aggregation requests

## How this was tested
* Added unit tests


network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,229 @@
...
+
+type Validator struct {
+	NodeID    ids.NodeID
+	PublicKey *bls.PublicKey
+	Weight    uint64
+}
+
+type indexedValidator struct {
+	Validator
+	I int
```
aaronbuchwald, line 47: nit: `Index`

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,229 @@
...
+			s.log.Debug(
+				"dropping response",
+				zap.Stringer("nodeID", nodeID),
+				zap.String("reason", "public key failed verification"),
+			)
+			return
+		}
+
+		lock.Lock()
+		signerBitSet.Add(validator.I)
```
aaronbuchwald, line 176: nit: `defer lock.Unlock()` here rather than prior to each return and end of the function

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// AggregateSignatures blocks until stakeWeightThreshold of validators signs the
+// provided message.
+func (s *SignatureAggregator) AggregateSignatures(
+	ctx context.Context,
+	message *warp.UnsignedMessage,
+	justification []byte,
+	validators []Validator,
+	stakeWeightThreshold uint64,
```
aaronbuchwald, line 77: Would it make more sense to pass in the required numerator/denominator values since the caller is more likely to be checking that the returned signature meets that threshold rather than a specific stake weight?

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,229 @@
...
+		err error,
+	) {
+		// We are guaranteed a response from a node in the validator set
+		validator := nodeIDsToValidator[nodeID]
+
+		defer func() {
+			lock.Lock()
+			attemptedStakeWeight += validator.Weight
+			remainingStakeWeight := totalStakeWeight - attemptedStakeWeight
+			failed := remainingStakeWeight < stakeWeightThreshold
```
aaronbuchwald, line 126: Should this be:

```suggestion
			failed := remainingStakeWeight + aggregatedStakeWeight < stakeWeightThreshold
```

joshua-kim: This relied on a hacky assumption that this value isn't read from the channel on the happy-path when we're not failing. I refactored this to make this less weird

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,294 @@
...
+// validators are requested to be aggregated into a warp message.
+func (s *SignatureAggregator) AggregateSignatures(
+	ctx context.Context,
+	message *warp.UnsignedMessage,
+	justification []byte,
+	validators []Validator,
+	quorumNumMin uint64,
+	quorumDenMin uint64,
+	quorumNumMax uint64,
+	quorumDenMax uint64,
```
joshua-kim, line 80: So the inclusion of these lower + upper bound parameters was to support DSMR's use-case where we want to meet a minimum threshold of signatures, but want to ideally replicate to more if possible. Another alternative to having these parameters here is to have a signature like
```
AggregateSignatures(ctx context.Context, msg warp.Message, justification []byte, validators []Validator, quorumNum, quorumDen uint64) (warp.Message, error)
```

If we pass in a signed message we can request signatures that aren't already in the bitset (will require a refactor to expose the concrete type of the `BitSetSignature` type on `warp.Message`). That way the caller could cancel the context on a timeout and decide if it wants to continue aggregation or not.
```

File: network/p2p/acp118/aggregator.go [hunk 6]
```diff
@@ -0,0 +1,294 @@
...
+) (*warp.Message, error) {
```
joshua-kim, line 81: So the inclusion of these lower + upper bound parameters was to support DSMR's use-case where we want to meet a minimum threshold of signatures, but want to ideally replicate to more if possible. Another alternative to having these parameters here is to have a signature like so:

```suggestion
func (s *SignatureAggregator) AggregateSignatures(
	ctx context.Context,
	message *warp.Message,
	justification []byte,
	validators []Validator,
	quorumNum uint64,
	quorumDen uint64,
) (*warp.Message, error) {
```

If we pass in a signed message we can request signatures that aren't already in the bitset (will require a refactor to expose the concrete type of the `BitSetSignature` type on `warp.Message`). That way the caller could cancel the context on a timeout and decide if it wants to continue aggregation or not.

cam-schultz: The [signature-aggregator](https://github.com/ava-labs/awm-relayer/tree/main/signature-aggregator) service implementation caches signatures per Warp message ID, indexed by the signer's public keys, allowing us to reconstruct signed messages when the canonical validator set changes in certain cases, without having to request them again. The proposed alternative would work well for that use case, in addition to allowing the caller to cancel as already described.

cam-schultz: Related - for the first call to `AggregateSignatures`, is the idea to pass a `warp.Message` with an default initialized (or nil) `BitSetSignature`?

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,294 @@
...
+
+	totalStakeWeight := uint64(0)
+	nodeIDsToValidator := make(map[ids.NodeID]indexedValidator)
+	for i, v := range validators {
+		totalStakeWeight, err = math.Add[uint64](totalStakeWeight, v.Weight)
+		if err != nil {
+			return nil, err
+		}
+
+		// Check that the validator set provided is valid
```
cam-schultz, line 103: A comment explaining that node IDs must be unique, but bls pubkeys may not be would be helpful.

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,294 @@
...
+		Signers:   a.signerBitSet.Bytes(),
+		Signature: [bls.SignatureLen]byte{},
+	}
+
+	copy(bitSetSignature.Signature[:], bls.SignatureToBytes(aggregateSignature))
+	return bitSetSignature, nil
+}
+
+func (a *aggregationJob) SendRequest(ctx context.Context) error {
+	if len(a.validators) == 0 {
```
cam-schultz, line 203: Should we add a check that `index < len(a.validators)` as well? It looks like it's possible to call `SendRequest` after all validators have been queried in the case where `s.maxPending == len(a.validators)`

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,294 @@
...
+	bitSetSignature := &warp.BitSetSignature{
+		Signers:   a.signerBitSet.Bytes(),
+		Signature: [bls.SignatureLen]byte{},
+	}
+
+	copy(bitSetSignature.Signature[:], bls.SignatureToBytes(aggregateSignature))
+	return bitSetSignature, nil
+}
+
+func (a *aggregationJob) SendRequest(ctx context.Context) error {
```
cam-schultz, line 202: How are timeouts handled?

network/p2p/acp118/aggregator.go
joshua-kim: Yeah, exaxctly. The normal usage would be to use a warp message with an empty signature.

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize message")
+	}
+
+	return warp.NewMessage(unsignedMessage, bitSetSignature)
+}
+
+func (a *aggregationJob) SendRequest(ctx context.Context) error {
+	// TODO dont rely on map iteration order and use weighted sampling instead
+	for nodeID, validator := range a.nodeIDsToValidator {
```
joshua-kim, line 227: Reminder to myself to move this code since this is a bug where it is currently

network/p2p/acp118/aggregator.go
joshua-kim: Added a comment

network/p2p/acp118/aggregator.go
joshua-kim: I refactored some code around so this edge-case doesn't exist anymore

network/p2p/acp118/aggregator.go
joshua-kim: Refactored some stuff around, but these should be handled as failed messages now

network/p2p/p2ptest/client.go
```diff
@@ -28,63 +29,100 @@ func NewClient(
...
+	return NewClientWithPeers(
+		t,
+		ctx,
+		clientNodeID,
+		map[ids.NodeID]p2p.Handler{serverNodeID: handler},
+	)
+}
 
+// NewClientWithPeers generates a client to communicate to a set of peers
+func NewClientWithPeers(
```
joshua-kim, line 41: I can make this a separate PR if requested - I've gotten feedback in the past about PRs with test utilities where it might be hard to understand why it's needed without corresponding usage.

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,363 @@
...
+				chainID,
+				[]byte("payload"),
+			)
+			require.NoError(err)
+			msg, err := warp.NewMessage(
+				unsignedMsg,
+				&warp.BitSetSignature{Signature: [bls.SignatureLen]byte{}},
+			)
+			require.NoError(err)
+			gotMsg, gotNum, gotDen, err := aggregator.AggregateSignatures(
```
aaronbuchwald, line 330: nit: should these be `aggregatedSignatureWeight` and `totalWeight` rather than num and den which suggests numerator/denominator?

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,363 @@
...
+				wantNum += tt.validators[i].Weight
+			}
+
+			wantDen := uint64(0)
+			for _, v := range tt.validators {
+				wantDen += v.Weight
+			}
+
+			require.Equal(wantNum, gotNum)
+			require.Equal(wantDen, gotDen)
```
aaronbuchwald, line 360: Same naming comment on numerator/denominator here

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,363 @@
...
+				{
+					NodeID:    nodeID0,
+					PublicKey: pk0,
+					Weight:    1,
+				},
+			},
+			wantSigners: []int{0},
+			quorumNum:   1,
+			quorumDen:   1,
+		},
```
aaronbuchwald, line 71: Super easy to read these test cases ❤️

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,363 @@
...
+		{
+			name: "aggregates from some validators - 1/3",
```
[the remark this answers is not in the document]
aaronbuchwald: Reading through the rest, this is probably fine as is, since there's already a convention to these names for success/failure, could just be more explicit in the success case

network/p2p/acp118/aggregator_test.go
joshua-kim: Yeah these names make a lot more sense

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,254 @@
...
+					zap.Error(err),
+				)
+
+				// Fast-fail if it's not possible to generate a signature that meets the
+				// minimum threshold
+				failedStakeWeight += result.Validator.Weight
+				if totalStakeWeight-failedStakeWeight < minThreshold {
+					return nil, 0, 0, ErrFailedAggregation
+				}
+				continue
```
StephenButtolph, line 179: Does this work with hypersdk's expected usage here? I thought that `num/dem` were going to be the maximum weights it would wait for, but that the minimum would be lower than that (meaning that if we are passing in the max here, we could be terminating when we realize we can't get the maximum... But we actually could have gotten the number that hypersdk wanted).

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,254 @@
...
+	signature, err := bls.SignatureFromBytes(response.Signature)
+	if err != nil {
+		r.results <- result{Validator: validator, Err: err}
+		return
+	}
+
+	if !bls.Verify(validator.PublicKey, signature, r.message.UnsignedMessage.Bytes()) {
+		r.results <- result{Validator: validator, Err: errFailedVerification}
+		return
+	}
```
StephenButtolph, line 251: nice

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,254 @@
...
+// the message. Caller is responsible for providing a well-formed canonical
+// validator set corresponding to the signer bitset in the message.
+func (s *SignatureAggregator) AggregateSignatures(
+	ctx context.Context,
+	message *warp.Message,
+	justification []byte,
+	validators []Validator,
+	quorumNum uint64,
+	quorumDen uint64,
+) (*warp.Message, uint64, uint64, error) {
```
StephenButtolph, line 75: I don't think this correctly handles the case that BLS public keys are shared across validators.

In Warp, only 1 signature is ever allowed from a BLS key in a warp message. If different nodeIDs have the same BLS key, their weights are aggregated for the BLS key's index

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,254 @@
...
+	results := make(chan result)
+	job := responseHandler{
+		message:            message,
+		nodeIDsToValidator: nodeIDsToValidator,
+		results:            results,
+	}
+
+	for _, nodeID := range sampleable {
+		// Avoid var shadowing in goroutine
+		nodeIDCopy := nodeID
```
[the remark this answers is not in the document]
joshua-kim: Ah - I forgot this was fixed.

network/p2p/acp118/aggregator.go
joshua-kim: Maybe it makes more sense for us to change the behavior so that this api blocks until all responses come back, or we reach the provided `num/den` threshold instead.

network/p2p/p2ptest/client.go
```diff
@@ -28,63 +29,100 @@ func NewClient(
...
-		// Send the request asynchronously to avoid deadlock when the server
-		// sends the response back to the client
-		go func() {
-			require.NoError(t, serverNetwork.AppRequest(ctx, clientNodeID, requestID, time.Time{}, requestBytes))
-		}()
+	clientSender.SendAppRequestF = func(ctx context.Context, nodeIDs set.Set[ids.NodeID], requestID uint32, requestBytes []byte) error {
+		for nodeID := range nodeIDs {
+			network, ok := peerNetworks[nodeID]
+			if !ok {
+				return fmt.Errorf("%s is not connected", nodeID)
```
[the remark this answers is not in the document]
joshua-kim: We could drop it - and this would give us more similar behavior to how the network sender implementation works. I think one downside is that we are only using this explicitly for testing and I think this makes tests easier to debug.

network/p2p/acp118/aggregator_test.go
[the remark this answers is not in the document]
joshua-kim: Changed naming for these test cases

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,363 @@
...
+					PublicKey: pk2,
+					Weight:    1,
+				},
+			},
+			wantSigners: []int{0},
+			quorumNum:   1,
+			quorumDen:   3,
+		},
+		{
+			name: "aggregates from some validators - 2/3",
```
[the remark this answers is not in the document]
joshua-kim: Added more test cases. I think there's one case that's hard to test (cancellation mid-way through aggregation) due to the api signature which we could test if we wanted to but would require a refactor.

network/p2p/acp118/aggregator.go
joshua-kim: Updated to use the `warp.Validator` type + added a test on shared public keys

network/p2p/p2ptest/client.go
```diff
@@ -19,72 +20,112 @@ import (
...
-				Message: errorMessage,
-			}))
-		}()
-
-		return nil
+	for nodeID := range peers {
+		peerSenders[nodeID].SendAppErrorF = func(ctx context.Context, _ ids.NodeID, requestID uint32, errorCode int32, errorMessage string) error {
+			// Send the request asynchronously to avoid deadlock when the server
+			// sends the response back to the client
+			nodeIDCopy := nodeID
```
aaronbuchwald, line 112: No longer needed with go 1.22 (https://go.dev/blog/loopvar-preview), but it looks like this shadows the loop variable incorrectly (according to < go1.22 recommended pattern to shadow loop variable).

The closure should capture `nodeID` and perform the copy when the function is invoked at which point `nodeID` can be overwritten by the loop.

Could we either remove the copy and rely on the fact that we're running > go1.22 or move the copy outside of the closure?

Needed to convince myself this was the case since I wasn't sure how golang would handle this case, so I wrote this program and get the following results:

```golang
package main

import (
	"fmt"
)

func main() {
	peers := make(map[int]struct{})
	peers[1] = struct{}{}
	peers[2] = struct{}{}
	peers[3] = struct{}{}

	funcs := make([]func() int, 0)
	for nodeID := range peers {
		funcs = append(funcs, func() int {
			nodeIDCopy := nodeID
			return nodeIDCopy
		})
	}

	for _, f := range funcs {
		fmt.Println(f())
	}
}
```

go1.21 result:
```
3
3
3
```

go1.22 result:
```
1
2
3
```

network/p2p/p2ptest/client.go
```diff
@@ -19,72 +20,112 @@ import (
...
-		go func() {
-			require.NoError(t, serverNetwork.AppGossip(ctx, clientNodeID, gossipBytes))
-		}()
+		for nodeID := range peers {
+			go func() {
+				require.NoError(t, peerNetworks[nodeID].AppGossip(ctx, nodeID, gossipBytes))
+			}()
+		}
 
 		return nil
```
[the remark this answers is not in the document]
joshua-kim: Since we're enforcing this now in AppRequest it makes sense to use the config now. Updated

network/p2p/p2ptest/client.go
joshua-kim: Removed the copy var

network/p2p/acp118/aggregator_test.go
```diff
@@ -0,0 +1,526 @@
...
+		signature           warp.BitSetSignature
+		validators          []*warp.Validator
+		quorumNum           uint64
+		quorumDen           uint64
+		wantTotalStake      int
+		wantMsg             *warp.Message
+		wantSigners         int
+		wantPossibleSigners []int
+		wantFinished        bool
+		wantErr             error
```
aaronbuchwald, line 65: It looks like a few fields are unused (`wantPossibleSigners` and `wantMsg`)

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	for _, pk := range nonSigners {
+		validator := publicKeysToValidators[pk]
+		for _, nodeID := range validator.NodeIDs {
+			if err := s.client.AppRequest(ctx, set.Of(nodeID), requestBytes, handler.HandleResponse); err != nil {
+				return nil, nil, nil, false, fmt.Errorf("failed to send aggregation request: %w", err)
+			}
+		}
+	}
```
aaronbuchwald, line 143: Should we eliminate the inner for loop and call `AppRequest` on all of the specified nodeIDs at once since `AppRequest` takes in a set?

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,260 @@
...
+
+	msg, err := newWarpMessage(message, signerBitSet, signatures)
+	if err != nil {
+		return nil, nil, nil, false, err
+	}
+
+	return msg, aggregatedStakeWeight, totalStakeWeight, true, nil
+}
```
aaronbuchwald, line 198: (optional nit) could we add a comment to explain the success/failure cases here? Reading through this, I have to pause to see why we handle each signature in the loop, but may exit the loop without having terminated.

I think I'm just used to seeing the previous style and what iirc is used in the EVM of checking for both terminate on reached required threshold and can no longer reach required threshold.

iiuc the new logic is terminate in the loop if we reach the termination threshold, otherwise we will process responses until either the context expires or we've processed every response (max out signature weight up to ctx expiration).

joshua-kim: > iiuc the new logic is terminate in the loop if we reach the termination threshold, otherwise we will process responses until either the context expires or we've processed every response (max out signature weight up to ctx expiration).

Yes that's correct. I can add comments to make this more clear.

network/p2p/acp118/aggregator.go
joshua-kim: Sure

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,260 @@
...
+	results                chan result
+}
+
+func (r *responseHandler) HandleResponse(
+	_ context.Context,
+	nodeID ids.NodeID,
+	responseBytes []byte,
+	err error,
+) {
+	validator := r.publicKeysToValidators[r.nodeIDsToPublicKeys[nodeID]]
```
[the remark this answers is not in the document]
joshua-kim: Removed both maps to just use a single map of node ids -> the indexed validator type

aaronbuchwald (approved):

cam-schultz (approved):
LGTM. This seems like it would really simplify the p2p integration in [signature-aggregator.](https://github.com/ava-labs/icm-services/tree/main/signature-aggregator) by leveraging the existing `p2p.Client`. We re-implement a lot of the response handling and routing logic.

network/p2p/acp118/aggregator.go
```diff
@@ -0,0 +1,255 @@
...
+	for i := 0; i < len(nonSigners); i++ {
+		select {
+		case <-ctx.Done():
+			// Try to return whatever progress we have if the context is cancelled
+			msg, err := newWarpMessage(message, signerBitSet, signatures)
+			if err != nil {
+				return nil, nil, nil, false, err
+			}
+
+			return msg, aggregatedStakeWeight, totalStakeWeight, false, nil
```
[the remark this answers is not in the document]
joshua-kim: Yeah I don't think we need this anymore because we also return the aggregated + total stake, so the caller can tell if the operation was canceled based off of that information as well as it won't match `quorumNum/quorumDen`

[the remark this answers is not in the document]
joshua-kim: Removed

PR #3395, opened 2024-09-17
ACP-77: Update `SecondsUntil` to prevent under-charging

StephenButtolph:
## Why this should be merged

Undercharging -> DoS
Overcharging -> Happy

Also limits fuzzing to an hour rather than a year. With race detection enabled, running for a full year can take too long.

## How this works

`SecondsUntil` -> `SecondsRemaining`

The block timestamp must only be able to advance to a point that validators are fully paying for that time.

## How this was tested

- [X] Updated unit tests

### DIFF
--- a/vms/platformvm/validators/fee/fee.go
+++ b/vms/platformvm/validators/fee/fee.go
@@ -91,10 +91,10 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
 	return cost
 }
 
-// SecondsUntil calculates the maximum number of seconds that a validator can
-// pay fees before their [costLimit] would be exceeded based on the dynamic fee
-// mechanism. The result is capped at [maxSeconds].
-func (s State) SecondsUntil(c Config, maxSeconds uint64, costLimit uint64) uint64 {
+// SecondsRemaining calculates the maximum number of seconds that a validator
+// can pay fees before their [costLimit] would be exceeded based on the dynamic
+// fee mechanism. The result is capped at [maxSeconds].
+func (s State) SecondsRemaining(c Config, maxSeconds uint64, costLimit uint64) uint64 {
 	// Because this function can divide by prices, we need to sanity check the
 	// parameters to avoid division by 0.
 	if c.MinPrice == 0 {
--- a/vms/platformvm/validators/fee/fee.go
+++ b/vms/platformvm/validators/fee/fee.go
@@ -43,7 +43,7 @@ func (s State) AdvanceTime(target gas.Gas, seconds uint64) State {
 }
 
 // CostOf calculates how much to charge based on the dynamic fee mechanism for
-// [seconds].
+// seconds.
 //
 // This implements the ACP-77 cost over time formula:
 func (s State) CostOf(c Config, seconds uint64) uint64 {
@@ -92,8 +92,8 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
 }
 
 // SecondsRemaining calculates the maximum number of seconds that a validator
-// can pay fees before their [costLimit] would be exceeded based on the dynamic
-// fee mechanism. The result is capped at [maxSeconds].
+// can pay fees before their costLimit would be exceeded based on the dynamic
+// fee mechanism. The result is capped at maxSeconds.
 func (s State) SecondsRemaining(c Config, maxSeconds uint64, costLimit uint64) uint64 {
 	// Because this function can divide by prices, we need to sanity check the
 	// parameters to avoid division by 0.
--- a/vms/platformvm/validators/fee/fee.go
+++ b/vms/platformvm/validators/fee/fee.go
@@ -91,35 +91,24 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
 	return cost
 }
 
-// SecondsUntil calculates the number of seconds that it would take to charge at
-// least [targetCost] based on the dynamic fee mechanism. The result is capped
-// at [maxSeconds].
-func (s State) SecondsUntil(c Config, maxSeconds uint64, targetCost uint64) uint64 {
+// SecondsUntil calculates the maximum number of seconds that a validator can
+// pay fees before their [costLimit] would be exceeded based on the dynamic fee
+// mechanism. The result is capped at [maxSeconds].
+func (s State) SecondsUntil(c Config, maxSeconds uint64, costLimit uint64) uint64 {
 	// Because this function can divide by prices, we need to sanity check the
 	// parameters to avoid division by 0.
 	if c.MinPrice == 0 {
-		if targetCost == 0 {
-			return 0
-		}
 		return maxSeconds
 	}
 
 	// If the current and target are the same, the price is constant.
 	if s.Current == c.Target {
-		price := gas.CalculatePrice(c.MinPrice, s.Excess, c.ExcessConversionConstant)
-		return secondsUntil(
-			uint64(price),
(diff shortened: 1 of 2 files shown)


vms/platformvm/validators/fee/fee.go
```diff
@@ -91,77 +91,46 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
...
-			return totalSeconds
+			return min(totalSeconds, maxSeconds)
 		}
 
-		seconds++
-		price := gas.CalculatePrice(c.MinPrice, s.Excess, c.ExcessConversionConstant)
-		cost, err = safemath.Add(cost, uint64(price))
-		if err != nil {
+		price := uint64(gas.CalculatePrice(c.MinPrice, s.Excess, c.ExcessConversionConstant))
+		if price > costLimit {
```
samliok, line 130: the unit tests won't ever hit this line, maybe its worth including some tests to cover this case?

samliok (approved):

vms/platformvm/validators/fee/fee.go
StephenButtolph: The existing unit tests hit this case

vms/platformvm/validators/fee/fee.go
```diff
@@ -91,77 +91,46 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
 	return cost
 }
 
-// SecondsUntil calculates the number of seconds that it would take to charge at
-// least [targetCost] based on the dynamic fee mechanism. The result is capped
-// at [maxSeconds].
-func (s State) SecondsUntil(c Config, maxSeconds uint64, targetCost uint64) uint64 {
+// SecondsRemaining calculates the maximum number of seconds that a validator
+// can pay fees before their [costLimit] would be exceeded based on the dynamic
```
ARR4N, line 95: (nit,fyi) The bracket notation in comments has [special treatment](https://tip.golang.org/doc/comment#doclinks) that won't apply here.

vms/platformvm/validators/fee/fee.go
```diff
@@ -91,77 +91,46 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
...
+// fee mechanism. The result is capped at [maxSeconds].
+func (s State) SecondsRemaining(c Config, maxSeconds uint64, costLimit uint64) uint64 {
 	// Because this function can divide by prices, we need to sanity check the
 	// parameters to avoid division by 0.
 	if c.MinPrice == 0 {
-		if targetCost == 0 {
-			return 0
-		}
 		return maxSeconds
```
ARR4N, line 101: Would the correct fail-safe here be to return 0? It depends on how this will be used, but my guess is that zero will be more likely to result in a bailout while `maxSeconds` might result in unintended behaviour.

vms/platformvm/validators/fee/fee.go
```diff
@@ -91,77 +91,46 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
...
-				zeroExcessCost,
-			)
-
+			secondsWithZeroExcess := costLimit / uint64(c.MinPrice)
 			totalSeconds, err := safemath.Add(seconds, secondsWithZeroExcess)
-			if err != nil || totalSeconds >= maxSeconds {
+			if err != nil {
+				// This is technically unreachable, but makes the code more
+				// clearly correct.
 				return maxSeconds
```
ARR4N, line 124: As above re `0` instead of `maxSeconds`

vms/platformvm/validators/fee/fee.go
```diff
@@ -91,77 +91,46 @@ func (s State) CostOf(c Config, seconds uint64) uint64 {
...
 
-		seconds++
-		price := gas.CalculatePrice(c.MinPrice, s.Excess, c.ExcessConversionConstant)
-		cost, err = safemath.Add(cost, uint64(price))
-		if err != nil {
+		price := uint64(gas.CalculatePrice(c.MinPrice, s.Excess, c.ExcessConversionConstant))
+		if price > costLimit {
 			return seconds
 		}
+		costLimit -= price
```
ARR4N, line 133: (nit) Determining if the `if price > costLimit` block is valid depends on knowledge of this subtraction, which makes the logic non-linear. Consider adding a `fundsRemaining := costLimit` and using that value instead throughout the loop.

vms/platformvm/validators/fee/fee_test.go
```diff
@@ -222,9 +221,9 @@ var (
 				MinPrice:                 minPrice,
 				ExcessConversionConstant: excessConversionConstant,
 			},
-			expectedSeconds: 11,
-			expectedCost:    math.MaxUint64, // Should not overflow
-			expectedExcess:  math.MaxUint32 * 11,
+			expectedSeconds: 10,
```
ARR4N, line 224: Why is `expectedSeconds` reducing? I expected the changes in test values to be non-negative because this new logic finds the _last_ time point at which they can pay the fee.

vms/platformvm/validators/fee/fee.go
StephenButtolph: To clarify a bit - this function is well defined for all input values. Returning `maxSeconds` here isn't intended as a flag that the input was invalid (if that was the case, this function would return an error). `maxSeconds` is the correct value to return in this case (as every second, the fee is equal to `0 nAVAX`. So, the validator can run forever).

vms/platformvm/validators/fee/fee.go
StephenButtolph: If this value were to overflow, it would mean that `maxSeconds` should be retuned. `0` would not be correct to return (if this value weren't well defined, we'd need to return an error rather than an ambiguous `0`)

vms/platformvm/validators/fee/fee_test.go
StephenButtolph: Previously we returned the first time that our balance went 0 or negative. Now we return the last time that we have a non-negative balance. So the tests should either be unchanged (if the value were hitting 0 exactly) or be reduced by one (which is the case here).

vms/platformvm/validators/fee/fee.go
StephenButtolph: til

vms/platformvm/validators/fee/fee.go
StephenButtolph: To clarify a bit, this function is essentially implementing:

"If you need to pay $x every second, how long does $y last?"

If you need to pay $0 every second, then it lasts forever.

vms/platformvm/validators/fee/fee.go
StephenButtolph: Similar clarification to above - this function is essentially implementing:

"If you need to pay $x every second, how long does $y last? If the answer is more than z then just say 1 minute."

If $x would last longer than MaxUint64, returning 0 is incorrect. We should return 1 minute.

vms/platformvm/validators/fee/fee.go
StephenButtolph: I ended up renaming `costLimit` to `fundsRemaining` entirely

vms/platformvm/validators/fee/fee_test.go
ARR4N: Ahh gotcha; I forgot what the previous behaviour was.

ARR4N (approved):

PR #3398, opened 2024-09-18
Fix node initialization in snowman test

yacovm:
## Why this should be merged

The topological consensus tests were supposed to have the instances initialized with a random tree of blocks, where each block points to a random set of descendants.

Instead, because of a bug that influenced the parent reference, the blocks were added to the consensus instance as a single block with the rest of the blocks as descendants.

That's not what we want to test, so this commit fixes this and further simplifies the initialization logic.

## How this works

Properly sets the parent of the block.

### DIFF
--- a/snow/consensus/snowman/network_test.go
+++ b/snow/consensus/snowman/network_test.go
@@ -74,26 +74,11 @@ func (n *Network) AddNode(t testing.TB, sm Consensus) error {
 	}
 
 	n.shuffleColors()
-	deps := map[ids.ID]ids.ID{}
 	for _, blk := range n.colors {
-		myDep, found := deps[blk.ParentV]
-		if !found {
-			myDep = blk.Parent()
-		}
-		myBlock := &snowmantest.Block{
-			Decidable: snowtest.Decidable{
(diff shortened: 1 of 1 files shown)


snow/consensus/snowman/network_test.go
```diff
@@ -74,26 +74,20 @@ func (n *Network) AddNode(t testing.TB, sm Consensus) error {
...
 				IDV:    blk.ID(),
 				Status: blk.Status,
 			},
-			ParentV: myDep,
+			ParentV: blk.ParentV,
 			HeightV: blk.Height(),
 			VerifyV: blk.Verify(context.Background()),
 			BytesV:  blk.Bytes(),
 		}
 		if err := sm.Add(myBlock); err != nil {
```
StephenButtolph, line 88: I think we can simplify this copy further by doing:
```golang
copiedBlk := *blk
if err := sm.Add(&copiedBlk); err != nil {
```

yacovm: Correct, done!

StephenButtolph (approved):

PR #3403, opened 2024-09-19
Make snowman use snowflake directly instead of snowball

yacovm:
## Why this should be merged

Snowball was introduced in order to improve the stability of snowflake, by taking into account the history when returning the preference.

However, if deployed with a configuration where the preference is small enough, and a network partition causes a 50-50 split of the stake, it can actually backfire, as described in 6e1a905.

Additionally, latest research [1] points out that snowflake suffices for snowman.

[1] https://arxiv.org/abs/2404.14250

## How this works

This commit removes snowball from snowman by substituting the snowball factory snowman uses, with a snowflake factory.

## How this was tested

Added a test that simulates a mixed network which runs snowman with and without snowball, and ensures that a mixed network still converges.

Also ran a modified node on Fuji and monitored its consensus related metrics, and it seemed functioning well.


snow/consensus/snowball/tree_test.go
```diff
@@ -137,14 +137,14 @@ func TestSnowballLastBinary(t *testing.T) {
...
-	tree := NewTree(SnowballFactory, params, zero)
+	tree := NewTree(SnowflakeFactory, params, zero)
 	tree.Add(one)
 
 	// Should do nothing
 	tree.Add(one)
 
-	expected := `SB(PreferenceStrength = 0, SF(Confidence = [0], Finalized = false)) Bits = [0, 255)
-    SB(Preference = 0, PreferenceStrength[0] = 0, PreferenceStrength[1] = 0, SF(Confidence = [0], Finalized = false, SL(Preference = 0))) Bit = 255`
+	expected := `SF(Confidence = [0], Finalized = false) Bits = [0, 255)
```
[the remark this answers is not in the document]
yacovm: I guess because it's hard to construct an object of a tree that is readable in the test because it contains too many fields, and is of recursive nature.

chains/manager.go
```diff
@@ -921,7 +922,7 @@ func (m *manager) createAvalancheChain(
 		return nil, fmt.Errorf("couldn't initialize snow base message handler: %w", err)
 	}
 
-	var snowmanConsensus smcon.Consensus = &smcon.Topological{}
+	var snowmanConsensus smcon.Consensus = &smcon.Topological{Factory: snowball.SnowflakeFactory}
```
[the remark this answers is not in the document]
yacovm: There is now snowflake package :-)

snow/consensus/snowball/flat_test.go
```diff
@@ -34,7 +34,7 @@ func TestFlat(t *testing.T) {
 
 	twoGreen := bag.Of(Green, Green)
 	require.True(f.RecordPoll(twoGreen))
-	require.Equal(Blue, f.Preference())
```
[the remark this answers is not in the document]
yacovm: It doesn't change, on the contrary, it remains the same. 

In this test we fed snowball with a threshold of blue, and then with a threshold of green. It needs another small push (line 41) to change its mind further.

snow/consensus/snowman/mixed_test.go
```diff
@@ -0,0 +1,55 @@
...
+				}
+
+				require.NoError(n.AddNode(t, sm))
+			}
+
+			for !n.Finalized() {
+				require.NoError(n.Round())
+			}
+
+			require.True(n.Agreement())
```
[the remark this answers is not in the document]
yacovm: I think it is sufficient to check at this package level, as just snowman alone is pretty high level and complex, and encapsulates in it quite a lot of things, such as transitive voting based on topological sort of the block graph, and this is essentially what is tested here.

I think the only thing not tested here, is what happens if nodes are disconnected or unreachable, but I don't think this should have any difference, because snowball doesn't handle this aspect, but delegates it to snowflake.

snow/consensus/snowman/topological.go
```diff
@@ -29,16 +29,20 @@ var (
 )
 
 // TopologicalFactory implements Factory by returning a topological struct
-type TopologicalFactory struct{}
+type TopologicalFactory struct {
+	factory snowball.Factory
```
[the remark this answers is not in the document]
yacovm: both snowflake and snowball factories are in the snowball package. Their interface is the same, and therefore there is a single Factory interface which can initialize both.

snow/consensus/snowball/binary_snowball_test.go
```diff
@@ -19,25 +19,26 @@ func TestBinarySnowball(t *testing.T) {
 	beta := 2
 	terminationConditions := newSingleTerminationCondition(alphaConfidence, beta)
 
-	sb := newBinarySnowball(alphaPreference, terminationConditions, red)
-	require.Equal(red, sb.Preference())
-	require.False(sb.Finalized())
+	sf := newBinarySnowflake(alphaPreference, terminationConditions, red)
```
StephenButtolph, line 22: Why is this test modified? It seems to be a duplicate of `TestBinarySnowflake` now.

snow/consensus/snowball/binary_snowball_test.go
StephenButtolph, line 1: I'm not sure why this file is being modified. We aren't changing the code that it was previously testing... If we feel like the `binarySnowflake` struct needs additional coverage shouldn't we be adding new tests to the `binary_snowflake_test.go` file rather than here?

snow/consensus/snowball/nnary_snowball_test.go
StephenButtolph, line 1: Because we aren't deleting the code that this was testing, `nnary_snowball.go`, why are we deleting the tests?

snow/consensus/snowball/unary_snowball_test.go
StephenButtolph, line 1: Because we aren't deleting the code that this was testing, `unary_snowball.go`, why are we deleting the tests?

snow/consensus/snowman/mixed_test.go
```diff
@@ -0,0 +1,55 @@
...
+				var sm Consensus
+				if i%2 == 0 {
+					factory := TopologicalFactory{factory: snowball.SnowflakeFactory}
+					sm = factory.New()
+				} else {
+					factory := TopologicalFactory{factory: snowball.SnowballFactory}
+					sm = factory.New()
+				}
+
+				require.NoError(n.AddNode(t, sm))
```
StephenButtolph, line 45: nit: Seems like we can cleanup the if block a bit
```suggestion
				var sbFactory snowball.Factory
				if i%2 == 0 {
					sbFactory = snowball.SnowflakeFactory
				} else {
					sbFactory = snowball.SnowballFactory
				}

				factory := TopologicalFactory{factory: sbFactory}
				sm := factory.New()
				require.NoError(n.AddNode(t, sm))
```

snow/consensus/snowball/tree_test.go
```diff
@@ -485,23 +485,44 @@ func TestSnowballTrinary(t *testing.T) {
 	require.Equal(Red, tree.Preference())
 	require.False(tree.Finalized())
 
+	//       *
+	//     1/ \
+	//     R   *
+	//        / \
+	//       G   B
```
StephenButtolph, line 492: What do the numbers on the edges mean? It seems like they are referencing Snowball counters but we are using snowflake here and not snowball.

snow/consensus/snowball/binary_snowball_test.go
yacovm: ok will revert

StephenButtolph (commented):
Code changes LGTM. Going to hold off on approving/merging until we've canaried this out on Fuji + Mainnet

StephenButtolph (approved):

PR #3404, opened 2024-09-19
ACP-77: Current validators API for SoV

ceyonur:
This pull request introduces a new gRPC method `GetCurrentValidatorSet` to the `ValidatorState` service, allowing clients to retrieve the current validator set for a given subnet. The changes span multiple files to implement this new functionality, including updates to the protobuf definitions, client and server interfaces, and associated tests.

### Additions to `ValidatorState` Service:

* [`proto/pb/validatorstate/validator_state_grpc.pb.go`](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR27): Added the `GetCurrentValidatorSet` method to both client and server interfaces, along with its handler and full method name constant. [[1]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR27) [[2]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR44-R46) [[3]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR93-R101) [[4]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR116-R118) [[5]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR138-R140) [[6]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR226-R243) [[7]](diffhunk://#diff-ca6cb9352fea44626eedb103d3289e4f115bebaa6f4c33a9f970bb7ce9cc8cbaR267-R270)
* [`proto/validatorstate/validator_state.proto`](diffhunk://#diff-ffc5ba8004111946c92ac7b94e459a8ffdfde36e731d52ba91662c44b8a30157R20-R22): Defined the `GetCurrentValidatorSetRequest` and `GetCurrentValidatorSetResponse` messages, and added the new RPC method to the `ValidatorState` service. [[1]](diffhunk://#diff-ffc5ba8004111946c92ac7b94e459a8ffdfde36e731d52ba91662c44b8a30157R20-R22) [[2]](diffhunk://#diff-ffc5ba8004111946c92ac7b94e459a8ffdfde36e731d52ba91662c44b8a30157R46-R74)

### Client and Server Implementation:

* [`snow/validators/gvalidators/validator_state_client.go`](diffhunk://#diff-50b66f7c335fe6f0f3bccbdf4647a09457628fb171bd67a208e0f63f66d9fe1cR97-R142): Implemented the `GetCurrentValidatorSet` method in the `Client` struct to call the new gRPC method.
* [`snow/validators/gvalidators/validator_state_server.go`](diffhunk://#diff-dd7d4ab59825a19909d2976e8c18b42131ef8349090dc3fe0b4905c681d7a5eaR82-R119): Implemented the `GetCurrentValidatorSet` method in the `Server` struct to handle incoming requests.

### State Interface and Mock Updates:

* [`snow/validators/state.go`](diffhunk://#diff-087fc55135fc29c2ef4729c6cefdf957d8da89b735813733e667a832a514fc7eR35-R43): Added the `GetCurrentValidatorSet` method to the `State` interface and implemented it in the `lockedState` struct. [[1]](diffhunk://#diff-087fc55135fc29c2ef4729c6cefdf957d8da89b735813733e667a832a514fc7eR35-R43) [[2]](diffhunk://#diff-087fc55135fc29c2ef4729c6cefdf957d8da89b735813733e667a832a514fc7eR90-R99) [[3]](diffhunk://#diff-087fc55135fc29c2ef4729c6cefdf957d8da89b735813733e667a832a514fc7eR113-R117)
* [`snow/validators/traced_state.go`](diffhunk://#diff-68d9bf7d6b7a2ad8b4283ba8fae3602c82c8e40c0ea3484ce92a2fb2a397edc1R75-R86): Added tracing support for the `GetCurrentValidatorSet` method.
* [`snow/validators/validator.go`](diffhunk://#diff-4a87bd156548ce0dd981328612ad4a82acaebf44dfdb80bdf523c888ad83acccR32-R41): Defined the `GetCurrentValidatorOutput` struct to represent the current validator data.
* [`snow/validators/validatorsmock/state.go`](diffhunk://#diff-851431ea89dc5c56162128dd0c772104755fa1799bcdfc5191f94aba71e71496R59-R75): Added mock implementations for the `GetCurrentValidatorSet` method.
* [`snow/validators/validatorstest/state.go`](diffhunk://#diff-e4b0eba4c5c0aef0699270fbefe7ab5544a66f8d65fa8c7a884bf232d5f102feR33-R39): Added test support for the `GetCurrentValidatorSet` method. [[1]](diffhunk://#diff-e4b0eba4c5c0aef0699270fbefe7ab5544a66f8d65fa8c7a884bf232d5f102feR33-R39) [[2]](diffhunk://#diff-e4b0eba4c5c0aef0699270fbefe7ab5544a66f8d65fa8c7a884bf232d5f102feR85-R97)

### Platform VM State Changes:

* [`vms/platformvm/state/state.go`](diffhunk://#diff-602ce6ed16158db7a4822a397925858d586f5f96cbe37af869676fb1b10161efR211-R212): Added the `GetCurrentValidatorSet` method to the `State` interface and implemented it in the `state` struct, including logic to determine if the subnet is L1 and fetch the current validators. [[1]](diffhunk://#diff-602ce6ed16158db7a4822a397925858d586f5f96cbe37af869676fb1b10161efR211-R212) [[2]](diffhunk://#diff-602ce6ed16158db7a4822a397925858d586f5f96cbe37af869676fb1b10161efR762-R849)
* [`vms/platformvm/state/mock_state.go`](diffhunk://#diff-047dab3b5dbb5f7d2c10f33d2b342b2a8b1b5c3fde79a39e422b3a43983d152aR427-R443): Added mock implementations for the `GetCurrentValidatorSet` method.


vms/platformvm/state/state.go
```diff
@@ -757,6 +759,94 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+			return nil, 0, false, err
+		}
+
+		validationID, err := ids.ToID(subnetIDIter.Value())
+		if err != nil {
+			return nil, 0, false, fmt.Errorf("failed to parse validation ID: %w", err)
+		}
+
+		// TODO: consider optimizing this to avoid hitting the inactiveDB if
+		// all validators are active (inactive weight is 0)
```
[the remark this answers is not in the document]
ceyonur: wouldn't that cause us to iterate over all validators? I guess that's still preferable over iterating DB keys?

vms/platformvm/state/state.go
```diff
@@ -757,6 +759,94 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+	}
+
+	return result, s.currentHeight, isL1, nil
+}
+
+func (s *state) IsL1(subnetID ids.ID) (bool, error) {
+	// short circuit if this is the primary network
+	if subnetID == constants.PrimaryNetworkID {
+		return false, nil
+	} else {
```
darioush, line 837: nit: else is not needed here

snow/validators/traced_state.go
```diff
@@ -72,3 +72,15 @@ func (s *tracedState) GetValidatorSet(
 
 	return s.s.GetValidatorSet(ctx, height, subnetID)
 }
+
+func (s *tracedState) GetCurrentValidatorSet(
+	ctx context.Context,
+	subnetID ids.ID,
+) (map[ids.ID]*GetCurrentValidatorOutput, uint64, bool, error) {
+	ctx, span := s.tracer.Start(ctx, s.getValidatorSetTag, oteltrace.WithAttributes(
```
darioush, line 80: seems that getSubnetID (Above) is using getValidatorSetTag, but there's a `getSubnetIDTag`.
Seems that we would make a new tag for this method as well

snow/validators/validatorstest/state.go
```diff
@@ -80,3 +82,16 @@ func (vm *State) GetValidatorSet(
...
+	ctx context.Context,
+	subnetID ids.ID,
+) (map[ids.ID]*validators.GetCurrentValidatorOutput, uint64, bool, error) {
+	if vm.GetCurrentValidatorSetF != nil {
+		return vm.GetCurrentValidatorSetF(ctx, subnetID)
+	}
+	if vm.CantGetCurrentValidatorSet && vm.T != nil {
+		require.FailNow(vm.T, "unexpectedly called GetCurrentValidatorSet")
+	}
+	return nil, 0, false, nil
```
darioush, line 96: nit: I suppose the pattern would be to make a new `errGetCurrentValidatorSet` as the final return value here.
not sure it matters except when vm.T is not set

vms/platformvm/state/state.go
```diff
@@ -757,6 +759,94 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+		validationID, err := ids.ToID(subnetIDIter.Value())
+		if err != nil {
+			return nil, 0, false, fmt.Errorf("failed to parse validation ID: %w", err)
+		}
+
+		// TODO: consider optimizing this to avoid hitting the inactiveDB if
+		// all validators are active (inactive weight is 0)
+		vdr, err := s.GetSubnetOnlyValidator(validationID)
+		if err == database.ErrNotFound {
+			// validator was deleted in modified
```
darioush, line 812: could we clarify this comment?

vms/platformvm/state/state.go
```diff
@@ -761,6 +763,71 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+		}
+
+		// TODO: consider optimizing this to avoid hitting the inactiveDB if
+		// all validators are active (inactive weight is 0)
+		vdr, err := s.GetSubnetOnlyValidator(validationID)
+		if err == database.ErrNotFound {
+			// if the validator is not found, it is removed from the validator set
+			// but not yet removed from the database
+			continue
+		}
```
StephenButtolph, line 811: I think this should just error in that case
```suggestion
```

proto/validatorstate/validator_state.proto
```diff
@@ -40,12 +43,32 @@ message GetValidatorSetRequest {
...
+  bytes subnet_id = 1;
+}
+
 message Validator {
   bytes node_id = 1;
   uint64 weight = 2;
   bytes public_key = 3;
 }
 
+message CurrentValidator {
```
iansuvak, line 56: Possibly dumb question but is there a good reason for not adding new fields to the existing `Validator` message instead of adding a new one? i.e. do we see the data returned by them becoming incompatible at some point? It seems like it should be compatible since protobuf fields are all optional

vms/platformvm/state/state_test.go
```diff
@@ -2088,3 +2088,257 @@ func TestSubnetOnlyValidatorAfterLegacyRemoval(t *testing.T) {
...
+					Weight:    0,
+					StartTime: time.Now(),
+				},
+			},
+			sovs: []SubnetOnlyValidator{
+				{
+					ValidationID: ids.GenerateTestID(),
+					SubnetID:     subnetID1,
+					NodeID:       ids.GenerateTestNodeID(),
+					StartTime:    uint64(time.Now().Unix()),
```
darioush, line 2232: could we define a single `now := time.Now()` to avoid the possibility of different return values for time.Now() for some unlucky test runs?

vms/platformvm/state/state_test.go
```diff
@@ -2088,3 +2088,257 @@ func TestSubnetOnlyValidatorAfterLegacyRemoval(t *testing.T) {
...
+			state := newTestState(t, db)
+
+			for _, staker := range test.initial {
+				primaryStaker := &Staker{
+					TxID:      ids.GenerateTestID(),
+					SubnetID:  constants.PrimaryNetworkID,
+					NodeID:    staker.NodeID,
+					PublicKey: staker.PublicKey,
+					Weight:    5,
+					StartTime: staker.StartTime.Add(-1 * time.Second),
```
darioush, line 2278: why do we add -1 seconds here? could we add a comment?

vms/platformvm/state/state_test.go
```diff
@@ -2088,3 +2088,257 @@ func TestSubnetOnlyValidatorAfterLegacyRemoval(t *testing.T) {
...
+				}
+				require.NoError(state.PutCurrentValidator(primaryStaker))
+				require.NoError(state.PutCurrentValidator(staker))
+			}
+
+			for _, sov := range test.sovs {
+				sov.RemainingBalanceOwner = []byte{}
+				sov.DeactivationOwner = []byte{}
```
darioush, line 2286: could we add a comment why these are being initialized to `[]byte{}` vs `nil`?

vms/platformvm/state/state_test.go
```diff
@@ -2088,3 +2088,257 @@ func TestSubnetOnlyValidatorAfterLegacyRemoval(t *testing.T) {
...
+				for _, expectedSOV := range sovBySubnetID[subnetID] {
+					currentValidator, ok := currentValidators[expectedSOV.ValidationID]
+					require.True(ok)
+					require.Equal(expectedSOV.ValidationID, currentValidator.ValidationID)
+					require.Equal(expectedSOV.NodeID, currentValidator.NodeID)
+					require.Equal(expectedSOV.PublicKey, currentValidator.PublicKey.Serialize())
+					require.Equal(expectedSOV.Weight, currentValidator.Weight)
+					require.Equal(expectedSOV.StartTime, currentValidator.StartTime)
+					require.Equal(expectedSOV.MinNonce, currentValidator.SetWeightNonce)
+					require.Equal(expectedSOV.isActive(), currentValidator.IsActive)
```
darioush, line 2338: should there be some inputs with `EndAccumulatedFee > 0` so sometimes .isActive returns true?

snow/validators/traced_state.go
darioush: Should we fix this in this PR? Seems simple enough to do so.

darioush (approved):

iansuvak (approved):
LGTM with the minor question on the protobuf

proto/validatorstate/validator_state.proto
ceyonur: It would mean a breaking change. Changing `Validator` would require changes to related functions + clients. I think if we want to we can deprecate old methods and types in favor of `CurrentValidator`. @StephenButtolph

ceyonur:
> My main question with this also in conjunction with [ava-labs/subnet-evm#1263](https://github.com/ava-labs/subnet-evm/pull/1263), is whether there is a possibility that the SoV validator would share the same nodeID as the ones initially set?

Both this API and Subnet-EVM relies on Validation ID rather than NodeID. The case where different Validation IDs mapped to same NodeID should already be handled by the state ([here](https://github.com/ava-labs/avalanchego/blob/test-only-current-validators-api/vms/platformvm/state/subnet_only_validator.go#L314-L317)), so this API should not return duplicated ones.

For the Subnet-EVM we first handle deletions when we handle the new set from P-Chain. It guarantees that we delete the uptime from the state, even if the same nodeID is being deleted + readded under different validationIDs in the same load period. After deletion we add new data with the same nodeID. See [here](https://github.com/ava-labs/subnet-evm/blob/uptime-tracking/plugin/evm/vm.go#L1465-L1497)

vms/platformvm/state/state_test.go
ceyonur: added

vms/platformvm/state/state_test.go
ceyonur: changed

vms/platformvm/state/state_test.go
ceyonur: added

vms/platformvm/state/state_test.go
ceyonur: yea good idea, added few

proto/validatorstate/validator_state.proto
yacovm: Why not just have a field of type `Validator` in `CurrentValidator`? then you don't need to copy the first three fields.

proto/validatorstate/validator_state.proto
```diff
@@ -40,12 +43,32 @@ message GetValidatorSetRequest {
...
+  bytes node_id = 1;
+  uint64 weight = 2;
+  bytes public_key = 3;
+  uint64 start_time = 4;
+  uint64 set_weight_nonce = 5;
```
yacovm, line 61: what is `set_weight_nonce` for?

snow/validators/gvalidators/validator_state_server.go
```diff
@@ -79,3 +79,41 @@ func (s *Server) GetValidatorSet(ctx context.Context, req *pb.GetValidatorSetReq
...
+			NodeId:         vdr.NodeID.Bytes(),
+			StartTime:      vdr.StartTime,
+			IsActive:       vdr.IsActive,
+			ValidationId:   vdr.ValidationID[:],
+			Weight:         vdr.Weight,
+			SetWeightNonce: vdr.SetWeightNonce,
+			IsSov:          vdr.IsSoV,
+		}
+		if vdr.PublicKey != nil {
+			// This is a performance optimization to avoid the cost of compression
```
yacovm, line 111: serializing in compressed form is cheap. It is the un-compression that is expensive.

vms/platformvm/state/state.go
```diff
@@ -831,6 +833,64 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
 	s.expiryDiff.DeleteExpiry(entry)
 }
 
+func (s *state) GetCurrentValidatorSet(ctx context.Context, subnetID ids.ID) (map[ids.ID]*validators.GetCurrentValidatorOutput, uint64, error) {
+	result := make(map[ids.ID]*validators.GetCurrentValidatorOutput)
```
yacovm, line 837: would it make sense to check `currentStakers` is not nil?

proto/validatorstate/validator_state.proto
ceyonur: it's the `MinNonce` of L1 validators

snow/validators/gvalidators/validator_state_server.go
ceyonur: literally copied from L73

vms/platformvm/state/state.go
ceyonur: currentStakers are initialized to be empty in `New`

snow/validators/gvalidators/validator_state_server.go
yacovm: Understood. 

Though I think that this comment is a bit misleading - what it's probably actually trying to say is that in order to avoid the expensive decompression in the client side, the server side is not compressing.

snow/validators/state.go
```diff
@@ -91,3 +110,8 @@ func NewNoValidatorsState(state State) State {
 func (*noValidators) GetValidatorSet(context.Context, uint64, ids.ID) (map[ids.NodeID]*GetValidatorOutput, error) {
 	return nil, nil
 }
+
+func (*noValidators) GetCurrentValidatorSet(context.Context, ids.ID) (map[ids.ID]*GetCurrentValidatorOutput, uint64, error) {
+	// TODO: should we return the actual height?
+	return nil, 0, nil
```
StephenButtolph, line 116: Feels like we should resolve this TODO.

snow/validators/traced_state.go
```diff
@@ -51,7 +53,7 @@ func (s *tracedState) GetCurrentHeight(ctx context.Context) (uint64, error) {
 }
 
 func (s *tracedState) GetSubnetID(ctx context.Context, chainID ids.ID) (ids.ID, error) {
-	ctx, span := s.tracer.Start(ctx, s.getValidatorSetTag, oteltrace.WithAttributes(
+	ctx, span := s.tracer.Start(ctx, s.getSubnetIDTag, oteltrace.WithAttributes(
```
StephenButtolph, line 56: super nit - but I think we could make this its own PR

snow/validators/validator.go
```diff
@@ -29,3 +29,14 @@ type GetValidatorOutput struct {
...
 	Weight    uint64
 }
+
+type GetCurrentValidatorOutput struct {
+	ValidationID   ids.ID
+	NodeID         ids.NodeID
+	PublicKey      *bls.PublicKey
+	Weight         uint64
+	StartTime      uint64
+	SetWeightNonce uint64
```
StephenButtolph, line 39: Same comment as in the proto - I think this should be `MinNonce`

vms/platformvm/state/state.go
```diff
@@ -831,6 +833,64 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+			Weight:         validator.Weight,
+			StartTime:      uint64(validator.StartTime.Unix()),
+			SetWeightNonce: 0,
+			IsActive:       true,
+			IsSoV:          false,
+		}
+	}
+
+	// this is L1 subnet, fetch all validation IDs from subnetIDNodeIDDB
+	// with subnetID as prefix
```
StephenButtolph, line 856: This comment seems out of date. We don't know if this is an L1 here or not (but we don't care)

vms/platformvm/state/state.go
```diff
@@ -831,6 +833,64 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+	subnetIDIter := s.subnetIDNodeIDDB.NewIteratorWithPrefix(
```
StephenButtolph, line 857: Should this be `validationIDIter`? or just `it`?

vms/platformvm/state/state.go
```diff
@@ -831,6 +833,64 @@ func (s *state) DeleteExpiry(entry ExpiryEntry) {
...
+			return nil, 0, err
+		}
+
+		validationID, err := ids.ToID(subnetIDIter.Value())
+		if err != nil {
+			return nil, 0, fmt.Errorf("failed to parse validation ID: %w", err)
+		}
+
+		// TODO: consider optimizing this to avoid hitting the subnetIDNodeIDDB and read from actives lookup
+		// if all validators are active (inactive weight is 0)
```
StephenButtolph, line 873: I think this TODO should be where we grab the iterator, not here.

RELEASES.md
```diff
@@ -9,6 +9,10 @@
   - `"l1-inactive-validators-cache-size"`
   - `"l1-subnet-id-node-id-cache-size"`
 
+### gRPC Interface
+
+- Added `GetCurrentValidatorSet` to the `validators.State` interface
+
```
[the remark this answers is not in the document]
ceyonur: thought this can be considered as API changes as we change the VM interface. removed it.

proto/validatorstate/validator_state.proto
ceyonur: Well I think it wasn't that much painful as I thought it would be. I was under impression that I also had to change `GetValidatorSet` return types, but turned out we can just leave unused fields empty for that function and keep the return type same.

snow/validators/traced_state.go
ceyonur: https://github.com/ava-labs/avalanchego/pull/3542/files

proto/validatorstate/validator_state.proto
```diff
@@ -17,6 +17,9 @@ service ValidatorState {
   // GetValidatorSet returns the weights of the nodeIDs for the provided
   // subnet at the requested P-chain height.
   rpc GetValidatorSet(GetValidatorSetRequest) returns (GetValidatorSetResponse);
+  // GetCurrentValidatorSet returnsthe  validator set for the provided
+  // subnet at the current P-chain height.
```
StephenButtolph, line 21:
```suggestion
  // GetCurrentValidatorSet returns the validator set for the provided subnet at
  // the current P-chain height.
```

snow/validators/state.go
```diff
@@ -32,6 +32,15 @@ type State interface {
 		height uint64,
 		subnetID ids.ID,
 	) (map[ids.NodeID]*GetValidatorOutput, error)
+
+	// GetCurrentValidatorSet returns the current validators of the provided subnet
+	// and the current P-Chain height.
+	// Map is keyed by ValidationID.
+	// The returned map should not be modified.
```
StephenButtolph, line 39: Why is this documented as not being able to be modified? I think everywhere this is implemented it can be modified.

vms/platformvm/validators/manager.go
```diff
@@ -409,3 +411,11 @@ func (m *manager) GetSubnetID(_ context.Context, chainID ids.ID) (ids.ID, error)
 func (m *manager) OnAcceptedBlockID(blkID ids.ID) {
 	m.recentlyAccepted.Add(blkID)
 }
+
+func (m *manager) GetCurrentValidatorSet(ctx context.Context, subnetID ids.ID) (map[ids.ID]*validators.GetCurrentValidatorOutput, uint64, error) {
+	vdrs, currentHeight, err := m.state.GetCurrentValidatorSet(ctx, subnetID)
+	if err != nil {
+		return nil, 0, err
+	}
+	return vdrs, currentHeight, nil
```
StephenButtolph, line 420:
```suggestion
	return m.state.GetCurrentValidatorSet(ctx, subnetID)
```

StephenButtolph (commented):
final nits

snow/validators/state.go
ceyonur: I think that's a left over from the previous attempt to use in-memory validators manager (from the `GetValidatorSet`). However wouldn't modifying the any validator in that map should still be illegal since we use currentStakers [here](https://github.com/ava-labs/avalanchego/blob/test-only-current-validators-api/vms/platformvm/state/state.go#L844-L853)?

StephenButtolph: We create a new map every time, so it can always be owned by the caller. I'm not sure why it would matter where the data comes from that is put into the map

ceyonur: removed it

PR #3407, opened 2024-09-20
Ensure slices are zeroed

StephenButtolph:
## Why this should be merged

Golang can only GC the memory if there are no remaining references to it. Because it's possible to re-extend slices, the memory doesn't get GCed.

## How this works

Uses `clear` (and deletes the `ZeroSlice` helper).

## How this was tested

N/A

### DIFF
--- a/utils/zero.go
+++ b/utils/zero.go
@@ -4,16 +4,6 @@
 package utils
 
 // Returns a new instance of a T.
-func Zero[T any]() T {
-	return *new(T)
-}
-
-// ZeroSlice sets all values of the provided slice to the type's zero value.
-//
-// This can be useful to ensure that the garbage collector doesn't hold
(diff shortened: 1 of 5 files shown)


utils/zero.go
```diff
@@ -4,16 +4,6 @@
...
-//
-// This can be useful to ensure that the garbage collector doesn't hold
-// references to values that are no longer desired.
-func ZeroSlice[T any](s []T) {
-	for i := range s {
-		s[i] = *new(T)
-	}
+func Zero[T any]() (_ T) {
+	return
 }
```
StephenButtolph, line 9: No clue why we had this and it was never used... weird

yacovm (approved):

tsachiherman (approved):

PR #3410, opened 2024-09-24
Make bootstrapping handle its own timeouts

yacovm:
## Why this should be merged

Currently, an engine registers timeouts into the handler, which schedules the timeouts on behalf of the the engine.

The handler then notifies the engine when the timeout expired.
    
However, the only engine that uses this mechanism is the bootstrapping engine, and not the other engine types such as the snowman and state sync engines.
    
It therefore makes sense that the bootstrapper handle its own timeouts instead of delegating them to the handler.
    
By moving the timeout handling into the bootstrapper, we can make the API of the common.Engine be slimmer by removing the Timeout() method from it.

## How this works

Just a refactoring which removes any timeout logic from the handler, and moves it into the bootstrapper instead.

## How this was tested

CI tests


snow/engine/common/timer.go
```diff
@@ -3,12 +3,87 @@
...
+		timeouts: timeoutBuff,
+		onTimeout: func() {
+			onTimeout()
+			<-timeoutBuff
+		},
+	}
+}
+
+// RegisterTimeout fires the function the timeout handler is initialized with no later than the given timeout.
+func (th *timeoutHandler) RegisterTimeout(d time.Duration) {
```
yacovm, line 62: I just moved this code section from `handler.go`

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -722,16 +731,16 @@ func (b *Bootstrapper) getLastAccepted(ctx context.Context) (snowman.Block, erro
...
-func (b *Bootstrapper) Timeout(ctx context.Context) error {
+func (b *Bootstrapper) Timeout() error {
 	if !b.awaitingTimeout {
 		return errUnexpectedTimeout
 	}
 	b.awaitingTimeout = false
 
 	if !b.Config.BootstrapTracker.IsBootstrapped() {
-		return b.restartBootstrapping(ctx)
+		return b.restartBootstrapping(context.TODO())
```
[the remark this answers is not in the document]
yacovm: Right, if you look at the previous code - the context that is passed is a [TODO](https://github.com/ava-labs/avalanchego/pull/3410/files/c25a68d2e6f65435c1e714d3b826674fa0e5fed3#diff-1b154e2d9b5b0909a66199417df6ff79ee9a18e83075b39ef2acf3634558342eL940) one.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,87 @@
...
 
-import "time"
+import (
+	"sync"
+	"time"
+)
 
-// Timer describes the standard interface for specifying a timeout
-type Timer interface {
+// PreemptionSignal signals when to preempt the timeouts of the timeout handler.
```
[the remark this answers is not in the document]
yacovm: I found the current initialization inside of the subnets - a bit ad-hoc and one that doesn't convey properly the intent.

Now the initialization logic is simpler, and isn't part of the subnets object, which I find to be more informative to the reader of the code.

snow/engine/common/bootstrap_tracker.go
```diff
@@ -11,8 +11,9 @@ type BootstrapTracker interface {
 	// Returns true iff done bootstrapping
 	IsBootstrapped() bool
 
-	// Bootstrapped marks the named chain as being bootstrapped
+	// Bootstrapped marks the named chain as being bootstrapped.
```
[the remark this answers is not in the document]
yacovm: ok I changed it accordingly.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,87 @@
...
+// NewTimeoutHandler constructs a new timeout handler with the given function to be invoked upon a timeout,
+// unless the preemptionSignal is closed and in which case it invokes the function immediately.
+func NewTimeoutHandler(onTimeout func(), preemptionSignal <-chan struct{}) *timeoutHandler {
+	timeoutBuff := make(chan struct{}, 1)
+	return &timeoutHandler{
+		preemptionSignal: preemptionSignal,
+		newTimer: func(d time.Duration) (<-chan time.Time, func() bool) {
+			timer := time.NewTimer(d)
+			return timer.C, timer.Stop
+		},
```
aaronbuchwald, line 52: What's the reason for specifying this function here rather than just using `time.NewTimer` directly?

It looks like we replace this with an identical implementation in one of our test cases

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{
+			desc: "Preemption makes the function be invoked immediately despite no clock tick",
+			body: func(th *timeoutHandler, wg *sync.WaitGroup, signal *PreemptionSignal) {
+				th.RegisterTimeout(time.Hour)
+				signal.Preempt()
+				wg.Wait()
+			},
+		},
+	} {
+		t.Run(testCase.desc, func(t *testing.T) {
```
aaronbuchwald, line 40: Is it possible to remove `clock` from these tests?

What do you think of adding test cases for multiple timeouts w/ and w/o preemption?

aaronbuchwald (dismissed):

snow/engine/common/timer.go
yacovm: Because I needed it for the tests :-)

snow/engine/common/timer_test.go
yacovm: I added two tests as requested, but I need the clock for these tests.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,87 @@
...
+		}
+
+		// If there is already a timeout ready to fire - just drop the
+		// additional timeout. This ensures that all goroutines that are spawned
+		// here are able to close if the chain is shutdown.
+		select {
+		case th.timeouts <- struct{}{}:
+			th.onTimeout()
+		default:
+		}
```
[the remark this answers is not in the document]
yacovm: There should be at least a single single timeout called, as `th.timeouts` is initialized to a buffered channel with a capacity of one. 

If two are called concurrently, then there is an attempt to only call one.

This code was not made in this PR, I just [moved](https://github.com/ava-labs/avalanchego/pull/3410/files/fe44d9533ba3fc3f4837679948f455765ccbb59f#diff-1b154e2d9b5b0909a66199417df6ff79ee9a18e83075b39ef2acf3634558342eL300) it around.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -139,7 +139,16 @@ func New(config Config, onFinished func(ctx context.Context, lastReqID uint32) e
...
 		executedStateTransitions: math.MaxInt,
 		onFinished:               onFinished,
-	}, err
+	}
+
+	timeout := func() {
+		if err := bs.Timeout(); err != nil {
+			bs.Config.Ctx.Log.Warn("Encountered error during bootstrapping: %w", zap.Error(err))
+		}
+	}
```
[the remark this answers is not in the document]
yacovm: So, you're saying that the current code synchronizes all the calls to Timeout and calls them via the handler, while this new code calls each function in its own goroutine, is that it?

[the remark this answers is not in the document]
yacovm: so, I think even without the context lock, this code is wrong.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,87 @@
...
+		close(ps.signal)
+	})
+}
+
+type timeoutHandler struct {
+	newTimer         func(time.Duration) (<-chan time.Time, func() bool)
+	timeouts         chan struct{}
+	onTimeout        func()
+	preemptionSignal <-chan struct{}
+}
```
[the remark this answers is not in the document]
yacovm: I made it this way because I didn't want to change the original code pattern, and to make as minimal changes as possible. 

I now re-implemented it. I hope it reads better and is easier to reason about.

snow/engine/common/timer.go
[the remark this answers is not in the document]
yacovm: I re-implemented this part, and added a lock on the context lock in the timeout callback

snow/engine/snowman/bootstrap/bootstrapper.go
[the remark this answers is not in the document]
yacovm: I'm now locking the context lock in the timeout callback. I am not thrilled about it, but since we already have a few dozens of places where we lock this context lock, i figured it makes sense to proceed than to revert and have a tighter coupling between the engine and the handler via the timeout feedback loop.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+
+// RegisterTimeout fires the function the timeout scheduler is initialized with no later than the given timeout.
+func (th *timeoutScheduler) RegisterTimeout(d time.Duration) {
+	acquiredToken := th.acquirePendingTimeoutToken()
+	preempted := th.preempted()
+
+	if !preempted && !acquiredToken {
+		return
+	}
```
StephenButtolph, line 66: Why is there special handling around the preemption here? Specifically, I'm not sure why we allow multiple goroutines to be scheduled if we have been preempted. I think just having:
```golang
if !th.acquirePendingTimeoutToken() {
	return
}
```
should be sufficient for our expected invariants.

If we did this, I think we could get rid of the `acquiredToken` argument in `scheduleTimeout` and remove the `preempted` helper entirely.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+
+	go th.scheduleTimeout(d, acquiredToken)
+}
+
+func (th *timeoutScheduler) scheduleTimeout(d time.Duration, acquiredToken bool) {
+	timer := th.newTimer(d)
+	defer timer.Stop()
+
+	defer th.onTimeout()
```
StephenButtolph, line 75: nit: I feel like this makes a bit more sense to just be called inline at the end of the function.

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+
+	select {
+	case <-timer.C:
+	case <-th.preemptionSignal:
+	}
+
+	if acquiredToken {
+		th.relinquishPendingTimeoutToken()
```
StephenButtolph, line 83: Can we comment why it is important to return the token prior to calling `onTimeout`?
```
// Allow the next timeout to be scheduled. It is important that this is
// performed prior to invoking the timeout handler to allow the timeout
// handler (or another goroutine synchronized with the timeout handler) to
// schedule a future timeout and be guaranteed it will fire.
```
or something

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+func (th *timeoutScheduler) relinquishPendingTimeoutToken() {
+	th.pendingTimeout <- struct{}{}
+}
+
+// TimeoutRegistrar describes the standard interface for specifying a timeout
+type TimeoutRegistrar interface {
 	// RegisterTimeout specifies how much time to delay the next timeout message
 	// by. If the subnet has been bootstrapped, the timeout will fire
-	// immediately.
+	// immediately via calling Preempt().
```
StephenButtolph, line 113: While we are here, it probably makes sense to expand on the interface comment a bit:
```golang
// RegisterTimeout specifies how much time to delay the next timeout message
// by.
//
// If there is already a pending timeout message, this call is a no-op.
// However, it is guaranteed that the timeout will fire at least once after
// calling this function.
//
// If the subnet has been bootstrapped, the timeout will fire immediately
// via calling Preempt().
```

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+// unless the preemptionSignal is closed and in which case it invokes the function immediately.
+func NewTimeoutScheduler(onTimeout func(), preemptionSignal <-chan struct{}, newTimer func(duration time.Duration) *time.Timer) *timeoutScheduler {
+	pendingTimout := make(chan struct{}, 1)
+	pendingTimout <- struct{}{}
+	return &timeoutScheduler{
+		preemptionSignal: preemptionSignal,
+		newTimer:         newTimer,
+		onTimeout:        onTimeout,
+		pendingTimeout:   pendingTimout,
+	}
```
StephenButtolph, line 56: 🙏 I think this is much cleaner than before - ty

snow/engine/common/timer.go
```diff
@@ -3,12 +3,113 @@
...
+type timeoutScheduler struct {
+	newTimer         func(duration time.Duration) *time.Timer
+	onTimeout        func()
+	preemptionSignal <-chan struct{}
+	pendingTimeout   chan struct{}
+}
+
+// NewTimeoutScheduler constructs a new timeout scheduler with the given function to be invoked upon a timeout,
+// unless the preemptionSignal is closed and in which case it invokes the function immediately.
+func NewTimeoutScheduler(onTimeout func(), preemptionSignal <-chan struct{}, newTimer func(duration time.Duration) *time.Timer) *timeoutScheduler {
```
StephenButtolph, line 48: q: since we only ever specify `newTimer` manually in tests in this package, would it make sense to remove this from the signature (initializing to `time.NewTimer` in this constructer) and then override the `newTimer` variable in the tests?

It seems like that would give a nicer UX for the prod code using this... But I don't have strong opinions here, just a small suggestion.

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+func TestTimeoutSchedulerConcurrentRegister(_ *testing.T) {
+	clock := make(chan time.Time, 2)
+	newTimer := func(_ time.Duration) *time.Timer {
+		// We use a duration of 0 to not leave a lingering timer
+		// after the test finishes.
+		// Then we replace the time channel to have control over the timer.
+		timer := time.NewTimer(0)
+		timer.C = clock
+		return timer
+	}
```
StephenButtolph, line 96: Should we make this a helper (since we define and use it twice?)

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+	ts := NewTimeoutScheduler(onTimeout, roChan, newTimer)
+
+	ts.RegisterTimeout(time.Hour) // First timeout is registered
+	ts.RegisterTimeout(time.Hour) // Second should not
+
+	// Clock ticks are after registering, in order to ensure onTimeout() isn't fired until second registration is invoked.
+	clock <- time.Now()
+	clock <- time.Now()
+
+	wg.Wait()
```
StephenButtolph, line 116: 1. I wasn't sure what `roChan` meant, felt like `preemptChan` was a more clear name.
2. We can just pass `wg.Done` rather than wrapping it into another function.
```suggestion
	var wg sync.WaitGroup
	wg.Add(1)

	preemptChan := make(<-chan struct{})

	ts := NewTimeoutScheduler(wg.Done, preemptChan, newTimer)

	ts.RegisterTimeout(time.Hour) // First timeout is registered
	ts.RegisterTimeout(time.Hour) // Second should not

	// Clock ticks are after registering, in order to ensure wg.Done() isn't fired until second registration is invoked.
	clock <- time.Now()
	clock <- time.Now()

	wg.Wait()
```

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+				ts.RegisterTimeout(time.Hour)
+				<-order
+			}
+
+			wg.Wait()
+		})
+	}
+}
+
+func TestTimeoutSchedulerConcurrentRegister(_ *testing.T) {
```
StephenButtolph, line 87: Should we comment that this test relies on `wg.Done` to panic if the counter goes negative? Typically a test that don't use the `*testing.T` isn't really a test (as it can't fail normally).

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+		initClock               func(chan time.Time)
+		advanceTime             func(chan time.Time)
+	}{
+		{
+			desc:                    "multiple pendingTimeout one after the other with preemption",
+			expectedInvocationCount: 10,
+			shouldPreempt:           true,
+			clock:                   make(chan time.Time, 1),
+			initClock:               func(_ chan time.Time) {},
+			advanceTime:             func(_ chan time.Time) {},
```
StephenButtolph, line 27: Super nit, but typically we don't put the `_` if we don't need to.
```suggestion
			initClock:               func(chan time.Time) {},
			advanceTime:             func(chan time.Time) {},
```

snow/engine/common/timer.go
yacovm: Yes makes sense. That was also how it was in the previous incarnation of this code.

snow/engine/common/timer.go
yacovm: Because the behavior of the previous code scheduled a new goroutine regardless. 
The trick in the bottom select doesn't work in most cases as it's very time reliant - it schedules a goroutine regardless, and then attempts to "cancel it" after the timeout. 

```
func (h *handler) RegisterTimeout(d time.Duration) {
	go func() {
		timer := time.NewTimer(d)
		defer timer.Stop()

		select {
		case <-timer.C:
		case <-h.preemptTimeouts:
		}

		// If there is already a timeout ready to fire - just drop the
		// additional timeout. This ensures that all goroutines that are spawned
		// here are able to close if the chain is shutdown.
		select {
		case h.timeouts <- struct{}{}:
		default:
		}
	}()
}
```

I agree though I should probably take advantage of the opportunity and not replicate a pattern if it does not add any value.

snow/engine/common/timer.go
yacovm: inlined

snow/engine/common/timer_test.go
yacovm: `roChan` means read-only channel. Agree `preemptChan` is better.

yacovm:
@StephenButtolph thanks for the review, addressed your comments.

snow/engine/common/timer.go
StephenButtolph: I think this inlining may need to be pushed from your local branch, it still seems to use the `defer`.

snow/engine/common/timer_test.go
```diff
@@ -0,0 +1,116 @@
...
+			onTimeout := func() {
+				order <- struct{}{}
+				wg.Done()
+				testCase.advanceTime(testCase.clock)
+			}
+
+			ts := NewTimeoutScheduler(onTimeout, ps)
+			ts.newTimer = newTimer
+
+			for i := 0; i < testCase.expectedInvocationCount; i++ {
```
StephenButtolph, line 71: Not asking to change this, but just pointing out that now that we are using go `1.22` this can be written as:
```suggestion
			for range testCase.expectedInvocationCount {
```

StephenButtolph (commented):
final nits then lgtm

snow/engine/common/timer.go
yacovm: oh I thought you meant to inline relinquishing the pending token.

snow/engine/common/timer_test.go
yacovm: I don't want to be the first to break incompatibility with 1.21 :-)

StephenButtolph (approved):

PR #3411, opened 2024-09-24
Bound GetNextStakerChangeTime for ACP-77

StephenButtolph:
## Why this should be merged

For ACP-77, when checking for the maximum allowed block time, we need to search for the first time that a SoV will need to be deactivated. To perform this search efficiently, we should bound that search as much as possible. This PR introduces a maximum time to `GetNextStakerChangeTime` which will be used to implement this efficient search.

## How this works

- Moves the `GetNextStakerChangeTime` verification last in the timestamp checks.
- Unifies the `Banff` block timestamp verification and the `AdvanceTimeTx` verification
  - This does change the verification rule for `AdvanceTimeTx`. However, it makes it less strict, and this transaction is no longer allowed to be issued. So this is a backwards compatible change.
- Introduces hourly polling to the P-chain block building logic.

## How this was tested

 - [x] Added unit tests for `GetNextStakerChangeTime` and `NextBlockTime`

### DIFF
--- a/vms/platformvm/txs/executor/advance_time_test.go
+++ b/vms/platformvm/txs/executor/advance_time_test.go
@@ -126,7 +126,8 @@ func TestAdvanceTimeTxTimestampTooEarly(t *testing.T) {
 	require.ErrorIs(err, ErrChildBlockEarlierThanParent)
 }
 
-// Ensure semantic verification fails when proposed timestamp is after next validator set change time
+// Ensure semantic verification fails when proposed timestamp is after next
+// validator set change time
 func TestAdvanceTimeTxTimestampTooLate(t *testing.T) {
 	require := require.New(t)
 	env := newEnvironment(t, upgradetest.ApricotPhase5)
@@ -167,8 +168,8 @@ func TestAdvanceTimeTxTimestampTooLate(t *testing.T) {
 	env.ctx.Lock.Lock()
 	defer env.ctx.Lock.Unlock()
 
-	// fast forward clock to 10 seconds before genesis validators stop validating
-	env.clk.Set(genesistest.DefaultValidatorEndTime.Add(-10 * time.Second))
+	// fast forward clock to when genesis validators stop validating
+	env.clk.Set(genesistest.DefaultValidatorEndTime)
 
 	{
 		// Proposes advancing timestamp to 1 second after genesis validators stop validating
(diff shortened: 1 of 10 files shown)


vms/platformvm/txs/executor/proposal_tx_executor.go
```diff
@@ -270,34 +269,17 @@ func (e *ProposalTxExecutor) AdvanceTimeTx(tx *txs.AdvanceTimeTx) error {
 		)
 	}
 
-	parentChainTime := e.OnCommitState.GetTimestamp()
-	if !newChainTime.After(parentChainTime) {
```
StephenButtolph, line 274: This used to verify that `parentChainTime < newChainTime`. Now the code verifies that `parentChainTime <= newChainTime`. Technically this verification could be entirely removed as `AdvanceTimeTx`s are now only verified during bootstrapping.

StephenButtolph: In bootstrapping, everything we are verifying has been attested to have been accepted by at least 51% of the network. So, we know that a majority of the network was able to verify and accept this operation. So, actually verifying that it is valid locally is really not needed.

During normal consensus, we need to verify blocks that haven't been accepted yet.. So it's important that we verify every rule condition.

In this case, this transaction was disabled during the Banff upgrade, so any new block containing this transaction would have already errored out from the `e.Config.UpgradeConfig.IsBanffActivated(newChainTime)` check above here... So we know that this code is never hit during normal operations.

yacovm (approved):

vms/platformvm/txs/executor/advance_time_test.go
```diff
@@ -166,8 +168,8 @@ func TestAdvanceTimeTxTimestampTooLate(t *testing.T) {
 	env.ctx.Lock.Lock()
 	defer env.ctx.Lock.Unlock()
 
-	// fast forward clock to 10 seconds before genesis validators stop validating
```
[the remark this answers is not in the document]
StephenButtolph: 10 seconds is the synchrony bound.. So verifying this transaction previously could have failed due to either:
1. The time is being proposed after the `NextStakerChangeTime`
2. The time is being proposed too far ahead of our wall-clock time.

Previously, we check the `NextStakerChangeTime` first, so this test returned that error as expected. However, now we check the proposed time against the wall-clock first... So the wall-clock needed to be updated to pass the wall-clock check.

vms/platformvm/txs/executor/proposal_tx_executor.go
StephenButtolph: This was the minimal diff... It could be removed

yacovm:
> Also, any pointers on how validation rules differ between bootstrapping and participation in consensus?

I think it shouldn't differ because otherwise a node that was offline for a while would fork its state. 

But I think that when you state sync, you just blindly accept the state and not really verify anything. 

@StephenButtolph is that correct?

PR #3412, opened 2024-09-24
ACP-77: Deactivate SoVs without sufficient fees

StephenButtolph:
## Why this should be merged

ACP-77 introduces a new type of validator: the `SubnetOnlyValidator`.

This PR introduces logic to disable this new validator type when a validator no longer has sufficient balance to pay additional fees.

## How this works

1. Block timestamps can not be advanced further than all SoVs have sufficient balance.
2. SoVs that run out of funds due to the timestamp increase are evicted prior to processing transactions
3. After processing transactions, SoVs that can not cover at least 1 seconds worth of fees are removed. (This prevents the chain timestamp from being frozen)

## How this was tested

Added unit tests.

There are no user facing changes.


vms/platformvm/state/chain_time_helpers.go
```diff
@@ -64,42 +75,83 @@ func GetNextStakerChangeTime(state Chain, defaultTime time.Time) (time.Time, err
...
+	}
+	return nextTime, nil
 }
 
 // PickFeeCalculator creates either a static or a dynamic fee calculator,
 // depending on the active upgrade.
 //
 // PickFeeCalculator does not modify [state].
-func PickFeeCalculator(cfg *config.Config, state Chain) fee.Calculator {
+func PickFeeCalculator(config *config.Config, state Chain) txfee.Calculator {
```
StephenButtolph, line 130: `cfg` was changed to `config` to keep consistency with the modified function above (and the rest of the repo)

vms/platformvm/block/executor/verifier.go
```diff
@@ -551,6 +557,57 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
...
+			return nil, nil, nil, err
+		}
+
+		var sovsToDeactivate []state.SubnetOnlyValidator
+		for sovIterator.Next() {
+			sov := sovIterator.Value()
+			// If the validator has exactly the right amount of fee for the next
+			// second we should not remove them here.
+			if sov.EndAccumulatedFee >= potentialAccruedFees {
+				break
```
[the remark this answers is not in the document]
StephenButtolph: According to the interface:

> GetActiveSubnetOnlyValidatorsIterator returns an iterator of all the active subnet-only validators in increasing order of EndAccumulatedFee.

vms/platformvm/state/chain_time_helpers.go
```diff
@@ -64,42 +75,83 @@ func GetNextStakerChangeTime(state Chain, defaultTime time.Time) (time.Time, err
...
 		time := it.Value().NextTime
-		if time.Before(defaultTime) {
-			defaultTime = time
+		if time.Before(nextTime) {
+			nextTime = time
 		}
 	}
-	return defaultTime, nil
+
+	sovIterator, err := state.GetActiveSubnetOnlyValidatorsIterator()
```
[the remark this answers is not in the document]
StephenButtolph: I moved the code into a separate function because I feel like it increases readability... But I feel like the existing tests for this logic are more comprehensive and easier to write than a test for the new helper.

vms/platformvm/txs/executor/state_changes.go
```diff
@@ -257,6 +251,67 @@ func advanceTimeTo(
...
+	// ParentState must not be modified before this iterator is released.
+	sovIterator, err := parentState.GetActiveSubnetOnlyValidatorsIterator()
+	if err != nil {
+		return nil, false, err
+	}
+	defer sovIterator.Release()
+
+	for sovIterator.Next() {
+		sov := sovIterator.Value()
+		if sov.EndAccumulatedFee > accruedFees {
```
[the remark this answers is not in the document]
StephenButtolph: As documented on the interface:
> GetActiveSubnetOnlyValidatorsIterator returns an iterator of all the active subnet-only validators in increasing order of EndAccumulatedFee.

vms/platformvm/block/executor/verifier.go
```diff
@@ -551,6 +557,57 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
...
 
+	// After processing all the transactions, deactivate any SoVs that might not
+	// have sufficient fee to pay for the next second.
+	//
+	// This ensures that SoVs are not undercharged for the next second.
+	if isEtna {
+		var (
+			accruedFees       = diff.GetAccruedFees()
+			validatorFeeState = validatorfee.State{
+				Current: gas.Gas(diff.NumActiveSubnetOnlyValidators()),
```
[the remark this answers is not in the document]
StephenButtolph: The spec is written here: https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#continuous-fee-mechanism. 

Specifically:

> After every second, $x$ will be updated:
> 
> $$x = \max(x + (V - T), 0)$$
> 
> Where:
> 
> - $V$ is the number of active L1 validators
> - $T$ is the target number of active L1 validators

vms/platformvm/block/executor/verifier.go
```diff
@@ -565,5 +569,70 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
...
+		return fmt.Errorf("could not iterate over active SoVs: %w", err)
+	}
+
+	var sovsToDeactivate []state.SubnetOnlyValidator
+	for sovIterator.Next() {
+		sov := sovIterator.Value()
+		// If the validator has exactly the right amount of fee for the next
+		// second we should not remove them here.
+		if sov.EndAccumulatedFee >= potentialAccruedFees {
+			break
```
cam-schultz, line 621: Should this be `continue`?

vms/platformvm/txs/executor/state_changes.go
```diff
@@ -257,10 +251,97 @@ func advanceTimeTo(
 		changes.DeleteExpiry(expiry)
 	}
 
+	if !backend.Config.UpgradeConfig.IsEtnaActivated(newChainTime) {
```
michaelkaplan13, line 254: Can't comment on the line, but the link on line 231 should be updated to: https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#registerl1validatortx

vms/platformvm/txs/executor/state_changes.go
```diff
@@ -257,10 +251,97 @@ func advanceTimeTo(
...
+		changes.SetTimestamp(newChainTime)
+		return changes, changed, nil
```
michaelkaplan13, line 256: Can/should this early return case be moved up above checking the expiries?

StephenButtolph: It can be... I did it this way to avoid needing to modify tests (as unfortunately this part of the code has heavily used mock testing... So non-functional changes can require test modifications.

vms/platformvm/txs/executor/state_changes.go
```diff
@@ -257,10 +251,97 @@ func advanceTimeTo(
...
+	sovIterator, err := parentState.GetActiveSubnetOnlyValidatorsIterator()
+	if err != nil {
+		return false, fmt.Errorf("could not iterate over active SoVs: %w", err)
+	}
+	defer sovIterator.Release()
+
+	var changed bool
+	for sovIterator.Next() {
+		sov := sovIterator.Value()
+		if sov.EndAccumulatedFee > accruedFees {
```
michaelkaplan13, line 327: I'm not 100% following how the math works out such that `EndAccumulatedFee` being:
```
equal to the amount of fees this validator is willing to pay plus the amount of globally accumulated fees when this validator started validating
```
also makes it such that
```
EndAccumulatedFee is the amount of globally accumulated fees that can accrue before this validator must be deactivated.
```

Thinking out loud...if the current global accumulated fee is 1 AVAX, and I register a validator with a balance of 1 AVAX, then my `EndAccumulatedFee` should be 2 AVAX. If advancing time 1 minute would require a cumulative of more than 1 AVAX to be charged across all validators, how is it that we know my validator would need to be charged more than 1 AVAX?

vms/platformvm/block/executor/verifier.go
```diff
@@ -565,5 +569,70 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
 		}
 	}
 
+	// After processing all the transactions, deactivate any SoVs that might not
```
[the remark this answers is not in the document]
StephenButtolph: This is deterministic. However, this check is not _accurate_.

Specifically, it's possible for another block to be issued at the same timestamp. This means that a later block technically could be issued with the same timestamp as this block and the balance of this SoV could be increased (for example).

Ideally this SoV would not have been marked as inactive in this case, because it was always able to pay fees for the time it was in the validator set. But we can't know for certain at this point that this will or will not happen, so we take the conservative approach here and disable the SoV.

vms/platformvm/block/executor/verifier.go
StephenButtolph: No, I think `break` is correct.

This is iterating over `sov`s (which means that `sov.EndAccumulatedFee` is monotonically increasing). Once we find the first SoV that has `sov.EndAccumulatedFee >= potentialAccruedFees`, we know that all following SoVs will have `sov.EndAccumulatedFee >= potentialAccruedFees`.

Technically a `continue` would still work here - but it would mean that we are iterating over SoVs that do not need to be iterated over.

vms/platformvm/block/executor/verifier.go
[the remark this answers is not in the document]
StephenButtolph: Added a comment.

vms/platformvm/block/executor/verifier.go
cam-schultz: Ah, I was missing that the sov's are ordered in that way. I think a comment explaining that any following sov's will also pass this check would be helpful.

StephenButtolph: Added

vms/platformvm/txs/executor/state_changes.go
StephenButtolph: The wording around `globally accumulated` is confusing. `accruedFees` is the fees that would have been paid by a validator that has been active since the Etna activation. _Not_ the total amount of fees that have been charged to all validators for all time.

vms/platformvm/block/executor/verifier.go
```diff
@@ -565,5 +569,73 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
...
+		// EndAccumulatedFee, so we can break early.
+		if sov.EndAccumulatedFee >= potentialAccruedFees {
+			break
+		}
+
+		sovsToDeactivate = append(sovsToDeactivate, sov)
+	}
+
+	// The iterator must be released prior to attempting to write to the
+	// diff.
```
darioush, line 631: nit: Could we also add the reason to the comment?

vms/platformvm/block/executor/verifier.go
```diff
@@ -565,5 +569,73 @@ func (v *verifier) processStandardTxs(txs []*txs.Tx, feeCalculator fee.Calculato
...
+	sovIterator.Release()
+
+	for _, sov := range sovsToDeactivate {
+		sov.EndAccumulatedFee = 0
+		if err := diff.PutSubnetOnlyValidator(sov); err != nil {
+			return fmt.Errorf("could not deactivate SoV %s: %w", sov.ValidationID, err)
```
darioush, line 637: for my understanding, this is an unexpected error?

vms/platformvm/txs/executor/state_changes.go
```diff
@@ -213,52 +221,136 @@ func advanceTimeTo(
...
+		}
+
+		sov.EndAccumulatedFee = 0 // Deactivate the validator
+		if err := changes.PutSubnetOnlyValidator(sov); err != nil {
+			return false, fmt.Errorf("could not deactivate SoV %s: %w", sov.ValidationID, err)
+		}
+		changed = true
+	}
+
+	//
```
darioush, line 349: nit: could we remove this or add the summary of the next lines here?

vms/platformvm/txs/executor/state_changes_test.go
```diff
@@ -214,3 +226,142 @@ func TestAdvanceTimeTo_RemovesStaleExpiries(t *testing.T) {
...
+				sovToKeep,
+			},
+			expectedModified: true,
+			expectedSoVs: []state.SubnetOnlyValidator{
+				sovToKeep,
+			},
+			expectedExcess: 6,
+		},
+		{
+			name: "chooses not to evict",
```
darioush, line 314: nit: not sure what "chooses" refers to here, maybe "no evictions"?

vms/platformvm/txs/executor/state_changes_test.go
```diff
@@ -214,3 +226,142 @@ func TestAdvanceTimeTo_RemovesStaleExpiries(t *testing.T) {
...
+				&Backend{
+					Config: &config,
+				},
+				s,
+				newTime,
+			)
+			require.NoError(err)
+			require.Equal(test.expectedModified, validatorsModified)
+
+			activeSoVs, err := s.GetActiveSubnetOnlyValidatorsIterator()
```
darioush, line 356: should this iterator be released?

michaelkaplan13 (approved):

vms/platformvm/state/chain_time_helpers.go
```diff
@@ -64,42 +75,93 @@ func GetNextStakerChangeTime(state Chain, defaultTime time.Time) (time.Time, err
...
+	// Calculate how many seconds the remaining funds can last for.
+	var (
+		currentTime = state.GetTimestamp()
+		maxSeconds  = uint64(nextTime.Sub(currentTime) / time.Second)
+	)
+	feeState := validatorfee.State{
+		Current: gas.Gas(state.NumActiveSubnetOnlyValidators()),
+		Excess:  state.GetSoVExcess(),
+	}
+	remainingSeconds := feeState.SecondsRemaining(
```
darioush, line 123: Just want to make sure we handle the case where a validator has remainingFunds for a fractional second properly.
It seems SecondsRemaining rounds down where the actual eviction time would be the next second, but I am not sure.

darioush (approved):

vms/platformvm/block/executor/verifier.go
StephenButtolph: Correct, this should never happen.

vms/platformvm/txs/executor/state_changes_test.go
StephenButtolph: `iterator.ToSlice` consumes the iterator (releases it).

vms/platformvm/state/chain_time_helpers.go
StephenButtolph: Added tests for fractional seconds.

vms/platformvm/block/executor/verifier.go
StephenButtolph: I added the comment to the `GetActiveSubnetOnlyValidatorsIterator` definition on the `SubnetOnlyValidators` interface. It is not safe to modify this iterator because it internally uses an `interator.tree` - which has a race condition if the tree is modified during iteration.

felipemadero (approved):

PR #3413, opened 2024-09-24
Set local C-Chain genesis timestamp to InitiallyActiveTime

michaelkaplan13:
## Why this should be merged
- Properly initializes the network upgrades on the C-Chain specified in the genesis
- Fixes https://github.com/ava-labs/coreth/issues/660

## How this works
Sets the local genesis C-Chain timestamp to the [InitiallyActiveTime](https://github.com/ava-labs/avalanchego/blob/master/upgrade/upgrade.go#L16)

## How this was tested
Running node with: `./build/avalanchego --sybil-protection-enabled=false --network-id=local`
And checking that there is a base fee returned for the genesis block (would previously return a null result):
```
$ curl --location 'http://127.0.0.1:9650/ext/bc/C/rpc' \
> --header 'Content-Type: application/json' \
> --data '{
>     "jsonrpc": "2.0",
>     "method": "eth_baseFee",
>     "params": [],
>     "id": 1
> }'
{"jsonrpc":"2.0","id":1,"result":"0x34630b8a00"}
```


genesis/genesis_local.json
```diff
@@ -91,6 +91,6 @@
...
+	"cChainGenesis": "{\"config\":{\"chainId\":43112,\"homesteadBlock\":0,\"daoForkBlock\":0,\"daoForkSupport\":true,\"eip150Block\":0,\"eip150Hash\":\"0x2086799aeebeae135c246c65021c82b4e15a2c451340993aacfd2751886514f0\",\"eip155Block\":0,\"eip158Block\":0,\"byzantiumBlock\":0,\"constantinopleBlock\":0,\"petersburgBlock\":0,\"istanbulBlock\":0,\"muirGlacierBlock\":0,\"apricotPhase1BlockTimestamp\":0,\"apricotPhase2BlockTimestamp\":0},\"nonce\":\"0x0\",\"timestamp\":\"0x5FCB13D0\",\"extraData\":\"0x00\",\"gasLimit\":\"0x5f5e100\",\"difficulty\":\"0x0\",\"mixHash\":\"0x0000000000000000000000000000000000000000000000000000000000000000\",\"coinbase\":\"0x0000000000000000000000000000000000000000\",\"alloc\":{\"8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC\":{\"balance\":\"0x295BE96E64066972000000\"}},\"number\":\"0x0\",\"gasUsed\":\"0x0\",\"parentHash\":\"0x0000000000000000000000000000000000000000000000000000000000000000\"}",
```
yacovm, line 94: do you mind explaining how `0x5FCB13D0` is calculated?

yacovm: Instead of using:

```	
//go:embed genesis_local.json
localGenesisConfigJSON []byte
```

We can just put a Golang object and add an `init()` function which marshals it into the bytes of `localGenesisConfigJSON`

yacovm: > It is [InitiallyActiveTime](https://github.com/ava-labs/avalanchego/blob/master/upgrade/upgrade.go#L16).Unix() represented in hexadecimal to be put in the genesis

It's possible to add a unit test that asserts it is the case.

yacovm (approved):

genesis/genesis_local.json
michaelkaplan13: Decided to leave it with the `//go:embed` approach for now to limit the changes for fixing this issue, but added a unit test that will break if the `InitiallyActiveTime` is ever changes.

StephenButtolph (approved):

PR #3417, opened 2024-09-25
Remove mocked iterator

StephenButtolph:
## Why this should be merged

While not all mocks are bad - these are. 🧹

## How this works

- Uses real iterator implementations rather than mocked implementations.
- Deletes the mocked iterator implementation.

## How this was tested

- [X] CI


ARR4N (approved):
> While not all mocks are bad 

Such diplomacy

michaelkaplan13 (approved):

PR #3421, opened 2024-09-26
ACP-77: Implement `SetSubnetValidatorWeightTx`

StephenButtolph:
## Why this should be merged

This PR introduces the `SetSubnetValidatorWeightTx` defined by [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets#setl1validatorweighttx).

This PR includes:
- `SetSubnetValidatorWeightTx` serialization definition
- `SetSubnetValidatorWeightTx` building and wallet support
- (Basic) complexity calculations for `SetSubnetValidatorWeightTx`
- Execution logic and rules for `SetSubnetValidatorWeightTx`
- Example usage of `SetSubnetValidatorWeightTx`

This PR does not include:
- Warp signature verification. This is added in #3423 
- Finalized ACP-103 complexity values for `SetSubnetValidatorWeightTx`

## How this works

This PR adds the new `SetSubnetValidatorWeightTx` type and implements all the required visitors.

## How this was tested


tests/e2e/p/permissionless_layer_one.go
```diff
@@ -346,6 +348,100 @@ var _ = e2e.DescribePChain("[Permissionless L1]", func() {
...
+			require.NoError(err)
+			require.True(ok)
+
+			tc.By("creating the signed warp message to increase the weight of the validator")
+			signers := set.NewBits()
+			signers.Add(0) // [signers] has weight from the genesis peer
+
+			var sigBytes [bls.SignatureLen]byte
+			copy(sigBytes[:], bls.SignatureToBytes(setSubnetValidatorWeightSignature))
+			registerSubnetValidator, err := warp.NewMessage(
```
darioush, line 388: nit: this name should be updated

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+		// The validator is currently active, we need to refund the remaining
+		// balance.
+		if sov.EndAccumulatedFee != 0 {
+			var remainingBalanceOwner message.PChainOwner
+			if _, err := txs.Codec.Unmarshal(sov.RemainingBalanceOwner, &remainingBalanceOwner); err != nil {
+				return err
+			}
+
+			accruedFees := e.State.GetAccruedFees()
+			if sov.EndAccumulatedFee <= accruedFees {
```
darioush, line 910: Could we expand this comment as to why the equal state is also invalid?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+		return err
+	}
+
+	sov, err := e.State.GetSubnetOnlyValidator(msg.ValidationID)
+	if err != nil {
+		return err
+	}
+	if msg.Nonce < sov.MinNonce {
+		return fmt.Errorf("expected nonce to be at least %d but got %d", sov.MinNonce, msg.Nonce)
+	}
```
darioush, line 876: Do we need to enforce that if `msg.Nonce == MaxUint64` then `msg.Weight == 0`?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+						Addrs:     remainingBalanceOwner.Addresses,
+					},
+				},
+			}
+			e.State.AddUTXO(utxo)
+		}
+	}
+
+	// If the weight is being set to 0, the validator is being removed and the
+	// nonce doesn't matter.
```
darioush, line 939:
```suggestion
	// If the weight is being set to 0, the validator is being removed and the
	// nonce doesn't matter. If not, msg.Nonce is less than MaxUInt64 and
	// can be incremented without overflow.
```

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+		}
+		if weight == sov.Weight {
+			return errRemovingLastValidator
+		}
+
+		// The validator is currently active, we need to refund the remaining
+		// balance.
+		if sov.EndAccumulatedFee != 0 {
+			var remainingBalanceOwner message.PChainOwner
+			if _, err := txs.Codec.Unmarshal(sov.RemainingBalanceOwner, &remainingBalanceOwner); err != nil {
```
cam-schultz, line 905: Just curious, is there a performance benefit to storing the `SubnetOnlyValidator` fields set as raw bytes rather than parsing them on instantiation?

vms/platformvm/txs/executor/standard_tx_executor.go
ceyonur: do we actually need that comment at all?

wallet/subnet/primary/examples/create-chain/main.go
```diff
@@ -22,7 +22,7 @@ func main() {
 	key := genesis.EWOQKey
 	uri := primary.LocalAPIURI
 	kc := secp256k1fx.NewKeychain(key)
-	subnetIDStr := "2DeHa7Qb6sufPkmQcFWG2uCd4pBPv9WB6dkzroiMQhd1NSRtof"
+	subnetIDStr := "2eZYSgCU738xN7aRw47NsBUPqnKkoqJMYUJexTsX19VdTNSZc9"
```
ceyonur, line 25: nit: I think we can just revert this?

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: I think the comment is important to make it clear that the potential overflow is not a problem.

ceyonur: but we still increment it regardless, so it looks a bit like comment is not in sync with the code. Maybe we should emphasize that it does not matter since it's handled in PutSubnetOnlyValidator.

wallet/subnet/primary/examples/set-subnet-validator-weight/main.go
```diff
@@ -0,0 +1,126 @@
...
+	validationID := ids.FromStringOrPanic("2Y3ZZZXxpzm46geqVuqFXeSFVbeKihgrfeXRDaiF4ds6R2N8M5")
+	nonce := uint64(1)
+	weight := uint64(2)
+
+	address, err := hex.DecodeString(addressHex)
+	if err != nil {
+		log.Fatalf("failed to decode address %q: %s\n", addressHex, err)
+	}
+
+	skBytes, err := os.ReadFile("/Users/stephen/.avalanchego/staking/signer.key")
```
yacovm, line 40: not sure we should have your local profile path in avalanchego

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: Sorry I'm not sure why the comment seems to be out of sync with the code. I'm expanding the comment to be:
```golang
// If the weight is being set to 0, the validator is being removed and the
// nonce doesn't matter. If not, [msg.Nonce] is enforced by [msg.Verify()]
// to be less than MaxUInt64 and can therefore be incremented without
// overflow.
```
Does that make sense?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+	addressedCall, err := payload.ParseAddressedCall(warpMessage.Payload)
+	if err != nil {
+		return err
+	}
+
+	msg, err := message.ParseSubnetValidatorWeight(addressedCall.Payload)
+	if err != nil {
+		return err
+	}
+	if err := msg.Verify(); err != nil {
```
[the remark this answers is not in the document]
StephenButtolph: Included this in the comment around overflow below.

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -805,6 +810,146 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+				},
+				Out: &secp256k1fx.TransferOutput{
+					Amt: remainingBalance,
+					OutputOwners: secp256k1fx.OutputOwners{
+						Threshold: remainingBalanceOwner.Threshold,
+						Addrs:     remainingBalanceOwner.Addresses,
+					},
+				},
+			}
+			e.State.AddUTXO(utxo)
```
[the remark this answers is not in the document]
StephenButtolph: All of the writes to the disk are performed atomically upon block acceptance, so the order doesn't matter here.

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: Probably, but I think we'll probably do a pass to remove this behavior and keep them deserialized (as a separate PR)

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: Changed to be:
```golang
// If the weight is being set to 0, it is possible for the nonce increment
// to overflow. However, the validator is being removed and the nonce
// doesn't matter. If weight is not 0, [msg.Nonce] is enforced by
// [msg.Verify()] to be less than MaxUInt64 and can therefore be incremented
// without overflow.
```
to be explicit about the overflow happening.

vms/platformvm/txs/fee/complexity.go
```diff
@@ -205,6 +205,13 @@ var (
 		gas.DBWrite: 0, // TODO
 		gas.Compute: 0, // TODO: Include PoP verification time
 	}
+	IntrinsicSetSubnetValidatorWeightTxComplexities = gas.Dimensions{
+		gas.Bandwidth: IntrinsicBaseTxComplexities[gas.Bandwidth] +
+			wrappers.IntLen, // message length
+		gas.DBRead:  0, // TODO
+		gas.DBWrite: 0, // TODO
+		gas.Compute: 0,
```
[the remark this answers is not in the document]
StephenButtolph: same response - ACP-103 will be finalized in a later PR.

tests/e2e/p/permissionless_layer_one.go
```diff
@@ -346,6 +348,100 @@ var _ = e2e.DescribePChain("[Permissionless L1]", func() {
...
+
+		tc.By("verifying the validator weight was increased", func() {
+			tc.By("verifying the validator set was updated", func() {
+				verifyValidatorSet(map[ids.NodeID]*snowvalidators.GetValidatorOutput{
+					subnetGenesisNode.NodeID: {
+						NodeID:    subnetGenesisNode.NodeID,
+						PublicKey: genesisNodePK,
+						Weight:    genesisWeight,
+					},
+					ids.EmptyNodeID: { // The validator is not active
```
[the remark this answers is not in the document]
StephenButtolph: Added execution tests :+1:

wallet/chain/p/builder/builder.go
```diff
@@ -933,6 +943,56 @@ func (b *builder) NewRegisterSubnetValidatorTx(
...
+		memo           = ops.Memo()
+		memoComplexity = gas.Dimensions{
+			gas.Bandwidth: uint64(len(memo)),
+		}
+	)
+	warpComplexity, err := fee.WarpComplexity(message)
+	if err != nil {
+		return nil, err
+	}
+	complexity, err := fee.IntrinsicRegisterSubnetValidatorTxComplexities.Add(
```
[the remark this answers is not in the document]
StephenButtolph: (Added unit test catches this bug btw)

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -982,37 +975,33 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
 	}
 
+	// Parse the warp message.
 	warpMessage, err := warp.ParseMessage(tx.Message)
 	if err != nil {
 		return err
 	}
-	if warpMessage.NetworkID != e.Ctx.NetworkID {
-		return fmt.Errorf("expected networkID %d but got %d", e.Ctx.NetworkID, warpMessage.NetworkID)
-	}
```
ceyonur: why we don't need this anymore?

StephenButtolph: This PR does not include warp signature verification. This is added in #3423

tests/e2e/p/l1.go
```diff
@@ -365,6 +367,97 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+
+		tc.By("verifying the validator weight was increased", func() {
+			tc.By("verifying the validator set was updated", func() {
+				verifyValidatorSet(map[ids.NodeID]*snowvalidators.GetValidatorOutput{
+					subnetGenesisNode.NodeID: {
+						NodeID:    subnetGenesisNode.NodeID,
+						PublicKey: genesisNodePK,
+						Weight:    genesisWeight,
+					},
+					ids.EmptyNodeID: { // The validator is not active
```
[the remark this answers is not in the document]
StephenButtolph: It's totally valid to modify the weights of non-active validators. Whether that be removing the validator, or just changing its weight (like in this test)

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -959,6 +957,142 @@ func (e *standardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+	)
+	if !upgrades.IsEtnaActivated(currentTimestamp) {
+		return errEtnaUpgradeNotActive
+	}
+
+	if err := e.tx.SyntacticVerify(e.backend.Ctx); err != nil {
+		return err
+	}
+
+	if err := avax.VerifyMemoFieldLength(tx.Memo, true /*=isDurangoActive*/); err != nil {
```
ceyonur, line 973: should this be a part of syntacticverify?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -959,6 +957,142 @@ func (e *standardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+	// Verify that the warp message was sent from the expected chain and
+	// address.
+	if err := verifyL1Conversion(e.state, sov.SubnetID, warpMessage.SourceChainID, addressedCall.SourceAddress); err != nil {
+		return err
+	}
+
+	txID := e.tx.ID()
+
+	// Check if we are removing the validator.
+	if msg.Weight == 0 {
```
ceyonur, line 1031: should we add extra gas if this is removing the validator rather than just changing the weight?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -959,6 +957,142 @@ func (e *standardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
 	return nil
 }
 
+func (e *standardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidatorWeightTx) error {
+	var (
+		currentTimestamp = e.state.GetTimestamp()
+		upgrades         = e.backend.Config.UpgradeConfig
+	)
+	if !upgrades.IsEtnaActivated(currentTimestamp) {
```
michaelkaplan13, line 965: I'm not super familiar with the code, but could the:
```
IsEtnaActivated()
SyntacticVerify()
VerifyMemoFieldLength()
CalculateFee()
VerifySpend()
```
flow be moved to a helper? Seems to be repeated (or at least close to repeated) for multiple tx types here.

michaelkaplan13 (commented):
LGTM, just two questions. One if a few of the standard calls for tx verification could be combined in a helper, the other the same as @ceyonur: will `SetSubnetValidatorWeightTx`s be considered to have the same complexity whether the weight is 0 or not? I guess it would always be charged for the worst case if so?

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: I think there are still a number of outstanding things to do w.r.t. ACP-103. Could we hold this convo until we do a pass of the final tx complexities?

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: Sorry I missed this when I was looking over reviews. I think we are likely to re-add support for memo fields, which is why these checks are in the execution rather than syntactic verification.

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -959,6 +957,142 @@ func (e *standardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
+			e.state.AddUTXO(utxo)
+		}
+	}
+
+	// If the weight is being set to 0, it is possible for the nonce increment
+	// to overflow. However, the validator is being removed and the nonce
+	// doesn't matter. If weight is not 0, [msg.Nonce] is enforced by
+	// [msg.Verify()] to be less than MaxUInt64 and can therefore be incremented
+	// without overflow.
+	sov.MinNonce = msg.Nonce + 1
```
ceyonur, line 1083: does that mean we can increase the MinNonce with the nonce in the msg? I.e it does not need to be a monotonic increase? I guess a higher nonce could invalidate all others?

StephenButtolph: Yes. We explicitly allow this so that consecutive messages can be combined into 1.

For example, say we
1. `SetWeight(nonce=5, weight=10)`
2. `SetWeight(nonce=6, weight=7)`
3. `SetWeight(nonce=7, weight=11)`

Rather than requiring all the messages to be delivered, we can just deliver the last message and the weight will be set correctly.

ceyonur (approved):

ceyonur (approved):

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: I do think this can be cleaned up... But I'd prefer it to be done holistically... Since this doesn't really impact correctness, I'm going to merge this PR and treat this as a followup.

PR #3422, opened 2024-09-26
Fix gossip unit test

joshua-kim:
## Why this should be merged

Fixed an assertion bug in the gossip unit tests

## How this works

We're supposed to assert that we are not gossiping. Clearing the validator set fails because we're not using pointer.

## How this was tested

UT


StephenButtolph (approved):

yacovm (approved):
Alternatively: 

```
	gossiper.Validators = testValidatorSet{}
```

PR #3423, opened 2024-09-26
ACP-77: Implement Warp message verification

StephenButtolph:
## Why this should be merged

This PR introduces warp verification for the `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` defined by [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets).

## How this works

Before block execution, all of the transactions are iterated over and the warp messages are verified for each transaction within the block.

For most transactions, this verification immediately returns a `nil` error.

For the transactions that contain warp messages, the message is verified.

## How this was tested

- [X] Added unit tests for the new verification rules
- [X] The existing E2E tests pass, which include sending transactions with warp messages.


vms/platformvm/block/executor/block.go
```diff
@@ -29,25 +29,49 @@ func (*Block) ShouldVerifyWithContext(context.Context) (bool, error) {
...
+	err := VerifyWarpMessages(
+		ctx,
+		b.manager.ctx.NetworkID,
+		b.manager.ctx.ValidatorState,
+		pChainHeight,
+		b,
+	)
+	if err != nil {
+		return err
 	}
```
StephenButtolph, line 71: Warp messages should probably only be verified after bootstrapping is done (although technically we could verify them during bootstrapping)

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -850,15 +845,10 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
 	if err != nil {
 		return err
 	}
-	if warpMessage.NetworkID != e.Ctx.NetworkID {
```
darioush, line 853: Why are the checks for NetworkID removed?

vms/platformvm/block/executor/manager.go
```diff
@@ -122,6 +123,21 @@ func (m *manager) VerifyTx(tx *txs.Tx) error {
...
 
+	recommendedPChainHeight, err := m.ctx.ValidatorState.GetMinimumHeight(context.TODO())
+	if err != nil {
+		return err
+	}
+	err = executor.VerifyWarpMessages(
+		context.TODO(),
+		m.ctx.NetworkID,
+		m.ctx.ValidatorState,
+		recommendedPChainHeight,
```
darioush, line 134: For my understanding VerifyTx verifies txs before inclusion in the mempool?

vms/platformvm/block/executor/block.go
darioush: Wouldn't it be expensive to calculate the validator sets for old blocks? IMO we should get this in for this PR if I understand correctly.

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: These checks are performed by the warp verification module now.

vms/platformvm/block/executor/manager.go
StephenButtolph: correct

vms/platformvm/block/executor/block.go
StephenButtolph: In this case, it isn't particularly expensive. On something like the C-chain, it is extremely expensive. But because the P-chain is the source of the validator sets, the referenced block height is always going to be close to the locally last accepted block.

darioush: Makes sense, we can add a TODO in the code if we plan to add this later.

darioush (approved):

vms/platformvm/block/builder/builder.go
```diff
@@ -239,13 +245,20 @@ func (b *builder) BuildBlock(context.Context) (snowman.Block, error) {
 		return nil, fmt.Errorf("could not calculate next staker change time: %w", err)
 	}
 
+	var pChainHeight uint64
```
cam-schultz, line 248: I'm assuming `pChainHeight=0` is a valid input to `buildBlock` and subsequent function calls? Is this documented anywhere? I traced the call as far as [BitSetSignature.Verify](https://github.com/ava-labs/avalanchego/blob/f17ea6a7ab4036c41b693e47b94d8f0c81cb69ec/vms/platformvm/warp/signature.go#L70) and wasn't seeing where this case was handled.

vms/platformvm/block/builder/builder.go
```diff
@@ -368,12 +396,14 @@ func buildBlock(
...
 
 func packDurangoBlockTxs(
+	ctx context.Context,
 	parentID ids.ID,
 	parentState state.Chain,
 	mempool mempool.Mempool,
 	backend *txexecutor.Backend,
 	manager blockexecutor.Manager,
 	timestamp time.Time,
+	pChainHeight uint64,
```
michaelkaplan13, line 406: For my own understanding, why are we adding the `pChainHeight` for packing Durango blocks when P-Chain Durango blocks aren't allowed to include ICM messages?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -691,15 +691,10 @@ func (e *StandardTxExecutor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetVal
...
```
michaelkaplan13, line 694: I'm assuming this was only ever added as a stopgap defensive check on the branch prior to ICM verification being added?

vms/platformvm/block/executor/warp_verifier.go
```diff
@@ -0,0 +1,95 @@
...
+	return nil
+}
+
+func (w *warpVerifier) BanffProposalBlock(blk *block.BanffProposalBlock) error {
+	for _, tx := range blk.Transactions {
+		if err := w.verify(tx); err != nil {
+			return err
+		}
+	}
+	return w.ApricotProposalBlock(&blk.ApricotProposalBlock)
```
ceyonur, line 55: Why do we need warpVerifier in banff blocks? Also aren't we repeating the same verification with w.verify in BanffProposalBlock + and one in the ApricotProposalBlock?

vms/platformvm/txs/executor/warp_verifier.go
```diff
@@ -0,0 +1,132 @@
...
+func (w *WarpVerifier) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetValidatorTx) error {
+	return w.verify(tx.Message)
+}
+
+func (w *WarpVerifier) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidatorWeightTx) error {
+	return w.verify(tx.Message)
+}
+
+func (w *WarpVerifier) verify(message []byte) error {
+	msg, err := warp.ParseMessage(message)
```
ceyonur, line 118: Is it concerning that we need to parse this multiple times for different verifiers/executors?

vms/platformvm/block/executor/block.go
```diff
@@ -29,25 +29,49 @@ func (*Block) ShouldVerifyWithContext(context.Context) (bool, error) {
...
 
 	blkID := b.ID()
 	if blkState, ok := b.manager.blkIDToState[blkID]; ok {
 		if !blkState.verifiedHeights.Contains(pChainHeight) {
-			// PlatformVM blocks are currently valid regardless of the ProposerVM's
-			// PChainHeight. If this changes, those validity checks should be done prior
-			// to adding [pChainHeight] to [verifiedHeights].
+			// Only the validity of warp messages need to be verified because
+			// this block has already passed verification with a different
+			// height.
```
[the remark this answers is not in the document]
StephenButtolph: I tried to make the comment a bit more clear with:
```golang
// Only the validity of warp messages need to be verified because
// this block was already executed with a different block context.
```

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: After rebasing, this diff no longer exists :+1:.

vms/platformvm/block/executor/warp_verifier.go
StephenButtolph: > Why do we need warpVerifier in banff blocks?

These are called `Banff` blocks because they were introduced in `Banff`. They are the latest block type (so the P-chain will continue to issue `Banff` blocks after `Etna` is activated).

> Also aren't we repeating the same verification with w.verify in BanffProposalBlock + and one in the ApricotProposalBlock?

`w.verify` takes in a transaction, for a single block, we call `w.verify` for every transaction in the block.

vms/platformvm/txs/executor/warp_verifier.go
StephenButtolph: I was a bit annoyed by this... It isn't a problem... But it is a bit weird.

vms/platformvm/block/executor/block.go
StephenButtolph: Modified the code to handle it.

vms/platformvm/block/builder/builder.go
```diff
@@ -270,59 +283,74 @@ func (b *builder) PackAllBlockTxs() ([]*txs.Tx, error) {
 		return nil, fmt.Errorf("could not calculate next staker change time: %w", err)
 	}
 
+	recommendedPChainHeight, err := b.txExecutorBackend.Ctx.ValidatorState.GetMinimumHeight(context.TODO())
```
ceyonur, line 286: that could be a very dumb question but why don't we use preferredHeight instead?

vms/platformvm/block/executor/block.go
```diff
@@ -29,34 +29,49 @@ func (*Block) ShouldVerifyWithContext(context.Context) (bool, error) {
...
+		if err != nil {
+			return err
 		}
+	}
 
-		// This block has already been verified.
+	// If the block was previously executed, we don't need to execute it again,
+	// we can just mark that the warp messages are valid at this height.
+	if previouslyExecuted {
+		blkState.verifiedHeights.Add(blockContext.PChainHeight)
```
ceyonur, line 55: why we only mark this height as verified if it's previouslyExecuted? Do we expect VerifyWithContext to be always called multiple times? Or do we handle verifiedHeights later at same point?

vms/platformvm/block/builder/builder.go
StephenButtolph: `GetMinimumHeight` is estimating the height that will be used by the proposervm

vms/platformvm/block/builder/builder.go
```diff
@@ -270,59 +283,74 @@ func (b *builder) PackAllBlockTxs() ([]*txs.Tx, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	if !b.txExecutorBackend.Config.UpgradeConfig.IsEtnaActivated(timestamp) {
 		return packDurangoBlockTxs(
+			context.TODO(),
```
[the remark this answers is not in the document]
StephenButtolph: > Is the intention to pass an actual context at some point

Yes

> to limit the scope of execution for a given operation?

The intention is to improve tracing support rather than aborting execution.

> A quick survey suggest that current context usage is a mix of TODO and Background.

Every (real) context starts as a `context.Background()` at some point, additional information gets added to it (like tracing information).

vms/platformvm/txs/executor/warp_verifier.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type warpVerifier struct {
+	context        context.Context
+	networkID      uint32
+	validatorState validators.State
+	pChainHeight   uint64
+}
+
+func (*warpVerifier) AddValidatorTx(*txs.AddValidatorTx) error {
```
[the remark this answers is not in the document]
StephenButtolph: The reason there isn't a default visitor is the same reason we use the visitor pattern at all - to require changes to the visitor interface to be propagated to all the visitors (and not allow a bug to be silently introduced)

vms/platformvm/txs/executor/warp_verifier_test.go
```diff
@@ -0,0 +1,206 @@
...
+		warpSignature,
+	)
+	require.NoError(t, err)
+
+	tests := []struct {
+		name        string
+		tx          txs.UnsignedTx
+		expectedErr error
+	}{
+		{
```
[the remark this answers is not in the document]
StephenButtolph: I didn't really think that this was testing unimplemented code, this is testing that all of these transaction types pass warp verification

ceyonur (approved):

PR #3428, opened 2024-09-28
ACP-77: Implement Warp message signing

StephenButtolph:
## Why this should be merged

This PR introduces P-Chain warp signing for all the messages that the P-Chain is willing to send to subnet managers according to [ACP-77](https://github.com/avalanche-foundation/ACPs/blob/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets/README.md#p-chain-warp-message-payloads). 

## How this works

1. Adds an ACP-118 handler to the P-Chain p2p network.
2. Implements ACP-118 verification for:
  - `SubnetConversion`
  - `SubnetValidatorRegistration`
  - `SubnetValidatorWeight`

## How this was tested

- [X] Added unit tests for the new verification rules
- [X] Added E2E tests for each warp message signing


vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+		return s.verifySubnetValidatorCanNotValidate(msg.ValidationID, preimage.RegisterSubnetValidatorMessage)
+	default:
+		return &common.AppError{
+			Code:    ErrInvalidJustificationType,
+			Message: fmt.Sprintf("invalid justification type: %T", justification.Preimage),
+		}
+	}
+}
+
+// verifySubnetValidatorCanNotValidate verifies that the validationID is
```
michaelkaplan13, line 165:
```suggestion
// verifySubnetValidatorRegistered verifies that the validationID is
```

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return &common.AppError{
+			Code:    common.ErrUndefined.Code,
+			Message: "failed to get subnet only validator: " + err.Error(),
+		}
+	}
+	return nil
+}
+
+// verifySubnetValidatorCanNotValidate verifies that the validationID is not
```
michaelkaplan13, line 190:
```suggestion
// verifySubnetValidatorNotCurrentlyRegistered verifies that the validationID is not
```

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+// currently a validator.
+func (s signatureRequestVerifier) verifySubnetValidatorNotCurrentlyRegistered(
```
michaelkaplan13, line 192: Maybe we should rename this `verifyInitialSubnetValidatorRemoved`? It can't be used to verify any Subnet validator is not currently registered, only those validators added in the `ConverSubnetToL1Tx`.

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+	}
+
+	currentTimeUnix := uint64(s.state.GetTimestamp().Unix())
+	if justification.Expiry <= currentTimeUnix {
+		// The validator is not registered and the expiry time has passed
+		return nil
+	}
+
+	hasExpiry, err := s.state.HasExpiry(state.ExpiryEntry{
```
michaelkaplan13, line 278:
```suggestion
    // If the validation ID was successfully registered and then removed, 
    // then it can never be re-used again even if its expiry has not yet passed.
	hasExpiry, err := s.state.HasExpiry(state.ExpiryEntry{
```

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+		return &common.AppError{
+			Code:    ErrMismatchedValidationID,
+			Message: fmt.Sprintf("validationID %q != justificationID %q", validationID, justificationID),
+		}
+	}
+
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	// Verify that the validator does not and can never exists
```
darioush, line 257: nit
```suggestion
	// Verify that the validator does not and can never exist
```

proto/platformvm/platformvm.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package platformvm;
+
+option go_package = "github.com/ava-labs/avalanchego/proto/pb/platformvm";
+
+message SubnetValidatorRegistrationJustification {
+  oneof preimage {
+    // Validator was added to the Subnet during the ConvertSubnetTx.
```
darioush, line 9:
```suggestion
    // This should be set to obtain attestation that a validator specified in
    // the ConvertSubnetTx is not currently registered.
```

proto/platformvm/platformvm.proto
```diff
@@ -0,0 +1,21 @@
...
+    SubnetIDIndex convert_subnet_tx_data = 1;
+    // Validator was registered to the Subnet after the ConvertSubnetTx.
+    // The SubnetValidator is being removed from the Subnet
```
darioush, line 12:
```suggestion
    // This should be set to obtain attestation that a validator added after
    // the ConvertSubnetTx is not a validator and cannot be added.
    // (i.e., attestation that the validator was successfully added then later
    // removed, or never added before the RegisterSubnetValidatorMessage
    // expired)
```

darioush (approved):

vms/platformvm/network/warp.go
StephenButtolph: I think that `verifySubnetValidatorNotCurrentlyRegistered` is a bit more descriptive of the actual verification that is being performed than `verifyInitialSubnetValidatorRemoved`.

For example, if the `ConverSubnetToL1Tx` only registered 1 validator, then I would expect `verifyInitialSubnetValidatorRemoved` only to allow signing `subnetID.Append(0)`. But this function actually allows signing `subnetID.Append(1)` as well.

@michaelkaplan13 is this behavior acceptable? When I was writing this I had felt like it was similar to the case where a `SubnetValidatorRegistration` with `false` can be sent even though the subnet never requested it.

proto/platformvm/platformvm.proto
```diff
@@ -0,0 +1,21 @@
...
+    bytes register_subnet_validator_message = 2;
+  }
+  bytes filter = 3;
```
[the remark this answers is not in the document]
StephenButtolph: removed

vms/platformvm/network/warp.go
michaelkaplan13: Fair enough. Definitely agree with the behavior, don't love the name but see how my suggestion would be less accurate.

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+
+	ErrImpossibleNonce
+	ErrWrongNonce
+	ErrWrongWeight
+)
+
+var _ acp118.Verifier = (*signatureRequestVerifier)(nil)
+
+type signatureRequestVerifier struct {
+	stateLock sync.Locker
```
yacovm, line 51: same comment - maybe create an interface with only the required methods, and then create a struct that locks only for those, instead of ah-hoc locking in all methods?

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+	subnetID, err := ids.ToID(justification.GetSubnetId())
+	if err != nil {
+		return &common.AppError{
+			Code:    ErrFailedToParseSubnetID,
+			Message: "failed to parse subnetID: " + err.Error(),
+		}
+	}
+
+	justificationID := subnetID.Append(justification.GetIndex())
+	if validationID != justificationID {
```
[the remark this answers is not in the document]
StephenButtolph: The `mismatched convert subnet validationID` test shows why this case is needed.

If this check were not here, then it would be possible to get nodes to sign a warp message saying that a node will not be registered to the validator set, even though it could be.

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+		return &common.AppError{
+			Code:    common.ErrUndefined.Code,
+			Message: "failed to get subnet only validator: " + err.Error(),
+		}
+	case msg.Nonce+1 != sov.MinNonce:
+		return &common.AppError{
+			Code:    ErrWrongNonce,
+			Message: fmt.Sprintf("provided nonce %d != expected nonce (%d - 1)", msg.Nonce, sov.MinNonce),
+		}
+	case msg.Weight != sov.Weight:
```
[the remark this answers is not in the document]
StephenButtolph: Correct

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+		return &common.AppError{
+			Code:    common.ErrUndefined.Code,
+			Message: "failed to lookup subnet only validator: " + err.Error(),
+		}
+	}
+	return nil
+}
+
+// verifySubnetValidatorCanNotValidate verifies that the validationID does not
+// currently and can never become a validator.
```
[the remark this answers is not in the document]
StephenButtolph: changed to:
```
// verifySubnetValidatorCanNotValidate verifies that the validationID is not
// currently and can never become a validator.
```

proto/platformvm/platformvm.proto
StephenButtolph: Did a pass to clarify 👍

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+) *common.AppError {
+	subnetID, err := ids.ToID(justification)
+	if err != nil {
+		return &common.AppError{
+			Code:    ErrFailedToParseJustification,
+			Message: "failed to parse justification: " + err.Error(),
+		}
+	}
+
+	s.stateLock.Lock()
```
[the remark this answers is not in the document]
StephenButtolph: We aren't grabbing a read lock here (but that is really left up to the lock being passed in)

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,336 @@
...
+func (s signatureRequestVerifier) verifySubnetValidatorRegistration(
+	msg *message.SubnetValidatorRegistration,
+	justificationBytes []byte,
+) *common.AppError {
+	if msg.Registered {
+		return s.verifySubnetValidatorRegistered(msg.ValidationID)
+	}
+
+	var justification platformvm.SubnetValidatorRegistrationJustification
+	if err := proto.Unmarshal(justificationBytes, &justification); err != nil {
```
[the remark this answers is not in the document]
StephenButtolph: Proto is used for all messages that are not posted on-chain.

vms/platformvm/vm.go
```diff
@@ -193,6 +193,9 @@ func (vm *VM) Initialize(
 		mempool,
 		txExecutorBackend.Config.PartialSyncPrimaryNetwork,
 		appSender,
+		&chainCtx.Lock,
```
ceyonur, line 196: I should probably have asked [this](https://github.com/ava-labs/avalanchego/pull/3428#discussion_r1831468981) here. 

Should we pass an `RLock` here? Or alternatively pass full RWMutex and do a readlock in warp verifier?

StephenButtolph: It wouldn't be incorrect to pass in `chainCtx.Lock.RLocker()`. But we are pretty liberal with grabbing the write lock in the P-chain, so I felt like this followed the rest of the code...

Looking into the actual state implementation of the methods that are called:
- `GetSubnetConversion`
- `GetSubnetOnlyValidator`
- `GetTimestamp`
- `HasExpiry`

They all seem to be safe to be called with the read lock... We haven't really done a pass to ensure that it's safe to use Rlocks for the getters of the state

ceyonur: I think we should probably start using read locks and do a pass to ensure we're not overusing write locks. Given that we started to use them in network/p-chain APIs etc.

vms/platformvm/warp/message/payload.go
```diff
@@ -43,7 +43,7 @@ func Parse(bytes []byte) (Payload, error) {
 	return p, nil
 }
 
-func initialize(p Payload) error {
+func Initialize(p Payload) error {
```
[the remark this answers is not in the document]
StephenButtolph: Without duplicating the struct definition - I don't think it's possible to implement that interface. If you had an example of how to do this in mind I'd be pretty interested

vms/platformvm/network/network.go
```diff
@@ -156,6 +164,17 @@ func New(
 		return nil, err
 	}
 
+	// We allow all peers to request warp messaging signatures
+	signatureRequestVerifier := signatureRequestVerifier{
+		stateLock: stateLock,
```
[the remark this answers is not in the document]
StephenButtolph: For most of these tests, we need atomic lookups for all of these values. Specifically, if we lookup `GetSubnetOnlyValidator` -> and it is followed by `HasExpiry` (like in `verifySubnetValidatorCanNotValidate`) then we could actually incorrectly sign a message if the validator was added between those 2 calls.

We might be able to re-write the code to handle volatile access... But that seems potentially problematic for maintenance.. I think the alternative would be implementing very specific methods that give all of the combinations of state-lookups that could be needed by the verifier... But I think that would end up looking basically the same as the verifier currently is.

vms/platformvm/vm.go
StephenButtolph: Do you think it makes sense to do this in this PR? or as a followup?

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	// Verify that the provided validationID either:
+	// - Is in the current state
+	// - Was removed from the current state
+	// - Was not included in the subnet conversion
```
michaelkaplan13, line 219: At first pass, this comment confused me because I thought we should be verifying the the validationID is **not** in the current state.

```suggestion
	// Verify that the provided subnetID has been converted 
	// and that the provided validationID is not in the current state.
	// This means the validationID was either not included in the subnet
	// conversion, or that it was removed from the current state.
```

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,358 @@
...
+			Code:    ErrImpossibleNonce,
+			Message: "impossible nonce",
+		}
+	}
+
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	sov, err := s.state.GetSubnetOnlyValidator(msg.ValidationID)
+	switch {
```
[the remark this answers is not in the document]
StephenButtolph: I just thought it was cleaner... I can change it to `if`s if people find it confusing.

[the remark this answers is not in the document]
michaelkaplan13: Nah, just caught my attention.

michaelkaplan13 (approved):

vms/platformvm/network/network.go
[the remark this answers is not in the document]
yacovm:
```
For most of these tests, we need atomic lookups for all of these values. Specifically, if we lookup GetSubnetOnlyValidator -> and it is followed by HasExpiry (like in verifySubnetValidatorCanNotValidate) then we could actually incorrectly sign a message if the validator was added between those 2 calls.
```

Then there is nothing we can do in this PR but to callously drag the context lock here too.

vms/platformvm/vm.go
ceyonur: I think we can change the lock here to readlock and then we can do a followup to clean rest of them

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,356 @@
...
+		return &common.AppError{
+			Code:    common.ErrUndefined.Code,
+			Message: "failed to get subnet only validator: " + err.Error(),
+		}
+	case msg.Nonce+1 != sov.MinNonce:
+		return &common.AppError{
+			Code:    ErrWrongNonce,
+			Message: fmt.Sprintf("provided nonce %d != expected nonce (%d - 1)", msg.Nonce, sov.MinNonce),
+		}
+	case msg.Weight != sov.Weight:
```
ceyonur, line 348: nit: should we check msg.Weight != 0 before? (to shortcircuit fetching from the state)

vms/platformvm/network/warp.go
```diff
@@ -0,0 +1,356 @@
...
+
+	// Verify that the validator does not currently exist
+	_, err = s.state.GetSubnetOnlyValidator(validationID)
+	if err == nil {
+		return &common.AppError{
+			Code:    ErrValidationExists,
+			Message: fmt.Sprintf("validation %q exists", validationID),
+		}
+	}
+	if err != database.ErrNotFound {
```
ceyonur, line 284: do we also need to check subnet conversion here?

vms/platformvm/network/warp.go
StephenButtolph: I had considered doing this... The only reason we check the nonce is to prevent overflow... We could check this, but I felt like it didn't really prevent any DoSes... And it just adds more code.

tests/e2e/p/l1.go
```diff
@@ -223,6 +225,17 @@ var _ = e2e.DescribePChain("[L1]", func() {
 				tc.WithDefaultContext(),
 			)
 			require.NoError(err)
+
+			tc.By("ensuring the genesis peer has accepted the tx at "+subnetGenesisNode.URI, func() {
+				var (
+					client = platformvm.NewClient(subnetGenesisNode.URI)
+					txID   = tx.ID()
+				)
+				require.Eventually(func() bool {
```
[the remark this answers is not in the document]
StephenButtolph: ah - interesting. Thanks for the ref.

tests/e2e/p/l1.go
```diff
@@ -267,6 +280,36 @@ var _ = e2e.DescribePChain("[L1]", func() {
 					},
 				})
 			})
+
+			tc.By("fetching the subnet conversion attestation", func() {
+				unsignedSubnetConversion := must[*warp.UnsignedMessage](tc)(warp.NewUnsignedMessage(
```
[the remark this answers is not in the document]
StephenButtolph: 1. Yeah - I think it makes sense to factor `must` out in a separate PR.
2. The main reason `must` was added was so that initializing more complex types would read more easily.
 
I waffled on the usage of `must` in the case that you described above. _but_ in the case where we do:
```golang
value1, err := make1(
    must[myType](make0(...)),
)
require.NoError(t, err)
```
vs
```golang
value0, err := make0(...)
require.NoError(t, err)

value1, err := make1(value0)
require.NoError(t, err)
```
I think this is nicer - as it reads like nested struct creation (to me).

Not a hill I'll die on though. If other people find it more difficult to read, we can remove it.

vms/platformvm/network/warp.go
ceyonur: I think it should be fine since `hasExpiry` returns false in that case.

vms/platformvm/network/warp.go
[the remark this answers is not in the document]
yacovm: why aren't we grabbing a read lock? I know the Locker is a regular lock interface but why not make a new interface accordingly and grab a read lock?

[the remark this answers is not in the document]
yacovm: After all, all we do with the state is data reads

vms/platformvm/vm.go
StephenButtolph: I changed the locker to be a RLock and documented the functions as being safe for concurrent reads.

vms/platformvm/network/warp.go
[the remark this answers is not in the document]
StephenButtolph: Updated to use the RLock.

vms/platformvm/vm.go
```diff
@@ -193,7 +193,7 @@ func (vm *VM) Initialize(
 		mempool,
 		txExecutorBackend.Config.PartialSyncPrimaryNetwork,
 		appSender,
-		&chainCtx.Lock,
+		chainCtx.Lock.RLocker(),
```
ceyonur, line 196: 🤩

ceyonur (approved):

network/p2p/acp118/handler.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
 )
 
+const HandlerID = p2p.SignatureRequestHandlerID
```
joshua-kim, line 21: Why do we re-define this constant here?

joshua-kim (approved):

PR #3420, opened 2024-09-25
ACP-77: Implement `RegisterSubnetValidatorTx`

StephenButtolph:
## Why this should be merged

This PR introduces the `RegisterSubnetValidatorTx` defined by [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets#registerl1validatortx).

This PR includes:
- `RegisterSubnetValidatorTx` serialization definition
- `RegisterSubnetValidatorTx` building and wallet support
- (Basic) complexity calculations for `RegisterSubnetValidatorTx`
- Execution logic and rules for `RegisterSubnetValidatorTx`
- Example usage of `RegisterSubnetValidatorTx`

This PR does not include:
- Warp signature verification. This is added in #3423 
- Finalized ACP-103 complexity values for `RegisterSubnetValidatorTx`

## How this works

This PR adds the new `RegisterSubnetValidatorTx` type and implements all the required visitors.

## How this was tested

- [X] Added unit tests for all new code
- [X] Added e2e test for new transaction type

PR #3429, opened 2024-09-30
ACP-77: Implement IncreaseBalanceTx

StephenButtolph:
## Why this should be merged

This PR introduces the `IncreaseBalanceTx` defined by [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets#increasel1validatorbalancetx).

This PR includes:
- `IncreaseBalanceTx` serialization definition
- `IncreaseBalanceTx` building and wallet support
- (Basic) complexity calculations for `IncreaseBalanceTx`
- Execution logic and rules for `IncreaseBalanceTx`
- Example usage of `IncreaseBalanceTx`

This PR does not include the finalized ACP-103 complexity values for `IncreaseBalanceTx`

## How this works

This PR adds the new `IncreaseBalanceTx` type and implements all the required visitors.

## How this was tested


vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -933,6 +933,74 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
+
+	sov, err := e.State.GetSubnetOnlyValidator(tx.ValidationID)
+	if err != nil {
+		return err
+	}
+
+	// If the validator is currently inactive, we are activating it.
+	if sov.EndAccumulatedFee == 0 {
+		sov.EndAccumulatedFee = e.State.GetAccruedFees()
+	}
```
StephenButtolph, line 985: Must validate active SoVs

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -940,6 +940,78 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
+
+	sov, err := e.State.GetSubnetOnlyValidator(tx.ValidationID)
+	if err != nil {
+		return err
+	}
+
+	// If the validator is currently inactive, we are activating it.
+	if sov.EndAccumulatedFee == 0 {
+		if gas.Gas(e.State.NumActiveSubnetOnlyValidators()) >= e.Backend.Config.ValidatorFeeConfig.Capacity {
```
darioush, line 991: not sure I understand why `.Capacity` is a gas.Gas type?

wallet/subnet/primary/examples/increase-balance/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ctx := context.Background()
+
+	// MakeWallet fetches the available UTXOs owned by [kc] on the network that
+	// [uri] is hosting and registers [subnetID].
+	walletSyncStartTime := time.Now()
+	wallet, err := primary.MakeWallet(ctx, &primary.WalletConfig{
+		URI:          uri,
+		AVAXKeychain: kc,
+		EthKeychain:  kc,
```
darioush, line 32: same comment as here https://github.com/ava-labs/avalanchego/pull/3420#discussion_r1829802028

darioush (approved):

vms/platformvm/txs/executor/standard_tx_executor.go
yacovm: me neither but i also don't understand what we're comparing here at all - the left side is a number of validators, how can you cast it to a gas? and what's the capacity?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -940,6 +940,78 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
+			return errMaxNumActiveValidators
+		}
+
+		sov.EndAccumulatedFee = e.State.GetAccruedFees()
```
yacovm, line 995: `sov.EndAccumulatedFee = e.State.GetAccruedFees()` - why do we do this? 

In another PR you do:

`remainingBalance := sov.EndAccumulatedFee - accruedFees` 

Why do you set this here to be equal to it?

vms/platformvm/txs/executor/standard_tx_executor.go
darioush: Yes, this enforces the hard cap on the number of active validators as far as I can see

vms/platformvm/txs/executor/standard_tx_executor.go
darioush: when the validator is activated it starts paying fees, having a 0 value in EndAccumulatedFee means the validator is inactive, so a fee paying validator (active validator) balance is equal to `sov.EndAccumulatedFee - e.State.GetAccruedFees()`

StephenButtolph: By storing the `EndAccumulatedFee` for each validator how we do, we are able to charge _every_ validator some amount `x` simply by increasing `e.State.GetAccruedFees()` by `x`. Otherwise we would need to iterate over every validator and explicitly deduct their balance.

yacovm (approved):

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -940,6 +940,78 @@ func (e *StandardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
+	); err != nil {
+		return err
+	}
+
+	sov, err := e.State.GetSubnetOnlyValidator(tx.ValidationID)
+	if err != nil {
+		return err
+	}
+
+	// If the validator is currently inactive, we are activating it.
```
ceyonur, line 989: should this cost extra gas for reactivations vs balance topping (in still active state)?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -1093,6 +1093,78 @@ func (e *standardTxExecutor) SetSubnetValidatorWeightTx(tx *txs.SetSubnetValidat
...
+		return errEtnaUpgradeNotActive
+	}
+
+	if err := e.tx.SyntacticVerify(e.backend.Ctx); err != nil {
+		return err
+	}
+
+	if err := avax.VerifyMemoFieldLength(tx.Memo, true /*=isDurangoActive*/); err != nil {
+		return err
+	}
```
ceyonur, line 1111: same question as https://github.com/ava-labs/avalanchego/pull/3421#discussion_r1835653120

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: This is kind of a chaining of the same reasoning over and over... We could change it if we wanted to...
- `Capacity` was chosen to be of `gas.Gas` so that it would have the same type as `Target`.
- `Target` was chosen to be of `gas.Gas` so that it would have the same type as `Excess`.
- `Excess` and `ExcessConversionConstant` were chosen to be of `gas.Gas` because the validator pricing is performed using the same calculations as ACP-103.

We could technically have `Capacity` and `Target` be specified as `int` for easier usage and have the casting be performed as an implementation detail of the fees package... Although then we might want to at least make them `uint` so that it is clear the casting can't overflow....

I felt like this was the cleanest... But open to suggestions (as a separate PR) to cleanup the types.

ceyonur (approved):

PR #3434, opened 2024-10-01
ACP-118: Implement p2p handler

joshua-kim:
## Why this should be merged

Refactors handler implementation out of https://github.com/ava-labs/avalanchego/pull/3394

## How this works

Adds handler with a server-specified `Attestor` that defines how whether to sign messages

## How this was tested

UT


network/acp118/handler.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/ava-labs/avalanchego/proto/pb/sdk"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+)
+
+var _ p2p.Handler = (*Handler)(nil)
+
+// Attestor defines whether to a warp message payload should be attested to
+type Attestor interface {
+	Attest(message *warp.UnsignedMessage, justification []byte) (bool, *common.AppError)
```
StephenButtolph, line 24: Is the `bool` really needed? I feel like just having `*common.AppError` would be a cleaner implementation. (`nil` meaning that signing is allowed)

network/acp118/handler.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/proto/pb/sdk"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+)
+
+var _ p2p.Handler = (*Handler)(nil)
+
+// Attestor defines whether to a warp message payload should be attested to
+type Attestor interface {
```
joshua-kim, line 23: Not sure if there's a better name for this interface... the other idea was `Signer`.

network/acp118/handler.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if !ok {
+		return nil, p2p.ErrAttestFailed
+	}
+
+	signature, err := h.signer.Sign(msg)
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    p2p.ErrUnexpected.Code,
```
StephenButtolph, line 77: Would it make sense to special-case `warp.ErrWrongSourceChainID` and `warp.ErrWrongNetworkID` for specific error codes?

This would only work for avalanchego VMs as those errors aren't communicated over the rpcchainvm... So I see reasons both to and not to. 🤷

network/acp118/handler.go
joshua-kim: In the original pass i was returning `(bool, error)`. I felt like it was weird to return an error for expected behavior, but I guess we're going to send back an `AppError` in either scenario.... so maybe it does make sense to drop the `bool`.

network/acp118/handler.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/ava-labs/avalanchego/proto/pb/sdk"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+)
+
+var _ p2p.Handler = (*Handler)(nil)
+
+// Attestor defines whether to a warp message payload should be attested to
+type Attestor interface {
+	Attest(message *warp.UnsignedMessage, justification []byte) (bool, *common.AppError)
```
StephenButtolph, line 24: I think we should pass the `context.Context` through into this function (for tracing)

network/acp118/handler.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, &common.AppError{
+			Code:    p2p.ErrUnexpected.Code,
+			Message: fmt.Sprintf("failed to unmarshal request: %s", err),
+		}
+	}
+
+	msg, err := warp.ParseUnsignedMessage(request.Message)
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    p2p.ErrUnexpected.Code,
```
StephenButtolph, line 60: Does it make sense to expose specific errors for these? This isn't really an unexpected error... these can definitely happen

network/acp118/handler.go
StephenButtolph: `acp118.Verifier`?

This is verifying that the message should be signed... So that would make sense to me

network/acp118/handler.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/proto/pb/sdk"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+)
+
+var _ p2p.Handler = (*Handler)(nil)
+
+// Verifier defines whether a warp message payload should be verified
```
StephenButtolph, line 22:
```suggestion
// Verifier verifies that a warp message should be signed
```

network/acp118/handler.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package acp118
```
StephenButtolph, line 4: Should this package live in `network/p2p` rather than just `network`? It seems fairly similar to `gossip`

network/acp118/handler.go
joshua-kim: Diff outdated because of file move so applied manually

network/p2p/acp118/handler.go
```diff
@@ -0,0 +1,94 @@
...
+			Code:    p2p.ErrUnexpected.Code,
+			Message: fmt.Sprintf("failed to unmarshal request: %s", err),
+		}
+	}
+
+	msg, err := warp.ParseUnsignedMessage(request.Message)
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    p2p.ErrUnexpected.Code,
+			Message: fmt.Sprintf("failed to initialize warp unsigned message: %s", err),
```
StephenButtolph, line 65:
```suggestion
			Message: fmt.Sprintf("failed to parse warp unsigned message: %s", err),
```

network/p2p/acp118/handler_test.go
```diff
@@ -0,0 +1,115 @@
...
+	tests := []struct {
+		name           string
+		verifier       Verifier
+		expectedErr    error
+		expectedVerify bool
+	}{
+		{
+			name:        "signature fails verification",
+			verifier:    &testVerifier{Err: &common.AppError{Code: int32(123)}},
+			expectedErr: &common.AppError{Code: int32(123)},
```
StephenButtolph, line 34:
```suggestion
			verifier:    &testVerifier{Err: &common.AppError{Code: 123}},
			expectedErr: &common.AppError{Code: 123},
```

network/p2p/acp118/handler_test.go
```diff
@@ -0,0 +1,115 @@
...
+				require.Equal(tt.expectedVerify, bls.Verify(pk, signature, request.Message))
+			}
+
+			require.NoError(c.AppRequest(ctx, set.Of(clientNodeID), requestBytes, onResponse))
+			<-done
+		})
+	}
+}
+
+// The zero value of testVerifier verifies
```
StephenButtolph, line 104:
```suggestion
// The zero value of testVerifier allows signing
```

network/acp118/handler.go
StephenButtolph: Seems like the app error codes are a bit inconsistent rn. Will take this convo out of this PR.

network/acp118/handler.go
StephenButtolph: Seems like the app error codes are a bit inconsistent rn. Will take this convo out of this PR.

StephenButtolph (approved):

darioush (approved):

iansuvak (approved):

network/acp118/handler.go
joshua-kim: Followup: https://github.com/ava-labs/avalanchego/issues/3438

PR #3440, opened 2024-10-02
ACP-77: Implement DisableSubnetValidatorTx

StephenButtolph:
## Why this should be merged

This PR introduces the `DisableSubnetValidatorTx` defined by [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets#disablel1validatortx).

This PR includes:
- `DisableSubnetValidatorTx` serialization definition
- `DisableSubnetValidatorTx` building and wallet support
- (Basic) complexity calculations for `DisableSubnetValidatorTx`
- Execution logic and rules for `DisableSubnetValidatorTx`
- Example usage of `DisableSubnetValidatorTx`

This PR does not include the finalized ACP-103 complexity values for `DisableSubnetValidatorTx`

## How this works

This PR adds the new `DisableSubnetValidatorTx` type and implements all the required visitors.

## How this was tested


vms/platformvm/txs/disable_subnet_validator_tx.go
```diff
@@ -0,0 +1,45 @@
...
+func (tx *DisableSubnetValidatorTx) SyntacticVerify(ctx *snow.Context) error {
+	switch {
+	case tx == nil:
+		return ErrNilTx
+	case tx.SyntacticallyVerified:
+		// already passed syntactic verification
+		return nil
+	}
+
+	if err := tx.BaseTx.SyntacticVerify(ctx); err != nil {
```
yacovm, line 32: `BaseTx.SyntacticVerify` doesn't turn on the `SyntacticallyVerified` of itself. 
Since we have syntactically verified the baseTx, shall we activate it here?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -1012,6 +1012,114 @@ func (e *StandardTxExecutor) IncreaseBalanceTx(tx *txs.IncreaseBalanceTx) error
...
+	}
+
+	var remainingBalanceOwner message.PChainOwner
+	if _, err := txs.Codec.Unmarshal(sov.RemainingBalanceOwner, &remainingBalanceOwner); err != nil {
+		return err
+	}
+
+	accruedFees := e.State.GetAccruedFees()
+	if sov.EndAccumulatedFee <= accruedFees {
+		// This check should be unreachable. However, including it ensures
```
yacovm, line 1093: well you should have this check also because of the next line (1098)

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -1012,6 +1012,114 @@ func (e *StandardTxExecutor) IncreaseBalanceTx(tx *txs.IncreaseBalanceTx) error
...
+	avax.Consume(e.State, tx.Ins)
+	// Produce the UTXOS
+	avax.Produce(e.State, txID, tx.Outs)
+
+	// If the validator is already disabled, there is nothing to do.
+	if sov.EndAccumulatedFee == 0 {
+		return nil
+	}
+
+	var remainingBalanceOwner message.PChainOwner
```
yacovm, line 1086: Why is this raw bytes if it's a field of the SoV?

vms/platformvm/txs/executor/standard_tx_executor.go
```diff
@@ -1012,6 +1012,114 @@ func (e *StandardTxExecutor) IncreaseBalanceTx(tx *txs.IncreaseBalanceTx) error
...
+	if _, err := txs.Codec.Unmarshal(sov.RemainingBalanceOwner, &remainingBalanceOwner); err != nil {
+		return err
+	}
+
+	accruedFees := e.State.GetAccruedFees()
```
yacovm, line 1091: the `GetAccruedFees` is the fees only for the SoV? why?

wallet/chain/p/builder/builder.go
```diff
@@ -1053,6 +1062,63 @@ func (b *builder) NewIncreaseBalanceTx(
...
+	memo := ops.Memo()
+	memoComplexity := gas.Dimensions{
+		gas.Bandwidth: uint64(len(memo)),
+	}
+	authComplexity, err := fee.AuthComplexity(disableAuth)
+	if err != nil {
+		return nil, err
+	}
+
+	complexity, err := fee.IntrinsicDisableSubnetValidatorTxComplexities.Add(
```
yacovm, line 1088: this is safe because the receiver is not a pointer, right?

wallet/chain/p/wallet/backend_visitor.go
```diff
@@ -63,18 +67,47 @@ func (b *backendVisitor) RemoveSubnetValidatorTx(tx *txs.RemoveSubnetValidatorTx
...
-	b.b.setSubnetOwner(
+	b.b.setOwner(
 		tx.Subnet,
 		tx.Owner,
 	)
 	return b.baseTx(&tx.BaseTx)
 }
 
 func (b *backendVisitor) ConvertSubnetTx(tx *txs.ConvertSubnetTx) error {
+	for i, vdr := range tx.Validators {
```
yacovm, line 78: can you explain what we're doing here?

yacovm: nvm i think i got it

wallet/chain/p/wallet/backend_visitor.go
```diff
@@ -63,18 +67,47 @@ func (b *backendVisitor) RemoveSubnetValidatorTx(tx *txs.RemoveSubnetValidatorTx
...
+				Threshold: vdr.DeactivationOwner.Threshold,
+				Addrs:     vdr.DeactivationOwner.Addresses,
+			},
+		)
+	}
 	return b.baseTx(&tx.BaseTx)
 }
 
 func (b *backendVisitor) RegisterSubnetValidatorTx(tx *txs.RegisterSubnetValidatorTx) error {
+	warpMessage, err := warp.ParseMessage(tx.Message)
```
yacovm, line 91: where do we validate/verify the Warp message?

wallet/subnet/primary/examples/disable-subnet-validator/main.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary"
+)
+
+func main() {
+	key := genesis.EWOQKey
+	uri := primary.LocalAPIURI
+	kc := secp256k1fx.NewKeychain(key)
+	validationID := ids.FromStringOrPanic("9FAftNgNBrzHUMMApsSyV6RcFiL9UmCbvsCu28xdLV2mQ7CMo")
```
yacovm, line 21: where is this ID taken from?

wallet/chain/p/wallet/backend_visitor.go
```diff
@@ -63,18 +67,47 @@ func (b *backendVisitor) RemoveSubnetValidatorTx(tx *txs.RemoveSubnetValidatorTx
...
 		tx.Subnet,
 		tx.Owner,
 	)
 	return b.baseTx(&tx.BaseTx)
 }
 
 func (b *backendVisitor) ConvertSubnetTx(tx *txs.ConvertSubnetTx) error {
+	for i, vdr := range tx.Validators {
+		b.b.setOwner(
+			tx.Subnet.Prefix(uint64(i)),
```
darioush, line 80: Should this be Append? (also should we add a name like `validationID` for this?)

darioush (approved):

wallet/chain/p/builder/builder.go
StephenButtolph: correct

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: This check should be impossible to be reached.

We already know that `sov.EndAccumulatedFee != 0` (as otherwise the function would have already returned).

If `sov.EndAccumulatedFee < accruedFees`, then the balance on the SoV is negative, which should clearly never happen.

If `sov.EndAccumulatedFee == accruedFees` then there must have been a bug in `AdvanceTimeTo`, as we should have set the `EndAccumulatedFee` to be equal to `0` there already.

The only reason this check is here is as a last-chance verification to ensure that a bug in execution doesn't cause the AVAX balance to underflow (which could be catastrophic)

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: This is a fair question... I had rationalized that this would be more efficient because we often won't need these values to be parsed... But it does make the usage more annoying. I'll pull this out into a separate PR to change the `SubnetOnlyValidator` type accordingly.

vms/platformvm/txs/executor/standard_tx_executor.go
StephenButtolph: Sorry I don't think I fully understand your question. Does [this](https://github.com/ava-labs/avalanchego/pull/3429#discussion_r1830454985) reply clarify things?

wallet/chain/p/builder/builder.go
StephenButtolph: yes

wallet/chain/p/wallet/backend_visitor.go
StephenButtolph: In the wallet, transactions aren't verified (it is assumed that the API the wallet is interacting with is virtuous).

At this point in the wallet code, the API has told us that the transaction was confirmed.

As far as where Warp message verification happens in the node, it is added in this PR: #3423

wallet/subnet/primary/examples/disable-subnet-validator/main.go
StephenButtolph: This is just an example, so running this against an actual network would likely need to change this value.

This value was taken from the result of either running the `convert-subnet` example or the `register-subnet-validator` example.

wallet/chain/p/wallet/backend_visitor.go
StephenButtolph: Yes - this is incorrect.

vms/platformvm/txs/executor/standard_tx_executor.go
yacovm: yeah

vms/platformvm/txs/executor/standard_tx_executor.go
yacovm: > The only reason this check is here is as a last-chance verification to ensure that a bug in execution doesn't cause the AVAX balance to underflow (which could be catastrophic)

Yes that's what I meant.

wallet/chain/p/builder/builder.go
```diff
@@ -1813,12 +1879,12 @@ func (b *builder) spend(
 	return s.inputs, s.changeOutputs, s.stakeOutputs, nil
 }
 
-func (b *builder) authorizeSubnet(subnetID ids.ID, options *common.Options) (*secp256k1fx.Input, error) {
-	ownerIntf, err := b.backend.GetSubnetOwner(options.Context(), subnetID)
+func (b *builder) authorize(ownerID ids.ID, options *common.Options) (*secp256k1fx.Input, error) {
```
ceyonur, line 1882: why we renamed it to ownerID?

wallet/chain/p/wallet/backend_visitor.go
ceyonur: I'm also a bit confused with Prefix. I also thought it should be Append.

wallet/chain/p/builder/builder.go
StephenButtolph: because it can either be a `subnetID` or a `validationID`

ceyonur: hmm seems we have commented it 👍

ceyonur (approved):

wallet/chain/p/wallet/backend_visitor.go
StephenButtolph: Updated

PR #3443, opened 2024-10-03
[testing] Fix bootstrap monitor e2e image build error check

marun:
testify was panicking on the previous error check which was preventing diagnosis of the underlying cause of image build flakes.


StephenButtolph (approved):
Not sure this is going to fix the issue - but the code looks better this way

yacovm (approved):

PR #3446, opened 2024-10-07
[tmpnet] Refactor bootstrap monitor kubernetes functions for reuse

marun:
## Why this should be merged

The bootstrap monitor added support for deployment of avalanchego nodes to kubernetes to support e2e testing. Moving/refactoring this to tmpnet can enable reuse of the same functionality in deploying tmpnet networks to kubernetes.

## How this works

Moves reusable functions for interacting with kubernetes from the bootstrap monitor to tmpnet package.

## How this was tested

CI

### DIFF
--- a/tests/fixture/tmpnet/kube.go
+++ b/tests/fixture/tmpnet/kube.go
@@ -237,7 +237,7 @@ func WaitForPodStatus(
 				return nil
 			}
 		case <-ctx.Done():
-			return errors.New("timeout waiting for pod readiness")
+			return fmt.Errorf("timeout waiting for pod readiness: %w", ctx.Err())
 		}
 	}
 }


tests/fixture/tmpnet/kube.go
```diff
@@ -0,0 +1,299 @@
...
+			pod, ok := event.Object.(*corev1.Pod)
+			if !ok {
+				continue
+			}
+
+			if acceptable(&pod.Status) {
+				return nil
+			}
+		case <-ctx.Done():
+			return errors.New("timeout waiting for pod readiness")
```
StephenButtolph, line 240: We used to return `ctx.Err()` here. Why aren't we anymore?
```suggestion
			return fmt.Errorf("timeout waiting for pod readiness: %w", ctx.Err())
```

StephenButtolph (approved):

PR #3447, opened 2024-10-07
ACP-77: Expand registered bool into registration status

StephenButtolph:
## Why this should be merged

Currently it is not possible to evict SoVs that were registered from the `ConvertSubnetTx`. This expands the `Registered` bool into an `Enum` that can be used to remove validators that were known to previously be validators.

## How this works

- `bool` -> `enum`
- Implements json marshalling of the enum

## How this was tested

- [X] CI


StephenButtolph:
Going to do a pass to see if we can expand the justification field to not include changing this warp message

StephenButtolph:
Decided to just change the justification.

PR #3452, opened 2024-10-09
Remove minimum peer error in health check

sukantoraymond:
## Why this should be merged

When setting up private network (e.g. Local Devnet), when there is only 1 validator node, node will remain unhealthy with the message "not connected to a minimum of %d peer(s) only". This PR removes this error message.

## How this works
NA

## How this was tested
Ran a local Devnet with this version of avalancheGo with 1 Validator only and node passes health check

### DIFF
--- a/.github/workflows/build-macos-release.yml
+++ b/.github/workflows/build-macos-release.yml
@@ -18,7 +18,7 @@ jobs:
   # This workflow contains a single job called "build"
   build-mac:
     # The type of runner that the job will run on
-    runs-on: macos-12
+    runs-on: macos-14
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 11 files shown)


network/network.go
```diff
@@ -410,9 +410,6 @@ func (n *network) HealthCheck(context.Context) (interface{}, error) {
 	}
 
 	var errorReasons []string
-	if !isConnected {
-		errorReasons = append(errorReasons, fmt.Sprintf("not connected to a minimum of %d peer(s) only %d", n.config.HealthConfig.MinConnectedPeers, connectedTo))
-	}
```
StephenButtolph, line 415: I think we need to remove more here to actually make the node report healthy.

PR #3454, opened 2024-10-09
Add Etna logging

StephenButtolph:
## Why this should be merged

Because I worked hard on it.

## How this works

Displays the log after accepting the first block that was accepted after Etna was activated (and suppresses the log if that block was accepted during bootstrapping or if the initial state was after Etna)

## How this was tested

N/A


darioush (approved):

ceyonur (approved):
LGTM 
```
  ,_ o
  / //\,
   \>> |
    \\
```

iansuvak (approved):

PR #3455, opened 2024-10-09
Remove unexpected block unwrapping

StephenButtolph:
## Why this should be merged

Currently we actually pass `ApricotProposalBlock` and `ApricotStandardBlock`s into `state.AddStatelessBlock` upon accept. This ends up not being a bug because the `Banff.*Block`s serialized format is still returned from `.Bytes()` and the parsed version of the block is not cached.

However, this is a very weird flow that seems very close to causing a bug.

## How this works

Passes the wrapped block types into the helper functions.

## How this was tested

- [X] Added additional test coverage that fails against master.

### DIFF
--- a/vms/platformvm/block/executor/verifier_test.go
+++ b/vms/platformvm/block/executor/verifier_test.go
@@ -1184,8 +1184,9 @@ func TestBlockExecutionWithComplexity(t *testing.T) {
 			}
 
 			require.Contains(verifier.blkIDToState, blkID)
-			onAcceptState := verifier.blkIDToState[blkID].onAcceptState
-			require.Equal(test.expectedFeeState, onAcceptState.GetFeeState())
+			blockState := verifier.blkIDToState[blkID]
+			require.Equal(blk, blockState.statelessBlock)
+			require.Equal(test.expectedFeeState, blockState.onAcceptState.GetFeeState())
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


vms/platformvm/block/executor/verifier_test.go
```diff
@@ -1184,8 +1184,9 @@ func TestBlockExecutionWithComplexity(t *testing.T) {
 			}
 
 			require.Contains(verifier.blkIDToState, blkID)
-			onAcceptState := verifier.blkIDToState[blkID].onAcceptState
-			require.Equal(test.expectedFeeState, onAcceptState.GetFeeState())
+			blockState := verifier.blkIDToState[blkID]
+			require.Equal(blk, blockState.statelessBlock)
```
ARR4N, line 1188: I don't know enough about what's going on to give a thoughtful review of all this. Would this have otherwise failed?

I'm looking for something in the tests that indicates to someone knowledgable about this code that there's been a proper fix here.

StephenButtolph: > Would this have otherwise failed?

Yes.

`require.Equal(blk, blockState.statelessBlock)` fails on master and passes on this branch.

`require.Equal(&blk.ApricotStandardBlock, blockState.statelessBlock)` would have passed on master.

ARR4N (approved):

vms/platformvm/block/executor/verifier_test.go
ARR4N: In that case, LGTM

> [!CAUTION]
> Once you start using GitHub Markdown alerts, you can't stop.

PR #3458, opened 2024-10-10
Add Etna P-chain metrics

StephenButtolph:
## Why this should be merged

Adds useful metrics:
```
# HELP avalanche_platformvm_gas_consumed Cumulative amount of gas consumed by transactions
# TYPE avalanche_platformvm_gas_consumed counter
avalanche_platformvm_gas_consumed{chain="P"} 482042
# HELP avalanche_platformvm_gas_capacity Minimum amount of gas that can be consumed in the next block
# TYPE avalanche_platformvm_gas_capacity gauge
avalanche_platformvm_gas_capacity{chain="P"} 999333
# HELP avalanche_platformvm_active_sovs Number of active Subnet only Validators
# TYPE avalanche_platformvm_active_sovs gauge
avalanche_platformvm_active_sovs{chain="P"} 9
# HELP avalanche_platformvm_excess Excess usage of a resource over the target usage
# TYPE avalanche_platformvm_excess gauge
avalanche_platformvm_excess{chain="P",resource="gas"} 667
avalanche_platformvm_excess{chain="P",resource="validators"} 0
# HELP avalanche_platformvm_price Price (in nAVAX) of a resource
# TYPE avalanche_platformvm_price gauge
avalanche_platformvm_price{chain="P",resource="gas"} 1
avalanche_platformvm_price{chain="P",resource="validators"} 1
```

## How this works

Adds a `metrics.Block` that is passed into `metrics.Metrics#MarkAccepted` which includes useful information related to the Etna state.

## How this was tested

- [X] Ran node on the devnet and metrics are exposed as expected.


vms/platformvm/metrics/metrics.go
```diff
@@ -106,19 +169,36 @@ type metrics struct {
...
 	validatorSetsCreated    prometheus.Counter
 	validatorSetsHeightDiff prometheus.Gauge
 	validatorSetsDuration   prometheus.Gauge
 }
 
-func (m *metrics) MarkAccepted(b block.Block) error {
-	return b.Visit(m.blockMetrics)
+func (m *metrics) MarkAccepted(b Block) error {
+	m.gasConsumed.Add(float64(b.GasConsumed))
+	m.gasCapacity.Add(float64(b.GasState.Capacity))
```
StephenButtolph, line 193:
```suggestion
	m.gasCapacity.Set(float64(b.GasState.Capacity))
```

vms/platformvm/block/executor/acceptor_test.go
```diff
@@ -113,12 +113,6 @@ func TestAcceptorVisitAtomicBlock(t *testing.T) {
 	)
 	require.NoError(err)
 
-	// Set expected calls on the state.
-	// We should error after [commonAccept] is called.
-	s.EXPECT().SetLastAccepted(blk.ID()).Times(1)
-	s.EXPECT().SetHeight(blk.Height()).Times(1)
-	s.EXPECT().AddStatelessBlock(blk).Times(1)
```
darioush, line 120: assuming these changes are only relevant in the error case, therefore the behavior does not change overall

darioush (approved):

vms/platformvm/block/executor/acceptor_test.go
StephenButtolph: Yeah, this test is now failing sooner than it was before, so we are no longer expecting to hit these calls in this case.

These calls still exist in the happy path where the block acceptance succeeds.

PR #3459, opened 2024-10-10
[ACP 151] Use current block's P-Chain height as context for verifying state of the inner block

iansuvak:
## Why this should be merged
Implements [ACP 151](https://github.com/avalanche-foundation/ACPs/pull/151) to improve UX of PChain and L1 interop for flows defined in ACP-77.

## How this works

Passes in current block's P-Chain height into `block.Context`'s `PChainHeight` field instead of the parent if Etna is activated. 

## How this was tested
Existing test mock expect was updated to reflect new behavior and a new test case for Pre-Etna behavior on the build side was added.

TODO: add testing for pre-etna verify


StephenButtolph (approved):

michaelkaplan13 (approved):
👍 LGTM.

We should link this PR as the reference implementation in the ACP and mark it as implementable also.

PR #3460, opened 2024-10-10
Evict recentlyAccepted blocks based on wall-clock time 

iansuvak:
## Why this should be merged

In combination with [ACP-115](https://github.com/avalanche-foundation/ACPs/pull/151) this would remove requirements for issuing dummy transaction on P-Chain for ACP-77 flows on networks that don't have frequent P-Chain block production like Fuji and Devnet. 

## How this works

Changes `RWLock` to `Lock` and evicts nodes based on wall-clock time on all operations. 
The `Window` is used here in [validators.Manager](https://github.com/ava-labs/avalanchego/blob/use-current-block-height/vms/platformvm/validators/manager.go#L131) and the change affects the output of `Manager.GetMinimumHeight`

## How this was tested
Modified existing test cases from negative to positive cases since the behavior of the window is changing here.


utils/window/window_test.go
```diff
@@ -275,3 +277,65 @@ func TestMinCapacity(t *testing.T) {
...
 }
+
+// Tests timed eviction of elements using [EvictEvery] method
+func TestEvictEvery(t *testing.T) {
+	require := require.New(t)
+
+	clock := &mockable.Clock{}
+	// Using sync clock for this test to avoid race conditions
+	// with the eviction goroutine
+	clock.Sync()
```
iansuvak, line 288: I assume we don't want to add locks to the clock since it's used everywhere.
To pass the data race detection we might be able to use atomics to store `clock.faked` and `clock.time` if we changed clock.time to stare `int64` unix representation instead of `time.Time`

This is a hack since changing two values is not atomic but these are only ever set for testing purposes when mocking the clock and would replace some sleeps in this test with `clock.Set` although we would still need to sleep for at least `2 *testEvictFrequency` in between expected evictions.

utils/window/window.go
```diff
@@ -95,18 +97,31 @@ func (w *window[T]) Length() int {
...
+		select {
+		case <-ticker.C:
+			w.lock.Lock()
+			w.removeStaleNodes()
+			w.lock.Unlock()
+		case <-ctx.Done():
+			return
+		}
+	}
+}
```
StephenButtolph, line 114: Thoughts on just using a `Mutex` rather than a `RWMutex` and then calling `removeStaleNodes` at the start of every function invocation - rather than adding this periodic polling?

iansuvak: Thanks -- that's way better.

StephenButtolph (approved):

joshua-kim (approved):

ceyonur (approved):

PR #3465, opened 2024-10-11
[tmpnet] Add --start-network to support hypersdk MODE=run

marun:
## Why this should be merged

Hypersdk tooling has long supported supplying `MODE=run` to its `run.sh` script to start a network and leave it running without executing any tests. When hypersdk switched to tmpnet, the run script was updated to specify `--reuse-network` and run a trivial test, but the use of `--reuse-network` risks confusion if a user expects the resulting network to reflect the latest binary state. Adding explicit support for always starting a new network avoids this risk. 

## How this works

- Add new flag `--start-network` to the tmpnet env flags
- If the flag is provided, start a new network, avoid stopping it on cleanup, and exit before running any tests

## How this was tested

CI for regression, locally to check that a new network was started and stayed running

### DIFF
--- a/tests/fixture/e2e/env.go
+++ b/tests/fixture/e2e/env.go
@@ -135,6 +135,7 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
 			flagVars.AvalancheGoExecPath(),
 			flagVars.PluginDir(),
 			flagVars.NetworkShutdownDelay(),
+			flagVars.StartNetwork(),
 			flagVars.ReuseNetwork(),
 		)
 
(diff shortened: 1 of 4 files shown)


tests/fixture/e2e/env.go
```diff
@@ -158,6 +159,10 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
 		}, DefaultTimeout, DefaultPollingInterval, "failed to see all chains bootstrap before timeout")
 	}
 
+	if flagVars.StartNetwork() {
+		os.Exit(0)
+	}
```
aaronbuchwald, line 164: Ah so this prevents any tests from running even if starting with ginkgo by using `os.Exit(0)` directly

This should be much better than our previous way of using `ginkgo.focus` to run a single test (simple ping)

aaronbuchwald (approved):
LGTM

yacovm (approved):

PR #3474, opened 2024-10-15
cache signatures only in acp118 handler

ceyonur:
## Why this should be merged

Cache might be needed to shared with other components. Currently we cache whole proto signature response, a plain signature caching can be more straightforward to be returned from shared cache for other components.  

For instance Subnet-EVM can use this signature cache in API calls and in legacy p2p handler. 

## How this works

Caches and returns direct signature bytes rather than proto `SignatureResponse` 

## How this was tested

Added UT


network/p2p/acp118/handler_test.go
```diff
@@ -125,6 +125,13 @@ func TestHandler(t *testing.T) {
 				require.NoError(err)
 
 				require.True(bls.Verify(pk, signature, request.Message))
+
+				if tt.cacher.Len() > 0 {
+					// Ensure the cache is populated with correct signature
+					sig, ok := tt.cacher.Get(unsignedMessage.ID())
+					require.True(ok)
+					require.Equal(sig, response.Signature)
+				}
```
StephenButtolph, line 134: Rather than checking the length (which seems fairly fragile) we should just assert that the signature is as expected if the cacher returns the signature.

network/p2p/acp118/handler.go
```diff
@@ -100,6 +100,11 @@ func (h *Handler) AppRequest(
 		}
 	}
 
+	h.signatureCache.Put(msgID, signature)
+	return responseBytesFromSignature(signature)
+}
+
+func responseBytesFromSignature(signature []byte) ([]byte, *common.AppError) {
```
StephenButtolph, line 107:
```suggestion
func signatureToResponse(signature []byte) ([]byte, *common.AppError) {
```

ceyonur: done

network/p2p/acp118/handler_test.go
ceyonur: done

StephenButtolph (approved):

PR #3476, opened 2024-10-16
Add request to update `releases.md` in PR template

ceyonur:
## Why this should be merged

Updating Release notes for every PR can be easier to capture context, reduces the risk of forgetting things and can be easier than documenting all changes just before the release.   

## How this works

adds a reminder to pull request template

## How this was tested

No need


yacovm (approved):

RELEASES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## [v1.11.12](https://github.com/ava-labs/avalanchego/releases/tag/v1.11.21)
+
+## Misc
+
+- Added reminder for documenting RELEASE.md in the pull request template
+
```
StephenButtolph, line 8: `v1.11.12` has already been released. Also we typically don't include non-user facing things into the release notes.

ceyonur: I added this to present an example

StephenButtolph (approved):

PR #3477, opened 2024-10-16
Properly initialize metrics in rpcchainVM

yacovm:
## Why this should be merged

The RPC chain VM gRPC dispatcher initializes the metrics at `Initialize()` but it doesn't need any input from the `Initialize` method's parameters. It's better to initialize the metrics at the construction of the object.

## How this works

Just a refactoring.

## How this was tested

CI

### DIFF
--- a/vms/rpcchainvm/vm_server.go
+++ b/vms/rpcchainvm/vm_server.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"google.golang.org/protobuf/types/known/emptypb"
 
@@ -74,7 +73,7 @@ type VMServer struct {
 
 	allowShutdown *utils.Atomic[bool]
 
-	metrics prometheus.Gatherer
+	metrics metrics.MultiGatherer
 	db      database.Database
 	log     logging.Logger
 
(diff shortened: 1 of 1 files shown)


vms/rpcchainvm/vm_server.go
```diff
@@ -87,14 +86,19 @@ type VMServer struct {
...
 	ssVM, _ := vm.(block.StateSyncableVM)
-	return &VMServer{
+	vmSrv := &VMServer{
 		vm:            vm,
 		bVM:           bVM,
 		ssVM:          ssVM,
 		allowShutdown: allowShutdown,
 	}
+
+	vmSrv.metrics = pluginMetrics
```
aaronbuchwald, line 100: Can we switch to just inlining this for the smallest diff? ie. add `metrics: metrics.NewPrefixGatherer()` as the next line in the struct definition?

aaronbuchwald (approved):
Left one nit, then this LGTM

vms/rpcchainvm/vm_server.go
```diff
@@ -87,14 +86,18 @@ type VMServer struct {
 
 // NewServer returns a vm instance connected to a remote vm instance
 func NewServer(vm block.ChainVM, allowShutdown *utils.Atomic[bool]) *VMServer {
+	pluginMetrics := metrics.NewPrefixGatherer()
+
 	bVM, _ := vm.(block.BuildBlockWithContextChainVM)
 	ssVM, _ := vm.(block.StateSyncableVM)
-	return &VMServer{
+	vmSrv := &VMServer{
+		metrics:       pluginMetrics,
```
darioush, line 94: nit: could just be `metrics: metrics.NewPrefixGatherer()`

darioush (approved):

StephenButtolph (approved):

aaronbuchwald:
This fixes a race condition while initializing an instance of the RPCChainVM that triggered a panic in AvalancheGo prior to this PR:

```
/go/pkg/mod/github.com/ava-labs/avalanchego@v1.11.11/vms/rpcchainvm/vm_server.go:563 +0x1a
```

The path that triggered this bug is registering VMClient (reference held by the chain manager) as a gatherer here during `vm.Initialize(...)`: https://github.com/ava-labs/avalanchego/blob/v1.11.11/vms/rpcchainvm/vm_client.go#L159

This is called before calling `Initialize` on the VM server with all of the required arguments from the vm client: https://github.com/ava-labs/avalanchego/blob/v1.11.11/vms/rpcchainvm/vm_client.go#L211. This is before the VM server would have initialized its own metrics field: https://github.com/ava-labs/avalanchego/blob/v1.11.11/vms/rpcchainvm/vm_server.go#L137.

This can trigger a panic because `Gather` can be called at any time, which can call `Gather` on the VMClient as soon as it registers itself and in the middle of `vm.Initialize(...)`.

PR #3486, opened 2024-10-21
database: add applicable dbtests for linkeddb

darioush:

### DIFF
--- a/database/dbtest/dbtest.go
+++ b/database/dbtest/dbtest.go
@@ -22,12 +22,20 @@ import (
 	"github.com/ava-labs/avalanchego/utils/units"
 )
 
+// TestsBasic is a list of all basic database tests that require only
+// a KeyValueReaderWriter.
+var TestsBasic = map[string]func(t *testing.T, db database.KeyValueReaderWriterDeleter){
+	"SimpleKeyValue":       TestSimpleKeyValue,
+	"OverwriteKeyValue":    TestOverwriteKeyValue,
+	"EmptyKey":             TestEmptyKey,
+	"KeyEmptyValue":        TestKeyEmptyValue,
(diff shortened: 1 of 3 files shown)


database/dbtest/dbtest.go
```diff
@@ -22,12 +22,20 @@ import (
 	"github.com/ava-labs/avalanchego/utils/units"
 )
 
+// TestsBasic is a list of all basic database tests that require only
+// a KeyValueReaderWriter.
```
darioush, line 26:
```suggestion
// a KeyValueReaderWriterDeleter.
```

database/linkeddb/linkeddb.go
```diff
@@ -281,7 +281,7 @@ func (ldb *linkedDB) getHeadKey() ([]byte, error) {
 func (ldb *linkedDB) putHeadKey(key []byte) error {
 	ldb.headKeyIsUpdated = true
 	ldb.updatedHeadKeyExists = true
-	ldb.updatedHeadKey = key
+	ldb.updatedHeadKey = slices.Clone(key)
```
StephenButtolph, line 284: On mobile right now - but I think there are other instances in the linkeddb that we need to perform a copy

StephenButtolph: In `Put` we hold a direct reference to `key` which can be problematic:
```golang
oldHead.Previous = key
```

StephenButtolph: I think putting the copy in `putHeadKey` results in an unnecessary copy during `Delete` as well. Maybe we could copy the `key` in `Put` once we know we are going to hold a reference to it (aka after we check that we aren't overwriting an existing node)?

darioush: updated

StephenButtolph (approved):

PR #3494, opened 2024-10-24
ACP-77: Update P-chain state staker tests

StephenButtolph:
## Why this should be merged

Factored out of #3487.

1. Rewrites `TestPersistStakers` as `TestState_writeStakers` which uses test tables rather than extensive callbacks.
2. Fixes `TestState_writeStakers` to not break the invariants around subnet validators being a sub-set of the primary network validator set.
3. Expands `TestStateAddRemoveValidator` as `TestState_ApplyValidatorDiffs` which includes additional cases, such as adding an removing validators in the same diff.

## How this works

Generally cleans up existing tests and improves coverage.

## How this was tested

Only tests are modified.

### DIFF
--- a/vms/platformvm/state/state_test.go
+++ b/vms/platformvm/state/state_test.go
@@ -278,15 +278,13 @@ func TestState_writeStakers(t *testing.T) {
 			addStakerTx:              addSubnetValidator,
 			expectedCurrentValidator: subnetCurrentValidatorStaker,
 			expectedValidatorSetOutput: &validators.GetValidatorOutput{
-				NodeID:    subnetCurrentValidatorStaker.NodeID,
-				PublicKey: primaryNetworkCurrentValidatorStaker.PublicKey,
-				Weight:    subnetCurrentValidatorStaker.Weight,
+				NodeID: subnetCurrentValidatorStaker.NodeID,
+				Weight: subnetCurrentValidatorStaker.Weight,
 			},
 			expectedWeightDiff: &ValidatorWeightDiff{
 				Decrease: false,
 				Amount:   subnetCurrentValidatorStaker.Weight,
 			},
-			expectedPublicKeyDiff: maybe.Some[*bls.PublicKey](nil),
 		},
 		"delete current primary network validator": {
 			initialStakers: []*Staker{primaryNetworkCurrentValidatorStaker},
@@ -344,7 +342,6 @@ func TestState_writeStakers(t *testing.T) {
 				Decrease: true,
 				Amount:   subnetCurrentValidatorStaker.Weight,
 			},
-			expectedPublicKeyDiff: maybe.Some[*bls.PublicKey](primaryNetworkCurrentValidatorStaker.PublicKey),
 		},
 	}
 
--- a/vms/platformvm/state/state_test.go
+++ b/vms/platformvm/state/state_test.go
@@ -5,7 +5,6 @@ package state
 
 import (
 	"context"
-	"fmt"
 	"math"
 	"math/rand"
 	"sync"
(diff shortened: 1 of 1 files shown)


vms/platformvm/state/state_test.go
```diff
@@ -109,620 +110,384 @@ func TestStateSyncGenesis(t *testing.T) {
...
-			checkValidatorsSet: func(r *require.Assertions, s *state, staker *Staker) {
-				val, err := s.GetCurrentValidator(staker.SubnetID, staker.NodeID)
-				r.NoError(err)
-
-				valsMap := s.validators.GetMap(staker.SubnetID)
-				r.Contains(valsMap, staker.NodeID)
-				valOut := valsMap[staker.NodeID]
-				r.Equal(valOut.NodeID, staker.NodeID)
-				r.Equal(valOut.Weight, val.Weight+staker.Weight)
+			expectedPublicKeyDiff: maybe.Some[*bls.PublicKey](nil),
```
samliok, line 242: why would adding a validator not update the `publicKeyDiff`?

vms/platformvm/state/state_test.go
```diff
@@ -109,620 +110,384 @@ func TestStateSyncGenesis(t *testing.T) {
...
+		},
+		"add current primary network delegator": {
+			initialStakers:            []*Staker{primaryNetworkCurrentValidatorStaker},
+			initialTxs:                []*txs.Tx{addPrimaryNetworkValidator},
+			staker:                    primaryNetworkCurrentDelegatorStaker,
+			addStakerTx:               addPrimaryNetworkDelegator,
+			expectedCurrentValidator:  primaryNetworkCurrentValidatorStaker,
+			expectedCurrentDelegators: []*Staker{primaryNetworkCurrentDelegatorStaker},
+			expectedValidatorSetOutput: &validators.GetValidatorOutput{
+				NodeID:    primaryNetworkCurrentDelegatorStaker.NodeID,
```
samliok, line 252: to keep things consistent should this be `primaryNetworkCurrentValidatorStaker.NodeID`? It should be the same value though

vms/platformvm/state/state_test.go
```diff
@@ -109,620 +110,384 @@ func TestStateSyncGenesis(t *testing.T) {
...
+				PublicKey: primaryNetworkCurrentValidatorStaker.PublicKey,
+				Weight:    primaryNetworkCurrentDelegatorStaker.Weight + primaryNetworkCurrentValidatorStaker.Weight,
+			},
+			expectedWeightDiff: &ValidatorWeightDiff{
+				Decrease: false,
+				Amount:   primaryNetworkCurrentDelegatorStaker.Weight,
 			},
```
samliok, line 259: for these tests why not also sanity check the `expectedPublicKeyDiff`

vms/platformvm/state/state_test.go
```diff
@@ -988,43 +753,49 @@ func TestValidatorWeightDiff(t *testing.T) {
...
+	for i, primaryStaker := range primaryStakers {
+		subnetStakers[i] = Staker{
+			TxID:            ids.GenerateTestID(),
+			NodeID:          primaryStaker.NodeID,
+			PublicKey:       nil, // Key is inherited from the primary network
+			SubnetID:        subnetID,
+			Weight:          uint64(i + 1),
+			StartTime:       primaryStaker.StartTime,
+			EndTime:         primaryStaker.EndTime,
+			PotentialReward: uint64(i + 1),
```
samliok, line 793: doesn't seem like we use `PotentialReward` anyway, but if start time and end time are the same, then shouldn't the potential reward be 0?

samliok (approved):

vms/platformvm/state/state_test.go
StephenButtolph: This is specifying a key diff (providing a `maybe.Some` means there is a key diff present). Adding a validator results in a key diff where the prior value was `nil`.

vms/platformvm/state/state_test.go
StephenButtolph: These tests are verifying that there is no public key diff.

vms/platformvm/state/state_test.go
StephenButtolph: > if start time and end time are the same

`endTime` is 24h after `startTime` throughout this test, but you're right these values don't really matter for this test.

PR #3496, opened 2024-10-24
Validate improper TLS RSA certificates early-on

yacovm:
## Why this should be merged

This commit makes validation of TLS certificates with either too big RSA keys, or the wrong exponent, fail as soon as the remote node presents its TLS certificate, in contrast to after the TLS handshake.

## How this works

Activates the `VerifyConnection` in the `tls.Config`, which makes the TLS handshake code in the golang TLS stack invoke the function on the TLS certificate chain presented by the remote node. 

## How this was tested

Added a unit test to added code.

Added unit tests for existing code, because why not?

No.


network/peer/tls_config_test.go
```diff
@@ -0,0 +1,73 @@
...
+	require.NoError(t, err)
+
+	basicCert := basicCert()
+	certBytes, err := x509.CreateCertificate(rand.Reader, basicCert, basicCert, &ecKey.PublicKey, ecKey)
+	require.NoError(t, err)
+
+	ecCert, err := x509.ParseCertificate(certBytes)
+	require.NoError(t, err)
+
+	for _, testCase := range []struct {
```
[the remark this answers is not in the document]
yacovm: I'm not sure I understand, can you be more specific and explicit?

staking/parse.go
```diff
@@ -136,18 +136,8 @@ func parsePublicKey(oid asn1.ObjectIdentifier, publicKey asn1.BitString) (crypto
 			return nil, ErrInvalidRSAPublicExponent
 		}
 
-		if pub.N.Sign() <= 0 {
```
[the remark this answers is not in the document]
yacovm: > Are lines 129:137 being called in the call chain of clients of ValidateRSAPublicKeyIsWellFormed?

I am not sure I follow, I just exported a few lines to a method because I want to call them from another method - `ValidateRSACertificate` in `tls_config.go`.

> Why is it necessary for the code moved to ValidateRSAPublicKeyIsWellFormed not performed by that same call chain?

I also don't quite understand what you're asking, can you be more clear? Which call chain are you referring to, exactly?

I exported the method to avoid code duplication and get better code coverage via tests.

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+			var wg sync.WaitGroup
+			wg.Add(1)
+
+			go func() {
+				defer wg.Done()
+				conn, err := listener.Accept()
+				require.NoError(t, err)
+
+				_, _, _, err = upgrader.Upgrade(conn)
```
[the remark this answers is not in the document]
yacovm: No, this test is not validating what you're referring to. It's easy to be confused as it requires some domain specific knowledge of how Golang's TLS stack works. 

Notice the client does `conn.Handshake()`. It therefore initiates a TLS connection. What is tested here, is that the function `ValidateRSACertificate` is called by the server side TLS handshake.

yacovm:
> Were this change deployed, would it be possible for the rsa certs of existing validators to fail this new validation? If so, what would be the impact?

No, because all I did was moving the code that is called [after the TLS handshake](https://github.com/ava-labs/avalanchego/blob/master/network/peer/upgrader.go#L72), [during the TLS handshake](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_server.go#L669) when we parse the TLS key of the client but before we [use](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_server.go#L740) that TLS key.

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,82 @@
...
+				x509Cert := makeRSACertAndKey(t, key)
+				return tls.ConnectionState{PeerCertificates: []*x509.Certificate{&x509Cert.cert}}
+			},
+		},
+		{
+			description: "No TLS certs given",
+			input: func() tls.ConnectionState {
+				return tls.ConnectionState{}
+			},
+			expectedErr: errors.New("no certificates sent by peer"),
```
tsachiherman, line 41: any particular reason you don't want to use the error object you've already been allocating ? ( ErrNoCertsSent )

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,221 @@
...
+			require.NoError(t, err)
+			defer listener.Close()
+
+			var wg sync.WaitGroup
+			wg.Add(1)
+
+			go func() {
+				defer wg.Done()
+				conn, err := listener.Accept()
+				require.NoError(t, err)
```
tsachiherman, line 117: I suspect that this would generate a race error in case an error is returned, since the go testing framework wouldn't allow you failing from an inner go-routine.
to "solve" this issue, I'd suggest using an error channel instead of the waitgroup and test it's result on the main go-routine.

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+
+	_ "embed"
+
+	"github.com/ava-labs/avalanchego/network/peer"
+	"github.com/ava-labs/avalanchego/staking"
+)
+
+// 8192RSA.pem is used here because it's too expensive
```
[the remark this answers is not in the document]
yacovm: I would say by putting a key in a file, implicitly it [shouldn't be used in anything but tests](https://en.wikipedia.org/wiki/Kerckhoffs%27s_principle) ;-)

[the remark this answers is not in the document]
yacovm: I changed accordingly though

staking/parse.go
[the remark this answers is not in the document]
yacovm: > Is the validation on the lines mentioned in my previous comment (L129-137) performed by the golang tls connection code? If so, presumably that same stdlib codepath is not performing the extra validation represented by ValidateRSAPublicKeyIsWellFormed - why not?


Right, so all this change is doing is essentially adding some checks to the TLS handshake such as not supporting huge keys (8192 bits) or with an exponent that is not the default exponent of Golang's TLS stack.
You can see in my test that is indeed does not do these checks, as the error we get are returned by code that I added, and not by code that was already there in the Go TLS layer. 

As for why not - well I don't know, but I do know that previous versions of avalanchego also didn't support such keys. All we're doing here is making the checks more efficient - by moving them to the TLS handshake itself and not to after it had already happened.

network/peer/upgrader_test.go
StephenButtolph: I don't think this causes a race condition, but the way that `t.FailNow()` works is by calling `panic` - which is normally caught by the testing harness managed when running a `Test` or in `t.Run(...)` but here that `panic` will escape the goroutine and cause the test to crash.

I doubt this is the only place in avalanchego where this could happen... so 🤷 on if we should fix it.

staking/parse.go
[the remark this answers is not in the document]
StephenButtolph: > As for why not - well I don't know, but I do know that previous versions of avalanchego also didn't support such keys.

These restrictions were imposed to ensure that "all reasonable RSA implementations would agree on the validity of the signatures with reasonable performance."

network/peer/tls_config_test.go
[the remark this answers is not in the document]
yacovm: It has become part of my personality by now :D

StephenButtolph (commented):
Should we consider fully validating the peer certificate prior to the handshake rather than just RSA keys?

We currently require:
1. TLS certifications are no larger than 2048 bytes.
2. A public key is provided
3. RSA modulus is eitther 2048 or 4096 bits
4. RSA modulus is not even
5. RSA exponent is 65537
6. ECDSA keys are on the P-256 curve.
7. Ed25519 keys are disallowed (will be allowed later)
8. DSA keys are disallowed

This PR introduces checks `2-5`, but not `1, 6-8`.

Essentially should we be trying to guarantee that `x509 parsing -> custom verification` can never produce a cert that isn't parsable by `custom parsing`?

network/peer/tls_config.go
```diff
@@ -26,5 +36,29 @@ func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
...
+
+	if cs.PeerCertificates[0] == nil {
+		return ErrEmptyCert
+	}
+
+	pk := cs.PeerCertificates[0].PublicKey
+
+	switch rsaKey := pk.(type) {
+	case *rsa.PublicKey:
+		return staking.ValidateRSAPublicKeyIsWellFormed(rsaKey)
```
StephenButtolph, line 58: This is definitely me just being overly defensive, but should we attempt to handle the case that `rsaKey` is `nil` here?

Test would look like:
```golang
{
	description: "nil RSA key",
	input: func() tls.ConnectionState {
		key, err := rsa.GenerateKey(rand.Reader, 2048)
		require.NoError(t, err)

		x509CertWithNilPK := makeRSACertAndKey(t, key)
		x509CertWithNilPK.cert.PublicKey = (*rsa.PublicKey)(nil)
		return tls.ConnectionState{PeerCertificates: []*x509.Certificate{&x509CertWithNilPK.cert}}
	},
	expectedErr: peer.ErrEmptyPublicKey,
},
```

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, testCase := range []struct {
+		description string
+		input       func() tls.ConnectionState
+		expectedErr error
+	}{
+		{
+			description: "Valid TLS cert",
+			input: func() tls.ConnectionState {
+				key, err := rsa.GenerateKey(rand.Reader, 2048)
+				require.NoError(t, err)
```
StephenButtolph, line 30: I think these functions are referencing the wrong `t`. These would call operations on the `TestValidateRSACertificate` `t` rather than the inner `t.Run(...)` `t`.

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package peer_test
```
StephenButtolph, line 4: Out of curiosity, why is it beneficial to compile these as a testing package? Typically we've only done this to avoid import cycles.

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+				_, _, _, err = upgrader.Upgrade(conn)
+
+				require.ErrorIs(t, err, testCase.expectedErr)
+			}()
+
+			conn, err := tls.Dial("tcp", listener.Addr().String(), &clientConfig)
+			require.NoError(t, err)
+
+			require.NoError(t, conn.Handshake())
```
StephenButtolph, line 128: I'm a bit surprised that this passes, is this just because the client signature verification is performed last?

yacovm:
> Should we consider fully validating the peer certificate prior to the handshake rather than just RSA keys?
> 
> We currently require:
> 
> 1. TLS certifications are no larger than 2048 bytes.
> 2. A public key is provided
> 3. RSA modulus is eitther 2048 or 4096 bits
> 4. RSA modulus is not even
> 5. RSA exponent is 65537
> 6. ECDSA keys are on the P-256 curve.
> 7. Ed25519 keys are disallowed (will be allowed later)
> 8. DSA keys are disallowed
> 
> This PR introduces checks `2-5`, but not `1, 6-8`.
> 
> Essentially should we be trying to guarantee that `x509 parsing -> custom verification` can never produce a cert that isn't parsable by `custom parsing`?

1. There is a check in the TLS code that the payload of each message is less than 16MB. When we invoke the method `VerifyConnection`, we have already read the certificate in its full size, so I'm not sure what is gained by restricting the certificate size after we have parsed it. 
6. I can do that, however - can the rest of the codebase handle P384? If so, how can we be sure there aren't networks with P384 that would get locked out as a result? Do we just assume everybody use whatever is in `staking/tls.go` ? 
7. Same question as above. 
8. DSA is not available in TLS 1.3, so they cannot verify such a signature, hence the handshake would most likely fail at the cipher suite negotiation phase which is before the signature check.

network/peer/tls_config.go
yacovm: Added test + check in `ValidateRSAPublicKeyIsWellFormed`.

network/peer/tls_config_test.go
yacovm: It's not necessary here. 

The `_test` can sometimes be useful because it forces the writer of the test to only use public APIs and not be able to look at internal state. 

I can change it to a non `_test` if it's an issue.

network/peer/tls_config_test.go
yacovm: yes you're right. I moved them because of code review comments and didn't notice it.

network/peer/upgrader_test.go
yacovm: In TLS1.3 we first negotiate the encryption, and do the authentication last. The client is the last party to send its authentication, and the implementation doesn't wait for the acknowledgement from the server, i guess for performance reasons.

yacovm:
> Essentially should we be trying to guarantee that x509 parsing -> custom verification can never produce a cert that isn't parsable by custom parsing?

In this PR I'm just addressing low hanging fruits that make sense to handle early-on to avoid nonsensical exotic configuration that shouldn't be used, such as an avalanchego node not using the RSA verification exponent that is built-in in Golang, or using an exceedingly large modulus of 8K bits. 

Of course, the space of certificates valid by x509 is a much larger superset than what we need to accept.

PR #3498, opened 2024-10-25
Add `Deregister` to `metrics.MultiGatherer` interface

StephenButtolph:
## Why this should be merged

Metrics are registered throughout avalanchego, deregistering all of them individually would require extensive additional code flows during the shutdown process. However, there are only a handful of metrics namespaces. This PR allows easily deregistering an entire namespace of metrics.

## How this works

Adds `Deregister` to the `metrics.MultiGatherer` interface.

## How this was tested

- [X] Added unit tests.

No.

### DIFF
--- a/api/metrics/label_gatherer_test.go
+++ b/api/metrics/label_gatherer_test.go
@@ -138,9 +138,13 @@ func TestLabelGatherer_Gather(t *testing.T) {
 	}
 }
 
-func TestLabelGatherer_Register(t *testing.T) {
+func TestLabelGatherer_Registration(t *testing.T) {
+	const (
+		firstName  = "first"
+		secondName = "second"
+	)
 	firstLabeledGatherer := &labeledGatherer{
-		labelValue: "first",
+		labelValue: firstName,
 		gatherer:   &testGatherer{},
 	}
 	firstLabelGatherer := func() *labelGatherer {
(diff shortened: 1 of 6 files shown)


api/metrics/label_gatherer_test.go
```diff
@@ -214,4 +230,64 @@ func TestLabelGatherer_Register(t *testing.T) {
...
+			expectedLabelGatherer: onlySecondLabeledGatherer,
+		},
+	}
+	for _, test := range deregisterTests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			removed := test.labelGatherer.Deregister(test.labelValue)
+			require.Equal(test.expectedRemoved, removed)
+			require.Equal(test.expectedLabelGatherer, test.labelGatherer)
```
yacovm, line 290: `require.Equal(test.expectedLabelGatherer, test.labelGatherer)` ?

StephenButtolph: Sorry I don't understand the question

utils/slice.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package utils
+
+// DeleteIndex moves the last element in the slice to index [i] and shrinks the
```
yacovm, line 6: I understand this is O(1) and not O(n) like the regular `slices.Delete` but it is less code and less complexity to just use the aforementioned builtin `slices.Delete` over adding a new method. 
`Deregister` isn't called yet in the code, and if called - it will be probably be called seldomly, so I'm not sure if it's worth adding it from a code maintenance perspective.

yacovm (approved):

darioush (approved):

api/metrics/label_gatherer_test.go
```diff
@@ -214,4 +230,64 @@ func TestLabelGatherer_Register(t *testing.T) {
...
+			name:                  "remove only first name",
+			labelGatherer:         secondLabelGatherer(),
+			labelValue:            firstName,
+			expectedRemoved:       true,
+			expectedLabelGatherer: onlySecondLabeledGatherer,
+		},
+	}
+	for _, test := range deregisterTests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
```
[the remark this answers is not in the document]
StephenButtolph: We do this shadowing pretty much everywhere... I feel like the shadowing here is almost beneficial, as there is no reason to ever use any of the functions in the package after this point.

utils/slice.go
StephenButtolph: This definitely isn't performance critical... But I also feel like I've re-implemented this same code like 5 times... So I feel like it is a reasonable helper to add

PR #3499, opened 2024-10-25
ACP-77: Add subnetIDNodeID struct

StephenButtolph:
## Why this should be merged

Factored out of #3388.

## How this works

Adds a struct to merge `subnetID`s and `nodeID`s together. This struct supports directly marshaling into bytes because it will be used as a key into the database.

## How this was tested

- [X] Added fuzz tests.

No.

### DIFF
--- a/vms/platformvm/state/subnet_id_node_id_test.go
+++ b/vms/platformvm/state/subnet_id_node_id_test.go
@@ -1,0 +1,66 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package state
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/thepudds/fzgen/fuzzer"
+)
+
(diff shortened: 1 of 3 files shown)


vms/platformvm/state/subnet_id_node_id_test.go
```diff
@@ -0,0 +1,66 @@
...
+func FuzzSubnetIDNodeIDMarshalOrdering(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		var (
+			v0 subnetIDNodeID
+			v1 subnetIDNodeID
+		)
+		fz := fuzzer.NewFuzzer(data)
+		fz.Fill(&v0, &v1)
+
+		if v0.subnetID == v1.subnetID {
```
yacovm, line 54: is this return statement and if condition essential?

yacovm (approved):

vms/platformvm/state/subnet_id_node_id_test.go
StephenButtolph: If the subnetIDs are equal, then the comparison depends on the nodeIDs as well.

ceyonur (approved):

PR #3500, opened 2024-10-26
Split `writeCurrentStakers` into multiple functions

StephenButtolph:

### DIFF
--- a/vms/platformvm/state/stakers.go
+++ b/vms/platformvm/state/stakers.go
@@ -5,6 +5,7 @@ package state
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/google/btree"
 
@@ -271,6 +272,34 @@ type diffValidator struct {
 
 	addedDelegators   *btree.BTreeG[*Staker]
 	deletedDelegators map[ids.ID]*Staker
+}
+
+func (d *diffValidator) WeightDiff() (ValidatorWeightDiff, error) {
+	weightDiff := ValidatorWeightDiff{
+		Decrease: d.validatorStatus == deleted,
+	}
+	if d.validatorStatus != unmodified {
+		weightDiff.Amount = d.validator.Weight
+	}
+
+	for _, staker := range d.deletedDelegators {
+		if err := weightDiff.Add(true, staker.Weight); err != nil {
+			return ValidatorWeightDiff{}, fmt.Errorf("failed to decrease node weight diff: %w", err)
+		}
+	}
+
+	addedDelegatorIterator := iterator.FromTree(d.addedDelegators)
+	defer addedDelegatorIterator.Release()
+
+	for addedDelegatorIterator.Next() {
+		staker := addedDelegatorIterator.Value()
+
+		if err := weightDiff.Add(false, staker.Weight); err != nil {
+			return ValidatorWeightDiff{}, fmt.Errorf("failed to increase node weight diff: %w", err)
+		}
+	}
+
+	return weightDiff, nil
 }
 
 // GetValidator attempts to fetch the validator with the given subnetID and
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -4,6 +4,7 @@
 package state
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


vms/platformvm/state/stakers.go
```diff
@@ -273,6 +274,34 @@ type diffValidator struct {
 	deletedDelegators map[ids.ID]*Staker
 }
 
+func (d *diffValidator) WeightDiff() (ValidatorWeightDiff, error) {
```
StephenButtolph, line 277: This was previously implemented in `writeCurrentStakersSubnetDiff`.

vms/platformvm/state/state.go
```diff
@@ -2040,47 +2044,75 @@ func (s *state) writeExpiry() error {
 	return nil
 }
 
-func (s *state) writeCurrentStakers(updateValidators bool, height uint64, codecVersion uint16) error {
+// getInheritedPublicKey returns the primary network validator's public key.
+//
+// Note: This function may return a nil public key and no error if the primary
+// network validator does not have a public key.
+func (s *state) getInheritedPublicKey(nodeID ids.NodeID) (*bls.PublicKey, error) {
```
StephenButtolph, line 2051: This logic previously existed as part of `writeCurrentStakersSubnetDiff`.

vms/platformvm/state/state.go
```diff
@@ -1792,7 +1794,9 @@ func (s *state) write(updateValidators bool, height uint64) error {
 	return errors.Join(
 		s.writeBlocks(),
 		s.writeExpiry(),
-		s.writeCurrentStakers(updateValidators, height, codecVersion),
+		s.updateValidatorManager(updateValidators),
+		s.writeValidatorDiffs(height),
+		s.writeCurrentStakers(codecVersion),
```
StephenButtolph, line 1799: While splitting these out is technically less efficient (and contains similar code). The functions were just getting too complex to understand. This is explicitly trading off some (imo insignificant) level of performance for clarity.

StephenButtolph (commented):
There are no testing changes because this is a pure refactor.

vms/platformvm/state/state.go
```diff
@@ -2040,47 +2044,75 @@ func (s *state) writeExpiry() error {
 	return nil
 }
 
-func (s *state) writeCurrentStakers(updateValidators bool, height uint64, codecVersion uint16) error {
+// getInheritedPublicKey returns the primary network validator's public key.
+//
+// Note: This function may return a nil public key and no error if the primary
```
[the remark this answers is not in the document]
michaelkaplan13: Prior to Durango, the `AddValidatorTx` was still supported and allowed validators to be registered without a BLS key (as opposed to [`AddPermisionlessValidatorTx` ](https://docs.avax.network/api-reference/p-chain/txn-format#unsigned-add-permissionless-validator-tx) which requires the public key).

So any validators previously added using `AddValidtorTx` that still haven't reached their end time will not have public keys. After the 1 year maximum staking duration from the activation of Durango, all primary network validators will have public keys

vms/platformvm/state/state.go
```diff
@@ -2094,185 +2126,153 @@ func (s *state) writeCurrentStakers(updateValidators bool, height uint64, codecV
...
-					// The primary network validator is being removed during
-					// writing.
-					pk = vdr.validator.PublicKey
-				} else {
-					// This should never happen as the primary network diffs are
-					// written last and subnet validator times must be a subset
-					// of the primary network validator times.
-					return fmt.Errorf("%w: %s", errMissingPrimaryNetworkValidator, nodeID)
-				}
+// writeValidatorDiffs writes the validator set diff contained by the pending
```
michaelkaplan13, line 2129: Just for my own understanding, is `writeValidatorDiffs` meant to only handle changes to the validators set of subnets before `ConvertSubnetToL1Tx`s? I'm not clear on its usage vs  `updateValidatorManager`

StephenButtolph: `writeValidatorDiffs` will include the logic for L1 diffs in a later PR.

michaelkaplan13 (approved):

vms/platformvm/state/state.go
```diff
@@ -2094,185 +2126,153 @@ func (s *state) writeCurrentStakers(updateValidators bool, height uint64, codecV
...
+// validator set changes to disk.
+//
+// This function must be called prior to writeCurrentStakers.
+func (s *state) writeValidatorDiffs(height uint64) error {
```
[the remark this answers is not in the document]
StephenButtolph: `TestState_writeStakers` is the test for this behavior.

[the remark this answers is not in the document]
StephenButtolph: Ended up splitting up the function and expanded `TestState_writeStakers`

PR #3501, opened 2024-10-27
Further restrict public keys in TLS handshake

yacovm:
## Why this should be merged

This commit further restricts the public keys allowed in the TLS certificates of clients:
    
    - No ed25519
    - For ECDSA, only P256 is supported

## How this works

Expands `ValidateCertificate` (formally `ValidateRSACertificate`) by adding extra checks

## How this was tested

Added some test cases in the corresponding unit test.

I don't think so, because anyone that uses avalanchego from the official release shouldn't have non P256 EC keys.


network/peer/tls_config.go
```diff
@@ -26,5 +42,38 @@ func TLSConfig(cert tls.Certificate, keyLogWriter io.Writer) *tls.Config {
...
+			return ErrCurveMismatch
+		}
+		return nil
+	case *rsa.PublicKey:
+		return staking.ValidateRSAPublicKeyIsWellFormed(key)
+	case nil:
+		return ErrEmptyPublicKey
+	default:
+		return nil
 	}
```
StephenButtolph, line 78: Feels like we can reduce the number of cases in this flow by only specifying the types that we want to allow.
```suggestion
	switch key := pk.(type) {
	case *ecdsa.PublicKey:
		if key == nil {
			return ErrEmptyPublicKey
		}
		if key.Curve != elliptic.P256() {
			return ErrCurveMismatch
		}
		return nil
	case *rsa.PublicKey:
		return staking.ValidateRSAPublicKeyIsWellFormed(key)
	default:
		return ErrUnsupportedKeyType
	}
```

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,131 @@
...
+				require.NoError(t, err)
+
+				basicCert := basicCert()
+				certBytes, err := x509.CreateCertificate(rand.Reader, basicCert, basicCert, &ecKey.PublicKey, ecKey)
+				require.NoError(t, err)
+
+				ecCert, err := x509.ParseCertificate(certBytes)
+				require.NoError(t, err)
+
+				ecCert.PublicKey = nil
```
StephenButtolph, line 104: Is this supposed to hit the ecdsa public key nil check?
```suggestion
				ecCert.PublicKey = (*ecdsa.PublicKey)(nil)
```

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,131 @@
...
+				ecCert, err := x509.ParseCertificate(certBytes)
+				require.NoError(t, err)
+				return tls.ConnectionState{PeerCertificates: []*x509.Certificate{ecCert}}
+			},
+		},
+		{
+			description: "EC cert with empty key",
+			expectedErr: peer.ErrEmptyPublicKey,
+			input: func(t *testing.T) tls.ConnectionState {
+				ecKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
StephenButtolph, line 94: Should we add a test where we use a different curve?

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return tlsCertServer
+}
+
+type certAndKey struct {
+	cert x509.Certificate
+	key  *rsa.PrivateKey
+}
+
+func makeRSACertAndKey(t *testing.T, privKey *rsa.PrivateKey) certAndKey {
```
StephenButtolph, line 190: Why does this function return `certAndKey`? The caller already has the `privKey` (they passed it into this function). Seems like we could just:
```suggestion
func makeRSACert(t *testing.T, privKey *rsa.PrivateKey) x509.Certificate {
```

network/peer/upgrader_test.go
```diff
@@ -0,0 +1,222 @@
...
+	return &x509.Certificate{
+		SerialNumber:          big.NewInt(0).SetInt64(100),
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().Add(time.Hour).UTC(),
+		BasicConstraintsValid: true,
+	}
+}
+
+type mockPrometheusCounter struct {
+	t *testing.T
```
StephenButtolph, line 215: `t` never seems to be used
```suggestion
```

network/peer/tls_config_test.go
```diff
@@ -0,0 +1,131 @@
...
+				require.NoError(t, err)
+
+				basicCert := basicCert()
+				certBytes, err := x509.CreateCertificate(rand.Reader, basicCert, basicCert, &ecKey.PublicKey, ecKey)
+				require.NoError(t, err)
+
+				ecCert, err := x509.ParseCertificate(certBytes)
+				require.NoError(t, err)
+				return tls.ConnectionState{PeerCertificates: []*x509.Certificate{ecCert}}
```
StephenButtolph, line 87: Should we add a `makeECDSACert` function as well to align with the `makeRSACert` (`AndKey`) function?

staking/parse_test.go
```diff
@@ -31,3 +34,73 @@ func BenchmarkParse(b *testing.B) {
...
+			description: "even modulus",
+			getPK: func() rsa.PublicKey {
+				evenN := new(big.Int).Set(validKey.N)
+				evenN.Add(evenN, big.NewInt(1))
+				return rsa.PublicKey{
+					N: evenN,
+					E: 65537,
+				}
+			},
+			expectErr: ErrRSAModulusIsEven,
```
StephenButtolph, line 63: This is going to make my OCD really obvious lol
```suggestion
			expectErr: ErrRSAModulusIsEven,
			getPK: func() rsa.PublicKey {
				evenN := new(big.Int).Set(validKey.N)
				evenN.Add(evenN, big.NewInt(1))
				return rsa.PublicKey{
					N: evenN,
					E: 65537,
				}
			},
```

network/peer/tls_config_test.go
yacovm: yes

network/peer/upgrader_test.go
yacovm: because I took it from the [PoC](https://github.com/yacovm/avalanchego/blob/tls/network/p2p/tls/tls_test.go#L87-L105) I made...

yacovm: will change it, good call

staking/parse_test.go
yacovm: I turn off mine when it comes to test code ;-)

network/peer/tls_config_test.go
yacovm: No, I'll just make `makeRSACert` handle both.

network/peer/tls_config_test.go
yacovm: done

network/peer/tls_config.go
yacovm: [Copied](https://github.com/ava-labs/avalanchego/pull/3501/commits/e45a72ae90a2c993915320789a299e78557fccac#diff-566b013fa8a33a9c2561ca4b8951d78da4fa0bb3f63fc3b48499d425f9c80217L60-L77) your code snippet.

network/peer/upgrader_test.go
yacovm: [Done](https://github.com/ava-labs/avalanchego/pull/3501/commits/e45a72ae90a2c993915320789a299e78557fccac#diff-904154d6e72be15a120d196a65b860c273801bb294ffca48bdf08e377a5f33a3L186-R193)

network/peer/upgrader_test.go
yacovm: [Fixed](https://github.com/ava-labs/avalanchego/pull/3501/commits/e45a72ae90a2c993915320789a299e78557fccac#diff-904154d6e72be15a120d196a65b860c273801bb294ffca48bdf08e377a5f33a3L215)

network/peer/tls_config_test.go
yacovm: [Done](https://github.com/ava-labs/avalanchego/pull/3501/commits/e45a72ae90a2c993915320789a299e78557fccac#diff-3eb82f83d630d7aeeeba49e200bdb1fad994fff6297717e10dc1faaa54bc0ce6L81-L85)

staking/parse_test.go
yacovm: [Fixed](https://github.com/ava-labs/avalanchego/pull/3501/commits/e45a72ae90a2c993915320789a299e78557fccac#diff-f15b6d29bdd07ca1a1e4bb7a665f7c67eaf0ec5f2f6fae634c3619cc90efdc75R55)

StephenButtolph (approved):

PR #3502, opened 2024-10-28
Use subnet public key diffs after Etna is activated

StephenButtolph:
## Why this should be merged

When existing networks upgrade from Pre-Etna to Post-Etna, the nodes in that network are only guaranteed to have populated the public key diffs for subnets after the transition.

This means that we can only safely use this new index after Etna has been activated.

However, we must use the new index after Etna is activated because the public keys are not guaranteed to be inherited from the primary network after Etna.

## How this works

Uses the new public key diffs after Etna is activated (but not before).

## How this was tested

Added multiple new unit tests.

There are no user facing changes.

### DIFF
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -86,6 +86,7 @@ var (
 	ExpiryReplayProtectionPrefix  = []byte("expiryReplayProtection")
 	SingletonPrefix               = []byte("singleton")
 
+	EtnaHeightKey      = []byte("etna height")
 	TimestampKey       = []byte("timestamp")
 	FeeStateKey        = []byte("fee state")
 	SoVExcessKey       = []byte("sov excess")
@@ -144,6 +145,9 @@ type State interface {
 	Chain
 	uptime.State
 	avax.UTXOReader
+
+	// TODO: Remove after Etna is activated
+	GetEtnaHeight() (uint64, error)
 
 	GetLastAccepted() ids.ID
 	SetLastAccepted(blkID ids.ID)
@@ -293,6 +297,7 @@ type stateBlk struct {
  * '-. singletons
  *   |-- initializedKey -> nil
  *   |-- blocksReindexedKey -> nil
+ *   |-- etnaHeightKey -> height
  *   |-- timestampKey -> timestamp
  *   |-- feeStateKey -> feeState
  *   |-- sovExcessKey -> sovExcess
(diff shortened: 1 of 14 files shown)


vms/platformvm/state/state.go
```diff
@@ -2516,7 +2525,13 @@ func (s *state) writeChains() error {
 	return nil
 }
 
-func (s *state) writeMetadata() error {
+func (s *state) writeMetadata(height uint64) error {
+	if !s.upgrades.IsEtnaActivated(s.persistedTimestamp) && s.upgrades.IsEtnaActivated(s.timestamp) {
+		if err := database.PutUInt64(s.singletonDB, EtnaHeightKey, height); err != nil {
+			return fmt.Errorf("failed to write etna height: %w", err)
```
ceyonur, line 2531: nit: should we check if we have already written it? In case where we can call the commit without advancing the chain (like disconnect?).

ceyonur (commented):
just a nit

vms/platformvm/validators/manager.go
```diff
@@ -283,19 +292,22 @@ func (m *manager) makePrimaryNetworkValidatorSet(
...
 ) (map[ids.NodeID]*validators.GetValidatorOutput, uint64, error) {
-	primaryMap := m.cfg.Validators.GetMap(constants.PrimaryNetworkID)
+	subnetMap := m.cfg.Validators.GetMap(subnetID)
 	currentHeight, err := m.getCurrentHeight(ctx)
-	return primaryMap, currentHeight, err
+	return subnetMap, currentHeight, err
 }
 
+// TODO: Once Etna has been activated, remove this function and use
+// makeValidatorSet for all validator set lookups.
```
michaelkaplan13, line 310: The TODO on line 365 is no longer relevant since now after Etna `makeValidatorSet` should be used instead of `makeSubnetValidatorSet`, right?

vms/platformvm/state/state.go
StephenButtolph: `persistedTimestamp` is updated to equal `timestamp` immediately after this, and both of them are monotonically increasing. So this condition can only happen once.

vms/platformvm/validators/manager.go
StephenButtolph: 🦅 👁️

michaelkaplan13 (approved):

PR #3505, opened 2024-10-30
Clarify partial sync flag

michaelkaplan13:
## Why this should be merged
Clarifies that nodes can using the `partial-sync-primary-network` flag can be L1 validators post-Etna activation.

## How this works
Updates documentation.

## How this was tested
N/A

No

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -176,7 +176,9 @@ must be the same with the number of given `--state-sync-ids`.
 
 #### `--partial-sync-primary-network` (string)
 
-Partial sync enables non-validators to optionally sync only the P-chain on the primary network.
+Partial sync enables nodes that are not primary validators to optionally sync only
+the P-chain on the primary network. Nodes that use this option can still track 
+Subnets. After the Etna upgrade, nodes that use this option can also validate L1s.
 
 ## Chain Configs
 


config/config.md
```diff
@@ -176,7 +176,9 @@ must be the same with the number of given `--state-sync-ids`.
 
 #### `--partial-sync-primary-network` (string)
 
-Partial sync enables non-validators to optionally sync only the P-chain on the primary network.
+Partial sync enables nodes that are not primary validators to optionally sync only
```
StephenButtolph, line 179: nit: I think we typically say "primary network validators" rather than "primary validators"

past that lgtm

config/config.md
```diff
@@ -176,7 +176,9 @@ must be the same with the number of given `--state-sync-ids`.
...
+the P-chain on the primary network. Nodes that use this option can still track 
+Subnets. After the Etna upgrade, nodes that use this option can also validate L1s.
```
meaghanfitzgerald:
```suggestion
Partial sync enables nodes that do not validate the [Primary Network](https://docs.avax.network/learn/primary-network) to sync only the P-Chain. Nodes that use this option can still track 
Subnets. After the Etna upgrade, nodes that use this option can also sync and/or validate L1s.
```

meaghanfitzgerald: phrasing seemed confusing, added a link to the doc that explains the primary network as well.

michaelkaplan13: If they track any Subnets they wouldn't only be syncing the P-Chain though 😅

meaghanfitzgerald: true 😵‍💫

StephenButtolph (approved):

meaghanfitzgerald (approved):

PR #3506, opened 2024-10-30
Update BLST to v0.3.13

yacovm:
## Why this should be merged

Ethereum upgraded its BLST dependency to 0.3.13, and current BLST (0.3.11) is broken on Mac 14.7 and higher but is fixed at [later](https://github.com/ethereum/go-ethereum/issues/30494#issuecomment-2373421802) versions.

## How this works

Just updates the dependency.

## How this was tested

CI

No


StephenButtolph (approved):
wcgw

PR #3508, opened 2024-10-31
[testing] Enable bootstrap testing of partial sync

marun:
## Why this should be merged

Previously there were only 2 modes for bootstrap testing - state sync and full sync. This change enables support for partial sync which only processes the history of the P-Chain and is expected to be used by L1 validators.

## How this works

Replaced the `StateSyncEnabled` field in the bootstrap test config with `SyncMode` and 3 initial values: `full-sync`, `c-chain-state-sync` and `p-chain-full-sync-only`. The value of `SyncMode` is determined from the set of env vars defined for an avalanchego container.

## How this was tested

Add unit tests to validate determination of sync mode.

This is not a user facing change.

### DIFF
--- a/tests/fixture/bootstrapmonitor/README.md
+++ b/tests/fixture/bootstrapmonitor/README.md
@@ -34,8 +34,8 @@ flag.
 #### Full Sync
 
 In this mode, the full history of the X-, P- and C-Chains will be
-processed. This is configured by including `state-sync-enabled:
-false` in the C-Chain configuration.
+processed. This is configured by including
+`state-sync-enabled:false` in the C-Chain configuration.
 
 ## Overview
 
--- a/tests/fixture/bootstrapmonitor/README.md
+++ b/tests/fixture/bootstrapmonitor/README.md
@@ -65,7 +65,9 @@ and initiates a new test when one is found.
      - The `full-sync` mode is enabled by including
        `state-sync-enabled:false` in the
        `AVAGO_CHAIN_CONFIG_CONTENT` env var set for the avalanchego
-       container.
+       container and either not including a value for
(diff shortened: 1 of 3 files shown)


tests/fixture/bootstrapmonitor/README.md
```diff
@@ -48,9 +56,16 @@ and initiates a new test when one is found.
    - The network targeted by the test is determined by the value of
      the `AVAGO_NETWORK_NAME` env var set for the avalanchego
      container.
-   - Whether state sync is enabled is determined by the value of the
-     `AVAGO_CHAIN_CONFIG_CONTENT` env var set for the avalanchego
-     container.
+   - By default, the sync mode will be `full-sync`. The other sync
```
michaelkaplan13, line 59: Above, it says that C-Chain State Sync is the default mode, which seems to be contradicted here.

tests/fixture/bootstrapmonitor/README.md
```diff
@@ -48,9 +56,16 @@ and initiates a new test when one is found.
...
+     modes require providing additional configuration:
+     - The `only-p-chain-full-sync` mode is enabled by setting the
+       `AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK` env var to `"true"` for
+       the avalanchego container. If this mode is enabled, the state
+       sync configuration for the C-Chain will be ignored.
+     - The `c-chain-state-sync` mode is enabled by including
+       `state-sync-enabled:false` in the
```
michaelkaplan13, line 66: This seems to suggest that state sync is the default. If that's the case, we may want to specify to enable full sync by setting `state-sync-enabled:false`, and not setting `only-p-chain-full-sync` option

michaelkaplan13 (approved):

StephenButtolph (approved):

PR #3509, opened 2024-10-31
ACP-77: Filter the inactive validator from block proposals and tx gossip

StephenButtolph:
## Why this should be merged

After ACP-77, inactive validators are grouped into a single `EmptyNodeID` entry. This `EmptyNodeID` should not be considered for block production or transaction gossip.

## How this works

Filters the `EmptyNodeID` from the validator set prior to the usage for tx gossip and generation of the proposal list.

## How this was tested

Added unit tests.

This is not a user facing change.

### DIFF
--- a/network/p2p/validators_test.go
+++ b/network/p2p/validators_test.go
@@ -296,6 +296,31 @@ func TestValidatorsTop(t *testing.T) {
 				nodeID1,
 			},
 		},
+		{
+			name: "top ignores inactive validators",
+			validators: []validator{
+				{
+					nodeID: ids.EmptyNodeID,
+					weight: 4,
+				},
+				{
+					nodeID: nodeID1,
+					weight: 2,
+				},
+				{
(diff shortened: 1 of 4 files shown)


network/p2p/validators_test.go
```diff
@@ -296,6 +296,31 @@ func TestValidatorsTop(t *testing.T) {
 				nodeID1,
 			},
 		},
+		{
+			name: "top ignores inactive validators",
```
joshua-kim, line 300: Can we add a similar test for `TestValdiatorsSample`?

StephenButtolph: `Sample` already ignores disconnected validators and it's unexpected for `EmptyNodeID` to be marked as connected.

So, we could add that case, but it felt a bit odd to me.

If you want it added I'm fine adding it.

joshua-kim (approved):

PR #3511, opened 2024-11-01
[testing] Update bootstrap monitor to output disk usage in bytes

marun:
Previously the output was in human-readable form which is fine for logs but not easily graphable.

### DIFF
--- a/tests/fixture/bootstrapmonitor/wait.go
+++ b/tests/fixture/bootstrapmonitor/wait.go
@@ -197,7 +197,7 @@ func getDiskUsage(log logging.Logger, dir string) uint64 {
 	usage, err := strconv.ParseUint(rawUsage, 10, 64)
 	if err != nil {
 		log.Error("Failed to parse disk usage",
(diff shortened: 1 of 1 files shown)


tests/fixture/bootstrapmonitor/wait.go
```diff
@@ -148,9 +149,9 @@ func WaitForCompletion(
 	return nil
 }
 
-// Determines the current disk usage for the specified directory
-func getDiskUsage(log logging.Logger, dir string) string {
-	cmd := exec.Command("du", "-sh", dir)
+// Determines the current disk usage (in bytes) for the specified directory
+func getDiskUsage(log logging.Logger, dir string) uint64 {
+	cmd := exec.Command("du", "-sb", dir)
```
yacovm, line 154: unrelated to this PR, but, perhaps it's better to use an[ API call ](https://www.includehelp.com/golang/get-the-disk-usage-information.aspx)instead of using `du`?

PR #3512, opened 2024-11-01
Return shallow copy of validator set in platformVM's validator manager

yacovm:
## Why this should be merged

PlatformVM's manager returns validators as a map, and caches the same instance. Then, it might be that different goroutines try to access this map instance concurrently, which is not safe and causes a data race.

## How this works

This commit simply shallow clones the map when returning it from within the validator manager.

## How this was tested

CI

No

### DIFF
--- a/vms/platformvm/validators/manager.go
+++ b/vms/platformvm/validators/manager.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"time"
 
 	"github.com/ava-labs/avalanchego/cache"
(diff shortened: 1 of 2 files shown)


vms/platformvm/validators/manager.go
```diff
@@ -230,7 +231,7 @@ func (m *manager) GetValidatorSet(
 	m.metrics.IncValidatorSetsCreated()
 	m.metrics.AddValidatorSetsDuration(duration)
 	m.metrics.AddValidatorSetsHeightDiff(currentHeight - targetHeight)
-	return validatorSet, nil
+	return maps.Clone(validatorSet), nil
```
StephenButtolph, line 234: I think the issue also comes from L203

StephenButtolph (commented):
Is it possible to add a regression test for this?

StephenButtolph (approved):

PR #3514, opened 2024-11-01
Add `ValidatorWeightDiff` `Add` and `Sub` helpers

StephenButtolph:
## Why this should be merged

Currently `Add(true/false, amount)` is pretty clunky to use. It was implemented this way to avoid duplicate code in the implementation, but adds needless complexity during the usage.

## How this works

This PR un-exports the current `Add(bool, uint64)` function and replaces it with `Add(uint64)` and `Sub(uint64)`.

## How this was tested

Updated existing tests.

No user facing changes.

### DIFF
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -425,7 +425,15 @@ type ValidatorWeightDiff struct {
 	Amount   uint64 `serialize:"true"`
 }
 
-func (v *ValidatorWeightDiff) Add(negative bool, amount uint64) error {
+func (v *ValidatorWeightDiff) Add(amount uint64) error {
+	return v.add(false, amount)
+}
+
+func (v *ValidatorWeightDiff) Sub(amount uint64) error {
+	return v.add(true, amount)
+}
+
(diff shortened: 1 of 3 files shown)


vms/platformvm/state/state.go
```diff
@@ -425,7 +425,15 @@ type ValidatorWeightDiff struct {
...
-func (v *ValidatorWeightDiff) Add(negative bool, amount uint64) error {
+func (v *ValidatorWeightDiff) Add(amount uint64) error {
+	return v.add(false, amount)
+}
+
+func (v *ValidatorWeightDiff) Sub(amount uint64) error {
+	return v.add(true, amount)
+}
+
+func (v *ValidatorWeightDiff) add(negative bool, amount uint64) error {
```
samliok, line 436: i feel like `update(add bool, amount uint64)` is a more descriptive function header but doesn't matter too much

samliok (approved):

vms/platformvm/state/state.go
samliok: or maybe `addOrSubtract`

samliok (approved):

PR #3517, opened 2024-11-04
Add script to configure metrics and log collection from a local node

marun:
## Why this should be merged

Collecting logs and metrics shouldn't always require running on monitored infrastructure when collection to the CI monitoring stack from a local node is possible.

## How this works

Adds promtail and prometheus configuration for a local node to the path expected by the instances started by the `scripts/run_prometheus.sh` and `scripts/run_promtail.sh` scripts.

## How this was tested

Manually

N/A

### DIFF
--- a/scripts/configure-local-metrics-collection.sh
+++ b/scripts/configure-local-metrics-collection.sh
@@ -7,12 +7,12 @@ set -euo pipefail
 
 API_PORT="${API_PORT:-9650}"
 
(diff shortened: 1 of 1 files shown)


meaghanfitzgerald (approved):
lgtm, thanks for writing this script!

darioush (approved):
where is this script intended to be invoked from?

PR #3397, opened 2024-09-18
ACP-77: Update `ConvertSubnetTx`

StephenButtolph:

PR #3518, opened 2024-11-05
ACP-77: Implement ids.ID#Append

StephenButtolph:
## Why this should be merged

Factored out of #3397.

This is used to generate `validationID`s according to [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/e333b335c34c8692d84259d21bd07b2bb849dc2c/ACPs/77-reinventing-subnets#convertsubnettol1tx).

## How this works

Adds a new helper `Append` to `ids.ID` - similar to `Prefix`.

## How this was tested

Added tests for both `Prefix` and `Append`.

No.


yacovm (approved):

ceyonur (approved):

ids/id.go
```diff
@@ -107,6 +107,22 @@ func (id ID) Prefix(prefixes ...uint64) ID {
 	return hashing.ComputeHash256Array(packer.Bytes)
 }
 
+// Append this id to create a more selective id.
```
darioush, line 110:
```suggestion
// Append suffixes to this id and re-hash the resulting bytes to create a more selective id.
```

ids/id.go
```diff
@@ -107,6 +107,22 @@ func (id ID) Prefix(prefixes ...uint64) ID {
...
+//
+// This is used to generate the ACP-77 validationIDs.
```
michaelkaplan13, line 112:
```suggestion
// This is used to generate the ACP-77 validationIDs.
// This will return a new id and not modify the original id.
```

darioush (approved):

michaelkaplan13 (approved):

PR #3520, opened 2024-11-05
ACP-103: Document and update genesis test fee configs

StephenButtolph:
## Why this should be merged

While these are not the final values, this change documents the meanings of each parameter and makes the default values (used in our tests) more useful.

Currently, It is possible for the tx fee to increase instantaneously by a factor of `e^200`. Which is just not suitable for testing.

## How this works

Significantly increases the `MaxPerSecond`, `TargetPerSecond`, and `ExcessConversionConstant` to produce more stable fees (using the same `MaxCapacity`).

One unit test needed to be modified because it didn't properly account for `MaxCapacity`.

## How this was tested

N/A

No.

### DIFF
--- a/genesis/genesis_fuji.go
+++ b/genesis/genesis_fuji.go
@@ -46,7 +46,7 @@ var (
 				MaxPerSecond:    250_000,
 				TargetPerSecond: 125_000, // Target block size ~125KB
 				MinPrice:        1,
-				// ExcessConversionConstant = (Capacity - Target) * NumberOfSecondsPerDoubling / ln(2)
+				// ExcessConversionConstant = (MaxPerSecond - TargetPerSecond) * NumberOfSecondsPerDoubling / ln(2)
 				//
 				// ln(2) is a float and the result is consensus critical, so we
 				// hardcode the result.
--- a/genesis/genesis_fuji.go
+++ b/genesis/genesis_fuji.go
@@ -42,11 +42,15 @@ var (
 					gas.DBWrite:   1,
 					gas.Compute:   1,
 				},
-				MaxCapacity:              1_000_000,
-				MaxPerSecond:             1_000,
-				TargetPerSecond:          500,
-				MinPrice:                 1,
-				ExcessConversionConstant: 5_000,
+				MaxCapacity:     1_000_000, // Max block size ~1MB
+				MaxPerSecond:    250_000,
+				TargetPerSecond: 125_000, // Target block size ~125KB
+				MinPrice:        1,
+				// ExcessConversionConstant = (Capacity - Target) * NumberOfSecondsPerDoubling / ln(2)
+				//
+				// ln(2) is a float and the result is consensus critical, so we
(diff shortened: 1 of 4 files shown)


genesis/genesis_local.go
```diff
@@ -60,11 +60,15 @@ var (
...
-				MaxCapacity:              1_000_000,
-				MaxPerSecond:             1_000,
-				TargetPerSecond:          500,
-				MinPrice:                 1,
-				ExcessConversionConstant: 5_000,
+				MaxCapacity:     1_000_000, // Max block size ~1MB
+				MaxPerSecond:    250_000,
+				TargetPerSecond: 125_000, // Target block size ~125KB
+				MinPrice:        1,
+				// ExcessConversionConstant = (Capacity - Target) * NumberOfSecondsPerDoubling / ln(2)
```
yacovm, line 67: (1000000 - 125000) * (30 / ln(2)) is not 5.4 million.

StephenButtolph: Sorry, was supposed to read as `Capacity (per second) - Target (per second)` (like in the SoV calculations below).

Updated to use the exact names in the config.

StephenButtolph: (The calculation is: `(250_000 - 125_000) * (30 / ln(2))`

genesis/genesis_fuji.go
```diff
@@ -42,11 +42,15 @@ var (
...
-				ExcessConversionConstant: 5_000,
+				MaxCapacity:     1_000_000, // Max block size ~1MB
+				MaxPerSecond:    250_000,
+				TargetPerSecond: 125_000, // Target block size ~125KB
+				MinPrice:        1,
+				// ExcessConversionConstant = (MaxPerSecond - TargetPerSecond) * NumberOfSecondsPerDoubling / ln(2)
+				//
+				// ln(2) is a float and the result is consensus critical, so we
+				// hardcode the result.
+				ExcessConversionConstant: 5_410_106, // Double every 30s
```
darioush, line 53: Should we remove the `// TODO` comment above if we consider these values reasonable (or are the gas weights still TODO)?

yacovm (approved):

geoff-vball (approved):

genesis/genesis_fuji.go
```diff
@@ -42,11 +42,15 @@ var (
...
 					gas.Compute:   1,
 				},
-				MaxCapacity:              1_000_000,
-				MaxPerSecond:             1_000,
-				TargetPerSecond:          500,
-				MinPrice:                 1,
-				ExcessConversionConstant: 5_000,
+				MaxCapacity:     1_000_000, // Max block size ~1MB
+				MaxPerSecond:    250_000,
+				TargetPerSecond: 125_000, // Target block size ~125KB
```
darioush, line 47: For my understanding, what is the relation between block size and gas consumption?

darioush (approved):

PR #3521, opened 2024-11-05
ACP-77: Allow legacy validator removal after conversion

StephenButtolph:
## Why this should be merged

ACP-77 was updated to allow removal of legacy subnet validators after the conversion.

## How this works

Removes the checks.

## How this was tested

updates the test

No.


vms/platformvm/txs/executor/standard_tx_executor_test.go
```diff
@@ -1997,39 +2009,6 @@ func TestStandardExecutorRemoveSubnetValidatorTx(t *testing.T) {
...
-						FlowChecker:  env.flowChecker,
-						Ctx:          &snow.Context{},
-					},
-					Tx:    env.tx,
-					State: env.state,
-				}
-				e.Bootstrapped.Set(true)
-				return env.unsignedTx, e
-			},
-			expectedErr: ErrRemoveValidatorManagedSubnet,
```
darioush, line 2031: Is it possible to retain the test with `expectedErr` as nil instead?

StephenButtolph: The first test in the vector was modified to do this

ceyonur (approved):

PR #3522, opened 2024-11-05
ACP-77: Refactor e2e test

StephenButtolph:
## Why this should be merged

This test is heavily extended by all of the remaining ACP-77 PRs. To simplify the review process, the non-behavioral changes that were included in those PRs were refactored into this single PR.

## How this works

Just moving code around to make the test more extendable.

## How this was tested

N/A

No.


tests/e2e/p/l1.go
```diff
@@ -0,0 +1,146 @@
...
+			_, err := pWallet.IssueConvertSubnetTx(
+				subnetID,
+				chainID,
+				address,
+				tc.WithDefaultContext(),
+			)
+			require.NoError(err)
+		})
+
+		tc.By("verifying the Permissioned Subnet was converted to an L1", func() {
```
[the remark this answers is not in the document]
StephenButtolph: The followup PRs include multiple inner calls. (which is why I split it out like this.) If you'd prefer for this PR to have them be just a single call and then break it into multiple calls then I'm happy to change that

tests/e2e/p/l1.go
```diff
@@ -0,0 +1,146 @@
...
+				tc.WithDefaultContext(),
+			)
+			require.NoError(err)
+
+			chainID = chainTx.ID()
+		})
+
+		address := []byte{}
+		tc.By("issuing a ConvertSubnetTx", func() {
+			_, err := pWallet.IssueConvertSubnetTx(
```
michaelkaplan13, line 116: (No action needed) Curious/Surprised that `IssueConvertSubnetTx` doesn't take the initial L1 validators...is this just because the test is for the old spec that included `RecoverSubnetTx`?

michaelkaplan13 (approved):
LGTM, just one question regarding `IssueConvertSubnetTx`

tests/e2e/p/l1.go
StephenButtolph: Correct. This PR was factored out of #3397 which adds the initial L1 validators.

darioush (approved):

PR #3530, opened 2024-11-08
Add `platform.getProposedHeight` API

iansuvak:
## Why this should be merged
Together with `platform.GetValidatorsAt` this allows for processes constructing warp signatures to more reliably construct it using a validator set against which it will actually be evaluated.  Supersedes #3497 

## How this works
Adds `GetProposedHeight` methods to the service and the client. 
## How this was tested
New unit test
Added an entry


vms/platformvm/service.go
```diff
@@ -99,6 +99,20 @@ func (s *Service) GetHeight(r *http.Request, _ *struct{}, response *api.GetHeigh
 	return err
 }
 
+func (s *Service) GetProposedHeight(r *http.Request, _ *struct{}, reply api.GetHeightResponse) error {
```
cam-schultz, line 102:
```suggestion
func (s *Service) GetProposedHeight(r *http.Request, _ *struct{}, reply *api.GetHeightResponse) error {
```
This needs to be a reference parameter, no?

StephenButtolph (approved):

PR #3531, opened 2024-11-08
Add "proposed" optional flag to `getValidatorsAt`

iansuvak:
## Why this should be merged
I'm not convinced that it should but opening this up for discussion. 

Adding this interface was discussed together with adding `platform.GetProposedHeight` interface in #3530 This is a convenience method to allow for passing "proposed" instead of a numeric height literal to `GetValidatorsAt`. 

The idea is to make it safer to publicly expose `GetValidatorsAt` interface. This by itself doesn't accomplish it but a different proxy service could use this same interface and call `GetProposedHeight` and return the cached value if this was called with the same resolved height and subnet. 

## How this works
Makes the GetValidatorsAt accept either numeric height or a "proposed" string literal that first calls `getMinimumHeight` and uses that instead of the numerical value. The service side is backward compatible but the client interface has changed. 

I'm also not sure if this solution is cleaner than just adding a new boolean flag field to the request. I don't think that the `platformVM` service uses this kind of type overloading anywhere else. 
  
## How this was tested
Added a unit test.

Added a note to the `RELEASES.md` but this also introduces a breaking change to the client and not sure what's the best way to handle communication around that

### DIFF
--- a/tests/e2e/p/validator_sets.go
+++ b/tests/e2e/p/validator_sets.go
@@ -16,9 +16,10 @@ import (
 	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/vms/platformvm"
-	platformapi "github.com/ava-labs/avalanchego/vms/platformvm/api"
 	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
 	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+
+	platformapi "github.com/ava-labs/avalanchego/vms/platformvm/api"
 )
 
 var _ = e2e.DescribePChain("[Validator Sets]", func() {


iansuvak:
> * Why was it necessary to propose a replacement for 3497 instead of evolving that PR?

The pivot to this approach was to re-use the already existing field and make it more easily cacheable on the backend and keep the number of externally exposed RPC methods limited if they are accomplishing similar things. Given the pivot in strategy we opted for a new PR.

> * I think it would be cleaner to add a new field instead of adding the complexity of a multi-type field. What suggested this approach?

I'm open to either solution. The reasoning behind accepting multi-type field was since it's already an established pattern in the industry, see [eth_getBlockByNumber](https://www.quicknode.com/docs/ethereum/eth_getBlockByNumber) which accepts `latest`, `earliest` etc. but if the consensus is to have a separate field I'd be happy to add a separate optional flag or boolean field

vms/platformvm/service.go
```diff
@@ -1689,7 +1689,7 @@ func (s *Service) GetTimestamp(_ *http.Request, _ *struct{}, reply *GetTimestamp
 
 // GetValidatorsAtArgs is the response from GetValidatorsAt
 type GetValidatorsAtArgs struct {
-	Height   avajson.Uint64 `json:"height"`
+	Height   avajson.Height `json:"height"`
```
ceyonur, line 1692: I guess this is to prevent breaking change for that args. However I wonder if a separate bool flag or even using negative values would make this more straightforward.

ceyonur (commented):
generally LGTM, we probably need to update RELEASES.md

vms/platformvm/service.md
```diff
@@ -1771,13 +1771,14 @@ Get the validators and their weights of a Subnet or the Primary Network at a giv
 ```sh
 platform.getValidatorsAt(
     {
-        height: int,
+        height: string,
```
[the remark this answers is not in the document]
iansuvak: I updated to `[string|int]` if we decide to stick with this approach

vms/platformvm/service.md
```diff
@@ -1771,13 +1771,14 @@ Get the validators and their weights of a Subnet or the Primary Network at a giv
...
         subnetID: string, // optional
     }
 )
 ```
 
-- `height` is the P-Chain height to get the validator set at.
+- `height` is the P-Chain height to get the validator set at, or the string literal "proposed"
```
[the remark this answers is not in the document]
iansuvak: Added

vms/platformvm/service.go
iansuvak: Yeah indeed, that + custom parsing is to allow either json field. The idea was that eth supports this kind of thing already in [eth_getBlockByNumber](https://www.quicknode.com/docs/ethereum/eth_getBlockByNumber)

iansuvak: But happy to hear the votes either way!

ceyonur: It's probably better to have backwards compatibility, so I'm fine with this or using negative values. Should we also add `latest` here?

utils/json/height.go
```diff
@@ -0,0 +1,32 @@
...
+// See the file LICENSE for licensing terms.
+
+package json
+
+import "strconv"
+
+type Height struct {
+	Numeric    Uint64
+	IsProposed bool
+}
```
StephenButtolph, line 11: Similarly to Cey's comments, I wonder if rather than using a struct for this if it would be nicer to have special values for `uint64`.

Specifically - `type Height uint64`

- When json marshaling - `Height == MaxUint64` gets mapped to `proposed`
- When json unmarshaling - `proposed` gets mapped to `MaxUint64`
- When json unmarshaling - `MaxUint64` is considered invalid

That way in the client we could just pass in `api.ProposedHeight` rather than adding a new bool arg.

(Also I think this should probably live in the `api` package rather than the `json` package, this feels more API specific than general to json helpers)

vms/platformvm/client.go
```diff
@@ -119,7 +121,7 @@ type Client interface {
 	// GetTimestamp returns the current chain timestamp
 	GetTimestamp(ctx context.Context, options ...rpc.Option) (time.Time, error)
 	// GetValidatorsAt returns the weights of the validator set of a provided
-	// subnet at the specified height.
+	// subnet at the specified height or at proposerVM height if [isProposed] is true.
```
StephenButtolph, line 124: I think this comment needs to be updated now.

Also - thoughts on changing `height uint64` to `height api.Height`? It feels like right now the `MaxUint64` behavior might feel odd

vms/platformvm/service.go
```diff
@@ -1876,6 +1876,17 @@ func (s *Service) GetValidatorsAt(r *http.Request, args *GetValidatorsAtArgs, re
...
+	var height uint64
+
+	if args.Height.IsProposed() {
+		height, err = s.vm.GetMinimumHeight(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get proposed height: %w", err)
+		}
+	} else {
+		height = uint64(args.Height)
+	}
```
StephenButtolph, line 1888: Super small nit
```suggestion
	height := uint64(args.Height)
	if args.Height.IsProposed() {
		height, err = s.vm.GetMinimumHeight(ctx)
		if err != nil {
			return fmt.Errorf("failed to get proposed height: %w", err)
		}
	}
```

StephenButtolph (commented):
Couple small nits then lgtm

vms/platformvm/client.go
iansuvak: I made the change since I don't feel strongly one way or another. I agree that the behavior might be unintuitive but accepting `height uint64` had the benefit of not changing the interface at all.

StephenButtolph (approved):

vms/platformvm/service_test.go
```diff
@@ -819,6 +819,99 @@ func TestGetCurrentValidators(t *testing.T) {
...
+		service.vm.ctx.AVAXAssetID,
+		rewardsOwner,
+		rewardsOwner,
+		0,
+		common.WithMemo([]byte{}),
+	)
+
+	require.NoError(err)
+
+	service.vm.ctx.Lock.Unlock()
```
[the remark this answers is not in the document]
iansuvak: I was just sandwiching the `vm.Network.IssueTxFromRPC` following call between `service.vm.ctx.Lock.Unlock/Lock` to indicate that we need it to be unlocked for that.

PR #3545, opened 2024-11-13
Clarify benched field

samliok:
## Why this should be merged

Clarifies readme

## How this works

N/A

## How this was tested

N/A

No

### DIFF
--- a/api/info/service.md
+++ b/api/info/service.md
@@ -539,7 +539,7 @@ info.peers({
 - `version` shows which version the peer runs on.
 - `lastSent` is the timestamp of last message sent to the peer.
 - `lastReceived` is the timestamp of last message received from the peer.
-- `benched` shows chain IDs that the peer is currently benching.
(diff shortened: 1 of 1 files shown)


api/info/service.md
```diff
@@ -539,7 +539,7 @@ info.peers({
 - `version` shows which version the peer runs on.
 - `lastSent` is the timestamp of last message sent to the peer.
 - `lastReceived` is the timestamp of last message received from the peer.
-- `benched` shows chain IDs that the peer is being benched.
+- `benched` shows chain IDs that the peer is currently benching.
```
StephenButtolph, line 542: So, the `benched` field is the list of chainIDs that have benched this peer due to being unresponsive. Would it make more sense to describe this as
```suggestion
- `benched` shows chain IDs that the peer is currently benched on.
```
? Not really happy with any of the options thus far

yacovm (approved):

StephenButtolph (approved):

PR #3548, opened 2024-11-14
ACP-103: Finalize complexity calculations

StephenButtolph:
## Why this should be merged

This PR resolves the TODOs in tx complexity calculations for [ACP-103](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/103-dynamic-fees/README.md).

## How this works

Introduces the complexity dimension (loosely treating 1 complexity unit as 1 micro-second of CPU time.)

```
TimePerSECP256K1Verification = 200 us
TimePerBLSAggregation        = 5 us
TimePerBLSVerification       = 1000 us

GasPerSECP256K1Verification = TimePerSECP256K1Verification * GasPerComputeTime =   800 gas/signature
GasPerBLSAggregation        = TimePerBLSAggregation * GasPerComputeTime        =    20 gas/key
GasPerBLSVerification       = TimePerBLSVerification * GasPerComputeTime       = 4_000 gas/signature

GasOverheadPerWarp     = 20 * GasPerDBRead
GasPerWarpVerification = GasPerBLSAggregation * numKeys + GasPerBLSVerification + GasOverheadPerWarp
```

This PR also updates and documents reads and writes for various transactions

## How this was tested

- [X] Updated unit tests
- [X] Expanded and added benchmarks

No.


vms/platformvm/txs/fee/complexity.go
```diff
@@ -64,6 +65,8 @@ const (
 	intrinsicSECP256k1FxSignatureBandwidth = wrappers.IntLen + // signature index
 		secp256k1.SignatureLen // signature length
```
yacovm, line 67: maybe state which AWS hardware was used to set these operation durations?

vms/platformvm/txs/fee/complexity.go
```diff
@@ -103,42 +112,32 @@ var (
...
-		gas.DBRead:  0,
-		gas.DBWrite: 1,
-		gas.Compute: 0,
+		gas.DBWrite: 1, // write subnet owner
 	}
 	IntrinsicImportTxComplexities = gas.Dimensions{
 		gas.Bandwidth: IntrinsicBaseTxComplexities[gas.Bandwidth] +
 			ids.IDLen + // source chainID
 			wrappers.IntLen, // num importing inputs
-		gas.DBRead:  0,
```
yacovm, line 121: what about the imported inputs?

yacovm: is that handled as part of I/O complexity?

StephenButtolph: These are added here: https://github.com/ava-labs/avalanchego/blob/acp-103-finalize-complexities/vms/platformvm/txs/fee/complexity.go#L588

vms/platformvm/txs/fee/complexity.go
```diff
@@ -103,42 +112,32 @@ var (
...
 		gas.Bandwidth: IntrinsicBaseTxComplexities[gas.Bandwidth] +
 			ids.NodeIDLen + // nodeID
 			ids.IDLen + // subnetID
 			wrappers.IntLen + // subnetAuth typeID
 			wrappers.IntLen, // subnetAuthCredential typeID
-		gas.DBRead:  2,
-		gas.DBWrite: 1,
-		gas.Compute: 0,
+		gas.DBRead:  1, // read subnet auth
+		gas.DBWrite: 3, // delete validator + write weight diff + write pk diff
```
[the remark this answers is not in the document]
yacovm: let me guess it's covered by `AuthComplexity`?

vms/platformvm/txs/fee/complexity.go
```diff
@@ -182,9 +178,6 @@ var (
 			wrappers.IntLen + // number of inputs
 			wrappers.IntLen + // length of memo
 			wrappers.IntLen, // number of credentials
-		gas.DBRead:  0,
-		gas.DBWrite: 0,
-		gas.Compute: 0,
 	}
 	IntrinsicConvertSubnetTxComplexities = gas.Dimensions{
```
yacovm, line 182: this might seem like an overkill and redundant, so i'm not sure it's worth investing the time, but - would it not be possible to annotate the functions in `standard_tx_executor.go` with "W" and "R" to have better book-keeping of this file and where we pay for each R/W operation?

vms/platformvm/txs/fee/complexity.go
```diff
@@ -194,42 +187,38 @@ var (
...
 	IntrinsicDisableSubnetValidatorTxComplexities = gas.Dimensions{
 		gas.Bandwidth: IntrinsicBaseTxComplexities[gas.Bandwidth] +
 			ids.IDLen + // validationID
 			wrappers.IntLen + // auth typeID
 			wrappers.IntLen, // authCredential typeID
-		gas.DBRead:  0, // TODO
-		gas.DBWrite: 0, // TODO
-		gas.Compute: 0,
+		gas.DBRead:  1, // read staker
+		gas.DBWrite: 6, // write remaining balance utxo + weight diff + deactivated weight diff + public key diff + delete staker + write staker
```
yacovm, line 221: could not find the weight diff and PK diff. can you point out where they are?

vms/platformvm/txs/fee/complexity.go
```diff
@@ -103,42 +112,32 @@ var (
 			wrappers.IntLen + // genesis length
 			wrappers.IntLen + // subnetAuth typeID
 			wrappers.IntLen, // subnetAuthCredential typeID
-		gas.DBRead:  1,
-		gas.DBWrite: 1,
-		gas.Compute: 0,
+		gas.DBRead:  3, // get subnet auth + check for subnet transformation + check for subnet conversion
+		gas.DBWrite: 1, // put chain
```
yacovm, line 116: i tried following `CreateChainTx` and oddly enough, i don't see that it [writes the chain](https://github.com/ava-labs/avalanchego/blob/92f0f596f154233b9d9a1e3288b2b32a964c7a16/vms/platformvm/txs/executor/standard_tx_executor.go#L235) to the DB, only to memory. Can you point out where it writes it to the DB?

vms/platformvm/txs/fee/complexity.go
StephenButtolph: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/state.go#L2607

vms/platformvm/txs/fee/complexity.go
StephenButtolph: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/state.go#L3041

vms/platformvm/txs/fee/complexity.go
yacovm: nevermind, since some of the writes are being performed via `state.write()` i think it's not that feasible.

vms/platformvm/txs/fee/complexity.go
```diff
@@ -149,29 +148,26 @@ var (
...
 		gas.Bandwidth: IntrinsicBaseTxComplexities[gas.Bandwidth] +
 			intrinsicValidatorBandwidth + // validator
 			ids.IDLen + // subnetID
 			wrappers.IntLen + // num stake outs
 			wrappers.IntLen, // delegator rewards typeID
-		gas.DBRead:  1,
-		gas.DBWrite: 1,
-		gas.Compute: 0,
+		gas.DBRead:  1, // get staking config
+		gas.DBWrite: 2, // put current staker + write weight diff
```
yacovm, line 161: don't both this and add validator happen via `putStaker` ?  why not pk diff here too?

vms/platformvm/txs/fee/complexity.go
```diff
@@ -72,6 +77,11 @@ const (
 		wrappers.IntLen + // deactivation owner threshold
 		wrappers.IntLen // deactivation owner num addresses
 
+	intrinsicBLSAggregateCompute = 5     // BLS public key aggregation time is around 5us
+	intrinsicBLSVerifyCompute    = 1_000 // BLS verification time is around 1000us
```
yacovm, line 81: by the way I just realized the verifying a PoP also involves un-compressing a compressec EC key and an EC signature. 

I guess we could say 1ms is good enough for both of them, but it's worth to mention it in code review since we also count the aggregation (which is linear, but still...)

yacovm (approved):

vms/platformvm/txs/fee/complexity.go
StephenButtolph: Delegator changes can never result in the weight going from `x -> 0` or `0 -> x` so it can never modify the public key.

vms/platformvm/txs/fee/complexity.go
StephenButtolph: Added the time to decompress the key, the signature decompression time should be considered part of the signature verification time :+1:.

yacovm (approved):

vms/platformvm/txs/fee/complexity.go
```diff
@@ -72,6 +77,13 @@ const (
 		wrappers.IntLen + // deactivation owner threshold
 		wrappers.IntLen // deactivation owner num addresses
 
+	intrinsicBLSAggregateCompute           = 5     // BLS public key aggregation time is around 5us
+	intrinsicBLSVerifyCompute              = 1_000 // BLS verification time is around 1000us
+	intrinsicBLSPublicKeyValidationCompute = 50    // BLS public key validation time is around 50us
+	intrinsicBLSPoPVerifyCompute           = intrinsicBLSPublicKeyValidationCompute + intrinsicBLSVerifyCompute
+
+	intrinsicWarpDBReads = 3 + 20 // chainID -> subnetID mapping + apply weight diffs + apply pk diffs + diff application reads
```
michaelkaplan13, line 85: For my own understanding, where does the 3 and 20 come from exactly?

StephenButtolph: - 1 DB read to map the chainID to the subnetID
- 1 DB read to create the weight diff iterator
- 1 DB read to create the pk diff iterator
- 20 DB reads (heuristic) to apply the diffs

michaelkaplan13 (approved):
LGTM aside from merge conflicts

PR #3552, opened 2024-11-16
Standardize `api.Height` handling

StephenButtolph:
## Why this should be merged

> By convention, to approximate the behavior of [Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) itself, Unmarshalers implement UnmarshalJSON([]byte("null")) as a no-op.

ref: https://pkg.go.dev/encoding/json#Unmarshaler

## How this works

If `null` is passed in, then we return immediately.

## How this was tested

This changed all of the existing ad-hoc tests into test vectors which (I think) are easier to see what exactly the expected behavior is (and what is being tested)

No

### DIFF
--- a/vms/platformvm/api/height.go
+++ b/vms/platformvm/api/height.go
@@ -4,17 +4,16 @@
 package api
 
 import (
-	"encoding/json"
 	"errors"
 	"math"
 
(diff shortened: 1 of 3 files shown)


iansuvak (approved):
Thank you for the standardization fix and test improvements. I'll do vectorized tests from now on.

PR #3554, opened 2024-11-19
[antithesis] Enable reuse of banff e2e test for antithesis testing

marun:
## Why this should be merged

Having to duplicate existing test coverage to benefit from execution with antithesis is not ideal. Better to be able to trivially refactor compatible e2e coverage for reuse. This PR refactors the banff e2e test for reuse with antithesis to demonstrate that this is possible. 

## How this works

 - adds new `APITestFunction` type that a compatible e2e test can implement to allow for execution with both ginkgo and antithesis
 - adds `ExecuteAPITest` helper to simplify execution of an `APITestFunction` with ginkgo
 - Refactors the banff e2e test into an `APITestFunction` executed by `ExecuteAPITest`
 - Refactors the avalanchego antithesis test setup to support execution of the banff test
   - Moves test dispatch into a new function so that panics can be recovered via a deferred function
      - This supports use of require by e2e tests without risking an unhandled panic exiting the test process

## How this was tested

CI

N/A

### DIFF
--- a/tests/antithesis/avalanchego/main.go
+++ b/tests/antithesis/avalanchego/main.go
@@ -218,7 +218,11 @@ func (w *workload) executeTest(ctx context.Context) {
 	defer tc.Recover()
 	require := require.New(tc)
 
-	val, err := rand.Int(rand.Reader, big.NewInt(5))
+	// Ensure this value matches the number of tests + 1 to offset
+	// 0-based + 1 for sleep case in the switch statement for flowID
+	testCount := int64(6)
+
+	val, err := rand.Int(rand.Reader, big.NewInt(testCount))
 	require.NoError(err, "failed to read randomness")
 
 	flowID := val.Int64()
--- a/tests/antithesis/avalanchego/main.go
+++ b/tests/antithesis/avalanchego/main.go
@@ -170,7 +170,7 @@ func (w *workload) run(ctx context.Context) {
 	tc := w.newTestContext(ctx)
 	// Any assertion failure from this test context will result in process exit due to the
 	// panic being rethrown. This ensures that failures in test setup are fatal.
-	defer tc.Recover(true /* rethrow */)
+	defer tc.RecoverAndRethrow()
 
 	xAVAX, pAVAX := e2e.GetWalletBalances(tc, w.wallet)
 	assert.Reachable("wallet starting", map[string]any{
@@ -215,7 +215,7 @@ func (w *workload) run(ctx context.Context) {
 func (w *workload) executeTest(ctx context.Context) {
 	tc := w.newTestContext(ctx)
 	// Panics will be recovered without being rethrown, ensuring that test failures are not fatal.
-	defer tc.Recover(false /* rethrow */)
+	defer tc.Recover()
 	require := require.New(tc)
 
 	val, err := rand.Int(rand.Reader, big.NewInt(5))
--- a/tests/antithesis/avalanchego/main.go
+++ b/tests/antithesis/avalanchego/main.go
@@ -45,7 +45,6 @@ import (
 
 const NumKeys = 5
 
-// TODO(marun) Switch to using zap for logging
 // TODO(marun) Extract the common elements of test execution for reuse across test setups
 
 func main() {
@@ -223,7 +222,6 @@ func (w *workload) executeTest(ctx context.Context) {
 	val, err := rand.Int(rand.Reader, big.NewInt(6))
 	require.NoError(err, "failed to read randomness")
 
-	// TODO(marun)
 	flowID := val.Int64()
 	switch flowID {
 	case 0:
--- a/tests/antithesis/avalanchego/main.go
+++ b/tests/antithesis/avalanchego/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/tests"
 	"github.com/ava-labs/avalanchego/tests/antithesis"
+	"github.com/ava-labs/avalanchego/tests/e2e/banff"
 	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
 	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 	"github.com/ava-labs/avalanchego/utils/constants"
@@ -44,10 +45,13 @@ import (
 
 const NumKeys = 5
 
+// TODO(marun) Switch to using zap for logging
+// TODO(marun) Extract the common elements of test execution for reuse across test setups
+
 func main() {
 	// TODO(marun) Support choosing the log format
-	tc := tests.NewTestContext(tests.NewDefaultLogger(""))
-	defer tc.Cleanup()
+	tc := antithesis.NewInstrumentedTestContext(tests.NewDefaultLogger(""))
+	defer tc.RecoverAndExit()
 	require := require.New(tc)
 
 	c := antithesis.NewConfig(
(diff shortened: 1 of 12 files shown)


tests/simple_test_context.go
```diff
@@ -10,83 +10,165 @@ import (
...
+}
+
+// Recover is intended to be deferred in a function executing a test
+// whose assertions may result in panics. Such a panic is intended to
+// be recovered to allow cleanup functions to be called. A panic can
+// be optionally rethrown by setting `rethrow` to true.
+func (tc *SimpleTestContext) Recover(rethrow bool) {
+	// Recover from test failure
+	var panicData any
+	if panicData := recover(); panicData != nil {
```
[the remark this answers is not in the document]
maru-ava: The original var needs to be set rather than creating a new one.

tests/simple_test_context.go
```diff
@@ -10,83 +10,165 @@ import (
...
-func (tc *SimpleTestContext) By(_ string, _ ...func()) {
-	tc.Errorf("By not yet implemented")
-	tc.FailNow()
+func (tc *SimpleTestContext) By(msg string, callback ...func()) {
+	tc.log.Info("Step: " + msg)
+
+	if len(callback) == 1 {
+		callback[0]()
+	} else if len(callback) > 1 {
+		tc.Errorf("just one callback per By, please")
```
RodrigoVillar, line 183: q: Is there a reason why only one callback is supported?

tests/simple_test_context.go
```diff
@@ -10,83 +10,165 @@ import (
...
+	if exitNonZero {
+		os.Exit(1)
+	}
+}
+
+// Recover is intended to be deferred in a function executing a test
+// whose assertions may result in panics. Such a panic is intended to
+// be recovered to allow cleanup functions to be called. A panic can
+// be optionally rethrown by setting `rethrow` to true.
+func (tc *SimpleTestContext) Recover(rethrow bool) {
```
RodrigoVillar: The documentation here is clear but the usage of `Recover` is unintuitive IMO. When I took a first stab at this, reading `defer tc.Recover(false)` make me think that a panic will terminate the process, when in reality it just means that the panic will not be rethrown (the process will recover from the panic). 

One idea I have is to split this function into `Recover` and `RecoverAndRethrow`, but also open to other suggestions or why we should keep `Recover` as is.

tests/test_context.go
```diff
@@ -31,6 +31,9 @@ type TestContext interface {
 	DefaultContext() context.Context
 	WithDefaultContext() common.Option
 
+	// The parent context (if non-nil) to use as the parent of default contexts
+	GetDefaultContextParent() context.Context
```
RodrigoVillar, line 35: (nit / no-action-necessary) I would expect `GetDefaultContextParent()` to return `context.Background()` or something similar by virtue of the `Default` keyword. However, we set this value to the top-level context (e.g. context with cancel on `SIGTERM`) in places like the Antithesis tests. Perhaps rename this to `GetContextParent()`?

tests/simple_test_context.go
maru-ava: This mirrors what ginkgo supports.

tests/test_context.go
maru-ava: Is the comment not clear?

tests/simple_test_context.go
maru-ava: Done

tests/test_context.go
RodrigoVillar: After further reflection, I understand the comment now (`GetDefaultContextParent` returns the source context for the context methods above).

maru-ava:
Approving my own PR is a treat (because my personal account proposed it), but I'll hold off on merging until someone else approves.

tests/load2/generator.go
```diff
@@ -87,10 +88,18 @@ func (l LoadGenerator) Run(
...
 		})
 	}
 
 	_ = eg.Wait()
 }
+
+// execTestWithRecovery ensures assertion-related panics encountered during test execution are recovered
+// and that deferred cleanups are always executed before returning.
+func execTestWithRecovery(ctx context.Context, log logging.Logger, test Test, wallet *Wallet) {
+	tc := tests.NewTestContext(log)
```
maru-ava, line 98: Creating a new TestContext here ensures that recovery executes only the cleanups registered to this context.

tests/load2/generator.go
```diff
@@ -85,10 +85,7 @@ func (l LoadGenerator) Run(
 				default:
 				}
 
-				ctx, cancel := context.WithTimeout(ctx, testTimeout)
```
maru-ava, line 87: Moved to execTestWithRecovery to ensure that cancelation is localized to test execution rather than only being performed on goroutine exit.

maru-ava:
> This way, `Run()` only deals with just one context which is easier to read + having two separate contexts isn't necessary as a result of this PR.

Done

maru-ava:
I'd like to rename SimpleTestContext to RecoverableTestContext, but would prefer to do that in a follow-on PR.

tests/antithesis/avalanchego/main.go
```diff
@@ -176,32 +179,30 @@ func (w *workload) run(ctx context.Context) {
...
+				"worker": w.id,
+				"err":    err,
+			})
+			rawExecutionDelay = defaultExecutionDelay
+		}
+		executionDelay := time.Duration(rawExecutionDelay.Int64())
+		w.log.Info("waiting",
+			zap.Duration("duration", executionDelay),
+		)
+		timer.Reset(executionDelay)
```
StephenButtolph: I don't have a problem with this change, but just wanted to confirm my understanding.

Previously `require.NoError` was utilized on the returned error from reading the randomness. There is now a comment that fetching randomness is not critical.

Is this a required change in this PR? Or was this something that was cleaned up while we were here? If it is required, why is it required? It seems like this sleep in the xsvm test isn't modified.

tests/antithesis/avalanchego/main.go
```diff
@@ -210,6 +211,44 @@ func (w *workload) run(ctx context.Context) {
...
+		w.log.Info("executing issueXChainOperationTx")
+		w.issueXChainOperationTx(ctx)
+	case 3:
+		w.log.Info("executing issueXToPTransfer")
+		w.issueXToPTransfer(ctx)
+	case 4:
+		w.log.Info("executing issuePToXTransfer")
+		w.issuePToXTransfer(ctx)
+	case 5:
+		w.log.Info("sleeping")
```
StephenButtolph, line 231: This case is unreachable
```suggestion
```

tests/test_context.go
```diff
@@ -31,6 +31,9 @@ type TestContext interface {
 	DefaultContext() context.Context
 	WithDefaultContext() common.Option
 
+	// The parent context (if non-nil) to use as the parent of default contexts
```
StephenButtolph: I think it is generally unexpected to pass around `nil` contexts. Can we change the implementations of this interface to return `context.Background()` rather than expecting the callers of this function to do that handling?

I think we just need to change:
```golang
func (*GinkgoTestContext) GetDefaultContextParent() context.Context {
	return nil
}
```

to
```golang
func (*GinkgoTestContext) GetDefaultContextParent() context.Context {
	return context.Background()
}
```

tests/antithesis/avalanchego/main.go
maru-ava: Updated to make this case reachable

tests/test_context.go
maru-ava: Done

tests/antithesis/avalanchego/main.go
maru-ava: After reconsidering, I'll revert this. 

My previous thinking was that failure to read randomness shouldn't be fatal, and using requires.NoError with a panic handler that rethrows would effectively make it fatal.

But given that one of the main selling points of antithesis is deterministic behavior - including randomness - any failure to return randomness would represent a system fault that could preclude reproduction. So, it _should_ be fatal.

StephenButtolph (approved):

PR #3562, opened 2024-11-21
Use JSON logs during Antithesis runs

StephenButtolph:
## Why this should be merged

Currently the antithesis tests perform string matching rather than looking at log levels.

## How this works

Sets the log format to json.

## How this was tested

Manually:
```
avalanche-node-1          | {"level":"trace","timestamp":"2024-11-21T19:39:31.316Z","logger":"P Chain","caller":"snowman/topological.go:594","msg":"accepting block","blkID":"RQSTTsQ9WAurX8M8fiDt6injR3GtiUETEDjHyDLT7L9uhAYBc","height":1,"timestamp":"2024-11-21T19:39:31.000Z"}
```

The logs for the nodes (not the workload) are now json 🥳 

No.

### DIFF
--- a/config/flags.go
+++ b/config/flags.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/compression"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/dynamicip"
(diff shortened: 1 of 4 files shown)


ARR4N (changes requested):
Please add tests

ARR4N (approved):

PR #3563, opened 2024-11-21
Antithesis: Skip checks if tx confirmation fails

StephenButtolph:
## Why this should be merged

The Antithesis tests have been consistently failing because the workload incorrectly skips transaction confirmation.

Specifically:
1. A transaction is issued to the network and marked as accepted on the API node.
2. A node is killed by the antithesis framework.
3. The transaction confirmation check is cancelled because the killed node is unresponsive
4. We attempt to verify that the expected state transition happened even though the transaction may not have been accepted on all of the nodes yet.

[Errors this PR should resolve](https://avalanche.antithesis.com/report/QLVyGekjFsEWhTZOXtyVGwqR/wnp_QMrNeLQuyu99oBTuukfWKjucchUeOYvKoPwGeAE.html?auth=v2.public.eyJuYmYiOiIyMDI0LTExLTE1VDA0OjQ4OjU3WiIsInNjb3BlIjp7IlJlcG9ydFNjb3BlVjEiOnsiYXNzZXQiOiJ3bnBfUU1yTmVMUXV5dTk5b0JUdXVrZldLanVjY2hVZU9ZdktvUHdHZUFFLmh0bWwiLCJyZXBvcnRfaWQiOiJRTFZ5R2VrakZzRVdoVFpPWHR5Vkd3cVIifX1936JcYwGmUSOCq20rYyt_zuTXfQ7aQPNTSnp90lmhIxh24ZlaOwg0V3p8yHVrIAsmqEnXpwnOLl2-yVaVtdo7Cw):
- `failed to verify that X-chain UTXO was deleted`
- `failed to verify that P-chain UTXO was deleted`

## How this works

If the transaction confirmation check is cancelled, we do not perform verification of the expected state transition.

## How this was tested

Ran locally and didn't see any regressions.

No.


tests/antithesis/avalanchego/main.go
```diff
@@ -409,9 +425,26 @@ func (w *workload) issueXChainOperationTx(ctx context.Context) {
...
+	if err := w.confirmXChainTx(ctx, operationTx); err != nil {
+		w.log.Warn("failed to confirm transaction",
+			zap.String("chain", "X"),
+			zap.String("txType", "operation"),
+			zap.Error(err),
+		)
+		return
+	}
+
 	w.verifyXChainTxConsumedUTXOs(ctx, operationTx)
```
samliok, line 448: should this method be refactored to return an error? It seems to just log an error if one occurs during but then returns nothing. 

I could create a follow up PR if so

samliok (approved):

tests/antithesis/avalanchego/main.go
StephenButtolph: Because this function doesn't actually perform any modifications, it isn't _required_ to propagate the error up... So that's why I didn't include that in this PR... But it certainly _could_.

PR #3565, opened 2024-11-22
Support UnmarshalJSON for `ExporterType`

RodrigoVillar:
## Why this should be merged

Currently, if you try calling `json.Marshal()` and `json.Unmarshal()` on a field of type `ExporterType`, this will fail as, when unmarshaling, it will try to convert a string (e.g. `"grpc"`, `"http"`, `"unknown"`) into a byte.

This is solved by adding a useful default value to the type and implementing `UnmarshalJSON`.

## How this works

This PR fixes the above by adding an `UnmarshalJSON()`, which handles conversions from the strings mentioned above to their respective `ExporterType` value.

## How this was tested

Added unit tests.

Modified configs have been documented.


trace/exporter_type.go
```diff
@@ -33,6 +33,16 @@ func (t ExporterType) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + t.String() + `"`), nil
 }
 
+func (t *ExporterType) UnmarshalJSON(b []byte) error {
+	exporterTypeStr := strings.Trim(string(b), `"`)
+	exporterType, err := ExporterTypeFromString(exporterTypeStr)
+	if err != nil && !errors.Is(err, errUnknownExporterType) {
+		return err
+	}
```
aaronbuchwald, line 41: (optional) I think we could remove the special case error check and just return an error on unknown

aaronbuchwald (approved):

trace/exporter_type.go
```diff
@@ -33,6 +33,16 @@ func (t ExporterType) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + t.String() + `"`), nil
 }
 
+func (t *ExporterType) UnmarshalJSON(b []byte) error {
+	exporterTypeStr := strings.Trim(string(b), `"`)
```
StephenButtolph, line 37: Should we be erroring if the quotes are formatted unexpectedly? For instance:
- `grpc`
- `"grpc`
- `""grpc`
- `""grpc"""`

would all be unmarshalled without issue, but non of them are formatted as expected.

trace/exporter_type.go
```diff
@@ -33,6 +33,16 @@ func (t ExporterType) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + t.String() + `"`), nil
 }
 
+func (t *ExporterType) UnmarshalJSON(b []byte) error {
```
StephenButtolph, line 36: By convention, `"null" == string(b)` should be a noop.

StephenButtolph, line 36: Should we add some test vectors for marshal and unmarshal?

RodrigoVillar: Done

RodrigoVillar: Done

trace/exporter_type.go
RodrigoVillar: Done

StephenButtolph (approved):

PR #3567, opened 2024-11-25
Add partial-sync support to the wallet

StephenButtolph:
## Why this should be merged

Currently, it is not easy to create a wallet that can issue transactions against a node that is running `--partial-sync-primary-network`.

This PR allows creating an instance of the P-chain wallet against such a node.

## How this works

- Removes the usage of the X-chain API when fetching the AVAX asset ID.
- Adds a new function `MakePWallet` that does not attempt to fetch UTXOs from the X-chain or C-chain (or shared memory).

## How this was tested

Issuing transactions against a partial synced node.

No.

### DIFF
--- a/wallet/chain/p/context.go
+++ b/wallet/chain/p/context.go
@@ -7,7 +7,7 @@ import (
 	"context"
 
 	"github.com/ava-labs/avalanchego/api/info"
-	"github.com/ava-labs/avalanchego/vms/avm"
+	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/vms/platformvm"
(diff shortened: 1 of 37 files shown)


wallet/chain/p/context.go
```diff
@@ -21,42 +21,40 @@ const gasPriceMultiplier = 2
 
 func NewContextFromURI(ctx context.Context, uri string) (*builder.Context, error) {
 	infoClient := info.NewClient(uri)
-	xChainClient := avm.NewClient(uri, "X")
```
StephenButtolph, line 24: In order to support partial sync, the P-chain wallet can not depend on any X-chain APIs.

yacovm (approved):

PR #3570, opened 2024-11-25
Allow issuing transactions when using partial-sync

StephenButtolph:
## Why this should be merged

Currently if `--partial-sync-primary-network` is set, the `platform.issueTx` API is disabled. This PR allows issuing transactions that can be verified by the node. 

## How this works

When `--partial-sync-primary-network` is enabled, ban `ImportTx`s. Otherwise, allow transactions to be issued as normal.

## How this was tested

Locally issued P-chain transactions to Fuji through a partial synced node.

There is currently no reference to `--partial-sync-primary-network` in the docs for `issueTx`.

### DIFF
--- a/vms/platformvm/block/executor/manager.go
+++ b/vms/platformvm/block/executor/manager.go
@@ -23,7 +23,8 @@ import (
 var (
 	_ Manager = (*manager)(nil)
 
-	ErrChainNotSynced = errors.New("chain not synced")
+	ErrChainNotSynced              = errors.New("chain not synced")
+	ErrImportTxWhilePartialSyncing = errors.New("issuing an import tx is not allowed while partial syncing")
 )
 
 type Manager interface {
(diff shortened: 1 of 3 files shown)


vms/platformvm/block/executor/manager.go
```diff
@@ -124,6 +125,15 @@ func (m *manager) VerifyTx(tx *txs.Tx) error {
 		return ErrChainNotSynced
 	}
 
+	// If partial sync is enabled, this node isn't guaranteed to have the full
```
yacovm, line 128: >  If partial sync is enabled, this node isn't guaranteed to have the full UTXO set from shared memory

Can you explain why? I thought the partial sync flag just skips syncing every chain but the P-chain. But we're still only processing P-chain UTXOs in platformVM... no?

yacovm: oh nevermind I think I got it, we're importing UTXOs from other chains, is that it?

yacovm, line 128: just to double check - this function is only called at the transaction admission path, right? It's not part of the block processing logic, otherwise I guess we need some kind of pre-etna, post etna check?

yacovm (approved):

vms/platformvm/block/executor/manager.go
StephenButtolph: Correct. `VerifyTx` is not used during block execution. From the method comment:

```golang
// VerifyTx verifies that the transaction can be issued based on the currently
// preferred state. This should *not* be used to verify transactions in a block.
```

yacovm: oh, the interface method was not part of the diff so I didn't notice that.

Thanks.

PR #3571, opened 2024-11-25
Remove redundant field in platformVM/network's Network

yacovm:
## Why this should be merged

The field `txVerifier` is unused 

## How this works

Removed the field

## How this was tested

Compile time check

No


StephenButtolph (approved):
🦅 👁️

darioush (approved):

PR #3573, opened 2024-11-26
Expose test network cfg

cam-schultz:
## Why this should be merged
The TestNetwork utility exposes the p2p stack so that non-node applications can communicate with nodes over p2p. This change makes this utility more flexible by allowing the user to modify the default configuration. 

## How this works
Adds `NewTestNetworkConfig` and modifies `NewTestNetwork` to accept a `network.Config` parameter.

## How this was tested
CI

No


geoff-vball (approved):
👍

yacovm (approved):

michaelkaplan13 (approved):

network/test_network.go
```diff
@@ -73,23 +73,12 @@ func (*noopListener) Addr() net.Addr {
...
-func NewTestNetwork(
-	log logging.Logger,
+func NewTestNetworkConfig(
 	networkID uint32,
 	currentValidators validators.Manager,
 	trackedSubnets set.Set[ids.ID],
-	router router.ExternalHandler,
-) (Network, error) {
+) (*Config, error) {
 	metrics := prometheus.NewRegistry()
```
joshua-kim, line 81: I think this should be passed in by the caller so that the same metrics instance can be passed into `NewTestNetwork`

cam-schultz: Good call. Made this change.

joshua-kim (approved):

StephenButtolph (approved):

PR #3574, opened 2024-11-27
move genesis docs to genesis package and add note on initial staker handling

aaronbuchwald:
This PR moves the documentation for genesis JSON into the `genesis/` package and adds an explanation of how funds for `initialStakers` are handled.

Specifically, if an address included in the genesis is marked as an `initialStaker` then the UTXOs described by its `unlockSchedule` are included in the the initial staker transaction instead of creating a UTXO in the genesis state of the P-Chain.

For reference, the flow is:
1. Initial stakers are added to a set [here](https://github.com/ava-labs/avalanchego/blob/94be34ac64cba099d7be198965f0dd551f7cdefb/genesis/genesis.go#L344)
2. When creating initial P-Chain UTXOs, we skip over any address included in `initialStakers` [ref](https://github.com/ava-labs/avalanchego/blob/94be34ac64cba099d7be198965f0dd551f7cdefb/genesis/genesis.go#L357)
3. We create the initial validators with their locked stakes [here](https://github.com/ava-labs/avalanchego/blob/94be34ac64cba099d7be198965f0dd551f7cdefb/genesis/genesis.go#L387)


genesis/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+The genesis JSON file contains the following properties:
+
+- `networkID`: A unique identifier for the blockchain, must be a number in the range [0, 2^32).
+- `allocations`: The list of initial addresses, their initial balances and the unlock schedule for each.
+- `startTime`: The time of the beginning of the blockchain, it must be a Unix
+  timestamp and it can't be a time in the future.
+- `initialStakeDuration`: The stake duration, in seconds, of the validators that exist at network genesis.
+- `initialStakeDurationOffset`: The offset, in seconds, between the start times
+  of the validators that exist at genesis.
```
StephenButtolph, line 13: This was wrong before, the offset is for the endTime - not the startTime.
```suggestion
- `initialStakeDurationOffset`: The offset, in seconds, between the end times
  of the validators that exist at genesis.
```

genesis/README.md
```diff
@@ -0,0 +1,32 @@
...
+Each allocation contains the following fields:
+
+- `ethAddr`: Annotation of corresponding Ethereum address holding an ERC-20 token
+- `avaxAddr`: X/P Chain address to receive the allocation
+- `initialAmount`: Initial unlocked amount minted to the `avaxAddr` on the X-Chain
+- `unlockSchedule`: List of locked, stakeable UTXOs minted to `avaxAddr` on the P-Chain
+
+Note: if an `avaxAddr` from allocations is included in `initialStakers`, the genesis includes
+all the UTXOs specified in the `unlockSchedule` as part of the locked stake of the corresponding
+genesis validator. Otherwise, the locked UTXO is created directly on the P-Chain.
```
StephenButtolph, line 32: All of the funds are use for the genesis validators, not just the tokens in the unlock schedule. That being said, on mainnet, there are no genesis validators with unlocked stake (so this distinction doesn't impact mainnet)
```suggestion
Note: if an `avaxAddr` from allocations is included in `initialStakedFunds`, the genesis allocates all the UTXOs for that `avaxAddr` as part of the stake of the genesis validators; rather than on the X-Chain and P-Chain.
```

genesis/README.md
```diff
@@ -0,0 +1,32 @@
+# Genesis
+
+The Genesis package converts formatted JSON files into the genesis of the Primary Network. For the simplest example, see the [Local Genesis](./genesis_local.json) JSON file.
```
meaghanfitzgerald, line 3:
```suggestion
The Genesis package converts formatted JSON files into the genesis of the Primary Network. For the simplest example, see the [Local Genesis](https://github.com/ava-labs/avalanchego/blob/master/genesis/genesis_local.json) JSON file.
```
This suggestion is only necessary if we also want to import this page into the docs. imo we should cc: @ashucoder9

meaghanfitzgerald:
merging to combine with other edits to config docs

PR #3576, opened 2024-11-27
encapsulate signer

richardpringle:
Depends on https://github.com/ava-labs/coreth/pull/693. But https://github.com/ava-labs/coreth/pull/693 also depends on this. Not sure about merge order... 

## Why this should be merged

Currently, signing happens all over the place and should definitely be abstracted away, especially when dealing with sensitive data like a private key. 

Once this abstraction lands, we could further enhance security by using secure hardware and signing outside of the avalanchego application and potentially integrate platforms like cubist. 

## How this works

Encapsulate signing such that the secret-key does not leave our `bls` utility package. 
The main changes are in this file: `utils/crypto/bls/secret.go`

## How this was tested

CI

I don't think so?


go.mod
```diff
@@ -190,3 +190,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/ava-labs/coreth v0.13.9-rc.1 => github.com/ava-labs/coreth v0.13.9-rc.1.0.20241127225719-c225e43551e1
```
richardpringle, line 194: Not sure what the process is here

richardpringle, line 194: Not sure what the process is here

network/config.go
```diff
@@ -128,7 +128,7 @@ type Config struct {
 	// TLSKey is this node's TLS key that is used to sign IPs.
 	TLSKey crypto.Signer `json:"-"`
 	// BLSKey is this node's BLS key that is used to sign IPs.
-	BLSKey *bls.SecretKey `json:"-"`
+	BLSKey bls.Signer `json:"-"`
```
richardpringle, line 131: I assume this means the `BLSKey` property gets serialized as the string, `"-"`?

chains/manager.go
```diff
@@ -186,7 +186,7 @@ type ManagerConfig struct {
 	SybilProtectionEnabled bool
 	StakingTLSSigner       crypto.Signer
 	StakingTLSCert         *staking.Certificate
-	StakingBLSKey          *bls.SecretKey
+	StakingBLSKey          bls.Signer
```
richardpringle, line 189: Do we want to change the property name here too?

config/config.go
```diff
@@ -694,7 +694,7 @@ func getStakingSigner(v *viper.Viper) (*bls.SecretKey, error) {
 		return nil, fmt.Errorf("couldn't create path for signing key at %s: %w", signingKeyPath, err)
 	}
 
-	keyBytes := bls.SecretKeyToBytes(key)
+	keyBytes := key.ToBytes()
```
richardpringle, line 697: I'll have to change this eventually to fully abstract the signer. I'm open to ideas on how best to do this

network/network_test.go
```diff
@@ -175,7 +175,7 @@ func newTestNetwork(t *testing.T, count int) (*testDialer, []*testListener, []id
 		require.NoError(t, err)
 		nodeID := ids.NodeIDFromCert(cert)
 
-		blsKey, err := bls.NewSecretKey()
+		blsKey, err := bls.NewSigner()
```
richardpringle, line 178: 🤔 If we want to change all the `*Key` names (variables or properties), I would recommend/request that I do that in a follow-up to limit the scope of both changes.

yacovm:
> Once this abstraction lands, we could further enhance security by using secure hardware and signing outside of the avalanchego application and potentially integrate platforms like cubist.


Do they support BLS? Most hardware security modules only support ECDSA / ed25519...

richardpringle:
> > Once this abstraction lands, we could further enhance security by using secure hardware and signing outside of the avalanchego application and potentially integrate platforms like cubist.
> 
> Do they support BLS? Most hardware security modules only support ECDSA / ed25519...

Yeah, they specifically have support for Ethereum validators

utils/crypto/bls/secret.go
```diff
@@ -25,49 +25,64 @@ var (
 
 type SecretKey = blst.SecretKey
```
richardpringle, line 26: I should get rid of this type

network/config.go
joshua-kim: IIRC this actually omits it from serialization, i.e it's not serialized or deserialized. This is done because we use the `Config` struct to also hold fields and not just config values.... so our hack is to ignore serialization for non-config values (which is weird IMO).

richardpringle: Makes sense. 

Yeah, I generally prefer to split up raw config data vs actual dependent services/structs, but 🤷. I've seen this pattern before too.

utils/crypto/bls/secret.go
joshua-kim, line 1: We should probably rename this to something like `signer.go`

utils/crypto/bls/secret.go
```diff
@@ -25,49 +25,64 @@ var (
...
+var _ Signer = (*signer)(nil)
+
+type signer struct {
+	sk *SecretKey
+}
+
 // NewSecretKey generates a new secret key from the local source of
 // cryptographically secure randomness.
-func NewSecretKey() (*SecretKey, error) {
+func NewSigner() (Signer, error) {
```
joshua-kim, line 44: nit: export `signer` and return it as a concrete type here so that the calling code can decide whether to use it as an interface/as a concrete type. We'll have to figure out a fix/good naming scheme since it'll collide w/ the interface name as it currently is though. We could also as an alternative argue that the interface is premature and just use the struct.

utils/crypto/bls/secret.go
richardpringle: My preference would be to make that change in a follow up if you're cool with that? Just trying to make the absolute minimal possible changes here. 

However, it's effectively impossible to mess that up, so if you feel strongly, I can do that before merging this.

utils/crypto/bls/secret.go
richardpringle: 🤔  I don't think the interface is premature though, I don't plan on adding any functionality for the other implementations. What I'm trying to get is encapsulation so that you can't access the secret-key outside of this package. I think want to make a `LocalSigner`, then some "cloud-based" signers (there are a couple of options here that will each have their own concrete types).   

Having it generic from day 1 minimizes the impact of converting what's here to a `LocalSigner` (which would also be used in all tests), and other signer types would be created based on config values on application start. 

Thoughts?

utils/crypto/bls/secret.go
joshua-kim: I think it's fine to do that in whatever PR we do the var renames in

snow/context.go
```diff
@@ -95,3 +99,59 @@ type ConsensusContext struct {
 	// True iff this chain is currently state-syncing
 	StateSyncing utils.Atomic[bool]
 }
+
+func NewTestContext() *Context {
```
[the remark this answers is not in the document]
richardpringle: It's used differently than `snowtest.Context`. I think it makes sense to move this back to `coreth` though

joshua-kim (approved):
Another idea is to cache the key bytes and public key but not going to block on it.

go.mod
```diff
@@ -190,3 +190,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/ava-labs/coreth v0.13.9-rc.1 => github.com/ava-labs/coreth v0.13.9-rc.1.0.20241129155209-a857e7c5a2da
```
richardpringle, line 194:
```suggestion
```

aaronbuchwald (commented):
Could we update to a coreth rc rather than a commit before merging to master? After that, this LGTM

config/config.go
richardpringle: Yeah, looking at the code again, signer creation would be specific. A `LocalSigner` should actually write the key to disk before returning from a constructor. We could have an `EphemeralSigner` that's similar but doesn't save the key and is used for testing. 

With that said, I would probably create a 
```go
func initializeSigner(signerConfig) (Signer) { /* ... */ }
```
and deal with saving or not saving there. It should not be dealt with here IMO. 

This is all for a future PR though

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/DataDog/zstd v1.5.2
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/coreth v0.13.9-rc.1
+	github.com/ava-labs/coreth v0.13.9-rc.1.0.20241204232641-e73e655356b7
```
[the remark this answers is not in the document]
richardpringle: yep

network/config.go
richardpringle: one day...

go.mod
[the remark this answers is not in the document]
richardpringle: done

utils/crypto/bls/secret.go
```diff
@@ -19,55 +19,69 @@ var (
...
-	return sk.Serialize()
+// ToBytes returns the big-endian format of the secret key.
+func (s *LocalSigner) ToBytes() []byte {
+	return s.sk.Serialize()
 }
 
 // SecretKeyFromBytes parses the big-endian format of the secret key into a
 // secret key.
-func SecretKeyFromBytes(skBytes []byte) (*SecretKey, error) {
+func SecretKeyFromBytes(skBytes []byte) (Signer, error) {
```
StephenButtolph, line 62:
```suggestion
func SecretKeyFromBytes(skBytes []byte) (*LocalSigner, error) {
```

utils/crypto/bls/secret.go
```diff
@@ -19,55 +19,69 @@ var (
...
 )
 
 type SecretKey = blst.SecretKey
 
+type Signer interface {
+	PublicKey() *PublicKey
+	Sign(msg []byte) *Signature
+	SignProofOfPossession(msg []byte) *Signature
+	// TODO: delete me
+	ToBytes() []byte
```
StephenButtolph, line 34: If we adopt the suggested change on `SecretKeyFromBytes` - then this can be deleted now.
```suggestion
```

utils/crypto/bls/secret.go
```diff
@@ -19,55 +19,69 @@ var (
...
+	ciphersuiteProofOfPossession        = []byte("BLS_POP_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_")
+	_                            Signer = (*LocalSigner)(nil)
 )
 
 type SecretKey = blst.SecretKey
 
+type Signer interface {
+	PublicKey() *PublicKey
+	Sign(msg []byte) *Signature
+	SignProofOfPossession(msg []byte) *Signature
```
StephenButtolph, line 32: I feel like these functions are going to need to return `error` once we want to use them for remote signing. Are we waiting to add `error` to the return value for when that other implementation is introduced?

richardpringle: Yeah, I don't think there's any point in pre-maturely defining the interface here. It's unclear to me at this point whether or not we'll want the specific signer implementation to handle its own errors or wether we would want the caller to handle errors (I'm missing some details on how to gracefully shutdown). 

I'm interested to hear your thoughts here, but I definitely don't think this is something that we should deal with yet.

utils/crypto/bls/secret.go
richardpringle: done fd7f3268a81550f73917cbe36baa1ccef46a639d

utils/crypto/bls/secret.go
StephenButtolph: Sounds good. I think the only complex question will be around the PoP in the networking code... Everything else should be fairly straightforward I think... But 🤷 we can deal with that when we need to I suppose.

StephenButtolph (approved):

meaghanfitzgerald:
https://github.com/ava-labs/avalanchego/issues/3640 Re:
> Within the config, I'd expect us to have a switch that offers different BLS Signers ie. generate and store a new BLS key on the machine or use an external signing service.
> I'd think this is either AvalancheGo defining the options it wants to support directly here or it supports a gRPC service that dials an external signer and handles signature requests.

PR #3581, opened 2024-12-05
use pebble nosync by default

ceyonur:
## Why this should be merged

Adds an option for sync/nosync in pebble and changed default to `NoSync`

## How this works

Adds a bool for `writeOption` in pebble and performs operations based on the configured `writeOption`

## How this was tested

Locally

Documented

### DIFF
--- a/database/pebbledb/db.go
+++ b/database/pebbledb/db.go
@@ -43,6 +43,7 @@ var (
 		MemTableSize:                defaultCacheSize / 4,
 		MaxOpenFiles:                4096,
 		MaxConcurrentCompactions:    1,
+		Sync:                        false,
 	}
 )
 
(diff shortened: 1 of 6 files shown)


darioush (approved):

database/pebbledb/db.go
```diff
@@ -64,7 +67,7 @@ type Config struct {
 }
 
 // TODO: Add metrics
-func New(file string, configBytes []byte, log logging.Logger, _ prometheus.Registerer) (database.Database, error) {
+func New(file string, useSyncWrites bool, configBytes []byte, log logging.Logger, _ prometheus.Registerer) (database.Database, error) {
```
StephenButtolph, line 70: I think the `useSyncWrites` should be part of the config.

ceyonur: updated

StephenButtolph (approved):

PR #3592, opened 2024-12-13
refactor: unify the error handling methods in the crypto package that are different from the project style

ChengenH:


StephenButtolph:
Thanks for opening this PR. However, this instances of direct equality were intentional as they depend on the database interface which specifies exact (not wrapped) values.

PR #3598, opened 2024-12-16
GetTXFee API Refactor

samliok:
## Why this should be merged

Removes the `info.GetFeeTx` endpoint which is now deprecated. The api is moved to the `XChainClient` which only returns relevant fields now. 

## How this was tested

Added a test to CI.

Yes


meaghanfitzgerald:
looks good so far, would just double check that none of the `wallet` service functionality for the p-chain is broken by removing the endpoint. It shouldn't, as a new tx fee getter was already implemented pre etna, but it doesn't hurt to check that all those tests are passing

vms/avm/client.go
```diff
@@ -385,6 +389,17 @@ func (c *client) GetAllBalances(
 	return res.Balances, err
 }
 
+type GetTxFeeReply struct {
+	TxFee            uint64 `json:"txFee"`
+	CreateAssetTxFee uint64 `json:"createAssetTxFee"`
+}
```
StephenButtolph, line 395: I think we typically define these in the service file - not the client file.

vms/avm/service.md
```diff
@@ -783,6 +783,47 @@ curl -X POST --data '{
...
+```
+
+**Example Response**:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "result": {
+    "txFee": "1000000",
+    "createAssetTxFee": "10000000"
```
StephenButtolph, line 822: Currently the service returns `uint64` and not the `json.Uint64` helper - so I don't think these are quoted in the response... Right? I think either this doc should be updated or we should be returning quoted strings.

In the past we returned quoted numbers to avoid floating point inaccuracies in javascript (because all numbers in js are floating point and they lose precision after 2^53).

wallet/chain/p/context.go
```diff
@@ -45,38 +44,15 @@ func NewContextFromClients(
...
-	staticFeeConfig, err := infoClient.GetTxFee(ctx)
+	_, gasPrice, _, err := chainClient.GetFeeState(ctx)
 	if err != nil {
 		return nil, err
 	}
 
 	return &builder.Context{
-		NetworkID:   networkID,
-		AVAXAssetID: avaxAssetID,
-		StaticFeeConfig: fee.StaticConfig{
```
StephenButtolph, line 71: Can we remove `StaticFeeConfig` entirely? Perhaps we should add a TODO to the context to remove that field (as I don't think we need it anymore).

samliok: created this, will tackle in another pr
https://github.com/ava-labs/avalanchego/issues/3601

vms/avm/service.md
```diff
@@ -783,6 +783,47 @@ curl -X POST --data '{
 }
 ```
 
+### `info.getTxFee`
```
meaghanfitzgerald, line 786:
```suggestion
### `avm.getTxFee`
```

vms/avm/service.md
```diff
@@ -783,6 +783,47 @@ curl -X POST --data '{
...
+
+Get the fees of the network.
+
+**Signature**:
+
+```
+info.getTxFee() ->
```
meaghanfitzgerald, line 793:
```suggestion
avm.getTxFee() ->
```

vms/avm/service.md
meaghanfitzgerald, line 786: the api docs list the endpoints in alphabetical order. `avm.getTxFee` should come after [avm.getTx](https://docs.avax.network/api-reference/x-chain/api#avmgettx)

meaghanfitzgerald (approved):
lgtm, e2e pre etna is failling but i think that is expected?

samliok:
> lgtm, e2e pre etna is failling but i think that is expected?

yes its expected, but not sure why `e2e_existing_network` is failing

meaghanfitzgerald:
lets see if it fails after merging master

samliok:
closing this pr for #3610 since a more extensive pr is needed for the e2e tests to pass.

meaghanfitzgerald:
is there a PR that supersedes this?

PR #3604, opened 2024-12-17
Verify tx gas isn't too large in VerifyTx

StephenButtolph:
## Why this should be merged

Currently the mempool is allowed to contain transactions with too large of a gas usage.

## How this works

Enforce that the gas usage of a tx is <= the current gas capacity.

### DIFF
--- a/vms/platformvm/block/executor/manager.go
+++ b/vms/platformvm/block/executor/manager.go
@@ -135,18 +135,6 @@ func (m *manager) VerifyTx(tx *txs.Tx) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


yacovm (approved):

ARR4N (approved):
Logic equivalent to [current check in block building](https://github.com/ava-labs/avalanchego/blob/ae038f1a08b37c01df06a0f1f98161cc7fe24d05/vms/platformvm/block/builder/builder.go#L494), modulo the use of max capacity as discussed.

qdm12 (approved):

PR #3605, opened 2024-12-17
Improve block building and verification logging

StephenButtolph:
## Why this should be merged

Makes debugging issues easier.

## How this works

Adds logs and formats errors.

## How this was tested

N/A

No.

### DIFF
--- a/vms/components/gas/state.go
+++ b/vms/components/gas/state.go
@@ -5,6 +5,7 @@ package gas
 
 import (
 	"errors"
+	"fmt"
 	"math"
 
 	safemath "github.com/ava-labs/avalanchego/utils/math"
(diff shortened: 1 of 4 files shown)


vms/components/gas/state.go
```diff
@@ -44,7 +45,7 @@ func (s State) AdvanceTime(
 func (s State) ConsumeGas(gas Gas) (State, error) {
 	newCapacity, err := safemath.Sub(uint64(s.Capacity), uint64(gas))
 	if err != nil {
-		return State{}, ErrInsufficientCapacity
+		return State{}, fmt.Errorf("%w: capacity (%d) < gas (%d)", ErrInsufficientCapacity, s.Capacity, gas)
```
aaronbuchwald, line 48: ❤️

vms/platformvm/block/builder/builder.go
```diff
@@ -355,6 +355,9 @@ func buildBlock(
 		)
 	}
 	if err != nil {
+		builder.txExecutorBackend.Ctx.Log.Warn("failed to pack block transactions",
+			zap.Error(err),
+		)
```
aaronbuchwald, line 360: Note: engine will log at debug, but not warn level since block build failures can happen regularly if the VM cleared its mempool and decides not to produce an empty block.

aaronbuchwald (approved):

iansuvak (approved):

PR #3606, opened 2024-12-17
Pass the nodeID downstream of acp118 verifier

tsachiherman:
## Why this should be merged

Adding the previously verified source of a message allow us to cross-validate the application level signature.

## How this works

Add the nodeID to the verifier interface and pass it along.

## How this was tested

Unit test was updated to validate the correctness of the code.

No.


network/p2p/acp118/handler_test.go
```diff
@@ -26,25 +26,23 @@ func TestHandler(t *testing.T) {
 	tests := []struct {
 		name         string
 		cacher       cache.Cacher[ids.ID, []byte]
-		verifier     Verifier
```
joshua-kim, line 29: Don't think we need this diff? The `testVerifier` type now takes in a slice of errors when it's created

network/p2p/acp118/handler_test.go
```diff
@@ -143,18 +142,31 @@ func TestHandler(t *testing.T) {
...
-	*warp.UnsignedMessage,
-	[]byte,
+	_ context.Context,
+	_ *warp.UnsignedMessage,
+	justification []byte,
+	nodeID ids.NodeID,
 ) *common.AppError {
-	if len(t.Errs) == 0 {
+	require.Equal(t.t, t.clientNodeID, nodeID)
+	require.Equal(t.t, []byte("justification"), justification)
```
joshua-kim, line 165: I think we should pass in the expected justification instead of hard-coding here

network/p2p/acp118/handler_test.go
```diff
@@ -143,18 +142,31 @@ func TestHandler(t *testing.T) {
 
 // The zero value of testVerifier allows signing
 type testVerifier struct {
-	Errs []*common.AppError
+	t            *testing.T
+	errs         []*common.AppError
+	clientNodeID ids.NodeID
```
joshua-kim, line 147: I prefer exporting these values because it makes refactoring easier. These aren't visible outside outside of the package anyways because they're in an unexported type + it's in a test file, but if we ever choose to move this into a testing file (say `acp118test` it makes the diff smaller for the future).

network/p2p/acp118/handler_test.go
```diff
@@ -143,18 +142,31 @@ func TestHandler(t *testing.T) {
...
+}
+
+func newTestVerifier(t *testing.T, clientNodeID ids.NodeID, errs []*common.AppError) *testVerifier {
+	return &testVerifier{
+		t:            t,
+		errs:         errs,
+		clientNodeID: clientNodeID,
+	}
 }
```
joshua-kim, line 156: nit: This is a matter of preference but I don't like adding these constructor methods unless give us something (e.g add initialization logic or something), but since these are just wiring parameters I'd prefer to not have this

network/p2p/acp118/handler_test.go
```diff
@@ -131,6 +128,8 @@ func TestHandler(t *testing.T) {
 				if ok {
 					require.Equal(sig, response.Signature)
 				}
+
+				require.Equal(serverNodeID, nodeID)
```
joshua-kim, line 132: nit: We don't need to check for this IMO because this is guaranteed to match the responding peer from `p2p` and we already have tests on this in the upstream package - even if this line broke it wouldn't be possible to fix it from the perspective of acp118. If we did want to keep this... we should assert this after checking for a non-nil appErr before we check the response on L118 IMO.

tsachiherman: I'm not really picky on this one, so I'll just omit it for now.

network/p2p/acp118/handler_test.go
```diff
@@ -26,25 +26,23 @@ func TestHandler(t *testing.T) {
...
 				&common.AppError{Code: 123},
 			},
 		},
 		{
-			name:     "signature signed",
-			cacher:   &cache.Empty[ids.ID, []byte]{},
-			verifier: &testVerifier{},
+			name:         "signature signed",
+			cacher:       &cache.Empty[ids.ID, []byte]{},
+			verifierErrs: nil,
```
joshua-kim, line 45: This line can be deleted entirely iirc because the zero value of a slice is safe to use

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/DataDog/zstd v1.5.2
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/coreth v0.13.9-rc.2-encapsulate-signer
```
joshua-kim, line 13: Is there a corresponding PR to coreth for this change? Wondering what this diff is for

tsachiherman: quite minimal. I can't really create a final version until this one is merged since I'd need to tag it.. but these are the expected changes : https://github.com/ava-labs/coreth/pull/710

StephenButtolph:
We talked about this offline. While it is possible that someone would want to extend this in the future, there is no need currently to provide the nodeID to the verifier.

PR #3613, opened 2024-12-19
Add networked-signer tests

richardpringle:
- **Move local-signer to its own package**
- **Split bls tests into signing tests and other generic bls operations**
- **Add test for networked signer**

## Why this should be merged

I moved the tests to make it easier to test multiple signer implementations. 

## How this works

I split the tests up. There are tests in the `bls` package as well as the individual signers. There are also a couple of tests in the `blstest` package as well. 

## How this was tested

See above

nope


utils/crypto/bls/test/bls_benchmark_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package bls
+package bls_test
```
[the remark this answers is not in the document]
yacovm: This is a special package, the test package. It forces you to use only exported APIs.

utils/crypto/bls/signers/local/local.go
```diff
@@ -0,0 +1,74 @@
...
+// NewSecretKey generates a new secret key from the local source of
+// cryptographically secure randomness.
+func NewSigner() (*LocalSigner, error) {
+	var ikm [32]byte
+	_, err := rand.Read(ikm[:])
+	if err != nil {
+		return nil, err
+	}
+	sk := blst.KeyGen(ikm[:])
+	ikm = [32]byte{} // zero out the ikm
```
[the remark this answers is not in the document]
richardpringle: Gotta zero out the bytes used to initialize the secret key

utils/crypto/bls/test/bls_benchmark_test.go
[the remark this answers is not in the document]
richardpringle: It looks like the convention is to just put these files in the same directory as the parent package (`<parent_package_name>_test`). Want me to do that? I do need the `_test` otherwise it's a completely separate package.

richardpringle:
> (No action required) Would it make sense to separate the functional changes introduced by this PR from the considerable number of non-functional changes (e.g. renames) to simplify review of such an otherwise large PR?

Great question... maybe? 

I would argue that the overall cost of review doesn't change that much as most files just contain renames. I would have had to change the tests no matter what. I could have added the json-rpc signer in a second PR, but is that really beneficial?

I can do that if you (@marun) or any other reviewers would like, I don't mind.

utils/crypto/bls/signer.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package bls
+
+type Signer interface {
+	PublicKey() *PublicKey
+	Sign(msg []byte) *Signature
+	SignProofOfPossession(msg []byte) *Signature
```
joshua-kim, line 9: I feel like this interface looks cleaner if we just merge these two function signatures and just add the ciphertext as a parameter. Having a separate function signature for each ciphertext feels like it just bloats the interface... but not sure what others think.

utils/crypto/bls/blstest/bls_benchmark_test.go
```diff
@@ -35,7 +37,7 @@ var (
 )
 
 func BenchmarkSign(b *testing.B) {
```
joshua-kim, line 39: Something awkward about this test structure from the perspective of `localsigner` is that this test benchmarks the `localsigner` implementation (knows how to sign), not `bls` (knows operations across bls keys like aggregation/verification), so it feels like this belongs under something like `signer/localsigner/local_signer_test.go` since `bls` isn't even used here.

Maybe an alternative to this structure would be to have:
1. `bls/bls_test.go` - tests `bls` (operations across keys)
2. `signer/signertest/signertest.go` - has a common test suite that you can plug in any implementation of `Signer` into to verify that the interface implementation is correct
3. `signer/localsigner/localsigner_test.go` - uses the test suite in `signertest`
4. `signer/jsonrpc` - uses the test suite in `signertest`

utils/crypto/bls/blstest/bls_benchmark_test.go
```diff
@@ -51,7 +53,7 @@ func BenchmarkSign(b *testing.B) {
 }
 
 func BenchmarkVerify(b *testing.B) {
-	privateKey, err := NewSigner()
+	privateKey, err := localsigner.NewSigner()
```
joshua-kim, line 56: One thing that's awkward about this test structure from the perspective of `bls` is that `bls` tests are coupled to the implementation of `localsigner`

utils/crypto/bls/signers/json-rpc/client.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/gorilla/rpc/v2/json2"
+
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+)
+
+var _ bls.Signer = (*Client)(nil)
+
+type Client struct {
+	// http client
```
joshua-kim, line 21: nit: unnecessary comment

utils/crypto/bls/signers/json-rpc/client.go
joshua-kim, line 1: nit: `signer/jsonrpc` instead of `signer/json-rpc` for the path name, I don't think I've seen kebab-casing before in our repo

utils/crypto/bls/signers/json-rpc/client.go
```diff
@@ -0,0 +1,105 @@
...
+	return json2.DecodeClientResponse(resp.Body, result)
+}
+
+func (c *Client) PublicKey() *bls.PublicKey {
+	reply := new(PublicKeyReply)
+
+	err := c.call("Signer.PublicKey", []interface{}{PublicKeyArgs{}}, reply)
+	if err != nil {
+		panic(err)
+	}
```
joshua-kim, line 63: Should we add this error to the i/f definition? Panic'ing seems not graceful here since the caller can't decide how to handle the error without using `recover`. Or are we trying to avoid the error handling as part of this PR to make the diff smaller?

utils/crypto/bls/signers/json-rpc/client.go
```diff
@@ -0,0 +1,105 @@
...
+	defer resp.Body.Close()
+
+	return json2.DecodeClientResponse(resp.Body, result)
+}
+
+func (c *Client) PublicKey() *bls.PublicKey {
+	reply := new(PublicKeyReply)
+
+	err := c.call("Signer.PublicKey", []interface{}{PublicKeyArgs{}}, reply)
```
joshua-kim, line 60: nit: use `any` instead of `interface{}` (code before go1.18 uses the `interface{}` keyword since `any` wasn't introduced until then).

utils/crypto/bls/signers/json-rpc/client.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// deserialize the public key
+	sig := new(bls.Signature)
+	sig = sig.Deserialize(reply.Signature)
+
+	// can be nil if the public key is invalid
+	return sig
```
joshua-kim, line 104: Should we also consider adding an error to the i/f on `Sign*`? It seems like it'll be an annoying dx to rememer to check for a nil signature as opposed to getting an error.

utils/crypto/bls/signers/json-rpc/messages.go
joshua-kim, line 1: nit: `message.go`, or just define these types `server.go`

utils/crypto/bls/signers/json-rpc/server.go
```diff
@@ -0,0 +1,112 @@
...
+type signerService struct {
+	signer *localsigner.LocalSigner
+}
+
+type Server struct {
+	httpServer *http.Server
+	listener   net.Listener
+}
+
+func NewSignerService() *signerService {
```
joshua-kim, line 29: Nit: export `signerService` since it's being used from an exported function

utils/crypto/bls/signers/json-rpc/server.go
```diff
@@ -0,0 +1,112 @@
...
+	signer, err := localsigner.NewSigner()
+	if err != nil {
+		panic(err)
+	}
```
joshua-kim, line 33: Can we bubble up the error here instead of panic'ing? I think for the other panics it's reasonable if we don't want to make the diff bigger because it's an interface change + the panic's wont trigger because the network signer isn't actually used.

utils/crypto/bls/blstest/bls_test.go
```diff
@@ -0,0 +1,555 @@
...
+	client := jsonrpc.NewClient(url)
+
+	return jsonRPCSigner{server: server, Client: client}, nil
+}
+
+var signerFns = []func() (Signer, error){
+	localSignerFn,
+	serverSignerFn,
+}
+
```
joshua-kim, line 69: I think the tests in this file are a bit complex and also are now testing integration across the `jsonrpc` signer and `localsigner` which I feel like is unnecessary - if they correctly implement the `Signer` interface they're just returning plain data types and not interfaces so the operations in `bls` should still work across them. I think if we go with the earlier comment's suggestion we can probably get rid of a lot of these tests.

utils/crypto/bls/signers/json-rpc/server.go
```diff
@@ -0,0 +1,112 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package jsonrpc
```
joshua-kim, line 4: nit: this package name is inconsistent with `localsigner`... maybe we rename `localsigner` to `local` or `jsonrpc` to `jsonrpcsigner`. Alternatively we can pull this out of the PR and put it into a subsequent one.

utils/crypto/bls/blstest/blstest.go
```diff
@@ -0,0 +1,21 @@
...
+	if err != nil {
+		return false, err
+	}
+	aggPK, err := bls.AggregatePublicKeys(publicKeys)
+	if err != nil {
+		return false, err
+	}
+
+	return bls.Verify(aggPK, aggSig, message), nil
+}
```
richardpringle, line 21: Should I just expose this function from the `bls` library?

utils/crypto/bls/signer.go
richardpringle: I actually think that we should have a `ProofOfPossession()` method. We shouldn't really be doing proof-of-possession signatures of anything other than the pubkey. 

Regardless, that change would be beyond the scope of this PR.

utils/crypto/bls/bls_benchmark_test.go
```diff
@@ -10,55 +10,18 @@ import (
...
-			message := utils.RandomBytes(messageSize)
-
-			b.ResetTimer()
-
-			for n := 0; n < b.N; n++ {
-				_ = privateKey.Sign(message)
-			}
-		})
-	}
-}
```
richardpringle, line 51: Moved to individual implementations

utils/crypto/bls/signature.go
```diff
@@ -46,12 +46,25 @@ func SignatureFromBytes(sigBytes []byte) (*Signature, error) {
...
+	if err != nil {
+		return false, err
+	}
+	aggPK, err := AggregatePublicKeys(publicKeys)
+	if err != nil {
+		return false, err
+	}
+
+	return Verify(aggPK, aggSig, message), nil
+}
```
richardpringle, line 70: This function is only used in tests right now. I'm not sure if there's a better way to expose it

utils/crypto/bls/signers/json-rpc/client.go
richardpringle: I didn't want to add error handling to this PR, as it's already pretty big. I can add it if you (or anyone else) think I should.

utils/crypto/bls/signers/json-rpc/server.go
richardpringle: 🤔. If we're happy with how testing is split up, I don't mind pulling it out. The only thing is that the way I split up the testing might look a little silly with only a single implementation. 

I'm easy though, either way works. 

I also think `jsonrpcsigner` makes the most sense here and will make that change regardless.

utils/crypto/bls/signers/json-rpc/server.go
richardpringle: The constructor isn't actually part of the interface. Easy to add this, 👍

utils/crypto/bls/signer.go
richardpringle: Just to elaborate here, the reason that I prefer to have both APIs over a single function where one has to pass in the digest is because I think this should be opaque to the caller. It's an implementation detail that these two methods only differ in the digest used for signing.

utils/crypto/bls/signature.go
cam-schultz: The preferred approach (in this repo) is to use `*test` packages for test-only code: https://github.com/ava-labs/avalanchego/pull/3238

As for this functions utility outside of tests, I can't think of a use case off the top of my head where the caller would be both the signer and verifier.

utils/crypto/bls/signers/json-rpc/benchmark_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/blstest"
+)
+
+func BenchmarkSign(b *testing.B) {
+	server, privateKey, _ := NewSigner(require.New(b))
```
cam-schultz, line 17: What "server" is running here? Trying to understand what this benchmark measures distinct from the local signer benchmark.

utils/crypto/bls/signers/json-rpc/client.go
cam-schultz, line 1: What's the motivation to use `gorilla/rpc/v2/json2` instead of `avalanchego/utils/rpc` to build out this package? Most other rpc packages in Avalanchego use `avalanchego/utils/rpc`/`avalanchego/utils/json`, both of which use `gorilla` internally.

cam-schultz (commented):
The structure makes sense to me, and I appreciate the comprehensive test coverage you included. I left some comments that should hopefully be straightforward to answer/address.

utils/crypto/bls/ciphersuite.go
```diff
@@ -0,0 +1,12 @@
...
+
+package bls
+
+var (
+	// The ciphersuite is more commonly known as G2ProofOfPossession.
+	// There are two digests to ensure that message space for normal
+	// signatures and the proof of possession are distinct.
+	CiphersuiteSignature         = []byte("BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_")
+	CiphersuiteProofOfPossession = []byte("BLS_POP_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_")
+)
```
StephenButtolph, line 12: Is there any way to make these `const`s if we are exporting them?
```suggestion
const (
	// The ciphersuite is more commonly known as G2ProofOfPossession.
	// There are two digests to ensure that message space for normal
	// signatures and the proof of possession are distinct.
	CiphersuiteSignature         = "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_"
	CiphersuiteProofOfPossession = "BLS_POP_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_"
)
```
I think it was done as a var because I was concerned that a `[]byte("")` cast would perform memory allocations... But I suppose we could use unexported vars for that as well.

utils/crypto/bls/public.go
```diff
@@ -9,7 +9,10 @@ import (
 	blst "github.com/supranational/blst/bindings/go"
 )
 
-const PublicKeyLen = blst.BLST_P1_COMPRESS_BYTES
+const (
+	SecretKeyLen = blst.BLST_SCALAR_BYTES
```
StephenButtolph, line 13: It feels weird to me to export the `SecretKeyLen` from a package that doesn't handle secret keys anymore (because the length of the key is going to be representation specific).

Would it make more sense to export this from the `localsigner` package?

utils/crypto/bls/signer/localsigner/localsigner.go
```diff
@@ -0,0 +1,73 @@
...
+
+type secretKey = blst.SecretKey
+
+type LocalSigner struct {
+	sk *secretKey
+}
+
+// NewSecretKey generates a new secret key from the local source of
+// cryptographically secure randomness.
+func NewSigner() (*LocalSigner, error) {
```
StephenButtolph, line 29: `localsigner.NewSigner` kind of stutters - should this just be `New`?
```suggestion
func New() (*LocalSigner, error) {
```

utils/crypto/bls/public.go
```diff
@@ -9,7 +9,9 @@ import (
 	blst "github.com/supranational/blst/bindings/go"
 )
 
-const PublicKeyLen = blst.BLST_P1_COMPRESS_BYTES
+const (
+	PublicKeyLen = blst.BLST_P1_COMPRESS_BYTES
+)
```
StephenButtolph, line 14: Just to reduce the diff
```suggestion
const PublicKeyLen = blst.BLST_P1_COMPRESS_BYTES
```

utils/crypto/bls/signer/localsigner/localsigner.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// ToBytes returns the big-endian format of the secret key.
+func (s *LocalSigner) ToBytes() []byte {
+	return s.sk.Serialize()
+}
+
+// SecretKeyFromBytes parses the big-endian format of the secret key into a
+// secret key.
+func SecretKeyFromBytes(skBytes []byte) (*LocalSigner, error) {
```
StephenButtolph, line 48: Sorry to keep doing this to you
```suggestion
func FromBytes(skBytes []byte) (*LocalSigner, error) {
```

PR #3615, opened 2024-12-21
[tmpnet] Enable deployment to kube

marun:
## PR Chain: tmpnet+kube

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3945
- #3947
- #3973
- **>>>>>>** #3615 **<<<<<<** 
- #3794

## TODO

- [x] Merge #3973 
- [x] Clean up changes for maintainability
- [ ] Update tmpnet README
- [ ] post-merge: Make new CI job required

### DIFF
--- a/tests/fixture/tmpnet/kube_runtime.go
+++ b/tests/fixture/tmpnet/kube_runtime.go
@@ -192,11 +192,7 @@ func (p *KubeRuntime) Start(ctx context.Context) error {
 		zap.String("statefulSet", statefulSetName),
 	)
 	_, err = clientset.AppsV1().StatefulSets(namespace).Get(ctx, statefulSetName, metav1.GetOptions{})
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to retrieve StatefulSet %s/%s: %w", namespace, statefulSetName, err)
-		}
-	} else {
+	if err == nil {
 		// Stateful exists - make sure it is scaled up and running
 
 		log.Debug("attempting to retrieve scale for existing StatefulSet",
@@ -241,6 +237,8 @@ func (p *KubeRuntime) Start(ctx context.Context) error {
 		)
 
 		return nil
+	} else if !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to retrieve StatefulSet %s/%s: %w", namespace, statefulSetName, err)
 	}
 
 	// StatefulSet does not exist - create it


tests/fixture/tmpnet/start_kind_cluster.go
```diff
@@ -8,49 +8,137 @@ import (
...
+		)
+		return false, nil
+	}
+	if err != nil {
+		return false, fmt.Errorf("failed to check kubeconfig path %s: %w", configPath, err)
+	}
+
+	clientset, err := GetClientset(log, configPath, configContext)
+	if err != nil {
+		if strings.Contains(err.Error(), missingContextMsg) {
```
Elvis339, line 87: Seems like `GetClientset` could return a custom error like `ErrMissingContext` which you could check with `errors.Is`. 

I think this change could make it more discoverable without having to read implementation (for contributors both external and internal). Also if the error message changes this functionality can break without notice - having it defined as error at least communicates that this specific error condition is part of the expected API and should be explicitly handled.

Edit: I saw you left todo comment for yourself. 👍

Elvis339 (approved):
LGTM

tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -0,0 +1,768 @@
...
+	}
+
+	p.setNotRunning()
+
+	return p.ensureBootstrapIP(ctx)
+}
+
+// IsHealthy checks if the node is running and healthy.
+//
+// TODO(marun) Add WaitForHealthy as a runtime method to minimize API calls requires and
```
RodrigoVillar, line 549: (not neccessary) nit: `requires` => `required`

RodrigoVillar (approved):

tests/fixture/tmpnet/start_kind_cluster.go
```diff
@@ -8,49 +8,137 @@ import (
...
-func StartKindCluster(ctx context.Context, log logging.Logger, configPath string, configContext string) error {
-	err := CheckClusterRunning(log, configPath, configContext)
-	if err == nil {
-		log.Info("kubernetes cluster already running",
+// isKindClusterRunning determines if a kind cluster is running with context kind-kind
+func isKindClusterRunning(log logging.Logger, configPath string) (bool, error) {
+	configContext := KindKubeconfigContext
+
+	_, err := os.Stat(configPath)
+	if os.IsNotExist(err) {
```
StephenButtolph, line 75: We aren't consistent here, but these error checks are documented as being less preferred rather than `errors.Is`:

> This function predates `errors.Is`. It only supports errors returned by the os package. New code should use `errors.Is(err, fs.ErrNotExist)`.

StephenButtolph (approved):

PR #3617, opened 2024-12-21
[testing] Update golangci-lint to latest version

marun:
Updates golangci-lint to latest version.  

## TODO
 - [x] Resolve lint failures

## How this was tested

CI

N/A

### DIFF
--- a/chains/atomic/gsharedmemory/shared_memory_client.go
+++ b/chains/atomic/gsharedmemory/shared_memory_client.go
@@ -67,8 +67,6 @@ func (c *Client) Apply(requests map[ids.ID]*atomic.Requests, batches ...database
 		Batches:  make([]*sharedmemorypb.Batch, len(batches)),
 	}
 	for key, value := range requests {
-		key := key
-
 		chainReq := &sharedmemorypb.AtomicRequest{
 			RemoveRequests: value.RemoveRequests,
 			PutRequests:    make([]*sharedmemorypb.Element, len(value.PutRequests)),
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,12 @@
 module github.com/ava-labs/avalanchego
 
-// Changes to the minimum golang version must also be replicated in
-// CONTRIBUTING.md
-// README.md
-// go.mod (here)
+// - Changes to the minimum golang version must also be replicated in:
+//   - CONTRIBUTING.md
+//   - README.md
+//   - go.mod (here)
+//
+// - If updating between minor versions (e.g. 1.22.x -> 1.23.x):
+//   - Maybe update the version of golangci-lint (in scripts/lint.sh) in case the new golang version requires an updated linter?
 go 1.22.8
 
 require (
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ module github.com/ava-labs/avalanchego
 //   - go.mod (here)
 //
 // - If updating between minor versions (e.g. 1.22.x -> 1.23.x):
-//   - Maybe update the version of golangci-lint (in scripts/lint.sh) in case the new golang version requires an updated linter?
+//   - Consider updating the version of golangci-lint (in scripts/lint.sh).
 go 1.22.8
 
 require (
--- a/utils/hashing/hashing.go
+++ b/utils/hashing/hashing.go
@@ -58,7 +58,8 @@ func ComputeHash160Array(buf []byte) Hash160 {
 // ComputeHash160 computes a cryptographically strong 160 bit hash of the input
 // byte slice.
 func ComputeHash160(buf []byte) []byte {
-	// See comment on the import for why this is not a security risk
+	// See the comment on the ripemd160 import as to why the risk of use is
+	// considered acceptable.
 	ripe := ripemd160.New() //nolint:gosec
 	_, err := io.Writer(ripe).Write(buf)
 	if err != nil {
--- a/utils/hashing/hashing.go
+++ b/utils/hashing/hashing.go
@@ -9,7 +9,8 @@ import (
 	"fmt"
 	"io"
 
-	"golang.org/x/crypto/ripemd160"
+	// TODO(marun) Can this be replaced?
+	"golang.org/x/crypto/ripemd160" //nolint:gosec
 )
 
 const (
@@ -51,7 +52,7 @@ func ComputeHash160Array(buf []byte) Hash160 {
 // ComputeHash160 computes a cryptographically strong 160 bit hash of the input
 // byte slice.
 func ComputeHash160(buf []byte) []byte {
-	ripe := ripemd160.New()
+	ripe := ripemd160.New() //nolint:gosec
 	_, err := io.Writer(ripe).Write(buf)
 	if err != nil {
 		panic(err)
--- a/utils/hashing/hashing.go
+++ b/utils/hashing/hashing.go
@@ -9,7 +9,10 @@ import (
 	"fmt"
 	"io"
 
-	// TODO(marun) Can this be replaced?
+	// This algorithm does not represent a security risk since it is
+	// only used to convert a public key into an address during "Does
+	// this key own these funds" checks in the P-chain, X-chain, and
+	// C-chain.
 	"golang.org/x/crypto/ripemd160" //nolint:gosec
 )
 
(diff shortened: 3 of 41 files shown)


utils/hashing/hashing.go
```diff
@@ -9,7 +9,8 @@ import (
 	"fmt"
 	"io"
 
-	"golang.org/x/crypto/ripemd160"
+	// TODO(marun) Can this be replaced?
```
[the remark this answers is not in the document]
StephenButtolph: This can not be replaced. This is used to convert a public key into an address.

go.mod
```diff
@@ -1,9 +1,12 @@
...
-// CONTRIBUTING.md
-// README.md
-// go.mod (here)
+// - Changes to the minimum golang version must also be replicated in:
+//   - CONTRIBUTING.md
+//   - README.md
+//   - go.mod (here)
+//
+// - If updating between minor versions (e.g. 1.22.x -> 1.23.x):
+//   - Maybe update the version of golangci-lint (in scripts/lint.sh) in case the new golang version requires an updated linter?
```
StephenButtolph, line 9:
```suggestion
//   - Consider updating the version of golangci-lint (in scripts/lint.sh).
```

vms/avm/service_test.go
```diff
@@ -3029,7 +3029,7 @@ func TestServiceGetBlock(t *testing.T) {
 			expectedJSON, err := json.Marshal(expected)
 			require.NoError(err)
 
-			require.Equal(json.RawMessage(expectedJSON), reply.Block)
+			require.JSONEq(string(json.RawMessage(expectedJSON)), string(reply.Block))
```
StephenButtolph, line 3032:
```suggestion
			require.JSONEq(string(expectedJSON), string(reply.Block))
```

vms/avm/service_test.go
```diff
@@ -3235,7 +3235,7 @@ func TestServiceGetBlockByHeight(t *testing.T) {
...
```
StephenButtolph, line 3238:
```suggestion
			require.JSONEq(string(expectedJSON), string(reply.Block))
```

vms/platformvm/service_test.go
```diff
@@ -1276,7 +1276,7 @@ func TestServiceGetBlockByHeight(t *testing.T) {
 			expectedJSON, err := json.Marshal(expected)
 			require.NoError(err)
 
-			require.Equal(json.RawMessage(expectedJSON), reply.Block)
+			require.JSONEq(string(json.RawMessage(expectedJSON)), string(reply.Block))
```
StephenButtolph, line 1279:
```suggestion
			require.JSONEq(string(expectedJSON), string(reply.Block))
```

vms/platformvm/txs/convert_subnet_to_l1_tx_test.go
```diff
@@ -540,7 +540,7 @@ func TestConvertSubnetToL1TxSerialization(t *testing.T) {
 
 			txJSON, err := json.MarshalIndent(test.tx, "", "\t")
 			require.NoError(err)
-			require.Equal(
+			require.JSONEq(
 				// Normalize newlines for Windows
 				strings.ReplaceAll(string(test.expectedJSON), "\r\n", "\n"),
```
StephenButtolph, line 545: Do we still need to normalize these with the usage of `JSONEq`?

vms/platformvm/txs/disable_l1_validator_tx_test.go
```diff
@@ -315,7 +315,7 @@ func TestDisableL1ValidatorTxSerialization(t *testing.T) {
 
 	txJSON, err := json.MarshalIndent(unsignedTx, "", "\t")
 	require.NoError(err)
-	require.Equal(
+	require.JSONEq(
 		// Normalize newlines for Windows
 		strings.ReplaceAll(string(disableL1ValidatorTxJSON), "\r\n", "\n"),
```
StephenButtolph, line 320: Same comment w.r.t. if this is needed anymore

vms/platformvm/txs/increase_l1_validator_balance_tx_test.go
```diff
@@ -309,7 +309,7 @@ func TestIncreaseL1ValidatorBalanceTxSerialization(t *testing.T) {
 
 	txJSON, err := json.MarshalIndent(unsignedTx, "", "\t")
 	require.NoError(err)
-	require.Equal(
+	require.JSONEq(
 		// Normalize newlines for Windows
 		strings.ReplaceAll(string(increaseL1ValidatorBalanceTxJSON), "\r\n", "\n"),
```
StephenButtolph, line 314: Same comment w.r.t. if this is needed anymore

vms/platformvm/txs/register_l1_validator_tx_test.go
```diff
@@ -327,7 +327,7 @@ func TestRegisterL1ValidatorTxSerialization(t *testing.T) {
 
 	txJSON, err := json.MarshalIndent(unsignedTx, "", "\t")
 	require.NoError(err)
-	require.Equal(
+	require.JSONEq(
 		// Normalize newlines for Windows
 		strings.ReplaceAll(string(registerL1ValidatorTxJSON), "\r\n", "\n"),
```
StephenButtolph, line 332: Same comment w.r.t. if this is needed anymore

vms/platformvm/txs/set_l1_validator_weight_tx_test.go
```diff
@@ -299,7 +299,7 @@ func TestSetL1ValidatorWeightTxSerialization(t *testing.T) {
 
 	txJSON, err := json.MarshalIndent(unsignedTx, "", "\t")
 	require.NoError(err)
-	require.Equal(
+	require.JSONEq(
 		// Normalize newlines for Windows
 		strings.ReplaceAll(string(setL1ValidatorWeightTxJSON), "\r\n", "\n"),
```
StephenButtolph, line 304: Same comment w.r.t. if this is needed anymore

StephenButtolph (approved):

utils/hashing/hashing.go
```diff
@@ -9,7 +9,11 @@ import (
 	"fmt"
 	"io"
 
-	"golang.org/x/crypto/ripemd160"
+	// This algorithm does not represent a security risk since it is
```
yacovm, line 12: Why does this prove this isn't a security risk?

chains/atomic/gsharedmemory/shared_memory_client.go
```diff
@@ -67,8 +67,6 @@ func (c *Client) Apply(requests map[ids.ID]*atomic.Requests, batches ...database
 		Batches:  make([]*sharedmemorypb.Batch, len(batches)),
 	}
 	for key, value := range requests {
-		key := key
```
yacovm, line 70: RIP reference bug

yacovm (commented):
LGTM modulo comment on ripemd

utils/hashing/hashing.go
yacovm: I guess what we should say is something like "this is what is used in bitcoin and other cryptocurrencies", but the fact that we use this to convert a public key to an address doesn't mean this is inherently secure. 

For instance, suppose that ripemd160 would map half of its possible inputs to the same output. Then, it wouldn't be secure, because you could steal funds from every second address (on average).

yacovm: I think we should say, that we only use ripemd160 to derive addresses, in the same way Bitcoin does, and thus in this aspect we're as secure as Bitcoin is. 

I just commented because address derivation has security importance, so we should be precise in reasoning we put as comments :-) 

@StephenButtolph thoughts?

StephenButtolph: Yeah I think it would make sense to essentially say:
1. ripemd160 is also used by bitcoin in a similar way to generate an address from a public key.
2. ripemd160 is not generally recommended for use. However, it is only used to hash small payloads which is generally harder to attack than large payload size collisions.

StephenButtolph: We could also link to: [this paper](https://online.tugraz.at/tug_online/voe_main2.getvolltext?pCurrPk=17675) which is I beleive the most up-to-date research on collision resistance of ripemd160.

yacovm (approved):

PR #3618, opened 2024-12-21
[testing] Always use the go.mod version of ginkgo

marun:
## Why this should be merged

Previously, ginkgo was installed with an explicit version which made it easy for the ginkgo command and the test code to be using different versions. 

## How this works

- add new script `scripts/ginkgo.sh` that wraps `go run` execution with the go.mod version of ginkgo
- simplify all test scripts to execute tests with new script
- also: simplify test execution by running tests directly rather than via a test binary that needs to be built
  - this provides a nicer example for development usage since on-disk changes are always included

## How this was tested

CI

N/A

### DIFF
--- a/scripts/ginkgo.sh
+++ b/scripts/ginkgo.sh
@@ -2,5 +2,5 @@
 
 set -euo pipefail
 
-# Run the ginkgo version from go.mod
+# If an explicit version is not specified, go run uses the ginkgo version from go.mod
 go run github.com/onsi/ginkgo/v2/ginkgo "${@}"
--- a/scripts/ginkgo.sh
+++ b/scripts/ginkgo.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Run the ginkgo version from go.mod
+go run github.com/onsi/ginkgo/v2/ginkgo "${@}"
--- a/scripts/tests.e2e.bootstrap_monitor.sh
+++ b/scripts/tests.e2e.bootstrap_monitor.sh
@@ -63,5 +63,4 @@ ensure_command "kind-with-registry.sh" "https://raw.githubusercontent.com/kubern
 # call them without a qualifying path.
 PATH="${PWD}/bin:$PATH" bash -x "${PWD}/bin/kind-with-registry.sh"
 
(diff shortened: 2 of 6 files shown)


scripts/ginkgo.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Run the ginkgo version from go.mod
+go run github.com/onsi/ginkgo/v2/ginkgo "${@}"
```
qdm12, line 6: Wouldn't that run the latest v2 ginkgo version instead of the one defined in go.mod?

qdm12 (approved):

scripts/ginkgo.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Run the ginkgo version from go.mod
```
darioush, line 5:
```suggestion
# If an explicit version is not specified, go run uses the ginkgo version from go.mod
```

darioush (approved):

scripts/ginkgo.sh
qdm12: Oh good to know! 👍 
Thanks!

scripts/tests.e2e.bootstrap_monitor.sh
```diff
@@ -63,7 +63,5 @@ ensure_command "kind-with-registry.sh" "https://raw.githubusercontent.com/kubern
 # call them without a qualifying path.
 PATH="${PWD}/bin:$PATH" bash -x "${PWD}/bin/kind-with-registry.sh"
 
-# TODO(marun) Factor out ginkgo installation to avoid duplicating it across test scripts
-go install -v github.com/onsi/ginkgo/v2/ginkgo@v2.13.1
-
-KUBECONFIG="$HOME/.kube/config" PATH="${PWD}/bin:$PATH" ginkgo -v ./tests/fixture/bootstrapmonitor/e2e
+KUBECONFIG="$HOME/.kube/config" PATH="${PWD}/bin:$PATH" \
+          ./scripts/ginkgo.sh -v ./tests/fixture/bootstrapmonitor/e2e
```
StephenButtolph, line 67: was this indentation intended?

StephenButtolph (approved):

PR #3626, opened 2024-12-29
small corrections of different importance

DeVikingMark:


StephenButtolph:
Looking through this - I don't think this PR improves the readability of the code. Thanks for opening the PR though.

PR #3628, opened 2024-12-31
chore(all): mocks generation improved

qdm12:
Happy new year 🎉 🎆 🆕 

## Why this should be merged

💁 This PR is created because of similar PRs created on coreth (https://github.com/ava-labs/coreth/pull/721) and subnet-evm (https://github.com/ava-labs/subnet-evm/pull/1413)

ℹ️  Each commit can be separated in a different PR, and the few last commits can be dropped if necessary.

- [x] **Local** mockgen commands in the package where they are needed
- [x] Generate mocks from your IDE directly
- [x] Platform independent generation of mocks using `go` only
- [x] Simplify CI check for up-to-date mocks and to detect mocks without a source controlled generation command

## How this works

- New `mocks_generate_test.go` file per package where interface(s) need a corresponding mock(s) generated, containing only `//go:generate` commands for mock generation. Each command is relative to the current package directory.
- Use `//go:generate go run go.uber.org/mock/mockgen@v0.4` to avoid requiring to pre-install mockgen 
- no shell script, just `go` command needed with `go generate -run "mockgen" ./...` - remove now unneeded `scripts/mocks.gen.sh`, `scripts/mocks.mockgen.txt` and `scripts/mocks.mockgen.source.txt`
- Simplify CI mocks check, it only uses `grep`, `xargs` and `git`. This is now for CI only since the developer should mostly care about re-generating mocks and not detecting outdated mocks. The developer can still use the CI job inlined commands locally if needed.
- Documentation mocks section updated in CONTRIBUTING.md and referenced in README.md
- Remove unused generated mocks (if it's used in another repository, they should generate the mocks in their own repository ideally)
- Rename local test only mock files (1) to `mocks_test.go` to reduce Go production code namespace pollution
- mockgen upgraded from v0.4 to v0.5 in order to support generic code and drop the usage of the source mode mock generation (6 generate commands) in favor of the reflect mode
- ~mock files in the same directory are now combined into a single mock file "mocks.go" to reduce the number of mockgen calls to make full re-generation a bit faster~ removed to reduce 4000+ lines of diffs, will do later eventually

## How this was tested

- Locally with `go generate -run "go.uber.org/mock/mockgen" ./...`
- CI failing, for example: https://github.com/ava-labs/avalanchego/actions/runs/12560957420/job/35019102796

## Need to documented in RELEASES.md?

Not really


.github/workflows/ci.yml
```diff
@@ -155,9 +155,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-go-for-project
       - shell: bash
-        run: scripts/mock.gen.sh
-      - shell: bash
-        run: .github/workflows/check-clean-branch.sh
+        run: |
+          grep -lr -E '^// Code generated by MockGen\. DO NOT EDIT\.$' . | xargs -r rm
+          go generate -run "go.uber.org/mock/mockgen" ./...
+          git diff --exit-code
```
qdm12, line 161: Note: we can revert to use `.github/workflows/check-clean-branch.sh` instead of `git diff --exit-code` but really I think `git diff --exit-code` does it all and we shouldn't need to call an extra script

api/server/servermock/server.go
```diff
@@ -1,148 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
qdm12, line 1: Unused mock so removed

chains/atomic/atomicmock/shared_memory.go
```diff
@@ -42,10 +43,10 @@ func (m *SharedMemory) EXPECT() *SharedMemoryMockRecorder {
 }
 
 // Apply mocks base method.
-func (m *SharedMemory) Apply(arg0 map[ids.ID]*atomic.Requests, arg1 ...database.Batch) error {
+func (m *SharedMemory) Apply(requests map[ids.ID]*atomic.Requests, batches ...database.Batch) error {
```
qdm12, line 46: Note mockgen v0.5 does set the argument names correctly (nice to have!), but that also introduces most of the diffs in this PR.

CONTRIBUTING.md
```diff
@@ -45,11 +45,39 @@ To start developing on AvalancheGo, you'll need a few things installed.
...
+    // Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+    // See the file LICENSE for licensing terms.
+
+    package mypackage
+
+    //go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE} -destination=mocks_test.go . YourInterface
+    ```
+
+    Notes:
+    1. Ideally generate all mocks to `mocks_test.go` for the package you need to use the mocks for and do not export mocks to other packages. This reduces package dependencies, reduces production code pollution and forces to have locally defined narrow interfaces.
```
qdm12, line 74: This is really the best for the reasons enumerated, at the cost of having some generated mock code duplication here and there. Note I'm not advising to migrate the current mock file/exported setup, but it's best to use this approach for new code though.

snow/engine/snowman/block/mocks_generate_test.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package block
+
+//go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/build_block_with_context_chain_vm.go -mock_names=BuildBlockWithContextChainVM=BuildBlockWithContextChainVM . BuildBlockWithContextChainVM
+//go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/chain_vm.go -mock_names=ChainVM=ChainVM . ChainVM
+//go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/state_syncable_vm.go -mock_names=StateSyncableVM=StateSyncableVM . StateSyncableVM
+//go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/with_verify_context.go -mock_names=WithVerifyContext=WithVerifyContext . WithVerifyContext
```
qdm12, line 9: Note: these can be combined to generate a single file, reducing calls to mockgen from 4 to 1 (parsing only once) which makes mock generation faster overall. I however decided to drop my commit (still stashed locally) to reduce the git deltas by 4000+ lines for this PR.

chains/atomic/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
qdm12, line 2: Question on the copyright header:

1. Should we keep it for mocks_generate_test.go files?
2. Should we also add it to generated mocks or not really?

.github/workflows/ci.yml
StephenButtolph: I don't think `git diff --exit-code` reports untracked files.

qdm12: TIL thanks and indeed! We could do `git add -N` right before, but yeah it's getting ugly. I'll revert this to use the previous script

chains/atomic/mocks_generate_test.go
```diff
@@ -0,0 +1,6 @@
...
+
+package atomic
+
+//go:generate go run go.uber.org/mock/mockgen@v0.5 -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/shared_memory.go -mock_names=SharedMemory=SharedMemory . SharedMemory
```
StephenButtolph, line 6: Is it not possible to get:
```suggestion
//go:generate go run go.uber.org/mock/mockgen -package=${GOPACKAGE}mock -destination=${GOPACKAGE}mock/shared_memory.go -mock_names=SharedMemory=SharedMemory . SharedMemory
```
to work correctly somehow?

I know support for [this is explicitly added in go1.24](https://go.dev/doc/modules/managing-dependencies#tools) but they seem to suggest that the current norm for doing this is by including a `tools.go` file with the required dependencies.
> This removes the need for the previous workaround of adding tools as blank imports to a file conventionally named “tools.go”.

I feel like changing the mockgen version shouldn't require changing a ton of files explicitly.

chains/atomic/mocks_generate_test.go
StephenButtolph: We already don't have the header for the generated `pb.go` code - so I don't think we really need it on the generated mocks.

As far as the `mocks_generate_test.go` files, feels like we might as well have it.

CONTRIBUTING.md
StephenButtolph: I'm hoping to continue to reduce the usage of mocking in general tbh. It has been a very poor addition to our testing process (lots of maintenance for low value tests).

Not saying there isn't a time and a place... but it was way over done.

qdm12: Documented you should **REDUCE** usage of mocks in a35197abb697473f38c1c1f43ffdcc2df42284e9

chains/atomic/mocks_generate_test.go
qdm12: If we don't specify the version, it requires to run `go get go.uber.org/mock/mockgen@v0.5` to add the dependency `golang.org/x/mod v0.18.0 // indirect` in `go.mod`, which gets removed by a `go mod tidy`. So indeed the solution is to add a `tools.go` see commit 0ae62d225ec10318c11370e7669063478573d1f3

I think it's nicer, although feel free to suggest a better location for tools.go since the root of the project might not be the best place.

StephenButtolph:
Once linting passes this lgtm

chains/atomic/mocks_generate_test.go
qdm12: Created issue https://github.com/ava-labs/avalanchego/issues/3633 to track this and use the future `tool` directive.

StephenButtolph (approved):

PR #3629, opened 2025-01-02
chore(ci): define Github labels as code with a workflow

qdm12:
## Why this should be merged

Allow to track in source control the Github labels, and to easily migrate Github labels (i.e. renames).
This might be extended to the coreth and subnet-evm repositories in order to add/change labels in a source controlled/reviewable way.

## How this works

New Github workflow `.github/workflows/labels.yml` using `github/labels.yml` as a source for the labels. All the label data was extracted from [this API call](https://api.github.com/repos/ava-labs/avalanchego/labels?per_page=100) to keep it the exact same.

## How this was tested

[Dry run of labeler job on this PR](https://github.com/ava-labs/avalanchego/actions/runs/12581506804/job/35065425531?pr=3629#step:3:125).

No


.github/workflows/labels.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: crazy-max/ghaction-github-labeler@v5
```
StephenButtolph, line 22: We haven't typically used "less trusted" 3rd parties like this. What are the guarantees around which repository secrets are provided to jobs like this?

StephenButtolph: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#about-secrets

> GitHub Actions can only read a secret if you explicitly include the secret in a workflow.

Seems like it should be fine.

StephenButtolph (approved):

.github/workflows/labels.yml
qdm12: Yep 😉 It may not always had been the case, so good checking!
[crazy-max](https://github.com/crazy-max) does a lot of things in the docker/CI space, you may see his username around again!

PR #3634, opened 2025-01-03
fix LRU sized cache: consistent size at element removal

rrazvan1:
## Why this should be merged
LRU sized cache is using provided **_size_** fn to compute the size of elements when being inserted, or **removed**.

If the LRU size cache is using pointers to structs and the stored element is mutated after insertion in a way that provided _**size**_ fn is returning a different value at removal, the cache will have a corrupted state (more specifically _**currentSize**_ value).

## How this works
Elements hashmap is now using a **_sizedElement_** struct which includes the size, so it is no longer recomputed at the time of removal.

This will give a better performance, but will increase the memory load.

## How this was tested
Old cache unit tests passing + extra unit tests for checking possible wrong size into the cache's state because of element mutation after insertion.

No.


yacovm (approved):

cache/lru_sized_cache.go
```diff
@@ -12,20 +12,28 @@ import (
 
 var _ Cacher[struct{}, any] = (*sizedLRU[struct{}, any])(nil)
 
+// sizedElement is used to store the element + size, so we dont call size fn when removing
+// (and also is solving issues of mutating values after they were inserted,
+// which may alter the size)
```
StephenButtolph, line 17:
```suggestion
// sizedElement is used to store the element with its size, so we don't
// calculate the size multiple times.
//
// This ensures that any inconsistencies returned by the size function can not
// corrupt the cache.
```

StephenButtolph (approved):
super nit cleanup on the comment - rest lgtm

PR #3656, opened 2025-01-16
Reintroduce the deprecated `info.getTxFee` API

StephenButtolph:
## Why this should be merged

This API can't be removed without a deprecation window. It'll likely be removed in a few releases.

## How this works

Reintroduces `info.getTxFee` in a minimally invasive way and marks it as deprecated.

## How this was tested

Manually.

Done.

### DIFF
--- a/api/info/service.go
+++ b/api/info/service.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/json"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/utils/units"
 	"github.com/ava-labs/avalanchego/version"
 	"github.com/ava-labs/avalanchego/vms"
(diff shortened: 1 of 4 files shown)


samliok (approved):

api/info/service.go
```diff
@@ -53,6 +84,9 @@ type Parameters struct {
 	NetworkID uint32
 	VMManager vms.Manager
 	Upgrades  upgrade.Config
+
```
meaghanfitzgerald, line 87:
```suggestion
```

node/node.go
```diff
@@ -1409,6 +1409,9 @@ func (n *Node) initInfoAPI() error {
 			NetworkID: n.Config.NetworkID,
 			VMManager: n.VMManager,
 			Upgrades:  n.Config.UpgradeConfig,
+
```
meaghanfitzgerald, line 1412:
```suggestion
```

meaghanfitzgerald (approved):

PR #3658, opened 2025-01-17
optimize historical range

rrazvan1:
## Why this should be merged
Direct optimization:
- how combined changes are computed between 2 roots
- getting changes to a specific root
- changes iterator using `startKey` and/or `prefix`

Indirect optimization
- change proofs
- range proofs
- view changes iterator

Fixes:
- `getChangesToGetToRoot(..)` no-ops being removed from output

## How this works
- `changeSummary` struct has a new field for having the changed keys in a sorted slice
- `getChangesToGetToRoot(..)` -> by having the `sortedKeys`, we can search (binary search) for the `startKey`, and also easily stop iterating when we are after `endKey`.
- `getValueChanges(..)` -> we can easily get change values between `startRoot` and `endRoot`, with keys within `[startKey, endKey]` in the following way:
  1) init a minheap where we store a root traverse information: `changes`, `insertNumber` and `index`. Min => the root with the min key and min insertNumber (in this way, by popping out of the minheap, we traverse all the keys in ASC order by `[key, insertNumber]`)
  2) iterate through each root's changes, and find (binary search) the index of the first key within `[startKey, endKey]`, and push that initial state of each root into the heap (or not, if there are no keys inside that interval).
  3) pop elements out of minheap, and while the key is same, we merge the changes and store the final combined change.

**IMPORTANT improvement for getValueChanges(..)**: we can stop whenever there are `maxLength` key changes found.

## How this was tested
Using the existing unit tests.
Adding new unit tests, or modifying existing ones to properly cover the new code.

### DIFF
--- a/x/merkledb/db.go
+++ b/x/merkledb/db.go
@@ -346,9 +346,9 @@ func newDatabase(
 		rootChange: change[maybe.Maybe[*node]]{
 			after: trieDB.root,
 		},
-		keyIndexes: map[Key]int{},
+		sortedKeys: []Key{},
 		nodes:      map[Key]*change[*node]{},
-		keyChanges: []*keyChange{},
+		keyChanges: map[Key]*change[maybe.Maybe[[]byte]]{},
 	})
 
 	// mark that the db has not yet been cleanly closed
@@ -747,22 +747,22 @@ func (db *merkleDB) GetChangeProof(
 		return nil, database.ErrClosed
 	}
 
-	// [changes] contains a subset of the keys that were added or had their
+	// [valueChanges] contains a subset of the keys that were added or had their
 	// values modified between [startRootID] to [endRootID].
-	changes, err := db.history.getValueChanges(startRootID, endRootID, start, end, maxLength)
+	valueChanges, err := db.history.getValueChanges(startRootID, endRootID, start, end, maxLength)
 	if err != nil {
 		return nil, err
 	}
 
 	result := &ChangeProof{
-		KeyChanges: make([]KeyChange, len(changes)),
+		KeyChanges: make([]KeyChange, len(valueChanges)),
 	}
 
-	for i, keyChange := range changes {
+	for i, valueChange := range valueChanges {
 		result.KeyChanges[i] = KeyChange{
-			Key: keyChange.key.Bytes(),
+			Key: valueChange.key.Bytes(),
 			// create a copy so edits of the []byte don't affect the db
-			Value: maybe.Bind(keyChange.after, slices.Clone[[]byte]),
+			Value: maybe.Bind(valueChange.change.after, slices.Clone[[]byte]),
 		}
 	}
 
@@ -1360,9 +1360,9 @@ func (db *merkleDB) Clear() error {
 	db.history = newTrieHistory(db.history.maxHistoryLen)
 	db.history.record(&changeSummary{
 		rootID:     db.rootID,
-		keyIndexes: map[Key]int{},
+		sortedKeys: []Key{},
 		nodes:      map[Key]*change[*node]{},
-		keyChanges: make([]*keyChange, 0),
+		keyChanges: map[Key]*change[maybe.Maybe[[]byte]]{},
 	})
 	return nil
 }
--- a/x/merkledb/history.go
+++ b/x/merkledb/history.go
@@ -51,11 +51,6 @@ type changeSummaryAndInsertNumber struct {
 	insertNumber uint64
 }
 
-type keyChange struct {
-	*change[maybe.Maybe[[]byte]]
-	key Key
-}
-
 // Tracks all the node and value changes that resulted in the rootID.
 type changeSummary struct {
 	// The ID of the trie after these changes.
@@ -65,23 +60,15 @@ type changeSummary struct {
 	rootChange change[maybe.Maybe[*node]]
 	nodes      map[Key]*change[*node]
 
-	keyIndexes map[Key]int
-	keyChanges []*keyChange // sorted
-}
-
-func (cs *changeSummary) getChange(key Key) (*change[maybe.Maybe[[]byte]], bool) {
-	if index, ok := cs.keyIndexes[key]; ok {
-		return cs.keyChanges[index].change, true
-	}
-
-	return nil, false
+	keyChanges map[Key]*change[maybe.Maybe[[]byte]]
+	sortedKeys []Key
 }
 
 func newChangeSummary(estimatedSize int) *changeSummary {
 	return &changeSummary{
 		nodes:      make(map[Key]*change[*node], estimatedSize),
-		keyIndexes: make(map[Key]int, estimatedSize),
-		keyChanges: make([]*keyChange, 0, estimatedSize),
+		keyChanges: make(map[Key]*change[maybe.Maybe[[]byte]], estimatedSize),
+		sortedKeys: make([]Key, 0, estimatedSize),
 		rootChange: change[maybe.Maybe[*node]]{},
 	}
 }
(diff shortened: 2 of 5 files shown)


joshua-kim (commented):
We should have some benchmark results (either manual or preferably through a benchmark test) as part of this PR to verify the results

x/merkledb/history.go
```diff
@@ -178,62 +197,107 @@ func (th *trieHistory) getValueChanges(
...
+	return combinedKeyChanges, nil
 }
 
 // Returns the changes to go from the current trie state back to the requested [rootID]
 // for the keys in [start, end].
 // If [start] is Nothing, all keys are considered > [start].
 // If [end] is Nothing, all keys are considered < [end].
 func (th *trieHistory) getChangesToGetToRoot(rootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[[]byte]) (*changeSummary, error) {
-	// [lastRootChange] is the last change in the history resulting in [rootID].
+	// [lastRootChange] is the last change in the historyChanges resulting in [rootID].
```
joshua-kim, line 300: Did we mean to update this comment?

x/merkledb/history_test.go
```diff
@@ -646,12 +792,183 @@ func TestHistoryRecord(t *testing.T) {
...
+					before: maybe.Some([]byte("value1b")),
+					after:  maybe.Some([]byte("value1a")),
+				},
+				key: ToKey([]byte("key1")),
+			},
+		},
+	}
+
+	maxHistoryLen := len(keyChangesSets)
+	history := newTrieHistory(maxHistoryLen)
```
joshua-kim, line 844: Same comment w.r.t unexported code

x/merkledb/view_iterator.go
```diff
@@ -29,21 +29,31 @@ func (v *view) NewIteratorWithStartAndPrefix(start, prefix []byte) database.Iter
...
-		if len(start) > 0 && startKey.Greater(key) || !key.HasPrefix(prefixKey) {
+	startKeyIndex := 0
+	if len(start) > 0 {
+		// Binary search for [startKey] index.
+		startKeyIndex, _ = slices.BinarySearchFunc(v.changes.sortedKeyChanges, startKey, func(k *keyChange, key Key) int {
+			return k.key.Compare(key)
+		})
+	}
+
+	for _, kChange := range v.changes.sortedKeyChanges[startKeyIndex:] {
```
joshua-kim, line 40: Won't this panic if startKeyIndex is out of bounds?

x/merkledb/view_test.go
```diff
@@ -103,3 +105,47 @@ func Benchmark_HashChangedNodes(b *testing.B) {
...
+			Key:   key,
+			Value: value,
+		})
+	}
+
+	ctx := context.Background()
+	viewIntf, err := db.NewView(ctx, ViewChanges{BatchOps: ops})
+	require.NoError(b, err)
+
+	view := viewIntf.(*view)
```
joshua-kim, line 138: This cast isn't necessary

x/merkledb/db.go
```diff
@@ -1365,9 +1359,10 @@ func (db *merkleDB) Clear() error {
 	// Clear history
 	db.history = newTrieHistory(db.history.maxHistoryLen)
 	db.history.record(&changeSummary{
-		rootID: db.rootID,
-		values: map[Key]*change[maybe.Maybe[[]byte]]{},
-		nodes:  map[Key]*change[*node]{},
+		rootID:           db.rootID,
+		values:           map[Key]*keyChange{},
+		nodes:            map[Key]*change[*node]{},
+		sortedKeyChanges: make([]*keyChange, 0),
```
joshua-kim, line 1365: Similar comment, but I wonder if it makes sense for us to have two copies of `keyChange` since we have to worry about them always being in sync. Maybe values could be updated to be a map of keys to indicies and do a lookup in `sortedKeyChanges`?

x/merkledb/history.go
```diff
@@ -156,20 +165,30 @@ func (th *trieHistory) getValueChanges(
 		}
 	}
 
-	var (
-		// Keep track of changed keys so the largest can be removed
-		// in order to stay within the [maxLength] limit if necessary.
-		changedKeys = set.Set[Key]{}
+	// historyChangesIndex is used for tracking keyChanges index from each historical root.
+	type historyChangesIndex struct {
```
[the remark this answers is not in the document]
rrazvan1: This package is full of structs. Since this is a small struct, and only used in this function, I prefer keeping it isolated.

x/merkledb/history.go
```diff
@@ -52,22 +54,29 @@ type changeSummaryAndInsertNumber struct {
 	insertNumber uint64
```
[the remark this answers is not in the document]
rrazvan1: For tracing purpuses, even though I mentioned privately to you: I will be using a map of the `rootIDs` -> `insert number` because indexes from history double queue are changing, and by having an insert number of a root id, we can compute its index from `insertNumber`.

x/merkledb/history.go
```diff
@@ -178,62 +197,107 @@ func (th *trieHistory) getValueChanges(
...
+			startKeyIndex, _ = slices.BinarySearchFunc(historyChanges.sortedKeyChanges, startKey, func(k *keyChange, m maybe.Maybe[Key]) int {
+				return k.key.Compare(m.Value())
+			})
 
-		// Add the changes from this commit to [combinedChanges].
-		for key, valueChange := range changes.values {
-			// The key is outside the range [start, end].
-			if (startKey.HasValue() && key.Less(startKey.Value())) ||
-				(end.HasValue() && key.Greater(endKey.Value())) {
+			if startKeyIndex >= len(historyChanges.sortedKeyChanges) {
```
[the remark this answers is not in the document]
rrazvan1: `slices.BinarySearchFunc` is returning the index, if the value **is present**, OR the index where the value **would appear** =>  `slices.BinarySearchFunc` can never return a negative value.

So that line, is skipping historical roots where there are no values from `[startKey, endKey]` in the `sortedKeyChanges`

x/merkledb/view_iterator.go
rrazvan1: `startKeyIndex` is between `0` and `len(v.changes.sortedKeyChanges)]`
In case `startKeyIndex` is `len(v.changes.sortedKeyChanges)`, `v.changes.sortedKeyChanges[startKeyIndex:]` is going to be an empty slice.

rrazvan1:
```
values := []int{0, 10, 20, 30}
idx, _ := slices.BinarySearch(values, 50)
fmt.Println(idx, values[idx:])
```
output:
```
4 []
```

x/merkledb/history.go
```diff
@@ -52,6 +51,11 @@ type changeSummaryAndInsertNumber struct {
 	insertNumber uint64
 }
 
+type keyChange struct {
+	*change[maybe.Maybe[[]byte]]
+	key Key
```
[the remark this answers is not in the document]
rrazvan1: I dont like the idea of using the bimap, bcs under the hood is actually adding a totally new map (reversed) (and only getting rid of a field inside a struct), but your comment gave me an idea about getting rid of this struct (which I hate for some reason):

```
keyIndexes map[Key]int
keyChanges []*keyChange // sorted ASC by key
```

replace with:

```
keyChanges map[Key]*change[maybe.Maybe[[]byte]]
sortedKeys []*Key
```

In this way we have access to:
- change by key
- change by index
- iterating through changes (+keys) being sorted

But we no longer need `keyChange` struct. 🤔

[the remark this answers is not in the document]
joshua-kim: Ah, yeah i think this works too 😄

PR #3660, opened 2025-01-19
[docker] Enable image builds from git worktrees

marun:
## Why this should be merged

This change enables docker image builds from [git worktrees](https://git-scm.com/docs/git-worktree). This required discoverying the git commit used to configure the avalanchego build script in advance of image build and providing it as an argument to image build.

Previously, it was impossible to build docker images from a git worktree:

 - the avalanchego build script needed to discover the git commit to configure the binary with for traceability
 - the build script assumed a .git directory to discover the git commit
 - the dockerfile copied the working tree into the image prior to running the avalanchego build script
 - if the working tree was for a git worktree, the resulting copy wouldn't have a .git directory to discovery the git commit from
   - instead of a regular .git directory, a non-primary worktree has a .git file referencing a path in the primary worktree's .git directory.

## How this was tested

CI

N/A

### DIFF
--- a/scripts/lib_build_antithesis_images.sh
+++ b/scripts/lib_build_antithesis_images.sh
@@ -66,7 +66,7 @@ function build_antithesis_images {
  --build-arg GO_VERSION=${go_version}\
  --build-arg BUILDER_IMAGE_TAG=${image_tag}\
  --build-arg BUILDER_WORKDIR=${builder_workdir}"
-  if [[ -n "${avalanchego_commit:-}" ]]; then
+  if [[ -n "${avalanchego_commit}" ]]; then
     docker_cmd="${docker_cmd} --build-arg AVALANCHEGO_COMMIT=${avalanchego_commit}"
   fi
 


scripts/git_commit.sh
michaelkaplan13, line 1: [No action] It was a little surprising to me that the script was not executable, but I guess this is not a new pattern introduced here (as opposed to echo'ing the variables and letting the caller export/source them)

scripts/lib_build_antithesis_images.sh
```diff
@@ -65,6 +66,9 @@ function build_antithesis_images {
  --build-arg GO_VERSION=${go_version}\
  --build-arg BUILDER_IMAGE_TAG=${image_tag}\
  --build-arg BUILDER_WORKDIR=${builder_workdir}"
+  if [[ -n "${avalanchego_commit:-}" ]]; then
```
michaelkaplan13, line 69: Relative bash novice here, and feel free to ignore either way, but is the :- necessary given that we already have `local avalanchego_commit=${10:-}` above?

michaelkaplan13 (approved):
TIL what git worktrees are. Confirmed the image builds work from a linked worktree on this branch (and saw the error when trying to build it from a worktree on `master`).

StephenButtolph (approved):

PR #3665, opened 2025-01-20
[tmpnet] Update URI and StakingAddress usage in support of kube

marun:
## Why this should be merged

Previously, `Node.URI` and `Node.StakingAddress` fields contained values for a local process and were always locally accessible. Nodes running in kubernetes may be running remotely, though, and only be accessible if e.g. a local port is forwarded to the port of a node's pod. To ensure compatibility with forwarded URIs and StakingAddresses, callers that aren't guaranteed to be running locally to a node are updated to access these values via functions that can ensure a remote endpoint is forwarded if the target node isn't running locally.

## How this was tested

CI

N/A

## TODO

- [x] Merge https://github.com/ava-labs/avalanchego/pull/3650

### DIFF
--- a/tests/fixture/e2e/helpers.go
+++ b/tests/fixture/e2e/helpers.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"net/netip"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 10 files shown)


tests/fixture/e2e/helpers.go
```diff
@@ -353,3 +354,26 @@ func GetRepoRootPath(suffix string) (string, error) {
...
 }
+
+// GetLocalURI retrieves the locally-accessible URI of the provided node. When a node
+// is running as a local process, this will be the URI exposed by the node. For a
+// node running remotely in kube, the URI will be a local address whose port is
+// forwarded to the node's URI through the kube API server.
+func GetLocalURI(tc tests.TestContext, node *tmpnet.Node) string {
+	uri, cancel, err := node.GetLocalURI(tc.DefaultContext())
+	require.NoError(tc, err)
+	tc.DeferCleanup(cancel)
```
yacovm, line 365: why not just invoke defer cancel() ?

yacovm (approved):

qdm12 (approved):

aaronbuchwald (approved):

PR #3668, opened 2025-01-22
Rename early_term_no_traversal(_test).go to early_term(_test).go

yacovm:
## Why this should be merged

In https://github.com/ava-labs/avalanchego/pull/3337 we introduce transitive voting for early termination of snow polls.
The files that were changed have no_traversal in their names, which implies there should be an implementation with a traversal. 

After https://github.com/ava-labs/avalanchego/pull/3337 was merged, we only have a single implementation, hence it doesn't make sense to keep the original file names.

However, changing the file names makes the diff of the squashed commit look as if all the lines in the files were introduced in that PR, which is misleading.

In order to preserve git blame and a per-line history, I made a new PR that just uses git mv on the files.

## How this works

Changing the file names

## How this was tested

Tested at compile time

No


StephenButtolph:
This has since been rolled into the original PR

PR #3672, opened 2025-01-24
Add F Upgrade Scaffolding. Post-Etna Cleanup

michaelkaplan13:
## Why this should be merged
- Adds scaffolding for future changes to be made as part of the F upgrade behind `IsFUpgradeActivated` checks.
- Removes `GetEtnaHeight` from the `validators.State` interface which is not needed now that Etna is activated on all networks.

## How this works
Self-explanatory. (The actual F upgrade name is to be decided later)

## How this was tested
CI

No, no functional change.


aaronbuchwald:
https://www.imdb.com/title/tt0083946/

upgrade/upgradetest/config.go
```diff
@@ -31,6 +31,8 @@ func GetConfigWithUpgradeTime(fork Fork, upgradeTime time.Time) upgrade.Config {
 // to the provided upgradeTime.
 func SetTimesTo(c *upgrade.Config, fork Fork, upgradeTime time.Time) {
 	switch fork {
+	case FUpgrade:
+		c.FUpgradeTime = upgradeTime
```
StephenButtolph, line 35:
```suggestion
		c.FUpgradeTime = upgradeTime
		fallthrough
```

vms/platformvm/state/state.go
```diff
@@ -1326,10 +1323,6 @@ func (s *state) GetStartTime(nodeID ids.NodeID) (time.Time, error) {
 	return staker.StartTime, nil
 }
 
-func (s *state) GetEtnaHeight() (uint64, error) {
-	return database.GetUInt64(s.singletonDB, EtnaHeightKey)
```
StephenButtolph, line 1330: Can we delete `EtnaHeightKey`?

vms/rpcchainvm/vm_client.go
```diff
@@ -198,6 +198,7 @@ func (vm *VMClient) Initialize(
 		CortinaXChainStopVertexId:     chainCtx.NetworkUpgrades.CortinaXChainStopVertexID[:],
 		DurangoTime:                   grpcutils.TimestampFromTime(chainCtx.NetworkUpgrades.DurangoTime),
 		EtnaTime:                      grpcutils.TimestampFromTime(chainCtx.NetworkUpgrades.EtnaTime),
+		FUpgradeTime:                  grpcutils.TimestampFromTime(chainCtx.NetworkUpgrades.FUpgradeTime),
```
StephenButtolph, line 201: I think we need to add this to the `vm_server.go` logic as well right?

vms/platformvm/state/state.go
michaelkaplan13: I wasn't sure looking at it initially since IIUC, existing nodes will still have that value in their state and new nodes will not. I guess there isn't any worry of mistakingly re-introducing that state key in the future though, even if we remove it from source.

StephenButtolph (approved):

darioush (approved):

PR #3678, opened 2025-01-27
minimize BitSetSignature.Verify chain scope

tsachiherman:
## Why this should be merged

This PR minimize the interface scope of the `pChainState` interface argument in `BitSetSignature.Verify`.

## How this works

Existing interface of `BitSetSignature.Verify` is 
```golang
func (s *BitSetSignature) Verify(
	ctx context.Context,
	msg *UnsignedMessage,
	networkID uint32,
	pChainState validators.State,
	pChainHeight uint64,
	quorumNum uint64,
	quorumDen uint64,
)
```
and the `validators.State` interface is 
```golang
type State interface {
	// GetMinimumHeight returns the minimum height of the block still in the
	// proposal window.
	GetMinimumHeight(context.Context) (uint64, error)
	// GetCurrentHeight returns the current height of the P-chain.
	GetCurrentHeight(context.Context) (uint64, error)

	// GetSubnetID returns the subnetID of the provided chain.
	GetSubnetID(ctx context.Context, chainID ids.ID) (ids.ID, error)

	// GetValidatorSet returns the validators of the provided subnet at the
	// requested P-chain height.
	// The returned map should not be modified.
	GetValidatorSet(
		ctx context.Context,
		height uint64,
		subnetID ids.ID,
	) (map[ids.NodeID]*GetValidatorOutput, error)

	// GetCurrentValidatorSet returns the current validators of the provided subnet
	// and the current P-Chain height.
	// Map is keyed by ValidationID.
	GetCurrentValidatorSet(
		ctx context.Context,
		subnetID ids.ID,
	) (map[ids.ID]*GetCurrentValidatorOutput, uint64, error)
}
```

However, only the `GetValidatorSet` and `GetSubnetID` methods are being used by `BitSetSignature.Verify`.

In this PR, I've created a partial interface that includes just these two functions:
```golang
type VerifiableValidatorState interface {
	// GetValidatorSet returns the validators of the provided subnet at the
	// requested P-chain height.
	GetValidatorSet(ctx context.Context, height uint64, subnetID ids.ID) (map[ids.NodeID]*validators.GetValidatorOutput, error)

	// GetSubnetID returns the subnetID of the provided chain.
	GetSubnetID(ctx context.Context, chainID ids.ID) (ids.ID, error)
}
```

## How this was tested

This PR retain full backward interface compatibility.

No.


vms/platformvm/warp/signature.go
```diff
@@ -23,6 +24,17 @@ var (
 	ErrParseSignature     = errors.New("failed to parse signature")
 )
 
+// VerifiableValidatorState allows the lookup of validator sets on specified subnets at the
+// requested P-chain height as well as the retrieval of subnet ids.
+type VerifiableValidatorState interface {
```
joshua-kim, line 29: We don't need to disambiguate across `validators.State` because this is its own package - can we just call this `ValidatorState` or `Validators` or something like that?

aaronbuchwald: I was thinking we should add the necessary functions to the existing `ValidatorState` interface here: https://github.com/ava-labs/avalanchego/blob/ab4619873028fd975dfa6e89e53e555dd53e0cdb/vms/platformvm/warp/validator.go#L32

The downside is ofc that the functions we're adding to it are not strictly needed.

Perhaps a better alternative is to modify the function signature of `*BitSetSIgnature.Verify(...)` to take in an already calculated canonical validator set and provide a helper function that converts from `validators.State` to the canonical validator set input that's required.

PR #3679, opened 2025-01-27
remove dependency of validator.State from BitSetSignature.Verify

tsachiherman:
## Why this should be merged

This PR minimize the interface scope of the `pChainState` interface argument in `BitSetSignature.Verify`.

## How this works
Existing interface of `BitSetSignature.Verify` is 
```golang
func (s *BitSetSignature) Verify(
	ctx context.Context,
	msg *UnsignedMessage,
	networkID uint32,
	pChainState validators.State,
	pChainHeight uint64,
	quorumNum uint64,
	quorumDen uint64,
)
```

This PR replaces the interface with
```golang
func (s *BitSetSignature) Verify(
	msg *UnsignedMessage,
	networkID uint32,
	validators CanonicalValidatorSet,
	quorumNum uint64,
	quorumDen uint64,
) error
```

with the `CanonicalValidatorSet` defined as
```golang
type CanonicalValidatorSet struct {
	Validators  []*Validator
	TotalWeight uint64
}
```

The CanonicalValidatorSet construction function was added as well :
```golang
func GetCanonicalValidatorSetFromState(
        ctx context.Context,
	pChainState validators.State,
	pChainHeight uint64,
	sourceChainID ids.ID,
) (CanonicalValidatorSet, error) 
```

## How this was tested

Existing testing were modified.

No.

## Corresponding changes

coreth : https://github.com/ava-labs/coreth/pull/769


StephenButtolph:
The comment:
```
// Because [signers] is a subset of [vdrs], this can never error.
sigWeight, _ := SumWeight(signers)
```
is no longer valid right?

aaronbuchwald (approved):

vms/platformvm/warp/validator.go
```diff
@@ -32,6 +32,11 @@ type ValidatorState interface {
 	GetValidatorSet(ctx context.Context, height uint64, subnetID ids.ID) (map[ids.NodeID]*validators.GetValidatorOutput, error)
 }
 
+type CanonicalValidatorSet struct {
+	Validators  []*Validator
+	TotalWeight uint64
+}
```
StephenButtolph, line 38: I dislike how we are opening up the ability for the user of this API to provide incorrect weights...

`FlattenValidatorSet` only returned the total weight to avoid needing to use safemath logic for creating the weight of the validators... It seems like a cleaner API if we removed the total weight from `FlattenValidatorSet` and calculate it later inside of `Verify`.

Alternatively, we could unexport `Validators` and `TotalWeight` so that we enforce the set is correctly calculated by this package (but that seems less ideal)

vms/platformvm/warp/validator.go
```diff
@@ -154,3 +159,17 @@ func AggregatePublicKeys(vdrs []*Validator) (*bls.PublicKey, error) {
 	}
 	return bls.AggregatePublicKeys(pks)
 }
+
+// GetCanonicalValidatorSetFromState returns the canonical validator set given a validators.State, pChain height and a sourceChainID.
+func GetCanonicalValidatorSetFromState(ctx context.Context,
```
StephenButtolph, line 164: `GetCanonicalValidatorSet` vs `GetCanonicalValidatorSetFromState`. Based on the names, I would expect `GetCanonicalValidatorSet` not to take in the `State` and for `GetCanonicalValidatorSetFromState` to take in the `State`.

They both take in states though... One of them taking in a chainID + state which can convert the chainID into a subnetID. The other taking in a subnetID + state.

Could we improve the names of these functions?

cam-schultz (approved):

vms/platformvm/warp/validator.go
tsachiherman: I think that I'd like to let the caller make the decision here, and like the idea of calculating the total weight inside the verify function.

vms/platformvm/warp/validator.go
tsachiherman: ok, done

StephenButtolph (approved):

PR #3680, opened 2025-01-28
chore(proposervm): timestamp metrics for block acceptance

ARR4N:
## Why this should be merged

Allows for alerting should `proposervm` block times drift too far from their inner blocks.

## How this works

Adds a new `proposervm` vector metric for last-accepted timestamp.

Lightly held opinion: although this introduces a redundant metric (`proposervm_last_accepted_timestamp` with `block_type=proposervm` is identical to `snowman_last_accepted_timestamp`), I don't think we should remove the latter because it has a natural partner in `snowman_last_accepted_height`. Also, while usage of `proposervm` is currently required, if this were to change then there would be a need for the `snowman` metric but it would likely be forgotten. In summary, I think a little redundancy is ok in return for snowman-metric completeness.

## How this was tested

Unit test.

Nope.


vms/proposervm/vm.go
```diff
@@ -231,10 +235,18 @@ func (vm *VM) Initialize(
 		// of comparing floating point of the same numerical value.
 		Buckets: []float64{0.5, 1.5, 2.5},
 	})
+	vm.lastAcceptedTimestampGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "last_accepted_timestamp",
+			Help: "timestamp of the last block to be accepted",
```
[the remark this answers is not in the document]
ARR4N: Yeah I meant "to have been" not "to be" (in the future). It's interesting because I have some idiosyncratic phrases and I wonder if this is one or if English is just vague in this regard.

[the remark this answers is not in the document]
ARR4N: > I have some idiosyncratic phrases and I wonder if this is one or if English is just vague in this regard

My wife: "I think you're both wrong" 🤣 

So I'm saying that it's a limitation of the language!

vms/proposervm/vm_test.go
```diff
@@ -2530,3 +2531,43 @@ func TestLocalParse(t *testing.T) {
...
 	}
 }
+
+func TestTimestampMetrics(t *testing.T) {
+	ctx := context.Background()
+
+	coreVM, _, proVM, _ := initTestProposerVM(t, time.Unix(0, 0), mockable.MaxTime, 0)
+	defer func() {
+		require.NoError(t, proVM.Shutdown(ctx))
+	}()
```
qdm12, line 2541:
```suggestion
	t.Cleanup(func() {
		require.NoError(t, proVM.Shutdown(ctx))
	})
```

vms/proposervm/vm_test.go
```diff
@@ -2530,3 +2531,43 @@ func TestLocalParse(t *testing.T) {
...
+
+	gaugeVec := proVM.lastAcceptedTimestampGaugeVec
+	tests := []struct {
+		blockType string
+		want      time.Time
+	}{
+		{innerBlockTypeMetricLabel, innerTime},
+		{outerBlockTypeMetricLabel, outerTime},
+	}
+	for _, tt := range tests {
```
qdm12, line 2566: nit why `tt` and not `test` 👀 I suppose not to conflict with `t`? 🤔  But `test` is nice and short too 😅  Just a general comment for future code, please no need to change it here.

vms/proposervm/post_fork_block.go
```diff
@@ -37,9 +38,20 @@ func (b *postForkBlock) Accept(ctx context.Context) error {
...
 	return nil
 }
 
+const (
+	innerBlockTypeMetricLabel = "inner"
+	outerBlockTypeMetricLabel = "proposervm"
+)
+
+func (b *postForkBlock) updateLastAcceptedTimestampMetric(blockTypeLabel string, t time.Time) {
+	b.vm.lastAcceptedTimestampGaugeVec.WithLabelValues(blockTypeLabel).Set(float64(t.Unix()))
```
qdm12, line 52: nit that line is pretty long (currently on half a laptop monitor, that's probably why 😄), maybe split with

```suggestion
	g := b.vm.lastAcceptedTimestampGaugeVec
	g.WithLabelValues(blockTypeLabel)
	g.Set(float64(t.Unix()))
```

qdm12 (approved):
👍  Just tiny minor comments

vms/proposervm/vm_test.go
ARR4N: I'll defer to @StephenButtolph on this due to consistency within the file.

Is there any real difference between the two? If anything I think it would be better to call `t.Cleanup()` inside `initTestProposerVM()` and remove it from everywhere else, but not in this PR.

vms/proposervm/vm_test.go
ARR4N: Yeah `tt` to not shadow `t`. It was common enough in the Google code base (along with `tc` for test case) that it was generally known exactly what it was (like `ctx`) so I picked it up as a habit, but appreciate that that may not be the case outside. I tend to keep using it over `test` because of the number of times it gets used.

vms/proposervm/vm_test.go
StephenButtolph: We switched to using `t.Cleanup` in some of our other packages (in holistic passes). Seems like something that could be pulled out into a followup

vms/proposervm/vm_test.go
StephenButtolph: I'd probably slightly prefer `test` as well - but I don't really care tbh

StephenButtolph (approved):
```
# HELP avalanche_proposervm_last_accepted_timestamp timestamp of the last block accepted
# TYPE avalanche_proposervm_last_accepted_timestamp gauge
avalanche_proposervm_last_accepted_timestamp{block_type="inner",chain="P"} 1.738164828e+09
avalanche_proposervm_last_accepted_timestamp{block_type="proposervm",chain="P"} 1.738164828e+09
```

PR #3685, opened 2025-01-30
[tmpnet] Minimize duration of tx acceptance for e2e testing

maru-ava:
## Why this should be merged

Faster tests mean faster iteration

## How this works

- Enable configuration of min block delay for primary subnet with the addition of node flag `--proposer-min-block-delay`
- Configure tmpnet to default the block delay to zero for all subnets
- Reduce the wallet poll interval from 100ms to 10ms
- Remove unnecessary sleep in virtuous test

## How this was tested

CI against regression, manually verified decreased runtime

New node flag should be documented.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1084,21 +1084,23 @@ func getSubnetConfigsFromDir(v *viper.Viper, subnetIDs []ids.ID) (map[ids.ID]sub
 	}
 
 	subnetConfigs := make(map[ids.ID]subnets.Config)
-	if len(subnetConfigPath) == 0 {
-		// subnet config path does not exist but not explicitly specified, so ignore it
-		return subnetConfigs, nil
-	}
 
 	// reads subnet config files from a path and given subnetIDs and returns a map.
 	for _, subnetID := range subnetIDs {
+		// Ensure default configuration
 		config := getDefaultSubnetConfig(v)
+		subnetConfigs[subnetID] = config
+
+		if len(subnetConfigPath) == 0 {
+			// subnet config path does not exist but not explicitly specified, so ignore it
+			continue
+		}
 
 		filePath := filepath.Join(subnetConfigPath, subnetID.String()+subnetConfigFileExt)
 		fileInfo, err := os.Stat(filePath)
 		switch {
 		case errors.Is(err, os.ErrNotExist):
-			// this subnet config does not exist, use the default configuration
-			subnetConfigs[subnetID] = config
+			// this subnet config does not exist, the default configuration will be used
 			continue
 		case err != nil:
 			return nil, err
@@ -1112,6 +1114,7 @@ func getSubnetConfigsFromDir(v *viper.Viper, subnetIDs []ids.ID) (map[ids.ID]sub
 			return nil, err
 		}
 
+		// Update the default config with the values from the file
 		if err := json.Unmarshal(file, &config); err != nil {
 			return nil, fmt.Errorf("%w: %w", errUnmarshalling, err)
 		}
--- a/config/config.go
+++ b/config/config.go
@@ -1132,7 +1132,7 @@ func getDefaultSubnetConfig(v *viper.Viper) subnets.Config {
 	return subnets.Config{
 		ConsensusParameters:         getConsensusConfig(v),
 		ValidatorOnly:               false,
-		ProposerMinBlockDelay:       proposervm.DefaultMinBlockDelay,
+		ProposerMinBlockDelay:       v.GetDuration(ProposerMinBlockDelayKey),
 		ProposerNumHistoricalBlocks: proposervm.DefaultNumHistoricalBlocks,
 	}
 }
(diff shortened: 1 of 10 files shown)


config/flags.go
```diff
@@ -384,6 +385,9 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.StringToString(TracingHeadersKey, map[string]string{}, "The headers to provide the trace indexer")
 
 	fs.String(ProcessContextFileKey, defaultProcessContextPath, "The path to write process context to (including PID, API URI, and staking address).")
+
+	// Primary Subnet Config
```
darioush, line 389: what does primary subnet here refer to? is it the primary network or a specific subnet?

tests/fixture/tmpnet/defaults.go
```diff
@@ -76,3 +78,11 @@ func DefaultChainConfigs() map[string]FlagsMap {
 		},
 	}
 }
+
+// A set of subnet configuration appropriate for testing.
+func DefaultSubnetConfig() FlagsMap {
```
qdm12, line 83: since this is only used in this package, can we unexport it?

qdm12 (approved):
1. What's the overall time saved?
2. `Reduce the wallet poll interval from 100ms to 10ms` - just as a word of warning, getting close to millisecond in tests can be risky on CI running on shared/slow runners, and produce flaky tests, just because the runner is slow and things take milliseconds to execute... think Pentium 4 running CI 😄

darioush (approved):

config/config.go
```diff
@@ -1132,7 +1132,7 @@ func getDefaultSubnetConfig(v *viper.Viper) subnets.Config {
 	return subnets.Config{
 		ConsensusParameters:         getConsensusConfig(v),
 		ValidatorOnly:               false,
-		ProposerMinBlockDelay:       proposervm.DefaultMinBlockDelay,
+		ProposerMinBlockDelay:       v.GetDuration(ProposerMinBlockDelayKey),
```
StephenButtolph, line 1135: This causes some weirdness in the defaults.

1. The default value of `ProposerMinBlockDelay` for all subnets is currently `proposervm.DefaultMinBlockDelay`.
2. With this change, the default value for subnets that do not specify a custom subnet config is still `proposervm.DefaultMinBlockDelay`.
3. With this change, the default value for subnets that do specify a custom subnet config is equal to `--proposer-min-block-delay`.

I think (2) and (3) should be unified.

config/keys.go
```diff
@@ -208,4 +208,5 @@ const (
 	TracingExporterTypeKey                             = "tracing-exporter-type"
 	TracingHeadersKey                                  = "tracing-headers"
 	ProcessContextFileKey                              = "process-context-file"
+	ProposerMinBlockDelayKey                           = "proposer-min-block-delay"
```
StephenButtolph, line 211: Should we include this in `RELEASES.md` (and maybe in `config.md` as well?)

StephenButtolph, line 211: Also wondering if this is only supposed to impact the primary network - if we should name it something like `--primary-network-min-block-delay`

config/config.go
```diff
@@ -1054,8 +1054,9 @@ func getSubnetConfigsFromFlags(v *viper.Viper, subnetIDs []ids.ID) (map[ids.ID]s
 
 	res := make(map[ids.ID]subnets.Config)
 	for _, subnetID := range subnetIDs {
+		config := getDefaultSubnetConfig(v)
```
aaronbuchwald, line 1057: I see, so here we guarantee that for each tracked `subnetID` we start with the default and override as provided.

config/config.go
```diff
@@ -1090,11 +1091,14 @@ func getSubnetConfigsFromDir(v *viper.Viper, subnetIDs []ids.ID) (map[ids.ID]sub
 
 	// reads subnet config files from a path and given subnetIDs and returns a map.
 	for _, subnetID := range subnetIDs {
+		config := getDefaultSubnetConfig(v)
```
aaronbuchwald, line 1094: The comment added to `chains/manager.go` implies that we now require a default config is present for each `subnetID` that we're tracking. Do we need to guarantee that here as well? It seems that if `subnetConfigPath` is empty, then we'll exit before getting to this point and may not guarantee that?

config/config_test.go
```diff
@@ -474,13 +476,6 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
-		"subnet is not tracked": {
-			givenJSON: `{"Gmt4fuNsGJAd2PX86LBvycGaBpgCYKbuULdCLZs3SEs1Jx1LU":{"validatorOnly":true}}`,
-			testF: func(require *require.Assertions, given map[ids.ID]subnets.Config) {
-				require.Empty(given)
-			},
-			expectedErr: nil,
-		},
```
aaronbuchwald, line 483: Should we keep this test as it's still relevant and should produce the same result as the new case `default config used when no config provided` ?

config/config.go
aaronbuchwald: It seems like this actually can't cause a problem because there are two behaviors:
- use the default path where this will be non-empty and therefore not exit early
- use a non-default path, where it will then check that the directory exists, which will cause an error if the subnetConfigPath is in fact an empty string

StephenButtolph (approved):

PR #3688, opened 2025-01-31
Fix change/range proofs + simplify the code

rrazvan1:
## Why this should be merged

Fixing proof verification bugs:
1. exclusion proofs -> passing when providing a proof path which has the last node: a sibling of an ancestor of the existing node
2. exclusion proofs -> passing when providing not fully extended proof
This issues were mainly affecting the simple proofs verification, but also change/range proofs.

And also, simplifying the code for better readability and maintainability.

## How this works
- moving the main logic of proofs validation inside **verifyProofPath(proofPath []ProofNode, key Key)** and simplifying the specific proof verifications.
- added new specific errors
- more details within comments
- unifying change/range proof flows

## How this was tested
UTs

### DIFF
--- a/x/merkledb/db.go
+++ b/x/merkledb/db.go
@@ -442,19 +442,19 @@ func (db *merkleDB) CommitRangeProof(ctx context.Context, start, end maybe.Maybe
 		return database.ErrClosed
 	}
 
-	ops := make([]database.BatchOp, len(proof.KeyValues))
-	keys := set.NewSet[string](len(proof.KeyValues))
-	for i, kv := range proof.KeyValues {
+	ops := make([]database.BatchOp, len(proof.KeyChanges))
+	keys := set.NewSet[string](len(proof.KeyChanges))
+	for i, kv := range proof.KeyChanges {
 		keys.Add(string(kv.Key))
 		ops[i] = database.BatchOp{
 			Key:   kv.Key,
-			Value: kv.Value,
+			Value: kv.Value.Value(),
 		}
 	}
 
 	largestKey := end
-	if len(proof.KeyValues) > 0 {
-		largestKey = maybe.Some(proof.KeyValues[len(proof.KeyValues)-1].Key)
+	if len(proof.KeyChanges) > 0 {
+		largestKey = maybe.Some(proof.KeyChanges[len(proof.KeyChanges)-1].Key)
 	}
 	keysToDelete, err := db.getKeysNotInSet(start, largestKey, keys)
 	if err != nil {
@@ -1098,39 +1098,28 @@ func (db *merkleDB) VerifyChangeProof(
 		return ErrEmptyProof
 	}
 
-	keyChanges := make(map[Key]maybe.Maybe[[]byte], len(proof.KeyChanges))
-	keys := make([]Key, len(proof.KeyChanges))
-	for i, keyValue := range proof.KeyChanges {
-		k := ToKey(keyValue.Key)
-		if k.hasPartialByte() {
-			return ErrProofKeyPartialByte
-		}
-
-		keyChanges[k] = keyValue.Value
-		keys[i] = k
-	}
-
-	// Validate proof.
-	if err := validateChangeProof(start, end, proof.StartProof, proof.EndProof, keys, db.tokenSize); err != nil {
-		return err
-	}
-
-	// [startProof] is an inclusion/exclusion proof of [startKey]
 	startProofKey := maybe.Bind(start, ToKey)
-
-	// [endProof] is an inclusion of the largest key in [keyValues],
-	// or an exclusion proof of [end] if [keyValues] is empty.
 	endProofKey := maybe.Bind(end, ToKey)
 
 	// Update [endProofKey] with the largest key in [keyValues].
 	if len(proof.KeyChanges) > 0 {
-		// If [proof] has key-value pairs, we should insert children
-		// greater than [largestProvenKey] to ancestors of the node containing
-		// [largestProvenKey] so that we get the expected root ID.
 		endProofKey = maybe.Some(ToKey(proof.KeyChanges[len(proof.KeyChanges)-1].Key))
 	}
 
-	// want to prevent commit writes to DB, but not prevent DB reads
+	// Validate proof.
+	if err := validateChangeProof(
+		startProofKey,
+		maybe.Bind(end, ToKey),
+		proof.StartProof,
+		proof.EndProof,
+		proof.KeyChanges,
+		endProofKey,
+		db.tokenSize,
+	); err != nil {
+		return err
+	}
+
+	// Prevent commit writes to DB, but not prevent DB reads
 	db.commitLock.RLock()
 	defer db.commitLock.RUnlock()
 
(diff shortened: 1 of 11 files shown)


x/merkledb/db.go
```diff
@@ -1078,53 +1078,43 @@ func (db *merkleDB) VerifyChangeProof(
...
-		return err
 	}
 
-	smallestKey := maybe.Bind(start, ToKey)
-
-	// Make sure the start proof, if given, is well-formed.
-	if err := verifyProofPath(proof.StartProof, smallestKey); err != nil {
-		return err
-	}
+	keyChanges := make(map[Key]maybe.Maybe[[]byte], len(proof.KeyChanges))
```
joshua-kim, line 1085: I think we can avoid the duplicate creation of this map every time `validateRangeChangeProof` is called by just doing it within `validateRangeChangeProof` if we pass in `proof.KeyChanges` as an arg.

x/merkledb/db.go
```diff
@@ -1078,53 +1078,43 @@ func (db *merkleDB) VerifyChangeProof(
...
+	// Validate proof.
+	err := validateRangeChangeProof(start, end, proof.StartProof, proof.EndProof, keys, db.tokenSize)
+	if err != nil {
 		return err
 	}
 
-	keyValues := make(map[Key]maybe.Maybe[[]byte], len(proof.KeyChanges))
-	for _, keyValue := range proof.KeyChanges {
-		keyValues[ToKey(keyValue.Key)] = keyValue.Value
+	var (
```
joshua-kim, line 1103: nit: I personally dislike the usage of `var` blocks unless they're strictly needed (say to force a type)... I would just use `:=` to create these...

x/merkledb/proof.go
```diff
@@ -247,24 +245,89 @@ type KeyValue struct {
...
+		return ErrEmptyProof
+	case end.IsNothing() && len(keys) == 0 && len(endProof) != 0:
+		return ErrUnexpectedEndProof
+	case start.IsNothing() && len(startProof) > 0:
+		return ErrUnexpectedStartProof
+	case len(endProof) == 0 && (end.HasValue() || len(keys) > 0):
+		return ErrNoEndProof
+	}
+
+	var (
```
joshua-kim, line 295: Same comment regarding the usage of `var`

x/merkledb/proof.go
```diff
@@ -283,71 +346,67 @@ func (proof *RangeProof) Verify(
...
+		keys[i] = k
+	}
+
+	// Validate proof.
+	err := validateRangeChangeProof(start, end, proof.StartProof, proof.EndProof, keys, tokenSize)
+	if err != nil {
+		return err
+	}
+
+	var (
```
joshua-kim, line 376: Same comment w/ `var` usage

x/merkledb/proof_test.go
```diff
@@ -263,49 +352,75 @@ func Test_RangeProof_MaxLength(t *testing.T) {
...
 	require.NoError(err)
 	require.NotNil(dbTrie)
+
+	var (
+		key  = []byte{0x6B, 0x65, 0x79}
+		key0 = []byte{0x6B, 0x65, 0x79, 0x30}
+		key1 = []byte{0x6B, 0x65, 0x79, 0x31}
+		key2 = []byte{0x6B, 0x65, 0x79, 0x32}
+		key3 = []byte{0x6B, 0x65, 0x79, 0x33}
+		key4 = []byte{0x6B, 0x65, 0x79, 0x34}
```
joshua-kim, line 368: Can we just use strings? These bytes values are hard to reason about/look ugly... and chars are 1 byte each + byte representations are visible in the debugger anyways.

x/merkledb/proof_test.go
```diff
@@ -1251,31 +1438,51 @@ func TestVerifyProofPath(t *testing.T) {
...
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := verifyProofPath(tt.path, tt.proofKey)
-			require.ErrorIs(t, err, tt.expectedErr)
+			require := require.New(t)
+
+			err := verifyProofPath(tt.path, tt.proofKey, tokenSize)
```
joshua-kim, line 1484: Since `tokenSize` is only used in this test body can we just define it here? The value is self-documenting since the parameter name is already `tokenSize`.
```suggestion
			err := verifyProofPath(tt.path, tt.proofKey, 8)
```

x/merkledb/proof.go
```diff
@@ -446,49 +283,104 @@ func (proof *RangeProof) UnmarshalProto(pbProof *pb.RangeProof) error {
...
-// in [[start], [end]] have the value given for that key in [keysValues].
-func verifyAllRangeProofKeyValuesPresent(
-	hasher Hasher,
-	proof []ProofNode,
-	start maybe.Maybe[Key],
-	end maybe.Maybe[Key],
-	keysValues map[Key][]byte,
+func (proof *RangeProof) AsChangeProof() *ChangeProof {
+	return (*ChangeProof)(proof)
+}
```
joshua-kim, line 299: This helper isn't needed imo - we can just do a type cast

x/merkledb/proof_test.go
```diff
@@ -263,49 +352,73 @@ func Test_RangeProof_MaxLength(t *testing.T) {
...
 	require.Equal(maybe.Some([]byte("value")), proof.Path[0].ValueOrHash)
-	require.Equal(ToKey([]byte("key0")).Take(28), proof.Path[1].Key)
-	require.True(proof.Path[1].ValueOrHash.IsNothing()) // intermediate node
-	require.Equal(ToKey([]byte("key1")), proof.Path[2].Key)
-	require.Equal(maybe.Some([]byte("value1")), proof.Path[2].ValueOrHash)
 
-	expectedRootID, err := trie.GetMerkleRoot(context.Background())
-	require.NoError(err)
-	require.NoError(proof.Verify(context.Background(), expectedRootID, dbTrie.tokenSize, dbTrie.hasher))
+	k := ToKey([]byte{0x6B, 0x65, 0x79, 0x30}).Take(28)
```
joshua-kim, line 416: Can we write this out as a string?

x/merkledb/proof_test.go
```diff
@@ -22,105 +22,194 @@ import (
...
+		})
 	}
+}
 
-	tests := []test{
+func Test_Proof_Inclusion(t *testing.T) {
+	tests := []struct {
+		name   string
+		modify func(*Proof)
+		expect error
```
joshua-kim, line 98: nit: use `wantErr` or `want` for this name - this file uses `expect/expectedErr` but it's not idiomatic in go.

x/merkledb/proof_test.go
```diff
@@ -22,105 +22,194 @@ import (
...
+	}
 }
 
-func Test_Proof_Simple(t *testing.T) {
-	require := require.New(t)
+func Test_Proof_Exclusion(t *testing.T) {
+	tests := []struct {
+		name   string
+		modify func(*Proof)
+		expect error
```
joshua-kim, line 56: Same nit

x/merkledb/proof_test.go
```diff
@@ -22,105 +22,194 @@ import (
 )
 
 func Test_Proof_Empty(t *testing.T) {
-	proof := &Proof{}
-	err := proof.Verify(context.Background(), ids.Empty, 4, DefaultHasher)
-	require.ErrorIs(t, err, ErrEmptyProof)
+	tests := []struct {
+		name   string
+		proof  *Proof
+		expect error
```
joshua-kim, line 28: Same nit

x/merkledb/proof_test.go
```diff
@@ -22,105 +22,194 @@ import (
...
+		})
+	}
 }
 
-func Test_Proof_Simple(t *testing.T) {
-	require := require.New(t)
+func Test_Proof_Exclusion(t *testing.T) {
+	tests := []struct {
+		name   string
+		modify func(*Proof)
```
joshua-kim, line 55: I would just write these as two separate tests - I think these `modify` functions (or any setup functions that pass around arguments in test tables for that matter) end up turning the test table into a test anyways.

x/merkledb/proof_test.go
```diff
@@ -263,49 +352,73 @@ func Test_RangeProof_MaxLength(t *testing.T) {
...
 	dbTrie, err := getBasicDB()
 	require.NoError(err)
 	require.NotNil(dbTrie)
+
+	key := []byte("key")
+	key0 := []byte("key0")
+	key1 := []byte("key1")
+	key2 := []byte("key2")
+	key3 := []byte("key3")
+	key4 := []byte("key4")
```
joshua-kim, line 367: Why do we make these variables? We don't end up using them anyways - can we just in-line them like the previous code did?

joshua-kim (approved):

StephenButtolph (approved):
(didn't actually review, `x/` shouldn't require multiple reviews.)

PR #3690, opened 2025-01-31
testing: improve e2e test bootstrapping

tsachiherman:
## Why this should be merged

This PR remove the redundant testing for bootstrapping nodes after the nodes have already been verified to be healthy.

## How this works

trivial.

## How this was tested

Tested using existing tests. 

no.

### DIFF
--- a/tests/fixture/e2e/env.go
+++ b/tests/fixture/e2e/env.go
@@ -160,9 +160,9 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
 				}
 			}
 		}
-		// Wait for chains to have bootstrapped on all nodes
+		// Wait for chains to have bootstrapped on all validators
 		tc.Eventually(func() bool {
-			allNodesBootstrapped := true
+			allValidatorsBootstrapped := true
 			for i := len(pendingBootstappingChains) - 1; i >= 0; i-- {
 				pending := pendingBootstappingChains[i]
 				uri, err := network.GetURIForNodeID(pending.validatorID)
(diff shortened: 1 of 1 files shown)


tests/fixture/e2e/env.go
```diff
@@ -144,23 +145,43 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
...
+			for i := len(pendingBootstappingChains) - 1; i >= 0; i-- {
+				pending := pendingBootstappingChains[i]
+				uri, err := network.GetURIForNodeID(pending.validatorID)
+				require.NoError(err)
+				infoClient := info.NewClient(uri)
+				isBootstrapped, err := infoClient.IsBootstrapped(tc.DefaultContext(), pending.chainID.String())
+				// Ignore errors since a chain id that is not yet known will result in a recoverable error.
+				if err != nil || !isBootstrapped {
+					allValidatorsBootstrapped = false
+					continue
```
aaronbuchwald, line 175: Why do we perform an `IsBootstrapped` check here rather than checking if the node is healthy? If the node reports healthy, all bootstrapping checks should pass and this will include correctly checking that dynamic state sync has completed (which is handled as a bit of an outlier).

tsachiherman: by the time we reach this point, the health check was already complete against all the node. Some differences:
- The health checking doesn't seem to be chain-specific, whereas the bootstrapping test is.
- When doing the health checking, the port would get dynamically refreshed. That's won't happen during the bootstrapping testing.

aaronbuchwald (approved):

PR #3691, opened 2025-02-01
[testing] Replace script-based tool installation with nix

maru-ava:
## Why this should be merged

Previously, binary tools like promtail and prometheus (enabling log and metrics collection) and kube and kind (enabling local kube clusters) were installed using bash scripts. A script-based approach, while simple, requires tedious and error-prone copying and committing to other repos (e.g. subnet-evm and hypersdk) that needed the same functionality.

Possible to rewrite the scripts to golang which allows usage via `go run` or `go install`. Simple enough if the versions never change, but still requires retrieving binaries, verifying their hashes for security, potentially dealing with tar archives. Still more complication if it becomes necessary to ensure the desired version is the one in the path.

Switching to [nix](https://nixos.org/), on the other hand, is an established way of solving the problem that doesn't involve maintaining code. Installing nix is a one-time operation not too different from homebrew, except that the result has more guarantees of reproducibility across linux and macos.
## How this works

 - Replaces manual tool installation with a nix flake
 - Updates github actions to install nix where needed

## How this was tested

CI, manually for direnv usage

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-go-for-project
+      - uses: cachix/install-nix-action@v30
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      # TODO(marun) Maybe figure out how to use `nix build` somehow i.e. make the default shell double as the default package
+      - run: nix develop --command echo "dependencies installed"
       - name: Build AvalancheGo Binary
         shell: bash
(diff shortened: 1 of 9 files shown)


StephenButtolph (approved):

.github/workflows/ci.yml
```diff
@@ -51,6 +51,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-go-for-project
+      - uses: cachix/install-nix-action@v30
+        with:
```
yacovm, line 55: why is this github token is needed while it was not needed in the previous version of this file?

PR #3692, opened 2025-02-02
[tmpnet] Deploy collectors with golang to simplify cross-repo use

maru-ava:
## Why this should be merged

Previously, prometheus and promtail for collecting from local tmpnet networks were installed and launched by bash scripts. Migrating installation to nix and launch to golang simplifies reuse from subnet-evm and hypersdk. No more having to copy and maintain copies of the scripts in multiple repos.

## How this works

 - migrates scripts launching prometheus and promtail to golang  
 - configures collector start via the tmpnet e2e flag `--start-collectors` 
 - enables start and stop of collectors via tmpnetctl

## How this was tested

- [x] CI for regression
- [x] Manually checked that metrics and logs were collected from local and CI networks with the chnages 
  - Automatic checks implemented in https://github.com/ava-labs/avalanchego/pull/3740

N/A

## TODO

- [x] Merge https://github.com/ava-labs/avalanchego/pull/3691
- [x] Refactor for review
- [x] Merge https://github.com/ava-labs/avalanchego/pull/3736


tests/fixture/tmpnet/start_collectors.go
```diff
@@ -0,0 +1,242 @@
...
+
+	fullCmd := "nohup " + cmdName + " " + args + " > " + cmdName + ".log 2>&1 & echo -n \"$!\" > " + pidFilename
+	log.Info("starting "+cmdName,
+		zap.String("workingDir", workingDir),
+		zap.String("fullCmd", fullCmd),
+	)
+
+	// TODO(marun) Figure out a way to redirect stdout and stderr of a detached child process without a bash shell
+	cmd := exec.Command("bash", "-c", fullCmd)
+	configureDetachedProcess(cmd) // Ensure the child process will outlive its parent
```
aaronbuchwald, line 184: I believe you can create the command and set `Stdout` and `Stderr` to nil directly (according to the docs)

aaronbuchwald: I may be wrong here, just remembered reading the documentation around this, so thought I'd provide the suggestion. May not behave as I expected from reading the docs (either due to a misread or incorrect docs).

aaronbuchwald: I think that I thought you were trying to ignore the output here: https://github.com/ava-labs/avalanchego/blob/15c496b09f92cc5ac23b5aa2937d17a258d9a14f/tests/fixture/tmpnet/start_collectors.go#L182, but you actually wanted to write it to the specified file. I think you could still do this by setting `Stdout` and `Stderr` to the corresponding file and avoid constructing a bash command.

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -0,0 +1,496 @@
...
+
+	// Check if the specified command is available in the path
+	if _, err := exec.LookPath(cmdName); err != nil {
+		return fmt.Errorf("%s command not found. Maybe run 'nix develop'?", cmdName)
+	}
+
+	// Write the collector config file
+	confFilename := cmdName + ".yaml"
+	confPath := filepath.Join(workingDir, confFilename)
+	log.Info("writing "+cmdName+" config",
```
StephenButtolph, line 298: I think it's generally frowned upon to have a dynamic log message with zap (and that fields should generally be used instead).

Although it doesn't seem too crazy here - so 🤷

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -0,0 +1,496 @@
...
+	cmd := exec.Command("bash", "-c", fullCmd)
+	configureDetachedProcess(cmd) // Ensure the child process will outlive its parent
+	cmd.Dir = workingDir
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("failed to start %s: %w", cmdName, err)
+	}
+
+	// Wait for PID file
+	var pid int
+	if err := pollUntilContextCancel(ctx, func(_ context.Context) (bool, error) {
```
StephenButtolph, line 415: Small style nit: I find the indentation here pretty odd. Specifically because the `return pid != 0, nil` happens at the same indentation level as `return err`, but they are actually referring to different functions entirely.

Personally, I try to avoid multi-line `if statement; condition {`s because it just looks really weird to me.

StephenButtolph (approved):

aaronbuchwald (approved):

PR #3696, opened 2025-02-03
Make `bls.Signer` api fallible

richardpringle:
- **Make ProofOfPossession creation fallible**
- **Make bls.Signer api fallible**

## Why this should be merged

If we want to add any networking in between the keys, we'll need a fallible API.

I've implemented the `rpcsigner` here as a grpc client with a server implemented in a `_test.go` file. The idea here is that anyone who implements the interface.

### Open Question:
There's one little gotchya here for anyone implementing the interface. It's that we expect the keys and signatures to be compressed when sent over the wire. I'm open to using the more general serialization OR adding the `compressed` prefix to the proto-message field names. 

### TODO: 
Further, we should probably add a minimal test suite that an external implementor can run, however, I think could probably be implemented in a follow up PR as it is not a priority at the moment.


utils/crypto/bls/signer.go
```diff
@@ -4,7 +4,7 @@
 package bls
 
 type Signer interface {
-	PublicKey() *PublicKey
-	Sign(msg []byte) *Signature
-	SignProofOfPossession(msg []byte) *Signature
+	PublicKey() (*PublicKey, error)
```
joshua-kim, line 7: Why does this need an error? Can't you the implementation cache the public key when the signer is initialized since it's always constant?

chains/manager.go
```diff
@@ -490,13 +490,15 @@ func (m *manager) buildChain(chainParams ChainParameters, sb subnets.Subnet) (*c
...
+	pubKey := m.StakingBLSKey.PublicKey()
+
 	ctx := &snow.ConsensusContext{
 		Context: &snow.Context{
 			NetworkID:       m.NetworkID,
 			SubnetID:        chainParams.SubnetID,
 			ChainID:         chainParams.ID,
 			NodeID:          m.NodeID,
-			PublicKey:       m.StakingBLSKey.PublicKey(),
+			PublicKey:       pubKey,
```
joshua-kim, line 501: I think we can undo this diff now that `PublicKey` isn't erroring (and in a bunch of other spots in this PR)

network/peer/ip.go
```diff
@@ -39,7 +39,15 @@ func (ip *UnsignedIP) Sign(tlsSigner crypto.Signer, blsSigner bls.Signer) (*Sign
...
-	blsSignature := blsSigner.SignProofOfPossession(ipBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	blsSignature, err := blsSigner.SignProofOfPossession(ipBytes)
+	if err != nil {
+		return nil, err
+	}
+
```
joshua-kim, line 50: Won't let me comment on the line, but we either need to return `nil` at the end of this signature (currently returns `err`), or keep the previous behavior where we don't do the `nil` check and return the `err` alongside `SignedIP`. I personally don't like returning errors with non-nil/zero values, but I think Stephen prefers it because he thinks it looks cleaner/less LOC.

utils/crypto/bls/signer/rpcsigner/bls_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	msg := []byte("TestSingleKeyAggregation local signer")
+
+	sig, err := signer.Sign(msg)
+	require.NoError(err)
+
+	isValid, err := AggregateAndVerify(pks, []*bls.Signature{sig}, msg)
+	require.NoError(err)
+	require.True(isValid)
+}
```
joshua-kim: Do we need tests on aggregation working? I feel like we already know aggregation will work because we are just implementing a type that produces `bls.Signature` and `bls.Publickey`, and those types + aggregation are already tested in the upstream package `bls`. I feel like coverage for `rpcsigner` only needs to cover that the `Sign*` functions produce signatures that are verifiable off of the `PublicKey`.

utils/crypto/bls/signer/rpcsigner/server_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
+)
+
+type server struct {
+	signer.UnimplementedSignerServer
+	localSigner *localsigner.LocalSigner
+}
+
+func newServer() (*server, error) {
```
joshua-kim, line 16: Should this be exported + rename this so it's not a test file?

utils/crypto/bls/signer/rpcsigner/bls_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+func TestVerifyValidProofOfPossession(t *testing.T) {
```
joshua-kim, line 176: Similar comment regarding test tables w/ the following test

proto/signer/signer.proto
```diff
@@ -0,0 +1,28 @@
...
+}
+
+message PublicKeyRequest {}
+message PublicKeyResponse {
+  bytes public_key = 1;
+}
+message SignatureRequest {
+  bytes message = 1;
+}
+message SignatureResponse {
```
joshua-kim, line 20: nit: `SignRequest/SignResponse`

proto/signer/signer.proto
```diff
@@ -0,0 +1,28 @@
...
+  bytes signature = 1;
+}
+message ProofOfPossessionSignatureRequest {
+  bytes message = 1;
+}
+message ProofOfPossessionSignatureResponse {
```
joshua-kim, line 26: nit: `SignProofOfPossessionSignatureRequest/Response`

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (c *Client) SignProofOfPossession(message []byte) (*bls.Signature, error) {
+	resp, err := c.client.SignProofOfPossession(context.Background(), &pb.ProofOfPossessionSignatureRequest{Message: message})
+	if err != nil {
+		return nil, err
+	}
+	return getSignatureFromResponse(resp)
+}
+
+func (c *Client) PublicKey() *bls.PublicKey {
```
joshua-kim, line 88: nit: put this before `Sign` to be consistent w/ ordering on the interface

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,90 @@
...
+	signature := resp.GetSignature()
+	if signature == nil {
+		return nil, errors.New("empty signature")
+	}
+
+	return bls.SignatureFromBytes(signature)
+}
+
+func (c *Client) Sign(message []byte) (*bls.Signature, error) {
+	resp, err := c.client.Sign(context.Background(), &pb.SignatureRequest{Message: message})
```
joshua-kim, line 72: Use `TODO` instead of `Background`

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,94 @@
...
+	conn   *grpc.ClientConn
+	pk     *bls.PublicKey
+}
+
+func NewClient(conn *grpc.ClientConn) (*Client, error) {
+	client := pb.NewSignerClient(conn)
+
+	pubkeyResponse, err := client.PublicKey(context.TODO(), &pb.PublicKeyRequest{})
+	if err != nil {
+		conn.Close()
```
joshua-kim, line 34: I feel weird about this code owning the lifecycle of the connection since it's provided by the caller. Some of these errors don't look [persistent](https://grpc.io/docs/guides/error/) + you can still re-use the connection since you can retry rpcs on different streams on the same http/2 connection even if this errored out.

You could probably make a strong case for some of these errors being persistent (i.e the empty public key case, the invalid pk case), but I would still think that you could handle all of these errors at the call-site through a connection close and have all the connection lifecycle managed there instead of having mixed ownership

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,94 @@
...
+	return &Client{
+		client: client,
+		conn:   conn,
+		pk:     pk,
+	}, nil
+}
+
+func (c *Client) Close() error {
+	return c.conn.Close()
+}
```
joshua-kim, line 60: Related to my last [comment](https://github.com/ava-labs/avalanchego/pull/3696/files#r1949626374) but I think this can be removed as well

utils/crypto/bls/signer/rpcsigner/client_test.go
```diff
@@ -0,0 +1,158 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			sig, err := client.SignProofOfPossession(test.msg)
+			require.Nil(t, sig)
+			require.EqualError(t, err, test.err.Error())
+		})
+	}
+}
+
+func (*testClient) PublicKey(_ context.Context, _ *signer.PublicKeyRequest, _ ...grpc.CallOption) (*signer.PublicKeyResponse, error) {
+	return nil, nil
```
joshua-kim, line 92: I think this is supposed to return the pk instead of nil - my guess is that this isn't caught because of the missing assertions

utils/crypto/bls/signer/rpcsigner/client_test.go
```diff
@@ -0,0 +1,158 @@
...
+	case string(uncompressedSignatureMsg):
+		sig, err := localSigner.Sign(in.Message)
+		if err != nil {
+			return nil, err
+		}
+
+		bytes := sig.Serialize()
+
+		return &signer.SignResponse{
+			Signature: bytes[:bls.SignatureLen],
```
joshua-kim, line 115: Why do we reslice here? Is this this sig otherwise not 96 bytes?

joshua-kim (approved):

utils/crypto/bls/signer/localsigner/bls_test.go
```diff
@@ -24,79 +24,89 @@ func AggregateAndVerify(publicKeys []*bls.PublicKey, signatures []*bls.Signature
...
+	return sk
+}
+
+func Map[T any, U any](arr []T, f func(T) U) []U {
+	result := make([]U, len(arr))
+	for i, val := range arr {
+		result[i] = f(val)
+	}
+	return result
 }
```
StephenButtolph, line 39: I think we should try to be in the habit of implementing helpers in better places. So if we actually want to add this it should probably be in a place that can be shared.

That being said, I don't really think map/filter are very standard for golang (they are more natural in a more functional language).

They were actually declined to be added to the stdlib here: https://github.com/golang/go/discussions/47203 (That being said, it wasn't a final decision. It was left as a potential future addition).

utils/crypto/bls/signer/localsigner/bls_test.go
```diff
@@ -120,19 +131,18 @@ func TestSingleKeyAggregation(t *testing.T) {
...
 	require := require.New(t)
-	sk1, pk1 := NewKeyPair(require)
-	sk2, pk2 := NewKeyPair(require)
-	sk3, pk3 := NewKeyPair(require)
-
-	pks := []*bls.PublicKey{pk1, pk2, pk3}
+	signers := []*LocalSigner{NewSigner(require), NewSigner(require), NewSigner(require)}
+	pks := Map(signers, func(signer *LocalSigner) *bls.PublicKey {
+		return signer.PublicKey()
+	})
```
StephenButtolph, line 137: This applies to a number of areas, but if we are going functional:
```suggestion
	pks := Map(signers, (*LocalSigner).PublicKey)
```

utils/crypto/bls/signer/localsigner/bls_test.go
```diff
@@ -141,20 +151,24 @@ func TestIncorrectMessageAggregation(t *testing.T) {
...
-		sk3.Sign(differentMsg),
-	}
+	diffSig, err := signers[0].Sign(differentMsg)
+	require.NoError(err)
+
+	sameSigs := Map(signers[1:], func(signer *LocalSigner) *bls.Signature {
+		sig, err := signer.Sign(msg)
+		require.NoError(err)
+		return sig
+	})
```
StephenButtolph, line 169: I personally find:
```suggestion
	sameSigs := make([]*bls.Signature, len(signers)-1)
	for i, signer := range signers[1:] {
		sig, err := signer.Sign(msg)
		require.NoError(err)
		sameSigs[i] = sig
	}
```
more natural for a golang dev. Using stdlib/language features should generally be preferred over (unless there is an actual gain out of said implementation).

Basically I think the boilerplate of:
```go
newSlice := make([]T, len(oldSlice))
for i, v := range oldSlice {
    // XXX
   newSlice[i] = newV
}
```

Should be preferred over:
```go
newSlice := Map(oldSlice, func(T) V {
    // XXX
   return newV
})
```

cc: @ARR4N would like to hear your thoughts here.

StephenButtolph: All that being said, I think the `(*testing.T, []byte, []*LocalSigner) -> []*bls.Signature` should be implemented as a helper function because it is used in a bunch of places.

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,87 @@
...
+		conn:   conn,
+		pk:     pk,
+	}, nil
+}
+
+type signatureResponse interface {
+	GetSignature() []byte
+}
+
+func getSignatureFromResponse[T signatureResponse](resp T) (*bls.Signature, error) {
```
StephenButtolph, line 59: The generic type parameter doesn't do anything here - so it should just be removed.
```suggestion
func getSignatureFromResponse(resp signatureResponse) (*bls.Signature, error) {
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -0,0 +1,87 @@
...
+	signature := resp.GetSignature()
+	if len(signature) == 0 {
+		return nil, ErrorEmptySignature
+	}
+
+	return bls.SignatureFromBytes(signature)
```
StephenButtolph, line 65: I'm not sure why we distinguish between invalid and empty signatures.
```suggestion
	signature := resp.GetSignature()
	return bls.SignatureFromBytes(signature)
```
(which tbh makes me wonder if we even need this helper)

utils/crypto/bls/signer/localsigner/bls_test.go
ARR4N: TL;DR I could be convinced of both arguments.

I'll think "out loud":

1. My ultimate goal in readability / style is to reduce cognitive load for the _reader_ of the code. This allows them (us) to more easily parse what's going on when trying to find bugs, make changes, etc.
2. Characteristics that feed into reduced cognitive load are idioms, abstractions, reduced nesting, blah blah blah. This could usually go unsaid, but I think it's important to call out idioms as "does this feel like Go" because if it doesn't then it's a distraction and that adds cognitive load.
3. For me, decisions like this boil down to cost-benefit analysis. For helper functions, the typical benefit is reducing code while the typical cost is distracting the reader by causing a cognitive branch as they read the helper. We don't have deep meat stacks in our skulls so that push/pop is expensive.

_Usually_ I'd say the saving here (the `make()`) isn't worth the distraction and that a little repetition isn't the end of the world (i.e. with @StephenButtolph). However `Map()`, `Reduce(), `Filter()` are all extremely well-known concepts so I don't think it's a distraction. If anything, it's a signal of intent to the reader, which is akin to the benefit of an idiom (i.e. with @richardpringle).

That was a very **long-winded way of saying that I'm on the fence** and quite stable but if I were to fall it would be towards a shared package that introduces those functions. Just please give it a name that communicates something (maybe `fn` because these are functional concepts) and not "underscore" like in JS! It's also not appropriate to export `bls.Map()`.

ARR4N: I missed that this was just in tests as I only read Stephen's comment. If this were my PR I'd keep the function but rename it to something like `mapped()`. `Map()` being exported is very suggestive of it being a function in the package, not a test helper, which is a bit jarring. As I said before, I really don't have strong opinions either way (but do want a green bike shed).

PR #3697, opened 2025-02-04
[tmpnet] Re-enable reuse of dynamically allocated API ports

maru-ava:
## Why this should be merged

By popular demand, tmpnet will once again attempt to reuse dynamically allocated API ports. The only difference is that this needs to be explicitly configured (via `Network.DefaultRuntimeConfig.ReuseDynamicPorts`) to allow for an error to be reported if a network so configured is started with a kube runtime.

## How this works

- Enables a dynamically-allocated API port to be saved in the node flags for use on a subsequent restart if `ReuseDynamicPorts` is set

## How this was tested

CI

N/A

### DIFF
--- a/tests/fixture/tmpnet/README.md
+++ b/tests/fixture/tmpnet/README.md
@@ -89,10 +89,13 @@ A temporary network can be managed in code:
 
 ```golang
 network := &tmpnet.Network{                   // Configure non-default values for the new network
+    DefaultRuntimeConfig{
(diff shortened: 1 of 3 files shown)


iansuvak (approved):
LGTM and thanks! I am guessing the error report if this is attempted to be used from Kube runtime would be handled elsewhere?

aaronbuchwald (approved):
LGTM - what's the reason for removing this feature and now bringing it back?

PR #3705, opened 2025-02-06
Remove RequestBuildBlock on P-Chain Mempool

joshua-kim:
## Why this should be merged

This function doesn't seem to belong on the mempool type because consensus is responsible for building the block. I think another alternative would be to put `RequestBuildBlock` onto a static function somewhere, but I'm not sure where it would live (probably not the mempool package) + I don't think the cost of duplicate code is bad here.

## How this works

Removes `RequestBuildBlock` and replaces the usage of it with usage of the `toEngine` channel

## How this was tested

CI

No

### DIFF
--- a/vms/platformvm/txs/mempool/mempool.go
+++ b/vms/platformvm/txs/mempool/mempool.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 
-	"github.com/ava-labs/avalanchego/snow/engine/common"
 	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
 
 	txmempool "github.com/ava-labs/avalanchego/vms/txs/mempool"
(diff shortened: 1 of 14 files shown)


Banjosa07:
This is strange.

Banjosa07:
Denial at it's highest peak

vms/platformvm/txs/mempool/mempool.go
```diff
@@ -23,37 +22,21 @@ var (
 
 type Mempool interface {
```
StephenButtolph, line 23: Seems like we shouldn't have this interface anymore.

StephenButtolph (commented):
1 nit + needs a rebase and then lgtm

vms/platformvm/txs/mempool/mempool.go
joshua-kim: Removed

StephenButtolph (approved):

rrazvan1 (approved):

PR #3709, opened 2025-02-07
Block context canoto

aaronbuchwald:
This PR adds Canoto serialization support for the `block.Context` type, so that downstream consumers that want to serialize it with Canoto can do so easily.

Updates minimum golang version to 1.22.9 as required by Canoto.


go.mod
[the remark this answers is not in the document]
aaronbuchwald: These were updated automatically after importing canoto. Reverting them manually and running `go mod tidy` reverts back to the current state, so I think these are the minimum versions of canoto.

Downgrading each within canoto repo, it seems that it is depending on these versions and downgrading to the previous versions required by AvalancheGo does cause a break.

The updated dependencies are:
- Cobra v1.8.1 (not the latest release) https://github.com/spf13/cobra/tree/v1.8.1
- testify v1.10.0 (latest release and November 2024)
- fzgen v0.4.3 (latest release as of 1 year ago)

Cobra and testify are very low risk imo. fzgen may be a bit higher risk to update since it is less commonly used.

snow/engine/snowman/block/block_context_vm.go
```diff
@@ -21,7 +21,9 @@ type Context struct {
 	//
 	// Because PreForkBlocks and PostForkOptions do not verify their execution
 	// against the P-chain's state, this context is undefined for those blocks.
-	PChainHeight uint64
+	PChainHeight uint64 `canoto:"fint64,1" json:"pChainHeight"`
```
StephenButtolph, line 24: I don't think this should be a `fint64` - I think this should just be an `int`. (The `PChainHeight` is not going to be > 2^56 for a long time). Even if this was `fint32` I think `int` is more efficient for the foreseeable future.

snow/engine/snowman/block/block_context_vm.go
StephenButtolph, line 1: Can we add a `go:generate` at the top of this file?

scripts/lint.sh
StephenButtolph, line 1: Do we run `go generate` during CI to ensure the generated files are up-to-date?

go.mod
```diff
@@ -12,6 +12,7 @@ go 1.23.6
 require (
 	github.com/DataDog/zstd v1.5.2
 	github.com/NYTimes/gziphandler v1.1.1
+	github.com/StephenButtolph/canoto v0.10.0
```
StephenButtolph, line 15: I'm a bit torn on if we should wait for `v1.0.0` before integrating (I think adding maps and https://github.com/StephenButtolph/canoto/issues/75 are all we'd want to do before calling it `v1`)

aaronbuchwald: I don't have a strong opinion, but this is blocking updating HyperSDK to the latest commit of AvalancheGo, which is required to use the latest tmpnet work.

scripts/lint.sh
aaronbuchwald: Nice https://github.com/ava-labs/avalanchego/actions/runs/13708501315/job/38339528523?pr=3709

.github/workflows/ci.yml
```diff
@@ -155,6 +155,16 @@ jobs:
...
+  check_canotogen:
+    name: Up-to-date canoto
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-go-for-project
+      - shell: bash
+        run: go generate -run "github.com/StephenButtolph/canoto/canoto" ./...
+      - shell: bash
+        run: .github/workflows/check-clean-branch.sh
```
aaronbuchwald, line 167: Copied from directly above

go.mod
StephenButtolph: no need to wait

PR #3711, opened 2025-02-08
perf(merkledb): optimize root hash calculation in view operations

VolodymyrBg:
## Why this should be merged

This PR optimizes the root hash calculation in the MerkleDB view operations by:
1. Eliminating redundant hash calculations of the root node
2. Caching the root hash during view initialization
3. Improving performance by avoiding recalculation during node operations

The change addresses a TODO comment in the codebase that highlighted the need for this optimization.

## How this works

The implementation:
1. Adds a new `rootID` field to the `view` struct to cache the root hash
2. Calculates the root hash once during view initialization in `newView()`
3. Uses the pre-calculated hash instead of recalculating it during node operations
4. Removes redundant hash calculation in node manipulation code

This change ensures that the root hash is calculated only when necessary and stored for reuse, improving performance especially in scenarios with frequent node operations.

## How this was tested

The changes were tested through:
1. Existing unit tests for view operations
2. Performance benchmarks comparing hash calculation frequency
3. Verification of correct hash propagation through view hierarchy
4. Edge cases testing with empty views and root modifications

Yes, this change should be documented in RELEASES.md under performance improvements:

### DIFF
--- a/x/merkledb/view.go
+++ b/x/merkledb/view.go
@@ -98,6 +98,10 @@ type view struct {
 	// The root of the trie represented by this view.
 	root maybe.Maybe[*node]
 
+	// The ID of the root node, calculated during view creation
+	// This avoids recalculating it during node operations
+	rootID ids.ID
+
 	tokenSize int
 }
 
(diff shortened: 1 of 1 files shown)


StephenButtolph:
Closing as this PR doesn't compile. Thanks for the submission

PR #3713, opened 2025-02-11
[ci] Fix metrics link annotation emitted for e2e and upgrade jobs

maru-ava:
## Why this should be merged

Previously, the upgrade job was configured to filter the metrics link annotation with `network_owner=avalanchego-e2e`. Not sure how it ended up that way, because only the e2e job should be so filtered. This was preventing the metrics link for both jobs from displaying results properly. This didn't effect links emitted as part of network start, only the link emitted as an annotation.

## How this works

- Moved `filter_by_owner` arg from the upgrade to the e2e job
- Also updated tmpnet README to better document the dashboard link annotation

## How this was tested

CI

N/A


qdm12 (approved):
💯 👍

StephenButtolph (approved):

PR #3719, opened 2025-02-13
Healthcheck for zero ingress connection count

yacovm:
## Why this should be merged

This PR adds a new health check that monitors if the node is a primary network validator and has no node connecting to it. 

Our current health checks only monitor whether the node is connected to enough stake. However, a node may be connected to enough stake by opening egress connections to all other validators, but also at the same time, not be reachable from any validator or non-validator node. 

This health check does just that - it alerts when:

1. The node is a primary network validator.
2. The node has zero ingress connections to it.
3. More than 10 minutes (configurable) passed since its startup.

## How this works

A node can establish a connection through two pathways:

1. It initiates a connection with a remote node
2. A remote node initiates connection to it.

We count the number of connections of the latter type via an atomic counter, and add a health check that checks the three conditions mentioned earlier. 

## How this was tested

Added unit tests and also did a manual test:

I launched a validator [node](https://snowpeer.io/validators/performance?nodeID=NodeID-49DuEYJ8Kub8pdHz6yucyVqKBhEw5urMZ&network=fuji) on Fuji (testnet) and I added it to the validators:

```
03-03|17:34:09.380] INFO <P Chain> validators/logger.go:50 node added to validator set {"subnetID": "11111111111111111111111111111111LpoYY", "nodeID": "NodeID-49DuEYJ8Kub8pdHz6yucyVqKBhEw5urMZ", "publicKey": "0x80f073c922fbfe9fe8ee22a5c4d47f8b1c30d6c95574d7f5fb9d7d673f5170141751d51b2c9e6ca0a1ca2f2acde2477b", "txID": "2XuGeB5LeKcrR5N2yjPvWMJj19NVE7kTiLk6gzU437sL9qjDZN", "weight": 2991109255}

```

```
~/avalanchego$ curl -s  -H 'Content-Type: application/json' --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"health.health",
    "params": {
        "tags": ["11111111111111111111111111111111LpoYY", "29uVeLPJB1eQJkzRemU8g8wZDw5uJRqpab5U2mX9euieVwiEbL"]
    }
}' 'http://localhost:9650/ext/health' | jq '.result.checks.network.message["primary network validator health"]'
{
  "ingressConnectionCount": 1708,
  "primary network validator": true
}


```

I then caused all other nodes to disconnect from it by blocking port 9651:

```
sudo iptables -I INPUT -p tcp --dport 9651 -j DROP
```

After a short time, the health check started failing:

```
[03-03|19:08:12.853] INFO health/worker.go:261 check started passing {"name": "health", "name": "bootstrapped", "tags": ["application"]}
[03-03|19:08:12.853] INFO health/worker.go:261 check started passing {"name": "readiness", "name": "bootstrapped", "tags": ["application"]}
[03-03|19:10:42.853] WARN health/worker.go:252 check started failing {"name": "health", "name": "network", "tags": ["application"], "error": "network layer is unhealthy reason: primary network validator is unreachable"}

```

and probing the health check API showed it: 

```
~/avalanchego$ curl -s  -H 'Content-Type: application/json' --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"health.health",
    "params": {
        "tags": ["11111111111111111111111111111111LpoYY", "29uVeLPJB1eQJkzRemU8g8wZDw5uJRqpab5U2mX9euieVwiEbL"]
    }
}' 'http://localhost:9650/ext/health' | jq '.result.checks.network'
{
  "message": {
    "connectedPeers": 227,
    "primary network validator health": {
      "ingressConnectionCount": 0,
      "primary network validator": true
    },
    "sendFailRate": 0,
    "timeSinceLastMsgReceived": "853.896432ms",
    "timeSinceLastMsgSent": "853.896432ms"
  },
  "error": "network layer is unhealthy reason: primary network validator is unreachable",
  "timestamp": "2025-03-03T19:11:12.853915018Z",
  "duration": 25832,
  "contiguousFailures": 2,
  "timeOfFirstFailure": "2025-03-03T19:10:42.853036554Z"
}


```

I then restored connectivity via deleting the `iptables` rule:

```
sudo iptables -D INPUT -p tcp --dport 9651 -j DROP
```

and observed that the health check recovered:

```
[03-03|19:12:12.853] INFO health/worker.go:261 check started passing {"name": "health", "name": "network", "tags": ["application"]}
```

```
~/avalanchego$ curl -s  -H 'Content-Type: application/json' --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"health.health",
    "params": {
        "tags": ["11111111111111111111111111111111LpoYY", "29uVeLPJB1eQJkzRemU8g8wZDw5uJRqpab5U2mX9euieVwiEbL"]
    }
}' 'http://localhost:9650/ext/health' | jq '.result.healthy'
true
```

Added a section about the health check added.


network/network.go
```diff
@@ -256,30 +267,34 @@ func NewNetwork(
 		ipTracker.ManuallyTrack(nodeID)
 	}
 
+	// ingressConnCount tracks connections to all remote peers,
+	// so for safety we set it as pointer, in case the config object is copied.
+	var ingressConnCount atomic.Uint32
```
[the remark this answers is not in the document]
tsachiherman: Separately, I'd suggest that you'll be using a Uint64 whenever possible and avoid Uint32. On most 64-bit machines, the 32-bit version tend to be slower.

network/network_test.go
```diff
@@ -282,6 +282,11 @@ func newFullyConnectedTestNetwork(t *testing.T, handlers []router.InboundHandler
 		if i != 0 {
 			config := configs[0]
 			net.ManuallyTrack(config.MyNodeID, config.MyIPPort.Get())
+			// Wait until the node is connected to the first node.
+			// This forces nodes to connect to each other in a deterministic order.
+			require.Eventually(func() bool {
+				return len(net.PeerInfo([]ids.NodeID{config.MyNodeID})) > 0
+			}, time.Minute, time.Millisecond)
```
tsachiherman, line 289: nit: On the success case, it would typically take 10-50ms before this statement completes.
On the fail case, it would take at least one minute. Would you consider dial the fail case threshold down to 10 seconds ?

network/network_test.go
```diff
@@ -306,6 +311,40 @@ func TestNewNetwork(t *testing.T) {
...
+		return result
+	}, time.Minute, time.Millisecond*10)
+
+	ingressConnections := make([]int, 0, len(networks))
+
+	for _, net := range networks {
+		ingressConnections = append(ingressConnections, net.IngressConnCount())
+	}
+
+	wg.Done()
```
tsachiherman, line 340: I think that the `wg.Done()` should be moved above the `Eventually` since it's a predecessor for the `Dispatch` call, which is required for the `Eventually`.

node/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,94 @@
...
+
+	connCount := nica.ingressConnections.IngressConnCount()
+	_, areWeValidator := nica.validators.GetValidator(constants.PrimaryNetworkID, nica.selfID)
+
+	if connCount > 0 {
+		nica.clearError(connCount, areWeValidator)
+		return nica.lastResult, nica.lastError
+	}
+
+	if !areWeValidator {
```
tsachiherman, line 78: nit: combine the two conditional cases.

tsachiherman: scratch that. Instead, move the `if connCount > 0 {`, right after the `connCount := nica. ...`.
That would prevent us from redundantly calling the `GetValidator`.

tsachiherman (commented):
Overall, looks great. I've added few comments.

node/no_ingress_conn_alert.go
yacovm: I made it this way on purpose, so we can also incorporate the "areWeValidator" in the health check output. 

The health check is executed once in half a minute, and `GetValidator` is just obtaining a read lock and returning a small struct. I think the extra information is worth it.

network/network.go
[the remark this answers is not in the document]
yacovm: > Why would a global ingress connection counter would be preferable over a network-specific connection counter ?

> I'm asking since it would prevent us from creating multiple networks on a single process and have them logically encapsulated.

Why is it global? the counter is created inside the constructor of a network instance and is saved inside the shared config for all peer instances the network creates.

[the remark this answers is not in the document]
yacovm: [Changed to uint64](https://github.com/ava-labs/avalanchego/pull/3719/commits/f476513314ad5c85650e1558b5586ef4f80f5d48#diff-96a7f5c352cd3d2d1ee3c7c55ed95e668c6edb28a4fa337e6c60270471bdab61L272)

network/network.go
```diff
@@ -48,6 +48,14 @@ const (
 	SendFailRateKey             = "sendFailRate"
 )
 
+type connectionType uint8
+
+const (
+	connectionTypeUnknown connectionType = iota
+	connectionTypeIngress
```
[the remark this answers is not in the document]
yacovm: [Removed the connType type](https://github.com/ava-labs/avalanchego/pull/3719/commits/f476513314ad5c85650e1558b5586ef4f80f5d48#diff-96a7f5c352cd3d2d1ee3c7c55ed95e668c6edb28a4fa337e6c60270471bdab61L51-L58)

network/network_test.go
yacovm: [Moved the waitgroup](https://github.com/ava-labs/avalanchego/pull/3719/commits/f476513314ad5c85650e1558b5586ef4f80f5d48#diff-6534cc6f135999ac0a3c726bf471e812670d52b30707d8facd54fe5a61f9f58eL340)

tsachiherman:
Thanks. Will re-review shortly.

tsachiherman (approved):

node/node.go
```diff
@@ -1534,6 +1543,17 @@ func (n *Node) initHealthAPI() error {
 	)
 }
 
+func (n *Node) initIngressConnectionCountHealthCheck() error {
```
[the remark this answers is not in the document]
yacovm: I can't, it depends on both chain manager and health API, so I have to initialize it after both.
The chain manager is initialized after the health API is...

node/no_ingress_conn_alert.go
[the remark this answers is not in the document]
yacovm: A handshake can be both initiated by you and by someone else, no? So if you connect to someone else, that someone else also handshakes with you, I think.

node/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	now := nica.now()
+
+	if nica.lastCheck.IsZero() {
+		nica.lastCheck = now
+		return nica.lastResult, nica.lastError
+	}
+
+	if now.Sub(nica.lastCheck) < nica.minCheckInterval && nica.lastResult == nil {
```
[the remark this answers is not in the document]
yacovm: Exactly. We don't want to alert right away, only after some initial period of time. This is in place to prevent unforeseen false positives.

node/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,94 @@
...
+	minCheckInterval   time.Duration
+	now                func() time.Time
+	selfID             ids.NodeID
+	ingressConnections ingressConnectionCounter
+	validators         validatorRetriever
+	bootstrapping      bootstrappingTracker
+}
+
+func ingressConnResult(n int, areWeValidator bool) map[string]interface{} {
+	return map[string]interface{}{"ingressConnectionCount": n, "primary network validator": areWeValidator}
```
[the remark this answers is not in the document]
yacovm: Can you be more specific? which other health check fields?

node/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (nica *noIngressConnAlert) HealthCheck(context.Context) (interface{}, error) {
```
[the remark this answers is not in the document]
yacovm: That's not possible because I need to save state from invocation to invocation.
Specifically I need to save the last result and error, as well as the time of the last check.

config/flags.go
```diff
@@ -176,7 +176,7 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.Duration(NetworkReadHandshakeTimeoutKey, constants.DefaultNetworkReadHandshakeTimeout, "Timeout value for reading handshake messages")
 	fs.Duration(NetworkPingTimeoutKey, constants.DefaultPingPongTimeout, "Timeout value for Ping-Pong with a peer")
 	fs.Duration(NetworkPingFrequencyKey, constants.DefaultPingFrequency, "Frequency of pinging other peers")
-
+	fs.Duration(NoIngressValidatorConnectionTimeoutKey, constants.DefaultNoIngressValidatorConnectionTimeout, "Time after which nodes are expected to be connected to us if we are a primary network validator")
```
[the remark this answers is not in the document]
yacovm: Fixed.

node/no_ingress_conn_alert_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/validators"
+)
+
+type mockBootstrappingTracker struct {
```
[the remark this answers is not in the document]
yacovm: Fixed

node/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if len(nica.bootstrapping.Bootstrapping()) > 0 {
+		return nil, nil
+	}
+
+	now := nica.now()
+
+	if nica.lastCheck.IsZero() {
+		nica.lastCheck = now
+		return nica.lastResult, nica.lastError
```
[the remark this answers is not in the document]
yacovm: Fixed

node/no_ingress_conn_alert_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (m *mockIngressConnectionCounter) IngressConnCount() int {
+	return m.res
+}
+
+func TestNoIngressConnAlertHealthCheck(t *testing.T) {
+	start := time.Now()
+
+	for _, testCase := range []struct {
```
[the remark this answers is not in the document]
yacovm: Fixed

network/network.go
```diff
@@ -256,30 +259,34 @@ func NewNetwork(
 		ipTracker.ManuallyTrack(nodeID)
 	}
 
+	// ingressConnCount tracks connections to all remote peers,
+	// so for safety we set it as pointer, in case the config object is copied.
```
StephenButtolph, line 263: > for safety we set it as pointer, in case the config object is copied.

The config already includes atomic values (so it already isn't safe to copy). Also, `atomic.Uint64` includes `noCopy` - which means that CI will fail if it is ever copied... Seems like it would be cleaner to just put it in the config

network/network.go
```diff
@@ -256,30 +259,34 @@ func NewNetwork(
...
+	var ingressConnCount atomic.Uint64
```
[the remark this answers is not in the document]
yacovm: I'm not sure I follow, why is adding 11111...11 to subtract 1 a problem? A connection count is something that is non negative so I think it makes sense to use an unsigned integer.

network/network.go
yacovm: I did it because it's less work for me to change the tests :) 

```
network/peer/peer_test.go:180:32: copylocks: call of newRawTestPeer copies lock value: github.com/ava-labs/avalanchego/network/peer.Config contains sync/atomic.Uint64 contains sync/atomic.noCopy (govet)
```

yacovm: Also I don't want to rely on CI for something like this...

yacovm:
Parking this as a draft before I manually test it again after this code change

config/keys.go
```diff
@@ -190,6 +190,7 @@ const (
 	SystemTrackerDiskHalflifeKey                       = "system-tracker-disk-halflife"
 	SystemTrackerRequiredAvailableDiskSpaceKey         = "system-tracker-disk-required-available-space"
 	SystemTrackerWarningThresholdAvailableDiskSpaceKey = "system-tracker-disk-warning-threshold-available-space"
+	NoIngressValidatorConnectionTimeoutKey             = "no-ingress-validator-connection-timeout"
```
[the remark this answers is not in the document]
yacovm: [Moved](https://github.com/ava-labs/avalanchego/pull/3719/commits/1cf5056e7b6460f0127492102254d8fb5e9ae0b0#diff-673d1b52af0b61d77a9bef15f54ccd5fcda4554a96356d6cd4fc7a67f70e36ccR123) it to the network group of flags.

network/network.go
yacovm: [Made](https://github.com/ava-labs/avalanchego/pull/3719/commits/1cf5056e7b6460f0127492102254d8fb5e9ae0b0#diff-635834cc4e56923e40bf037154f3c89d2d64e8cafe4286e3f086356b5eb2525fR29-R32) it a non pointer

node/no_ingress_conn_alert.go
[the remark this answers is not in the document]
yacovm: [Simplified](https://github.com/ava-labs/avalanchego/pull/3719/commits/1cf5056e7b6460f0127492102254d8fb5e9ae0b0#diff-a717b2d70449f71cde25c97e15008b7a6f8b21c9be34bb89c16d77a871546b20R31-R47) the check and made it to be part of the network health check.

yacovm:
@StephenButtolph the PR is ready again for re-review.

network/network.go
```diff
@@ -427,9 +448,18 @@ func (n *network) HealthCheck(context.Context) (interface{}, error) {
 	if !isMsgFailRate {
 		errorReasons = append(errorReasons, fmt.Sprintf("messages failure send rate %g > %g", sendFailRate, n.config.HealthConfig.MaxSendFailRate))
 	}
+
+	if !reachablePrimaryNetworkValidator {
+		errorReasons = append(errorReasons, "primary network validator is unreachable")
```
StephenButtolph, line 453: I think this is the issue we are trying to catch, but I think the error should be more specific (to make potentially false positives easier to diagnose as false positives)
```suggestion
		errorReasons = append(errorReasons, "primary network validator has no inbound connections")
```

network/network.go
```diff
@@ -401,6 +413,15 @@ func (n *network) HealthCheck(context.Context) (interface{}, error) {
 	details[SendFailRateKey] = sendFailRate
 	n.metrics.sendFailRate.Set(sendFailRate)
 
+	reachablePrimaryNetworkValidator := true
+	// Make sure if we're a primary network validator, we have ingress connections
```
StephenButtolph, line 417:
```suggestion
	// If we're a primary network validator, make sure we have ingress connections
```

network/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if connCount > 0 {
+		return ingressConnResult(connCount, areWeValidator), nil
+	}
+
+	if !areWeValidator {
+		return ingressConnResult(connCount, areWeValidator), nil
+	}
+
+	return ingressConnResult(connCount, areWeValidator), ErrNoIngressConnections
```
StephenButtolph, line 47: I feel like `ingressConnResult` is not a useful abstraction here:
```suggestion
	result := map[string]interface{}{
		"ingressConnectionCount":    connCount,
		"primary network validator": areWeValidator,
	}
	if connCount > 0 || !areWeValidator {
		return result, nil
	}
	return result, ErrNoIngressConnections
```

network/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type noIngressConnAlert struct {
+	selfID             ids.NodeID
+	ingressConnections ingressConnectionCounter
+	validators         validatorRetriever
+}
+
+func ingressConnResult(n int, areWeValidator bool) map[string]interface{} {
+	return map[string]interface{}{"ingressConnectionCount": n, "primary network validator": areWeValidator}
```
StephenButtolph, line 32: It feels odd that "one of these has spaces" and the other is "camelCase".

network/no_ingress_conn_alert.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (nica *noIngressConnAlert) checkHealth() (interface{}, error) {
```
StephenButtolph, line 35: Would it make more sense for this to just be a function that takes in:
```
	selfID             ids.NodeID
	ingressConnections ingressConnectionCounter
	validators         validatorRetriever
```
rather than defining the whole `noIngressConnAlert` struct?

network/network.go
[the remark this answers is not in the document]
StephenButtolph: It isn't a correctness problem - its a readability problem. `x.Add(^0)` is much less readable than `x.Add(-1)` imo

network/network_test.go
```diff
@@ -306,6 +311,50 @@ func TestNewNetwork(t *testing.T) {
...
+		_, err := net.HealthCheck(context.Background())
+		healthCheckErrors = append(healthCheckErrors, err)
+	}
+
+	// First node has all nodes connected to it.
+	// Second node has only the third node connected to it.
+	// Third node has no node connected to it, as it connects to the first and second node.
+	require.Equal([]int{2, 1, 0}, ingressConnections)
+	require.Equal([]error{nil, nil}, healthCheckErrors[:2])
+	require.EqualError(healthCheckErrors[2], "network layer is unhealthy reason: primary network validator is unreachable")
```
StephenButtolph, line 355: Can we add a TODO here to remove this string check? I'd love to not have this - but it seems like a bigger refactor than should be included in this PR.

config/node/config.go
```diff
@@ -196,6 +196,10 @@ type Config struct {
 	// Larger halflife --> disk usage metrics change more slowly.
 	SystemTrackerDiskHalflife time.Duration `json:"systemTrackerDiskHalflife"`
 
+	// NoIngressValidatorConnectionTimeout denotes the time after which at least one
+	// node is expected to be connected to us if we validate the primary network, otherwise a health check fails.
+	NoIngressValidatorConnectionTimeout time.Duration `json:"noIngressValidatorConnectionTimeout"`
+
```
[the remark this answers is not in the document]
yacovm: Thanks, missed it. Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17L1441).

network/network.go
yacovm: I thought unreachable implies just that, that it's not reachable, but doesn't say anything about whether it can reach other nodes. 

Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-96a7f5c352cd3d2d1ee3c7c55ed95e668c6edb28a4fa337e6c60270471bdab61L453).

network/network.go
yacovm: Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-96a7f5c352cd3d2d1ee3c7c55ed95e668c6edb28a4fa337e6c60270471bdab61L417).

network/no_ingress_conn_alert.go
yacovm: Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-a717b2d70449f71cde25c97e15008b7a6f8b21c9be34bb89c16d77a871546b20L31-L48).

network/no_ingress_conn_alert.go
yacovm: Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-fd96d1e5d5ab19a672f4a33d6ab4a3698cca6103c509465354b2f07377f0dcc1L47).

network/no_ingress_conn_alert.go
yacovm: yes, in the previous incarnations this struct had state, so it had to be one. Now it can be a function as you pointed out. Fixed [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-a717b2d70449f71cde25c97e15008b7a6f8b21c9be34bb89c16d77a871546b20R25).

network/network.go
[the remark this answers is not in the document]
yacovm: [Changed](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-47c50dc44ebcefe91fa948a05bacec6e10a2f3edb0939bae4ef5a10ebbada254R64) to int64.

network/network_test.go
yacovm: Hopefully I improved it by referencing a variable instead, [here](https://github.com/ava-labs/avalanchego/pull/3719/commits/9c9ca56e1ad115a5e5ba025a4df288e978e4ec12#diff-6534cc6f135999ac0a3c726bf471e812670d52b30707d8facd54fe5a61f9f58eL355).

utils/constants/networking.go
```diff
@@ -38,13 +38,14 @@ const (
...
-	DefaultNetworkTimeoutCoefficient    = 2
-	DefaultNetworkReadHandshakeTimeout  = 15 * time.Second
+	DefaultNetworkInitialTimeout                   = 5 * time.Second
+	DefaultNetworkMinimumTimeout                   = 2 * time.Second
+	DefaultNetworkMaximumTimeout                   = 10 * time.Second
+	DefaultNetworkMaximumInboundTimeout            = 10 * time.Second
+	DefaultNetworkTimeoutHalflife                  = 5 * time.Minute
+	DefaultNetworkTimeoutCoefficient               = 2
+	DefaultNetworkReadHandshakeTimeout             = 15 * time.Second
+	DefaultNoIngressValidatorConnectionGracePeriod = time.Minute * 10
```
StephenButtolph, line 48: Just for consistency
```suggestion
	DefaultNoIngressValidatorConnectionGracePeriod = 10 * time.Minute
```

yacovm: [Fixed](https://github.com/ava-labs/avalanchego/pull/3719/commits/23850fd3a9b753dd73bcf7e473bdf8fb5bfe5d9e#diff-0f86939a30a343f4c6e9de6b3efcbdbee14db63808ee96a1a1d6669cb7dab71aR48).

StephenButtolph (approved):

network/network.go
StephenButtolph: The reason I prefer "has no inbound connections" is because it doesn't assume that no inbound connections implies unreachable. (For example, maybe the node is reachable.. But the public IP isn't set correctly. So the node could be reachable - just not at the broadcasted public IP).

PR #3725, opened 2025-02-18
RPC-signer configuration

richardpringle:
## Why this should be merged

The enables the rpc-signer configuration

## How this works

If the proper flag is set, a gprc-client is instantiated and used for signing. The `node` will fail to start if there isn't a grpc-server implmentation at supplied connection-string.

## How this was tested

A table test was added in `config/config_test.go` that tests all possible signing configurations. This only tests the instantiation of the signer as the actual signing tests live in the respective signer packages. 

Maybe? Do we have some kind of experimental flag? I believe we should have some actual experience running live nodes with an rpc-signer before we really document the usage here.

### DIFF
--- a/utils/crypto/bls/signer/rpcsigner/client.go
+++ b/utils/crypto/bls/signer/rpcsigner/client.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
@@ -28,6 +29,8 @@ func NewClient(ctx context.Context, url string) (*Client, func() error, error) {
 	// TODO: figure out the best parameters here given the target block-time
 	opts := grpc.WithConnectParams(grpc.ConnectParams{
 		Backoff: backoff.DefaultConfig,
+		// same as grpc default
+		MinConnectTimeout: 20 * time.Second,
 	})
 
 	// the rpc-signer client should call a proxy server (on the same machine) that forwards


config/config.go
```diff
@@ -639,47 +644,85 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
-		key, err := localsigner.New()
+func getStakingSigner(ctx context.Context, v *viper.Viper) (bls.Signer, error) {
+	ephemeralSignerEnabled := v.GetBool(StakingEphemeralSignerEnabledKey)
+	contentKeyExists := v.IsSet(StakingSignerKeyContentKey)
+	keyPathIsSet := v.IsSet(StakingSignerKeyPathKey)
+	rpcSignerURLExists := v.IsSet(StakingRPCSignerKey)
+
+	signingKeyPath := getExpandedArg(v, StakingSignerKeyPathKey)
+	_, err := os.Stat(signingKeyPath)
+	keyFileNotFound := errors.Is(err, fs.ErrNotExist)
```
joshua-kim, line 655: I think this looks better + reduces the diff if we just keep this within the `case` statement it's used in (+ we don't need to name the path not existing error).

config/config.go
```diff
@@ -639,47 +644,85 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
-		return key, nil
-	}
 
-	if v.IsSet(StakingSignerKeyContentKey) {
+		return signer, nil
+
+	case !ephemeralSignerEnabled && contentKeyExists && !keyPathIsSet && !rpcSignerURLExists:
 		signerKeyRawContent := v.GetString(StakingSignerKeyContentKey)
 		signerKeyContent, err := base64.StdEncoding.DecodeString(signerKeyRawContent)
+
```
joshua-kim, line 670: nit: unneeded diff, generally errors are checked on the following line without an extra line break

config/config.go
```diff
@@ -639,47 +644,85 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+		// the rpc-signer client should call a proxy server (on the same machine) that forwards
+		// the request to the actual signer instead of relying on tls-credentials
+		conn, err := grpc.NewClient(rpcSignerURL, grpc.WithTransportCredentials(insecure.NewCredentials()))
+		if err != nil {
+			return nil, fmt.Errorf("couldn't create rpc signer client: %w", err)
+		}
+
+		signer, err := rpcsigner.NewClient(ctx, conn)
+		if err != nil {
+			conn.Close()
```
joshua-kim, line 713: I think the way we manage the lifecycle of this conn is supposed to live somewhere else (upstream of this code). I think we need to hold a reference to this conn and upon a fatal error and call `Close` to prevent the server from relying on a timeout to terminate the connection

config/config.go
```diff
@@ -639,47 +644,85 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+		}
+
+		return signer, nil
+
+	case ephemeralSignerEnabled || contentKeyExists || keyPathIsSet || rpcSignerURLExists:
+		return nil, errInvalidSignerConfig
 	}
 
-	key, err := localsigner.New()
+	// no config values explicitly set,
```
joshua-kim, line 723: This is the existing code, but I think we should just fix it here. If the usage of the ephemeral signer is the default, should this be the default case in the switch + merged w/ the ephemeral signer flag is specified case?

config/config_test.go
```diff
@@ -541,6 +550,106 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+		require.NoError(t, err)
+	}()
+
+	type config map[string]any
+	type test map[string]struct {
+		viperKeys          string
+		config             config
+		expectedSignerType reflect.Type
+		expectedErr        error
+	}
```
joshua-kim, line 591: nit: you can define these anonymously like so (and also avoid the map-style initialization that is inconsistent w/ the way we have most of our tests written)
````
	tests := []struct{
		name string
		foo     bar
	} {
		{
			name: "foo",
			foo:  bar{},
		},
	}

	for _, tt := range tests {
      // test here
	}
````

File: config/config_test.go [hunk 6]
```diff
@@ -541,6 +550,106 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+			config: config{
+				StakingEphemeralSignerEnabledKey: true,
+				StakingSignerKeyContentKey:       testKey,
+			},
+			expectedErr: errInvalidSignerConfig,
+		},
+	}
+
+	// required for proper write permissions for the default signer-key location
+	t.Setenv("HOME", t.TempDir())
```
joshua-kim, line 631: Should we have a defer step to clean up stuff? Or is this overkill?

config/flags.go
```diff
@@ -271,6 +271,7 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.Bool(StakingEphemeralSignerEnabledKey, false, "If true, the node uses an ephemeral staking signer key")
 	fs.String(StakingSignerKeyPathKey, defaultStakingSignerKeyPath, fmt.Sprintf("Path to the signer private key for staking. Ignored if %s is specified", StakingSignerKeyContentKey))
 	fs.String(StakingSignerKeyContentKey, "", "Specifies base64 encoded signer private key for staking")
+	fs.String(StakingRPCSignerKey, "", "Specifies the gRPC endpoint of the staking signer")
```
joshua-kim, line 274: Can we be consistent w/ `rpc/grpc` naming here

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -21,7 +24,24 @@ type Client struct {
 	pk     *bls.PublicKey
 }
 
-func NewClient(ctx context.Context, conn *grpc.ClientConn) (*Client, error) {
+func NewClient(ctx context.Context, rpcSignerURL string) (*Client, error) {
+	// TODO: figure out the best parameters here given the target block-time
+	opts := grpc.WithConnectParams(grpc.ConnectParams{
+		Backoff: backoff.DefaultConfig,
+	})
```
joshua-kim, line 31: Can we let the caller doesn't provide these options? Or was there a reason we chose to not pass this in as a parameter

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -21,7 +24,24 @@ type Client struct {
...
+
+	// the rpc-signer client should call a proxy server (on the same machine) that forwards
+	// the request to the actual signer instead of relying on tls-credentials
+	conn, err := grpc.NewClient(rpcSignerURL, opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, fmt.Errorf("couldn't create rpc signer client: %w", err)
```
joshua-kim, line 37: nit: I generally avoid special characters where possible in logs errors since you usually have to escape them when grepping through logs. Can we make this something like `could not` or `failed to` or anything along those lines? Our old code doesn't follow this pattern but I think we should avoid it in new code we write

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -46,22 +66,58 @@ func (c *Client) PublicKey() *bls.PublicKey {
 	return c.pk
 }
 
+// Sign a message. The [Client] already handles transient connection errors. If this method fails, it will
+// render the client in an unusable state and the client should be discarded.
 func (c *Client) Sign(message []byte) (*bls.Signature, error) {
+	var err error
+	defer func() {
+		if err != nil {
+			c.Close()
```
joshua-kim, line 75: Another idea I'm wondering is that if we can do better than the current ux of forcing the caller to re-create the client when this errors and manage that within the `Client` type. What if we started a goroutine in `New` that would keep trying to re-connect if we closed the connection, and `Sign*` if we're not connected could just fast-fail and return some "disconnected error"?

config/config_test.go
```diff
@@ -541,6 +550,106 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			require := require.New(t)
+			v := setupViperFlags()
+
+			for key, value := range test.config {
+				v.Set(key, value)
+			}
+
+			signer, err := getStakingSigner(context.Background(), v)
```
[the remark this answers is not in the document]
joshua-kim: This is testing code not exported by the actual package though (i.e, `getStakingSigner` is an implementation detail of the package and not part of its public API) ... shouldn't we test this through `GetNodeConfig` instead?

config/config.go
```diff
@@ -639,73 +642,103 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
 
-func getStakingSigner(v *viper.Viper) (bls.Signer, error) {
-	if v.GetBool(StakingEphemeralSignerEnabledKey) {
-		key, err := localsigner.New()
+func getStakingSigner(ctx context.Context, v *viper.Viper) (bls.Signer, error) {
+	ephemeralSignerEnabled := v.GetBool(StakingEphemeralSignerEnabledKey)
+	contentKeyIsSet := v.IsSet(StakingSignerKeyContentKey)
+	keyPathIsSet := v.IsSet(StakingSignerKeyPathKey)
+	rpcSignerURLIsSet := v.IsSet(StakingRPCSignerKey)
+
```
StephenButtolph, line 650: Would it make sense to validate only one of these is set prior to the switch? The boolean checks are pretty annoying:
```golang
bools := bag.Of(ephemeralSignerEnabled, contentKeyIsSet, keyPathIsSet, rpcSignerURLIsSet)
if bools.Count(true) > 1 {
	return nil, errInvalidSignerConfig
}
```

config/config.go
```diff
@@ -639,73 +642,103 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+		keyBytes := signer.ToBytes()
+		if err := os.WriteFile(signingKeyPath, keyBytes, perms.ReadWrite); err != nil {
+			return nil, fmt.Errorf("couldn't write new signing key to %s: %w", signingKeyPath, err)
+		}
+
+		if err := os.Chmod(signingKeyPath, perms.ReadOnly); err != nil {
+			return nil, fmt.Errorf("couldn't restrict permissions on new signing key at %s: %w", signingKeyPath, err)
+		}
+
+		return signer, nil
```
StephenButtolph, line 729: This isn't the same (default) behavior as before. Previously the behavior was:

> If the staking information is present, use it. Otherwise, generate new staking information, save it, and use it.

Running a node twice with this code causes a FATAL (because the file isn't writeable after the permissions are restricted)

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -21,7 +24,24 @@ type Client struct {
...
+	})
+
+	// the rpc-signer client should call a proxy server (on the same machine) that forwards
+	// the request to the actual signer instead of relying on tls-credentials
+	conn, err := grpc.NewClient(url, opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create rpc signer client: %w", err)
+	}
+	defer func() {
+		if err != nil {
```
ARR4N, line 40: The typical way to clarify this is to change the function signature to:

```go
func NewClient(ctx context.Context, url string) (_ *Client, retErr error) {
```

and then here

```go
defer func() {
    if retErr != nil {
        conn.Close()
    }
}()
```

You'll never explicitly set `retErr` so it can't be accidentally overwritten and is known to be the returned error.

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -21,7 +24,24 @@ type Client struct {
...
+	// the request to the actual signer instead of relying on tls-credentials
+	conn, err := grpc.NewClient(rpcSignerURL, opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, fmt.Errorf("couldn't create rpc signer client: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			conn.Close()
+		}
+	}()
```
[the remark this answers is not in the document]
ARR4N: The typical way to clarify this is to change the function signature to:

```go
func NewClient(ctx context.Context, url string) (_ *Client, retErr error) {
```

and then here

```go
defer func() {
    if retErr != nil {
        conn.Close()
    }
}()
```

You'll never explicitly set `retErr` so it can't be accidentally overwritten and is known to be the returned error.

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -46,22 +66,58 @@ func (c *Client) PublicKey() *bls.PublicKey {
...
+// Sign a message. The [Client] already handles transient connection errors. If this method fails, it will
+// render the client in an unusable state and the client should be discarded.
 func (c *Client) Sign(message []byte) (*bls.Signature, error) {
+	var err error
+	defer func() {
+		if err != nil {
+			c.Close()
+		}
+	}()
+
```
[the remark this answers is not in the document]
ARR4N: Options as I see them:

1. You add a `Close() error` method to the `bls.Signer` interface. This is clean but pollutes it and forces all other concrete types to implement it (as a no-op).
2. `getStakingSigner()` returns a `cleanup` function as a middle parameter. That pollutes the call stack as it has to be bubbled up to the point in the code that knows when the signer is no longer needed.
3. Use `runtime.SetFinalizer()` in `NewClient()` and then change it to `runtime.AddCleanup()` once using Go 1.24 because `SetFinalizer()` has many foot guns.

(1) feels the most idiomatic to me but if there's a strong preference against polluting the `bls.Signer` signature then I'm not entirely against (3) with a pattern like:

```go
c := &Client{
    client: client,
    conn:   conn,
    pk:     pk,
}
runtime.SetFinalizer(c, func(*Client){
  conn.Close() // safe to drop the error because operations are read-only
})
return c, nil
```

Note that the finalizer may never actually be called (why this is a bad idea for non-read-only connections) and that limiting access to the `conn` to inside `Client` avoids [this foot gun](https://pkg.go.dev/runtime#SetFinalizer:~:text=In%20order%20to%20use%20finalizers%20correctly%2C%20the%20program%20must%20ensure%20that%20the%20object%20is%20reachable%20until%20it%20is%20no%20longer%20required.).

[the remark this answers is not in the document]
StephenButtolph: I guess I'm still confused.

As the code is currently written, we `Close` the `conn` if a request to the `client` fails. What is the _goal_ of this code?

If the code is attempting to cleanup the connection, it doesn't really seem to do that (because there doesn't seem to be any guarantee that the client will be called and have an error returned during shutdown).

If we are trying to ensure that the connection is closed when we exit, it feels like it should be the job of the caller `NewClient` to close it when it is done using it.

I feel like @ARR4N's 2nd suggestion (or a modification of it to be somewhere between 1 and 2) aligns with what we'd actually want to do the best.

Please don't use `runtime.SetFinalizer` or `runtime.SetFinalizer` haha.

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
+	// TODO: figure out the best parameters here given the target block-time
+	opts := grpc.WithConnectParams(grpc.ConnectParams{
+		Backoff: backoff.DefaultConfig,
+	})
+
+	// the rpc-signer client should call a proxy server (on the same machine) that forwards
+	// the request to the actual signer instead of relying on tls-credentials
+	conn, err := grpc.NewClient(url, opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, func() error { return nil }, fmt.Errorf("failed to create rpc signer client: %w", err)
```
StephenButtolph, line 37:
```suggestion
		return nil, nil, fmt.Errorf("failed to create rpc signer client: %w", err)
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
+	}
+
+	cleanup := func() error {
+		return conn.Close()
+	}
+
```
StephenButtolph, line 43:
```suggestion
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
 	client := pb.NewSignerClient(conn)
 
 	pubkeyResponse, err := client.PublicKey(ctx, &pb.PublicKeyRequest{})
 	if err != nil {
-		return nil, err
+		return nil, nil, errors.Join(err, cleanup())
```
StephenButtolph, line 48:
```suggestion
		return nil, nil, errors.Join(err, conn.Close())
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
 	}
 
 	pkBytes := pubkeyResponse.GetPublicKey()
 	pk, err := bls.PublicKeyFromCompressedBytes(pkBytes)
 	if err != nil {
-		return nil, err
+		return nil, nil, errors.Join(err, cleanup())
```
StephenButtolph, line 54:
```suggestion
		return nil, nil, errors.Join(err, conn.Close())
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
 	}
 
 	return &Client{
 		client: client,
-		conn:   conn,
 		pk:     pk,
-	}, nil
+	}, cleanup, nil
```
StephenButtolph, line 60:
```suggestion
	}, conn.Close, nil
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
 }
 
 func (c *Client) PublicKey() *bls.PublicKey {
 	return c.pk
 }
 
+// Sign a message. The [Client] already handles transient connection errors. If this method fails, it will
+// render the client in an unusable state and the client should be discarded.
```
StephenButtolph, line 68:
```suggestion
// Sign a message. The [Client] already handles transient connection errors.
```

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -17,51 +21,69 @@ var _ bls.Signer = (*Client)(nil)
...
-	conn   *grpc.ClientConn
 	pk     *bls.PublicKey
 }
 
-func NewClient(ctx context.Context, conn *grpc.ClientConn) (*Client, error) {
+func NewClient(ctx context.Context, url string) (*Client, func() error, error) {
+	// TODO: figure out the best parameters here given the target block-time
+	opts := grpc.WithConnectParams(grpc.ConnectParams{
+		Backoff: backoff.DefaultConfig,
+	})
```
StephenButtolph, line 31: Can we better describe what this is doing? On a quick look - it doesn't seem to do anything. But on closer inspection, it seems to be lowering the `MinConnectTimeout` from `20s` to `0`. Can we explicitly set the `MinConnectTimeout: 0` in the `ConnectParams` and comment that this is why we are specifying the params? It is fairly common for unspecified (0) to be overridden to the default value... But that doesn't seem to happen here.

Alternatively, if the goal wasn't to modify the `MinConnectTimeout` - should we delete this option?

main/main.go
```diff
@@ -51,7 +53,9 @@ func main() {
 		os.Exit(0)
 	}
 
-	nodeConfig, err := config.GetNodeConfig(v)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	nodeConfig, cleanup, err := config.GetNodeConfig(ctx, v)
```
StephenButtolph, line 57: ngl - the more I look at how we pass `cleanup` around, the more I think we need to do a refactor of the `config` package.

The `config` package should just be populating the config... Doing file system stuff and now networking stuff is just terrible... That being said.. It feels a bit unreasonable to expect a total refactor of node initialization just to add this one feature in... So I think this is fine for now... But we really need to clean this up because it's getting unmanageable.

config/config_test.go
```diff
@@ -549,6 +557,130 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+				}
+			}()
+
+			require.ErrorIs(err, tt.expectedErr)
+			require.IsType(tt.expectedSignerType, config.StakingSigningKey)
+		})
+	}
+}
+
+func TestDefaultConfigInitializtionUsesExistingDefaultKey(t *testing.T) {
```
StephenButtolph, line 667:
```suggestion
func TestDefaultConfigInitializationUsesExistingDefaultKey(t *testing.T) {
```

config/config.go
```diff
@@ -640,91 +644,150 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+	}
+
+	cleanup := func() error { return nil }
+
+	switch {
+	case ephemeralSignerEnabled:
+		signer, err := localsigner.New()
 		if err != nil {
-			return nil, fmt.Errorf("couldn't generate ephemeral signing key: %w", err)
+			return nil, nil, fmt.Errorf("couldn't generate ephemeral signing signer: %w", err)
```
StephenButtolph, line 666:
```suggestion
			return nil, nil, fmt.Errorf("couldn't generate ephemeral signer: %w", err)
```

config/config.go
```diff
@@ -640,91 +644,150 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+		return signer, cleanup, nil
+
+	default:
+		_, err := os.Stat(signingKeyPath)
+
+		// if the file exists, try to load the key from the file
+		if !errors.Is(err, fs.ErrNotExist) {
+			signer, err := createSignerFromFile(signingKeyPath)
+			if err != nil {
+				return nil, nil, errors.Join(err, cleanup())
```
StephenButtolph, line 719: I don't think cleanup is doing anything here.
```suggestion
				return nil, nil, err
```

config/config.go
```diff
@@ -640,91 +644,150 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+
+	case keyPathIsSet:
+		// If the key is set, but a user-file isn't provided, we don't create one.
+		// The siging key is only stored to the default file-location if it's created
+		// and saved by the current application run.
+		_, err := os.Stat(signingKeyPath)
+
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, nil, errMissingStakingSigningKeyFile
+		}
```
StephenButtolph, line 693: Why do we check if the file exists here? `createSignerFromFile` seems to handle the case that the file doesn't exist just fine (and this technically introduces some weird races).

config/config.go
```diff
@@ -640,91 +644,150 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
+			return nil, nil, fmt.Errorf("couldn't create rpc signer client: %w", err)
+		}
+
+		return signer, cleanup, nil
+
+	default:
+		_, err := os.Stat(signingKeyPath)
+
+		// if the file exists, try to load the key from the file
+		if !errors.Is(err, fs.ErrNotExist) {
```
StephenButtolph, line 716: Why do we stat the file first? I'd imagine `fs.ErrNotExist` is what gets returned from `os.ReadFile` if the file doesn't exist right?

StephenButtolph (commented):
I think this is about as clean as is reasonable... LGTM once the comments are addressed. Although I'm definitely not super happy with the `config` package as a whole.

PR #3729, opened 2025-02-19
Add With and WithOptions receivers to the Logger interface

iansuvak:
## Why this should be merged
Being able to use hooks in `WithOptions` to test logging, and `With` to create child loggers with particular contexts is useful. 
## How this works
Creates a clone of the log struct  with the internal loggers replaced with returns of zap methods but keeping the same wrapped cores to preserve existing functionality. 

Originally included `WithLazy` as well but removed it since I didn't see a good way of testing that it's actually doing lazy modification properly with our setup being behind the interface.

## How this was tested
Existing tests should pass and new unit tests added. They are not vectorized since they only have two cases but happy to change it if that's the preference. 

No

### DIFF
--- a/utils/logging/log_test.go
+++ b/utils/logging/log_test.go
@@ -20,7 +20,7 @@ func (w *testLogWriter) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
-func (w *testLogWriter) Close() error {
+func (*testLogWriter) Close() error {
 	return nil
 }
 
@@ -59,7 +59,7 @@ func TestWithOption(t *testing.T) {
 
 	hookTriggered := new(bool)
 
-	hook := func(entry zapcore.Entry) error {
+	hook := func(_ zapcore.Entry) error {
 		*hookTriggered = true
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


utils/logging/log_test.go
```diff
@@ -23,3 +38,39 @@ func TestLog(t *testing.T) {
...
+	parentLogger.Info("parent message")
+	require.Len(t, testWriter.logs, 2)
+	require.NotContains(t, string(testWriter.logs[1]), `"bool": true`)
+}
+
+func TestWithOption(t *testing.T) {
+	testWriter := &testLogWriter{}
+	parentLogger := NewLogger("", NewWrappedCore(Info, testWriter, Plain.ConsoleEncoder()))
+
+	hookTriggered := new(bool)
```
yacovm, line 60: you don't need to have a pointer, you can have a regular bool.

iansuvak: Indeed -- updated.

yacovm (approved):

StephenButtolph (approved):

PR #3730, opened 2025-02-20
Libevm avalanchegov.1.12 bump

ceyonur:


StephenButtolph:
I don't think this is needed anymore.

PR #3731, opened 2025-02-20
[nix] Enable binary and image build

maru-ava:


maru-ava:
Obviated by the planned adoption of bazel

PR #3732, opened 2025-02-20
Migrate X-Chain from LevelDB to Firewood and PebbleDB

### LINKED ISSUE #4493: Merkle-ize X-Chain
Migrate the state on the X-Chain to use Firewood to support a merkle state that can be state synced to.

joshua-kim:
## Why this should be merged

Implements a database migration from the legacy database format into a database format that supports a merkle-ized prefix of state

## How this works

Changes db schema

## How this was tested

Added a migration unit test + tested state migration on Fuji

```
[03-20|16:01:06.597] INFO <X Chain> avm/vm.go:173 VM config initialized {"config": {"network":{"max-validator-set-staleness":60000000000,"target-gossip-size":20480,"push-gossip-percent-stake":0.9,"push-gossip-num-validators":100,"push-gossip-num-peers":0,"push-regossip-num-validators":10,"push-regossip-num-peers":0,"push-gossip-discarded-cache-size":16384,"push-gossip-max-regossip-frequency":30000000000,"push-gossip-frequency":500000000,"pull-gossip-poll-size":1,"pull-gossip-frequency":1500000000,"pull-gossip-throttling-period":10000000000,"pull-gossip-throttling-limit":2,"expected-bloom-filter-elements":8192,"expected-bloom-filter-false-positive-probability":0.01,"max-bloom-filter-false-positive-probability":0.05},"index-transactions":false,"index-allow-incomplete":false,"checksums-enabled":false}}
[03-20|16:01:06.597] INFO <X Chain> avm/vm.go:598 fee asset is established {"alias": "AVAX", "assetID": "U8iRqJoiJm8xZHAacmvYyZVwqQx6uDNtQeP3CQ6fcgQk3JqnK"}
[03-20|16:01:06.597] INFO <X Chain> avm/vm.go:257 address transaction indexing is disabled
[03-20|16:01:06.599] INFO <X Chain> snowman/engine.go:96 initializing consensus engine
[03-20|16:01:06.599] INFO <X Chain> bootstrap/bootstrapper.go:320 starting bootstrap
[03-20|16:01:06.601] INFO <X Chain> bootstrap/bootstrapper.go:600 executing transactions
[03-20|16:01:06.601] INFO <X Chain> queue/jobs.go:223 executed operations {"numExecuted": 0}
[03-20|16:01:06.601] INFO <X Chain> bootstrap/bootstrapper.go:612 executing vertices
[03-20|16:01:06.601] INFO <X Chain> queue/jobs.go:223 executed operations {"numExecuted": 0}
[03-20|16:01:06.603] INFO <X Chain> avm/state_migration.go:143 starting state migration
[03-20|16:04:53.784] INFO <X Chain> avm/state_migration.go:321 migration complete
[03-20|16:04:53.790] INFO <X Chain> proposervm/vm.go:209 initialized proposervm {"state": "after fork", "forkHeight": 1, "lastAcceptedHeight": 32356}
[03-20|16:04:53.790] INFO <X Chain> bootstrap/bootstrapper.go:182 starting bootstrapper {"lastAcceptedID": "8o6VfSzPHhjLXwr7FEZDxuVAwXYVCx6dP7vrwLoW9mwYRPZi4", "lastAcceptedHeight": 32356}
[03-20|16:05:42.318] INFO <X Chain> bootstrap/bootstrapper.go:387 starting to fetch blocks {"numKnownBlocks": 99, "numAcceptedBlocks": 1, "numMissingBlocks": 100}
[03-20|16:05:42.857] INFO <X Chain> bootstrap/storage.go:194 executing blocks {"numToExecute": 45}
[03-20|16:05:43.116] INFO <X Chain> bootstrap/storage.go:186 executed blocks {"numExecuted": 45, "numToExecute": 45, "halted": false, "duration": "258.784792ms"}
```

Yes

### DIFF
--- a/database/memdb/db.go
+++ b/database/memdb/db.go
@@ -5,6 +5,7 @@ package memdb
 
 import (
 	"context"
+	"fmt"
 	"slices"
 	"strings"
 	"sync"
@@ -36,6 +37,21 @@ type Database struct {
 // New returns a map with the Database interface methods implemented.
 func New() *Database {
 	return NewWithSize(DefaultSize)
+}
+
+// Copy returns a Database with the same key-value pairs as db
+func Copy(db *Database) (*Database, error) {
+	db.lock.Lock()
+	defer db.lock.Unlock()
+
+	result := New()
+	for k, v := range db.db {
+		if err := result.Put([]byte(k), v); err != nil {
+			return nil, fmt.Errorf("failed to insert key: %w", err)
+		}
+	}
+
+	return result, nil
 }
 
 // NewWithSize returns a map pre-allocated to the provided size with the
--- a/vms/avm/block/builder/builder_test.go
+++ b/vms/avm/block/builder/builder_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 
+	"github.com/ava-labs/avalanchego/chains/atomic"
 	"github.com/ava-labs/avalanchego/codec"
 	"github.com/ava-labs/avalanchego/codec/codecmock"
 	"github.com/ava-labs/avalanchego/database/memdb"
@@ -291,8 +292,11 @@ func TestBuilderBuildBlock(t *testing.T) {
 				manager.EXPECT().VerifyUniqueInputs(preferredID, gomock.Any()).Return(nil)
 				// Assert created block has one tx, tx1,
 				// and other fields are set correctly.
-				manager.EXPECT().NewBlock(gomock.Any()).DoAndReturn(
-					func(block *block.StandardBlock) snowman.Block {
+				manager.EXPECT().NewBlock(gomock.Any(), gomock.Any()).DoAndReturn(
+					func(
+						block *block.StandardBlock,
+						_ atomic.SharedMemory,
+					) snowman.Block {
 						require.Len(t, block.Transactions, 1)
 						require.Equal(t, tx1, block.Transactions[0])
 						require.Equal(t, preferredHeight+1, block.Height())
@@ -350,8 +354,11 @@ func TestBuilderBuildBlock(t *testing.T) {
 				manager.EXPECT().GetState(preferredID).Return(preferredState, true)
 				manager.EXPECT().VerifyUniqueInputs(preferredID, gomock.Any()).Return(nil)
 				// Assert that the created block has the right timestamp
-				manager.EXPECT().NewBlock(gomock.Any()).DoAndReturn(
-					func(block *block.StandardBlock) snowman.Block {
+				manager.EXPECT().NewBlock(gomock.Any(), gomock.Any()).DoAndReturn(
+					func(
+						block *block.StandardBlock,
+						_ atomic.SharedMemory,
+					) snowman.Block {
 						require.Equal(t, preferredTimestamp.Unix(), block.Timestamp().Unix())
 						return nil
 					},
@@ -422,8 +429,11 @@ func TestBuilderBuildBlock(t *testing.T) {
 				manager.EXPECT().GetState(preferredID).Return(preferredState, true)
 				manager.EXPECT().VerifyUniqueInputs(preferredID, gomock.Any()).Return(nil)
 				// Assert that the created block has the right timestamp
-				manager.EXPECT().NewBlock(gomock.Any()).DoAndReturn(
-					func(block *block.StandardBlock) snowman.Block {
+				manager.EXPECT().NewBlock(gomock.Any(), gomock.Any()).DoAndReturn(
+					func(
+						block *block.StandardBlock,
+						_ atomic.SharedMemory,
+					) snowman.Block {
 						require.Equal(t, now.Unix(), block.Timestamp().Unix())
 						return nil
 					},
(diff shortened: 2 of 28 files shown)


vms/avm/block/builder/builder_test.go
```diff
@@ -514,8 +517,22 @@ func TestBlockBuilderAddLocalTx(t *testing.T) {
 	}
 
 	baseDB := versiondb.New(memdb.New())
+	metadataDB := prefixdb.New([]byte("metadata"), baseDB)
```
joshua-kim, line 520: I was thinking all of non-state would live under this prefix... as an alternative we could just have each db live under their own prefix under the root db instead of sharing a common metadata prefix. I don't see a reason to do/not do one over the other so open to suggestions here

vms/avm/vm.go
```diff
@@ -56,6 +59,13 @@ var (
 	errGenesisAssetMustHaveState = errors.New("genesis asset must have non-empty state")
 
 	_ vertex.LinearizableVMWithEngine = (*VM)(nil)
+
+	// Metadata DB Prefixes
+	dbPrefixMetadata = []byte{0x0}
+	dbPrefixIndexer  = []byte{0x1}
+
+	// State DB Prefixes
+	dbPrefixState = []byte{0x1}
```
joshua-kim, line 68: I was thinking we'd have two prefixes, one for state and one for non-state. As an alternative, we could just have each db live under their own prefix under the root db instead of sharing a common metadata prefix. I don't see a reason to do/not do one over the other so open to suggestions here

vms/components/avax/utxo_state.go
```diff
@@ -251,6 +247,25 @@ func (s *utxoState) UTXOIDs(addr []byte, start ids.ID, limit int) ([]ids.ID, err
...
+func (s *utxoState) UTXOs() iter.Seq2[*UTXO, error] {
+	return func(yield func(*UTXO, error) bool) {
+		itr := s.utxoDB.NewIterator()
+		defer itr.Release()
+
+		for itr.Next() {
+			u := &UTXO{}
+			_, err := s.codec.Unmarshal(itr.Value(), u)
+			if err != nil {
+				return
```
joshua-kim, line 259: todo: error handling

database/memdb/db.go
```diff
@@ -38,6 +39,21 @@ func New() *Database {
 	return NewWithSize(DefaultSize)
 }
 
+// Copy returns a Database with the same key-value pairs as db
+func Copy(db *Database) (*Database, error) {
```
joshua-kim, line 43: We need this utility to be able to copy the memdb between VM instances, since closing a database makes it so that it's unusable afterwards.

vms/avm/state/state.go
```diff
@@ -38,7 +42,7 @@ var (
 	timestampKey     = []byte{0x01}
 	lastAcceptedKey  = []byte{0x02}
```
joshua-kim, line 43: This isn't technically needed as part of this PR, but I would suggest changing these from bytes to plain strings (i.e `"is_initialized"`, `"timestamp"`, etc). I don't think the overhead is an issue + it makes it much easier to read in the debugger.

vms/avm/state/state.go
```diff
@@ -110,46 +114,76 @@ type State interface {
...
  */
-type state struct {
+type PreGState struct {
 	parser block.Parser
 	db     *versiondb.Database
 
 	modifiedUTXOs map[ids.ID]*avax.UTXO // map of modified UTXOID -> *UTXO if the UTXO is nil, it has been removed
-	utxoDB        database.Database
-	utxoState     avax.UTXOState
+	// TODO un-export databases post-migration
```
joshua-kim, line 122: Another idea is to just make the calling code use `NewWithFormat` instead of exporting these dbs

vms/avm/vm_test.go
```diff
@@ -716,3 +721,202 @@ func TestClearForceAcceptedExportTx(t *testing.T) {
...
+	for _, txID := range wantTxs {
+		_, err := vm.GetTx(txID)
+		require.ErrorIs(err, database.ErrNotFound)
+	}
+
+	for _, utxo := range wantUTXOs {
+		_, err := vm.GetUTXO(utxo.InputID())
+		require.ErrorIs(err, database.ErrNotFound)
+	}
+}
```
joshua-kim, line 925: Need to test that indexer data is also migrated correctly

joshua-kim: We're going to consider deprecating the indexers, so we won't have to migrate this data.

vms/avm/state/state.go
```diff
@@ -389,29 +430,34 @@ func (s *state) Commit() error {
...
+}
+
+func (s *State) Close() error {
 	return errors.Join(
-		s.utxoDB.Close(),
-		s.txDB.Close(),
-		s.blockIDDB.Close(),
-		s.blockDB.Close(),
+		s.UTXODB.Close(),
+		s.UTXOIndexDB.Close(),
```
joshua-kim, line 451: Previously the `avax` package but it didn't close this db... so this is a behavior change.

vms/avm/state/state.go
```diff
@@ -402,29 +443,34 @@ func (s *state) Commit() error {
...
 	}
 	return s.db.CommitBatch()
 }
 
-func (s *state) Close() error {
+func (s *State) UTXOs(startingUTXOID ids.ID) iter.Seq2[*avax.UTXO, error] {
+	return s.utxoState.UTXOs(startingUTXOID)
+}
+
+func (s *State) Close() error {
```
rrazvan1: should we wrap each possible error, so we debug easily? otherwise we may get for example **failed to flush on disk**, and we dont have any idea which db did that.

```
errs := []error{}
if err := s.UTXODB.Close(); err != nil {
   errs = append(errs, fmt.Errorf("failed to close UTXO DB: %w", err))
}
........
return errors.Join(errs...)
```

joshua-kim: Yeah I can update this to wrap the individual db errors.

joshua-kim:
Need to come up with a way to break up this diff. Currently I think that we could break this into:
1. Introducing read-only shared memory
2. `firewood` package
3. Diff in `state`

PR #3746, opened 2025-02-25
ProposerVM Epochs POC

cam-schultz:
## Why this should be merged

Implements ProposerVM epochs as described in https://github.com/avalanche-foundation/ACPs/pull/181

## How this works
Adds a field `PChainEpochHeight` to proposervm blocks that remains fixed for the epoch's duration. This height is passed to the innerVM block, rather than the windowed P-Chain height.

## How this was tested

Added unit test for the epoch calculation, but unit/integration tests for the rest of the changed are still TODO

Yes

### DIFF
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -73,7 +73,8 @@ var (
 		CortinaXChainStopVertexID:    ids.Empty,
 		DurangoTime:                  InitiallyActiveTime,
 		EtnaTime:                     InitiallyActiveTime,
-		FUpgradeTime:                 UnscheduledActivationTime,
+		FUpgradeTime:                 InitiallyActiveTime,
+		FUpgradeEpochDuration:        30,
 	}
 
 	ErrInvalidUpgradeTimes = errors.New("invalid upgrade configuration")
@@ -95,6 +96,7 @@ type Config struct {
 	DurangoTime                  time.Time `json:"durangoTime"`
 	EtnaTime                     time.Time `json:"etnaTime"`
 	FUpgradeTime                 time.Time `json:"fUpgradeTime"`
+	FUpgradeEpochDuration        uint64    `json:"fUpgradeEpochDuration"`
 }
 
 func (c *Config) Validate() error {
--- a/vms/proposervm/block.go
+++ b/vms/proposervm/block.go
@@ -59,6 +59,7 @@ type Block interface {
 	buildChild(context.Context) (Block, error)
 
 	pChainHeight(context.Context) (uint64, error)
+	pChainEpochHeight(context.Context) (uint64, error)
 }
 
 type PostForkBlock interface {
@@ -163,7 +164,9 @@ func (p *postForkCommonComponents) Verify(
 	}
 
 	var contextPChainHeight uint64
-	if p.vm.Upgrades.IsEtnaActivated(childTimestamp) {
+	if p.vm.Upgrades.IsFUpgradeActivated(childTimestamp) {
+		contextPChainHeight = child.PChainEpochHeight()
+	} else if p.vm.Upgrades.IsEtnaActivated(childTimestamp) {
 		contextPChainHeight = childPChainHeight
 	} else {
 		contextPChainHeight = parentPChainHeight
@@ -203,6 +206,64 @@ func (p *postForkCommonComponents) buildChild(
 		return nil, err
 	}
 
+	var pChainEpochHeight, pChainEpochHeightParent uint64
+	if p.vm.Upgrades.IsFUpgradeActivated(newTimestamp) {
+		parent, err := p.vm.getBlock(ctx, parentID)
+		if err != nil {
+			p.vm.ctx.Log.Error("unexpected build block failure",
+				zap.String("reason", "failed to fetch parent block"),
+				zap.Stringer("parentID", parentID),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+		parent2, err := p.vm.getBlock(ctx, parent.Parent())
+		if err != nil {
+			p.vm.ctx.Log.Error("unexpected build block failure",
+				zap.String("reason", "failed to fetch second parent block"),
+				zap.Stringer("parentID", parentID),
+				zap.Error(err),
+			)
+			return nil, err
+		}
+
+		if p.vm.GetEpoch(parent.Timestamp()) == p.vm.GetEpoch(parent2.Timestamp()) {
+			pChainEpochHeightParent, err = parent.pChainEpochHeight(ctx)
+			if err != nil {
+				p.vm.ctx.Log.Error("unexpected build block failure",
+					zap.String("reason", "failed to get parent epoch height"),
+					zap.Stringer("parentID", parentID),
+					zap.Error(err),
+				)
+				return nil, err
+			}
+			pChainEpochHeight = pChainEpochHeightParent
+
+			p.vm.ctx.Log.Info(
+				"continuing epoch",
+				zap.Uint64("pChainHeight", pChainHeight),
+				zap.Uint64("pChainEpochHeight", pChainEpochHeight),
+				zap.Uint64("epoch", p.vm.GetEpoch(newTimestamp)),
+			)
+		} else {
+			pChainEpochHeight, err = parent.pChainHeight(ctx)
+			if err != nil {
+				p.vm.ctx.Log.Error("unexpected build block failure",
+					zap.String("reason", "failed to get parent p-chain height"),
+					zap.Stringer("parentID", parentID),
+					zap.Error(err),
+				)
+				return nil, err
+			}
+			p.vm.ctx.Log.Info(
+				"epoch has advanced",
+				zap.Uint64("pChainHeight", pChainHeight),
+				zap.Uint64("pChainEpochHeight", pChainEpochHeight),
+				zap.Uint64("epoch", p.vm.GetEpoch(newTimestamp)),
+			)
+		}
+	}
+
 	var shouldBuildSignedBlock bool
 	if p.vm.Upgrades.IsDurangoActivated(parentTimestamp) {
 		shouldBuildSignedBlock, err = p.shouldBuildSignedBlockPostDurango(
@@ -226,7 +287,9 @@ func (p *postForkCommonComponents) buildChild(
 	}
 
 	var contextPChainHeight uint64
-	if p.vm.Upgrades.IsEtnaActivated(newTimestamp) {
+	if p.vm.Upgrades.IsFUpgradeActivated(newTimestamp) {
+		contextPChainHeight = pChainEpochHeight
+	} else if p.vm.Upgrades.IsEtnaActivated(newTimestamp) {
 		contextPChainHeight = pChainHeight
 	} else {
 		contextPChainHeight = parentPChainHeight
@@ -251,6 +314,7 @@ func (p *postForkCommonComponents) buildChild(
 			parentID,
 			newTimestamp,
 			pChainHeight,
+			pChainEpochHeight,
 			p.vm.StakingCertLeaf,
 			innerBlock.Bytes(),
 			p.vm.ctx.ChainID,
@@ -261,6 +325,7 @@ func (p *postForkCommonComponents) buildChild(
 			parentID,
 			newTimestamp,
 			pChainHeight,
+			pChainEpochHeight,
 			innerBlock.Bytes(),
 		)
 	}
--- a/vms/proposervm/block/build.go
+++ b/vms/proposervm/block/build.go
@@ -18,15 +18,17 @@ func BuildUnsigned(
 	parentID ids.ID,
 	timestamp time.Time,
 	pChainHeight uint64,
+	pChainEpochHeight uint64,
 	blockBytes []byte,
 ) (SignedBlock, error) {
 	var block SignedBlock = &statelessBlock{
 		StatelessBlock: statelessUnsignedBlock{
-			ParentID:     parentID,
-			Timestamp:    timestamp.Unix(),
-			PChainHeight: pChainHeight,
-			Certificate:  nil,
-			Block:        blockBytes,
+			ParentID:          parentID,
+			Timestamp:         timestamp.Unix(),
+			PChainHeight:      pChainHeight,
+			PChainEpochHeight: pChainEpochHeight,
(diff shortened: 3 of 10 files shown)


upgrade/upgrade.go
```diff
@@ -99,6 +100,7 @@ type Config struct {
 	EtnaTime                     time.Time `json:"etnaTime"`
 	FortunaTime                  time.Time `json:"fortunaTime"`
 	GraniteTime                  time.Time `json:"graniteTime"`
+	GraniteEpochDuration         uint64    `json:"graniteEpochDuration"`
```
a1k0n: for consistency, why not make this a time.Time also, and the default be 30 * time.Second? does it absolutely need to be an integer number of seconds?

cam-schultz: Thanks for the suggestion. I've made this change.

upgrade/upgrade.go
```diff
@@ -77,28 +77,30 @@ var (
...
+	ApricotPhase6Time            time.Time     `json:"apricotPhase6Time"`
+	ApricotPhasePost6Time        time.Time     `json:"apricotPhasePost6Time"`
+	BanffTime                    time.Time     `json:"banffTime"`
+	CortinaTime                  time.Time     `json:"cortinaTime"`
+	CortinaXChainStopVertexID    ids.ID        `json:"cortinaXChainStopVertexID"`
+	DurangoTime                  time.Time     `json:"durangoTime"`
+	EtnaTime                     time.Time     `json:"etnaTime"`
+	FortunaTime                  time.Time     `json:"fortunaTime"`
+	GraniteTime                  time.Time     `json:"graniteTime"`
+	GraniteEpochDuration         time.Duration `json:"graniteEpochDuration"`
```
yacovm, line 103: shouldn't we add this to `Mainnet` and `Fuji` too?

yacovm, line 103: shouldn't we add this to `Mainnet` and `Fuji` too?

vms/proposervm/block.go
```diff
@@ -59,6 +59,9 @@ type Block interface {
 	buildChild(context.Context) (Block, error)
 
 	pChainHeight(context.Context) (uint64, error)
+	pChainEpochHeight(context.Context) (uint64, error)
```
yacovm, line 62: what is the use of the context here? Isn't the implementation just memory access?

vms/proposervm/block.go
```diff
@@ -79,6 +82,51 @@ func (p *postForkCommonComponents) Height() uint64 {
 	return p.innerBlk.Height()
 }
 
+// Calculates a block's P-Chain epoch height based on its ancestor's epoch membership
+func (p *postForkCommonComponents) getPChainEpoch(ctx context.Context, parentID ids.ID) (uint64, uint64, time.Time, error) {
+	parent, err := p.vm.getBlock(ctx, parentID)
```
yacovm, line 87: what happens if we call this on the genesis block?

vms/proposervm/block.go
```diff
@@ -163,9 +212,23 @@ func (p *postForkCommonComponents) Verify(
 	}
 
 	var contextPChainHeight uint64
-	if p.vm.Upgrades.IsEtnaActivated(childTimestamp) {
+	switch {
+	case p.vm.Upgrades.IsGraniteActivated(childTimestamp):
```
yacovm, line 224: what happens if the parent doesn't have granite activated but the child did? How should the system behave?

vms/proposervm/block.go
```diff
@@ -163,9 +212,23 @@ func (p *postForkCommonComponents) Verify(
...
+		pChainEpochHeight, _, _, err := p.getPChainEpoch(ctx, child.Parent())
+		if err != nil {
```
yacovm, line 226: We're checking the PchainEpochHeight is what it should be, what about the EpochNumber and EpochStartTime? shouldn't we check that too?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var _ = e2e.DescribeCChain("[ProposerVM Epoch]", func() {
+	tc := e2e.NewTestContext()
+	require := require.New(tc)
+
+	const txAmount = 10 * units.Avax // Arbitrary amount to send and transfer
+
+	ginkgo.It("should advance the proposervm epoch according to the upgrade config epoch duration", func() {
+		// TODO: Skip this test if Granite is not activated
```
yacovm, line 27: > // TODO: Skip this test if Granite is not activated

Why? Can't we force its activation in the test?

yacovm: We can just do `upgrades.GraniteTime = upgrade.InitiallyActiveTime` and it will make it activated.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,100 @@
...
+
+			// Issue enough txs to activate the proposervm form and ensure we advance the epoch (duration is 4s)
+			const numTxs = 7
+			txCount := 0
+
+			initialEpochNumber, _, _, err := proposerClient.GetEpoch(tc.DefaultContext())
+			require.NoError(err)
+
+			for range ticker.C {
+				acceptedNonce, err := ethClient.AcceptedNonceAt(tc.DefaultContext(), senderEthAddress)
```
yacovm, line 59: These lines aren't a critical part of the test, they're just for sending a transaction to advance the chain. Can we extract the lines 59-79 to a separate function?

vms/proposervm/service.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+// GetProposerBlockArgs is the parameters supplied to the GetProposerBlockWrapper API
+type GetProposerBlockArgs struct {
+	ProposerID ids.ID              `json:"proposerID"`
+	Encoding   formatting.Encoding `json:"encoding"`
+}
+
+func (p *ProposerAPI) GetProposerBlockWrapper(r *http.Request, args *GetProposerBlockArgs, reply *api.GetBlockResponse) error {
```
yacovm, line 39: I don't think we really need this API. Let's remove it as discussed.

vms/proposervm/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+var _ Client = (*client)(nil)
+
+type Client interface {
+	// GetProposedHeight returns the current height of this node's proposer VM.
+	GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error)
+	// GetProposerBlockWrapper returns the ProposerVM block wrapper
+	GetProposerBlockWrapper(ctx context.Context, proposerID ids.ID, options ...rpc.Option) ([]byte, error)
+	// GetEpoch returns the current epoch number, start time, and P-Chain height.
+	GetEpoch(ctx context.Context, options ...rpc.Option) (uint64, uint64, uint64, error)
```
yacovm, line 24: In the server side we have a nice struct (see below), let's also use it here:

```
type GetEpochResponse struct {
	Number       avajson.Uint64 `json:"Number"`
	StartTime    avajson.Uint64 `json:"StartTime"`
	PChainHeight avajson.Uint64 `json:"pChainHeight"`
}
```

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,100 @@
...
+
+		proposerClient := proposervm.NewClient(nodeURI.URI, "C")
+
+		tc.By("issuing C-Chain transactions to advance the epoch", func() {
+			// Issue enough C-Chain transactions to observe the epoch advancing
+			ticker := time.NewTicker(1 * time.Second)
+			defer ticker.Stop()
+
+			// Issue enough txs to activate the proposervm form and ensure we advance the epoch (duration is 4s)
+			const numTxs = 7
```
yacovm, line 52: if the epoch duration is 4 seconds why do we need 7 transactions separated by a second? Can't we do 1 transaction, wait for it to be finalized, sleep 4 seconds, then another transaction?

vms/proposervm/block.go
```diff
@@ -79,6 +82,59 @@ func (p *postForkCommonComponents) Height() uint64 {
 	return p.innerBlk.Height()
 }
 
+// Calculates a block's P-Chain epoch height based on its ancestor's epoch membership
+func (p *postForkCommonComponents) getPChainEpoch(ctx context.Context, parentID ids.ID) (uint64, uint64, time.Time, error) {
```
yacovm, line 86: can we:

1. add the parent block as a parameter to this function instead of retrieving it inside the method
2. move the logger to be a parameter to the method
3. make this a function and not an instance method, so it can be easily tested
4. make a unit test that tests this method directly

?

vms/proposervm/block.go
```diff
@@ -89,6 +145,7 @@ func (p *postForkCommonComponents) Height() uint64 {
 // 7) [child]'s timestamp is within its proposer's window
 // 8) [child] has a valid signature from its proposer
 // 9) [child]'s inner block is valid
+// 10) [child] has the expected P-Chain epoch height
```
yacovm, line 148: what about child has the expected epoch number and epoch start time? 

We need to verify every field we encode.

vms/proposervm/block.go
yacovm: This is the ` epochStartTime.IsZero()` case inside `getPChainEpoch` right?

vms/proposervm/block.go
```diff
@@ -163,9 +220,23 @@ func (p *postForkCommonComponents) Verify(
...
 	var contextPChainHeight uint64
-	if p.vm.Upgrades.IsEtnaActivated(childTimestamp) {
+	switch {
+	case p.vm.Upgrades.IsGraniteActivated(childTimestamp):
+		pChainEpochHeight, _, _, err := p.getPChainEpoch(ctx, child.Parent())
+		if err != nil {
+			p.vm.ctx.Log.Error("unexpected build verification failure",
+				zap.String("reason", "failed to get P-Chain epoch height"),
+				zap.Stringer("parentID", child.Parent()),
+				zap.Error(err),
```
yacovm, line 230: don't we need to fail fast and return an error here?

vms/proposervm/block/build.go
```diff
@@ -18,15 +18,21 @@ func BuildUnsigned(
...
 		StatelessBlock: statelessUnsignedBlock{
-			ParentID:     parentID,
-			Timestamp:    timestamp.Unix(),
-			PChainHeight: pChainHeight,
-			Certificate:  nil,
-			Block:        blockBytes,
+			ParentID:          parentID,
+			Timestamp:         timestamp.Unix(),
+			PChainHeight:      pChainHeight,
+			PChainEpochHeight: pChainEpochHeight,
```
yacovm, line 31: These new fields are being serialized even if we haven't activated granite yet (they'll be zero), right? 

Since the validators don't all upgrade the software version at the same second - how would it work? how can a validator in the current avalanchego version deserialize a block built by a validator in the new avalanchego version?

vms/proposervm/block.go
```diff
@@ -163,9 +220,23 @@ func (p *postForkCommonComponents) Verify(
...
+			)
+		}
+		if childHeight := child.PChainEpochHeight(); pChainEpochHeight != childHeight {
```
yacovm, line 233: can we extract the ICM epoch attribute validation into a separate function and then just check everything we need there, and not in this function?

vms/proposervm/service.go
```diff
@@ -0,0 +1,106 @@
...
+	PChainHeight avajson.Uint64 `json:"pChainHeight"`
+}
+
+func (p *ProposerAPI) GetEpoch(r *http.Request, _ *struct{}, reply *GetEpochResponse) error {
+	p.vm.ctx.Log.Debug("API called",
+		zap.String("service", "proposervm"),
+		zap.String("method", "getEpoch"),
+	)
+
+	lastAccepted, err := p.vm.LastAccepted(r.Context())
```
yacovm, line 79: Isn't this struct supposed to get accessed concurrently? I think we need some kind of synchronization if so.

vms/proposervm/service.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (p *ProposerAPI) GetProposerBlockWrapper(r *http.Request, args *GetProposerBlockArgs, reply *api.GetBlockResponse) error {
+	p.vm.ctx.Log.Debug("API called",
+		zap.String("service", "proposervm"),
+		zap.String("method", "getProposerBlockWrapper"),
+		zap.String("proposerID", args.ProposerID.String()),
+		zap.String("encoding", args.Encoding.String()),
+	)
+
+	block, err := p.vm.GetBlock(r.Context(), args.ProposerID)
```
yacovm, line 47: I am not sure VM access is thread safe

vms/proposervm/service.md
```diff
@@ -0,0 +1,178 @@
...
+    encoding: string // optional
+}) -> {
+    block: string,
+    encoding: string
+}
+```
+
+**Request:**
+
+- `proposerID` is the proposer ID. It should be in cb58 format.
```
yacovm, line 69: let's change this parameter name, it's the block ID.

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package proposervm
+
+import (
+	"testing"
+	"time"
+)
```
yacovm, line 9: seems like the tests are missing here.

vms/proposervm/block.go
geoff-vball: The Block interface ends up getting used in postForkBlock and postForkOption. postForkOption needs to get the parent block of the "option", which requires the context.

vms/proposervm/block.go
geoff-vball: Yes, that's correct. We've change the behaviour to actually check for epoch `0` instead of checking the starttime

tests/e2e/c/proposervm_epoch.go
geoff-vball: `upgrades.GraniteTime` gets set outside of the test. The current test suite runs all the tests once without granite activated, and once with them activated.

michaelkaplan13:
Closing in favor of #4238

PR #3747, opened 2025-02-26
Support caller-defined namespaces in merkledb

joshua-kim:
## Why this should be merged

Currently you're not able to create more than one merkledb instance, because we use a hard-coded metrics namespace into the provided metrics registerer. If a caller wants multiple db instances, it must create a separate metrics registry for each db instance.

We will need to have multiple merkledbs when we merkle-ize X/P-Chain State, so that we can support sub-trees

## How this works

Adds a parameter to specify the metrics namespace

## How this was tested

CI

No


yacovm (approved):

x/merkledb/db.go
```diff
@@ -244,8 +244,13 @@ type merkleDB struct {
 }
 
 // New returns a new merkle database.
-func New(ctx context.Context, db database.Database, config Config) (MerkleDB, error) {
-	metrics, err := newMetrics("merkledb", config.Reg)
+func New(
+	ctx context.Context,
+	db database.Database,
+	config Config,
+	namespace string,
```
StephenButtolph, line 251: Should this namespace be put in the config rather than passed alongside it?

x/merkledb/db.go
```diff
@@ -244,8 +244,13 @@ type merkleDB struct {
...
+) (MerkleDB, error) {
+	metrics, err := newMetrics(namespace, config.Reg)
```
StephenButtolph, line 253: Should all of the metrics be prefixed with `merkledb` so that passing in `""` as the namespace is the same behavior as before?

x/merkledb/db.go
joshua-kim: I think my rationale for not putting this in the `Config` is that this is probably something we want to explicitly configure and not depend on a default for...

x/merkledb/db.go
joshua-kim: We could maintain the previous behavior but I don't see a reason to because nothing is using `merkledb` at the moment. I also feel like the `merkledb` prefix/being forced into the metric name is gross, it feels like it bleeds the implementation into the metric.

x/merkledb/metrics.go
```diff
@@ -93,11 +94,22 @@ type prometheusMetrics struct {
...
+	return strings.Join([]string{prefix, namespace}, "_")
+}
+
+func newMetrics(prefix string, reg prometheus.Registerer) (metrics, error) {
 	// TODO: Should we instead return an error if reg is nil?
 	if reg == nil {
 		return &mockMetrics{}, nil
 	}
+
+	namespace := getMetricsNamespace(prefix, "merkledb")
```
StephenButtolph, line 111:
```suggestion
	namespace := metric.AppendNamespace(prefix, "merkledb")
```

StephenButtolph (approved):

PR #3748, opened 2025-02-26
Expose merkledb defaults

joshua-kim:
## Why this should be merged

When we merkle-ize the P and X Chains, it's likely that we're going to use similar merkledb configurations so this PR exposes defaults to avoid some boilerplate setup code.

## How this works

Exposes `NewConfig` that contains a `Config` with default values

## How this was tested

CI

No

### DIFF
--- a/x/merkledb/db.go
+++ b/x/merkledb/db.go
@@ -159,6 +159,21 @@ type MerkleDB interface {
 	Prefetcher
 }
 
+func NewConfig() Config {
+	return Config{
+		BranchFactor:                BranchFactor16,
+		Hasher:                      DefaultHasher,
+		RootGenConcurrency:          0,
+		HistoryLength:               300,
+		ValueNodeCacheSize:          units.MiB,
+		IntermediateNodeCacheSize:   units.MiB,
+		IntermediateWriteBufferSize: units.KiB,
+		IntermediateWriteBatchSize:  256 * units.KiB,
+		TraceLevel:                  InfoTrace,
+		Tracer:                      trace.Noop,
+	}
+}
+
 type Config struct {
 	// BranchFactor determines the number of children each node can have.
 	BranchFactor BranchFactor
--- a/x/merkledb/db_test.go
+++ b/x/merkledb/db_test.go
@@ -14,21 +14,17 @@ import (
 	"testing"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/database/dbtest"
 	"github.com/ava-labs/avalanchego/database/memdb"
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/trace"
 	"github.com/ava-labs/avalanchego/utils/hashing"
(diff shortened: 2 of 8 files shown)


x/merkledb/db_test.go
```diff
@@ -40,19 +39,11 @@ func newDB(ctx context.Context, db database.Database, config Config) (*merkleDB,
...
-		ValueNodeCacheSize:          units.MiB,
-		IntermediateNodeCacheSize:   units.MiB,
-		IntermediateWriteBufferSize: units.KiB,
-		IntermediateWriteBatchSize:  256 * units.KiB,
-		Reg:                         prometheus.NewRegistry(),
-		TraceLevel:                  InfoTrace,
-		Tracer:                      trace.Noop,
-	}
+	config := NewConfig()
+	config.HistoryLength = defaultHistoryLength
```
StephenButtolph, line 43: Why is this overridden?

x/merkledb/db.go
```diff
@@ -159,6 +159,21 @@ type MerkleDB interface {
 	Prefetcher
 }
 
+func NewConfig() Config {
```
StephenButtolph, line 162: It feels a bit odd to default to populating `mockMetrics` - I'm wondering if the `newDefaultConfig` needs to exist at all or if we should just use this. (and populate the metrics reg here)..

Although that feels a bit odd as well... so 🤷 maybe the issue is with the `mockMetrics` as a whole...

It feels like whoever made `mockMetrics` didn't know that `"github.com/prometheus/client_golang/prometheus/testutil"` existed.

x/merkledb/db_test.go
joshua-kim: This was because some of the tests do checks against the history. Maybe a better idea than what we're doing now would be for us to just use `NewConfig` in the tests and manually override there instead of doing it here.

joshua-kim: Removed this function entirely and tests that rely on the history length override the default config.

StephenButtolph (approved):

x/merkledb/proof_test.go
```diff
@@ -1906,8 +1914,8 @@ func FuzzChangeProofVerification(f *testing.F) {
 			require,
 			rand,
 			[]database.Database{db},
-			numKeyValues,
-			deletePortion,
+			config.HistoryLength/2,
+			0.25,
```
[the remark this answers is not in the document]
joshua-kim: I just moved it into the function signature because the function signature self-documents what these variables mean

x/merkledb/db_test.go
```diff
@@ -903,13 +883,13 @@ const (
...
+	config := NewConfig()
+	config.BranchFactor = tokenSizeToBranchFactor[tokenSize]
+	db, err := New(context.Background(), memdb.New(), config)
 	require.NoError(err)
 
-	const (
-		maxProofLen  = 100
-		maxPastRoots = defaultHistoryLength
-	)
+	maxProofLen := 100
```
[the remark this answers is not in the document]
joshua-kim: I don't know when we're supposed to use var blocks... it seems to make sense for package-level vars, when you want a short-hand for the zero value, or where you want to choose the type of something, but when those criteria aren't met I usually don't use them (and for zero values I usually explicitly assign to the zero value...).

PR #3749, opened 2025-02-27
Add ACP-176 e2e tests

StephenButtolph:


tests/e2e/c/hashing_contract.go
```diff
@@ -1,11 +0,0 @@
-// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
-// See the file LICENSE for licensing terms.
-
-// AUTOMATICALLY GENERATED. DO NOT EDIT!
-// Generated from hashing.sol by compile-contract.sh
```
StephenButtolph, line 5: This file doesn't seem to have ever been checked in to the repo (even going back to when this file was originally added to the kurtosis e2e tests).

tests/fixture/e2e/helpers.go
```diff
@@ -183,7 +183,12 @@ func SendEthTransaction(tc tests.TestContext, ethClient ethclient.Client, signed
 		return true
 	}, DefaultTimeout, DefaultPollingInterval, "failed to see transaction acceptance before timeout")
 
-	require.Equal(types.ReceiptStatusSuccessful, receipt.Status)
```
StephenButtolph, line 186: This seems to imply that we would only want to issue successful transactions. The gas consumption in this PR actually expects the transaction to fail.

tests/e2e/c/dynamic_fees.go
```diff
@@ -60,71 +83,107 @@ var _ = e2e.DescribeCChain("[Dynamic Fees]", func() {
...
 
-		var gasPrice *big.Int
-		tc.By("calling the expensive contract repeatedly until a gas price increase is detected", func() {
+		latest, err := ethClient.HeaderByNumber(tc.DefaultContext(), nil)
+		require.NoError(err)
+
+		initialGasPrice := latest.BaseFee
+		targetGasPrice := new(big.Int).Set(initialGasPrice)
+		targetGasPrice.Mul(targetGasPrice, big.NewInt(expectedGasPriceIncreaseNumerator))
+		targetGasPrice.Div(targetGasPrice, big.NewInt(expectedGasPriceIncreaseDenominator))
```
StephenButtolph, line 142: This should round up

tests/e2e/c/consume_gas.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity = 0.8.6;
+
+contract ConsumeGas {
+  function run() public {
+    while (true){}
```
aaronbuchwald, line 7: I'm partial to hand-writing EVM bytecode to accomplish this goal, but I guess this is fine 😉

aaronbuchwald (approved):

tests/e2e/c/hashing_contract.go
StephenButtolph: I didn't actually use a CLI/golang tool to generate this... I Just used remix (browser app)...

I don't think it's worthwhile to work to setup CI to compile solidity code unless we start writing a bunch of non-trivial solidity and issuing it in CI.

Although the Interop team may already do this to check the teleporter contracts

PR #3755, opened 2025-03-02
Remove unused `ForceCreateChain` function from `testManager`

comfsrt:
## Why this should be merged
The `ForceCreateChain` function in `testManager` was unused and did not conform to the `Manager` interface. Removing it removes "dead" code, does not affect test logic, and simplifies maintenance.


chains/test_manager.go
StephenButtolph, line 1: Seems like the `SubnetID` function can be removed as well.

StephenButtolph (approved):

PR #3756, opened 2025-03-03
fix: typo

LesCyber:
fix typo


StephenButtolph:
I don't think this makes the code more readable.

PR #3767, opened 2025-03-05
[ci] Stop emitting grafana link as an annotation

maru-ava:
## Why this should be merged

Instead the same job-wide link will only be emitted to the job output. This ensures that any annotation that appears on a workflow run can be perceived as actionable rather than informational.

## How this was tested

CI

N/A


tests/fixture/tmpnet/README.md
```diff
@@ -468,11 +468,9 @@ metrics and logs.
 #### CI
 
 Collection of logs and metrics is enabled for CI jobs that use
-tmpnet. Each job will execute a step titled `Notify of metrics
-availability` that emits a link to grafana parameterized to show
-results for the job. This link is emitted as a github actions
-annotation that appears on the summary page for a test run to increase
-the visibility of collected metrics and logs.
+tmpnet. Each job will execute a step including the text
```
darioush, line 471:
```suggestion
tmpnet. Each job will execute a step including the script
```
?

darioush (approved):

StephenButtolph (approved):

PR #3774, opened 2025-03-06
Remove support for AVM tx checksums

joshua-kim:
## Why this should be merged

Once we merge https://github.com/ava-labs/avalanchego/pull/3732, checksums will be updated to be return the merkle root instead. We have to make a decision to either continue to maintain the current tx checksumming, or store our txs in a merkle trie. The only current use-case of the tx checksum is for debugging/logging purposes, so this PR removes support for it entirely.

## How this works

Removes support for tx checksumming

## How this was tested

CI

Yes

### DIFF
--- a/vms/avm/block/executor/block.go
+++ b/vms/avm/block/executor/block.go
@@ -252,7 +252,7 @@ func (b *Block) Accept(context.Context) error {
 		zap.Stringer("blkID", blkID),
 		zap.Uint64("height", b.Height()),
 		zap.Stringer("parentID", b.Parent()),
-		zap.Stringer("checksum", checksum),
+		zap.Stringer("checksum", b.manager.state.Checksum()),
 	)
 	return nil
 }
--- a/vms/avm/block/executor/block.go
+++ b/vms/avm/block/executor/block.go
@@ -246,14 +246,13 @@ func (b *Block) Accept(context.Context) error {
 		return err
 	}
 
-	txChecksum, utxoChecksum := b.manager.state.Checksums()
+	checksum := b.manager.state.Checksum()
 	b.manager.backend.Ctx.Log.Trace(
 		"accepted block",
 		zap.Stringer("blkID", blkID),
 		zap.Uint64("height", b.Height()),
 		zap.Stringer("parentID", b.Parent()),
-		zap.Stringer("txChecksum", txChecksum),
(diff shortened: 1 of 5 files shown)


vms/platformvm/block/executor/acceptor.go
```diff
@@ -110,7 +110,7 @@ func (a *acceptor) ApricotAtomicBlock(b *block.ApricotAtomicBlock) error {
 		zap.Stringer("blkID", blkID),
 		zap.Uint64("height", b.Height()),
 		zap.Stringer("parentID", b.Parent()),
-		zap.Stringer("utxoChecksum", a.state.Checksum()),
```
joshua-kim, line 113: The files in `platformvm` feel like an unrelated change... so I can take them out if requested. The rationale for coupling them in this PR was to keep the logging consistent.

bentrnr21:
Thanks for this update! Removing AVM tx checksums seems like a logical step to streamline the codebase and improve maintainability. I took a look at the changes, and everything appears clean. However, I have a few questions:

	1.	Are there any edge cases where removing checksum validation could lead to unexpected issues, such as in legacy transactions?
	2.	Should we update the documentation or migration notes to notify developers who might still rely on these checksums?
	3.	Have all relevant test cases been covered, particularly for scenarios where old transactions might still exist in some nodes?

Overall, this looks like a solid improvement. Thanks again for the work on this!

vms/avm/block/executor/block.go
```diff
@@ -246,14 +246,13 @@ func (b *Block) Accept(context.Context) error {
...
-	txChecksum, utxoChecksum := b.manager.state.Checksums()
+	checksum := b.manager.state.Checksum()
 	b.manager.backend.Ctx.Log.Trace(
 		"accepted block",
 		zap.Stringer("blkID", blkID),
 		zap.Uint64("height", b.Height()),
 		zap.Stringer("parentID", b.Parent()),
-		zap.Stringer("txChecksum", txChecksum),
-		zap.Stringer("utxoChecksum", utxoChecksum),
+		zap.Stringer("checksum", checksum),
```
StephenButtolph, line 255: Super nit to align with the other fields
```suggestion
		zap.Stringer("checksum", b.manager.state.Checksum()),
```

StephenButtolph (approved):

PR #3781, opened 2025-03-10
[ci] Disable monitoring for jobs of PRs of fork branches

maru-ava:
## Why this should be merged

Jobs executing on PRs based on fork branches won't have access to the secrets required to configure collectors for logs and metrics. Disabling the collection in the custom action avoids failing such jobs. This was the behavior prior to the switch to starting collectors with golang.

## How this works

- makes collector start and collection checks in the run-monitored-tmpnet-cmd custom action conditional on credentials being non-empty
- emits an annotation if collector credentials are missing in the hopes of notifying interested parties of unintended misconfiguration 

## How this was tested

- [x] CI (should show no change)
- [ ] Passing run from a fork branch

N/A


aaronbuchwald (approved):

yacovm (approved):
Thanks!

PR #3786, opened 2025-03-11
Add canoto annotations for warp bit set signature

aaronbuchwald:
## Why this should be merged

Add canoto annotations to the warp bit set signature type, so that it can be used in HyperSDK DSMR without conversion to using the AvalancheGo codec.

## How this works

Adds canoto annotations to `warp.BitSetSignature` and the private `canotoData` field and run `go generate ./...`.


StephenButtolph:
Closing due to staleness.

PR #3792, opened 2025-03-12
Timely halt Avalanche engine

yacovm:


snow/engine/avalanche/bootstrap/config.go
```diff
@@ -39,4 +39,7 @@ type Config struct {
 	// If StopVertexID is empty, the engine will generate the stop vertex based
 	// on the current state.
 	StopVertexID ids.ID
+
+	// Haltable signals when the engine is stopped
+	common.Haltable
```
StephenButtolph, line 44: The snowman bootstrapper includes `ShouldHalt func() bool` and now this includes `common.Haltable` - can we align these to be the same? If feels weird to do the same thing 2 different ways.

fwiw I do think passing around a `Haltable` is more clear than passing around a `func() bool`.

yacovm: I agree, changed accordingly.

StephenButtolph (approved):

aaronbuchwald (approved):

PR #3798, opened 2025-03-13
[docker] Update all images to debian12/bookworm

maru-ava:
## Why this should be merged

With the EOL of debian11-based Ubuntu 20.04/focal and the deprecation and impending removal of 20.04 runners from github actions, an upgrade to our image userland to debian12/bookworm ensures our images use a userland that is being tested with the debian12-based 22.04.

This will also ensure compatibility of subnet-evm binaries built with goreleaser on 22.04 runners (and so requiring a debian12 userland) with avalanchego-based images that will now also have a debian12 userland.

## How this was tested

CI

Maybe?


StephenButtolph (approved):
lgtm

PR #3799, opened 2025-03-13
Fix flake TestIngressConnCount

yacovm:
## Why this should be merged

The test spawns three instances of test network one after the other. In an overwhelming number of cases, they connect to each other according to their order of creation.

However, in some rare cases, they connect to each other out of order.

I made the test use set semantics instead of list semantics in order to not rely on their order of creation.

## How this works

Unit test flake fix

## How this was tested

Hopefully it won't flake in CI anymore

No.

### DIFF
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -338,21 +338,24 @@ func TestIngressConnCount(t *testing.T) {
 		return result
 	}, time.Minute, time.Millisecond*10)
 
-	ingressConnections := make([]int, 0, len(networks))
-	healthCheckErrors := make([]error, 0, len(networks))
+	ingressConnCount := set.Of[int]()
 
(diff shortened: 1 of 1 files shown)


network/network_test.go
```diff
@@ -338,21 +338,27 @@ func TestIngressConnCount(t *testing.T) {
 		return result
 	}, time.Minute, time.Millisecond*10)
 
-	ingressConnections := make([]int, 0, len(networks))
-	healthCheckErrors := make([]error, 0, len(networks))
+	ingressConnCount := make(map[int]struct{})
```
StephenButtolph, line 341: Can we use the `set.Set` here instead?

StephenButtolph (approved):

PR #3803, opened 2025-03-14
refactor: use `libevm/core/types` directly

ARR4N:
## Why this should be merged

All of the `coreth/core/types.*` identifiers used in AvalancheGo are just aliases of the `libevm` ones anyway. Removing them here allows for the aliases to be deleted in `coreth`.

## How this works

Find and replace.

## How this was tested

Existing tests.

No


darioush (approved):
lgtm pending CI

PR #3809, opened 2025-03-17
Fix typos in iterator.go

Marcofann:
Fixed a typo in `iterator.go`:
  - Changed `allow use to use` → `allow us to use`
  - Changed `ie.` → `i.e.`
  - Changed `without causing error` → `without causing an error`

## How this was tested
No functional changes were introduced. The changes only affect comments and error messages.

No, these are minor comment and message fixes that do not impact functionality.


genesis/genesis.go
```diff
@@ -41,11 +41,11 @@ var (
 	errNoCChainGenesis                 = errors.New("C-Chain genesis cannot be empty")
 	errNoTxs                           = errors.New("genesis creates no transactions")
```
yacovm, line 42: i think the typos in this file are fine to keep, in case you think otherwise - can you please fix the typos in this file in a different pull request, and leave only the typo fixes of the iterator?

Marcofann: done

PR #3811, opened 2025-03-18
[ci] Source shellcheck from nix instead of installing via script

maru-ava:
## Why this should be merged

We're already using nix and it's simpler not to have to maintain a multi-platform installer.

## How this works

- Adds shellcheck to the flake devshell
- Removes installation from the shellcheck script
- Updates the CI job to use nix 

## How this was tested

CI

N/A

### DIFF
--- a/scripts/shellcheck.sh
+++ b/scripts/shellcheck.sh
@@ -12,42 +12,10 @@ if ! [[ "$0" =~ scripts/shellcheck.sh ]]; then
   exit 255
 fi
 
-VERSION="v0.9.0"
-
-function get_version {
-  local target_path=$1
-  if command -v "${target_path}" > /dev/null; then
-    echo "v$("${target_path}" --version | grep version: | awk '{print $2}')"
-  fi
-}
-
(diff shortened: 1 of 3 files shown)


scripts/shellcheck.sh
```diff
@@ -12,42 +12,10 @@ if ! [[ "$0" =~ scripts/shellcheck.sh ]]; then
...
-    cp /tmp/shellcheck-"${VERSION}"/shellcheck "${SHELLCHECK}"
-  fi
-fi
-
 # `find *` is the simplest way to ensure find does not include a
 # leading `.` in filenames it emits. A leading `.` will prevent the
 # use of `git apply` to fix reported shellcheck issues. This is
 # compatible with both macos and linux (unlike the use of -printf).
 #
 # shellcheck disable=SC2035
```
ARR4N, line 20: It's shellcheck all the way down!

ARR4N (approved):
FYI there's a [shellcheck action](https://github.com/ludeeus/action-shellcheck) that avoids the indirection via a script.

StephenButtolph (approved):

PR #3812, opened 2025-03-18
Use IPv4 addresses if possible

StephenButtolph:
## Why this should be merged

Saves 12 bytes per IP!

## How this works

On the receiving side of these IPs, we already handle IPv4 addresses. However for compatibility reasons, we only send IPv6 addresses currently. This is no longer required (as of Etna activation). So we can save the 12 bytes.

## How this was tested

The e2e upgrade test verifies that this is backwards compatible.


joshua-kim (approved):

StephenButtolph:
Will merge after the `v1.13.0` release cadence

PR #3822, opened 2025-03-21
[ci] Use SHAs instead of tags for 3rd-party github actions

maru-ava:
## Why this should be merged

We should be able to trust that built-in actions (actions/*) have stable tags, but for everything else a SHA should be preferred to a tag to avoid the possibility of a supply-chain attack.

## How this works

Updates 3rd party github action references to use SHAs.

## How this was tested

CI

N/A

### DIFF
--- a/.github/actions/install-nix/action.yml
+++ b/.github/actions/install-nix/action.yml
@@ -10,7 +10,7 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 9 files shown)


ARR4N (approved):
I haven't manually cross-checked the SHAs vs exact versions. Happy to if you'd like but didn't think it was necessary.

PR #3824, opened 2025-03-21
[nix] Add package for official golang binaries

maru-ava:
## Why this should be merged

Ensures the availability of the required golang version in the nix dev shell.

## How this works

- Adds a nix derivation for the official golang binaries
- Adds the package to the dev shell
- Ensures nix-using custom actions replicate the caching behaviors of setup-go
- Removes usage of setup-go-for-project from jobs using nix-using custom actions

## How this was tested

- CI (including verifying that caching is active)

N/A


JuanLeon2 (approved):

JuanLeon2:
Can you expand why this is not risky: "Attempts to use in CI are not recommended since it doesn't support dependency caching like action/setup-go."  What if someone tries to use it in CI?

JuanLeon2:
Actually, this seems to have been superseded by PR 3826 https://github.com/ava-labs/avalanchego/pull/3826

True? 

Can you remove this PR in that case?

StephenButtolph:
@JuanLeon2 I don't think this PR has been superseded. #3826 seems to depend on this PR, not supersede it.

.github/actions/install-nix/action.yml
```diff
@@ -15,3 +15,19 @@ runs:
         github_access_token: ${{ inputs.github_token }}
     - run: nix develop --command echo "dependencies installed"
       shell: bash
+    # Cache Go modules and build cache
+    - uses: actions/cache@v4
```
maru-ava, line 19: This has to be inlined in both custom actions rather than encapsulated in its own custom action due to composite github custom actions supporting either relative references to repo-local custom actions or being usable from outside the repo. This can be revisited if/when coreth/subnet-evm are no longer living in separate repos.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -108,8 +107,6 @@ runs:
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
 
-    - uses: ./.github/actions/install-nix
```
maru-ava, line 116: I noticed that this step was duplicating the nix installation already performed by the run-monitored-tmpnet-cmd custom action.

.github/actions/install-nix/action.yml
```diff
@@ -15,3 +15,23 @@ runs:
         github_access_token: ${{ inputs.github_token }}
     - run: nix develop --command echo "dependencies installed"
       shell: bash
+    # Cache Go modules (architecture-independent)
+    - uses: actions/cache@v4
+      id: go-mod-cache
```
maru-ava, line 20: Setting this field enables checking for a cache hit in subsequent steps

maru-ava:
> Can you expand why this is not risky: "Attempts to use in CI are not recommended since it doesn't support dependency caching like action/setup-go." What if someone tries to use it in CI?

I've ensured caching equivalent to that provided by setup-go for jobs that use nix-provided golang.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -108,8 +107,6 @@ runs:
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
```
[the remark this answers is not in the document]
maru-ava: The run-monitored-tmpnet-cmd step installs nix.

joshua-kim (approved):

PR #3826, opened 2025-03-22
[nix] Migrate linter installation to nix

maru-ava:
## Why this should be merged

Migrate linter installation to nix to reduce installation time and ensure consistency between CI and local usage.

If seeing VCS errors when running lint.sh under a nix shell on macos, the following commands should fix it:

```bash
sudo xcode-select --reset
sudo xcode-select --switch /Library/Developer/CommandLineTools
```

Reference: https://github.com/NixOS/nixpkgs/issues/355486#issuecomment-2488329223

## How this works

- adds a package for golangci-lint that downloads release binaries
- uses nixos packages for shellcheck and actionlint  

## How this was tested

CI

N/A

## TODO

- [ ] Merge https://github.com/ava-labs/avalanchego/pull/3824

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,19 +122,18 @@ jobs:
       #     loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
   Lint:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: nix develop --command bash -x {0}
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup-go-for-project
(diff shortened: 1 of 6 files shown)


JuanLeon2:
@StephenButtolph are you ok with this?
Secondarily: why does this require Stephen?  Because it changes a github action?

StephenButtolph:
> are you ok with this?

Yes.

> why does this require Stephen? Because it changes a github action?

Unless otherwise specified, I am the only person that default as an owner of a file. The change in the `go.mod` marks me as a required reviewer

PR #3831, opened 2025-03-23
Make sure inner state summary accept is called

aaronbuchwald:
## Why this should be merged

This PR is in response to https://github.com/ava-labs/hypersdk/issues/1993

This fixes a potential path where the ProposerVM skips calling `Accept` on the inner VM's state sync summary when the inner VM relies on that call to populate its state.

The previous behavior is problematic for VMs that implement dynamic state sync as outlined here: https://github.com/ava-labs/avalanchego/pull/3831#issuecomment-2751474021

## How this works

This switches from skipping the `Accept` call to the innerVM's state sync summary to calling it and re-aligning the ProposerVM's indices to match the post-accept state of the innerVM (via `SetState`).

## How this was tested

This was tested with the existing unit tests and modifying `TestStateSummaryAcceptOlderBlock` to confirm the new behavior.

No


aaronbuchwald:
https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/vms/proposervm/vm.go#L443

The ProposerVM implements `repairAcceptedChainByHeight` which rolls the ProposerVM (outer VM) last accepted height back to match the inner VM's last accepted height.

This occurs during `Initialize`:

https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/vms/proposervm/vm.go#L194

and when transitioning into `StateSyncing` in the call to `SetState`:

https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/vms/proposervm/vm.go#L280

This means that when we hit the state sync summary, the ProposerVM's last accepted height has already been rolled back to the last accepted height first reported by the inner VM.

aaronbuchwald:
When we accept a state summary, we handle it in one of three ways (https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/snow/engine/snowman/syncer/state_syncer.go#L403):

1) StateSyncSkipped
2) StateSyncStatic
3) StateSyncDynamic

Each of these cases invokes `onDoneStateSyncing(...)` https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/snow/engine/snowman/syncer/state_syncer.go#L57, which is set by the chain manager to start the next engine gear: the bootstrapper.

https://github.com/ava-labs/avalanchego/blob/bf7f283679417285fb97b999343907dae381cb36/chains/manager.go#L1386

Actually each engine gear will re-fetch the last accepted height from the VM giving each engine gear equal opportunity to move the last accepted height forward/backward independently as long as this does not violate the invariants that must be upheld for the VM / engine (ie. cannot roll back consensus or roll back a block that applied a change to atomic memory on the primary network).

Therefore we can rollback the ProposerVM's height to match the state sync summary as proposed in this PR.

We need to defer to the inner VM to address the following path:

1) Start dynamic state sync to block 100 and progress to block 200
2) Restart and inner VM reports its last accepted height as 200, which matches the ProposerVM
3) ProposerVM skips calling `Accept` on the state sync summary because the inner VM reports a last accepted height of 200 even though it does not have the full state at this height and relies on the call to `Accept` the state sync summary

vms/proposervm/state_summary.go
```diff
@@ -62,5 +56,8 @@ func (s *stateSummary) Accept(ctx context.Context) (block.StateSyncMode, error)
 	// innerSummary.Accept may fail with the proposerVM block and index already
 	// updated. The error would be treated as fatal and the chain would then be
 	// repaired upon the VM restart.
+	// After the inner summary is accepted, the engine transitions to bootstrapping
+	// and SetState is responsible for re-aligning the ProposerVM to the height reported
+	// by the inner VM after handling state sync.
```
aaronbuchwald, line 61: If we are transitioning out of state sync, then we already re-align the proposervm to the inner vm during that exit: https://github.com/ava-labs/avalanchego/blob/66490e40149bb64cdaf750c88e6fb763e7595bc8/vms/proposervm/vm.go#L272

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -564,12 +565,28 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+	// test Accept summary invokes innerVM
+	calledInnerAccept := false
 	innerSummary.AcceptF = func(context.Context) (block.StateSyncMode, error) {
+		innerVM.LastAcceptedF = func(context.Context) (ids.ID, error) {
+			return innerSummary.ID(), nil
+		}
+		innerVM.GetBlockF = func(context.Context, ids.ID) (snowman.Block, error) {
+			return innerBlk, nil
+		}
+		calledInnerAccept = true
```
aaronbuchwald, line 579: We're changing the behavior from the innerVM Accept never getting called to updating the last accepted block, so we need to update the mocked functions.

This ensures that in `SetState`, we correctly align the ProposerVM to the innerVM's updated height.

yacovm:
> Start dynamic state sync to block 100 and progress to block 200

Can you clarify the difference between "state sync to block" and "progress to block" ?

>  ProposerVM skips calling Accept on the state sync summary because the inner VM reports a last accepted height of 200 even though it does not have the full state at this height and relies on the call to Accept the state sync summary

could you perhaps point to where in the code it skips calling accept on the inner state sync, and why / how can the inner VM report a last accepted height of 200 even though it doesn't have the full state at that height? 

AFAIK you can have state without having the blocks, but not having blocks without having the corresponding state...

vms/proposervm/state_summary.go
```diff
@@ -40,12 +40,6 @@ func (s *stateSummary) Height() uint64 {
 }
 
 func (s *stateSummary) Accept(ctx context.Context) (block.StateSyncMode, error) {
-	// If we have already synced up to or past this state summary, we do not
-	// want to sync to it.
-	if s.vm.lastAcceptedHeight >= s.Height() {
-		return block.StateSyncSkipped, nil
-	}
-
```
aaronbuchwald, line 48: If the VM's last accepted height is >= the summary's height, then we return skipped and never call `Accept` on the inner VM's summary.

aaronbuchwald:
> > Start dynamic state sync to block 100 and progress to block 200
> 
> Can you clarify the difference between "state sync to block" and "progress to block" ?
> 
> > ProposerVM skips calling Accept on the state sync summary because the inner VM reports a last accepted height of 200 even though it does not have the full state at this height and relies on the call to Accept the state sync summary
> 
> could you perhaps point to where in the code it skips calling accept on the inner state sync, and why / how can the inner VM report a last accepted height of 200 even though it doesn't have the full state at that height?
> 
> AFAIK you can have state without having the blocks, but not having blocks without having the corresponding state...

https://github.com/ava-labs/avalanchego/pull/3831/files#r2012712067

State sync to block means set an initial target of block 100. After this, dynamic state sync will continue accepting blocks, so "progress to block" means that it will keep updating its target as blocks get accepted and eventually the up-to-date target will be block 200.

Here's another summary that may be more helpful and provide context on dynamic state sync:

In Dynamic State Sync, the VM accepts a state sync summary at an initial target height T. It then immediately transitions into normal consensus, so that it can dynamically update the target it's syncing to.

During this time, the VM verifies blocks vacuously - treating them all as valid without verifying them since it does not have the state necessary to verify them properly and updating its sync target when blocks are accepted.

This allows the VM to start fetching the state for an initial target T and update it as new blocks are accepted by consensus. Assuming the majority of the network is online and fully verifying blocks, the dynamic state sync node will only ever accept blocks correctly decided by the network (despite potentially marking invalid blocks as verified).

However, the ProposerVM continues to verify/accept/reject blocks in the same way ie. we still verify the proposervm headers and update the accepted indices when accepting the proposervm blocks.

As a result we have the following path:
Start a dynamic state sync with target T
InnerVM starts async fetching the required state for target T and updates it to T+K as new blocks are accepted
When the consensus engine accepts blocks, the proposerVM updates its accepted index and the innerVM updates its target, but does not have the full state until dynamic state sync completes.

Now if we restart the node, we may observe the following:
Restart and the proposerVM sees its last accepted height at T+K (from the last run)
State sync is enabled so it attempts to fetch a state sync summary from the network.
If the proposervm then skips that state sync summary because it has already accepted it (which is possible), then it may never call Accept on the innerVM

This means that the innerVM cannot count on a call to accept a state sync summary on startup to initialize the state sync process even if it's left in a state that requires it. We can handle this in two ways, we can either force the ProposerVM to always call Accept on the innerVM's summary, or we can force the innerVM to handle all paths through Initialize or SetState to detect when it's reached this state

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -564,12 +565,28 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
 
 	summary, err := vm.GetStateSummary(context.Background(), reqHeight)
 	require.NoError(err)
+	require.Equal(summary.Height(), reqHeight)
 
-	// test Accept skipped
+	// test Accept summary invokes innerVM
+	calledInnerAccept := false
```
yacovm, line 571: why the nested callback override? 

Why not do:

```
	innerVM.LastAcceptedF = func(context.Context) (ids.ID, error) {
		return innerSummary.ID(), nil
	}
	innerVM.GetBlockF = func(context.Context, ids.ID) (snowman.Block, error) {
		return innerBlk, nil
	}
	innerSummary.AcceptF = func(context.Context) (block.StateSyncMode, error) {
		calledInnerAccept = true
		return block.StateSyncStatic, nil
	}
```

?

yacovm (approved):

vms/proposervm/state_syncable_vm_test.go
aaronbuchwald: It is only after `Accept` has been invoked that the other two functions should be updated, so I override them within `AcceptF`

StephenButtolph (approved):
lgtm. Can we update the documentation on `StateSummary#Accept`? Something along the lines of:
```golang
	// Accept notifies the VM that this is a canonically accepted state summary.
	// The VM is expected to return how it plans on handling this summary via
	// the [StateSyncMode].
	//
	// Invariant: The VM must be able to correctly handle the acceptance of a
	// state summary that is older than its current state.
	//
	// See [StateSyncSkipped], [StateSyncStatic], and [StateSyncDynamic] for the
	// expected invariants the VM must maintain based on the return value.
	Accept(context.Context) (StateSyncMode, error)
```

aaronbuchwald:
> lgtm. Can we update the documentation on `StateSummary#Accept`? Something along the lines of:
> 
> ```go
> 	// Accept notifies the VM that this is a canonically accepted state summary.
> 	// The VM is expected to return how it plans on handling this summary via
> 	// the [StateSyncMode].
> 	//
> 	// Invariant: The VM must be able to correctly handle the acceptance of a
> 	// state summary that is older than its current state.
> 	//
> 	// See [StateSyncSkipped], [StateSyncStatic], and [StateSyncDynamic] for the
> 	// expected invariants the VM must maintain based on the return value.
> 	Accept(context.Context) (StateSyncMode, error)
> ```

Added https://github.com/ava-labs/avalanchego/pull/3831/commits/08fd4b800b9b6fd959f3641d0f10c872fd2fdf3e

vms/proposervm/state_summary.go
aaronbuchwald: Removing this leads to this issue: https://github.com/ava-labs/avalanchego/issues/3948

The goal of this PR is to ensure that the innerVM gets notified about the state sync summary accept call. If the proposerVM skips it without notifying the innerVM, then it results in a very hard to follow DevX for VM implementers if they depend on a clear place to start state sync.

We have conflicting needs here. The ProposerVM will skip the state summary out of band if it is below its current last accepted height. This prevents the bug because we'll never rewind the proposervm's last accepted height and maintain the desired guarantee. Alternatively, we can handle this special case, but that seems much less desirable.

The problem is the ProposerVM needs to be able to skip a state summary instead of accepting and re-winding, but this requires that it can force the innerVM to accept the summary. This breaks in two ways.

1) If the node dies between rolling back the ProposerVM and calling Accept on the innerVM, we have broken the invariant
2) We must allow the innerVM to support "skips", but that means we need to check if it's going to skip, so we can avoid an invalid rollback of the ProposerVM

PR #3838, opened 2025-03-25
fix function batch_test.go

futreall:
## Why this should be merged
// The variable name 'pebbleByteOverHead' contains a typo:
// 'overhead' should be spelled as one word, e.g. 'pebbleByteOverhead'.


StephenButtolph:
This PR doesn't compile. Feel free to make a new one with this change that does compile.

PR #3839, opened 2025-03-25
Proactively detect and report duplicate gossiped/malformed messages

tsachiherman:
## Why this should be merged

This PR modify the way the existing gossip implementation reports dropped incoming transaction. 
When reporting the `gossip_count` and `gossip_bytes` for received messages, an additional label was added : `dropped`. This label would specify the reason for a transaction to be dropped ( `malformed` / `duplicate` / `other` ) in case it was dropped. ( `not` otherwise ). 

## How this works

To support the existing avalanchego VMs, a simplified `mempool` was created. This mempool maintain a map of all the pending transaction, and would label a duplicate transaction as such. The two VMs that are using the gossip, `avm` and `platformvm` were modified to use this as the backing mempool in addition to the existing `vms/txs/mempool`.

Supporting the duplicate transaction on the coreth proven to be slightly more complicated, due to it's subpool architecture. In order to support both cases, the `gossip.Metrics` was extended to allow labeling transaction during the `mempool.Add` execution.

The existing, and updated metrics related logic was moved from `gossip.go` into `metrics.go`. This cleaned up the former from metrics related logic.

## Related PRs ?

The coreth PR is https://github.com/ava-labs/coreth/pull/935

No


vms/platformvm/network/gossip.go
```diff
@@ -115,7 +114,7 @@ func (g *gossipMempool) Add(tx *txs.Tx) error {
 		)
 
 		g.Mempool.MarkDropped(txID, err)
-		return fmt.Errorf("failed verification: %w", err)
+		return fmt.Errorf("failed verification: %w", gossip.ErrGossipableMalformed)
```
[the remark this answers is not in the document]
tsachiherman: done

network/p2p/gossip/gossip.go
```diff
@@ -76,6 +120,10 @@ var (
 	ErrInvalidDiscardedSize     = errors.New("discarded size cannot be negative")
 	ErrInvalidTargetGossipSize  = errors.New("target gossip size cannot be negative")
 	ErrInvalidRegossipFrequency = errors.New("re-gossip frequency cannot be negative")
+	// ErrGossipableAlreadyKnown is returned by the Set's Add method if the gossipable is already present.
+	ErrGossipableAlreadyKnown = errors.New("gossipable already known")
+	// ErrGossipableMalformed indicates a gossipable with invalid structure or content.
+	ErrGossipableMalformed = errors.New("gossipable malformed")
```
joshua-kim: Instead of defining drop reasons + metrics in the `gossip` package, should we define them in the vm's mempool implementation? One example where this is awkward is we're using `ErrGossipableMalformed` when failing verification in the mempool, when this error feels more like an error intended for parsing failures.

I think one way we could do this is to have a mempool implementation in this package that handles dropping duplicate/txs that fail to parse, that is then wrapped by the vm mempool implementation which adds its own metrics.

vms/avm/network/gossip.go
```diff
@@ -119,7 +118,7 @@ func (g *gossipMempool) Add(tx *txs.Tx) error {
 	// Verify the tx at the currently preferred state
 	if err := g.txVerifier.VerifyTx(tx); err != nil {
 		g.Mempool.MarkDropped(txID, err)
-		return err
+		return fmt.Errorf("%w: transaction %s verification failed - %s", gossip.ErrGossipableMalformed, txID, err.Error())
```
joshua-kim: Wonder if we need a separate error type for malformed vs verification failing

network/p2p/gossip/gossip.go
```diff
@@ -34,6 +34,16 @@ const (
 	unsentType = "unsent"
 	sentType   = "sent"
 
+	// omitted indicate that the gossipable element was not added to the set due to some reason.
+	// for sent message, we'll use notReceive below.
+	omittedLabel = "omitted"
```
joshua-kim, line 39: nit: In other parts of the mempool code we use the term "dropped" for this

network/p2p/gossip/gossip.go
```diff
@@ -34,6 +34,17 @@ const (
...
+
+	omittedFailedVerification = "failed-verification"
```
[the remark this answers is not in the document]
tsachiherman: thanks. updated.

network/p2p/gossip/gossip.go
```diff
@@ -34,6 +34,17 @@ const (
...
+	omittedDuplicate          = "duplicate"
+	omittedMalformed          = "malformed"
+	omittedOther              = "other"
+	omittedNot                = "not"
+	notReceive                = "n/a"
```
joshua-kim, line 46: I'm surprised this works... I didn't think prometheus allowed special characters other than colons/underscores.

network/p2p/gossip/gossip.go
tsachiherman: Yes, I was thinking the same. The reason I held back was that it seems to be a bit "too much" for what we're trying to achieve. I did update the labels to separate the Malformed ( parsing error ) and FailedVerification ( for verification failure ) as @StephenButtolph suggested.
i.e. the goal in this PR is to be able to identify if we have any of these fail cases at all.

Please LMK if you'd want to proceed with the code as is or the refactor as you suggested above.

network/p2p/gossip/gossip.go
tsachiherman: I'll rename to dropped.

network/p2p/gossip/gossip.go
joshua-kim: I think refactoring would be better - otherwise I think the point of this PR will be lost (increasing visibility) if we bucket unrelated drop reasons into things like malformed/other IMO.

network/p2p/gossip/gossip.go
tsachiherman: changed to "not_receive" to avoid potential issues.

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	pullType   = "pull"
+	unsentType = "unsent"
+	sentType   = "sent"
+
+	// dropped indicate that the gossipable element was not added to the set due to some reason.
+	// for sent message, we'll use notReceive below.
+	droppedLabel = "dropped"
+
+	droppedMalformed = "malformed"
+	DroppedDuplicate = "duplicate"
```
joshua-kim, line 32: Why is this exported?

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+		metrics: metrics,
+	}
+	return m, nil
+}
+
+func (m *mempool[T]) Add(tx T) error {
+	txID := tx.GossipID()
+	if m.Has(txID) {
+		m.metrics.ObserveIncomingGossipable(txID, DroppedDuplicate)
+		return fmt.Errorf("%w: %s", ErrDuplicateTx, txID)
```
joshua-kim, line 46: We only return an error in the duplicate case... so do we even need to return an error if it's always the same case? I think we could just return a true/false bool here

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (m *mempool[T]) Remove(txs ...T) {
+	for _, tx := range txs {
+		delete(m.Txs, tx.GossipID())
+	}
+}
+
+func (m *mempool[T]) Len() int {
+	return len(m.Txs)
+}
```
joshua-kim, line 61: What is this used for?

vms/avm/network/gossip.go
```diff
@@ -94,8 +102,11 @@ type gossipMempool struct {
 	txVerifier TxVerifier
 	parser     txs.Parser
 
-	lock  sync.RWMutex
-	bloom *gossip.BloomFilter
+	bloomLock sync.RWMutex
+	bloom     *gossip.BloomFilter
+
+	gossipMempoolLock sync.RWMutex
```
joshua-kim, line 108: Why can't we just use a single lock for this struct?

vms/platformvm/network/gossip.go
```diff
@@ -75,29 +74,48 @@ func newGossipMempool(
...
+	if err != nil {
+		return nil, err
+	}
+	gMempool, err := gossip.NewMempool[*txs.Tx](gossipMetrics)
+	if err != nil {
+		return nil, err
+	}
+	if err != nil {
+		return nil, err
+	}
```
joshua-kim, line 89:
```suggestion
```

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+// See the file LICENSE for licensing terms.
+
+package gossip
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+
+	xmempool "github.com/ava-labs/avalanchego/vms/txs/mempool"
```
joshua-kim, line 11: We should avoid having a dependency on `vms`... I think one way to avoid this is to just remove the error type we're importing but if we really wanted to keep an error we should just define it here instead.

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var ErrDuplicateTx = xmempool.ErrDuplicateTx
+
+type Mempool[T Gossipable] interface {
```
joshua-kim, line 16: Do we need this interface?

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	droppedOther     = "other"
```
joshua-kim, line 33: Can we remove this entirely? I'm wondering if the mempool the VM defines can just get a reference to `Metrics` so that it can define the label here instead of bucketing things under `other` which will lose visibility.

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+
+	typeLabel  = "type"
+	pushType   = "push"
+	pullType   = "pull"
+	unsentType = "unsent"
+	sentType   = "sent"
+
+	// dropped indicate that the gossipable element was not added to the set due to some reason.
+	// for sent message, we'll use notReceive below.
+	droppedLabel = "dropped"
```
joshua-kim, line 29: I find the naming for these to be extremely confusing because we use "dropped" even for outbound gossip, for which we can't ever actually drop it (and that's why we have the `not_receive` label). Without looking at the code I wouldn't understand how to interpret these metric names.

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	Add(tx T) error
```
joshua-kim, line 17: I don't mind us changing names to be `tx` instead of `gossipable` going forwards, but whatever we do we should be consistent in this diff (inconsistent w/ `handleIncomingGossipables`). I actually prefer `tx` naming, the only reason we use `gossipables` in the current code is because i thought we were going to use this for ip gossip, but we only ended up using this for txs.

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+type mempool[T Gossipable] struct {
+	Txs     map[ids.ID]bool
+	metrics *Metrics
+}
+
+func NewMempool[T Gossipable](
+	metrics *Metrics,
+) (*mempool[T], error) {
```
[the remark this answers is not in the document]
tsachiherman: good find; what I meant what to return an interface and embed that. example:
https://go.dev/play/p/W6dW5mPEceD

But I'm open to eliminate the interface here and use the struct directly. I'll change the struct pointer with interface. Please let me know if you think it should be an exported struct instead.

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+
+	xmempool "github.com/ava-labs/avalanchego/vms/txs/mempool"
+)
+
+var ErrDuplicateTx = xmempool.ErrDuplicateTx
+
+type Mempool[T Gossipable] interface {
+	Add(tx T) error
+	Remove(txs ...T)
```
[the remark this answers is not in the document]
tsachiherman: I figured that one of the next steps would be to use this new implementation as a replacement for `avalanchego/vms/txs/mempool/mempool.go`'s Mempool, which is defined as : 
```golang
type Mempool[T Tx] interface {
	Add(tx T) error
	Get(txID ids.ID) (T, bool)
	// Remove [txs] and any conflicts of [txs] from the mempool.
	Remove(txs ...T)

	// Peek returns the oldest tx in the mempool.
	Peek() (tx T, exists bool)

	// Iterate iterates over the txs until f returns false
	Iterate(f func(tx T) bool)

	// Note: dropped txs are added to droppedTxIDs but are not evicted from
	// unissued decision/staker txs. This allows previously dropped txs to be
	// possibly reissued.
	MarkDropped(txID ids.ID, reason error)
	GetDropReason(txID ids.ID) error

	// Len returns the number of txs in the mempool.
	Len() int
}
```

I completely agree that it's enough to pass in the GossipID in this case and think we should change that on the next PR.

network/p2p/gossip/gossip.go
```diff
@@ -285,10 +192,10 @@ func (p *PullGossiper[_]) handleResponse(
 // NewPushGossiper returns an instance of PushGossiper
 func NewPushGossiper[T Gossipable](
 	marshaller Marshaller[T],
-	mempool Set[T],
+	set Set[T],
 	validators p2p.ValidatorSubset,
 	client *p2p.Client,
-	metrics Metrics,
+	metrics *Metrics,
```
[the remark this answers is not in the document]
tsachiherman: the `Metrics` structure contains a lock, which synchronize the access to the `gossipablesLabels`. This allow us to perform the gossip on both the pull and push concurrently, without any collisions ( note : I do expect that there would be collisions.. )

network/p2p/gossip/metrics.go
tsachiherman: We don't *have to* export this label. Beside the code inside the gossip package, we use it in the coreth implementation along with the `ObserveIncomingGossipable` call. We can definitely define a separate label text ( doesn't have to be `duplicate` ). As a starting point, I think that keeping it the same is good enough. We can always change that later on if we see that it's unused/unjustified/etc.

network/p2p/gossip/mempool.go
tsachiherman: If we're going to use this struct to be a replacement for the real mempool, I'd say "yes, let's keep it".
Keep in mind that it also save the `gossipMempool` in platformvm and avm from having the need to format that message.

vms/avm/network/gossip.go
tsachiherman: The two are unrelated, and combining them would result in longer locking time.

`gossipMempoolLock` ensures that we have operational atomicity between the gossip.mempool and the txs/mempool instance.

The `bloomLock` ensures that the bloom filter read/write access is being honored.

network/p2p/gossip/mempool.go
tsachiherman: ok, I'll "port" the error type into here.

network/p2p/gossip/mempool.go
tsachiherman: I think that an interface better define the functionality we want to provide, in contrast to a struct, which is more of a data-type. It's also more aligned with the rest of our codebase. But I'm definitely open to change that if you think it's over-engineering.

network/p2p/gossip/metrics.go
tsachiherman: The "other" bucket would only include transactions that the VMs didn't label. While an ideal VM would label all it's transaction ( in case these were dropped ), we don't have to force it. Having "other" would let us know if this actually happens or not.

network/p2p/gossip/metrics.go
tsachiherman: The alternative is to have sentPushLabels and receivedPushLabels defined as a separate counter vector.

network/p2p/gossip/mempool.go
tsachiherman: let me name the `tx` in the gossip package `gossipable` for now. We can make a change on the next PR. I want to keep the code consistent and make continues improvements.

network/p2p/gossip/mempool.go
joshua-kim: > I think that an interface better define the functionality we want to provide, in contrast to a struct, which is more of a data-type.

Why? This is not how interfaces are used in Go because interfaces are defined implicitly. Interfaces are generally defined in the package that consumes a type - not defined by the package that produces the type

> It's also more aligned with the rest of our codebase.

This is true, but the reason earlier parts of our codebase were not written with idiomatic Go in mind is because at the time the team did not understand Go style. We should not continue writing non-idiomatic Go for the sake of consistency.

network/p2p/gossip/mempool.go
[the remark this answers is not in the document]
joshua-kim: We should just export the struct + return it instead of returning an interface or an unexported struct here. This is because the interface is not needed and unexported structs are very awkward to use from the calling package.

network/p2p/gossip/mempool.go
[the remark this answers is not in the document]
joshua-kim: We can define the api without the common mempool package in mind, because the common mempool package can just write a wrapper over this. There isn't a reason we need to couple this package's api design to the api design of the calling package.

vms/avm/network/gossip.go
joshua-kim: I think this is a pre-mature optimization. Unless we actually see data that indicates that lock contention is something we need to be aware of, I don't think it's worth adding complexity to the code.

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+
+// observeReceivedMessage summarizes the metrics for a given received gossip message and update the metrics accordingly.
+func (m *Metrics) observeReceivedMessage(typeLabel string, gossipables map[ids.ID]int, malformedBytes int, malformedCount int) error {
+	labelBytes := make(map[string]int, len(gossipables))
+	labelCount := make(map[string]int, len(gossipables))
+
+	labelBytes[droppedMalformed] = malformedBytes
+	labelCount[droppedMalformed] = malformedCount
+
+	m.labelsLock.Lock()
```
joshua-kim, line 164: Why don't we just lock at the start of this function and use `defer` to unlock? Unless we have data to indicate that lock contention is an issue I don't think it's worth making the code more confusing to read.

vms/avm/network/gossip.go
```diff
@@ -104,9 +115,15 @@ type gossipMempool struct {
 // transaction to push gossip as well.
 func (g *gossipMempool) Add(tx *txs.Tx) error {
 	txID := tx.ID()
-	if _, ok := g.Mempool.Get(txID); ok {
-		return fmt.Errorf("attempted to issue %w: %s ", mempool.ErrDuplicateTx, txID)
+
+	g.gossipMempoolLock.Lock()
```
joshua-kim, line 119: Can we use `defer` here to unlock?

network/p2p/gossip/mempool.go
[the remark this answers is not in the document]
tsachiherman: ok, done.

network/p2p/gossip/mempool.go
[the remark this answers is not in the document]
tsachiherman: ok, done.

network/p2p/gossip/mempool.go
tsachiherman: given your previous statement where "we define the interface at the destination", I think that I can drop this for now.

vms/avm/network/gossip.go
tsachiherman: ok, done.

network/p2p/gossip/mempool.go
tsachiherman: dropped interface, exported struct.

network/p2p/gossip/metrics.go
tsachiherman: ok, done.

vms/avm/network/gossip.go
tsachiherman: yes

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,49 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+var ErrDuplicateTx = errors.New("duplicate tx")
+
+type Mempool[T Gossipable] struct {
+	Txs     map[ids.ID]bool
```
joshua-kim, line 16: 1. Do we want to expose this?
2. Same comment regarding naming consistency

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,49 @@
...
+	if m.Has(txID) {
+		m.metrics.ObserveIncomingGossipable(txID, DroppedDuplicate)
+		return fmt.Errorf("%w: %s", ErrDuplicateTx, txID)
+	}
+
+	m.Txs[txID] = true
+	return nil
+}
+
+func (m *Mempool[T]) Remove(gossipables ...T) {
```
joshua-kim, line 40: I think this comment from my previous review wasn't addressed, but we can get rid of the variadic param + only take in an id.

network/p2p/gossip/mempool.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+var ErrDuplicateTx = errors.New("duplicate tx")
+
+type Mempool[T Gossipable] struct {
```
joshua-kim, line 15: Since we're exposing a new type out of the package we should be unit testing this.

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// ObserveIncomingGossipable allows external-package mempools to label a custom gossipable drop-reason and have it
+// included in the gossip metrics.
+func (m *Metrics) ObserveIncomingGossipable(txID ids.ID, label string) {
+	m.labelsLock.Lock()
+	defer m.labelsLock.Unlock()
+	_, ok := m.gossipablesLabels[txID]
+	if ok {
+		// we already have this gossipable labeled; don't re-label.
```
joshua-kim, line 149: This seems like a confusing ux to me... why handle de-duping for the caller? If the caller wants de-duped behavior, they can just check the error returned from `Add` before adding their metric. At the very least if we kept this behavior it should be explicitly documented.

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+// only use a single instance of Metrics.
+type Metrics struct {
+	count                   *prometheus.CounterVec
+	bytes                   *prometheus.CounterVec
+	tracking                *prometheus.GaugeVec
+	trackingLifetimeAverage prometheus.Gauge
+	topValidators           *prometheus.GaugeVec
+
+	labelsLock        sync.Mutex
+	gossipablesLabels map[ids.ID]string
```
joshua-kim, line 68: Do we even need this map + lock? Can't we just update the drop metrics similarly to how we're updating the existing metrics?

network/p2p/gossip/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	_, ok := m.gossipablesLabels[txID]
+	if ok {
+		// we already have this gossipable labeled; don't re-label.
+		return
+	}
+	m.gossipablesLabels[txID] = label
+}
+
+// observeReceivedMessage summarizes the metrics for a given received gossip message and update the metrics accordingly.
+func (m *Metrics) observeReceivedMessage(typeLabel string, gossipables map[ids.ID]int, malformedBytes int, malformedCount int) error {
```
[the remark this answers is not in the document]
tsachiherman: given that there is a notion of "receive" vs. "sent", I'd opt into changing this to "updateReceivedMetrics" ( and update the other direction as well )

network/p2p/gossip/metrics.go
tsachiherman: this behavior was explicitly meant for the "Others" group. Given that we don't have it anymore, I'll remove it.

network/p2p/gossip/mempool.go
tsachiherman: added unit tests.

network/p2p/gossip/mempool.go
joshua-kim: I apologize if my wording was confusing but I only explained that we could use a `map[ids.ID]struct{}` to clarify that it's equivalent to our existing `set.Set` implementation, which we should use instead.

tsachiherman: sounds good. updated.

joshua-kim (commented):
I think we're still also looking to get rid of the `droppedNot` label as per https://github.com/ava-labs/avalanchego/pull/3839/files#r2064383603

PR #3842, opened 2025-03-26
Set TestNetwork minCompatibleTime to Fortuna Activation

michaelkaplan13:
## Why this should be merged
Aligns `TestNetwork` compatibility with the `Network` created for full AvalancheGo nodes [here](https://github.com/ava-labs/avalanchego/blob/master/node/node.go#L632). This should be updated in the future when required network upgrades are scheduled.

## How this works
Uses the Fortuna activation time instead of `InitiallyActiveTime` as the `minCompatibleTime` of `TestNetwork` instances. Setting the `minCompatibleTime` to `InitiallyActiveTime` resulted in applications using `TestNetwork` to be unable to connect with nodes running the previous minor AvalancheGo version prior the network upgrade activating.

## How this was tested
N/A

No

### DIFF
--- a/network/test_network.go
+++ b/network/test_network.go
@@ -11,7 +11,6 @@ import (
 	"net/netip"
 	"runtime"
 	"sync"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -26,6 +25,7 @@ import (
 	"github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/staking"
 	"github.com/ava-labs/avalanchego/subnets"
+	"github.com/ava-labs/avalanchego/upgrade"
 	"github.com/ava-labs/avalanchego/utils"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
@@ -210,7 +210,6 @@ func NewTestNetwork(
 	metrics prometheus.Registerer,
 	cfg *Config,
 	router router.ExternalHandler,
-	minCompatibleTime time.Time,
 ) (Network, error) {
 	msgCreator, err := message.NewCreator(
 		logging.NoLog{},
@@ -224,7 +223,7 @@ func NewTestNetwork(
 
 	return NewNetwork(
 		cfg,
-		minCompatibleTime,
+		upgrade.GetConfig(cfg.NetworkID).FortunaTime,
 		msgCreator,
 		metrics,
 		log,
--- a/network/test_network.go
+++ b/network/test_network.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 	"runtime"
 	"sync"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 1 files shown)


cam-schultz (commented):
Could we instead remove minCompatibilityTime from `NewNetwork` altogether, and use FortunaTime in all instances? That way the activation time used by the test network is covered by general `Network` upgrade tests (i.e. e2e tests that run full nodes).

michaelkaplan13:
> Could we instead remove minCompatibilityTime from `NewNetwork` altogether, and use FortunaTime in all instances? That way the activation time used by the test network is covered by general `Network` upgrade tests (i.e. e2e tests that run full nodes).

I debated that, or adding the `minCompatibleTime` to the `upgrade.Config` somehow, so that nothing in the `Network` code needs to be changed for future network upgrades.

I think the `minCompatibleTime` can be defined as the time of the next scheduled network upgrade activation or, if there are no scheduled upgrades in the future, the time the last network upgrade activated.

Either way, I agree minimizing the number of places updates need to be made should be the goal in order to reduce the chance that any are missed in the future.

michaelkaplan13:
> Could we instead remove minCompatibilityTime from `NewNetwork` altogether, and use FortunaTime in all instances? That way the activation time used by the test network is covered by general `Network` upgrade tests (i.e. e2e tests that run full nodes).

I updated the PR to this approach now. I agree it's less error prone since there's only one place where the minCompatibleTime is defined, and should be covered in the E2E tests. The downside is needing to pass the full `upgrade.Config` to `NewNetwork`, and the minCompatibleTime definition being embedded in the implementation is harder to find.

I'm pretty indifferent between the options myself really, open to everyone's input.

cam-schultz (approved):

network/network.go
```diff
@@ -175,7 +176,7 @@ type network struct {
 // NewNetwork returns a new Network implementation with the provided parameters.
 func NewNetwork(
 	config *Config,
-	minCompatibleTime time.Time,
+	upgradeConfig upgrade.Config,
```
joshua-kim, line 179: This change in the to depend on `upgrade` looks unrelated to the intent of this PR to me... can we remove changes from `network.go` and `network_test.go` from the diff?

michaelkaplan13: See [Cam's comment above](https://github.com/ava-labs/avalanchego/pull/3842#issuecomment-2755807516) for the context here. We can do this just by updating `test_network.go` (as we had originally), but that would make it such that multiple places need to be updated for every network upgrade to ensure compatibility, which is more error prone. This approach makes it so that the `minCompatibleTime` only needs to be updated in a single place, but requires then passing the full `upgrade.Config` to `NewNetwork`.

Open to whichever is preferred on your side.

joshua-kim: From the perspective of `network` it just seems like an abstraction that it doesn't need to know about - `upgrade.Config` defines all fork times but the networking layer only needs to know when a peer isn't compatible.

If the problem we're trying to solve is to make sure that all tests use the same network timestamp, I don't know if this solves our problem because passing in the wrong timestamp seems to be as error prone as passing in the wrong config

michaelkaplan13: I switched back to passing the `minCompatibleTime`, which also minimize the diff. We'll just need to remember to update it in `TestNetwork` alongside any bumps for network upgrades made in `node.go`. 🫡

joshua-kim (approved):

StephenButtolph (approved):

PR #3843, opened 2025-03-26
Add L1 validators to getCurrentValidators response

ceyonur:
## Why this should be merged

L1 Validators  are not supported in `client.GetCurrentValidators` response. This PR adds `ClientL1Validator` to `ClientPermissionlessValidator` so that it can show L1 validators fields in the response. 

## How this works

This pull request includes several changes to the `vms/platformvm` package to enhance the representation and handling of L1 validators. The most important changes involve adding new types and refactoring existing ones to support the new L1 validator structure.

### Enhancements to L1 Validator Representation:

* Added `APIL1Validator` and `BaseL1Validator` types to represent L1 validators with additional fields such as `PublicKey`, `RemainingBalanceOwner`, `DeactivationOwner`, `MinNonce`, and `Balance`. (`vms/platformvm/api/static_service.go`)
* Updated `ClientL1Validator` type to include fields for `ValidationID`, `RemainingBalanceOwner`, `DeactivationOwner`, `MinNonce`, and `Balance`. (`vms/platformvm/client_permissionless_validator.go`)

### Refactoring and Code Simplification:

* Refactored the `GetL1Validator` function to handle optional fields and convert them appropriately. (`vms/platformvm/client.go`) [[1]](diffhunk://#diff-07adbc0edb62e0f0bd7cbcfc65a4e6d8be518002233ca81a8410b0e0ba5bbbbcR379-R387) [[2]](diffhunk://#diff-07adbc0edb62e0f0bd7cbcfc65a4e6d8be518002233ca81a8410b0e0ba5bbbbcL393-R405)
* Removed the deprecated `RewardOwner` field from `PermissionlessValidator` and replaced it with `ValidationRewardOwner` and `DelegationRewardOwner`. (`vms/platformvm/api/static_service.go`)

### API and Service Updates:

* Updated the `convertL1ValidatorToAPI` function to use the new `platformapi.APIL1Validator` type and handle the conversion of nested fields. (`vms/platformvm/service.go`)

## How this was tested

Existing tests should cover the service changes. We dont have any client UT, locally tested it.

Updated


vms/platformvm/api/static_service.go
```diff
@@ -72,16 +73,11 @@ func (utxo UTXO) Compare(other UTXO) int {
...
+// Staker is the representation of a staker sent via APIs.
 // [TxID] is the txID of the transaction that added this staker.
-// [Amount] is the amount of tokens being staked.
-// [StartTime] is the Unix time when they start staking
 // [Endtime] is the Unix time repr. of when they are done staking
 // [NodeID] is the node ID of the staker
-// [Uptime] is the observed uptime of this staker
+// [Weight] is the validator weight (stake) when sampling validators
 type Staker struct {
 	TxID      ids.ID      `json:"txID"`
```
ceyonur, line 82: TxID and EndTime will have zero values for L1 validators. I did not want to break any client implementation. If we think it's weird to show these for L1 validators I can set them to pointers. We can even break `Staker` into `BaseStaker` and `Staker` and move those fields there.

vms/platformvm/client.go
```diff
@@ -374,6 +376,15 @@ func (c *client) GetL1Validator(
...
 
+	var minNonce uint64
+	if res.MinNonce != nil {
+		minNonce = uint64(*res.MinNonce)
+	}
+	var balance uint64
+	if res.Balance != nil {
+		balance = uint64(*res.Balance)
+	}
+
```
meaghanfitzgerald, line 387: Is there a reason this wasn't also applied to the weight? I'm assuming you did this to check vars that are meant to change/be updated

ceyonur: This is to ensure they're not nil before de-referencing. `Weight` is not a pointer [here](https://github.com/ava-labs/avalanchego/pull/3843/files/5d1f16f5bf563bfacf860014896b18891cc6f439#diff-bc67b377b06e3e87764d1efb1a1dde6337ed4b902ceb0f0adba9b7257397da18R102) thus cannot be nil.

meaghanfitzgerald: [no action required, just curious] why is weight stored as a non-pointer?

meaghanfitzgerald:
afaict this pr does not change the response object for [getCurrentValidators](https://build.avax.network/docs/api-reference/p-chain/api#platformgetcurrentvalidators), lgtm

vms/platformvm/client.go
ceyonur: Weight is common in all validator types (stakers vs l1 validators), so they will always be present in json. By using a pointer we can mark it them as being "nilable" in json representations. Otherwise they will be always shown with their zero values.

vms/platformvm/api/static_service.go
qdm12: Maybe just add a deprecated comment for TxID?

vms/platformvm/api/static_service.go
```diff
@@ -100,13 +96,33 @@ type Owner struct {
...
+	Balance *json.Uint64 `json:"balance,omitempty"`
+}
+
 // PermissionlessValidator is the repr. of a permissionless validator sent over
 // APIs.
 type PermissionlessValidator struct {
 	Staker
-	// Deprecated: RewardOwner has been replaced by ValidationRewardOwner and
-	//             DelegationRewardOwner.
-	RewardOwner *Owner `json:"rewardOwner,omitempty"`
```
qdm12, line 109: Should this be kept around for retro-compatibility, even-though deprecated?

qdm12 (approved):

meaghanfitzgerald (approved):
> afaict this pr does not change the response object for [getCurrentValidators](https://build.avax.network/docs/api-reference/p-chain/api#platformgetcurrentvalidators), lgtm

still lgtm

vms/platformvm/api/static_service.go
ceyonur: These will still apply for non L1 stakers (subnets/primary validators)

vms/platformvm/api/static_service.go
ceyonur: It's been deprecated for 3 years, I think it was already long due that we should have removed it.

StephenButtolph (commented):
1 comment then lgtm

StephenButtolph (approved):

PR #3794, opened 2025-03-13
[tmpnet] Enable monitoring of nodes running in kube

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3945
- #3947
- #3973
- #3615
- **>>>>>>** #3794 **<<<<<<** 

## Why this should be merged

- Ensures nodes deployed to kube have labels and annotations to enable the same labeling as nodes running a processes. 
- Enables deployment of prometheus and promtail to kube to ensure collection of logs and metrics from nodes deployed on the cluster.

## How this works

- adds labels and annotations to node pods 
- enables deployment of the collectors via `tmpnetctl start-kind-cluster --start-metrics-collector --start-logs-collector` 
- updates the kube e2e ci job to check that logs and metrics were collected 
- defines collector deployment in yaml for compatibility with argocd usage
 
## How this was tested

CI configures a metrics check to verify that logs and metrics were collected

## Need to be documented in RELEASES.md?

N/A

## TODO

- [x] merge https://github.com/ava-labs/avalanchego/pull/3615
- [x] Verify that metrics and logs are collected from nodes running in kubernetes
```
[DeferCleanup (Suite)] 
/home/runner/work/avalanchego/avalanchego/tests/fixture/e2e/ginkgo_test_context.go:96

  Timeline >>
   [05-28|19:02:40.088] INFO tmpnet/check_monitoring.go:79 checking if logs exist {"url": "https://loki-poc.avax-dev.network/", "query": "sum(count_over_time({network_uuid=\"27080d29-7152-4a5c-a883-af4b6a47700b\"}[1h]))"}
   [05-28|19:02:40.507] INFO tmpnet/check_monitoring.go:45 collected count is non-zero {"type": "logs", "count": 381224}
  << Timeline
------------------------------
[DeferCleanup (Suite)] PASSED [0.291 seconds]
[DeferCleanup (Suite)] 
/home/runner/work/avalanchego/avalanchego/tests/fixture/e2e/ginkgo_test_context.go:96

  Timeline >>
[...]

PR #3857, opened 2025-03-31
[tmpnet] Provide genesis, subnet and chain config via content flags

maru-ava:
## PR Chain: tmpnet+kube

- #3854
- **>>>>>>** #3857 **<<<<<<**
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

Previously supplied genesis, subnet and chain config to nodes via `--*-file-content` flags pointing to files on disk. Switching to use the equivalent `--*-content` flags simplifies configuring nodes deployed to kube where configuration via files is more complicated.

To support this, Node.Flags is now used only for generated configuration like keys and user-supplied node-specific config. Network defaults and automatic configuration (like bootstrap IPs/IDs) are not included. Instead, flags.json for each node is generated each time a node is started to ensure that the configuration is up-to-date.

## How this was tested

CI, local validation of `tmpnetctl start-network`

N/A

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -66,6 +66,8 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 			Owner: "avalanchego-e2e",
 			DefaultFlags: tmpnet.FlagsMap{
 				config.UpgradeFileContentKey: upgradeBase64,
+				// Ensure a min stake duration compatible with testing staking logic
+				config.MinStakeDurationKey: tmpnet.DefaultMinStakeDuration,
 			},
 			Nodes:   nodes,
 			Subnets: subnets,
--- a/tests/fixture/tmpnet/defaults.go
+++ b/tests/fixture/tmpnet/defaults.go
@@ -60,7 +60,6 @@ func DefaultTmpnetFlags() FlagsMap {
 		config.LogDisplayLevelKey: logging.Off.String(), // Display logging not needed since nodes run headless
 		config.LogLevelKey:        logging.Debug.String(),
 		// Specific to e2e testing
-		config.MinStakeDurationKey:           DefaultMinStakeDuration.String(),
 		config.ProposerVMUseCurrentHeightKey: true,
 		// Reducing this from the 1s default speeds up tx acceptance
 		config.ProposerVMMinBlockDelayKey: "0s",


tests/fixture/tmpnet/node.go
```diff
@@ -154,6 +155,11 @@ func ReadNodes(networkDir string, includeEphemeral bool) ([]*Node, error) {
 			continue
 		}
 
+		// Ensure the node ID is set
+		if err := node.EnsureNodeID(); err != nil {
```
qdm12, line 159: nit comment feels unneeded

```suggestion
		if err := node.EnsureNodeID(); err != nil {
```

tests/fixture/tmpnet/node.go
```diff
@@ -154,6 +155,11 @@ func ReadNodes(networkDir string, includeEphemeral bool) ([]*Node, error) {
...
+			return nil, err
```
qdm12, line 160: nit wrap the error

```suggestion
			return nil, fmt.Errorf("checking node id: %w", err)
```

qdm12 (approved):
Nice 👍 Just leaving a few nitpicky comments which you can ignore as well 😉 !

tests/fixture/tmpnet/flags.go
```diff
@@ -30,14 +30,20 @@ func ReadFlagsMap(path string, description string) (FlagsMap, error) {
 	return flagsMap, nil
 }
 
+// SetDefault ensures the effectiveness of a flag override by only
+// setting a value supplied whose key is not already explicitly set.
+func (f FlagsMap) SetDefault(key string, value any) {
```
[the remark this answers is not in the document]
qdm12: Isn't it enough to have SetDefaults exported already?

tests/fixture/tmpnet/network.go
```diff
@@ -793,11 +769,11 @@ func (n *Network) CreateSubnets(ctx context.Context, log logging.Logger, apiURI
 			return err
 		}
 
-		// Persist the chain configuration
-		if err := subnet.Write(n.GetSubnetDir(), n.GetChainConfigDir()); err != nil {
+		// Persist the subnet
```
qdm12, line 772: nit maybe a tiny bit clearer:

```suggestion
		// Persist the subnet tmpnet config
```

Persisting/writing subnet is a bit odd I think 🤔 Maybe I'm being too nitpicky as well!

tests/fixture/tmpnet/network.go
```diff
@@ -888,6 +864,129 @@ func (n *Network) GetPluginDir() (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal genesis: %w", err)
+	}
+	return base64.StdEncoding.EncodeToString(bytes), nil
+}
+
+// GetSubnetConfigContent returns the base64-encoded and
+// JSON-marshaled map of subnetID to subnet configuration.
+func (n *Network) GetSubnetConfigContent() (string, error) {
+	subnetConfigs := maps.Clone(n.PrimarySubnetConfigs)
```
qdm12, line 880: `n.PrimarySubnetConfigs` seems to never be set, I suppose this is coming in an upcoming PR? Might be worth moving that field to the relevant PR where it's starting to be used? 🤔

tests/fixture/tmpnet/network.go
```diff
@@ -888,6 +864,129 @@ func (n *Network) GetPluginDir() (string, error) {
...
+func (n *Network) GetGenesisFileContent() (string, error) {
+	bytes, err := json.Marshal(n.Genesis)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal genesis: %w", err)
+	}
+	return base64.StdEncoding.EncodeToString(bytes), nil
+}
+
+// GetSubnetConfigContent returns the base64-encoded and
+// JSON-marshaled map of subnetID to subnet configuration.
```
qdm12, line 878: nit `and` is a bit confusing imo (same below)

```suggestion
// GetSubnetConfigContent returns the base64-encoded of the
// JSON-marshaled map of subnetID to subnet configuration.
```

tests/fixture/tmpnet/network_config.go
```diff
@@ -39,9 +41,6 @@ func (n *Network) Write() error {
 	if err := n.writeGenesis(); err != nil {
```
felipemadero, line 41: Genesis should not be wrote for fuji/mainnet

felipemadero (commented):
Overall LGTM. Just a naming nit and a couple of suggestions which scope anyway may belong to a different PR.

tests/fixture/tmpnet/network_config.go
felipemadero: If may be a problem after reading it again, for when some initialization logic checks genesis for not being null. Probably not the case atm but maybe better to avoid the future  case.

felipemadero (approved):

tests/fixture/tmpnet/network_config.go
felipemadero: Made https://github.com/ava-labs/avalanchego/pull/3868 to cover for this

PR #3858, opened 2025-04-02
chore: fix `ErrHelp` check

santamasa:
## **Why this should be merged**  
The current implementation uses `errors.Is(err, pflag.ErrHelp)`, but `ErrHelp` from `spf13/pflag` is not wrapped—it's a direct error value. This means the check might not work as intended. Updating it to `err == pflag.ErrHelp` ensures correctness.  

## **How this works**  
The change replaces `errors.Is(err, pflag.ErrHelp)` with `err == pflag.ErrHelp`, aligning with how `spf13/pflag` defines `ErrHelp`. This ensures that the check properly detects when the help flag is triggered.  

## **How this was tested**  
Manually tested by running the application with `--help` and verifying that it exits correctly without errors.
No other behavior is affected.  

## **Need to be documented in RELEASES.md?**  
No, this is a minor internal fix that does not affect the user-facing behavior.

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -21,7 +21,7 @@ func main() {
 	fs := config.BuildFlagSet()
 	v, err := config.BuildViper(fs, os.Args[1:])
 
-	if errors.Is(err, pflag.ErrHelp) {
+	if err == pflag.ErrHelp {
 		os.Exit(0)
 	}
 


StephenButtolph:
This PR doesn't seem to actually fix a bug.

PR #3863, opened 2025-04-06
[tooling] Add support for the Task (go-task) task runner

maru-ava:
## Why this should be merged

This repo has a lot of tooling but previously had no easy entrypoint other than its github action workflows. This complicated discoverability and CI reproduceability.

This change adds support for invoking every command intended to be run in CI via the [Task](https://taskfile.dev) task runner. Since `Task` is written in golang, it is easily installable without additional dependency.

The intention is to ensure that every development task can be invoked with `scripts/run_task.sh [task]` and that the logic of task execution - even for tasks primarily intended for usage in CI - be defined in a given task rather than in a github action workflow. 

Alternatives considered:
 - [GNU Make](https://www.gnu.org/software/make/) is a build system, and has a lot more complexity and features than we need for a task runner.
 - [Just](https://github.com/casey/just) has a similar featureset to Task but since it is written in Rust it isn't possible to `go run` on a system without it. `task` also supports parallel execution which may prove useful in the future.

## How this was tested

CI, locally for tasks that aren't tested by CI (e.g. build-image)

N/A

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,6 @@
 {
   # To use:
-  #  - install nix: `./scripts/run_task.sh install_nix`
+  #  - install nix: `./scripts/run_task.sh install-nix`
   #  - run `nix develop` or use direnv (https://direnv.net/)
   #    - for quieter direnv output, set `export DIRENV_LOG_FORMAT=`
 
--- a/.envrc
+++ b/.envrc
@@ -1,5 +1,5 @@
 # Requires nix to be installed and available in the PATH.
-# To install nix: `./scripts/run_task.sh install_nix`
+# To install nix: `./scripts/run_task.sh install-nix`
 use flake
 
 # Repo-local commands like ginkgo and tmpnetctl


.github/workflows/ci.yml
```diff
@@ -33,32 +33,28 @@ jobs:
         shell: bash
         if: matrix.os == 'windows-2022'
         run: echo "TIMEOUT=240s" >> "$GITHUB_ENV"
-      - name: build_test
+      - name: test-unit
```
yacovm, line 36: why not unit-test ?

scripts/tests.e2e.sh
```diff
@@ -20,10 +20,6 @@ fi
 # the instructions to build non-portable BLST.
 source ./scripts/constants.sh
 
-# Enable subnet testing by building xsvm
```
yacovm, line 23: why is this removed?

flake.nix
```diff
@@ -33,6 +33,9 @@
         default = pkgs.mkShell {
           # The Nix packages provided in the environment
           packages = with pkgs; [
+            # Task runner
+            go-task
```
yacovm, line 37: so we have nix install go-task and go-task install nix?

scripts/actionlint.sh
```diff
@@ -3,3 +3,22 @@
...
 go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.1 "${@}"
+
+echo "Checking use of scripts/* in GitHub Actions workflows..."
+SCRIPT_USAGE=
+for file in .github/workflows/*.{yml,yaml}; do
+  # Skip if no matches found (in case one of the extensions doesn't exist)
+  [[ -f "$file" ]] || continue
+
+  # Search for scripts/* except for scripts/run_task.sh
+  MATCHES=$(grep -H -n -P "scripts/(?!run_task\.sh)" "$file" || true)
```
yacovm, line 14: so this essentially ensures we only use `scripts/run_task.sh` and nothing else in `scripts`?

tsachiherman (approved):
looks good. One thing that I'd rather seeing is having the `task` tool itself living within our repo ( or a clone that we maintain, which can be used across multiple repos )

yacovm:
> > looks good. One thing that I'd rather seeing is having the `task` tool itself living within our repo ( or a clone that we maintain, which can be used across multiple repos )
> 
> Why would we want to maintain the tool in our repo? We depend on many other tools that live outside our repo, I'm not sure what makes `task` special. `scripts/run_task.sh` ensures that it can be executed even if not installed, and people have the option of manually installing it in their environment or invoking from within a nix dev shell.

I think if we're worried about the task repository compromised, we can always fork it to our organization and then carefully review the upstream commits and upgrade the forked repository.

flake.nix
```diff
@@ -1,6 +1,6 @@
 {
   # To use:
-  #  - install nix: https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#install-nix
+  #  - install nix: `./scripts/run_task.sh install_nix`
```
StephenButtolph, line 3:
```
task: Task "install_nix" does not exist
```

StephenButtolph (approved):
pretty cool

PR #3869, opened 2025-04-09
[tmpnet] Update rpc version check to tolerate usage of `go run`

maru-ava:
## Why this should be merged

When avalanchego is invoked with `go run`, the go toolchain may emit informational messages that corrupt the `--version-json` output. Being tolerant of such output enables downstream repos to use tmpnet with avalanchego invoked with `go run`.

## How this was tested

Manually tested with https://github.com/ava-labs/hypersdk/pull/2015

N/A

### DIFF
--- a/tests/fixture/tmpnet/network.go
+++ b/tests/fixture/tmpnet/network.go
@@ -1027,7 +1027,7 @@ func getRPCVersion(log logging.Logger, command string, versionArgs ...string) (u
 
 	// Ignore output before the opening brace to tolerate the case of a command being invoked
 	// with `go run` and the go toolchain emitting diagnostic logging before the version output.
-	if idx := bytes.IndexByte(output, '{'); idx != -1 {
+	if idx := bytes.IndexByte(output, '{'); idx > 0 {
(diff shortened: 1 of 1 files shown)


tests/fixture/tmpnet/network.go
```diff
@@ -1017,12 +1018,23 @@ type RPCChainVMVersion struct {
...
 		return 0, fmt.Errorf("command %q failed with output: %s", command, output)
 	}
+
+	// Ignore output before the opening brace to tolerate the case of a command being invoked
+	// with `go run` and the go toolchain emitting diagnostic logging before the version output.
+	if idx := bytes.IndexByte(output, '{'); idx != -1 {
+		log.Info("ignoring leading bytes of JSON version output in advance of opening `{`",
+			zap.String("command", command),
+			zap.String("ignoredLeadingBytes", string(output[:idx])),
+		)
```
StephenButtolph, line 1034: I think this will log even if `idx == 0` (aka there are no leadingBytes). Is that expected? Perhaps the if should be `idx > 0` rather than `idx != -1`

StephenButtolph (approved):

yacovm (approved):

PR #3867, opened 2025-04-08
[tmpnet] Refactor runtime configuration in preparation for kube

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- **>>>>>>** #3867 **<<<<<<**
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

Adding the kube runtime for tmpnet requires being able to configure it via cli args. This PR refactors the existing runtime configuration to better support the introduction of the kube runtime configuration.

## How this was tested

CI, manual use of tmpnetctl

## Need to be documented in RELEASES.md?

N/A

## TODO

- [x] Merge https://github.com/ava-labs/avalanchego/pull/3870

PR #3870, opened 2025-04-09
[tmpnet] Ensure Node has a reference to Network

maru-ava:
## PR Chain: tmpnet+kube

- #3854
- #3857
- **>>>>>>** #3870 **<<<<<<**
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

# Why this should be merged

Previously tmpnet.Node was serialized with a subset of tmpnet.Network fields like UUID and Owner. Intead, tmpnet.Node gets a reference to its owning tmpnet.Network on read or start to ensure it has full access to all network details.

## How this was tested

CI

N/A

### DIFF
--- a/tests/fixture/tmpnet/network.go
+++ b/tests/fixture/tmpnet/network.go
@@ -542,7 +542,7 @@ func (n *Network) RestartNode(ctx context.Context, log logging.Logger, node *Nod
 // Stops all nodes in the network.
 func (n *Network) Stop(ctx context.Context) error {
 	// Target all nodes, including the ephemeral ones
-	nodes, err := ReadNodes(n.Dir, true /* includeEphemeral */)
+	nodes, err := ReadNodes(n, true /* includeEphemeral */)
 	if err != nil {
 		return err
 	}
@@ -585,11 +585,8 @@ func (n *Network) Restart(ctx context.Context, log logging.Logger) error {
 // no action will be taken.
 // TODO(marun) Reword or refactor to account for the differing behavior pre- vs post-start
 func (n *Network) EnsureNodeConfig(node *Node) error {
-	// Ensure nodes can label their metrics with the network uuid
-	node.NetworkUUID = n.UUID
-
-	// Ensure nodes can label metrics with an indication of the shared/private nature of the network
-	node.NetworkOwner = n.Owner
+	// Ensure the node has access to network configuration
+	node.network = n
 
 	if err := node.EnsureKeys(); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


tests/fixture/tmpnet/network.go
```diff
@@ -542,7 +542,7 @@ func (n *Network) RestartNode(ctx context.Context, log logging.Logger, node *Nod
 // Stops all nodes in the network.
 func (n *Network) Stop(ctx context.Context) error {
 	// Target all nodes, including the ephemeral ones
-	nodes, err := ReadNodes(n.Dir, true /* includeEphemeral */)
+	nodes, err := ReadNodes(n, true /* includeEphemeral */)
```
qdm12, line 545: nit you could use a named constant instead of a comment. I like the attention to documenting, because it is indeed not clear what `true` is without using an IDE. I tend to use constants for all these arguments, sometimes even nil variables (i.e. `someName := []byte(nil)`) to have a better call documentation. The worst is having a function call `f(nil, false, nil, true, 0)` and have trouble matching each of the argument by position, this is so cognitively unneeded imo. Anyway, rant over 😄 

```suggestion
	const includeEphemeral = true
	nodes, err := ReadNodes(n, includeEphemeral)
```

Same suggestion for other calls of ReadNodes

tests/fixture/tmpnet/network.go
```diff
@@ -585,11 +585,8 @@ func (n *Network) Restart(ctx context.Context, log logging.Logger) error {
 // no action will be taken.
 // TODO(marun) Reword or refactor to account for the differing behavior pre- vs post-start
 func (n *Network) EnsureNodeConfig(node *Node) error {
-	// Ensure nodes can label their metrics with the network uuid
-	node.NetworkUUID = n.UUID
-
-	// Ensure nodes can label metrics with an indication of the shared/private nature of the network
-	node.NetworkOwner = n.Owner
+	// Ensure the node has access to network configuration
```
qdm12, line 588: nit maybe a `the` missing?

```suggestion
	// Ensure the node has access to the network configuration
```

tests/fixture/tmpnet/node.go
```diff
@@ -89,6 +79,9 @@ type Node struct {
 
 	// Initialized on demand
 	runtime NodeRuntime
+
+	// Intended to be set by the network
```
qdm12, line 83: nit such that it links up to the relevant functions setting the network field

```suggestion
	// Intended to be set by [Network.EnsureNodeConfig] or [ReadNodes]
```

tests/fixture/tmpnet/network.go
```diff
@@ -585,11 +585,8 @@ func (n *Network) Restart(ctx context.Context, log logging.Logger) error {
...
+	node.network = n
```
qdm12, line 589: maybe a bit out of scope, but shouldn't the network field be set only if it's unset? The name of the function EnsureNodeConfig suggests this 🤔

qdm12 (approved):

darioush (approved):

joshua-kim (approved):

PR #3874, opened 2025-04-10
Set `subnets.Config.ConsensusParameters` to serialize with omitempty

maru-ava:
## Why this should be merged

This results in serializing with a zero-value of nil (valid) rather than a zero value of `snowball.Parameters` with zero-value fields (invalid). This change allows tmpnet to use `subnet.Config` to define configuration intended to be serialized.

## How this was tested

CI

N/A

### DIFF
--- a/subnets/config.go
+++ b/subnets/config.go
@@ -24,7 +24,7 @@ type Config struct {
 	// AllowedNodes is the set of node IDs that are explicitly allowed to connect to this Subnet when
 	// ValidatorOnly is enabled.
 	AllowedNodes        set.Set[ids.NodeID] `json:"allowedNodes"        yaml:"allowedNodes"`
-	ConsensusParameters snowball.Parameters `json:"consensusParameters" yaml:"consensusParameters",omitempty`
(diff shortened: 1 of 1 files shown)


joshua-kim (approved):

subnets/config.go
```diff
@@ -24,7 +24,7 @@ type Config struct {
 	// AllowedNodes is the set of node IDs that are explicitly allowed to connect to this Subnet when
 	// ValidatorOnly is enabled.
 	AllowedNodes        set.Set[ids.NodeID] `json:"allowedNodes"        yaml:"allowedNodes"`
-	ConsensusParameters snowball.Parameters `json:"consensusParameters" yaml:"consensusParameters"`
+	ConsensusParameters snowball.Parameters `json:"consensusParameters" yaml:"consensusParameters",omitempty`
```
StephenButtolph, line 27:
```suggestion
	ConsensusParameters snowball.Parameters `json:"consensusParameters" yaml:"consensusParameters,omitempty"`
```
?

StephenButtolph (approved):

PR #3878, opened 2025-04-11
[tooling] Add scripts that build+run tools and put them in the path

maru-ava:
## Why this should be merged

`go run` turns out to be slower than installing and running tooling. This change adds such scripts for ginkgo, avalanchego and tmpnetctl and symlinks them to the bin path so that they are in the path with direnv. This ensures executed binaries are always up-to-date when developing locally.

## How this was tested

All local

N/A

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -3,11 +3,8 @@ use flake
 # Repo-local commands like ginkgo and tmpnetctl
 PATH_add bin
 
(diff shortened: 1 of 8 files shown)


StephenButtolph (approved):
lgtm

joshua-kim (approved):

PR #3880, opened 2025-04-11
Simplify P-Chain block has changes check

StephenButtolph:
## Why this should be merged

This removes some special case handling that was added for the Fortuna activation.

## How this works

Because all networks are post Fortuna activation now, there is no need to enforce additional pre-fortuna verification rules.

## How this was tested

Updated unit tests.

No.

### DIFF
--- a/vms/platformvm/block/executor/verifier_test.go
+++ b/vms/platformvm/block/executor/verifier_test.go
@@ -1365,7 +1365,6 @@ func TestDeactivateLowBalanceL1ValidatorBlockChanges(t *testing.T) {
 			currentFork:       upgradetest.Etna,
 			durationToAdvance: time.Second,
 			networkID:         constants.UnitTestID,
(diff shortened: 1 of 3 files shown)


vms/platformvm/block/executor/verifier_test.go
```diff
@@ -1365,7 +1365,6 @@ func TestDeactivateLowBalanceL1ValidatorBlockChanges(t *testing.T) {
 			currentFork:       upgradetest.Etna,
 			durationToAdvance: time.Second,
 			networkID:         constants.UnitTestID,
-			expectedErr:       ErrStandardBlockWithoutChanges,
```
StephenButtolph, line 1368: The code now allows these block to be issued pre-Fortuna (as they should have been originally). This isn't a breaking change to the network because no networks will verify new (blocks that weren't already accepted) pre-Fortuna blocks.

joshua-kim (approved):

PR #3877, opened 2025-04-11
[tmpnet] Switch back to using maps for subnet config

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- **>>>>>>** #3877 **<<<<<<**
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

Using subnet.Config to define subnet configuration initially seemed like a good idea - typing for the win - but went down in flames due to a combination the semantics of JSON marshaling and how subnet config defaults are set by the node. So, back to maps we go.

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

PR #3871, opened 2025-04-09
[tmpnet] Unify start network flag usage between e2e and tmpnetctl

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- **>>>>>>** #3871 **<<<<<<**
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

Cleans up flag usage and configuration network start to ensure consistency and simplify maintenance across e2e and tmpnetctl.

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

## TODO

 - [x] Merge https://github.com/ava-labs/avalanchego/pull/3867

PR #3881, opened 2025-04-11
[tmpnet] Avoid serializing the node data directory

maru-ava:
## PR Chain: tmpnet+kube

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- **>>>>>>** #3881 **<<<<<<**
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

This is intended to simplify usage by CLI. It also moves responsibility for setting the data dir to the runtime since the dir will differ between the node (local fs) and kube runtimes (pvc).

## How this was tested

CI, local testing of tmpnetctl

N/A

## TODO

- [x] Merge https://github.com/ava-labs/avalanchego/pull/3871

### DIFF
--- a/tests/antithesis/init_db.go
+++ b/tests/antithesis/init_db.go
@@ -45,7 +45,7 @@ func initBootstrapDB(network *tmpnet.Network, destPath string) error {
 	}
 
 	// Copy the db state from the bootstrap node to the compose volume path.
-	sourcePath := filepath.Join(network.Nodes[0].GetDataDir(), "db")
+	sourcePath := filepath.Join(network.Nodes[0].DataDir, "db")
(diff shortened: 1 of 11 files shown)


felipemadero (approved):

michaelkaplan13 (approved):
The code all LGTM. 

Just one high level understanding question on the intended usage: Previously, a `Node`'s data directory was included in its `Flags`, which would get serialized when `Write` was called for a given `Network`, and now that's not the case. How does that simplify the CLI's usage of tmpnet? Is it that networks/nodes can be restarted using different data directories than before?

PR #3899, opened 2025-04-18
Move database creation to database factory package

ceyonur:
## Why this should be merged

It moves database creation function to `database/factory` pkg to be used from external pkgs/modules (like subnet-evm)

### Database Configuration Refactor:

* **New `database/factory` package**: Introduced a new `DatabaseConfig` struct and a `NewDatabase` function in `database/factory/factory.go`. This function centralizes the creation of LevelDB, MemDB, and PebbleDB instances, as well as their associated wrappers (e.g., corruptable and meter DBs).

### Code Simplification in `node` Package:

* **Refactored database initialization**: The `initDatabase` method in `node/node.go` now uses the `NewDatabase` function from the `database/factory` package, reducing duplicated logic and improving maintainability. 

## How this was tested

e2e tests should cover this + locally

No


database/factory/factory.go
```diff
@@ -0,0 +1,97 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package factory
```
ceyonur, line 4: I found it most convenient to put it here. Importing whole node pkg for the database factory seemed too much.

database/factory/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ava-labs/avalanchego/database/corruptabledb"
+	"github.com/ava-labs/avalanchego/database/leveldb"
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/avalanchego/database/meterdb"
+	"github.com/ava-labs/avalanchego/database/pebbledb"
+	"github.com/ava-labs/avalanchego/database/versiondb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+type DatabaseConfig struct {
```
StephenButtolph, line 20: Should this just be 
```suggestion
type Config struct {
```
?

database/factory/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	// Path to config file
+	Config []byte `json:"-"`
+}
+
+// NewDatabase creates a new database instance based on the provided configuration.
+// It supports LevelDB, MemDB, and PebbleDB as database types.
+// It also wraps the database with a corruptable DB and a meter DB.
+// [dbMetricsPrefix] is used to create a new metrics registerer for the database.
+// [meterDBRegName] is used to create a new metrics registerer for the meter DB.
+func NewDatabase(dbConfig DatabaseConfig, gatherer metrics.MultiGatherer, logger logging.Logger, dbMetricsPrefix, meterDBRegName string) (database.Database, error) {
```
StephenButtolph, line 39:
```suggestion
func New(config DatabaseConfig, gatherer metrics.MultiGatherer, logger logging.Logger, metricsPrefix, meterDBRegName string) (database.Database, error) {
```

node/node.go
```diff
@@ -757,60 +754,23 @@ func (n *Node) Dispatch() error {
...
-		return fmt.Errorf(
-			"db-type was %q but should have been one of {%s, %s, %s}",
-			n.Config.DatabaseConfig.Name,
-			leveldb.Name,
-			memdb.Name,
-			pebbledb.Name,
-		)
+		dbFolderName = "db"
 	}
+	n.Config.DatabaseConfig.Path = filepath.Join(n.Config.DatabaseConfig.Path, dbFolderName)
```
StephenButtolph, line 768: I don't know that there is a better option, but it is pretty gross that we have to modify the config here.

ceyonur: changed it

database/factory/factory.go
```diff
@@ -0,0 +1,85 @@
...
+// It also wraps the database with a corruptable DB and a meter DB.
+// [dbName] is the name of the database.
+// [dbPath] is the path to the database folder.
+// [readOnly] indicates if the database should be read-only.
+// [dbConfig] is the database configuration in JSON format.
+// [dbMetricsPrefix] is used to create a new metrics registerer for the database.
+// [meterDBRegName] is used to create a new metrics registerer for the meter DB.
+func New(name string, path string, readOnly bool, config []byte,
+	gatherer metrics.MultiGatherer, logger logging.Logger, metricsPrefix, meterDBRegName string,
+) (database.Database, error) {
```
StephenButtolph, line 31:
```suggestion
// New creates a new database instance based on the provided configuration.
//
// It also wraps the database with a corruptable DB and a meter DB.
//
// dbName is the name of the database, either leveldb, memdb, or pebbledb.
// dbPath is the path to the database folder.
// readOnly indicates if the database should be read-only.
// dbConfig is the database configuration in JSON format.
// dbMetricsPrefix is used to create a new metrics registerer for the database.
// meterDBRegName is used to create a new metrics registerer for the meter DB.
func New(
	name string,
	path string,
	readOnly bool,
	config []byte,
	gatherer metrics.MultiGatherer,
	logger logging.Logger,
	metricsPrefix string,
	meterDBRegName string,
) (database.Database, error) {
```

StephenButtolph (approved):

database/factory/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	logger logging.Logger,
+	metricsPrefix string,
+	meterDBRegName string,
+) (database.Database, error) {
+	dbRegisterer, err := metrics.MakeAndRegister(
+		gatherer,
+		metricsPrefix,
+	)
+	if err != nil {
+		return nil, err
```
tsachiherman, line 45: nit: could you create more specific formatted errors to the ones here ?
i.e. fmt.Errorf("failed to MakeAndRegister the metricPrefix %s : %w". metricsPrefix, err)

same for the others error in this function.

joshua-kim: I agree with this comment but I don't think it makes sense to block this PR on it because this is a strict refactor.

PR #3900, opened 2025-04-19
use atomicvm version for coreth

ceyonur:
## Why this should be merged

Bumps coreth to atomicvm version

## How this works

Bumps coreth and uses coreth/factory to initialize the factory

## How this was tested

e2e tests
bootstrap fuji

No


StephenButtolph:
This PR was essentially included in #3999.

PR #3905, opened 2025-04-21
wait until local p-chain has caught up with the p-chain height of inner VM blocks

yacovm:

### DIFF
--- a/vms/proposervm/vm.go
+++ b/vms/proposervm/vm.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


vms/proposervm/vm.go
```diff
@@ -249,6 +246,62 @@ func (vm *VM) Initialize(
...
+	go vm.ctx.Log.RecoverAndPanic(func() {
+		// Since new blocks built must have a P-chain height that is at least the P-chain height
+		// of their ancestors, if our local P-chain height is behind the P-chain height of our inner VM,
+		// we cannot build blocks.
+		// We therefore wait until the local P-chain has caught up with the P-chain height of the last block
+		// of our inner VM.
+		for !vm.isLocalPlatformChainInSync() {
+			time.Sleep(pChainPostBootstrapSyncPollInterval)
+		}
+		scheduler.Dispatch(time.Now())
```
StephenButtolph, line 263: By deferring the call to `Dispatch` the scheduler is going to block upon calls to `scheduler.SetBuildBlockTime`. However, `scheduler.SetBuildBlockTime` is called with the context lock held in `SetPreference`.

I think that can cause a deadlock if we are unlucky enough for the P-chain to have caught back up before `SetPreference` is called but without `isLocalPlatformChainInSync` having returned `true`.

yacovm: good catch, fixed.

PR #3909, opened 2025-04-23
Vm sdk

aaronbuchwald:

### DIFF
--- a/vms/sdk/event/event.go
+++ b/vms/sdk/event/event.go
@@ -1,0 +1,85 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package event
+
+import (
+	"context"
+	"errors"
+)
+
+var (
+	_ Subscription[struct{}]        = (*SubscriptionFunc[struct{}])(nil)
+	_ SubscriptionFactory[struct{}] = (*SubscriptionFuncFactory[struct{}])(nil)
+)
+
+// SubscriptionFactory returns an instance of a concrete Subscription
+type SubscriptionFactory[T any] interface {
+	New() (Subscription[T], error)
+}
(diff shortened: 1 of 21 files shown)


vms/sdk/event/event.go
```diff
@@ -0,0 +1,93 @@
...
+				if err := sub.Close(); err != nil {
+					errs = append(errs, err)
+				}
+			}
+			return errors.Join(errs...)
+		},
+	}
+}
+
+// Map transforms a subscription from an output sub to an input typed sub
```
StephenButtolph, line 85: I waffled a lot on this description. It really isn't the format I'm used to for a Map function... Typically I would think `Map` would take in a `Stream[I]` and `f(I) O` and return a `Stream[O]`. But instead this takes in an `EventHandler[O]` and `f(I) O` and returns an `EventHandler[I]`...

Perhaps:
```suggestion
// Map returns a new subscription which transforms an event with f and notifies
// the original subscription of the result.
//
// The original subscription will be closed if the returned subscription is closed.
```

vms/sdk/event/event_test.go
```diff
@@ -0,0 +1,78 @@
...
+	r.NoError(aggregateSub.Close())
+	<-sub1Ch
+	<-sub2Ch
+}
+
+func TestMapSub(t *testing.T) {
+	r := require.New(t)
+	sub, subCh := newTestSubscription()
+
+	mappedSub := Map(func(eventID string) ids.ID { return ids.FromStringOrPanic(eventID) }, sub)
```
StephenButtolph, line 72:
```suggestion
	mappedSub := Map(ids.FromStringOrPanic, sub)
```

StephenButtolph (commented):
Didn't finish the review - but need to go get dinner + workout

PR #3910, opened 2025-04-23
fix rpcchainvm handling for arbitrary length http body

joshua-kim:
## Why this should be merged

Fixes a bug in rpcchainvm where a arbitrary length http body will hang when attempting to copy the body into protobuf.

## How this works

Instead of copying the body into a proto message, wrap the http body in an implementation of `io.Reader` that streams the message body. 

## How this was tested

Added a regression test

Yes


vms/rpcchainvm/ghttp/grequest/request_client.go
```diff
@@ -0,0 +1,34 @@
...
+type Client struct {
+	ctx    context.Context
+	client request.RequestClient
+}
+
+func (c *Client) Read(p []byte) (n int, err error) {
+	reply, err := c.client.Body(c.ctx, &request.BodyRequest{N: uint32(len(p))})
+	if err != nil {
+		return 0, err
+	}
```
joshua-kim, line 31: I think it's fine for us to just return the error here. `io.EOF` is a special error for both grpc and `io.Reader` ([ref](https://pkg.go.dev/io?utm_source=godoc#Reader)). `io.EOF`'s invariants on `io.Reader` are particularly confusing though.

proto/http/request/request.proto
StephenButtolph, line 1: Can we reuse the [io.Reader](https://github.com/ava-labs/avalanchego/blob/master/vms/rpcchainvm/ghttp/greader/reader_client.go) implementation?

StephenButtolph:
Also - linting seems to be failing

vms/rpcchainvm/ghttp/http_test.go
```diff
@@ -49,3 +61,44 @@ func TestConvertWriteResponse(t *testing.T) {
...
+
+func (fooServer) Foo(context.Context, *test.FooRequest) (*test.FooResponse, error) {
+	return &test.FooResponse{}, nil
+}
+
+type infiniteStream struct{}
+
+func (i infiniteStream) Read(p []byte) (n int, err error) {
+	return len(p), nil
+}
```
StephenButtolph, line 104: I don't think we need this `fooServer` - or the `test` proto file at all
```suggestion
func TestRequestClientArbitrarilyLongBody(t *testing.T) {
	require := require.New(t)

	listener := bufconn.Listen(0)
	server := grpc.NewServer()
	httppb.RegisterHTTPServer(server, &httppb.UnimplementedHTTPServer{})

	go func() {
		require.NoError(server.Serve(listener))
	}()

	conn, err := grpc.NewClient(listener.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
	require.NoError(err)

	client := NewClient(httppb.NewHTTPClient(conn))

	w := &httptest.ResponseRecorder{}
	r := &http.Request{
		Header: map[string][]string{
			"Upgrade": {"foo"}, // Make this look like a streaming request
		},
		Body: io.NopCloser(infiniteStream{}),
	}

	client.ServeHTTP(w, r) // Shouldn't block forever reading the body
}

type infiniteStream struct{}

func (infiniteStream) Read(p []byte) (n int, err error) {
	return len(p), nil
}
```

joshua-kim: I forgot to unstage this... originally was trying to test this through a dummy service but ended up not using it

proto/pb/test/foo_grpc.pb.go
StephenButtolph, line 1: Should this have been deleted?

proto/pb/test/foo.pb.go
StephenButtolph, line 1: Should this have been deleted?

vms/rpcchainvm/ghttp/http_body.go
StephenButtolph, line 1: Can't we use the existing implementation for this?

proto/pb/test/foo_grpc.pb.go
joshua-kim: Removed https://github.com/ava-labs/avalanchego/pull/3910/commits/edabb0b404d4396c04fe755868b6fb19f4f4e4c8

vms/rpcchainvm/ghttp/http_body.go
joshua-kim: Updated: https://github.com/ava-labs/avalanchego/pull/3910/commits/3c93340cffe2d9429749a816a06527db11f4108f

StephenButtolph (approved):

tsachiherman (approved):
looks functionally correct to me.

PR #3911, opened 2025-04-24
docs: fix flag name to `--proposervm-min-block-delay`

ceyonur:
Fixes `--proposervm-min-block-delay` flag name in docs


yacovm (approved):

ceyonur:
> how can we make sure it also gets propagated here https://build.avax.network/docs/nodes/configure/configs-flags?

It should be automatically fetched by the docs site.

PR #3917, opened 2025-04-29
Fix bug in rpcchainvm where writes in a the plugin would not be propagated back to the client

joshua-kim:
## Why this should be merged

Factored out of https://github.com/ava-labs/avalanchego/pull/3294

## How this works

Introduces a rpc to sync the headers from the plugin vm. Our current implementation does not work because the headers is just a map/not an interface, so any modifications to the map are not synced across avalanchego <-> the plugin.

## How this was tested

Added a regression test

Yes


vms/rpcchainvm/ghttp/http_test.go
```diff
@@ -84,6 +86,58 @@ func TestRequestClientArbitrarilyLongBody(t *testing.T) {
...
+	client := NewClient(httppb.NewHTTPClient(conn))
+
+	recorder := &httptest.ResponseRecorder{
+		Body: bytes.NewBuffer(nil),
+	}
+
+	request := &http.Request{
+		Body: io.NopCloser(strings.NewReader("bar")),
+		Header: http.Header{
+			"Upgrade": {"upgrade"},
```
joshua-kim, line 125: We have to specify this header because this fixes the bug when we're not using http simple. I'm also wondering if we should just remove `handleHttpSimple` entirely...

proto/http/http.proto
```diff
@@ -147,6 +145,11 @@ message HTTPRequest {
   Request request = 2;
 }
 
+message HttpResponse {
```
StephenButtolph, line 148: To align with `HTTPRequest`:
```suggestion
message HTTPResponse {
```

vms/rpcchainvm/ghttp/http_test.go
```diff
@@ -84,6 +88,57 @@ func TestRequestClientArbitrarilyLongBody(t *testing.T) {
...
+		Trailer: http.Header{
+			"bar": {"bar"},
+		},
+		RemoteAddr: "bar",
+	}
+
+	client.ServeHTTP(recorder, request)
+
+	require.Equal(200, recorder.Code)
+	require.Equal([]string{"bar"}, recorder.Header()["Foo"])
```
StephenButtolph, line 138:
```suggestion
	require.Equal(
		http.Header{
			"Foo": []string{"bar"},
		},
		recorder.Header(),
	)
```

vms/rpcchainvm/ghttp/http_server.go
```diff
@@ -142,7 +141,21 @@ func (s *Server) Handle(ctx context.Context, req *httppb.HTTPRequest) (*emptypb.
...
+		return nil, fmt.Errorf("failed to close client conn: %w", err)
+	}
+
+	header := make([]*httppb.Element, len(writerHeaders))
+	for k, v := range writerHeaders {
+		header = append(header, &httppb.Element{
+			Key:    k,
+			Values: v,
+		})
+	}
```
StephenButtolph, line 154: I believe this will result in `header` having a final length of `2 * len(writerHeaders)` where the first `len(writerHeaders)` entries are `nil`.
```suggestion
	header := make([]*httppb.Element, 0, len(writerHeaders))
	for k, v := range writerHeaders {
		header = append(header, &httppb.Element{
			Key:    k,
			Values: v,
		})
	}
```

StephenButtolph: This doesn't panic, rather it ends up adding a `"": nil` entry to the headers.

vms/rpcchainvm/ghttp/http_test.go
joshua-kim: This is also nicer since it captures that bug you found....

StephenButtolph (approved):

vms/rpcchainvm/ghttp/http_server.go
```diff
@@ -142,7 +141,21 @@ func (s *Server) Handle(ctx context.Context, req *httppb.HTTPRequest) (*emptypb.
...
+		return nil, fmt.Errorf("failed to close client conn: %w", err)
+	}
+
+	header := make([]*httppb.Element, 0, len(writerHeaders))
+	for k, v := range writerHeaders {
+		header = append(header, &httppb.Element{
+			Key:    k,
+			Values: v,
+		})
+	}
```
StephenButtolph, line 154: Looks like we already have a helper for this
```suggestion
	header := grpcutils.GetHTTPHeader(writerHeaders)
```

vms/rpcchainvm/ghttp/http_client.go
```diff
@@ -156,9 +156,17 @@ func (c *Client) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	header := w.Header()
+	clear(header)
+
+	for _, h := range reply.Header {
+		header[h.Key] = h.Values
 	}
```
StephenButtolph, line 170: I think we also have a helper for this right now: `grpcutils.MergeHTTPHeader` That implementation doesn't clear the input header... But I think it should. Currently if the implementation deletes a header that is not honored.

StephenButtolph (approved):

vms/rpcchainvm/ghttp/http_test.go
```diff
@@ -84,6 +88,78 @@ func TestRequestClientArbitrarilyLongBody(t *testing.T) {
 	client.ServeHTTP(w, r) // Shouldn't block forever reading the body
 }
 
+// Tests that writes to the http response in the server are propagated to the
+// client
+func TestHttpResponse(t *testing.T) {
```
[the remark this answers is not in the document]
joshua-kim: I don't think so... we need to set up a client <-> server pair and I'm not aware of a simpler way to do so.

vms/rpcchainvm/ghttp/http_test.go
```diff
@@ -84,6 +88,78 @@ func TestRequestClientArbitrarilyLongBody(t *testing.T) {
...
+				grpc.WithTransportCredentials(insecure.NewCredentials()),
+			)
+			require.NoError(err)
+
+			recorder := &httptest.ResponseRecorder{
+				Body: bytes.NewBuffer(nil),
+			}
+
+			client := NewClient(httppb.NewHTTPClient(conn))
+			client.ServeHTTP(recorder, &tt.request)
```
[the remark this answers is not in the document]
joshua-kim: We should document this, but this is because rpcchainvm's implementation handles a request that specifies the upgrade header differently as a performance optimization. I can leave a comment here.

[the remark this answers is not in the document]
joshua-kim: Basically when handling a request, we translate it into a proto message to translate it across the avalanchego <-> vm boundary. We explicitly define two separate apis to translate request/responses, which we call http/httpSimple. I originally added the regression test specifying the upgrade header (http), but also added coverage on existing untested code for httpSimple.

vms/rpcchainvm/ghttp/http_test.go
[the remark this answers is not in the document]
joshua-kim: I agree that the test setup is complex... but I'm also not sure if grpc gives us any test helpers that make this actually possible within the scope of this PR. The silver lining is that we previously had no unit tests at all on this package before this sequence of PRs so there wasn't any tests to tell us that this package was hard to test.

vms/rpcchainvm/ghttp/http_client.go
joshua-kim: Updated, + added coverage on both http + http simple

PR #3918, opened 2025-04-30
Use libevm instead of coreth

ceyonur:
## Why this should be merged

Importing `customtypes` and `params` pkg causes a libevm side-effect which prevents us to use any libevm based (subnet-evm, coreth) VMs to use the e2e tests here. This PR reduces usage of side-effect pkg. 

## How this was tested

* Added a new test, `TestMustNotImport`, in `tests/fixture/imports_test.go`. This test ensures certain packages do not import restricted modules like `plugin/evm/customtypes` and `params` to avoid global configuration conflicts.

No


JuanLeon2:
May or may not conflict with pr 3923

ceyonur:
> May or may not conflict with pr 3923

Yes it conflicts with 3923. This one addresses comments in that PR.

tests/imports_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Start collecting dependencies
+	for _, pkg := range pkgs {
+		collectDeps(pkg)
+	}
+	return deps, nil
+}
+
+func TestMustNotImport(t *testing.T) {
+	withRepo := func(pkg string, repo string) string {
```
StephenButtolph, line 52: Super nit: I originally thought there was a bug here because the arg order is pkg, repo but (obv) the result is repo+pkg.

tests/imports_test.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Sprintf("%s/%s", repo, pkg)
```
StephenButtolph, line 53: Should this be a filepath.Join? If it should, should this helper be removed entirely?

tests/imports_test.go
```diff
@@ -0,0 +1,77 @@
...
+		"tests/...": {"plugin/evm/customtypes", "params"},
+	}
+
+	for packageName, forbiddenImports := range mustNotImport {
+		imports, err := getDependencies(withRepo(packageName, sourceRepo))
+		require.NoError(t, err)
+
+		for _, forbiddenImport := range forbiddenImports {
+			fullForbiddenImport := withRepo(forbiddenImport, targetRepo)
+			_, found := imports[fullForbiddenImport]
```
StephenButtolph, line 73: require.NotContains?

tests/imports_test.go
```diff
@@ -0,0 +1,77 @@
...
+	pkgs, err := packages.Load(cfg, packageName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load package: %w", err)
+	}
+
+	if len(pkgs) == 0 {
+		return nil, fmt.Errorf("no packages found for %s", packageName)
+	}
+
+	if pkgs[0].Errors != nil {
```
StephenButtolph, line 28: Should this be checked for every package?

tests/imports_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/tools/go/packages"
+)
+
+// getDependencies takes a fully qualified package name and returns a map of all
```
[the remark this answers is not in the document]
ceyonur: it was copied from coreth. I made it a testult and we can use it in different pkgs.

tests/imports_test.go
ceyonur: `filepath.Join` can have different separators for different OSes, I dont think it's a direct equivalent.

ceyonur: `path.Join` might be

utils/packages/dependencies.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package packages
```
StephenButtolph, line 4: This can't be called `package` because `package` is a keyword in golang.

StephenButtolph (approved):

PR #3932, opened 2025-05-02
refactor: replace []byte(fmt.Sprintf) with fmt.Appendf

findnature:
## Why this should be merged

Optimize code using a more modern writing style.   Official support from Go, for more details visit https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.

### DIFF
--- a/staking/tls_test.go
+++ b/staking/tls_test.go
@@ -21,7 +21,7 @@ func TestMakeKeys(t *testing.T) {
 	cert, err := NewTLSCert()
 	require.NoError(err)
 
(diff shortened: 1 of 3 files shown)


StephenButtolph:
@findnature Would it be possible to enable modernize in our existing golangci-lint tests?

findnature:
> After looking around - it doesn't seem trivial to include gopls suggestions into golangci-lint. Bummer.

Yes, it's still not very mature and has some bugs. The code processed by it may not compile successfully and will likely require manual intervention.

PR #3938, opened 2025-05-05
Add logging to corruptabledb closure

joshua-kim:
## Why this should be merged

Adds visibility to the closure reason in avalanchego's db

## How this works

Adds a log upon hitting an error

## How this was tested

Did not

No

### DIFF
--- a/database/factory/factory.go
+++ b/database/factory/factory.go
@@ -28,6 +28,7 @@ import (
 // dbMetricsPrefix is used to create a new metrics registerer for the database.
 // meterDBRegName is used to create a new metrics registerer for the meter DB.
 func New(
+	log logging.Logger,
 	name string,
 	path string,
 	readOnly bool,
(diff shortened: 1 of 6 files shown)


database/factory/factory.go
```diff
@@ -28,6 +28,7 @@ import (
 // dbMetricsPrefix is used to create a new metrics registerer for the database.
 // meterDBRegName is used to create a new metrics registerer for the meter DB.
 func New(
+	log logging.Logger,
```
StephenButtolph, line 31: The log is already passed in as `logger`

vms/rpcchainvm/vm_server.go
```diff
@@ -183,6 +183,7 @@ func (vm *VMServer) Initialize(ctx context.Context, req *vmpb.InitializeRequest)
 	vm.connCloser.Add(dbClientConn)
 	vm.db = corruptabledb.New(
 		rpcdb.NewClient(rpcdbpb.NewDatabaseClient(dbClientConn)),
+		logging.NoLog{},
```
StephenButtolph, line 186: Can we pass in the real logger here? (The logger is defined a on the next line)

StephenButtolph (approved):

tsachiherman (approved):

PR #1173, opened 2022-11-11
Avoiding panic on genesis with no unlock events

abhvious:
## Why this should be merged

This small change prevents a panic in the avalanchego binary when it is run on a valid genesis.json file that does not include an unlock event entry.

## How this works

It is not clear to me what the semantics of the splitAllocations function is.   [Line 369](https://github.com/ava-labs/avalanchego/blob/364581b756acc0d81fa331a577a0326387b57a10/genesis/genesis.go#L369) of genesis.go loops over the InitialStakers array, but accesses an "allAllocationNodes" array and expects that array to have the same number of entries as the InitialStakers.  Perhaps there is a better fix which avoids this linkage.

However, the change in this PR is the smallest one that avoids a panic when the genesis.json file does not have unlock events. See attached test file.

## How this was tested

Sample file.
[test_gen.json.txt](https://github.com/ava-labs/avalanchego/files/9992334/test_gen.json.txt)

In 1.9.2, there will be a panic.
```
./build/avalanchego --public-ip=127.0.0.1 --http-port=9650 --staking-port=9651 --db-dir=db/node0 --staking-tls-cert-file=../../staking/s0.crt --staking-tls-key-file=../../staking/s0.key --genesis=../../test_gen.json.txt --network-id=1337

panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/ava-labs/avalanchego/genesis.FromConfig(0x14000262900)
	/Users/shelat/ava/avalanchego/genesis/genesis.go:376 +0x12fc
github.com/ava-labs/avalanchego/genesis.FromFile(0x539, {0x16f7a3886, 0x10})
	/Users/shelat/ava/avalanchego/genesis/genesis.go:206 +0x24c
github.com/ava-labs/avalanchego/config.getGenesisData(0x3feccccccccccccd?, 0x539)
	/Users/shelat/ava/avalanchego/config/config.go:862 +0x16c
github.com/ava-labs/avalanchego/config.GetNodeConfig(_, {_, _})
	/Users/shelat/ava/avalanchego/config/config.go:1349 +0x5c8
main.main()
	/Users/shelat/ava/avalanchego/main/main.go:42 +0x210
```

In this PR, the expected behavior is this error:

```
couldn't load node config: unable to load genesis file: problem while building platform chain's genesis state: validator would have already unstaked
```

PR #3941, opened 2025-05-06
fix: validate allocations locked amount in genesis to prevent panic

### LINKED ISSUE #3901: Invalid genesis bytes may panic
## Description

Avalanchego currently does not ensure that genesis creation gracefully errors. This bug was originally reported in #1173, but that PR did not necessarily ensure that all output genesis bytes are well-formed for the rest of initialization.

## Reproduction

See the original PR for an example genesis file that causes a panic.

## Suggested Fix

I originally had a comment to add additional validation to the geneisis parsing [here](https://github.com/ava-labs/avalanchego/pull/1173#issuecomment-1319381906).

DracoLi:
## Why this should be merged

Prevents panics when the allocations locked amount in genesis cannot be split across the initial stakers.
Fixes #3901

## How this works

Add `validateLockedAllocationsAmount` to genesis `validateConfig`

## How this was tested

Manually tested via `avalanchego --genesis-file=` and new unit test.

### DIFF
--- a/genesis/genesis.go
+++ b/genesis/genesis.go
@@ -46,6 +46,7 @@ var (
 	errFutureStartTime                 = errors.New("startTime cannot be in the future")
 	errInitialStakeDurationTooLow      = errors.New("initial stake duration is too low")
 	errOverridesStandardNetworkConfig  = errors.New("overrides standard network genesis config")
+	errAllocationsLockedAmountTooLow   = errors.New("total allocations locked amount is too low")
 )
 
 // validateInitialStakedFunds ensures all staked
@@ -113,6 +114,27 @@ func validateInitialStakedFunds(config *Config) error {
 	return nil
 }
 
+// validateAllocationsLockedAmount ensures that the sum of all locked
+// allocation amounts is at least the number of initial stakers.
+func validateAllocationsLockedAmount(config *Config) error {
+	totalLocked := uint64(0)
+	for _, allocation := range config.Allocations {
+		for _, unlock := range allocation.UnlockSchedule {
+			totalLocked += unlock.Amount
+		}
+	}
+	stakersCount := len(config.InitialStakers)
+	if totalLocked < uint64(stakersCount) {
+		return fmt.Errorf(
+			"%w: %d locked < %d stakers",
+			errAllocationsLockedAmountTooLow,
+			totalLocked,
+			stakersCount,
+		)
+	}
+	return nil
+}
+
 // validateConfig returns an error if the provided
 // *Config is not considered valid.
 func validateConfig(networkID uint32, config *Config, stakingCfg *StakingConfig) error {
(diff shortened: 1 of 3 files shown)


genesis/genesis.go
```diff
@@ -533,7 +559,10 @@ func splitAllocations(allocations []Allocation, numSplits int) [][]Allocation {
 		}
 	}
 
-	return append(allNodeAllocations, currentNodeAllocation)
+	if len(currentNodeAllocation) > 0 {
+		allNodeAllocations = append(allNodeAllocations, currentNodeAllocation)
+	}
+	return allNodeAllocations
```
StephenButtolph, line 565: What does this modification do?

genesis/genesis_test.go
```diff
@@ -118,6 +118,20 @@ func TestValidateConfig(t *testing.T) {
 			}(),
 			expectedErr: errInitialStakeDurationTooLow,
 		},
+		"locked allocations amount too low": {
```
joshua-kim, line 121: This technically works... I'm not a fan of this existing test though because it tests through `validateConfig` which isn't a part of this package's exported API though. I think a majority of these tests are actually invalid configs, because they would all fail the network id check. Can we test this through the calling code which is `FromFile`? We have an existing test `TestGenesisFromFile` which is set up to do this... we could also use the user-provided genesis config that revealed this bug linked on https://github.com/ava-labs/avalanchego/pull/1173

genesis/genesis.go
DracoLi: not required for the fix and can be omitted. 
this modification ensures we don't return an array with an empty nested array when no locked amount. https://github.com/ava-labs/avalanchego/pull/3941/files/2842951c8860726232091bff268849ce0ff7117e#diff-2ce354664b078c8815c9e3644165471d1b47b509a4ce00763895428f2f1f384bR514-R515

genesis/genesis_test.go
DracoLi: sounds good, i'll move this test to be part of FromFile.

genesis/genesis_test.go
```diff
@@ -118,15 +121,6 @@ func TestValidateConfig(t *testing.T) {
...
 		},
-		"empty initial staked funds": {
-			networkID: 12345,
-			config: func() *Config {
-				thisConfig := LocalConfig
-				thisConfig.InitialStakedFunds = []ids.ShortID(nil)
-				return &thisConfig
-			}(),
-			expectedErr: errNoInitiallyStakedFunds,
-		},
```
joshua-kim, line 129: Did we mean to remove this? I think this is a different case than the insufficient allocations case

DracoLi: definitely not, i'll need to add this back

StephenButtolph (approved):

joshua-kim (approved):

PR #3942, opened 2025-05-06
Add Load Framework

RodrigoVillar:
## Why this should be merged

This PR migrates the load framework from the HyperSDK to AvalancheGo, enabling its usage in both the HyperSDK and any other project (e.g. Coreth, Subnet-EVM, etc.) which wishes to benchmark its TPS numbers.

## How this works

Migration of code from HyperSDK to AvalancheGo.

## How this was tested

Both the gradual and burst orchestrators were tested against HyperVMs in the HyperSDK. Furthermore, we have UTs for the gradual orchestrator.

No

### DIFF
--- a/tests/load/dependencies.go
+++ b/tests/load/dependencies.go
@@ -1,0 +1,57 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+import (
+	"context"
+)
+
+type Issuer[T comparable] interface {
+	// GenerateAndIssueTx generates and sends a tx to the network, and informs the
+	// tracker that it sent said transaction. It returns the sent transaction.
+	GenerateAndIssueTx(ctx context.Context) (tx T, err error)
+}
+
+type Listener[T comparable] interface {
+	// Listen for the final status of transactions and notify the tracker
+	// Listen stops if the context is done, an error occurs, or if it received
+	// all the transactions issued and the issuer no longer issues any.
+	// Listen MUST return a nil error if the context is canceled.
+	Listen(ctx context.Context) error
+
+	// RegisterIssued informs the listener that a transaction was issued.
+	RegisterIssued(tx T)
+
+	// IssuingDone informs the listener that no more transactions will be issued.
+	IssuingDone()
+}
+
+// Tracker keeps track of the status of transactions.
+// This must be thread-safe, so it can be called in parallel by the issuer(s) or orchestrator.
+type Tracker[T comparable] interface {
+	// Issue records a transaction that was submitted, but whose final status is
+	// not yet known.
+	Issue(T)
+	// ObserveConfirmed records a transaction that was confirmed.
+	ObserveConfirmed(T)
+	// ObserveFailed records a transaction that failed (e.g. expired)
+	ObserveFailed(T)
+
+	// GetObservedIssued returns the number of transactions that the tracker has
+	// confirmed were issued.
+	GetObservedIssued() uint64
+	// GetObservedConfirmed returns the number of transactions that the tracker has
+	// confirmed were accepted.
+	GetObservedConfirmed() uint64
+	// GetObservedFailed returns the number of transactions that the tracker has
+	// confirmed failed.
+	GetObservedFailed() uint64
+}
+
+// orchestrator executes the load test by coordinating the issuers to send
+// transactions, in a manner depending on the implementation.
+type orchestrator interface {
+	// Execute the load test
+	Execute(ctx context.Context) error
+}
--- a/tests/load/gradual_orchestrator.go
+++ b/tests/load/gradual_orchestrator.go
@@ -1,0 +1,268 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+import (
+	"context"
+	"errors"
+	"math"
+	"sync/atomic"
+	"time"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+var (
+	_ orchestrator = (*GradualOrchestrator[any, any])(nil)
+
+	ErrFailedToReachTargetTPS = errors.New("failed to reach target TPS")
+)
+
+type GradualOrchestratorConfig struct {
+	// The maximum TPS the orchestrator should aim for.
+	MaxTPS uint64
+	// The minimum TPS the orchestrator should start with.
+	MinTPS uint64
+	// The step size to increase the TPS by.
+	Step uint64
+
+	// The factor by which to pad the number of txs an issuer sends per second
+	// for example, if targetTPS = 1000 and numIssuers = 10, then each issuer
+	// will send (1000/10)*TxRateMultiplier transactions per second.
+	//
+	// Maintaining a multiplier above target provides a toggle to keep load
+	// persistently above instead of below target. This ensures load generation
+	// does not pause issuance at the target and persistently under-issue and
+	// fail to account for the time it takes to add more load.
+	TxRateMultiplier float64
+
+	// The time period which TPS is averaged over
+	// Similarly, the time period which the orchestrator will wait before
+	// computing the average TPS.
+	SustainedTime time.Duration
+	// The number of attempts to try achieving a given target TPS before giving up.
+	MaxAttempts uint64
+
+	// Whether the orchestrator should return if the maxTPS has been reached
+	Terminate bool
+}
+
+func DefaultGradualOrchestratorConfig() GradualOrchestratorConfig {
+	return GradualOrchestratorConfig{
+		MaxTPS:           5_000,
+		MinTPS:           1_000,
+		Step:             1_000,
+		TxRateMultiplier: 1.3,
+		SustainedTime:    20 * time.Second,
+		MaxAttempts:      3,
+		Terminate:        true,
+	}
+}
+
+// GradualOrchestrator tests the network by continuously sending
+// transactions at a given rate (currTargetTPS) and increasing that rate until it detects that
+// the network can no longer make progress (i.e. the rate at the network accepts
+// transactions is less than currTargetTPS).
+type GradualOrchestrator[T, U comparable] struct {
+	agents []Agent[T, U]
+
+	log logging.Logger
+
+	maxObservedTPS atomic.Uint64
+
+	observerGroup errgroup.Group
+	issuerGroup   *errgroup.Group
+
+	config GradualOrchestratorConfig
+}
+
+func NewGradualOrchestrator[T, U comparable](
+	agents []Agent[T, U],
+	log logging.Logger,
+	config GradualOrchestratorConfig,
+) (*GradualOrchestrator[T, U], error) {
+	return &GradualOrchestrator[T, U]{
+		agents: agents,
+		log:    log,
+		config: config,
+	}, nil
+}
+
+func (o *GradualOrchestrator[T, U]) Execute(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
+
+	// start a goroutine to confirm each issuer's transactions
+	for _, agent := range o.agents {
+		o.observerGroup.Go(func() error { return agent.Listener.Listen(ctx) })
+	}
+
+	// start the test and block until it's done
+	success := o.run(ctx)
+
+	var err error
+	if !success {
+		err = ErrFailedToReachTargetTPS
+	}
+
+	// stop the observers and issuers
+	cancel()
+
+	// block until both the observers and issuers have stopped
+	return errors.Join(o.issuerGroup.Wait(), o.observerGroup.Wait(), err)
+}
+
+// run the gradual load test by continuously increasing the rate at which
+// transactions are sent
+//
+// run blocks until one of the following conditions is met:
+//
+// 1. an issuer has errored
+// 2. the max TPS target has been reached and we can terminate
+// 3. the maximum number of attempts to reach a target TPS has been reached
+func (o *GradualOrchestrator[T, U]) run(ctx context.Context) bool {
+	var (
+		prevConfirmed        = GetTotalObservedConfirmed(o.agents)
+		prevTime             = time.Now()
+		currTargetTPS        = new(atomic.Uint64)
+		attempts      uint64 = 1
+		// true if the orchestrator has reached the max TPS target
+		achievedTargetTPS bool
+	)
+
+	currTargetTPS.Store(o.config.MinTPS)
+
+	issuerGroup, issuerCtx := errgroup.WithContext(ctx)
+	o.issuerGroup = issuerGroup
+
+	o.issueTxs(issuerCtx, currTargetTPS)
+
+	for {
+		// wait for the sustained time to pass or for the context to be cancelled
+		select {
+		case <-time.After(o.config.SustainedTime):
+		case <-issuerCtx.Done(): // the parent context was cancelled or an issuer errored
+		}
+
+		if issuerCtx.Err() != nil {
+			break // Case 1
+		}
+
+		currConfirmed := GetTotalObservedConfirmed(o.agents)
+		currTime := time.Now()
+
+		tps := computeTPS(prevConfirmed, currConfirmed, currTime.Sub(prevTime))
+		o.setMaxObservedTPS(tps)
+
+		// if max TPS target has been reached and we don't terminate, then continue here
+		// so we do not keep increasing the target TPS
+		if achievedTargetTPS && !o.config.Terminate {
+			o.log.Info(
+				"current network state",
+				zap.Uint64("current TPS", tps),
+				zap.Uint64("max observed TPS", o.maxObservedTPS.Load()),
+			)
+			continue
+		}
+
+		if tps >= currTargetTPS.Load() {
+			if currTargetTPS.Load() >= o.config.MaxTPS {
+				achievedTargetTPS = true
+				o.log.Info(
+					"max TPS target reached",
+					zap.Uint64("max TPS target", currTargetTPS.Load()),
+					zap.Uint64("average TPS", tps),
+				)
+				if o.config.Terminate {
+					o.log.Info("terminating orchestrator")
+					break // Case 2
+				} else {
+					o.log.Info("orchestrator will now continue running at max TPS")
+					continue
+				}
+			}
+			o.log.Info(
+				"increasing TPS",
+				zap.Uint64("previous target TPS", currTargetTPS.Load()),
+				zap.Uint64("average TPS", tps),
+				zap.Uint64("new target TPS", currTargetTPS.Load()+o.config.Step),
+			)
+			currTargetTPS.Add(o.config.Step)
+			attempts = 1
+		} else {
+			if attempts >= o.config.MaxAttempts {
+				o.log.Info(
+					"max attempts reached",
+					zap.Uint64("attempted target TPS", currTargetTPS.Load()),
+					zap.Uint64("number of attempts", attempts),
+				)
+				break // Case 3
+			}
+			o.log.Info(
+				"failed to reach target TPS, retrying",
+				zap.Uint64("current target TPS", currTargetTPS.Load()),
+				zap.Uint64("average TPS", tps),
+				zap.Uint64("attempt number", attempts),
+			)
+			attempts++
+		}
+
+		prevConfirmed = currConfirmed
+		prevTime = currTime
+	}
+
+	return achievedTargetTPS
+}
+
+// GetObservedIssued returns the max TPS the orchestrator observed
+func (o *GradualOrchestrator[T, U]) GetMaxObservedTPS() uint64 {
+	return o.maxObservedTPS.Load()
+}
+
+// start a goroutine to each issuer to continuously send transactions
+// if an issuer errors, all other issuers will stop as well.
+func (o *GradualOrchestrator[T, U]) issueTxs(ctx context.Context, currTargetTPS *atomic.Uint64) {
+	for _, agent := range o.agents {
+		o.issuerGroup.Go(func() error {
+			for {
+				if ctx.Err() != nil {
+					return nil //nolint:nilerr
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.agents))) * o.config.TxRateMultiplier)
+				// always listen until listener context is cancelled, do not call agent.Listener.IssuingDone().
+				for range txsPerIssuer {
+					tx, err := agent.Issuer.GenerateAndIssueTx(ctx)
+					if err != nil {
+						return err
+					}
+					agent.Listener.RegisterIssued(tx)
+				}
+				diff := time.Second - time.Since(currTime)
+				if diff > 0 {
+					time.Sleep(diff)
+				}
+			}
+		})
+	}
+}
+
+// setMaxObservedTPS only if tps > the current max observed TPS.
+func (o *GradualOrchestrator[T, U]) setMaxObservedTPS(tps uint64) {
+	if tps > o.maxObservedTPS.Load() {
+		o.maxObservedTPS.Store(tps)
+	}
+}
+
+func computeTPS(initial uint64, final uint64, duration time.Duration) uint64 {
+	if duration <= 0 {
+		return 0
+	}
+
+	numTxs := final - initial
+	tps := float64(numTxs) / duration.Seconds()
+
+	return uint64(tps)
+}
--- a/tests/load/agent.go
+++ b/tests/load/agent.go
@@ -1,0 +1,30 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+type Agent[T, U comparable] struct {
+	Issuer   Issuer[T]
+	Listener Listener[T]
+	Tracker  Tracker[U]
+}
+
+func NewAgent[T, U comparable](
+	issuer Issuer[T],
+	listener Listener[T],
+	tracker Tracker[U],
+) Agent[T, U] {
+	return Agent[T, U]{
+		Issuer:   issuer,
+		Listener: listener,
+		Tracker:  tracker,
+	}
+}
+
+func GetTotalObservedConfirmed[T, U comparable](agents []Agent[T, U]) uint64 {
+	total := uint64(0)
+	for _, agent := range agents {
+		total += agent.Tracker.GetObservedConfirmed()
+	}
+	return total
+}
--- a/tests/load/gradual_orchestrator_test.go
+++ b/tests/load/gradual_orchestrator_test.go
@@ -1,0 +1,202 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+import (
+	"context"
+	"errors"
+	"math"
+	"testing"
+	"time"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+var _ Issuer[ids.ID] = (*mockIssuer)(nil)
+
+func TestGradualOrchestratorTPS(t *testing.T) {
+	tests := []struct {
+		name        string
+		serverTPS   uint64
+		config      GradualOrchestratorConfig
+		expectedErr error
+	}{
+		{
+			name:      "orchestrator achieves max TPS",
+			serverTPS: math.MaxUint64,
+			config: GradualOrchestratorConfig{
+				MaxTPS:           2_000,
+				MinTPS:           1_000,
+				Step:             1_000,
+				TxRateMultiplier: 1.5,
+				SustainedTime:    time.Second,
+				MaxAttempts:      2,
+				Terminate:        true,
+			},
+		},
+		{
+			name:      "orchestrator TPS limited by network",
+			serverTPS: 1_000,
+			config: GradualOrchestratorConfig{
+				MaxTPS:           2_000,
+				MinTPS:           1_000,
+				Step:             1_000,
+				TxRateMultiplier: 1.3,
+				SustainedTime:    time.Second,
+				MaxAttempts:      2,
+				Terminate:        true,
+			},
+			expectedErr: ErrFailedToReachTargetTPS,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := require.New(t)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			tracker, err := NewPrometheusTracker[ids.ID](prometheus.NewRegistry())
+			r.NoError(err)
+
+			agents := []Agent[ids.ID, ids.ID]{
+				NewAgent(
+					&mockIssuer{
+						generateTxF: func() (ids.ID, error) {
+							return ids.GenerateTestID(), nil
+						},
+						tracker: tracker,
+						maxTxs:  tt.serverTPS,
+					},
+					&mockListener{},
+					tracker,
+				),
+			}
+
+			orchestrator, err := NewGradualOrchestrator(
+				agents,
+				logging.NoLog{},
+				tt.config,
+			)
+			r.NoError(err)
+
+			r.ErrorIs(orchestrator.Execute(ctx), tt.expectedErr)
+
+			if tt.expectedErr == nil {
+				r.GreaterOrEqual(orchestrator.GetMaxObservedTPS(), tt.config.MaxTPS)
+			} else {
+				r.Less(orchestrator.GetMaxObservedTPS(), tt.config.MaxTPS)
+			}
+		})
+	}
+}
+
+// test that the orchestrator returns early if the txGenerators or the issuers error
+func TestGradualOrchestratorExecution(t *testing.T) {
+	var (
+		errMockTxGenerator = errors.New("mock tx generator error")
+		errMockIssuer      = errors.New("mock issuer error")
+	)
+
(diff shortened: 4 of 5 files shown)


tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+	for _, agent := range o.agents {
+		o.issuerGroup.Go(func() error {
+			for {
+				if ctx.Err() != nil {
+					return nil //nolint:nilerr
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.agents))) * o.config.TxRateMultiplier)
+				// always listen until listener context is cancelled, do not call agent.Listener.IssuingDone().
+				for range txsPerIssuer {
```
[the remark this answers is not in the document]
qdm12: Ah AI at its peak again 😄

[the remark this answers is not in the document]
RodrigoVillar: lol

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+
+		prevConfirmed = currConfirmed
+		prevTime = currTime
+	}
+
+	return achievedTargetTPS
+}
+
+// GetObservedIssued returns the max TPS the orchestrator observed
+func (o *GradualOrchestrator[T, U]) GetMaxObservedTPS() uint64 {
```
qdm12, line 224: nit since this is used in the orchestrator test only, should we unexport the method or even remove and use `o.maxObservedTPS.Load()` directly in tests? Or would you see another use for it?

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+	return o.maxObservedTPS.Load()
+}
+
+// start a goroutine to each issuer to continuously send transactions
+// if an issuer errors, all other issuers will stop as well.
```
qdm12, line 229: nit adding a full stop otherwise it sounds "send transactions if an issuer errors" 😄 

```suggestion
// start a goroutine to each issuer to continuously send transactions.
// if an issuer errors, all other issuers will stop as well.
```

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+				for range txsPerIssuer {
+					tx, err := agent.Issuer.GenerateAndIssueTx(ctx)
+					if err != nil {
+						return err
+					}
+					agent.Listener.RegisterIssued(tx)
+				}
+				diff := time.Second - time.Since(currTime)
+				if diff > 0 {
+					time.Sleep(diff)
```
qdm12, line 249: nit let's use timers so this can be context-aware

```suggestion
					timer := time.NewTimer(diff)
					select {
					case <-ctx.Done():
						timer.Stop()
						return nil
					case <-timer.C:
					}
```

michaelkaplan13 (commented):
I compared this code with what's in `hypersdk/load` by checking out the repos side by side and running:
```
diff -ruN --color=always hypersdk/load avalanchego/tests/load
```

Everything LGTM, I was just wondering if we also want/need to bring over the [DefaultListener](https://github.com/ava-labs/hypersdk/blob/main/load/listener.go) implementation?

RodrigoVillar:
> Everything LGTM, I was just wondering if we also want/need to bring over the [DefaultListener](https://github.com/ava-labs/hypersdk/blob/main/load/listener.go) implementation?

The DefaultListener is VM specific (uses HyperSDK ws) and therefore, can't be used here.

qdm12:
> Everything LGTM, I was just wondering if we also want/need to bring over the [DefaultListener](https://github.com/ava-labs/hypersdk/blob/main/load/listener.go) implementation?

It uses hypersdk specific elements, so it should stay in hypersdk. Maybe we can find a way later on to abstract things further to use a common issuer/listener for both eth-based-evms and hypersdk, but for now it's rather different.

michaelkaplan13:
> nice migration 👍 Just left a few nits on code here and there, nothing of big importance.

Since the objective of this PR is to directly port over the existing framework implementation, I think we should defer any changes to the code for a different PR if needed.

michaelkaplan13:
Thanks both for answering my question!

michaelkaplan13 (approved):

tests/load/dependencies.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// GetObservedIssued returns the number of transactions that the tracker has
+	// confirmed were issued.
+	GetObservedIssued() uint64
+	// GetObservedConfirmed returns the number of transactions that the tracker has
+	// confirmed were accepted.
+	GetObservedConfirmed() uint64
+	// GetObservedFailed returns the number of transactions that the tracker has
+	// confirmed failed.
+	GetObservedFailed() uint64
```
joshua-kim, line 47: Why does this need to be an interface?

joshua-kim (dismissed):

tests/load/dependencies.go
```diff
@@ -0,0 +1,55 @@
...
+// See the file LICENSE for licensing terms.
+
+package load
+
+import "context"
+
+type Issuer[T comparable] interface {
+	// GenerateAndIssueTx generates and sends a tx to the network, and informs the
+	// tracker that it sent said transaction. It returns the sent transaction.
+	GenerateAndIssueTx(ctx context.Context) (tx T, err error)
```
joshua-kim, line 11: 1. We don't need named return parameters here - they're typically only used if we have multiple return values of the same type and we need to disambiguate
2. Why is this generic? Since `GenerateAndIssueTx` is responsible for the tx issuance, this package doesn't care about the tx concrete type at all afaict. Couldn't we just hard-code this to use `ids.ID`?

tests/load/dependencies.go
RodrigoVillar: VMs shouldn't have to be forced to use a single tracker implementation (i.e. they can choose between a tracker that does the bare minimum, a tracker with Prometheus metrics, etc.)

tests/load/burst_orchestrator.go
```diff
@@ -0,0 +1,89 @@
...
+			return nil
+		})
+	}
+
+	// wait for all issuers to finish sending their transactions
+	if err := issuerGroup.Wait(); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, o.config.Timeout)
```
joshua-kim, line 76: Why don't we just pass in this context into `Listen` instead of creating the `observerCtx`?

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+	// computing the average TPS.
+	SustainedTime time.Duration
+	// The number of attempts to try achieving a given target TPS before giving up.
+	MaxAttempts uint64
+
+	// Whether the orchestrator should return if the maxTPS has been reached
+	Terminate bool
+}
+
+func DefaultGradualOrchestratorConfig() GradualOrchestratorConfig {
```
joshua-kim, line 54: nit: call this `New*` instead of `Default*` - the caller implicitly knows you're getting defaults

tests/load/dependencies.go
joshua-kim: 1. If you don't want to use prometheus metrics - I think this might be inconsistent w/ the rest of the vm code since avalanchego uses prometheus metrics
2. If you want to disable prometheus metrics, we can still do this without defining this interface by accepting a `prometheus.Registerer`
3. Do we actually have any VMs using a different tracker implementation?

tests/load/dependencies.go
qdm12: 1. True, this may disambiguate what `T` is though. But yeah we could remove.
2. eth-based evms use a `*types.Transaction` and `common.Hash` whereas hypersdk uses `*chain.Transaction` and `ids.ID`

tests/load/dependencies.go
qdm12: I actually have a no-op tracker in another branch, which is needed for an orchestrator to distribute funds amongst keys before load testing, so interfacing does help 😉 We could accept a prometheus.Registerer though, that's true.

tests/load/dependencies.go
joshua-kim: I think an eth-based evm could just cast `common.Hash` into an `ids.ID` for compatibility

tests/load/README.MD
joshua-kim, line 1: nit: use `*.md`

tests/load/dependencies.go
joshua-kim, line 1: The file structure of this package is pretty weird (too granular)... in Go it's totally fine to have files that have more than one struct/interface since it doesn't have any one-public interface/class per file like java does.

tests/load/dependencies.go
joshua-kim: We could even use `any` if we had to keep the generic, the constraint on `comparable` specifically is strange because we don't do anything with that type information

tests/load/burst_orchestrator.go
```diff
@@ -0,0 +1,89 @@
...
+// transactions en masse in a short timeframe.
+type BurstOrchestrator[T comparable] struct {
+	agents []Agent[T]
+	log    logging.Logger
+
+	config BurstOrchestratorConfig
+}
+
+func NewBurstOrchestrator[T comparable](
+	agents []Agent[T],
```
joshua-kim, line 32: Why do we need to take in multiple instances of issuers/listeners? My guess is that tx generation/issuance isn't a bottleneck and we I think we can call listen asynchronously... so I'm wondering if we can just have a single one.

tests/load/dependencies.go
```diff
@@ -0,0 +1,55 @@
...
+	// Listen stops if the context is done, an error occurs, or if it received
+	// all the transactions issued and the issuer no longer issues any.
+	// Listen MUST return a nil error if the context is canceled.
+	Listen(ctx context.Context) error
+
+	// RegisterIssued informs the listener that a transaction was issued.
+	RegisterIssued(tx T)
+
+	// IssuingDone informs the listener that no more transactions will be issued.
+	IssuingDone()
```
joshua-kim, line 25: I'm wondering if we can simplify this interface. Can't we just have a single function like `Wait(ctx context.Context, tx T) error`? We could just call `Wait` after we issue each transaction, and block on all goroutines running `Wait` when we're doing issuing txs.

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+// GradualOrchestrator tests the network by continuously sending
+// transactions at a given rate (currTargetTPS) and increasing that rate until it detects that
+// the network can no longer make progress (i.e. the rate at the network accepts
+// transactions is less than currTargetTPS).
+type GradualOrchestrator[T, U comparable] struct {
+	agents  []Agent[T]
+	tracker Tracker[U]
```
joshua-kim, line 72: When is `T` and `U` expected to not be the same type?

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+
+	for {
+		// wait for the sustained time to pass or for the context to be cancelled
+		select {
+		case <-time.After(o.config.SustainedTime):
+		case <-issuerCtx.Done(): // the parent context was cancelled or an issuer errored
+		}
+
+		if issuerCtx.Err() != nil {
+			break // Case 1
```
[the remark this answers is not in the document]
joshua-kim: I think it references the godoc on the function (1/2/3)

StephenButtolph:
> Since the objective of this PR is to directly port over the existing framework implementation, I think we should defer any changes to the code for a different PR if needed.

Unless we are planning on putting this in the `x/` folder, I don't think we should be migrating code without a proper review

tests/load/dependencies.go
qdm12: > Hmm I think we could either force the caller to use an `ids.ID`, or if we want them to be able to define their own hash we could also just make a a tx interface...

The issuer in this case returns a `*types.Transaction` (eth) or a `*chain.Transaction` (hypersdk). 

> the constraint on comparable specifically is strange because we don't do anything with that type information

Good point, although it can be used more as "documentation" to prevent someone from implementing an issuer with not-comparable transactions, given it has to be comparable in the listener. We could use ~[32]byte though, that's interesting.

tests/load/gradual_orchestrator.go
qdm12: Good point, I think we can have both as the same type (transaction). Might had been me using common.Hash / ids.ID for the tracker, although we can use the transaction directly (as it is now)

tests/load/dependencies.go
qdm12: We don't want to issue->listen for each tx sequentially, we want to issue and listen asynchronously.
Or do you mean to run a goroutine per transaction issuance - if so, I don't think it would scale well and would make code possibly more complex I think. Unless I misunderstood what you mean? 🤔

tests/load/burst_orchestrator.go
qdm12: Actually, from my load testing, issuance on a single goroutine is not enough to bottleneck a node, I need 3 🤷  There is also another perspective, I tend to have a few milliseconds sleeps between issuances (with multiple agents) in order to not have the base fee go through the roof 😄

tests/load/dependencies.go
qdm12: I personally like how it is, so I would say this is a bit subjective. Separating implementation per file did help me figure out what's what when going through the code personally

tests/load/gradual_orchestrator.go
qdm12: I disagree with this, explicit code is better; `New*` is for constructors, and would suggest to me it's a new empty config with for example initialized empty maps.

RodrigoVillar:
Given the above, going to address the reviews above instead of the original "migration only" attempt

tests/load/dependencies.go
RodrigoVillar: Relaxed the generic type constraint here: ccaff101c46b0bd5543308ff09f8801c3d3e2057

Given the above comment that TXs may use different types for their ids (`ids.ID` vs `common.Hash`), I think it makes the most sense to use a generic here.

tests/load/burst_orchestrator.go
```diff
@@ -0,0 +1,89 @@
...
+type BurstOrchestratorConfig struct {
+	TxsPerIssuer uint64
+	Timeout      time.Duration
+}
+
+// BurstOrchestrator tests the network by sending a fixed number of
+// transactions en masse in a short timeframe.
+type BurstOrchestrator[T comparable] struct {
+	agents []Agent[T]
+	log    logging.Logger
```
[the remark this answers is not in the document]
RodrigoVillar: Done 855bfb92014a9ecd17e3ecc5c323daafd6a1b25f

tests/load/README.MD
RodrigoVillar: Done 855bfb920

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+					zap.Uint64("max TPS target", currTargetTPS.Load()),
+					zap.Uint64("average TPS", tps),
+				)
+				if o.config.Terminate {
+					o.log.Info("terminating orchestrator")
+					break // Case 2
+				} else {
+					o.log.Info("orchestrator will now continue running at max TPS")
+					continue
+				}
```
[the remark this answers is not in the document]
RodrigoVillar: Done b5faf4f3d

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,272 @@
...
+type GradualOrchestrator[T, U comparable] struct {
+	agents  []Agent[T]
+	tracker Tracker[U]
+
+	log logging.Logger
+
+	maxObservedTPS atomic.Uint64
+
+	observerGroup errgroup.Group
+	issuerGroup   *errgroup.Group
```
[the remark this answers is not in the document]
RodrigoVillar: `issuerGroup` is created via `errgroup.WithContext(ctx)` which returns a pointer. We could have `observerGroup` be a pointer as well but that would require us to initialize it. Happy to change `observerGroup` to a pointer if you think it makes sense.

tests/load/gradual_orchestrator.go
RodrigoVillar: I think that for TPS tests that live outside of `load/`, it might make sense to continue exporting this method.

tests/load/burst_orchestrator.go
RodrigoVillar: In addition to the above, if we want to attempt to simulate real-world performance, having `N` issuers/listeners is more realistic than having a single entity issue TXs.

tests/load/dependencies.go
RodrigoVillar: >  Do we have any VMs using a different tracker implementation w/ hypersdk?

@qdm12 can probably add to this, but I'm almost sure that the EVM performance tests will need to have more detailed metrics than the current `PrometheusTracker`

tests/load/burst_orchestrator.go
```diff
@@ -0,0 +1,89 @@
...
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+var _ orchestrator = (*BurstOrchestrator[any])(nil)
+
+type BurstOrchestratorConfig struct {
```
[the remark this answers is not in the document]
RodrigoVillar: Burst sends `X` transactions and waits for all of them to be confirmed while Gradual sends `Y` transactions in a given period and checks if the network achieved a given TPS (i.e. TX failures are allowed if we hit the local TPS target). Perhaps you could fit the Burst orchestrator into Gradual, but I think it's cleaner to leave them as separate.

tests/load/burst_orchestrator.go
```diff
@@ -21,21 +19,18 @@ type BurstOrchestratorConfig struct {
...
+func NewBurstOrchestrator[T any](
 	agents []Agent[T],
-	log logging.Logger,
 	config BurstOrchestratorConfig,
 ) (*BurstOrchestrator[T], error) {
 	return &BurstOrchestrator[T]{
 		agents: agents,
-		log:    log,
 		config: config,
 	}, nil
```
qdm12, line 35: the returned error is always nil for both orchestrators, maybe worth removing at this point in time?

tests/load/dependencies.go
```diff
@@ -27,7 +27,7 @@ type Listener[T comparable] interface {
 
 // Tracker keeps track of the status of transactions.
 // This must be thread-safe, so it can be called in parallel by the issuer(s) or orchestrator.
-type Tracker[T comparable] interface {
+type Tracker[T any] interface {
```
qdm12, line 30: I think this needs to be comparable so implementation can track T in maps etc. - as in it's more for the implementation aspect of it, true, but I think it's good to restrict this.

tests/load/gradual_orchestrator.go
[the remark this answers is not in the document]
qdm12: All good then, my bad!

tests/load/gradual_orchestrator.go
```diff
@@ -0,0 +1,277 @@
...
+				diff := time.Second - time.Since(currTime)
+				if diff > 0 {
+					timer := time.NewTimer(diff)
+					select {
+					case <-ctx.Done():
+						timer.Stop()
+						return nil
+					case <-timer.C:
+					}
+				}
```
qdm12, line 255: nit could flip the if condition to reduce indentation

```suggestion
				if diff <= 0 {
					continue
				}
				timer := time.NewTimer(diff)
				select {
				case <-ctx.Done():
					timer.Stop()
					return nil
				case <-timer.C:
				}
```

qdm12 (approved):

tests/load/gradual_orchestrator.go
qdm12: Can't we use the `T` as map keys though (tx pointer)? 🤔

tests/load/dependencies.go
qdm12: Ah interesting point. I did have a few extra metrics until I realised these were somehow unneeded, and I decided to match the existing tracker implementation 😄
I think having a single tracker implementation could be an interesting simplification, and we can accept a Registerer interface (so I can have my no-op tracker!)

tests/load/burst_orchestrator.go
joshua-kim: To clarify I'm not saying that having concurrent issuers is bad, I'm asking why that's something this package cares about. Can't you just push that into the implementation?

tests/load/gradual_orchestrator.go
joshua-kim: > New* is for constructors, and would suggest to me it's a new empty config with for example initialized empty maps.

This is a constructor. It is idiomatic in Go for a constructor to optionally define default values. As an example, the Go stdlib does this: https://cs.opensource.google/go/go/+/master:src/net/http/request.go;l=894;drc=ec7d955abf8bbaa6fdc65b460d957767e13e6378?q=http%2Frequest.go&ss=go%2Fgo

> I think this is true only if the caller reads the source code (which they shouldn't have to). Having the prefix Default makes it obvious to the caller they're getting default values.

It is obvious to the caller that this has defaults. In Go it is normal for raw data types to have safe zero values, so initializing it through a constructor means something is being done by the package for you.

tests/load/dependencies.go
joshua-kim: I recommend removing this interface and just accepting a `prometheus.Registerer` to create the metrics

tests/load/burst_orchestrator.go
RodrigoVillar: Done 672f377cf5812097d51ac83a71a16c0a97c418ff

tests/load/gradual_orchestrator.go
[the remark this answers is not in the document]
RodrigoVillar: Done 0595e310e678125943f76eb3224bff85a2730f6e

tests/load/burst_orchestrator.go
RodrigoVillar: One of the goals of a library like `load` is to minimize duplicate code. Both the HyperSDK and the EVM load tests require/will require multiple issuers which is the reason why this logic belongs in `BurstOrchestrator` rather than a responsibility of the user.

tests/load/burst_orchestrator.go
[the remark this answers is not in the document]
RodrigoVillar: If we were to merge the two, we would need test the burst case within gradual and so we'd still need tests regardless. 

> Are these two traffic patterns so different that we need to expose two separate types for them?

As mentioned, Burst does not allow for failures while Gradual does. Regarding cleaner implementations, merging Burst into Gradual would increase the amount of nesting we have within `Execute()` and `issueTxs()` (one case for unbounded TPS and a case for bounded TPS) which leads to less legible code IMO. Also, because issuers operate async from the TPS calculation logic, we would need to add a mechanism which ignores TPS and instead focuses on total TXs accepted. Since Burst and Gradual operate on different units of measurements (total txs vs TPS), they should be separate.

tests/load/burst_orchestrator.go
joshua-kim: > One of the goals of a library like load is to minimize duplicate code

Is it? The goals of a library should be to define clear abstractions that have clear responsibilities and boundaries. As an example, the go stdlib offers multiple ways to copy bytes (`io.Copy`, `copy`, etc). The cost for copying code is not always that bad, and the cost to implement a concurrent sender in this example I believe to be trivial. Even with the current approach the caller still needs to do some-level of coordination between the agents, since as an example you don't want to have multiple agents issuing duplicate txs.

tests/load/burst_orchestrator.go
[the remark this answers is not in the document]
joshua-kim: > If we were to merge the two, we would need test the burst case within gradual and so we'd still need tests regardless.

Tests change from parameters on the gradual type vs having the new concrete type to test. This also increases the package api surface

> As mentioned, Burst does not allow for failures while Gradual does.

This can be fixed by adding a parameter

> Regarding cleaner implementation....

It is better to have complexity hidden from the consumer of the package

tests/load/gradual_orchestrator.go
RodrigoVillar: The above might have been misleading; while you can use use `*chain.Transaction` as comparable, the issue arises with the listener. Consider the following:

1. We instantiate a new orchestrator with `T = *chain.Transaction`
2. This means that the `Listener` uses `*chain.Transaction`; that is, it'll receive txs via `Register Issued` and send txs to the `Tracker` via `ObserveConfirmed()`/`ObserveFailed()`. 
3. However, in the case of the HyperSDK, we have that `ListenTx()` on the WS client returns `ids.ID`. As a result, this fails because the tracker was expecting a value of type `*chain.Transaction`

tests/load/dependencies.go
qdm12: On the interfaces in the same dependencies.go: this can be problematic once an interface is used in multiple files - where do you place the interface then? I personally use an `interfaces.go` (dependencies.go here) to define all exported interfaces for a package, such that there is no ambiguity as to where interfaces should be.

tests/load/burst_orchestrator.go
qdm12: > Can't you just push that into the implementation?

Yes but also the "run multiple issuer+listener goroutines" sounds like it should be part of the orchestrating implementation I think 🤔 Moving the parallelism to the issuer and listener implementation would also transfer the complexity to the issuer and listener implementation, which should be implemented at least twice (eth + hypersdk), so I would prefer to have in the orchestrator code (implemented once for all issuer+listener implementations).

> Even with the current approach the caller still needs to do some-level of coordination between the agents

I don't think this is the case in the eth or hypersdk implementations. Coordination should be done at the issuer/listener constructing time, for example don't use the same keys - I don't think we would ever need to coordinate agents.

tests/load/dependencies.go
```diff
@@ -0,0 +1,55 @@
...
+// See the file LICENSE for licensing terms.
+
+package load
+
+import "context"
+
+type Issuer[T any] interface {
+	// GenerateAndIssueTx generates and sends a tx to the network, and informs the
+	// tracker that it sent said transaction. It returns the sent transaction.
+	GenerateAndIssueTx(ctx context.Context) (T, error)
```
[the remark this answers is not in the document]
qdm12: 1. The tracker is injected in the issuer and the issuer calls tracker.Issue(T) to notify the tracker when a tx has been issued. The tracker is also injected in the listener and the listener calls tracker.ObserveConfirmed(T) and tracker.ObserveFailed(T) as it progresses. Yes the implementation should have a tracker injected and handle the tracker updates. I suppose we could move the tracker updates in the orchestrating code:
    - For the issuer, it's easy, just call tracker.Issue(T) with the T returned from GenerateAndIssueTx
    - For the listener, we would need to change the interface, something like `IssueReceived(ctx context.Context) ([]T, error)` that the orchestrator would call in a for loop in another goroutine and block on, similarly to GenerateAndIssueTx. That would be interesting, and possibly remove the need for `IssuingDone()` (to check).
2. We could use an iterator, but I'm also happy to keep it simple as it is. Either way sounds good to me. Please not a channel though, this is a single thread situation there is absolutely no need to use a channel, especially now that we have iterators.

tests/load/dependencies.go
RodrigoVillar: Changed tracker from an interface to being an implementation: 781bf6082dc13aec971719cb97df5b3456097373

tests/load/gradual_orchestrator.go
RodrigoVillar: Update: with the single tracker implementation, we can use a single comparable generic on the gradual orchestrator; the issuer would need to return `ids.ID` which the listener can then use: b0ae321dc

tests/load/gradual_orchestrator.go
joshua-kim: 1. I agree leaving a comment is correct to document default overrides
2. I disagree w/ Arran's comment that the `Default` prefix is expected for this use-case. When the Go stdlib uses the default prefix, it's usually for a global default that can be overridden by the caller (e.g `log.Default()`, `http.DefaultTransport`). I can however find instances where `New` is creating an instance of something that has defaults. `http.NewRequest` uses http1.1 by default, `net.Dial` creates a Dialer that defaults to 300ms fallback delay. I think it is reasonable to use the `NewDefault` prefix if you need to disambiguate from another `New*` function in the package, but this is not the case here.

tests/load/dependencies.go
RodrigoVillar: Removed `dependencies.go`: 3ba6372

tests/load/gradual_orchestrator.go
RodrigoVillar: Thanks for the insight @ARR4N ! Added documentation to `NewOrchestratorConfig()` to make it clear the user is getting defaults: 9f9eebd66

tests/load/burst_orchestrator.go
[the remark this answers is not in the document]
RodrigoVillar: Merged burst into gradual: 4daa6b083

tests/load/burst_orchestrator.go
RodrigoVillar: I agree with @qdm12 's point that, as consumers of this PR, it would be beneficial to both the EVM and HyperSDK load tests if the orchestrator could handle parallelism on our behalf. I also took a stab at switching from the multiple agents per orchestrator to one agent per orchestrator, and found that the current model is better because it computes network TPS, rather than per-agent TPS under the one agent per orchestrator model. It is possible to compute network TPS under the one agent per orchestrator model, but the implementation looks awfully similar to what the orchestrator already does and it begs the question of whether we need to be computing TPS twice (per agent and per network) when we can just do it once under the multiple agents per orchestrator model.

tests/load/README.md
```diff
@@ -0,0 +1,129 @@
...
+transactions or marking them as failed. As it receives transactions, it
+notifies the tracker of the transaction status.
+
+### Tracker
+
+The tracker is responsible for maintaining metrics for all sent txs. Since the
+tracker is used by both the issuers, listeners and the orchestrator, all methods of the
+tracker must be thread safe.
+
+## Default Orchestrators
```
qdm12, line 77: This section needs to be updated:

- short burst is no longer present
- "gradual load" should be changed to just "Gradual orchestrator"
- Gradual orchestrator supports bursting with -1 as the maxTPS as far as I understood

tests/load/orchestrator.go
```diff
@@ -0,0 +1,315 @@
...
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.agents))) * o.config.TxRateMultiplier)
+				// always listen until listener context is cancelled, do not call agent.Listener.IssuingDone().
+				for range txsPerIssuer {
+					tx, err := agent.Issuer.GenerateAndIssueTx(ctx)
+					if err != nil {
+						return err
+					}
+					agent.Listener.RegisterIssued(tx)
```
qdm12, line 275: the orchestrator could call o.tracker.Issue(tx) instead of the issuer calling the tracker

tests/load/orchestrator.go
```diff
@@ -0,0 +1,315 @@
...
+		o.issuerGroup.Go(func() error {
+			for {
+				if ctx.Err() != nil {
+					return nil //nolint:nilerr
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.agents))) * o.config.TxRateMultiplier)
+				// always listen until listener context is cancelled, do not call agent.Listener.IssuingDone().
+				for range txsPerIssuer {
+					tx, err := agent.Issuer.GenerateAndIssueTx(ctx)
```
qdm12, line 271: nit rename tx to txID since it's now just an id

tests/load/orchestrator.go
```diff
@@ -0,0 +1,315 @@
...
+// if an issuer errors, all other issuers will stop as well.
+func (o *Orchestrator[_]) issueTxs(ctx context.Context, currTargetTPS *atomic.Int64) {
+	for _, agent := range o.agents {
+		o.issuerGroup.Go(func() error {
+			for {
+				if ctx.Err() != nil {
+					return nil //nolint:nilerr
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.agents))) * o.config.TxRateMultiplier)
```
qdm12, line 268:
```suggestion
				targetTPSPerAgent := math.Ceil(float64(currTargetTPS.Load()) / float64(len(o.agents)))
				txsPerIssuer := uint64(targetTPSPerAgent * o.config.TxRateMultiplier)
				if txsPerIssuer == uint64(targetTPSPerAgent) {
					// For low target TPS and rate multiplier, the rate multiplier can have no effect.
					// For example, for 50 TPS target, 5 agents, and rate multiplier of 1.05, we have
					// uint64(10*1.05) = 10 therefore we add 1 to force an increase in txs issued.
					txsPerIssuer++
				}
```

qdm12, line 268: Thinking more on this, it would be quite great if the user would not be required to set a mintps, maxtps, step, rate multiplier and stability duration. It could be an option to set a maxtps to have the test fail if it doesn't get reached, but at this point I think we're more interested in just running the node at its max TPS and get a max TPS number. We could have some logic to handle this, for example:

1. Set `targetTPS = 1`
2. On success, `targetTPS = 2 * targetTPS`
3. On the first failure reaching the TPS, set `previousTargetTPS = targetTPS/2`, `lastFailedTPS = targetTPS` then `targetTPS = (previousTargetTPS + lastFailedTPS) / 2`. Repeat this on further failures. On success, set `targetTPS = (targetTPS + lastFailedTPS) / 2`. Stop this loop when `targetTPS == lastFailedTPS - 1`.

To make this faster, you can set the stability timeout to i.e. 5s and increase it once approaching the final max TPS.

qdm12: After more thoughts, this would be a nice addition, but not really needed I suppose, we can figure out the max tps manually still.

tests/load/README.md
RodrigoVillar: Done 762ca7a

tests/load/orchestrator.go
RodrigoVillar: Done 0d8c344

PR #3945, opened 2025-05-07
[tmpnet] Define reusable flags for configuring kubernetes client access

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882
- - **>>>>>>** #3945 **<<<<<<**  
- #3947
- #3615
- #3794 

## Why this should be merged

Previously `tmpnetctl start-kind-cluster` and the bootstrap-monitor determined their kubeconfig and context separately. Simpler to do it one way and this will be reusable by the kube node runtime.

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

PR #3897, opened 2025-04-18
[tmpnet] Delegate writing of the flag file to the runtime

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- **>>>>>>** #3897 **<<<<<<**
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

This is intended to enable the process and kube runtimes to vary how they compose the flags that configure a given node.

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

## TODO

- [x] Merge parent #3896

PR #3890, opened 2025-04-17
[tmpnet] Rename NodeProcess to ProcessRuntime

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- **>>>>>>** #3890 **<<<<<<**
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

This is intended to be consistent with the addition of KubeRuntime and the name of the runtime configuration (ProcessRuntimeConfig).

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

## TODO

- [x] Merge #3881

PR #3884, opened 2025-04-13
[tmpnet] Switch FlagsMap from map[string]any to map[string]string

maru-ava:
## PR Chain: tmpnet+kube

This PR chain enables tmpnet to deploy temporary networks to Kubernetes. Early PRs refactor tmpnet to support the addition in #3615 of a new tmpnet node runtime for kube.  

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- **>>>>>>** #3884 **<<<<<<**
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3615
- #3794 

## Why this should be merged

`map[string]any` complicates comparison of values converted to strings (e.g. when node configuration is provided as env vars to a kube pod) to their original golang form. Switching `FlagsMap` to use `map[string]string` forces the decision of how to serialize on the developer to ensure that all subsequent usage will be consistent.

This change also updates subnet, chain and monitor configuration to stop using `FlagsMap` since their use-cases still require `map[string]any`.

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

## TODO

- [x] Merge #3890

PR #3947, opened 2025-05-08
[tmpnet] Separate start of prometheus and promtail collectors

maru-ava:
- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3945
- **>>>>>>** #3947 **<<<<<<** 
- #3615
- #3794

## Why this should be merged

Previously local collectors were always started together. For kube, though, it makes sense to support only starting prometheus (e.g. to collect metrics from the load generation framework) since logs and metrics for nodes will be collected by agents within the kube cluster.

## How this was tested

CI

N/A


qdm12 (approved):
Nice catch 👍 
Except for retro-compatibility, do we want to continue supporting `TMPNET_START_COLLECTORS`? Should we just support log+metrics collection flags to have code a bit simpler? But it's also fine as is, so feel free to ignore.

StephenButtolph (approved):

PR #3949, opened 2025-05-08
refactor genesis building logic in avm and platformvm

### LINKED ISSUE #2613: Refactor static service structs into static functions
After https://github.com/ava-labs/avalanchego/pull/2589, we no longer need the static service structs and can expose creating genesis files via a static function that isn't bound to a struct.

DracoLi:
## Why this should be merged

Resolves #2613, we no longer need static structs to create genesis.

## How this works

- Removes `StaticService` from both `plaftformvm` and `avm`
- Move logic for building genesis to `BuildGenesis`

## How this was tested

unit test


vms/avm/genesis_builder.go
```diff
@@ -88,7 +80,7 @@ type BuildGenesisReply struct {
 
 // BuildGenesis returns the UTXOs such that at least one address in [args.Addresses] is
 // referenced in the UTXO.
-func (*StaticService) BuildGenesis(_ *http.Request, args *BuildGenesisArgs, reply *BuildGenesisReply) error {
+func BuildGenesis(args *BuildGenesisArgs, reply *BuildGenesisReply) error {
```
joshua-kim, line 83: Can we also get rid of all these args/reply structs and just make them parameters/return values/structs?

DracoLi: yep good point, i'll update this

vms/platformvm/api/validator.go
joshua-kim, line 1: Even though this is the old `static_service` this is showing up as a remove/add in the diff instead of a file rename. Can we try to see if we can preserve the git history here? I think we might need to re-do the diff where we have a separate commit for renaming the file.

vms/platformvm/genesis/genesis_builder.go
joshua-kim, line 1: Can we just merge this into `genesis.go`?

vms/platformvm/genesis/genesis_builder.go
```diff
@@ -0,0 +1,329 @@
...
+// [avaxAssetID] is the ID of the AVAX asset
+// [networkID] is the ID of the network
+// [allocations] are the UTXOs on the Platform Chain that exist at genesis.
+// [validators] are the validators of the primary network at genesis.
+// [chains] are the chains that exist at genesis.
+// [time] is the Platform Chain's time at network genesis.
+// [initialSupply] is the initial supply of the AVAX asset.
+// [message] is the message to be sent to the genesis UTXOs.
+// [encoding] is the encoding of the message & chains genesis data.
+func BuildGenesis(
```
joshua-kim, line 128: nit: rename to `genesis.New` (just name all constructors `New*` to avoid confusion)

vms/platformvm/api/validator.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// Staker is the representation of a staker sent via APIs.
+// [TxID] is the txID of the transaction that added this staker.
+// [Weight] is the amount of tokens being staked.
+// [StartTime] is the Unix time when they start staking
+// [Endtime] is the Unix time repr. of when they are done staking
+// [NodeID] is the node ID of the staker
+type Staker struct {
+	TxID      ids.ID      `json:"txID"`
```
joshua-kim, line 19: Do we still need these json serialization tags?

vms/platformvm/genesis/genesis_builder.go
```diff
@@ -0,0 +1,329 @@
...
+// Allocation is a UTXO on the Platform Chain that exists at the chain's genesis
+type Allocation struct {
+	Locktime uint64
+	Amount   uint64
+	Address  string
+	Message  string
+}
+
+// Compare compares two allocations
+func (utxo Allocation) Compare(other Allocation) int {
```
joshua-kim, line 48: nit: rename receiver to `a` (Go convention is first letter of the type)

vms/avm/static_service.go
joshua-kim, line 1: Same comment w.r.t making this a rename instead of a delete/new file if it's possible to do

vms/avm/genesis_builder.go
joshua-kim, line 1: Same comment w.r.t moving this to `genesis.go`

vms/avm/genesis_builder_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/formatting"
+	"github.com/ava-labs/avalanchego/utils/formatting/address"
+)
+
+var addrStrArray = []string{
```
joshua-kim, line 17: nit: move this to inside `TestBuildGenesis`. Shared information across tests tend to add more confusion imo.

vms/platformvm/genesis/genesis_builder.go
```diff
@@ -0,0 +1,329 @@
...
+	avaxAssetID ids.ID,
+	networkID uint32,
+	allocations []Allocation,
+	validators []GenesisPermissionlessValidator,
+	chains []Chain,
+	time uint64,
+	initialSupply uint64,
+	message string,
+	encoding formatting.Encoding,
+) (string, error) {
```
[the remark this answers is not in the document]
DracoLi: yeah i agree - i think we can probably just return the bytes, since that's what its being used right now.
Maybe also just change this function to `NewGenesisBytes`?

vms/platformvm/api/validator.go
DracoLi: yeah, this is used in the `GetCurrentValidators` API in platformvm

vms/platformvm/genesis/genesis_builder.go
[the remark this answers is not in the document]
joshua-kim: I think that sounds reasonable. I think returning the concrete genesis type is overkill for what we need right now since none of our tests depend on anything other than the bytes

vms/platformvm/api/validator.go
DracoLi: i was trying to get this to work before but no luck because of the amount of changes (git think its too different to be a rename).
But I've prepared the first commit to have no file changes to help with review - https://github.com/ava-labs/avalanchego/pull/3949/commits/ee89fc09a660071722d21c558359af1d051e4719

vms/platformvm/genesis/genesis.go
```diff
@@ -41,3 +62,291 @@ func Parse(genesisBytes []byte) (*Genesis, error) {
...
+// Allocation is a UTXO on the Platform Chain that exists at the chain's genesis
+type Allocation struct {
+	Locktime uint64
+	Amount   uint64
+	Address  string
+	Message  []byte
+}
+
+// Compare compares two allocations
+func (utxo Allocation) Compare(a Allocation) int {
```
joshua-kim, line 75: I thought I left this comment on my last review but it looks like I forgot to post it... can we update the receiver name since we renamed the struct (UTXO -> Allocation) to avoid the conflict with the existing UTXO type? Generally the Go convention is the first letter of the type, so `a` would be suitable here.

DracoLi: yep I saw the comment but forgot to make the change. thx for spotting this.

joshua-kim (approved):

vms/platformvm/api/validator.go
yacovm: I think it's much easier to review the code if we don't change the file names.
Can we defer changing the file names and moving code around to a different PR after this is merged or rebased on top of this one?

yacovm: > git think its too different to be a rename

I think the right strategy would be to first refactor keeping the files, and only then renaming in a different PR.

DracoLi: sounds good - let me try to separate this into two PRs

genesis/genesis.go
```diff
@@ -390,23 +368,19 @@ func FromConfig(config *Config) ([]byte, ids.ID, error) {
 		}
 		for _, unlock := range allocation.UnlockSchedule {
 			if unlock.Amount > 0 {
-				msgStr, err := formatting.Encode(defaultEncoding, allocation.ETHAddr.Bytes())
```
yacovm, line 393: why is this removed?

genesis/genesis.go
```diff
@@ -420,55 +394,46 @@ func FromConfig(config *Config) ([]byte, ids.ID, error) {
...
 
-		utxos := []api.UTXO(nil)
+		allocations := []api.Allocation(nil)
 		for _, allocation := range nodeAllocations {
 			addr, err := address.FormatBech32(hrp, allocation.AVAXAddr.Bytes())
 			if err != nil {
 				return nil, ids.Empty, err
 			}
 			for _, unlock := range allocation.UnlockSchedule {
-				msgStr, err := formatting.Encode(defaultEncoding, allocation.ETHAddr.Bytes())
```
yacovm, line 430: why is this removed?

vms/platformvm/api/static_service.go
```diff
@@ -36,30 +33,27 @@ var (
...
-// StaticService defines the static API methods exposed by the platform VM
-type StaticService struct{}
-
-// UTXO is a UTXO on the Platform Chain that exists at the chain's genesis.
-type UTXO struct {
-	Locktime json.Uint64 `json:"locktime"`
-	Amount   json.Uint64 `json:"amount"`
-	Address  string      `json:"address"`
-	Message  string      `json:"message"`
+// Allocation is a UTXO on the Platform Chain that exists at the chain's genesis
```
yacovm, line 39: Why the renaming?

vms/platformvm/api/validator.go
```diff
@@ -0,0 +1,73 @@
...
+
+package api
+
+import (
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/json"
+	"github.com/ava-labs/avalanchego/vms/platformvm/signer"
+)
+
+// Staker is the representation of a staker sent via APIs.
```
yacovm, line 12: Is it possible to keep these definitions in the static service file?

yacovm: and later rename it as we need

vms/platformvm/api/static_service.go
DracoLi: because of type conflict. `platformvm/genesis` has UXTO already. Since this type is specifically for creating the utxo when building genesis so I renamed it to be a bit more clear.

vms/platformvm/api/validator.go
DracoLi: these don't belong either in static service or genesis now. It's types used by the platformvm api in `vms/platformvm/service.go` so I think we can just extract it in this PR.

genesis/genesis.go
DracoLi: `defaultEncoding` is no longer needed for this functions. So no we no longer need to encode it the bytes. Same for above.

yacovm (approved):

vms/platformvm/api/static_service.go
```diff
@@ -350,59 +304,50 @@ func (*StaticService) BuildGenesis(_ *http.Request, args *BuildGenesisArgs, repl
...
-	}
-	reply.Bytes, err = formatting.Encode(args.Encoding, bytes)
+	genesisBytes, err := genesis.Codec.Marshal(genesis.CodecVersion, g)
 	if err != nil {
-		return fmt.Errorf("couldn't encode genesis as string: %w", err)
+		return nil, fmt.Errorf("couldn't marshal genesis: %w", err)
 	}
-	reply.Encoding = args.Encoding
-	return nil
+	return genesisBytes, nil
```
StephenButtolph, line 352: Similar comment to the AVM - I feel like this function would best return the actual genesis rather than the bytes

vms/platformvm/api/static_service.go
```diff
@@ -209,122 +149,136 @@ func bech32ToID(addrStr string) (ids.ShortID, error) {
...
+// NewGenesisBytes builds the genesis state of the P-Chain (and thereby the Avalanche network.)
+// [avaxAssetID] is the ID of the AVAX asset
+// [networkID] is the ID of the network
+// [allocations] are the UTXOs on the Platform Chain that exist at genesis.
+// [validators] are the validators of the primary network at genesis.
+// [chains] are the chains that exist at genesis.
+// [time] is the Platform Chain's time at network genesis.
+// [initialSupply] is the initial supply of the AVAX asset.
+// [message] is the message to be sent to the genesis UTXOs.
+// [encoding] is the encoding of the message & chains genesis data.
```
StephenButtolph, line 161:
```suggestion
```

vms/avm/static_service.go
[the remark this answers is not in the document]
yacovm: it will be renamed [here](https://github.com/ava-labs/avalanchego/pull/3963/commits/ad7851438fa397dc6739220a4fa16c30c2d40347#diff-94b54f961b2d58a7af99b4c4143e80eafa14274dc79ab74d01fa64975facfca1). Doing it in two steps simplifies the diff.

vms/platformvm/genesis/genesis.go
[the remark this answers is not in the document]
yacovm: renamed [here](https://github.com/ava-labs/avalanchego/pull/3963/commits/ad7851438fa397dc6739220a4fa16c30c2d40347#diff-193b110e31cff9b2ad6201c2778bfd289cb4f07c3daa81db97a8cf0a79c506f8).

StephenButtolph (approved):

Issue #3948, opened 2025-05-08
Failed to Repair ProposerVM Height after C-Chain State Sync on Fuji


aaronbuchwald:
Hitting this error after state syncing the C-Chain on Fuji:

```
[05-07|19:46:09.772] FATAL <C Chain> handler/handler.go:312 shutting down chain {"reason": "received an unexpected error", "error": "failed to notify VM that bootstrapping has started: failed to repair accepted chain height: proposervm height index (40288256) should never be lower than the inner height index (40297478) while processing sync message: get_accepted_state_summary_failed from NodeID-PT9g4vX1UAEE869Fed3kMFNR6Xjyx3RPb"}
```

After restart it fails a second time with:

```
[05-07|19:49:48.180] FATAL chains/manager.go:392 error creating required chain {"subnetID": "11111111111111111111111111111111LpoYY", "chainID": "yH8D7ThNJkxmtkuv2jgBa4P1Rn3Qpr4pPr7QYNfcdoS6k6HWp", "vmID": "mgj786NP7uDwBCcq6YwThhaN8FLyybkCa4zBWTQbNgmK6k9A6", "error": "error while creating new snowman vm failed to repair accepted chain by height: proposervm height index (40288256) should never be lower than the inner height index (40297478)"}
```

This appears to be hitting the same underlying error on two paths (`SetState` at the end of state sync and `Initialize` on restart).

The actual error occurs here: https://github.com/ava-labs/avalanchego/blob/a90944df3fc8a413e5cf14d7ba0c305899d47834/vms/proposervm/vm.go#L466

This indicates that the ProposerVM and Coreth end up breaking an invariant that the ProposerVM's last accepted height _must_ always be >= the inner VM's last accepted height.

Likely related to https://github.com/ava-labs/avalanchego/pull/3831

aaronbuchwald:
This occurs first at the end of static state sync of the C-Chain. The inner VM's expected behavior is to either skip state sync because the proposed state summary is too close to the tip (return `block.StateSyncSkipped`) or to handle the state sync summary, return `block.StateSyncStatic` and then start fetching the state summary to eventually mark that block as accepted.

The cause could be:
- proposervm accepts state summary at height 2000 and updates its last accepted block
- innerVM skips the state summary because it is too close to tip, so it keeps a last accepted block of 2001
- engine handles the skipped state sync and attempts to set the engine to the next state
- proposervm sees the broken invariant where proposervm height is lower than the innerVM height

[...]

PR #3950, opened 2025-05-08
Update proposervm summary to roll forward only

aaronbuchwald:
## Why this should be merged

This PR fixes https://github.com/ava-labs/avalanchego/issues/3948 caused by https://github.com/ava-labs/avalanchego/pull/3831 violating an invariant that the ProposerVM index must always be >= the innerVM index.

## How this works

Rather than the proposerVM always accepting the state sync summary, this PR ensures the ProposerVM index only moves forward. This means that if the accepted state sync summary is below the proposerVM's last accepted height, it allows the ProposerVM and InnerVM to diverge.

This will be corrected when state sync finishes and `SetState` is called and the ProposerVM re-aligns the two. Since we keep the ProposerVM roll forward only, we are still able to roll it back and guarantee that they are re-aligned before we entre consensus.

## How this was tested

Existing unit tests

No

### DIFF
--- a/vms/proposervm/state_summary.go
+++ b/vms/proposervm/state_summary.go
@@ -46,11 +46,16 @@ func (s *stateSummary) Accept(ctx context.Context) (block.StateSyncMode, error) 
 		return block.StateSyncSkipped, err
 	}
 
-	// We store the full proposerVM block associated with the summary
-	// and update height index with it, so that state sync could resume
-	// after a shutdown.
-	if err := s.block.acceptOuterBlk(); err != nil {
-		return block.StateSyncSkipped, err
+	// Mark the summary as accepted on the outerVM iff it rolls forward.
+	// We refuse to roll the proposerVM backward because it violates the invariant
+	// that the proposerVM index is always >= the innerVM index.
+	if s.vm.lastAcceptedHeight < s.Height() {
+		// We store the full proposerVM block associated with the summary
+		// and update height index with it, so that state sync could resume
+		// after a shutdown.
+		if err := s.block.acceptOuterBlk(); err != nil {
+			return block.StateSyncSkipped, err
+		}
 	}
 
 	// innerSummary.Accept may fail with the proposerVM block and index already
--- a/vms/proposervm/state_syncable_vm_test.go
+++ b/vms/proposervm/state_syncable_vm_test.go
@@ -649,7 +649,7 @@ func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
 			return nil, database.ErrNotFound
 		}
 	}
-	innerVM.GetStateSummaryF = func(_ context.Context, h uint64) (block.StateSummary, error) {
+	innerVM.GetStateSummaryF = func(context.Context, uint64) (block.StateSummary, error) {
 		return innerSummary1, nil
 	}
 	innerVM.ParseBlockF = func(_ context.Context, b []byte) (snowman.Block, error) {
@@ -716,8 +716,8 @@ func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
 	require.NoError(err)
 	require.Equal(block.StateSyncSkipped, status)
 	require.True(calledInnerAccept)
-
 	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+
 	require.Equal(innerBlk2.Height(), vm.lastAcceptedHeight)
 	lastAcceptedID, err := vm.LastAccepted(context.Background())
 	require.NoError(err)
--- a/vms/proposervm/state_syncable_vm_test.go
+++ b/vms/proposervm/state_syncable_vm_test.go
@@ -595,7 +595,12 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
 	require.Equal(proBlk.ID(), lastAcceptedID)
 }
 
-func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
+// TestStateSummaryAcceptOlderBlockSkipStateSync tests the case where we accept
+// a state summary older than the last accepted block. In this case, we should not
+// roll the ProposerVM back to match the state summary, but we should invoke the
+// innerVM to accept the state summary and re-align the ProposerVM with the innerVM
+// during the transition out of state sync.
+func TestStateSummaryAcceptOlderBlockSkipStateSync(t *testing.T) {
 	require := require.New(t)
 
 	innerVM, vm := helperBuildStateSyncTestObjects(t)
@@ -612,14 +617,7 @@ func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
 		ParentV: ids.GenerateTestID(),
 		HeightV: 1969,
 	}
-	innerBlk2 := &snowmantest.Block{
-		Decidable: snowtest.Decidable{
-			IDV: ids.GenerateTestID(),
-		},
-		BytesV:  []byte{2},
-		ParentV: innerBlk1.ID(),
-		HeightV: 1970,
-	}
+	innerBlk2 := snowmantest.BuildChild(innerBlk1)
 
 	innerSummary1 := &blocktest.StateSummary{
 		IDV:     innerBlk1.ID(),
@@ -638,12 +636,12 @@ func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
 	}
 
 	innerVM.GetBlockF = func(_ context.Context, blkID ids.ID) (snowman.Block, error) {
-		switch {
-		case blkID == snowmantest.GenesisID:
+		switch blkID {
+		case snowmantest.GenesisID:
 			return snowmantest.Genesis, nil
-		case blkID == innerBlk1.ID():
+		case innerBlk1.ID():
 			return innerBlk1, nil
-		case blkID == innerBlk2.ID():
+		case innerBlk2.ID():
 			return innerBlk2, nil
 		default:
 			return nil, database.ErrNotFound
@@ -712,6 +710,10 @@ func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
 	require.Equal(summary.Height(), innerBlk1.Height())
 
 	// Process a state summary that would rewind the chain
+	// ProposerVM should ignore the rollback and accept the inner state summary to
+	// notify the innerVM.
+	// This can result in the ProposerVM and innerVM diverging their last accepted block,
+	// which should be rolled forward and re-aligned in SetState.
 	status, err := summary.Accept(context.Background())
 	require.NoError(err)
 	require.Equal(block.StateSyncSkipped, status)
--- a/vms/proposervm/state_syncable_vm_test.go
+++ b/vms/proposervm/state_syncable_vm_test.go
@@ -529,7 +529,7 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
 
 	require.NoError(vm.SetForkHeight(innerSummary.Height() - 1))
 
-	// Set the last accepted block height to be higher that the state summary
+	// Set the last accepted block height to be higher than the state summary
 	// we are going to attempt to accept
 	vm.lastAcceptedHeight = innerSummary.Height() + 1
 
@@ -627,7 +627,7 @@ func TestStateSummaryAcceptOlderBlockSkipStateSync(t *testing.T) {
 
 	require.NoError(vm.SetForkHeight(innerSummary1.Height() - 1))
 
-	// Set the last accepted block height to be higher that the state summary
+	// Set the last accepted block height to be higher than the state summary
 	// we are going to attempt to accept
 	vm.lastAcceptedHeight = innerBlk2.Height()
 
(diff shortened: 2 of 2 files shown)


vms/proposervm/state_syncable_vm_test.go
```diff
@@ -52,7 +53,10 @@ func helperBuildStateSyncTestObjects(t *testing.T) (*fullVM, *VM) {
 	innerVM.LastAcceptedF = snowmantest.MakeLastAcceptedBlockF(
 		[]*snowmantest.Block{snowmantest.Genesis},
 	)
-	innerVM.GetBlockF = func(context.Context, ids.ID) (snowman.Block, error) {
+	innerVM.GetBlockF = func(_ context.Context, blkID ids.ID) (snowman.Block, error) {
+		if blkID != snowmantest.Genesis.ID() {
+			return nil, database.ErrNotFound
+		}
```
aaronbuchwald, line 59: Ensure that the default test setup does not silently ignore the `blkID` parameter and return the genesis

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,132 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		HeightV: 1969,
+	}
+	innerBlk2 := &snowmantest.Block{
+		Decidable: snowtest.Decidable{
+			IDV: ids.GenerateTestID(),
+		},
+		BytesV:  []byte{2},
+		ParentV: innerBlk1.ID(),
+		HeightV: 1970,
+	}
```
StephenButtolph, line 622:
```suggestion
	innerBlk2 := snowmantest.BuildChild(innerBlk1)
```

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,132 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		return innerBlk2.IDV, nil
+	}
+
+	innerVM.GetBlockF = func(_ context.Context, blkID ids.ID) (snowman.Block, error) {
+		switch {
+		case blkID == snowmantest.GenesisID:
+			return snowmantest.Genesis, nil
+		case blkID == innerBlk1.ID():
+			return innerBlk1, nil
+		case blkID == innerBlk2.ID():
```
StephenButtolph, line 646:
```suggestion
		switch blkID {
		case snowmantest.GenesisID:
			return snowmantest.Genesis, nil
		case innerBlk1.ID():
			return innerBlk1, nil
		case innerBlk2.ID():
```

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,132 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
 	require.NoError(err)
 	require.Equal(proBlk.ID(), lastAcceptedID)
 }
+
+func TestStateSummaryAcceptSkipOlderBlockInnerVM(t *testing.T) {
```
StephenButtolph, line 598: We should document what edge case this is actually testing.

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,132 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+	status, err := summary.Accept(context.Background())
+	require.NoError(err)
+	require.Equal(block.StateSyncSkipped, status)
+	require.True(calledInnerAccept)
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+
+	require.Equal(innerBlk2.Height(), vm.lastAcceptedHeight)
+	lastAcceptedID, err := vm.LastAccepted(context.Background())
+	require.NoError(err)
+	require.Equal(proBlk2.ID(), lastAcceptedID)
```
StephenButtolph, line 724: We should comment what we expect to be happening here

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+
+	summary, err := vm.GetStateSummary(context.Background(), innerBlk1.Height())
+	require.NoError(err)
+	require.Equal(summary.Height(), innerBlk1.Height())
+
+	// Process a state summary that would rewind the chain
+	// ProposerVM should ignore the rollback and accept the inner state summary to
+	// notify the innerVM.
+	// This can result in the ProposerVM and innerVM diverging their last accepted block,
+	// which should be rolled forward and re-aligned in SetState.
```
StephenButtolph, line 716: We do not roll forward. If the innerVM rolls back, the proposervm will roll back. This PR is ensuring that the proposervm _only_ rolls back if the innerVM rolls back.

aaronbuchwald, line 716:
```suggestion
	// This can result in the ProposerVM and innerVM diverging their last accepted block.
	// These are re-aligned in SetState before transitioning to consensus.
```

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		return innerSummary1, nil
+	}
+	innerVM.ParseBlockF = func(_ context.Context, b []byte) (snowman.Block, error) {
+		switch {
+		case bytes.Equal(b, innerBlk1.BytesV):
+			return innerBlk1, nil
+		case bytes.Equal(b, innerBlk2.BytesV):
+			return innerBlk2, nil
+		default:
+			panic("unexpected parse block")
```
joshua-kim, line 660: Can we err instead of panic'ing here

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+	calledInnerAccept := false
+	innerSummary1.AcceptF = func(context.Context) (block.StateSyncMode, error) {
+		calledInnerAccept = true
+		return block.StateSyncSkipped, nil
+	}
+
+	slb1, err := statelessblock.Build(
+		vm.preferred,
+		innerBlk1.Timestamp(),
+		100, // pChainHeight,
```
joshua-kim, line 673: These comments documenting the parameter name isn't needed since IDEs already can be configured to display parameter hints

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		}
+	}
+	// Setup innerSummary.AcceptF to return StateSyncSkipped and check it was called
+	calledInnerAccept := false
+	innerSummary1.AcceptF = func(context.Context) (block.StateSyncMode, error) {
+		calledInnerAccept = true
+		return block.StateSyncSkipped, nil
+	}
+
+	slb1, err := statelessblock.Build(
```
joshua-kim, line 670: Can we just call this `blk` or `statelessBlk` of `slb`? I don't think we've used this naming before

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		switch {
+		case bytes.Equal(b, innerBlk1.BytesV):
+			return innerBlk1, nil
+		case bytes.Equal(b, innerBlk2.BytesV):
+			return innerBlk2, nil
+		default:
+			panic("unexpected parse block")
+		}
+	}
+	// Setup innerSummary.AcceptF to return StateSyncSkipped and check it was called
```
joshua-kim, line 663: This comment can be removed/is overly verbose

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		postForkCommonComponents: postForkCommonComponents{
+			vm:       vm,
+			innerBlk: innerBlk2,
+		},
+	}
+	require.NoError(vm.acceptPostForkBlock(proBlk2))
+
+	summary, err := vm.GetStateSummary(context.Background(), innerBlk1.Height())
+	require.NoError(err)
+	require.Equal(summary.Height(), innerBlk1.Height())
```
joshua-kim, line 710: Shouldn't the order be reversed here? I think `innerBlk1` is actually the want value, not the got value

vms/proposervm/state_summary.go
```diff
@@ -46,11 +46,16 @@ func (s *stateSummary) Accept(ctx context.Context) (block.StateSyncMode, error)
...
+	// Mark the summary as accepted on the outerVM iff it rolls forward.
+	// We refuse to roll the proposerVM backward because it violates the invariant
+	// that the proposerVM index is always >= the innerVM index.
+	if s.vm.lastAcceptedHeight < s.Height() {
+		// We store the full proposerVM block associated with the summary
+		// and update height index with it, so that state sync could resume
+		// after a shutdown.
+		if err := s.block.acceptOuterBlk(); err != nil {
+			return block.StateSyncSkipped, err
+		}
```
yacovm, line 58: what happens if `err` is `nil`? We continue to accept the inner summary, and assume that now `s.vm.lastAcceptedHeight` is `s.Height()` ?

yacovm: ah yes that is what `acceptOuterBlk` is doing

vms/proposervm/state_syncable_vm_test.go
aaronbuchwald: We use this naming convention throughout the rest of the file. The only difference here is that we need two blocks and use 1/2 suffixes.

vms/proposervm/state_summary.go
aaronbuchwald: If we hit an error handling the outer block, we return a fatal error early.

Otherwise, we want to ensure that we'll invoke `s.innerSummary.Accept(ctx)` no matter what. We only update the proposervm index if it moves the index forward (we refuse to roll the proposervm index backwards).

Yes, accepting the outer block will update `s.vm.lastAcceptedHeight = s.Height()` https://github.com/ava-labs/avalanchego/blob/6e04bc0ac1d3d3cd1d0bf17c55a4eda41d094e88/vms/proposervm/vm.go#L651.

The innerVM is expected to update its height to match the summary after it completes state sync (more nuance here I won't go into for static vs. dynamic state sync). The important bit is that they should be aligned when we transition to normal operation, which is taken care of in `SetState` / `repairAcceptedChainByHeight`: https://github.com/ava-labs/avalanchego/blob/6e04bc0ac1d3d3cd1d0bf17c55a4eda41d094e88/vms/proposervm/vm.go#L277

`repairAcceptedChainByHeight` assumes the proposervm index is ahead of the innerVM (guaranteed by the order we apply accept on both blocks and the summary) and updates the proposervm last accepted index to match the innerVM. This handles the case that the innerVM died during accept. Since the proposervm has the block, we could alternatively invoke accept on the innerVM to move it forward, but the current behavior is to roll the ProposerVM back.

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,134 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+
+	innerSummary1 := &blocktest.StateSummary{
+		IDV:     innerBlk1.ID(),
+		HeightV: innerBlk1.Height(),
+		BytesV:  innerBlk1.BytesV,
+	}
+
+	require.NoError(vm.SetForkHeight(innerSummary1.Height() - 1))
+
+	// Set the last accepted block height to be higher that the state summary
```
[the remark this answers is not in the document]
aaronbuchwald: Nice catch, this was copied from the other test, will fix both.

vms/proposervm/state_syncable_vm_test.go
aaronbuchwald: Updated

vms/proposervm/state_syncable_vm_test.go
aaronbuchwald: Removed

vms/proposervm/state_syncable_vm_test.go
aaronbuchwald: Switched

yacovm (approved):

vms/proposervm/state_syncable_vm_test.go
```diff
@@ -590,3 +594,136 @@ func TestStateSummaryAcceptOlderBlock(t *testing.T) {
...
+		return innerSummary1, nil
+	}
+	innerVM.ParseBlockF = func(_ context.Context, b []byte) (snowman.Block, error) {
+		switch {
+		case bytes.Equal(b, innerBlk1.BytesV):
+			return innerBlk1, nil
+		case bytes.Equal(b, innerBlk2.BytesV):
+			return innerBlk2, nil
+		default:
+			require.FailNow("unexpected parse block")
```
joshua-kim, line 660: I originally had in mind returning an error here instead of using `FailNow` to avoid the unreachable code warning... does that not work beacuse of the way this is called for some reason?

aaronbuchwald: This uses `require.FailNow(...)` because it's unexpected for this to ever parse an invalid block.

The way these tests are written we don't have an actual serialization format we just hardcode bytes and then set `ParseBlock` to match that to the block we hardcoded it to (ie. `bytes.Equal(b, innerBlk2.BytesV)`). If we see bytes that we don't recognize here, we could return an error, but it would be hiding some code path that should never occur.

joshua-kim (approved):

StephenButtolph:
I tried rebasing the PR rather than introducing a merge comment (because merge commits mark me as an author).

Won't do that again...

PR #3953, opened 2025-05-10
Revert "Make sure inner state summary accept is called (#3831)"

ceyonur:
This reverts commit 428a4113b154288879b046c8249a5d1ddbd422be.

Testing only


StephenButtolph:
This regression was fixed.

PR #3956, opened 2025-05-12
fix(ids): `%x` formatting of `ID` and `ShortID`

ARR4N:
## Why this should be merged

`fmt.Sprintf("%x", ids.ID{})` returns the hex encoding of the `cb58` encoding, not of the raw bytes.

## How this works

Implement the `fmt.Formatter` interface for both `ids.ID` and `ids.ShortID`.

## How this was tested

Unit test of formatting with `%v`, `%s` and `%q` (equivalent to `String()`), plus `%x` and `%#x` (equivalent to `Hex()`).

Not sure. Maybe. I dunno `¯\_(ツ)_/¯`

### DIFF
--- a/ids/format.go
+++ b/ids/format.go
@@ -17,10 +17,15 @@ func (id ShortID) Format(s fmt.State, verb rune) {
 	format(s, verb, id)
 }
 
-// format implements the [fmt.Formatter] interface for [ID] and [ShortID].
-func format[T interface {
+type idForFormatting interface {
 	String() string
 	Hex() string
+}
+
+// format implements the [fmt.Formatter] interface for [ID] and [ShortID].
+func format[T interface {
+	idForFormatting
+	ID | ShortID
 }](s fmt.State, verb rune, id T) {
 	switch verb {
 	case 'x':
--- a/ids/format.go
+++ b/ids/format.go
@@ -22,9 +22,9 @@ func format[T interface {
 	switch verb {
 	case 'x':
 		if s.Flag('#') {
-			s.Write([]byte("0x"))
+			s.Write([]byte("0x")) //nolint:errcheck // [fmt.Formatter] doesn't allow for returning errors, and the implementation of [fmt.State] always returns nil on Write()
 		}
-		s.Write([]byte(id.Hex()))
+		s.Write([]byte(id.Hex())) //nolint:errcheck // See above
 
 	case 'q':
 		str := id.String()
@@ -32,9 +32,9 @@ func format[T interface {
 		buf[0] = '"'
 		buf[len(buf)-1] = '"'
 		copy(buf[1:], str)
-		s.Write(buf)
+		s.Write(buf) //nolint:errcheck // See above
 
 	default:
-		s.Write([]byte(id.String()))
+		s.Write([]byte(id.String())) //nolint:errcheck // See above
 	}
 }
--- a/ids/format.go
+++ b/ids/format.go
@@ -1,0 +1,40 @@
+package ids
+
+import "fmt"
+
+var _ = []fmt.Formatter{ID{}, ShortID{}}
+
+// Format implements the [fmt.Formatter] interface.
+func (id ID) Format(s fmt.State, verb rune) {
+	format(s, verb, id)
+}
+
+// Format implements the [fmt.Formatter] interface.
+func (id ShortID) Format(s fmt.State, verb rune) {
+	format(s, verb, id)
+}
+
+// format implements the [fmt.Formatter] interface for [ID] and [ShortID].
+func format[T interface {
+	String() string
+	Hex() string
+}](s fmt.State, verb rune, id T) {
+	switch verb {
+	case 'x':
+		if s.Flag('#') {
+			s.Write([]byte("0x"))
+		}
+		s.Write([]byte(id.Hex()))
+
+	case 'q':
+		str := id.String()
+		buf := make([]byte, len(str)+2)
+		buf[0] = '"'
+		buf[len(buf)-1] = '"'
+		copy(buf[1:], str)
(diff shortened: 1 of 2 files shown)


ids/format.go
```diff
@@ -0,0 +1,40 @@
+package ids
```
yacovm, line 1: license header missing

yacovm, line 1: why not put this in `id.go` ? it's only 40 LOC and the other formatting stuff are there.

ids/format_test.go
```diff
@@ -0,0 +1,57 @@
...
+		makeTestCase(GenerateTestID()),
+		makeTestCase(ShortID{}),
+		makeTestCase(GenerateTestShortID()),
+		makeTestCase(GenerateTestShortID()),
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.id.String(), func(t *testing.T) {
+			for format, want := range tt.want {
+				if got := fmt.Sprintf(format, tt.id); got != want {
```
yacovm, line 44: I'm curious to know the use case for this PR. We don't pass ID into Sprintf but use the logger, and there we use zap.Stringer.

ids/format.go
ARR4N: It would also need to be in `short.go` and then the common implementation would be separated from at least one use case. Here they naturally stay together and provide the reader with context.

ids/format_test.go
ARR4N: In an EVM implementation `ids.ID` is typically derived from a `common.Hash` (both `[32]byte`), which is idiomatically rendered as hex. My errors that include a block hash/ID use `%#x`.

ids/format_test.go
```diff
@@ -0,0 +1,61 @@
...
+					//nolint:forbidigo // require.Equal() is inappropriate as it unnecessarily stops future tests and `assert.Equal()` isn't allowed
+					t.Errorf("fmt.Sprintf(%q, %T) got %q; want %q", format, tt.id, got, want)
+				}
```
StephenButtolph, line 50:
```suggestion
				t.Run(format, func(t *testing.T) {
					got := fmt.Sprintf(format, tt.id)
					require.Equal(t, want, got)
				})
```

ids/format_test.go
```diff
@@ -0,0 +1,61 @@
...
+		})
+	}
+}
+
+func BenchmarkFormat(*testing.B) {
+	// %q uses a []byte so this is just to demonstrate that it's on the stack
+	// otherwise someone, not naming any names, might want to "fix" it.
+	_ = fmt.Sprintf("%q", ID{})
+	_ = fmt.Sprintf("%q", ShortID{})
+}
```
StephenButtolph, line 61: What is this doing?

ARR4N: It's purely so you can run it and see that there are no allocs. Very happy to remove it entirely.

ids/format_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"testing"
+)
+
+func TestFormat(t *testing.T) {
+	type (
+		idInterface interface {
+			String() string
+			Hex() string
+		}
```
StephenButtolph, line 16: This seems to be duplicated in the test and in the code, should we define this interface in the code and use it in both places? Perhaps `type formatted interface` or something

ids/format_test.go
ARR4N: `require` / `t.Fatal` stop all further tests, which IMO is bad practice when used as a blanket rule because it hides possibly useful information.

For example (granted, a very trivial one) did my test fail because I convert to hex incorrectly, or because I forgot to add the `0x` prefix? Seeing multiple failures gives the full context instead of forcing the engineer to play whack-a-mole.

WDYT? Is this a battle that's already been fought and won/lost in this code base?

ids/format_test.go
ARR4N: Whoops!

ids/format_test.go
StephenButtolph: > Is this a battle that's already been fought and won/lost in this code base?

Yes

ids/format_test.go
StephenButtolph:
```suggestion
func BenchmarkFormat(b *testing.B) {
	// %q uses a []byte so this is just to demonstrate that it's on the stack
	// otherwise someone, not naming any names, might want to "fix" it.
	b.Run("ID", func(b *testing.B) {
		for range b.N {
			_ = fmt.Sprintf("%q", ID{})
		}
	})
	b.Run("ShortID", func(b *testing.B) {
		for range b.N {
			_ = fmt.Sprintf("%q", ShortID{})
		}
	})
}
```

Outputs:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/ids
cpu: Apple M2 Max
BenchmarkFormat/ID-12 			 3499521	       350.7 ns/op	     272 B/op	       6 allocs/op
BenchmarkFormat/ShortID-12         	 3796848	       309.6 ns/op	     176 B/op	       6 allocs/op
PASS
coverage: 10.9% of statements in all
```

ids/format_test.go
ARR4N: image

ids/format.go
```diff
@@ -17,10 +17,15 @@ func (id ShortID) Format(s fmt.State, verb rune) {
...
-func format[T interface {
+type idForFormatting interface {
 	String() string
 	Hex() string
+}
+
+// format implements the [fmt.Formatter] interface for [ID] and [ShortID].
+func format[T interface {
+	idForFormatting
+	ID | ShortID
```
StephenButtolph, line 28: Why do we constrain the types to be either ID or ShortID?

PR #3957, opened 2025-05-12
Integrate Simplex within Avalanchego

samliok:


chains/manager.go
```diff
@@ -1412,6 +1436,213 @@ func (m *manager) createSnowmanChain(
...
+	}
+
+	chainConfig, err := m.getChainConfig(ctx.ChainID)
+	if err != nil {
+		return nil, fmt.Errorf("error while fetching chain config: %w", err)
+	}
+
+	if m.TracingEnabled {
+		vm = tracedvm.NewBlockVM(vm, primaryAlias, m.Tracer)
+	}
```
samliok, line 1500: @yacovm what is the purpose of setting the vm twice if tracing is enabled? 
```
	vm = tracedvm.NewBlockVM(vm, primaryAlias, m.Tracer)
```

PR #3959, opened 2025-05-12
ci: load test monitoring

Elvis339:
## Why this should be merged
This PR enables CI monitoring for the `load` package and specifically allows C-chain load tests to run in CI.

## How this works
- Added `solc` dependency to Nix: [[PR diff link](https://github.com/ava-labs/avalanchego/pull/3959/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0)](https://github.com/ava-labs/avalanchego/pull/3959/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0)
- Extended `load` Prometheus metric server with a utility function to dynamically generate Prometheus config files, enabling tmpnet discovery via file-based service discovery
- Enhanced tmpnet to support custom Grafana links. Previously, links were generic:
  ```
  https://grafana-poc.avax-dev.network/d/kBQpRdWnk/avalanche-main-dashboard?from=2025-05-14T16:01:56.594Z&to=2025-05-14T16:02:08.594Z&var-filter=network_uuid%7C%3D%7Ce22ec04d-11ab-47f4-9108-90aa375fb2a0&timezone=browser&orgId=1&var-datasource=P1809F7CD0C75ACF3&refresh=10s
  ```
  Now they are more specific
  ```
  https://grafana-poc.avax-dev.network/d/eabddd1d-0a06-4ba1-8e68-a44504e37535/C-Chain%20Load?from=1747242733195&to=1747242845205&var-filter=network_uuid%7C%3D%7C28a90298-93cd-486e-9a30-db820a383888
  ```

## How this was tested
- Locally and in CI

No


tests/load/c/ginkgo_test.go
```diff
@@ -98,9 +130,71 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
...
+
+// generateCollectorConfig creates the Prometheus configuration
+func generateCollectorConfig(targets []string, uuid string) ([]byte, error) {
+	return json.MarshalIndent([]map[string]any{
+		{
+			"targets": targets,
+			"labels": map[string]string{
+				"network_uuid":      uuid,
+				"network_owner":     "load-test",
+				"is_ephemeral_node": "false", // IMPORTANT: This hardcoded value prevents "No Data" in Grafana. Tmpnet generates URL with `is_ephemeral_node=false` by default (see: github.com/ava-labs/avalanchego/blob/e0304c57d/tests/fixture/tmpnet/network.go#L1055) but some metrics like load_* don't need it.
```
Elvis339, line 196: The load testing [metrics](https://github.com/ava-labs/avalanchego/blob/0f0fd0532f045586456aa873da0c68effa365adb/tests/load/tracker.go#L28) don't display in Grafana because `is_ephemeral_node` label is missing. 

When tmpnet generates links in CI, it automatically adds `is_ephemeral_node=false` as a query parameter, which adhoc applies filter in Grafana. Since load metrics don't include this label, any query with this filter returns no data for these metrics.

While solution somehwat works, it's not ideal as it requires specific domain knowledge if we are going to make a seamless performance integration. A better approach would be to extend tmpnet's API to provide a seamless way to integrate custom metrics with labels already applied to ensure compatibility. 

For example this [URL](https://grafana-poc.avax-dev.network/d/kBQpRdWnk/avalanche-main-dashboard?orgId=1&refresh=10s&var-filter=is_ephemeral_node%7C%3D%7Cfalse&var-filter=gh_repo%7C%3D%7Cava-labs%2Favalanchego&var-filter=gh_run_id%7C%3D%7C15022394281&var-filter=gh_run_attempt%7C%3D%7C1&var-filter=gh_job_id%7C%3D%7Ce2e_load&from=1747230494000&to=1747231394000&var-filter=network_owner%7C%3D%7Cavalanchego-e2e) applies: `is_ephemeral_nide=false, gh_repo=ava-labs/avalanchego, gh_run_id=15022394281, gh_run_attempt=1, gh_job_id=e2e_load, network_owner=avalanchego-e2e` it assumess all metrics have these filters thus dashboards showing "No Data".

Elvis339: I had a hacky implementation and I didn't like it so I extended tmpnet API: https://github.com/ava-labs/avalanchego/blob/da20606a53a709fc656e20bb6683abba3d9dd166/tests/fixture/tmpnet/network.go#L1068

Usage: https://github.com/ava-labs/avalanchego/blob/da20606a53a709fc656e20bb6683abba3d9dd166/tests/load/metrics_link.go

tests/load/c/ginkgo_test.go
```diff
@@ -1,17 +1,24 @@
-// Co// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: My batch replace has been rather exotic huh

tests/load/c/ginkgo_test.go
```diff
@@ -26,7 +33,10 @@ func init() {
 	flagVars = e2e.RegisterFlagsWithDefaultOwner("avalanchego-load")
 }
 
-const nodesCount = 3
+const (
+	blockchainID = "C"
```
qdm12, line 37: 👍

tests/load/c/ginkgo_test.go
```diff
@@ -53,21 +63,35 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 })
 
 var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
-	var network *tmpnet.Network
+	var (
+		network  *tmpnet.Network
+		tracker  *load.Tracker[common.Hash]
```
qdm12, line 68: Is it possible to have a tracker per subtest? Sharing the tracker between runs might not be a great idea given the orchestrator uses it to re-configure itself.
What's wrong with having a prometheus server start & stop per subtest? Especially given the last test would exit the server as a subtest would 🤔 Unless I'm missing something, what we could do is have the execute function sleep a few seconds on termination (context aware, log it out when sleeping), so it can be scrapped one last time? The sleep duration should ideally be the scrape period for the Prometheus load target

tests/load/c/ginkgo_test.go
```diff
@@ -53,21 +63,35 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
 		tc := e2e.NewTestContext()
+		r := require.New(tc)
 		env := e2e.GetEnv(tc)
+
 		network = env.GetNetwork()
 		network.Nodes = network.Nodes[:nodesCount]
 		for _, node := range network.Nodes {
 			err := node.EnsureKeys()
-			require.NoError(tc, err, "ensuring keys for node %s", node.NodeID)
+			r.NoError(err, "ensuring keys for node %s", node.NodeID)
```
qdm12, line 87: nit you (we) should prefer using the `require.Function(t` form instead of `r := require.New(t)` since this is error prone to pass the wrong `t`. I had a whole debate with Arran on that which I cannot find, but that's the bits I remember 😄  But feel free to leave unchanged, given avalanchego generally uses this form 🤷

.github/workflows/ci.yml
```diff
@@ -223,10 +223,10 @@ jobs:
       - uses: ./.github/actions/install-nix
       - name: Contract bindings are up to date
         shell: bash
-        run: ./scripts/run_task.sh check-generate-load-contract-bindings
+        run: nix develop --command ./scripts/run_task.sh check-generate-load-contract-bindings
```
qdm12, line 226: 👍 💯

tests/load/c/execute.go
```diff
@@ -53,13 +49,6 @@ func execute(ctx context.Context, preFundedKeys []*secp256k1.PrivateKey, config
 		return fmt.Errorf("ensuring minimum funds: %w", err)
 	}
 
-	registry := prometheus.NewRegistry()
```
qdm12, line 56: I don't think we really want to share the same registry for each execute call, each load test should be separated ideally 🤔

tests/load/c/ginkgo_test.go
```diff
@@ -64,30 +64,36 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	)
 
 	ginkgo.BeforeAll(func() {
-		tc := e2e.NewTestContext()
-		r := require.New(tc)
-		env := e2e.GetEnv(tc)
+		var (
+			tc  = e2e.NewTestContext()
+			env = e2e.GetEnv(tc)
+		)
```
qdm12, line 78: nit a bit unneeded diffs; fine if this is new code though. I'm a bit of a minimal-git-diffs maniac please forgive me 😄 (for somehow decent reasons though)

```suggestion
		tc := e2e.NewTestContext()
		env := e2e.GetEnv(tc)
```

tests/load/c/ginkgo_test.go
```diff
@@ -64,30 +64,36 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-		cleanup := setupMetricsServer(r, registry, network.UUID, network.Owner, logger)
+	// Setup metrics server and load tracker before each test run
+	// ensuring isolation
+	ginkgo.BeforeEach(func() {
+		promRegistry := prometheus.NewRegistry()
+		loadTracker, err := load.NewTracker[common.Hash](promRegistry)
+		require.NoError(tc, err)
+		tracker = loadTracker
+
+		cleanup := setupMetricsServer(tc, promRegistry, network.UUID, network.Owner, logger)
```
qdm12, line 96: What's preventing this from being in the execute function (to reduce coupling with ginkgo)? We could pass the network to the execute function I think?

tests/load/c/ginkgo_test.go
```diff
@@ -98,9 +119,38 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
...
+	r.NoError(err, "allocating dynamic port")
+
+	tcpAddr, ok := listener.Addr().(*net.TCPAddr)
+	r.True(ok, "allocating dynamic port")
+	port := tcpAddr.Port
+	r.NoError(listener.Close(), "closing listener on port %d", port)
+
+	metricsURI := fmt.Sprintf("127.0.0.1:%d", port)
+
+	metricsServer := load.NewPrometheusServer(metricsURI, registry, logger)
```
qdm12, line 144:
```suggestion
	metricsServer := load.NewPrometheusServer("127.0.0.1:0", registry, logger)
```

tests/load/c/ginkgo_test.go
```diff
@@ -98,9 +119,38 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
...
+	_, err = metricsServer.Start()
+	r.NoError(err, "failed to start metrics server")
+
+	monitoringConfigFilePath, err := metricsServer.GenerateMonitoringConfig(networkUUID, networkOwner)
+	r.NoError(err, "failed to generate monitoring config file")
+
+	// Return a cleanup function
+	return func() {
+		r.NoError(metricsServer.Stop(), "stopping metrics server")
```
qdm12, line 153: You can assign `metricsErrCh, err := metricsServer.Start()` and then listen to it in the cleanup

```suggestion
		select {
		case err := <-metricsErrCh:
			r.NoError(err, "metrics server error")
		default:
			r.NoError(metricsServer.Stop(), "stopping metrics server")
		}
```

Also, you might want to just inline this `setupMetricsServer` body directly in BeforeAll, since this function is getting a bit small

tests/load/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+		registry.Register(m.txsFailedCounter),
+		registry.Register(m.txLatency),
+	); err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+func (m *Metrics) IncIssuedTx() {
```
qdm12, line 55: Should we match the tracker method names? For example `Issue`, `ObserveConfirmed`, etc.? Or we should change the tracker to match this, but it sounds strange to have different naming for effectively the same event.

tests/load/prometheus.go
```diff
@@ -1,22 +1,31 @@
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
+	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
+const (
+	// relative to the user home directory
+	metricsFilePath = ".tmpnet/prometheus/file_sd_configs/load-test.json"
+)
```
qdm12, line 27: nit have this inlined within `GenerateMonitoringConfig`

tests/load/prometheus.go
```diff
@@ -79,3 +88,28 @@ func (s *MetricsServer) Stop() (err error) {
 	defer cancel()
 	return s.server.Shutdown(shutdownCtx)
 }
+
+// GenerateMonitoringConfig generates and writes the Prometheus collector configuration
+// so tmpnet can dynamically discover new scrape target via file-based service discovery
+func (s *MetricsServer) GenerateMonitoringConfig(networkUUID, networkOwner string) (string, error) {
```
qdm12, line 94:
```suggestion
// It returns the collector file path and an eventual error.
func (s *MetricsServer) GenerateMonitoringConfig(networkUUID, networkOwner string) (string, error) {
```

tests/load/metrics.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
```
qdm12, line 5: I personally like the metrics split apart from the tracker, but it is also I think an unnecessary change, as long as we share the prometheus registry right? It is clearer that metrics are shared though.

tests/load/c/ginkgo_test.go
Elvis339: Thanks, that is more readable and less code to maintain :D

https://github.com/ava-labs/avalanchego/pull/3959/commits/6dfba141e07e195d822389a6ee9e051b8b850a33

tests/load/c/ginkgo_test.go
Elvis339: Good insight, changed: https://github.com/ava-labs/avalanchego/pull/3959/commits/6dfba141e07e195d822389a6ee9e051b8b850a33

tests/load/prometheus.go
Elvis339: Change: https://github.com/ava-labs/avalanchego/pull/3959/commits/6dfba141e07e195d822389a6ee9e051b8b850a33

tests/load/metrics.go
qdm12: Indeed, as the code is right now it should be fine, but that feels like a future hard-to-debug problem if the orchestrator code changes (or other orchestrators are added) given we read the tracker state to adapt the orchestrator.

tests/load/prometheus.go
Elvis339: Is my comment verbose? I was verbose on purpose with relevant keywords such as (Prometheus, dynamic discovery, file-based service disovery) so people without domain can learn & understand what this func is doing.

qdm12: It's just we don't have a named return, so we don't really know what the string returned is. We could either use named returns or specify what we return in the comment

qdm12 (approved):

tests/load/prometheus.go
Elvis339: ahh, that makes sense updated 👍

PR #3964, opened 2025-05-14
Add Granite to the `upgrade.Config`

StephenButtolph:
## Why this should be merged

Allows us to start developing code targeting the G-upgrade.

## How this works

- Adds `Granite` to the `upgrade.Config`
- Converts the "post-fortuna" e2e run to a "post-granite" e2e run

## How this was tested

CI


api/info/service.md
```diff
@@ -695,7 +695,8 @@ curl -X POST --data '{
     "cortinaXChainStopVertexID": "11111111111111111111111111111111LpoYY",
     "durangoTime": "2020-12-05T05:00:00Z",
     "etnaTime": "2024-10-09T20:00:00Z",
-    "fortunaTime": "9999-12-01T05:00:00Z"
```
yacovm, line 698: why is Fortuna 9999?

yacovm (approved):

api/info/service.md
StephenButtolph: This is just some docs on a potential response... 9999 is the "unactivated" timestamp that we use.

yacovm: shouldn't we set it to 8th of April?

PR #3965, opened 2025-05-15
Enable Cubist Signer integration

geoff-vball:
## Why this should be merged
The enables the rpc-signer configuration.

This is based off of [this old PR](https://github.com/ava-labs/avalanchego/pull/3725).

## How this works
I moved the instantiation of the signer from the `config` package to `node.New()` where we instantiate everything else that needs filesystem/networking. This way we don't need to propagate `cleanup` all over the place.

I also opted for a `Shutdown()` method on `bls.Signer`, which is how the other resources attached to `node` are managed. Currently, resources are not shutdown if we fail in another part of `node.New()`, but this is how the rest of the resources are also managed. 

Out of scope for this PR, but we may want to consider some sort of registry of cleanup/shutdown functions that can be unwound whenever we fail. 

## How this was tested
Unit tests

Yes?


config/config.go
```diff
@@ -639,74 +639,15 @@ func getStakingTLSCert(v *viper.Viper) (tls.Certificate, error) {
...
 func getStakingConfig(v *viper.Viper, networkID uint32) (node.StakingConfig, error) {
 	config := node.StakingConfig{
 		SybilProtectionEnabled:        v.GetBool(SybilProtectionEnabledKey),
 		SybilProtectionDisabledWeight: v.GetUint64(SybilProtectionDisabledWeightKey),
 		PartialSyncPrimaryNetwork:     v.GetBool(PartialSyncPrimaryNetworkKey),
-		StakingKeyPath:                getExpandedArg(v, StakingTLSKeyPathKey),
-		StakingCertPath:               getExpandedArg(v, StakingCertPathKey),
-		StakingSignerPath:             getExpandedArg(v, StakingSignerKeyPathKey),
+		StakingTLSKeyPath:             getExpandedArg(v, StakingTLSKeyPathKey),
+		StakingTLSCertPath:            getExpandedArg(v, StakingCertPathKey),
```
geoff-vball, line 648: I renamed these to disambiguate from the bls signing items.

config/node/config.go
```diff
@@ -76,13 +75,26 @@ type StakingConfig struct {
...
+}
+
+type EphemeralSignerConfig struct{}
+
+type ContentKeyConfig struct {
+	SignerKeyRawContent string `json:"signerKeyRawContent"`
+}
+
+type SignerPathConfig struct {
+	SignerPathIsSet bool   `json:"signerPathIsSet"`
```
cam-schultz, line 92: Do we need a json tag on this field? It looks like this is set internally by `getStakingConfig`, and that the user is not expected to provide this value.

utils/crypto/bls/signer/signer.go
```diff
@@ -0,0 +1,114 @@
...
+	case node.SignerPathConfig:
+		_, err := os.Stat(cfg.SigningKeyPath)
+		if !errors.Is(err, fs.ErrNotExist) {
+			log.Info("creating signer from file")
+			return createSignerFromFile(cfg.SigningKeyPath)
+		}
+
+		if cfg.SignerPathIsSet {
+			return nil, errMissingStakingSigningKeyFile
+		}
```
cam-schultz, line 75: I don't love this pattern of embedding metadata into the config struct that's only used for internal control flow. One alternative would be to return a `signerPathIsSet` bool from `getStakingSignerConfig` and pass that around, but dedicating an entire return parameter that needs to be managed by the caller for one specific configuration type is also cumbersome. I mainly wanted to call this out as a slightly awkward implementation, that may not have a cleaner alternative.

cam-schultz (commented):
Treating the signer the same way as the other `Node` composite fields makes sense to me. I left a few questions/comments about the config parsing pattern.

utils/crypto/bls/signer/signer.go
geoff-vball: Yeah, I don't think passing a return parameter around is a good choice. Would renaming the boolean make more sense? I just kind of went off what was there, but I think "fromExistingFile" might make more sense?

iansuvak: I was thinking in `SignerPathConfig` but I now realize that this is in `node` package instead of `config` which would require a constructor or a setter.  I don't feel strongly about this either way but do prefer not serializing it (by omitting the json tag) if it's left public.

config/config.go
```diff
@@ -757,9 +699,50 @@ func getStakingConfig(v *viper.Viper, networkID uint32) (node.StakingConfig, err
 	} else {
 		config.StakingConfig = genesis.GetStakingConfig(networkID)
 	}
+
 	return config, nil
 }
 
+func getStakingSignerConfig(v *viper.Viper) (interface{}, error) {
```
joshua-kim: This PR uses this a lot, but as of go 1.18 it's idiomatic to use the `any` alias for `interface{}`

utils/crypto/bls/signer/signer.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/rpcsigner"
+	"github.com/ava-labs/avalanchego/utils/perms"
+)
+
+var errMissingStakingSigningKeyFile = errors.New("missing staking signing key file")
+
+// GetStakingSigner returns a BLS signer based on the provided configuration.
+func GetStakingSigner(
```
joshua-kim: 1. go's idiom is to name factory functions `New*`, so I think this should be `NewStakingSigner`
2. Location looks a bit awkward to me - the package is the unit of abstraction and this is the node deciding which signer to use based off of the user-provided config, so I think this should live as a non-exported function in `node` + all of the config types we created should also be non-exported.

config/config_test.go
```diff
@@ -549,6 +551,79 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
 	}
 }
 
+func TestGetStakingSigner(t *testing.T) {
+	testKey := "HLimS3vRibTMk9lZD4b+Z+GLuSBShvgbsu0WTLt2Kd4="
+	type cfg map[string]any
```
joshua-kim: nit: we don't need this type alias

config/config_test.go
```diff
@@ -549,6 +551,79 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+			v := setupViperFlags()
+
+			for key, value := range tt.config {
+				v.Set(key, value)
+			}
+
+			config, err := GetNodeConfig(v)
+
+			require.ErrorIs(err, tt.expectedErr)
+			require.IsType(tt.expectedSignerConfigType, config.StakingSignerConfig)
```
joshua-kim: Should we just do an check on the expected config instead of just a type check?

config/config_test.go
```diff
@@ -549,6 +551,79 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+
+	tests := []struct {
+		name                     string
+		viperKeys                string
+		config                   cfg
+		expectedSignerConfigType interface{}
+		expectedErr              error
+	}{
+		{
+			name:                     "default-signer",
```
joshua-kim: Can we remove the kebab-casing in these test names? Plain strings work fine and there's no reason to format it this way IMO

config/node/config.go
```diff
@@ -76,12 +75,27 @@ type StakingConfig struct {
...
-	StakingSigningKey             bls.Signer      `json:"-"`
 	SybilProtectionDisabledWeight uint64          `json:"sybilProtectionDisabledWeight"`
-	// not accessed but used for logging
-	StakingKeyPath    string `json:"stakingKeyPath"`
-	StakingCertPath   string `json:"stakingCertPath"`
-	StakingSignerPath string `json:"stakingSignerPath"`
+	StakingTLSKeyPath             string          `json:"stakingTLSKeyPath"`
+	StakingTLSCertPath            string          `json:"stakingTLSCertPath"`
+
+	// Not user supplied. This is set in order to instatiate the correct signer type.
```
joshua-kim: This comment is somewhat misleading since this configuration is defined by the user, but we determine the type at runtime

utils/crypto/bls/signer/signer.go
```diff
@@ -0,0 +1,114 @@
...
+		cancel()
+		if err != nil {
+			return nil, fmt.Errorf("couldn't create rpc signer client: %w", err)
+		}
+
+		return signer, nil
+
+	case node.SignerPathConfig:
+		_, err := os.Stat(cfg.SigningKeyPath)
+		if !errors.Is(err, fs.ErrNotExist) {
```
joshua-kim: Don't we need to handle non-nil errors that aren't of `ErrNotExist` here?

utils/crypto/bls/signer/signer.go
joshua-kim: Another idea is to just make a different type called `DefaultSignerConfig` so we could switch on that instead of having to do this hack + avoid the `ok` var passing

utils/crypto/bls/signer/signer.go
```diff
@@ -0,0 +1,114 @@
...
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/ava-labs/libevm/log"
```
joshua-kim: Should we be using the avalanchego node logger?

config/node/config.go
```diff
@@ -76,12 +75,27 @@ type StakingConfig struct {
...
+
+type EphemeralSignerConfig struct{}
+
+type ContentKeyConfig struct {
+	SignerKeyRawContent string
+}
+
+type SignerPathConfig struct {
+	SignerPathIsSet bool
+	SigningKeyPath  string
```
joshua-kim: nit: inconsistent w/ `SignerPath`, should this be `SignerKeyPath`?

config/keys.go
```diff
@@ -85,6 +85,7 @@ const (
 	StakingEphemeralSignerEnabledKey                   = "staking-ephemeral-signer-enabled"
 	StakingSignerKeyPathKey                            = "staking-signer-key-file"
 	StakingSignerKeyContentKey                         = "staking-signer-key-file-content"
+	StakingRPCSignerKey                                = "staking-rpc-signer"
```
joshua-kim: This is an endpoint - should this flag say something like that?

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create rpc signer client: %w", err)
+	}
+
 	client := pb.NewSignerClient(conn)
 
 	pubkeyResponse, err := client.PublicKey(ctx, &pb.PublicKeyRequest{})
 	if err != nil {
-		return nil, err
+		return nil, errors.Join(fmt.Errorf("failed to get pubkey response: %w", err), conn.Close())
```
joshua-kim:
```suggestion
		return nil, errors.Join(
		    fmt.Errorf("failed to get pubkey response: %w", err),
		    conn.Close(),
		)
```

joshua-kim: nit: I find this a bit easier to read

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
 	}
 
 	pkBytes := pubkeyResponse.GetPublicKey()
 	pk, err := bls.PublicKeyFromCompressedBytes(pkBytes)
 	if err != nil {
-		return nil, err
+		return nil, errors.Join(fmt.Errorf("failed to uncompress public key bytes: %w", err), conn.Close())
```
joshua-kim: Same nit

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
 	resp, err := c.client.Sign(context.TODO(), &pb.SignRequest{Message: message})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to sign message: %w", err)
 	}
-	signature := resp.GetSignature()
 
-	return bls.SignatureFromBytes(signature)
+	sigBytes := resp.GetSignature()
+	return bls.SignatureFromBytes(sigBytes)
```
joshua-kim: Since we're now wrapping the previous error I feel like it's weird for us to not wrap this error

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
 	resp, err := c.client.SignProofOfPossession(context.TODO(), &pb.SignProofOfPossessionRequest{Message: message})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to sign proof of possession: %w", err)
 	}
-	signature := resp.GetSignature()
 
-	return bls.SignatureFromBytes(signature)
+	sigBytes := resp.GetSignature()
+	return bls.SignatureFromBytes(sigBytes)
```
joshua-kim: Same nit here

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
+}
+
+func (c *Client) Shutdown() error {
+	return c.connection.Close()
```
joshua-kim: Similar nit here

utils/crypto/bls/signer/signer.go
```diff
@@ -0,0 +1,114 @@
...
+
+	default:
+		return nil, fmt.Errorf("unsupported signer type: %T", cfg)
+	}
+}
+
+func createSignerFromFile(signingKeyPath string) (bls.Signer, error) {
+	signingKeyBytes, err := os.ReadFile(signingKeyPath)
+	if err != nil {
+		return nil, err
```
joshua-kim: same wrapping nit

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
+		pk:         pk,
+		connection: conn,
 	}, nil
 }
 
 func (c *Client) PublicKey() *bls.PublicKey {
 	return c.pk
 }
 
+// Sign a message. The [Client] already handles transient connection errors.
```
joshua-kim: I would move the comment w.r.t transient conn errors to `NewClient` or `Client` since it's applicable to the type and not just `Sign`

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -16,50 +21,73 @@ import (
...
+		return nil, fmt.Errorf("failed to sign message: %w", err)
 	}
-	signature := resp.GetSignature()
 
-	return bls.SignatureFromBytes(signature)
+	sigBytes := resp.GetSignature()
+	return bls.SignatureFromBytes(sigBytes)
 }
 
+// [SignProofOfPossession] has the same behavior as [Sign] but will product a different signature.
```
joshua-kim: Square brackets are so confusing... my understanding is:
1. This is a newer convention that auto links in godoc
2. It only works if it's linked properly + exported out of the package, so stuff a package `foo` that can do `[Foo]` (if `Foo` is an exported type in `foo`) and `[bar.Bar]`, but if `Foo.Baz` is a method then you have to specify the type and can't just reference `Baz`.

So if you want to use square brackets I think you need to prefix the type (i.e `Client.SignProofOfPossession` and `Client.Sign`), or it'll break in godoc. I think you could alternatively just not use square brackets.

Regarding the actual comment, I think we can just say it produces a `ProofOfPossession` signature instead of different, since it's not clear what that difference is + we document PoP elsewhere in the repo iirc.

utils/crypto/bls/signer/signer.go
geoff-vball: So, this was the existing behaviour... What I understand to be the logic here - If we call this function twice with nothing set, we should create a signer from a new key and write that key to a file. If we call this function again, it should use the file that was written instead of creating a second new key.

If there's an error that isn't `fs.ErrNotExist`, it will error again in `createSignerFromFile` and bubble that error up anyway.

utils/crypto/bls/signer/rpcsigner/client.go
geoff-vball: Does wrapping an error in a function that only has one possible return provide any meaningful information?

config/config_test.go
geoff-vball: This was slightly annoying because of the way temp dirs work in the test suite. Let me know if you think it's better or worse.

utils/crypto/bls/signer/rpcsigner/client.go
geoff-vball: I think I removed them all. I haven't used these at all before - they're from the portion of the code Richard wrote

utils/crypto/bls/signer/rpcsigner/client.go
joshua-kim: Although there aren't any other errors in this scope, if we don't wrap this error it's not disambiguated from other grpc connection close errors. As an example, if we had another grpc conn in avalanchego and its conn close failed, we might lose visibility into which conn was the one we had to debug.

config/config.go
```diff
@@ -1450,6 +1431,7 @@ func GetNodeConfig(v *viper.Viper) (node.Config, error) {
 	nodeConfig.ProcessContextFilePath = getExpandedArg(v, ProcessContextFileKey)
 
 	nodeConfig.ProvidedFlags = providedFlags(v)
+
```
joshua-kim: This is _so_ pedantic... but I generally want to try to keep diffs clean and minimize unrelated changes. The other whitespace changes were in blocks of code that were in parts of the diff that were being touched anyways/were close to them so I'm fine letting them through, but this one is in a function that's unrelated to this PR.

node/signer.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/ava-labs/avalanchego/config/node"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/rpcsigner"
+	"github.com/ava-labs/avalanchego/utils/perms"
+)
+
+// NewStakingSigner returns a BLS signer based on the provided configuration.
+func NewStakingSigner(
```
joshua-kim: I don't think this has to be exported out of this package - it's only used by `Node` to figure out what signer to create based off of its config. If we do this I think this probably needs to be moved to `node.go`.

node/signer_test.go
```diff
@@ -0,0 +1,41 @@
...
+	t.Setenv("HOME", t.TempDir())
+
+	require := require.New(t)
+	v := setupViperFlags()
+
+	config1, err := config.GetNodeConfig(v)
+	require.NoError(err)
+	signer1, err := NewStakingSigner(config1.StakingSignerConfig)
+	require.NoError(err)
+	signer2, err := NewStakingSigner(config1.StakingSignerConfig)
```
joshua-kim: This feels like we're testing implementation details, since this code is not a part of the public node api. I think we should just create the node through `node.New`, pass in the configuration, and do a check against the `Signer` exposed on it. If we do that I think these tests belong in `node_test` instead of in their own file, since we generally shouldn't test unexported code by calling it explicitly

utils/crypto/bls/signer/rpcsigner/client.go
```diff
@@ -47,19 +75,33 @@ func (c *Client) PublicKey() *bls.PublicKey {
...
+	sigBytes := resp.GetSignature()
+	sig, err := bls.SignatureFromBytes(sigBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse signature: %w", err)
+	}
+	return sig, nil
+}
+
+func (c *Client) Shutdown() error {
+	return fmt.Errorf("failed to close connection: %w", c.connection.Close())
```
joshua-kim: Should we check if the err is nil or not here?

node/signer.go
```diff
@@ -0,0 +1,106 @@
...
+	config any,
+) (bls.Signer, error) {
+	switch cfg := config.(type) {
+	case node.EphemeralSignerConfig:
+		signer, err := localsigner.New()
+		if err != nil {
+			return nil, fmt.Errorf("couldn't generate ephemeral signer: %w", err)
```
joshua-kim: nit: in these error messages can we avoid special characters (aside from `:` which is common in go)? It makes it hard to parse logs with unix tools, since you have to escape special characters (so to be specific my feedback is to change "couldn't" to "could not").

config/config_test.go
```diff
@@ -549,6 +552,92 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
 	}
 }
 
+func TestGetStakingSigner(t *testing.T) {
+	testKey := "HLimS3vRibTMk9lZD4b+Z+GLuSBShvgbsu0WTLt2Kd4="
+	testKeyPath1 := filepath.Join(t.TempDir(), ".avalanchego/staking/signer.key")
+	testKeyPath2 := strings.Replace(testKeyPath1, "001", "002", 1) // Anticipate the new temp dir that will be created
```
joshua-kim: Is this safe to do? I think `t.TempDir` does clean up the dirs (so you can expect to start at 001), but if you hit power loss, a sig kill, or any other unexpected failure, this string replacement is brittle and can lead to flaky unit tests. It's also not documented as a guarantee on the interface for `TempDir`. Is it possible for us to get rid of this string matching?

geoff-vball: I think `t.TempDir` creates an entirely new parent directory per test run, so should be unaffected by power-loss, panic etc, but I did find a way to remove the string replacement

node/signer_test.go
geoff-vball: I've simplified this test to hopefully make things more clear. What we're testing is that when a signer is created with a new key, that that key is saved and used the next time a signer is instantiated.

config/config_test.go
joshua-kim: I think it is still affected, since if the dir fails to cleanup the hard-coded string "001" would not match since `testing` will create a new dir with a unique name ([ref](https://cs.opensource.google/go/go/+/master:src/testing/testing.go;l=1376)).

node/signer.go
joshua-kim, line 1: Could we move this into `node.go`? Typically in Go it's idiomatic to put stuff close to where they're used, so creating flat file structures where you have a dedicated file/test-file per abstraction isn't common. I think it can be okay in some circumstances, but for a non-exported struct I would lean away from it.

joshua-kim, line 1: Could we move this into `node.go`? Typically in Go it's idiomatic to put stuff close to where they're used, so creating flat file structures where you have a dedicated file/test-file per abstraction isn't common. I think it can be okay in some circumstances, but for a non-exported struct I would lean away from it.

node/signer.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not parse signing key: %w", err)
+		}
+
+		return signer, nil
+
+	case node.RPCSignerConfig:
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		signer, err := rpcsigner.NewClient(ctx, cfg.StakingSignerRPC)
+		cancel()
```
joshua-kim: nit: could we use `defer cancel()` instead of calling `cancel()` directly? it's technically the same but it's more idiomatic in go + helps us to avoid subtle bugs in the future if this code is modified.

node/signer_test.go
```diff
@@ -0,0 +1,31 @@
...
+	require := require.New(t)
+
+	config := node.DefaultSignerConfig{
+		SignerKeyPath: filepath.Join(tempDir, ".avalanchego/staking/signer.key"),
+	}
+
+	signer1, err := newStakingSigner(config)
+	require.NoError(err)
+	signer2, err := newStakingSigner(config)
+	require.NoError(err)
```
joshua-kim: Since `newStakingSigner` is a non-exported helper, ideally we wouldn’t test it directly since it's an implementation detail of `node`, not a part of its api. Would it be possible to cover this logic by going through `node.New` instead? That would make the test more realistic and aligned with how the code is actually used by the caller.

node/signer_test.go
joshua-kim, line 1: Similar comment w.r.t Go file structure, but this should probably live in `node_test` as well.

node/signer_test.go
joshua-kim: My concern isn't that the test is clear - it's that we're testing implementation details. ref https://github.com/ava-labs/avalanchego/pull/3965#discussion_r2192698611

config/config_test.go
geoff-vball: In this case I'm fairly certain that `testing` will create an entirely new parent directory for the subsequent test run

node/signer_test.go
geoff-vball: Testing the idempotence of creating a "default signer" by creating an entire `node` seems completely antithetical to to idea of unit testing. I can push down the [localsigner creation logic](https://github.com/ava-labs/avalanchego/blob/signers-config-wip/node/signer.go#L64-L68) into an exported helper in `localsigner` if that makes the separation cleaner

joshua-kim: > Testing the idempotence of creating a "default signer" by creating an entire node seems completely antithetical to to idea of unit testing.

This is not exported as part of the package, so it's an implementation detail of the package api. I can be okay with testing un-exported code when:
1. The codebase is legacy and it is not reasonable for us to refactor code to test a change
2. There is an incident/sensitive timeframe
3. The abstraction being tested is clearly encapsulated and is non-trivial logic and it is also not possible to test through a public api

In this case, although this is well-encapsulated code, it is still possible to test through the public api if we parse default viper flags. My understanding is that it is possible and simple enough to test this through `node.New` and in exchange our test is less brittle and gives us better coverage. The way this bug would occur would be if someone spun up a node twice - why not reflect that in the test?

> I can push down the [localsigner creation logic](https://github.com/ava-labs/avalanchego/blob/signers-config-wip/node/signer.go#L64-L68) into an exported helper in localsigner if that makes the separation cleaner

This is an option... but this initialization logic is only needed for the node since it needs to write/parse the key it generates on startup so it feels leaky to expose that api to all consumers of `localsigner` who only need it for unit tests.

api/server/server.go
```diff
@@ -300,12 +301,13 @@ func (s *server) AddAliasesWithReadLock(endpoint string, aliases ...string) erro
 
 func (s *server) Shutdown() error {
 	ctx, cancel := context.WithTimeout(context.Background(), s.shutdownTimeout)
-	err := s.srv.Shutdown(ctx)
+	listenerErr := s.listener.Close()
```
geoff-vball, line 293: This wasn't getting closed during `node.Shutdown()` which was causing an error starting up a new node in the same test.

node/node_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+func TestDefaultConfigInitializationUsesExistingDefaultKey(t *testing.T) {
+	require := require.New(t)
+
+	v := setupViperFlags(t)
+	conf, err := config.GetNodeConfig(v)
+	require.NoError(err)
```
StephenButtolph: I don't think we should be creating a default node in unit tests.

Specifically, there are a number of default behaviors that aren't suitable for a unit test, such as:
- Creating the `~/.avalanchego` dir
  - populating it with a database
  - staking file
  - ect
-  opening servers at `9650`, and `9651`

When I run this test locally, it fails due to `couldn't initialize API server: listen tcp 127.0.0.1:9650: bind: address already in use` because I have a node running.

node/node.go
```diff
@@ -1617,6 +1626,91 @@ func (n *Node) initDiskTargeter(
 	)
 }
 
+// newStakingSigner returns a BLS signer based on the provided configuration.
+func newStakingSigner(config any) (bls.Signer, error) {
+	switch cfg := config.(type) {
```
StephenButtolph: Why do we have multiple config types? We don't do this anywhere else afaik...

config/config_test.go
```diff
@@ -549,6 +551,95 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+					require.NoError(t, os.MkdirAll(
+						filepath.Dir(fileKeyPath),
+						os.ModePerm,
+					))
+
+					bytes, err := base64.StdEncoding.DecodeString(testKey)
+					require.NoError(t, err)
+					require.NoError(t, os.WriteFile(fileKeyPath, bytes, perms.ReadWrite))
+					return fileKeyPath
+				}(),
```
StephenButtolph:
```suggestion
				StakingSignerKeyPathKey: fileKeyPath,
```

config/config_test.go
```diff
@@ -549,6 +551,95 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
 	}
 }
 
+func TestGetStakingSigner(t *testing.T) {
+	testKey := "HLimS3vRibTMk9lZD4b+Z+GLuSBShvgbsu0WTLt2Kd4="
+	defaultSignerKeyTestDir := t.TempDir()
```
StephenButtolph: Can we call this the `dataDir`?

config/config_test.go
```diff
@@ -549,6 +551,95 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+
+	fileKeyPath := filepath.Join(t.TempDir(), "foobar/signer.key")
```
StephenButtolph:
```suggestion
	fileKeyPath := filepath.Join(t.TempDir(), "foobar", "signer.key")
```

config/config_test.go
```diff
@@ -549,6 +551,95 @@ func TestGetSubnetConfigsFromFlags(t *testing.T) {
...
+	defaultSignerKeyPath := filepath.Join(
+		defaultSignerKeyTestDir,
+		"/staking/signer.key",
+	)
```
StephenButtolph:
```suggestion
	defaultSignerKeyPath := filepath.Join(
		defaultSignerKeyTestDir,
		"staking",
		"signer.key",
	)
```

node/node.go
geoff-vball: It's a downstream consequence of moving the instantiation of the signer from `config` to `node`. 

`config` validates the config options and therefore knows what type of signer to instantiate and can pass the relevant information on. We could use `config` as just a translation layer between file and struct and move the validation logic to `node`, but that seemed like an even bigger change.

node/node_test.go
geoff-vball: I really didn't want to 😅 

Josh and I had conversations on this [here](https://github.com/ava-labs/avalanchego/pull/3965#discussion_r2180947569) and [here](https://github.com/ava-labs/avalanchego/pull/3965#discussion_r2192698611). What would you suggest doing?

StephenButtolph: Seems like the cleanest way to appease everyone is to move `newStakingSigner` into the `staking` package and export + test it correctly there.

node/node.go
StephenButtolph: My confusion is just why we use different types to indicate which signer to instantiate rather than interpreting a single config.

We can still do validation in `config`, but the way we convey the information to the `node` is pretty odd rn imo

node/node_test.go
geoff-vball: So, I think we originally started with that solution... It was a little awkward because then `staking` needed to import `config/node`. I've extracted the functionality we wanted to test into helper functions in `staking` - let me know what you think of that solution.

node/node.go
geoff-vball: I've refactor this in [this commit](https://github.com/ava-labs/avalanchego/pull/3965/commits/7a148a4d5c8bbdc0d269cfe83a663d4343de9a86)

api/server/server.go
```diff
@@ -287,12 +288,13 @@ func (s *server) AddAliasesWithReadLock(endpoint string, aliases ...string) erro
...
```
StephenButtolph, line 293: We should document that the listener is closed here in case `Serve` was never called on the server.

Tbh I feel like this is more of a patch than a fix, since there are still cases that we won't close the Listener correctly... But I suppose a patch is better then no patch 🤷

StephenButtolph (commented):
I don't think it's much easier to make this code better without re-writing the node package... so lgtm (modulo the one comment)

StephenButtolph (approved):

PR #3966, opened 2025-05-15
Remove dead code

StephenButtolph:
## Why this should be merged

I was trying to get linting to detect unused constants... I failed to do that... But I found a bunch of other dead code.

## How this works

Deletes dead code.

## How this was tested

CI

### DIFF
--- a/snow/engine/snowman/getter/getter.go
+++ b/snow/engine/snowman/getter/getter.go
@@ -42,7 +42,7 @@ func New(
 
 	var err error
 	gh.getAncestorsBlks, err = metric.NewAverager(
-		"bs_get_ancestors_blks",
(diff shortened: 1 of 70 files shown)


snow/engine/snowman/getter/getter.go
```diff
@@ -42,7 +42,7 @@ func New(
 
 	var err error
 	gh.getAncestorsBlks, err = metric.NewAverager(
-		"bs_get_ancestors_blks",
+		"get_ancestors_blks",
```
StephenButtolph, line 45: This is renamed because we remove the duplicate `get_ancestors_blks` metric below

joshua-kim (approved):

PR #3973, opened 2025-05-20
[tmpnet] Ensure GetNodeURIs returns locally-accessible URIs to ensure kube compatibility

maru-ava:
## PR Chain: tmpnet+kube

- #3854
- #3857
- #3870
- #3877
- #3867
- #3871
- #3881
- #3890
- #3884
- #3893
- #3894
- #3896
- #3897
- #3898
- #3882 
- #3945
- #3947
- **>>>>>>** #3973 **<<<<<<** 
- #3615
- #3794
## Why this should be merged

Previously it was possible to use `TestEnvironment.URIs` to access node API endpoints. If a node is running in a kube cluster, however, it won't be accessible via direct URI from outside that cluster. Adding `TestEnvironment.GetNodeURIs` and updating callers to use that ensures URIs that will work regardless of the location of nodes and test workloads.

## How this was tested

CI

N/A

### DIFF
--- a/tests/fixture/e2e/env.go
+++ b/tests/fixture/e2e/env.go
@@ -227,7 +227,7 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
 		nodeIDs := network.GetAvailableNodeIDs()
 		require.NotEmpty(nodeIDs, "network contains no nodes")
 		tc.Log().Info("network nodes are available. Not showing node URIs since kube nodes may be running remotely.",
-			zap.Any("nodeIDs", nodeIDs),
+			zap.Strings("nodeIDs", nodeIDs),
 		)
 	}
 


tests/fixture/tmpnet/utils.go
```diff
@@ -53,23 +53,55 @@ type NodeURI struct {
 	URI    string
 }
 
+// GetNodeURIs returns the URIs of the given nodes provided they are running and not ephemeral. Prefer
+// GetLocalNodeURIs if targeting nodes that could be running in a Kubernetes cluster.
```
RodrigoVillar, line 57: Do you think it'd make sense to elaborate on the function doc? From my understanding, `GetLocalNodeURIs()` returns the URIs of available nodes running locally (e.g. inside a process runtime). However, the current doc for `GetNodeURIs()` doesn't make clear (IMO) what is offers when compared to `GetLocalNodeURIs()`.

tests/fixture/tmpnet/utils.go
```diff
@@ -53,23 +53,55 @@ type NodeURI struct {
...
+		}
+		deferCleanupFunc(cancel)
+		uris = append(uris, NodeURI{
+			NodeID: node.NodeID,
+			URI:    uri,
+		})
+	}
+
+	return uris, nil
+}
```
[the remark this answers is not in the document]
Elvis339: As a tmpnet user, I'm trying to understand when this separation is needed. Having two functions increases the API surface that needs documentation and maintenance (seems negligible in this case). If we think these could be joined into one utility function, maybe it's better to do it now rather than later.

I'd prefer a single function that works regardless of environment (kube or local), but I understand that you might have context I don't have. I'm commenting only from a user perspective, so okay with your reasoning to keep both functions.

tests/fixture/tmpnet/utils.go
Elvis339: Agree with Rodrigo.

Elvis339 (approved):

tests/fixture/e2e/env.go
```diff
@@ -210,30 +207,85 @@ func NewTestEnvironment(tc tests.TestContext, flagVars *FlagVars, desiredNetwork
...
+		tc.Log().Info("network nodes are available",
+			zap.Any("uris", uris),
+		)
+	} else {
+		// Only display node IDs for kube-based networks since the nodes may not be network accessible and
+		// port-forwarded URIs are ephemeral
+		nodeIDs := network.GetAvailableNodeIDs()
+		require.NotEmpty(nodeIDs, "network contains no nodes")
+		tc.Log().Info("network nodes are available. Not showing node URIs since kube nodes may be running remotely.",
+			zap.Any("nodeIDs", nodeIDs),
```
RodrigoVillar, line 230: (not necessary): `zap.Any` => `zap.Strings`

RodrigoVillar (approved):

RodrigoVillar (approved):

PR #3976, opened 2025-05-22
Add Simplex To P2P Proto

samliok:
## Why this should be merged

This PR is the first of many to try and integrate [Simplex](https://github.com/ava-labs/simplex) in AvalancheGo. This first PR defines Simplex messages in `p2p.proto`. It also routes inbound Simplex messages, to a simplex engine that will be completed in a future PR. 

## How this was tested

Added a tests in `messages_test.go` and `engine_test.go` to check the new opcode & engines are functioning. Feels like a bit more testing is needed, maybe a test in `handler_test.go` should be added?


network/peer/peer.go
```diff
@@ -212,6 +213,7 @@ func Start(
 	messageQueue MessageQueue,
 	isIngress bool,
 ) Peer {
+	fmt.Println("Starting peer, id:", id, "engine:", config.VersionCompatibility.Version().String())
```
yacovm: can we get rid of the `fmt.Println`s everywhere?

network/peer/peer.go
```diff
@@ -385,7 +388,7 @@ func (p *peer) readMessages() {
 	for {
 		// Time out and close connection if we can't read the message length
 		if err := p.conn.SetReadDeadline(p.nextTimeout()); err != nil {
-			p.Log.Verbo(failedToSetDeadlineLog,
+			p.Log.Warn(failedToSetDeadlineLog,
```
yacovm: can we please revert back the log levels? I used them for debugging

proto/p2p/p2p.proto
```diff
@@ -419,3 +422,97 @@ message AppGossip {
   // Message body
   bytes app_bytes = 2;
 }
+
+message Simplex {
+  // Chain being responded from
```
yacovm, line 429: why responded from? what does that mean?

proto/p2p/p2p.proto
```diff
@@ -419,3 +422,97 @@ message AppGossip {
...
+  // Seq is the order of the block among all blocks in the blockchain.
+  // Cannot correspond to an empty block.
+  uint64 seq = 4;
+  // Prev returns the digest of the previous data block
+  bytes prev = 5;
+}
+
+message BlockHeader {
+  ProtocolMetadata metadata = 1;
+  bytes digest = 2;
```
yacovm, line 467: // digest is the digest of the inner block

snow/networking/router/chain_router.go
```diff
@@ -706,3 +714,92 @@ func (cr *ChainRouter) clearRequest(
...
 	return uniqueRequestID, &request
 }
+
+func (cr *ChainRouter) handleSimplexMessage(ctx context.Context, msg message.InboundMessage) {
+	m := msg.Message()
+	nodeID := msg.NodeID()
+	op := msg.Op()
+
+	if msg.Op() != message.SimplexOp {
+		cr.log.Debug("dropping message with invalid field",
```
yacovm: why are we dropping because it has an invalid field? aren't we dropping it because the op isn't compatible?

snow/networking/router/chain_router.go
```diff
@@ -706,3 +714,92 @@ func (cr *ChainRouter) clearRequest(
...
+		)
+		cr.metrics.droppedRequests.Inc()
+		msg.OnFinishedHandling()
+		return
+	}
+
+	engineType := p2p.EngineType_ENGINE_TYPE_UNSPECIFIED
+
+	// Pass the response to the chain
+	chain.Push(
```
yacovm, line 786: can we add a test? `Push` is an interface method, we can mock the handler and intercept it in the test.

yacovm:
Made a quick pass, will review again next week.

simplex/engine.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package simplex
```
yacovm, line 4: is it possible to strip out the engine code addition from this PR? I think the chain router and protobuf is enough to capture some functionality, while I'm not sure the engine changes provide significant functionality to justify being part of this PR.

snow/networking/handler/handler.go
```diff
@@ -709,7 +709,20 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
...
+		if engineType != p2ppb.EngineType_ENGINE_TYPE_SIMPLEX {
+			h.ctx.Log.Debug("dropping simplex message",
+				zap.String("reason", "engine type mismatch"),
+				zap.String("messageOp", op),
+				zap.Stringer("currentEngineType", currentState.Type),
+			)
+			return nil
+		}
 
+		return fmt.Errorf(
```
yacovm, line 722: I am not sure we want to return an error here, this will stop the chain, won't it?

yacovm (approved):

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+message ReplicationResponse {
+  repeated QuorumRound data = 1;
+  QuorumRound latest_round = 2;
+}
+
+message QuorumRound {
+  SimplexBlock block = 1;
+  Notarization notarization = 2;
+  EmptyNotarization empty_notarization = 3;
+  FinalizationCertificate finalization = 4;
```
StephenButtolph: Not an actionable comment, but when I had first looked at this I felt like putting the `empty_notarization` and `finalization` into a `OneOf` might be a nice idea (because in simplex a `FinalizationCertificate` implies that an `EmptyNotarization` can never exist.

But in our protocol they _can_ exist if a block was transitively finalized. So not relevant, but felt worthwhile to write down the thoughts.

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+}
+
+message ReplicationRequest {
+  repeated uint64 seqs = 1; // array of seqs being requested
+  uint64 latest_round = 2; // latest round, our node has
+}
+
+message ReplicationResponse {
+  repeated QuorumRound data = 1;
+  QuorumRound latest_round = 2;
```
StephenButtolph: This has the same meaning as in the request right? This is the latest round that the server has to be able to fetch?

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+  bytes quorum_certificate = 2;
+}
+
+message Notarization {
+  BlockHeader vote = 1;
+  bytes quorum_certificate = 2;
+}
+
+message EmptyNotarization {
+  BlockHeader empty_vote = 1;
```
StephenButtolph: Should this be:
```suggestion
  ProtocolMetadata empty_vote = 1;
```
?

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+
+message BlockHeader {
+  ProtocolMetadata metadata = 1;
+  // digest is the digest of the inner block data.
+  bytes digest = 2;
+}
+
+message Signature {
+  bytes signer = 1;
+  bytes value = 2;
```
StephenButtolph: Can we add some comments:
```suggestion
  // Signer identifies who the signature came from.
  bytes signer = 1;
  // Value is the actual cryptographic signature.
  bytes value = 2;
```

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+    Finalization finalization = 5;
+    Notarization notarization = 6;
+    EmptyNotarization empty_notarization = 7;
+    FinalizationCertificate finalization_certificate = 8;
+    ReplicationResponse replication_response = 9;
+    ReplicationRequest replication_request = 10;
+  }
+}
+
+message SimplexBlock {
```
StephenButtolph: I feel like `SimplexBlock` isn't a great name for this... Could we name it a `Proposal` rather than a `SimplexBlock`? In reality, it is the block + the vote for the block from the block proposer.

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+message Simplex {
+  bytes chain_id = 1;
+
+  oneof message {
+    SimplexBlock block = 2;
+    Vote vote = 3;
+    EmptyVote empty_vote = 4;
+    Finalization finalization = 5;
+    Notarization notarization = 6;
+    EmptyNotarization empty_notarization = 7;
```
StephenButtolph: I think I had commented on this in the past (and if we have decided against it that is fine), but I just feel like `Finalization` and `Notarization` are so similar that I always think that the `Finalization` is the aggregation of the `Finalize` messages. But instead we have a:

- `Vote` <-> `EmptyVote` <-> `Finalization`
- `Notarization` <-> `EmptyNotarization` <-> `FinalizationCertificate`

equivalency.

I'd prefer:
- `Vote` <-> `EmptyVote` <-> `FinalizeVote`
- `Notarization` <-> `EmptyNotarization` <-> `Finalization`

or:
- `Vote` <-> `EmptyVote` <-> `Finalize`
- `Notarization` <-> `EmptyNotarization` <-> `Finalization`

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+  Signature signature = 2;
+}
+
+message Finalization {
+  BlockHeader vote = 1;
+  Signature signature = 2;
+}
+
+message FinalizationCertificate {
+  Finalization finalization = 1;
```
StephenButtolph: Why does this include a `Finalization`? Shouldn't this be a `BlockHeader`?

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+
+message EmptyVote {
+  ProtocolMetadata vote = 1;
+  Signature signature = 2;
+}
+
+message Finalization {
+  BlockHeader vote = 1;
+  Signature signature = 2;
+}
```
StephenButtolph: imo `Vote` and `Finalization` should be merged into the same type along with `FinalizationCertificate` and `Notarization`.

They have different meanings, but it seems to me like those different meanings are conveyed in the `OneOf` in the `Simplex` message definition.

proto/p2p/p2p.proto
```diff
@@ -419,3 +424,97 @@ message AppGossip {
...
+  bytes quorum_certificate = 2;
+}
+
+message ReplicationRequest {
+  repeated uint64 seqs = 1; // array of seqs being requested
+  uint64 latest_round = 2; // latest round, our node has
+}
+
+message ReplicationResponse {
+  repeated QuorumRound data = 1;
```
StephenButtolph: Can we document if we expect the data to match the same ordering of the requested `seqs`?

snow/networking/handler/handler.go
```diff
@@ -485,7 +485,7 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
 
 	var engineType p2ppb.EngineType
 	switch msg.EngineType {
-	case p2ppb.EngineType_ENGINE_TYPE_AVALANCHE, p2ppb.EngineType_ENGINE_TYPE_SNOWMAN:
+	case p2ppb.EngineType_ENGINE_TYPE_AVALANCHE, p2ppb.EngineType_ENGINE_TYPE_SNOWMAN, p2ppb.EngineType_ENGINE_TYPE_SIMPLEX:
```
StephenButtolph: Note for me to come back to this. I am not immediately confident in the correctness of this change.

snow/networking/handler/handler.go
```diff
@@ -709,7 +709,22 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
 		}
 
 		return engine.Chits(ctx, nodeID, msg.RequestId, preferredID, preferredIDAtHeight, acceptedID, msg.AcceptedHeight)
+	case *p2ppb.Simplex:
```
StephenButtolph: Can we move this down to be after `QueryFailed`? This switch case is currently in the order of:
- Request
- Response
- RequestFailed

And currently the `Simplex` message is in-between the `Response` and `RequestFailed` for snow consensus messages.

message/inbound_msg_builder.go
```diff
@@ -314,6 +314,21 @@ func InboundAppResponse(
 	}
 }
 
+// NewInboundSimplexMessage creates a new InboundMessage for simplex messages.
+func InboundSimplexMessage(
+	chainID ids.ID,
+	nodeID ids.NodeID,
+	msg p2p.Message_Simplex,
```
StephenButtolph: Since we aren't able to reference the `OneOf` interface, I think the cleanest method signature here would be:
```suggestion
	nodeID ids.NodeID,
	msg *p2p.Simplex,
```
It feels weird to me to pass in `p2p.Message_Simplex` (because that is just the OneOf implementation for the `message` `OneOf`. Also I don't love passing in "duplicate" arguments, which is the `chainID` and the `msg` (which includes the `chainId`)

message/ops.go
```diff
@@ -59,6 +59,8 @@ const (
 	DisconnectedOp
 	NotifyOp
 	GossipRequestOp
+	// Simplex
+	SimplexOp
```
StephenButtolph, line 63: Should this be added to the `UnrequestedOps` set?

snow/networking/router/chain_router.go
```diff
@@ -208,6 +208,12 @@ func (cr *ChainRouter) HandleInbound(ctx context.Context, msg message.InboundMes
 	op := msg.Op()
 
 	m := msg.Message()
+
+	if msg.Op() == message.SimplexOp {
+		cr.handleSimplexMessage(ctx, msg)
+		return
+	}
```
StephenButtolph: I don't think we should introduce a new flow for this message type, I think we should just treat this message like we treat other `UnrequestedOps` messages.

proto/p2p/p2p.proto
```diff
@@ -214,6 +217,8 @@ enum EngineType {
   // Only the X-Chain uses avalanche consensus
   ENGINE_TYPE_AVALANCHE = 1;
   ENGINE_TYPE_SNOWMAN = 2;
+  // For simplex consensus
+  ENGINE_TYPE_SIMPLEX = 3;
```
StephenButtolph: The `EngineType` was only introduced to handle the `Avalanche` -> `Snowman` transition for the X-chain. It has now been removed from all messages except for the `GetAncestors` messages to support bootstrapping the DAG on the X-chain.

While we _are_ introducing a new engine type, I don't think we should modify this ENUM unless we are trying to support migrating from Snowman to Simplex (which while we may try to do that at some point, it seems like we should support creating Simplex standalone first)

snow/networking/handler/handler.go
StephenButtolph: See: https://github.com/ava-labs/avalanchego/pull/3976#discussion_r2122152274

proto/p2p/p2p.proto
yacovm: I think it makes sense, and I see how we can rename the code easily. 
Renaming the terms in our heads will be harder though (at least speaking for myself).

yacovm: (I'm in favor of this, even though it will take time/practice to get used to the new terminology)

proto/p2p/p2p.proto
yacovm: Correct... oops.

proto/p2p/p2p.proto
yacovm: yes

proto/p2p/p2p.proto
yacovm: We shouldn't make any assumptions on what we're returned from a foreign node. 

In the code, when we handle the response, we [sort](https://github.com/ava-labs/Simplex/blob/e9927d0012259c5d4999674b44fd3ac024a4eb6f/replication.go#L184) it before further processing.

proto/p2p/p2p.proto
yacovm: It's just a way of the remote node to tell you "hey I know you requested such and such blocks, but just so you know, I have advanced since you asked me, and this is the proof for me advancing"

proto/p2p/p2p.proto
yacovm: In the code, you are expected to only populate one of notarization, empty notarization, and finalizationCertificate. 

We actually enforce that it's a one-of and reject the message if it's not.

However I don't like the way the golang protobuf compiler implements the one-of, (because of the unexported intermediate interfaces) so I suggest we keep it this way

proto/p2p/p2p.proto
yacovm: I think it makes sense, but only for the protobuf messages. In the actual Simplex application we have to have these types separate, as we distinguish which context to sign with by [the type](https://github.com/ava-labs/Simplex/blob/e9927d0012259c5d4999674b44fd3ac024a4eb6f/msg.go#L113).

snow/networking/router/chain_router.go
samliok: makes sense, i did this because a couple lines down we try and grab call `GetRequestID(m)` and inside this function we have 
```
func GetRequestID(m any) (uint32, bool) {
...
// AppGossip is the only inbound message not containing a requestID. For
// ease of handling, imagine that it does have a requestID.
_, ok := m.(*p2p.AppGossip)
return 0, ok
```

I think i'd need to update this function in order for us not to error.

message/ops.go
samliok: added

proto/p2p/p2p.proto
samliok: since I've gotten used to the other verbage, i may be a bit confused by your equivalency bulletpoints. Currently a `FinalizationCertificate` is an aggregation of `Finalizations`, and `Finalization` is an aggregation of `Notarizations` and a `Notarization` is an aggregation of `Votes`.  

However, I'm confused about your first two bullets. `Finalization` seems out of place here, since its a collection of `Notarizations` not votes.

> Vote <-> EmptyVote <-> Finalization

I think it would make sense for `Finalization` to be the new name for `FinalizationCertificate` and `FinalizeVote` would be the new name for a `Notarization` but im not sure this is what you're suggesting

proto/p2p/p2p.proto
samliok: ah ok, makes sense. Thanks for the context 🫡

proto/p2p/p2p.proto
samliok: agree, I'll rename it to `BlockProposer` for slightly more clarity.

proto/p2p/p2p.proto
samliok: this makes sense. i feel like I currently use the term `Finalization` and `FinalizationCertificate` interchangeably even though they are quite different

message/inbound_msg_builder.go
```diff
@@ -314,6 +314,19 @@ func InboundAppResponse(
 	}
 }
 
+// NewInboundSimplexMessage creates a new InboundMessage for simplex messages.
+func InboundSimplexMessage(
+	nodeID ids.NodeID,
+	msg p2p.Message_Simplex,
```
StephenButtolph:
```suggestion
	msg *p2p.Simplex,
```

proto/p2p/p2p.proto
```diff
@@ -419,3 +422,91 @@ message AppGossip {
...
+
+message ReplicationResponse {
+  repeated QuorumRound data = 1; // requested seqs. not required to be in particular order
+  QuorumRound latest_round = 2; // latest round the responding node is aware of
+}
+
+// QuorumRound represents a round that has acheived quorum on either
+// (empty notarization), (block & notarization), or (block, finalization certificate)
+message QuorumRound {
+  BlockProposal block = 1;
```
StephenButtolph: Do we need the `Vote` inside the `BlockProposal` here? Or should this just be `bytes`?

proto/p2p/p2p.proto
```diff
@@ -419,3 +422,91 @@ message AppGossip {
...
+  oneof message {
+    BlockProposal block_proposal = 2;
+    Vote vote = 3;
+    EmptyVote empty_vote = 4;
+    Vote finalize_vote = 5;
+    QuorumCertificate notarization = 6;
+    EmptyNotarization empty_notarization = 7;
+    QuorumCertificate finalization = 8;
+    ReplicationResponse replication_response = 9;
+    ReplicationRequest replication_request = 10;
```
StephenButtolph: This is going to make my OCD really obvious, but could we humor me and list the request before the response?
```suggestion
    ReplicationRequest replication_request = 9;
    ReplicationResponse replication_response = 10;
```
It's super trivial but I just feel like it makes the code read very slightly more naturally.

snow/networking/router/chain_router.go
StephenButtolph: Hmmm good point, I'll make a separate PR to clean this up.

StephenButtolph (approved):

PR #3977, opened 2025-05-23
chore(deps): use coreth with geth-aligned ethclient package

qdm12:
## Why this should be merged

So we can have a textually aligned ethclient package with geth in coreth. This uses coreth from this master branch commit: https://github.com/ava-labs/coreth/commit/aac22d91be3fceeca235ce3e4f49415fd47887ff

~ℹ️  this is based on top of v1.13.1-rc.2 because that's what coreth is currently using (2025-05-23). I did try to base it on avalanchego master branch but that caused some tests to fail for unrelated reasons, and didn't have time to debug all this today.~

## How this works

Change using the `ethclient.Client` interface to the pointer `*ethclient.Client`. There was no point using an interface anyway.

## How this was tested

CI passing on both the coreth PR and this PR

No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/StephenButtolph/canoto v0.15.0
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/coreth v0.15.1-rc.0.0.20250523100307-aa2d3705c439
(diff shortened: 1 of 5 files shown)


qdm12:
Sorry I won't be around to handle PR review feedback, but it's rather straight forward to review I think.
Please feel free to push commits to that branch if you want.

StephenButtolph (approved):

PR #3982, opened 2025-05-29
feat: canoto tags for `gas.State`

ARR4N:
## Why this should be merged

Required for SAE shutdown recovery.

## How this works

Adds `canoto` field tags to `gas.State`.

## How this was tested

No explicit test.

No


vms/components/gas/state.go
```diff
@@ -13,9 +13,11 @@ import (
 
 var ErrInsufficientCapacity = errors.New("insufficient capacity")
 
+//go:generate go run github.com/StephenButtolph/canoto/canoto $GOFILE
```
StephenButtolph, line 16: I think we typically put the `//go:generate` command at the top of the file

PR #3986, opened 2025-06-02
[testing] Add local kube support for load tests

RodrigoVillar:
## Why this should be merged

This PR extends the functionality of the load tests by allowing it to run against a deployed Kube network.

## How this works

```bash
./scripts/run_task.sh test-load-kube
```

N/A

### DIFF
--- a/scripts/tests.load.kube.sh
+++ b/scripts/tests.load.kube.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Run load test against nodes deployed to a kind cluster
+
+if ! [[ "$0" =~ scripts/tests.load.kube.sh ]]; then
+    echo "must be run from repository root"
+    exit 255
+fi
+
+# This script will use kubeconfig arguments if supplied
+./scripts/start_kind_cluster.sh "$@"
+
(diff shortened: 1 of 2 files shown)


scripts/tests.load.kube.sh
```diff
@@ -0,0 +1,18 @@
...
+    exit 255
+fi
+
+# This script will use kubeconfig arguments if supplied
+./scripts/start_kind_cluster.sh "$@"
+
+# Build AvalancheGo image
+DOCKER_IMAGE=localhost:5001/avalanchego FORCE_TAG_LATEST=1 ./scripts/build_image.sh 
+
+go run ./tests/load/c/main --runtime=kube "$@"
```
maru-ava: You'll want to set `--kube-image` to the same value as `DOCKER_IMAGE` to ensure the test is using the image built from the current state of the tree. The test is still passing because the default image (`avaplatform/avalanchego:latest`) is currently compatible with the load test, but that may not always be the case.

RodrigoVillar: Thanks for pointing this out - I've updated the load test to set `kube-image`: d6f2166

PR #3988, opened 2025-06-03
[tmpnet] Enable exclusive scheduling

maru-ava:
## Why this should be merged

tmpnet previously supported deploying avalanchego nodes as kubernetes pods, but didn't provide a way to ensure that nodes had guaranteed reservations of compute and memory. This PR introduces a simple mechanism enabling each node in a network deployed with tmpnet to a kube cluster to be scheduled to its own dedicated and appropriately-sized kube node.  

## How this works

Setting `--kube-use-exclusive-scheduling` when using a network with tmpnet with `--runtime=kube` will ensure that each avalanchego node in the network will be scheduled to a beefy dedicated Kubernetes node (assuming the target cluster is properly configured). This avoids contention for resources between nodes and between nodes and other applications. Node selection is configured by `--kube-scheduling-label-key` and `--kube-scheduling-label-value`. 

The underlying scheduling mechanisms include:
 - [Node selection](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector)
    - Enables scheduling only to a specific node/node group
 - [Taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
    - A node/node group configured with a taint can only be scheduled to if a pod has a matching toleration. e.g.
       - a node/node group with _taint_ X will only be scheduled to by avalanchego pods with a _toleration_ for X
       - other pods that do not tolerate a node/node group's taint will never be scheduled to it
 - [Affinity/anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity)
    - This enables an avalanchego node to indicate that it should never be scheduled to the same kubernetes node as another avalanchego node.

## How this was tested

CI against regression

New behavior validated against internal test cluster:
 - [x] pods are scheduled to dedicated nodes
 - [x] metrics are collected 
 - [x] logs are collected

N/A

### DIFF
--- a/tests/fixture/tmpnet/node.go
+++ b/tests/fixture/tmpnet/node.go
@@ -57,11 +57,11 @@ type NodeRuntimeConfig struct {
 }
 
 // GetNetworkStartTimeout returns the timeout to use when starting a network.
-func (c *NodeRuntimeConfig) GetNetworkStartTimeout(nodeCount int) time.Duration {
+func (c *NodeRuntimeConfig) GetNetworkStartTimeout(nodeCount int) (time.Duration, error) {
 	switch {
 	case c.Process != nil:
 		// Processes are expected to start quickly, nodeCount is ignored
-		return DefaultNetworkTimeout
+		return DefaultNetworkTimeout, nil
 	case c.Kube != nil:
 		// Ensure sufficient time for scheduling and image pull
 		timeout := time.Duration(nodeCount) * time.Minute
@@ -71,9 +71,9 @@ func (c *NodeRuntimeConfig) GetNetworkStartTimeout(nodeCount int) time.Duration 
 			timeout *= 2
 		}
 
-		return timeout
+		return timeout, nil
 	default:
-		panic("no runtime configuration set")
+		return 0, errors.New("no runtime configuration set")
 	}
 }
 
--- a/tests/fixture/tmpnet/tmpnetctl/main.go
+++ b/tests/fixture/tmpnet/tmpnetctl/main.go
@@ -80,7 +80,10 @@ func main() {
 				DefaultRuntimeConfig: *nodeRuntimeConfig,
 			}
 
-			timeout := nodeRuntimeConfig.GetNetworkStartTimeout(nodeCount)
+			timeout, err := nodeRuntimeConfig.GetNetworkStartTimeout(nodeCount)
+			if err != nil {
+				return err
+			}
 			log.Info("waiting for network to start",
 				zap.Float64("timeoutSeconds", timeout.Seconds()),
 			)
--- a/tests/fixture/e2e/helpers.go
+++ b/tests/fixture/e2e/helpers.go
@@ -282,13 +282,14 @@ func StartNetwork(
 	require := require.New(tc)
 
 	nodeCount := len(network.Nodes)
-	timeout := network.DefaultRuntimeConfig.GetNetworkStartTimeout(nodeCount)
+	timeout, err := network.DefaultRuntimeConfig.GetNetworkStartTimeout(nodeCount)
+	require.NoError(err)
 	tc.Log().Info("waiting for network to start",
 		zap.Float64("timeoutSeconds", timeout.Seconds()),
 	)
 	ctx := tc.ContextWithTimeout(timeout)
 
-	err := tmpnet.BootstrapNewNetwork(
+	err = tmpnet.BootstrapNewNetwork(
 		ctx,
 		tc.Log(),
 		network,


tests/fixture/tmpnet/tmpnetctl/main.go
```diff
@@ -80,7 +81,24 @@ func main() {
...
+
+			// Kube-based nodes are expected to take longer
+			if nodeRuntimeConfig.Kube != nil {
+				// Ensure sufficient time for scheduling and image pull
+				timeout = time.Duration(nodeCount) * time.Minute
+
+				if nodeRuntimeConfig.Kube.UseExclusiveScheduling {
+					// Ensure sufficient time for the creation of autoscaled nodes
+					timeout *= 2
+				}
```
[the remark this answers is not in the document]
maru-ava: The duration scales with the node count, putting a maximum could preclude waiting long enough.

tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -764,3 +786,50 @@ func (p *KubeRuntime) getFlags() (FlagsMap, error) {
...
+			Key:      labelKey,
+			Operator: corev1.TolerationOpEqual,
+			Value:    labelValue,
+			Effect:   corev1.TaintEffectNoExecute,
+		},
+	}
+
+	// Configure anti-affinity to ensure only one pod per node
+	antiAffinityLabelKey := "tmpnet-scheduling"
+	antiAffinityLabelValue := "exclusive"
```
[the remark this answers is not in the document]
maru-ava: Done

tests/fixture/tmpnet/flags/kube_runtime.go
```diff
@@ -64,6 +68,24 @@ func (v *kubeRuntimeVars) register(stringVar varFunc[string], uintVar varFunc[ui
...
+	boolVar(
+		&v.useExclusiveScheduling,
+		"kube-use-exclusive-scheduling",
+		false,
+		kubeDocPrefix+"Whether to schedule each AvalancheGo node to a dedicated Kubernetes node",
+	)
+	stringVar(
+		&v.schedulingLabelKey,
+		"kube-scheduling-label-key",
+		"purpose",
```
RodrigoVillar, line 80: Shouldn't this default value be `""` to enforce that the caller provides a scheduling kv-pair if they enable exclusive scheduling?

maru-ava: This is the default on our test cluster. My thinking is that the `--kube-use-exclusive-scheduling` flag is sufficient to indicate intent and that having to additionally provide the label key and value doesn't add value. I chose to default it in the flags (rather that e.g. in the task invocation) so that it works for tmpnetctl too.

maru-ava: fwiw this is a middle ground - I considered just making the key and value constants. But figured I'd leave the door open to usage on clusters with different configuration. I could be convinced YAGNI though and just hard-code it.

tests/fixture/tmpnet/node.go
```diff
@@ -56,6 +56,27 @@ type NodeRuntimeConfig struct {
...
+		timeout := time.Duration(nodeCount) * time.Minute
+
+		if c.Kube.UseExclusiveScheduling {
+			// Ensure sufficient time for the creation of autoscaled nodes
+			timeout *= 2
+		}
+
+		return timeout
+	default:
+		panic("no runtime configuration set")
```
Elvis339: Consumers of this method handle errors properly
- https://github.com/ava-labs/avalanchego/blob/c3ff3cb0ff9d0dc5e2761160f19e7e3a81f638b5/tests/fixture/tmpnet/tmpnetctl/main.go#L83
- https://github.com/ava-labs/avalanchego/blob/c3ff3cb0ff9d0dc5e2761160f19e7e3a81f638b5/tests/fixture/e2e/helpers.go#L285

What's the rationale for panicking instead of returning an error? If panic is required, the method name should clearly indicate it can panic (e.g., `MustDoSomething`).

tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -764,3 +792,48 @@ func (p *KubeRuntime) getFlags() (FlagsMap, error) {
...
+	podSpec.NodeSelector[labelKey] = labelValue
+
+	// Configure toleration. Nodes are assumed to have a taint with the same
+	// key+value as the label used to select it.
+	podSpec.Tolerations = []corev1.Toleration{
+		{
+			Key:      labelKey,
+			Operator: corev1.TolerationOpEqual,
+			Value:    labelValue,
+			Effect:   corev1.TaintEffectNoExecute,
```
Elvis339, line 815: NoExecute will immediately evict existing pods. Is this the intended behavior, or would NoSchedule be safer for initial rollouts?

Elvis339 (commented):
Few questions:
- What's the failure mode when exclusive scheduling can't be satisfied?
- When tearing down the network, do we need special handling for the dedicated nodes?

tests/fixture/tmpnet/kube_runtime.go
maru-ava: There is no choice to be made here - we're _tolerating_ a taint that is configured `NoExecute`.

maru-ava:
> Few questions:
> 
> * What's the failure mode when exclusive scheduling can't be satisfied?

Same as before, failure to start the network will result in [the network being stopped](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/e2e/helpers.go#L290): 

> * When tearing down the network, do we need special handling for the dedicated nodes?

The autoscaling node group will automatically scale down members once they are idle for a specific duration (currently 30m).

tests/fixture/tmpnet/node.go
maru-ava: Done.

tests/fixture/tmpnet/kube_runtime.go
maru-ava: To be fair, we could ask infra to change this. I'm just targeting the existing configuration. But for test workloads I'm not sure it matters. If they have a reason to drain the node pool we only lose a test result that we could re-execute (vs a production workload where downtime might have a customer impact).

Elvis339:
> > Few questions:
> > 
> > * What's the failure mode when exclusive scheduling can't be satisfied?
> 
> Same as before, failure to start the network will result in [the network being stopped](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/e2e/helpers.go#L290):
> 
> > * When tearing down the network, do we need special handling for the dedicated nodes?
> 
> The autoscaling node group will automatically scale down members once they are idle for a specific duration (currently 30m).

---

> Same as before, failure to start the network will result in [the network being stopped](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/e2e/helpers.go#L290):

When exclusive scheduling fails, the network cleanup happens but the root cause isn't clear.
I could easily see situations where instead of focusing on performance improvements and gaining signal, I'm stuck debugging infrastructure issues. Would it be possible to address this in future iterations by surfacing Kubernetes scheduler events? The current implementation is fine, but better error reporting would significantly improve the debugging experience.

maru-ava:
> When exclusive scheduling fails, the network cleanup happens but the root cause isn't clear. I could easily see situations where instead of focusing on performance improvements and gaining signal, I'm stuck debugging infrastructure issues. Would it be possible to address this in future iterations by surfacing Kubernetes scheduler events? The current implementation is fine, but better error reporting would significantly improve the debugging experience.

Fair points all, and this is very much a work-in-progress. I'm inclined to fix problems as they come rather than try to anticipate them for a couple of reasons:

 - it is difficult if not impossible to enumerate all the potential failure modes that we might encounter
 - inducing potential failure modes is complicated, and error handling requires being able to reproduce those failure modes

Note that the reason we have CI jobs deploying to kube is to ensure that the majority of the functionality is validated repeatedly on every PR. This narrows the potential failure modes to e.g. scheduling to these dedicated node groups. For now, if/when scheduling issues are encountered, it should be possible to trivially reproduce them by deploying via tmpnetctl which does not tear down on failure so the network state on the cluster is inspectable. I'd much rather be able to investigate all the running elements of a network in the event of failure vs having a static snapshot that may not capture the full extent of a problem.

Elvis339:
To very large extent I agree that we should fix problems as they come rather than anticipating but that is not to say that we should discard everything we anticipate - I'd take a balanced approach.

> it should be possible to trivially reproduce them by deploying via tmpnetctl which does not tear down on failure so the network state on the cluster is inspectable

This is very much true. Thinking as the user I'd rather not spend time on inspection - sometimes I will have to 

> > When exclusive scheduling fails, the network cleanup happens but the root cause isn't clear. I could easily see situations where instead of focusing on performance improvements and gaining signal, I'm stuck debugging infrastructure issues. Would it be possible to address this in future iterations by surfacing Kubernetes scheduler events? The current implementation is fine, but better error reporting would significantly improve the debugging experience.
> 
> Fair points all, and this is very much a work-in-progress. I'm inclined to fix problems as they come rather than try to anticipate them for a couple of reasons:
> 
> * it is difficult if not impossible to enumerate all the potential failure modes that we might encounter
> * inducing potential failure modes is complicated, and error handling requires being able to reproduce those failure modes
> 
> Note that the reason we have CI jobs deploying to kube is to ensure that the majority of the functionality is validated repeatedly on every PR. This narrows the potential failure modes to e.g. scheduling to these dedicated node groups. For now, if/when scheduling issues are encountered, it should be possible to trivially reproduce them by deploying via tmpnetctl which does not tear down on failure so the network state on the cluster is inspectable. I'd much rather be able to investigate all the running elements of a network in the event of failure vs having a static snapshot that may not capture the full extent of a problem.

I largely agree with fixing problems as they come rather than over-anticipating, but I'd take a balanced approach, we shouldn't discard everything we can reasonably anticipate.

> it should be possible to trivially reproduce them by deploying via tmpnetctl which does not tear down on failure so the network state on the cluster is inspectable

This is true and helpful for debugging. However, I think this could become a problem when we enable engineers to work outside of CI - it is possible to hit these infrastructure issues more frequently and spend time on infra rather than their actual work.

I understand this is work-in-progress and we're exploring the problem space. For CI validation the current approach works well, but worth considering better error reporting for broader usage down the line.

Elvis339 (approved):

PR #3990, opened 2025-06-04
Introduce Simplex Components Into AvalancheGo

samliok:
## Why this should be merged

This PR should not be merged yet as it is a work in progress. More to be written in the description, once the PR is ready for a review.

The goal is to introduce the fundamental pieces of Simplex into AvalancheGo, so that future PRs could integrate them with rest of AvalancheGo.


simplex/bls.go
```diff
@@ -0,0 +1,119 @@
...
+var (
+	simplexLabel = []byte("<simplex>")
+)
+
+// BLSSigner signs messages encoded with the provided ChainID and NetworkID
+// using the SignBLS function.
+type BLSSigner struct {
+	chainID  ids.ID
+	subnetID ids.ID
+	signBLS  func(msg []byte) (*bls.Signature, error) // should we pass in a verifier function into BLSVerifier?
```
yacovm, line 21: the signer function here is because we support both software signing and hardware signing. When we verify, we only verify with software, so no function is needed

PR #3993, opened 2025-06-04
BLS Components for Simplex

samliok:
## Why this should be merged

Introduces `BLSSigner` and `BLSVerifier` in the new `simplex` package for signing and verifying messages (e.g., votes) in Simplex consensus. `BLSVerifier` will be extended in future PRs to support quorum certificate verification and additional functionality.

The newly created `Config` struct is provided during simplex engine initialization and contains only the fields necessary for the current functionality of the simplex package.

## How this works

In future PRs, the simplex package will be expanded to integrate simplex into AvalancheGo.

## How this was tested

Add unit tests in `bls_test.go`

No


simplex/bls.go
```diff
@@ -0,0 +1,127 @@
...
+	sig, err := s.signBLS(message2Sign)
+	if err != nil {
+		return nil, err
+	}
+
+	sigBytes := bls.SignatureToBytes(sig)
+	return sigBytes, nil
+}
+
+type encodedSimplexMessage struct {
```
yacovm: maybe `encodedSimplexSignedPayload` ?

simplex/bls.go
```diff
@@ -0,0 +1,127 @@
...
+	Message  []byte
```
yacovm: what about the `simplexLabel` ?

simplex/bls.go
```diff
@@ -0,0 +1,127 @@
...
+type SignFunc func(msg []byte) (*bls.Signature, error)
+
+// BLSSigner signs messages encoded with the provided ChainID and SubnetID.
+// using the SignBLS function.
+type BLSSigner struct {
+	chainID  ids.ID
+	subnetID ids.ID
+	// signBLS is passed in because we support both software and hardware BLS signing.
+	signBLS SignFunc
+	nodeID  ids.NodeID
```
yacovm: why do we need the `nodeID` ?

simplex/bls.go
```diff
@@ -0,0 +1,129 @@
...
+	verifier := BLSVerifier{
+		nodeID2PK: make(map[ids.NodeID]bls.PublicKey),
+		subnetID:  config.Ctx.SubnetID,
+		chainID:   config.Ctx.ChainID,
+	}
+
+	nodes := config.Validators.GetValidatorIDs(config.Ctx.SubnetID)
+	for _, node := range nodes {
+		validator, ok := config.Validators.GetValidator(config.Ctx.SubnetID, node)
+		if !ok {
```
yacovm: can we add a log here?

yacovm:
What about the [QC stuff](https://github.com/yacovm/avalanchego/blob/simplexIntegration/simplex/bls.go#L87-L201)? Is that a future PR?

yacovm (approved):
LGTM modulo two comments above

samliok:
> What about the [QC stuff](https://github.com/yacovm/avalanchego/blob/simplexIntegration/simplex/bls.go#L87-L201)? Is that a future PR?

Yep!

yacovm (approved):

simplex/bls.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+type encodedSimplexSignedPayload struct {
+	Message  []byte
+	ChainID  []byte
+	SubnetID []byte
+	Label    []byte
+}
+
+// encodesMessageToSign returns a byte slice [simplexLabel][chainID][networkID][message length][message].
```
StephenButtolph: Is this comment correct? I would have expected it to be in the reverse order.

simplex/bls.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	sig, err := s.signBLS(message2Sign)
+	if err != nil {
+		return nil, err
+	}
+
+	sigBytes := bls.SignatureToBytes(sig)
+	return sigBytes, nil
+}
```
StephenButtolph: This is expected to use the node's BLS key right? The same key used during Warp message signing and PoPs right?

Assuming this is the case, we need to be very careful around the format of the message being signed by this key (to ensure there are no [cross protocol attacks](https://www.startupdefense.io/cyberattacks/cross-protocol-attack)).

I am not extremely familiar with `asn1`. Would it make sense for us to use the normal avalanchego codec to clearly delineate between Warp messages and Simplex messages?

simplex/config.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/validators"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+type ValidatorInfo interface {
+	GetValidatorIDs(subnetID ids.ID) []ids.NodeID
+	GetValidator(subnetID ids.ID, nodeID ids.NodeID) (*validators.Validator, bool)
+}
```
StephenButtolph: Is this expected to be implemented by [`validators.Manager`](https://github.com/ava-labs/avalanchego/blob/master/snow/validators/manager.go#L40)?

If so, this doesn't really make sense to me, as the `validators.Manager` is _volatile_ and not safe to assume to be consistent across nodes.

simplex/test_util.go
```diff
@@ -0,0 +1,83 @@
...
+	if v.validators == nil {
+		return nil
+	}
+
+	ids := make([]ids.NodeID, 0, len(v.validators))
+	for id := range v.validators {
+		ids = append(ids, id)
+	}
+	return ids
+}
```
StephenButtolph:
```suggestion
func (v *testValidatorInfo) GetValidatorIDs(ids.ID) []ids.NodeID {
	return maps.Keys(v.validators)
}
```

simplex/test_util.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (v *testValidatorInfo) GetValidator(_ ids.ID, nodeID ids.NodeID) (*validators.Validator, bool) {
+	if v.validators == nil {
+		return nil, false
+	}
+
```
StephenButtolph:
```suggestion
```

simplex/bls.go
```diff
@@ -0,0 +1,131 @@
...
+func createVerifier(config *Config) BLSVerifier {
+	verifier := BLSVerifier{
+		nodeID2PK: make(map[ids.NodeID]bls.PublicKey),
+		subnetID:  config.Ctx.SubnetID,
+		chainID:   config.Ctx.ChainID,
+	}
+
+	nodes := config.Validators.GetValidatorIDs(config.Ctx.SubnetID)
+	for _, node := range nodes {
+		validator, ok := config.Validators.GetValidator(config.Ctx.SubnetID, node)
```
StephenButtolph: If it is expected for `Validators` to be volatile (kinda doubtful), then I think we should be using `GetMap` rather than these separate calls, as these are (logically) racy reads.

simplex/bls.go
yacovm: But we have the simplex label, shouldn't this diagonalize the message space signed?

StephenButtolph: It depends on where the simplex label is. For example, could someone issue a warp call on the C-chain with a payload containing the simplex label?

yacovm: Sure, **someone** might issue a warp call on the C-chain with a payload containing the Simplex label, but the question is - can an adversary coerce sufficient stake to do that?

simplex/bls.go
samliok: ah the comment was from before we were using `asn1` encoding. updated

simplex/bls.go
StephenButtolph: If a warp call is issued on the C-chain all nodes will be willing to sign that message

yacovm: The PoP is a public key which is different than what we have.

The Warp structure **looks** different to me:
```
	NetworkID     uint32 `serialize:"true"`
	SourceChainID ids.ID `serialize:"true"`
	Payload       []byte `serialize:"true"`
```

But I admit I haven't properly vetted that we cannot have a collision. 

However shouldn't we ask the same question regarding warp? Where do we diagonalize warp from other protocols?

simplex/config.go
samliok: yea rather than passing in the entire validators.Manager i wanted to strip out the methods we needed. Does that mean `GetMap` is not consistent across nodes? 

Its important that all nodes have a consistent view of the rest of the nodes in the network, so if we can't guarantee that with `validators.Manager` what should I use?

StephenButtolph: https://github.com/ava-labs/avalanchego/blob/master/snow/validators/state.go#L27-L34 will give a consistent validator set across nodes, but requires a P-chain height to be provided (which must be consistent across nodes)

samliok: sounds good, I updated and passed in a height of 0 for now. @yacovm for reconfiguration we will need to either create a new verifier or re-sync the `validators` map. Would it cause an issue since this height is based off the pchain height, and not the height of the finalized reconfiguration block? 🤔

samliok: the issue potentially being we would need to encode the pchain height into the finalized reconfiguration block & ensure the new validator is already part of the validator set at that encoded pchain height.

simplex/bls.go
StephenButtolph: Currently we only use the BLS keys for:
1. PoPs
2. Warp messages

simplex/bls.go
```diff
@@ -110,21 +113,22 @@ func (v BLSVerifier) Verify(message []byte, signature []byte, signer simplex.Nod
...
-func createVerifier(config *Config) BLSVerifier {
+func createVerifier(config *Config) (BLSVerifier, error) {
 	verifier := BLSVerifier{
 		nodeID2PK: make(map[ids.NodeID]bls.PublicKey),
 		subnetID:  config.Ctx.SubnetID,
 		chainID:   config.Ctx.ChainID,
 	}
 
-	nodes := config.Validators.GetValidatorIDs(config.Ctx.SubnetID)
+	nodes, err := config.Validators.GetValidatorSet(context.Background(), 0, config.Ctx.SubnetID)
```
StephenButtolph: Passing in `0` here is not really an option. We'll need the height to be provided in the config. Alternatively we could just pass in the entire `nodeID2PK` map in the config.

samliok: i like passing in the entire map. I think it makes the simplex config a bit simpler/clear. updated

simplex/bls.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func encodeMessageToSign(message []byte, chainID ids.ID, subnetID ids.ID) ([]byte, error) {
+	encodedSimplexMessage := encodedSimplexSignedPayload{
+		Message:  message,
+		ChainID:  chainID[:],
+		SubnetID: subnetID[:],
+		Label:    simplexLabel,
+	}
+	return asn1.Marshal(encodedSimplexMessage)
```
yacovm: So if we sign a [Codec encoded message](https://github.com/ava-labs/avalanchego/blob/f17ea6a7ab4036c41b693e47b94d8f0c81cb69ec/vms/platformvm/warp/unsigned_message.go#L50), we should do the same here.

simplex/bls.go
```diff
@@ -0,0 +1,125 @@
...
+	SubnetID []byte `serialize:"true"`
+	Label    []byte `serialize:"true"`
+}
+
+func encodeMessageToSign(message []byte, chainID ids.ID, subnetID ids.ID) ([]byte, error) {
+	encodedSimplexMessage := encodedSimplexSignedPayload{
+		Message:  message,
+		ChainID:  chainID[:],
+		SubnetID: subnetID[:],
+		Label:    simplexLabel,
```
StephenButtolph: What is the purpose of the simplexLabel? I think this is probably not needed because we use the codec version for namespace separation.

simplex/bls.go
```diff
@@ -0,0 +1,125 @@
...
+	ChainID  []byte `serialize:"true"`
+	SubnetID []byte `serialize:"true"`
+	Label    []byte `serialize:"true"`
+}
+
+func encodeMessageToSign(message []byte, chainID ids.ID, subnetID ids.ID) ([]byte, error) {
+	encodedSimplexMessage := encodedSimplexSignedPayload{
+		Message:  message,
+		ChainID:  chainID[:],
+		SubnetID: subnetID[:],
```
StephenButtolph: Why do we sign both the `ChainID` and the `SubnetID`? Perhaps this was intended to be the `NetworkID` rather than the `SubnetID`?

`ChainIDs` are unique for a given network, so we don't need a `SubnetID` here (in Warp messages we do not sign over a `SubnetID`). However, specifically the P-chain's `ChainID` is only unique within the some network. Unfortunately we defined the P-chain `ChainID` as the emptyID on _all_ networks. So we need to include the `NetworkID` to differentiate between Mainnet P-chain and Fuji P-chain (or other P-chain instances).

simplex/bls.go
```diff
@@ -0,0 +1,125 @@
...
+
+	sigBytes := bls.SignatureToBytes(sig)
+	return sigBytes, nil
+}
+
+type encodedSimplexSignedPayload struct {
+	Message  []byte `serialize:"true"`
+	ChainID  []byte `serialize:"true"`
+	SubnetID []byte `serialize:"true"`
+	Label    []byte `serialize:"true"`
```
StephenButtolph: This applies the below suggestions, as well as using `ids.ID` rather than `[]byte` when possible. `ids.ID` is packed more efficiently than `[]byte` (as the length is not included as it is fixed based on the type).
```suggestion
	NetworkID uint32 `serialize:"true"`
	ChainID   ids.ID `serialize:"true"`
	Message   []byte `serialize:"true"`
```

simplex/bls.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return Codec.Marshal(CodecVersion, &encodedSimplexMessage)
+}
+
+func (v BLSVerifier) Verify(message []byte, signature []byte, signer simplex.NodeID) error {
+	if len(signer) != ids.NodeIDLen {
+		return fmt.Errorf("expected signer to be %d bytes but got %d bytes", ids.NodeIDLen, len(signer))
+	}
+
+	key := ids.NodeID(signer)
```
StephenButtolph: nit: I probably would have used `ids.ToNodeID` here

simplex/bls.go
```diff
@@ -0,0 +1,125 @@
...
+// using the SignBLS function.
+type BLSSigner struct {
+	chainID  ids.ID
+	subnetID ids.ID
+	// signBLS is passed in because we support both software and hardware BLS signing.
+	signBLS SignFunc
+}
+
+type BLSVerifier struct {
+	nodeID2PK map[ids.NodeID]bls.PublicKey
```
StephenButtolph: Throughout the repo, we normally pass `bls.PublicKey`s as pointers. Would it make sense to stay consistent with that?

simplex/bls_test.go
```diff
@@ -0,0 +1,59 @@
...
+	dummyMsg := []byte("dummy message")
+	sig, err := config.SignBLS(dummyMsg)
+	require.NoError(t, err)
+
+	sigBytes := bls.SignatureToBytes(sig)
+
+	_, verifier := NewBLSAuth(config)
+	err = verifier.Verify(dummyMsg, sigBytes, config.Ctx.NodeID[:])
+	require.ErrorIs(t, err, errSignatureVerificationFailed)
+}
```
StephenButtolph: Could we convert these into test vectors for the `BLSVerifier`? In addition to `errSignerNotFound` and `errSignatureVerificationFailed`, I think we should have some tests for when the `NodeID` isn't the right length and when the signature is malformed.

simplex/test_util.go
```diff
@@ -0,0 +1,45 @@
...
+		NodeID:    nodeID,
+		PublicKey: ls.PublicKey(),
+	}
+
+	return &Config{
+		Ctx:        simplexChainContext,
+		Validators: newTestValidatorInfo([]validators.GetValidatorOutput{nodeInfo}),
+		SignBLS:    ls.Sign,
+	}, nil
+}
```
StephenButtolph: If the helper function is used later, feel free to keep it... But it looks unnecessary in this PR.
```suggestion
func newEngineConfig() (*Config, error) {
	ls, err := localsigner.New()
	if err != nil {
		return nil, err
	}

	nodeID := ids.GenerateTestNodeID()
	return &Config{
		Ctx: SimplexChainContext{
			NodeID:   nodeID,
			ChainID:  ids.GenerateTestID(),
			SubnetID: ids.GenerateTestID(),
		},
		Validators: map[ids.NodeID]*validators.GetValidatorOutput{
			nodeID: {
				NodeID:    nodeID,
				PublicKey: ls.PublicKey(),
			},
		},
		SignBLS: ls.Sign,
	}, nil
}
```

simplex/bls_test.go
samliok: changed these to table tests, and added a few more tests

simplex/test_util.go
samliok: in future prs i add more validators to `Validators` via the helper, so i think its worth keeping.

simplex/test_util.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	nodeID := ids.GenerateTestNodeID()
+
+	simplexChainContext := SimplexChainContext{
+		NodeID:    nodeID,
+		ChainID:   ids.GenerateTestID(),
+		NetworkID: constants.LocalID,
```
StephenButtolph:
```suggestion
		NetworkID: constants.UnitTestID,
```

simplex/config.go
```diff
@@ -0,0 +1,34 @@
...
+	// This tells the node about the current membership set, and should be consistent
+	// across all nodes in the subnet.
+	Validators map[ids.NodeID]*validators.GetValidatorOutput
+
+	// SignBLS is the signing function used for this node to sign messages.
+	SignBLS SignFunc
+}
+
+// Context is information about the current execution.
+// [SubnetID] is the ID of the subnet this context exists within.
```
StephenButtolph: Should be `NetworkID`

simplex/config.go
```diff
@@ -0,0 +1,34 @@
...
+// [ChainID] is the ID of the chain this context exists within.
+// [NodeID] is the ID of this node
```
StephenButtolph: I don't think the bracket notation is correct to use here, as these aren't godoc linkable symbols. (A good test for this is to run `godoc` locally and just look at how the docs are shown.)

simplex/bls_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	msg := []byte("Begin at the beginning, and go on till you come to the end: then stop")
+	tests := []struct {
+		name      string
+		expectErr error
+		nodeID    []byte
+		sig       []byte
+	}{
+		{
+			name:      "valid signature",
```
StephenButtolph: I have recently been convinced that test names should use `_` rather than spaces so that the golang test names displayed during test failures textually match the code for easier searching

simplex/bls_test.go
```diff
@@ -0,0 +1,88 @@
...
+			expectErr: errSignerNotFound,
+			nodeID:    otherNodeID[:],
+			sig: func() []byte {
+				sig, err := signer.Sign(msg)
+				require.NoError(t, err)
+				return sig
+			}(),
+		},
+		{
+			name:      "invalid encoding",
```
StephenButtolph:
```suggestion
			name:      "invalid_message_encoding",
```

simplex/bls.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return Codec.Marshal(CodecVersion, &encodedSimplexMessage)
+}
+
+func (v BLSVerifier) Verify(message []byte, signature []byte, signer simplex.NodeID) error {
+	if len(signer) != ids.NodeIDLen {
+		return fmt.Errorf("%w: expected signer to be %d bytes but got %d bytes", errInvalidNodeIDLength, ids.NodeIDLen, len(signer))
+	}
+
+	fmt.Println("verifying signature for signer", len(signer))
```
StephenButtolph:
```suggestion
```

simplex/bls.go
```diff
@@ -0,0 +1,130 @@
...
+		ChainID:   chainID,
+		NewtorkID: networkID,
+	}
+	return Codec.Marshal(CodecVersion, &encodedSimplexMessage)
+}
+
+func (v BLSVerifier) Verify(message []byte, signature []byte, signer simplex.NodeID) error {
+	if len(signer) != ids.NodeIDLen {
+		return fmt.Errorf("%w: expected signer to be %d bytes but got %d bytes", errInvalidNodeIDLength, ids.NodeIDLen, len(signer))
+	}
```
StephenButtolph: `ids.ToNodeID` already checks the length.

simplex/config.go
samliok: ah this is a cool command to know. I added brackets to keep this consistent with the other way we initialized the context

https://github.com/ava-labs/avalanchego/blob/65962974d6ea2fb946d884cfcb79c497dd4e1318/snow/context.go#L29-L34

should I keep it consistent or just update this/both?

StephenButtolph: I think we are trying to consistently only use brackets when the syntax is appropriate. There are definitely instances in some (not so) old code where we incorrectly used them... But for new code we should do it right

PR #3996, opened 2025-06-05
Load Framework 2 - POC

RodrigoVillar:
## Why this should be merged

This is a POC for what the load framework could look like and therefore should not be merged (currently)


tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
+		txsConfirmedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "txs_confirmed",
+			Help:      "Number of transactionsconfirmed",
+		}),
+		totalGasUsedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "total_gas_used",
+			Help:      "Sum of gas used for all confirmed transactions",
+		}),
```
joshua-kim: Is this a metric the client should track? Seems like we could just derive this from the node metrics

tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
 }
 
-// NewTracker returns a new Tracker instance which records metrics for the number
-// of transactions issued, confirmed, and failed. It also tracks the latency of
-// transactions.
-func NewTracker[T TxID](metrics *Metrics) *Tracker[T] {
-	return &Tracker[T]{
-		outstandingTxs: make(map[T]time.Time),
-		metrics:        metrics,
+func NewMetrics(registry *prometheus.Registry) (*Metrics, error) {
```
joshua-kim: I don't think we have to export the `Metrics` type out of this package - the caller just needs to specify the prometheus registry they pass in and we should be able to read the metrics from the metrics endpoint + write to the metrics are all handled within this package

tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
-	return t.txsConfirmed
+	return m, nil
 }
 
-// GetObservedFailed returns the number of transactions that the tracker has
-// confirmed failed.
-func (t *Tracker[_]) GetObservedFailed() uint64 {
-	t.lock.RLock()
-	defer t.lock.RUnlock()
+type Tracker struct {
```
joshua-kim: Similarly I don't think this needs to be exported either - this is basically a wrapper for our metrics. I think we can probably refactor this + `Metrics` under the same struct. We could probably even just have all of this under the orchestrator type.

tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
+	m := &Metrics{
+		txsIssuedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "txs_issued",
+			Help:      "Number of transactions issued",
+		}),
+		txsConfirmedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "txs_confirmed",
+			Help:      "Number of transactionsconfirmed",
```
joshua-kim: nit: typo

tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
-// of transactions issued, confirmed, and failed. It also tracks the latency of
-// transactions.
-func NewTracker[T TxID](metrics *Metrics) *Tracker[T] {
-	return &Tracker[T]{
-		outstandingTxs: make(map[T]time.Time),
-		metrics:        metrics,
+func NewMetrics(registry *prometheus.Registry) (*Metrics, error) {
+	m := &Metrics{
+		txsIssuedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
```
joshua-kim: Might be nice to support some prefixing here from the caller instead of forcing us to use the same metrics namespace

tests/load/sender.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+
+	ethereum "github.com/ava-labs/libevm"
+)
+
+type Builder interface {
+	// Create a valid transaction
+	BuildTx() (*types.Transaction, error)
+	// Increment the nonce of the issuer by 1
+	IncrementNonce()
```
joshua-kim: I feel like we don't need this... I think we could either provide the nonce from the caller if we read the initial nonce on startup, or just expect the implementation to just track the nonce itself since each call to `BuildTx` implies a nonce update

tests/load/options.go
```diff
@@ -0,0 +1,102 @@
...
+	return o
+}
+
+func (o *Options) applyOptions(ops []Option) {
+	for _, op := range ops {
+		op(o)
+	}
+}
+
+func (o *Options) Context() context.Context {
```
joshua-kim: Why is this an option? Can't we just force the caller to always specify a context?

tests/load/options.go
```diff
@@ -0,0 +1,102 @@
...
+	return context.Background()
+}
+
+func (o *Options) IssuanceHandler() func(IssuanceReceipt) {
+	return o.issuanceHandler
+}
+
+func (o *Options) ConfirmationHandler() func(ConfirmationReceipt) {
+	return o.confirmationHandler
+}
```
joshua-kim: I feel like this shouldn't be an option either, since we have to define these to do metric updates

tests/load/options.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (o *Options) PingFrequency() time.Duration {
+	return o.pingFrequency
+}
```
joshua-kim: If we delete this `Options` struct I think this can be moved to a param/field somewhere

tests/load/options.go
```diff
@@ -0,0 +1,102 @@
...
+	"context"
+	"time"
+
+	"github.com/ava-labs/libevm/core/types"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+// IssuanceReceipt is the information known after issuing a transaction.
+type IssuanceReceipt struct {
```
joshua-kim: 1. I think we can probably get rid of these receipt types if we get rid of the options type
2. I don't think we need them anyways, since we can just pass in the fields they contain as function params anyways

tests/load/sender.go
```diff
@@ -0,0 +1,112 @@
...
+	client *ethclient.Client
+}
+
+func NewSender(client *ethclient.Client) *Sender {
+	return &Sender{client: client}
+}
+
+func (w *Sender) SendTx(
+	ctx context.Context,
+	builder Builder,
```
joshua-kim: This function signature implies to me that we could have one sender with multiple builders... would it make more sense to just accept a `Builder` in the constructor instead of having it as a function param here?

tests/load/sender.go
```diff
@@ -0,0 +1,112 @@
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-ticker.C:
+		}
+
+		receipt, err := client.TransactionReceipt(ctx, txHash)
+		if err != nil {
+			if err == ethereum.NotFound {
+				continue
+			}
```
joshua-kim: Nit: use `errors.Is`, equality checks break with error wrapping

tests/load/tracker.go
```diff
@@ -4,94 +4,106 @@
...
+
+	g.metrics.txsIssuedCounter.Add(1)
+	g.metrics.txIssuanceLatency.Observe(float64(i.Duration))
 }
 
-// ObserveConfirmed records a transaction that was confirmed.
-func (t *Tracker[T]) ObserveConfirmed(txID T) {
-	t.lock.Lock()
-	defer t.lock.Unlock()
+func (g *Tracker) LogConfirmation(c ConfirmationReceipt) {
```
joshua-kim: nit: make receivers match the type (probably `t`), it looks like a bunch of structs in this package have a similar issue

tests/load/orchestrator.go
```diff
@@ -16,298 +13,83 @@ import (
...
+		g.tracker.LogConfirmation(c)
 	}
 
-	// stop the observers and issuers
-	cancel()
-
-	// block until both the observers and issuers have stopped
-	return errors.Join(o.issuerGroup.Wait(), o.observerGroup.Wait(), err)
-}
+	eg, cctx := errgroup.WithContext(ctx)
```
joshua-kim: To get around the weird `cctx` naming I would just name `ctx` -> `rootCtx` or `parentCtx` or call `cctx` -> `childCtx` or both.

tests/load/orchestrator.go
```diff
@@ -16,298 +13,83 @@ import (
...
+	builders []Builder
+	tracker  *Tracker
+	log      logging.Logger
 }
 
-func NewOrchestrator[T TxID](
-	agents []Agent[T],
-	tracker *Tracker[T],
+func NewOrchestrator(
+	wallets []*Sender,
```
joshua-kim: nit: wallets naming is weird to me since we call these `Sender`

tests/load/sender.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"time"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/ethclient"
+
+	"github.com/ava-labs/avalanchego/ids"
+
+	ethereum "github.com/ava-labs/libevm"
```
joshua-kim: Why alias the package here? Seems like there's no conflicts

RodrigoVillar: Hmm there's no conflicts but removing the pkg alias and trying to use `libevm.NotFound` fails because the pkg that lives at `github.com/ava-labs/libevm` is actually `ethereum`. I think you could make the case of keeping the pkg alias since it makes clear that `ethereum.NotFound` is derived from `libevm` but will defer to you for styling.

tests/load/orchestrator.go
RodrigoVillar: Done: 0faa5ce55

tests/load/tracker.go
RodrigoVillar: Done: 7997be7c8

tests/load/sender.go
RodrigoVillar: Done: 9a4d4bb82

tests/load/sender.go
RodrigoVillar: Changed so that the implementation tracks the nonce: 4d24e6ec4

tests/load/tracker.go
RodrigoVillar: Done: 7d8e7060e

tests/load/sender.go
joshua-kim: Ah, that's super confusing.

joshua-kim: Wait... could we just import it as libevm (no alias), but just still use the error type as `ethereum.NotFound`...? 👀

RodrigoVillar: Ya that's possible - just wondering if you thought if it was good style to have `ethereum.NotFound` but have it be derived from `"github.com/ava-labs/libevm"` (no type alias)

joshua-kim: Yeah IDK here, I'm fine keeping this package import. This seems to be a truly exceptional corner-case that I don't care enough to figure out what the right way to do this is. Explicit import seems fine to me.

tests/load/options.go
RodrigoVillar: Removed options altgother: c31ef977f

tests/load/options.go
RodrigoVillar: Removed options altgother: c31ef977f

tests/load/options.go
RodrigoVillar: Done: c31ef977f

tests/load/options.go
RodrigoVillar: Done: c31ef977f

tests/load/orchestrator.go
RodrigoVillar: Done: 1e95e58f7

tests/load/tracker.go
RodrigoVillar: Done: 0eeeb64f8

tests/load2/tracker.go
```diff
@@ -0,0 +1,41 @@
...
+
+type Tracker struct {
+	lock sync.RWMutex
+
+	totalGasUsed uint64
+	txsIssued    uint64
+	txsAccepted  uint64
+}
+
+func (t *Tracker) LogIssued(time.Duration) {
```
joshua-kim: 1. this parameter i unused, should we delete it?
2. Should we be using this parameter to track e2e latency?
3. This function is called `LogIssued` but doesn't actually perform logging, maybe we should call this `Issue`?

Similar thoughts on `LogAccepted`

tests/load2/main/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+	errFailedToCreateContract = errors.New("failed to create contract")
+)
+
+func init() {
+	flagVars = e2e.RegisterFlags()
+	flag.Parse()
+}
+
+func main() {
```
joshua-kim, line 48: What are we trying to do with this file? If we're trying to make a e2e test, I don't think the tests belong here since this isn't a test file. If we're trying to make a tool that runs against an endpoint, I don't think we need `tmpnet`.

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (g *Generator) Run(ctx context.Context) error {
+	tracker := &Tracker{}
+	issuerGroup, childCtx := errgroup.WithContext(ctx)
+
+	for i := range len(g.wallets) {
+		issuerGroup.Go(func() error {
+			for {
+				select {
+				case <-ctx.Done():
```
joshua-kim: Is this supposed to be the child context? This won't return if a sibling goroutine errors, which will only cancel the child and not the parent

tests/load2/wallet.go
```diff
@@ -0,0 +1,141 @@
...
+	ChainID() *big.Int
+	Signer() types.Signer
+}
+
+type backend struct {
+	privKey *ecdsa.PrivateKey
+	nonce   uint64
+	chainID *big.Int
+	signer  types.Signer
+}
```
joshua-kim: Do we need an interface here if we only have one implementation of it? It might make sense to just have a single struct for now and figure out what the interface needs to look like later if/when we need it

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+	"time"
+
+	"github.com/ava-labs/libevm/core/types"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+const pingFrequency = time.Millisecond
```
joshua-kim: This constant feels too small + is inconsistent behavior w/ the `Wallet` which allows a caller-defined frequency

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+
+type Generator struct {
+	log        logging.Logger
+	wallets    []*Wallet
+	txBuilders []TxBuilder
+}
+
+func NewGenerator(
+	log logging.Logger,
+	wallets []*Wallet,
```
joshua-kim: Why do we need pointers here? Seems like `Wallet` has no mutable state and the types it wraps have pointers inside of them anyways

tests/load2/tx_builders.go
```diff
@@ -0,0 +1,355 @@
...
+	return contractInstance.SimulateExternalCall(txOpts)
+}
+
+type txType struct {
+	txFunc func(Backend, *contracts.EVMLoadSimulator) (*types.Transaction, error)
+	name   string
+	weight uint
+}
+
+func pickWeightedRandom(txTypes []txType) (txType, error) {
```
joshua-kim: We already have this in avalanchego under `sampler`, can we avoid this implementation?

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load2
```
joshua-kim, line 4: I think we should either just do this in the `load` package or delete the old one as part of this PR - I don't know if we want the repo to be in a state where we maintain two copies of the same tests/load testing component

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+	log        logging.Logger
+	wallets    []*Wallet
+	txBuilders []TxBuilder
+}
+
+func NewGenerator(
+	log logging.Logger,
+	wallets []*Wallet,
+	txBuilders []TxBuilder,
+) (*Generator, error) {
```
joshua-kim: Some question - doesn't look like this type needs to be a pointer either

tests/load2/wallet.go
```diff
@@ -0,0 +1,141 @@
...
+	confirmationHandler(receipt, confirmationDuration)
+
+	w.backend.nonce++
+
+	return nil
+}
+
+func (w *Wallet) Backend() Backend {
+	return w.backend
+}
```
joshua-kim: The `Backend` type + this function on the `Wallet` feel like a code smell. It's not clear to me what a `Backend` is + the abstraction feels artificial. I think it's worth exploring to see if we can rethink the api to see if we can get rid of it.

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+
+		currTotalGasUsed := tracker.TotalGasUsed()
+		currTime := time.Now()
+
+		gps := computeDiff(prevTotalGasUsed, currTotalGasUsed, currTime.Sub(prevTime))
+		g.log.Info("stats", zap.Uint64("gps", gps))
+
+		prevTotalGasUsed = currTotalGasUsed
+		prevTime = currTime
+	}
```
joshua-kim: Wondering if we need this worker loop to emit gps logs. Is this a metric we should track from the client-side? This information I think is available from the EVM on the network's side, so I'm wondering if we should just have a single source of truth here.

tests/load2/generator.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/ava-labs/libevm/core/types"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+const pingFrequency = time.Millisecond
+
+type TxBuilder func(Backend) (*types.Transaction, error)
```
joshua-kim: 1. The use of functions over interfaces here is a bit awkward - it also doesn't behave well with IDEs since you can't look up types that satisfy this.
2. This type is awkward to use... we have to guarantee a 1:1 relationship with wallets and builders as well. I'm wondering if there's a way we can re-think the relationship between the tx builders and wallets/do some refactor so the caller only has to specify one of these.

tests/load2/random_issuer.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/ethclient"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/tests"
+	"github.com/ava-labs/avalanchego/tests/load/c/contracts"
+)
+
+func TestRandomTx(
```
joshua-kim: This test doesn't seem to be used

tests/load2/tracker.go
joshua-kim, line 1: Go doesn't require separate files for each type (types typically just live close to where they are used) - if we un-export this it would probably make more sense to just group this under the `generator.go` file

tests/load2/tx_builders.go
joshua-kim, line 1: What do we need this for?

tests/load2/main/main.go
maru-ava: The _tests_ run against one or more node URIs, yes. The load test (this) that orchestrates those tests is responsible for deploying a new network or reusing an existing one from which to provide those URIs. 

tmpnet enables a test suite to deploy networks configured to its liking.

tests/load2/tx_builders.go
RodrigoVillar: These are the functions that produce transactions for the load tests. We call `BuildRandomTx()` for the load tests, but as per the PRD, each tx type has its own function to allow testing of that individual tx.

tests/load2/random_issuer.go
RodrigoVillar: This was meant as an example of how the new framework can be reused for testing contexts besides that of load testing. Moving this to the PR description since it's not necessary for this PR: ac3d360

tests/load2/tracker.go
RodrigoVillar: Done: a56c461

tests/load2/generator.go
RodrigoVillar: Done: 9f60eee

tests/load2/generator.go
RodrigoVillar: Done: 9f60eee

tests/load2/tx_builders.go
RodrigoVillar: Done: af93208

tests/load2/tracker.go
joshua-kim: I think you may have forgotten to un-export this type?

tests/load2/generator.go
RodrigoVillar: Ah good catch - done: 4817eb2

tests/load2/tracker.go
RodrigoVillar: Done: e518a86

tests/load2/wallet.go
RodrigoVillar: Removed `Backend`: 6fa2323

tests/load2/wallet.go
RodrigoVillar: This is a good point - I've removed the `Backend` component entirely and merged it into the `Wallet`: 6fa2323

tests/load2/generator.go
RodrigoVillar: Changed `pingFrequency` to be a parameter set by the caller: 037cd06. 

For the duration (currently `time.Millisecond`), my rationale was that the wallet should block for the lowest amount of time necessary to hear about tx acceptance, which setting the ping frequency to a small value accomplishes. Open to changing the duration though.

tests/load2/generator.go
maru-ava: I requested this parallel implementation. I want the new load implementation to stand on its own rather than being an evolution of the original, and we can remove the original package in a subsequent PR.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/ava-labs/avalanchego/tests"
+)
+
+type Tracker struct {
+	lock sync.RWMutex
+
+	txsIssued   uint64
+	txsAccepted uint64
+
+	txsIssuedCounter      prometheus.Counter
```
maru-ava: (No action required) Are there avalanchego metrics for issued and accepted?

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+		}),
+		txsAcceptedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "txs_confirmed",
+			Help:      "Number of transactions confirmed",
+		}),
+		txIssuanceLatency: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "tx_issuance_latency",
+			Help:      "issuance latency of transactions",
```
maru-ava: Maybe ensure consistent casing of help messages across metrics?

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/ava-labs/avalanchego/tests"
+)
+
+type Tracker struct {
```
maru-ava: (No action required) Maybe move Tracker into its own file? It seems separable from Generator.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+type Generator struct {
+	wallets []Wallet
+	txTests []TxTest
+}
+
+func NewGenerator(
+	wallets []Wallet,
+	txTests []TxTest,
+) (Generator, error) {
+	if len(wallets) != len(txTests) {
```
maru-ava: (No action required) Why does there need to be a correspondence between the number of wallets and tests?

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/ava-labs/avalanchego/tests/load/c/contracts"
+	"github.com/ava-labs/avalanchego/tests/load2"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+
+	ethcommon "github.com/ava-labs/libevm/common"
+)
+
+const (
+	blockchainID = "C"
+	// invariant: nodesCount >= 5
```
maru-ava: Please remove this invariant, I don't think it makes sense. The number of nodes is likely to vary depending on testing requirements rather than be something static.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	nodesCount    = 5
```
maru-ava: Please update the flag handling to enable setting a default value provided by a test suite rather than hard-coding a value here.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	agentsPerNode = 5
+	agentsCount   = nodesCount * agentsPerNode
+	logPrefix     = "avalanchego-load-test"
+	namespace     = "load"
```
maru-ava: (No action required) Given how overloaded the term 'namespace' is, maybe use a more descriptive name (e.g. metricsNamespace)?

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	ethcommon "github.com/ava-labs/libevm/common"
+)
+
+const (
+	blockchainID = "C"
+	// invariant: nodesCount >= 5
+	nodesCount    = 5
+	agentsPerNode = 5
+	agentsCount   = nodesCount * agentsPerNode
+	logPrefix     = "avalanchego-load-test"
```
maru-ava: (No action required) Why is this log prefix suggested given that all log lines include it?

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	namespace     = "load"
+	pollFrequency = time.Millisecond
```
maru-ava: (No action required) What are the implications of polling for acceptance vs using the previous approach of SubscribeNewHead? I'm guessing this switch was to enable acceptance to be a detail internal to a given tx-issuing test, but I'm not sure why that needs to require polling.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+
+	require := require.New(tc)
+	ctx := context.Background()
+
+	nodes := tmpnet.NewNodesOrPanic(nodesCount)
+
+	keys, err := tmpnet.NewPrivateKeys(agentsCount)
+	require.NoError(err)
+	network := &tmpnet.Network{
+		Owner:         "avalanchego-load-test",
```
maru-ava: `Owner` can be removed - it's only desirable when a test suite expects to more than one network.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+
+	e2e.NewTestEnvironment(tc, flagVars, network)
+	tc.DeferCleanup(func() {
+		require.NoError(network.Stop(ctx), "failed to stop network")
+	})
+
+	wsURIs, err := tmpnet.GetNodeWebsocketURIs(ctx, network.Nodes, blockchainID, tc.DeferCleanup)
+	require.NoError(err)
+
+	registry := prometheus.NewRegistry()
```
maru-ava, line 73: (No action required) Maybe internalize registry instantiation to Tracker given that it isn't use anywhere else?

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+		wsURI := wsURIs[i%len(wsURIs)]
+		client, err := ethclient.Dial(wsURI)
+		require.NoError(err)
+
+		chainID, err := client.ChainID(ctx)
+		require.NoError(err)
+
+		wallet := load2.NewWallet(client, keys[i].ToECDSA(), 0, chainID)
+
+		contract, err := createContract(ctx, client, wallet)
```
maru-ava: Please simplify this PR by targeting the simplest possible transaction - e.g. a zero balance transfer. 

Integrating the load-simulating contract can be a subsequent PR, and one that avoids coupling test setup to contract usage e.g. with an interface that supports optional one-time setup e.g.

```golang
type TestIterator interface {
  SetUp(...) // Executed once during suite setup e.g. deploy contract
  Next(...) ReusableTest   // Returns the next test to execute against the contract
}
```

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+
+	wsURIs, err := tmpnet.GetNodeWebsocketURIs(ctx, network.Nodes, blockchainID, tc.DeferCleanup)
+	require.NoError(err)
+
+	registry := prometheus.NewRegistry()
+	tracker, err := load2.NewTracker(namespace, registry)
+	require.NoError(err)
+
+	issuanceF := func(i common.IssuanceReceipt) {
```
maru-ava: (No action required) Why is it desirable to define these anonymous functions rather than just supplying a tracker instance to the test-only wallet?

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	flag.Parse()
+}
+
+func main() {
+	log := tests.NewDefaultLogger(logPrefix)
+	tc := tests.NewTestContext(log)
+	defer tc.Cleanup()
+
+	require := require.New(tc)
+	ctx := context.Background()
```
maru-ava: Please use `tests.DefaultNotifyContext` to ensure cleanup (e.g. network shutdown) is performed when SIGTERM and SIGINT signals are recieved.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+
+	t.txsAccepted++
+
+	t.txTotalLatency.Observe(float64(totalDuration.Milliseconds()))
+	t.txConfirmationLatency.Observe(float64(confirmationDuration.Milliseconds()))
+}
+
+type TxTest func(tests.TestContext, context.Context, Wallet)
+
+type Generator struct {
```
maru-ava: (No action required) Why is 'generator' an appropriate name for this type? It looks more like it's executing tests than generating them.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+		txTests[i] = txTest
+	}
+
+	generator, err := load2.NewGenerator(wallets, txTests)
+	require.NoError(err)
+
+	generator.Run(tc, ctx)
+}
+
+func createContract(
```
maru-ava: (No action required) Why does this belong in the main entrypoint? This prevents its reuse or testing.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+		chainID, err := client.ChainID(ctx)
+		require.NoError(err)
+
+		wallet := load2.NewWallet(client, keys[i].ToECDSA(), 0, chainID)
+
+		contract, err := createContract(ctx, client, wallet)
+		require.NoError(err)
+
+		txTest := func(tc tests.TestContext, ctx context.Context, w load2.Wallet) {
+			load2.TestRandomTx(tc, ctx, w, contract)
```
maru-ava: Maybe ensure the use of contexts with timeouts when executing transactions to insure against an infinite loop in the wallet if a transaction gets dropped?

tests/load2/main/main.go
Elvis339: Should we deploy contract once only?

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+			defer wg.Done()
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
+
+				g.txTests[i](tc, ctx, g.wallets[i])
```
Elvis339: This will hammer the network as fast as possible. For load testing, we want to have configurable delays between transactions and backoff on errors (probably).

Is this intentional?

tests/load2/wallet.go
```diff
@@ -0,0 +1,172 @@
...
+	ethereum "github.com/ava-labs/libevm"
+	ethcommon "github.com/ava-labs/libevm/common"
+)
+
+var (
+	_ Wallet = (*wallet)(nil)
+	_ Wallet = (*walletWithOptions)(nil)
+)
+
+type Wallet interface {
```
Elvis339: What happens if a wallet runs out of funds during a long-running test?

tests/load2/generator.go
Elvis339: Should Tracker be renamed to Metrics? Imo Tracker implies internal state is used somewhere else whereas Metrics is pretty straightforward.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/ava-labs/avalanchego/tests"
+)
+
+type Tracker struct {
+	lock sync.RWMutex
+
+	txsIssued   uint64
+	txsAccepted uint64
```
Elvis339: `txsIssued` and `txsAccepted` fields are incremented but never read. They should probably be removed along with mutex. Prometheus counters are already thread-safe.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+			Help:      "Number of transactions issued",
+		}),
+		txsAcceptedCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "txs_confirmed",
+			Help:      "Number of transactions confirmed",
+		}),
+		txIssuanceLatency: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "tx_issuance_latency",
```
Elvis339: What does tx_issuance_latency actually measure? If it's just RPC call time, consider removing as it adds noise compared to confirmation latency.

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+			Namespace: namespace,
+			Name:      "tx_confirmation_latency",
+			Help:      "confirmation latency of transactions",
+		}),
+		txTotalLatency: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "tx_total_latency",
+			Help:      "total latency of transactions",
+		}),
+	}
```
Elvis339: What's the difference between those two?

tests/load2/generator.go
```diff
@@ -0,0 +1,138 @@
...
+		registry.Register(t.txConfirmationLatency),
+		registry.Register(t.txTotalLatency),
+	); err != nil {
+		return nil, err
+	}
+
+	return t, nil
+}
+
+func (t *Tracker) Issue(d time.Duration) {
```
Elvis339: Missing `t.txsIssuedCounter.Inc()`

tests/load2/wallet.go
```diff
@@ -0,0 +1,172 @@
...
+	pingFrequency time.Duration,
+) error {
+	ticker := time.NewTicker(pingFrequency)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
```
Elvis339: The function waits for the first tick before checking the transaction.

```go
// Check once immediately
if receipt, err := client.TransactionReceipt(ctx, txHash); err == nil {
   ...
}

// Then start polling
for {
    select {
    case <-ctx.Done():
        return ctx.Err() 
    case <-ticker.C: 
    }
}
```

tests/load2/main/main.go
RodrigoVillar: Was following the existing convention we have for C-Chain wallets: https://github.com/ava-labs/avalanchego/blob/a778bf376553999cb1eb12639dc3871321f0c03c/wallet/subnet/primary/common/options.go#L71-L72

However, it makes more sense to just have an implementation of the wallet that takes in a tracker.

tests/load2/generator.go
RodrigoVillar: The type is executing tests, but since this type will mainly be used to generate load against the network, the name `generator` made sense to me. 

We could rename it to `Executor` or something similar but I believe the new name wouldn't be reflective of its expected usage.

tests/load2/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+	require.NoError(err)
+	network := &tmpnet.Network{
+		Owner:         "avalanchego-load-test",
+		Nodes:         nodes,
+		PreFundedKeys: keys,
+	}
+
+	e2e.NewTestEnvironment(tc, flagVars, network)
+	tc.DeferCleanup(func() {
+		require.NoError(network.Stop(ctx), "failed to stop network")
```
maru-ava: It is both unnecessary and counterproductive to stop the network here. `e2e.NewTestEnvironment` is already registering a cleanup, and does so conditionally depending on flags to e.g. support `--reuse-network`.

maru-ava: Note that cleanup functions should always define their own context objects. As per the example here, the main context for the process is intended to control regular execution. At the point of a cleanup function being called, regular execution has ended and with it any guarantee that a context defined outside of the cleanup function hasn't been canceled.

tests/load2/generator.go
maru-ava: Maybe just remove the ambiguity with a more specific name - LoadGenerator?

tests/load2/main/main.go
maru-ava: Maybe consider one of the following rather than the current middle ground?

- evolve the existing c-chain wallet so that it supports the functionality required for load testing
- evolve the proposed implementation in this PR to be simpler and focused solely on the load testing use case

tests/load2/generator.go
maru-ava: Is it just RPC call time? When a system is overloaded responsiveness to tx issuance may suffer. Maybe not in the case where we overprovision API nodes, but is that likely to universally be the case in production?

tests/load2/generator.go
RodrigoVillar: There's metrics for accepted (`avalanche_evm_eth_chain_txs_accepted`) but not for issuance. I would prefer to keep our issuance/acceptance metrics for consistency but acknowledge the potential for multiple truths here.

tests/load2/main/main.go
RodrigoVillar: Done: 8ab1562

tests/load2/main/main.go
RodrigoVillar: Done: 8ab1562

tests/load2/generator.go
RodrigoVillar: Done: 175ce26

tests/load2/main/main.go
RodrigoVillar: Done: 93f035b89a911eb5c86e126ca4fb17cafe199e30

tests/load2/main/main.go
RodrigoVillar: Removed network stop here: 1e75103

tests/load2/main/main.go
RodrigoVillar: Done: 1e75103

tests/load2/main/main.go
RodrigoVillar: Removed log prefix: 803419be4ad91066059ab76c90378c63fbb5b330

tests/load2/main/main.go
RodrigoVillar: Renamed to `metricsNamespace`: 803419be4ad91066059ab76c90378c63fbb5b330

tests/load2/main/main.go
RodrigoVillar: Switched to using `node-count` flag via `tmpnet` - default number of nodes will be 2: d540b96b1798d9d44d7d31b4e2789a9782d1663f

tests/load2/wallet.go
RodrigoVillar: Done: 8910183e592ecd99d7c54ecf748a182921efdd00

tests/load2/main/main.go
RodrigoVillar: Resolving this since the test contract is no longer part of this PR.

tests/load2/generator.go
RodrigoVillar: Done: d93635f84518697305ffab29957b0cdc590a23c4

tests/load2/generator.go
RodrigoVillar: Great point - removed these fields: ad399463acf3e5c4093933d1d80e767b21d56668

tests/load2/generator.go
RodrigoVillar: Resolving since this field was removed here: ad399463acf3e5c4093933d1d80e767b21d56668

tests/load2/main/main.go
RodrigoVillar: Added a timeout for tests within the generator: e450c9a6c666903ffe62f9185063bd5ae082b1bd

tests/load2/main/main.go
RodrigoVillar: Took the second approach and simplified the wallet to support what's necessary for the load testing case: 02e8d6ae7eaf823ea8a9845de9c838b5f7e9bf94

tests/load2/generator.go
RodrigoVillar: This would be redundant since the package name itself is `load` - any client of the package (e.g. `main`) would need to specify `load.LoadGenerator`.

tests/load2/generator.go
RodrigoVillar: Confirmation latency: time between issuance and TX confirmation (does not include duration of issuance)
Total latency: time between start of the test and TX confirmation

tests/load2/generator.go
RodrigoVillar: Yes. For an MVP, we just need a tool that produces load against a network. In future PR's, we can modify the generator/add additional generators that produce the desired behaviors.

tests/load2/wallet.go
RodrigoVillar: If the wallet runs out of funds, then `client.SendTransaction()` should return an error, and therefore the test would fail.

tests/load2/generator.go
RodrigoVillar: Issuance latency is the RPC call time. We previously agreed on keeping this and I think it's a useful metric to have since it tells us whether our bottlenecks are coming from API calls or the system itself.

tests/load2/main/main.go
RodrigoVillar: Resolving since we've removed contracts from this PR.

tests/load2/generator.go
RodrigoVillar: Great point - since we're just using a single test for now, going to reduce the generator to take in a single test: f7f5b6b

tests/load2/main/main.go
RodrigoVillar: `SubscribeNewHead` ran in an async process (part of `Listener`) which we no longer have under the new issue-wait model.  I chose polling since it's what we currently do in the existing C-Chain wallet but I'm open to exploring utilizing `SubscribeNewHead` in a future PR.

tests/load2/generator.go
Elvis339: So confirmation latency excludes issuance time, but total latency includes everything? What's happening in issuance that's not captured by the issuance latency metric we were just [discussing](https://github.com/ava-labs/avalanchego/pull/3996#discussion_r2158849271)?

If we have:
- Issuance latency (RPC call time)
- Confirmation latency (issuance → confirmation, excluding issuance duration)
- Total latency (test start → confirmation)

Then total latency should just be issuance + confirmation latency, right? But this definition suggests there's some other unmeasured time component.

In a properly functioning system, `SendTransaction` should return almost immediately it's just accepting the TX into the mempool. If this call is taking significant time we are likely measuring network latency and serialization/deserialization overhead

These aren't the bottlenecks we should care about in load testing imo, the real bottlenecks are:
- Mempool saturation
- Gas price dynamics

Which we can derive from the system itself.

cc @maru-ava

tests/load2/generator.go
RodrigoVillar: Done: 3849da1

tests/load/tracker.go
RodrigoVillar: Removed `totalGasUsedCounter` from metrics

tests/load2/generator.go
RodrigoVillar: Removed GPS logs from `Generator`

tests/load2/generator.go
RodrigoVillar: Switched from using function types to interfaces here: 93f035b89a911eb5c86e126ca4fb17cafe199e30. Also removed the bijection between wallets and tests here: f7f5b6b231401b60af7195dd390d6ab0391166df

tests/load2/generator.go
RodrigoVillar: Moved `Metrics` to it's own file: 3849da1

tests/load2/main/main.go
RodrigoVillar: Going to keep registry instantiation in `main.go` since we use it for the metrics server.

tests/load/sender.go
RodrigoVillar: Switched from having a `Builder` to `SendTx()` just taking in a transaction.

tests/load2/generator.go
RodrigoVillar: Resolved offline; we decided to keep issuance time and will remove in the future if we deem it to not be useful.

tests/load2/generator.go
RodrigoVillar: Resolved offline; we decided to keep issuance time and will remove in the future if we deem it to not be useful.

tests/load2/tx_builders.go
RodrigoVillar: Removed contract here: d2436f7

tests/load2/tracker.go
RodrigoVillar: Removed the `Log` prefix from these methods and the parameters are being used now. I'm not sure what you mean in (2)

tests/load/tracker.go
RodrigoVillar: We have just a single metrics object now (`Metrics`). This type is exported since we need it in `main.go` for the prometheus server.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,19 @@ jobs:
           prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
           loki_username: ${{ secrets.LOKI_ID || '' }}
           loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
+  load2:
+    name: Run load2 test
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v4
+        - uses: ./.github/actions/setup-go-for-project
+        - uses: ./.github/actions/install-nix
```
maru-ava: It's not necessary to install nix, `run-monitored-tmpnet-cmd` takes care of that.

tests/load2/generator.go
maru-ava: By that reasoning, does every other type in this package also have an implicit 'Load' prefix?

I think my original point stands - just because a type will be prefixed by a package name in some uses doesn't remove the need to use a sufficiently descriptive name.

tests/load2/generator.go
maru-ava: What is the value of counting acceptance twice?

tests/load2/main/main.go
maru-ava: Is that default acceptable? As per my original comment it wouldn't be hard to enable the registration of flags with a default number of nodes.

tests/load2/main/main.go
maru-ava: This poll frequency requires making an API call 1,000 times per second for every issued tx for the duration between issuance and acceptance. And decreasing that to some lesser value(e.g. 10ms/100ms/1s/etc) would result in a consequent decrease in the accuracy of measurement.

I pointed to the existing wallets as a model of how to simplify tx construction to make tests easier to write. I did not intend to suggest that their current implementation was appropriate for pushing transactions at high-scale. Poll-based determination of acceptance latency is clearly unworkable for load testing.

I'd put event-based measurement of acceptance latency at the top of the list of things that need to get fixed in subsequent PRs.

tests/load2/main/main.go
maru-ava: As per our discussion offline, the notify context won't ensure cleanup until the recover methods from https://github.com/ava-labs/avalanchego/pull/3554 become available.

tests/load2/wallet.go
```diff
@@ -0,0 +1,116 @@
...
+	totalDuration := time.Since(startTime)
+	confirmationDuration := totalDuration - issuanceDuration
+	w.metrics.Accept(confirmationDuration, totalDuration)
+
+	w.nonce++
+
+	return nil
+}
+
+func (w *Wallet) PrivKey() *ecdsa.PrivateKey {
```
maru-ava: (No action required) The wallet might have had good reason to protect the its internals, but for test code, I think the value is a lot less. I'd rather see these fields become public if they need to be available externally than add accessor methods like this.

tests/load2/wallet.go
```diff
@@ -0,0 +1,116 @@
...
+	txHash common.Hash,
+	pingFrequency time.Duration,
+) error {
+	ticker := time.NewTicker(pingFrequency)
+	defer ticker.Stop()
+
+	for {
+		receipt, err := client.TransactionReceipt(ctx, txHash)
+		if err == nil {
+			if receipt.Status != 1 {
```
maru-ava: What does 1 mean in this instance? If geth doesn't provide a constant for this particular status whose name indicates its import, maybe add one locally. Filed under 'avoid magic numbers'.

tests/load2/main/main.go
```diff
@@ -0,0 +1,121 @@
...
+var (
+	flagVars *e2e.FlagVars
+
+	loadTimeout int
+)
+
+func init() {
+	flagVars = e2e.RegisterFlags()
+
+	flag.IntVar(
```
maru-ava: (No action required) Maybe use DurationVar to permit providing durations in denominations other than seconds?

tests/load2/generator.go
RodrigoVillar: Removed `txsIssuedCounter` here (i.e. going to use `avalanche_evm_eth_chain_txs_accepted`): 55b9bf9

tests/load2/generator.go
RodrigoVillar: Renamed generator here: 8495c17

tests/load2/main/main.go
RodrigoVillar: Great point - changed to using `DurationVar` here: c4ee87e

tests/load2/wallet.go
RodrigoVillar: The status code `1` represents a TX whose execution succeeded. Modified this to use the value used by `geth`: 550c701

.github/workflows/ci.yml
RodrigoVillar: Removed the `install-nix` step here: e9a62a7

tests/load2/main/main.go
RodrigoVillar: Noted - will add event-based acceptance in a subsequent PR.

tests/load2/wallet.go
RodrigoVillar: Removed accessor methods here: 461686a

tests/load2/generator.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ava-labs/avalanchego/tests"
+)
+
+type Test interface {
+	Run(tests.TestContext, context.Context, *Wallet)
+}
```
joshua-kim: nit: I like having named parameters so that when my IDE autogenerates code it can just name them whatever the interface named them
```suggestion
type Test interface {
	Run(tc tests.TestContext, ctx context.Context, wallet *Wallet)
}
```

tests/load2/generator.go
```diff
@@ -0,0 +1,69 @@
...
+	wallets []*Wallet,
+	test Test,
+) (LoadGenerator, error) {
+	return LoadGenerator{
+		wallets: wallets,
+		test:    test,
+	}, nil
+}
+
+func (g LoadGenerator) Run(
```
joshua-kim: nit: update receiver name. The convention is usually always lowercase, but there can be some differences based on what you're naming. In this case I'd probably choose` l` because I like single-letter names, but `lg` is also acceptable but it's less idiomatic imo.
* `f` for `(F)oo` (first letter, idiomatic w/ stdlib)
* `fb` for `(F)oo(B)ar` (first letter of each word, done in avalanchego in some spots like `vm` and `db` but is not idiomatic w/ stdlib)

tests/load2/generator.go
```diff
@@ -0,0 +1,69 @@
...
+	testTimeout time.Duration,
+) {
+	wg := sync.WaitGroup{}
+
+	childCtx := ctx
+	if loadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(ctx, loadTimeout)
+		childCtx = ctx
+		defer cancel()
+	}
```
joshua-kim, line 76: Some behavior of these timeouts are confusing, as an example we special-case `loadTimeout` but not `testTimeout` so that there's inconsistent behavior + a timeout of zero is ignored. I think if we go with the related comment to move the workers/metrics into `LoadGenerator`, we can remove the context from `Run`, which will clean up this code here.

tests/load2/generator.go
```diff
@@ -0,0 +1,69 @@
...
+type Test interface {
+	Run(tests.TestContext, context.Context, *Wallet)
+}
+
+type LoadGenerator struct {
+	wallets []*Wallet
+	test    Test
+}
+
+func NewLoadGenerator(
```
joshua-kim, line 34: nit: If a line isn't > our character limit (i think it's 80) you don't need to chop the parameters into separate lines

tests/load2/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+package load2
+
+import (
+	"errors"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Metrics struct {
```
joshua-kim: I'm still unclear on why this needs to be exported outside the package... it seems like a leaky abstraction to have the caller know about how to read/write to metrics (which the load generator actually owns) when it only cares about the prometheus registry for its metrics server.

1. The prometheus server only cares about the registry, so we can initialize the load generator with `NewLoadGenerator(..., prometheusRegistry)` so that the caller can still create the server since it will still know about the metrics registry
2. Currently the caller has to know about the `Metrics` type since it's shared across the `Wallet`s and the `LoadGenerator`.  Can we just make the `LoadGenerator` initialize the wallets, so that it can also own the `Metrics`? I think we can do this by having the caller specify the wallets it wants to create, and have the `LoadGenerator` create the metrics instance + wallets internally.

pseudocode:
```
// Test
registry := prometheus.NewRegistry()
wallets := []Wallet{
	{
		Key: pk1,
		Nonce: 1,
		ChainID: ids.ID{},
		client: client,
	},
	{
		Key: pk,
		Nonce: 2,
		ChainID: ids.ID{},
		client: client,
	},
	...
}

l := NewLoadGenerator(registry, wallets)
metricsServer := NewServer("localhost:1234", registry)
...


// Code
type worker struct {
	wallet Wallet
	metrics metrics
}

...

type LoadGenerator struct {
	workers []worker
	...
}
```

tests/load2/wallet.go
```diff
@@ -0,0 +1,101 @@
...
+	Signer  types.Signer
+
+	client  *ethclient.Client
+	metrics Metrics
+}
+
+func NewWallet(
+	privKey *ecdsa.PrivateKey,
+	nonce uint64,
+	chainID *big.Int,
```
joshua-kim, line 33: I don't know what our requirements for this are, but are we supporting testing against multiple chains simultaneously? Or should this be a `LoadGenerator`-level config, instead of a `Wallet`-level one?

tests/load2/transfer.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/ava-labs/libevm/crypto"
+	"github.com/ava-labs/libevm/params"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/tests"
+)
+
+var _ Test = (*ZeroTransferTest)(nil)
+
+type ZeroTransferTest struct {
```
joshua-kim: Why is this not defined in `main`? I'm guessing that this supposed to be a re-usable test?

tests/load2/transfer.go
```diff
@@ -0,0 +1,52 @@
...
+		GasTipCap: gasTipCap,
+		GasFeeCap: gasFeeCap,
+		Gas:       params.TxGas,
+		To:        &senderAddress,
+		Data:      nil,
+		Value:     common.Big0,
+	})
+	require.NoError(err)
+
+	require.NoError(wallet.SendTx(ctx, tx, z.PollFrequency))
```
joshua-kim: If we do the earlier suggestion of having the `LoadGenerator` own the metrics/workers, we can just return the tx here, and the `LoadGenerator` can send the tx for us which removes some complexity from the test implementation. We could also rename `Run` to `NewTx` so it becomes a factory pattern.

tests/load2/transfer.go
joshua-kim, line 1: nit: filename is a bit awkward because the type defined here is `ZeroTransfer`. Should we call this something like `tests` if this file was intended to hold all re-usable tests?

tests/load2/wallet.go
```diff
@@ -0,0 +1,101 @@
...
+		if err == nil {
+			if receipt.Status != types.ReceiptStatusSuccessful {
+				return fmt.Errorf("failed tx: %d", receipt.Status)
+			}
+			return nil
+		}
+
+		if err != ethereum.NotFound {
+			return err
+		}
```
joshua-kim: This will break with error wrapping - if this error is never wrapped it's fine, but otherwise I would go with using `errors.Is`

tests/load2/transfer.go
RodrigoVillar: Yes, these tests will eventually be used in contexts outside of load testing

tests/load2/transfer.go
maru-ava: Does there always need to be a direct correspondence between file name and the type(s) it defines?

tests/load2/generator.go
maru-ava: My fault for asking the name to be changed from Generator to LoadGenerator. tbh I don't feel all that strongly - I think this is bike shedding territory. Sometimes i'll choose the a letter other than the first for a single character receiver name.

tests/load2/wallet.go
RodrigoVillar: Done: 179467a

tests/load2/transfer.go
RodrigoVillar: Renamed test file: d7c4a9d

tests/load2/generator.go
RodrigoVillar: Done: e125729

tests/load2/generator.go
RodrigoVillar: Done: e125729

tests/load2/wallet.go
RodrigoVillar: Just like `ZeroTransfer`, `Wallet` is expected to be used in contexts outside of load testing. Having this be a `LoadGenerator`-level config prevents the `Wallet` from being used outside of load testing.

tests/load2/transfer.go
joshua-kim: > Does there always need to be a direct correspondence between file name and the type(s) it defines?

Not really... it just seemed like it was in a weird in-between of not being specific enough for a specific file name (if we're doing one file name per transfer test type), and not being vague enough if this was going to be a file that just held all of them

tests/load2/metrics.go
RodrigoVillar: Moved metrics into the `LoadGenerator` - the `LoadGenerator` now initializes the wallets + metrics, with the client having to pass in the workers it wants to use: 25bcb2e

tests/load2/wallet.go
joshua-kim: Do we have to bind these use-cases' wallets together? I'm trying to figure out if multi-chain load tests are something we want to do, or if it's just another potential configuration we can mess up. If we have some type which would be a struct over all of this sans chain id, the load generator can create the wallets with the same chain id.

tests/load2/transfer.go
RodrigoVillar: If we were to return the `tx`, this would no longer be a test but rather a constructor. For `load2` tests, we aimed for our tests to follow the design pattern set by the PR: 

https://github.com/ava-labs/avalanchego/pull/3554/files#diff-9ad649ff93aa380494097a550a8e2f9eb24f70414914c02eda2148ab728dbcf6

cc @maru-ava

tests/load2/wallet.go
RodrigoVillar: Ah I understand now - perfectly fine for the generator to setting the `chainID` for the wallets: 907654a

tests/load2/generator.go
RodrigoVillar: Even if we were to remove the `ctx` from `Run()`, we would still need to make a call to `context.WithTimeout()` for test timeouts (requested here: https://github.com/ava-labs/avalanchego/pull/3996#discussion_r2157475370)

The behavior of `loadTimeout` differs from `testTimeout` because they serve two different purposes:
- Load timeout: how long the long test should run for. The context timing out shouldn't error the tests because this is expected behavior ("I want to run the load test for X amount of time"). A load timeout of 0 sets the load test to run forever.
- Test timeout: how long a single test should run for. The context timing out should error the test because a test running past its timeout is unexpected behavior.

Elvis339 (approved):

tests/load2/generator.go
```diff
@@ -0,0 +1,99 @@
...
+	if loadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(ctx, loadTimeout)
+		childCtx = ctx
+		defer cancel()
+	}
+
+	for i := range l.wallets {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
joshua-kim: This pattern looks to be identical to an errgroup to me, should we use that instead here?

tests/load2/tests.go
```diff
@@ -0,0 +1,52 @@
...
+		GasTipCap: gasTipCap,
+		GasFeeCap: gasFeeCap,
+		Gas:       params.TxGas,
+		To:        &senderAddress,
+		Data:      nil,
+		Value:     common.Big0,
+	})
+	require.NoError(err)
+
+	require.NoError(wallet.sendTx(ctx, tx, z.PollFrequency))
```
joshua-kim: `sendTx` needs to be exported if this test is going to own sending, since otherwise we won't be able to have an implementation outside of this package implement the `Test` interface

RodrigoVillar: Done: d132d61

tests/load2/generator.go
RodrigoVillar: The function passed in `wg.Add()` does not produce any errors (errors are handled by `tc`) - `sync.WaitGroup` is equivalent to `errgroup.Group` in this context.

RodrigoVillar: Discussed offline - switched to `errgroup`: c4cfda7

tests/load2/generator.go
```diff
@@ -0,0 +1,96 @@
...
+		childCtx = ctx
+		defer cancel()
+	}
+
+	for i := range l.wallets {
+		eg.Go(func() error {
+			for {
+				select {
+				case <-childCtx.Done():
+					return nil
```
joshua-kim, line 83: Is this supposed to be the ctx error? I think if the context is canceled we drop the reason for cancelation here. If this is supposed to return the context's error then I think we'd also need to do a check on the result of `Wait`

joshua-kim (approved):

PR #3997, opened 2025-06-05
[tmpnet] Avoid port forwarding when running in a kube cluster

maru-ava:
## Why this should be merged

Previously port forwarding was always used to communicate with a node even when a test workload was running inside a kube cluster (assumed to be the same cluster hosting the avalanchego nodes). Now the pod IP will be used if running in a kube cluster (as indicated by a service account token is observed on the local filesystem). This avoids the overhead of tunneling through the kube API when not needed.

## How this was tested

CI

N/A


yacovm:
whoops, jumped the 🔫

PR #3998, opened 2025-06-05
Comm Component for Simplex

samliok:
## Why this should be merged

The `Comm` interface is used by simplex to send and broadcast messages. This PR builds off pr #3993 and should be merged after

## How this works

This PR adds two new fields to the Config struct—`Sender` and `OutboundMsgBuilder`—which enable the Simplex engine to construct and send outbound messages to other nodes.

## How this was tested

Using generated mocks for `ExternalSender` and `OutboundMessageBuilder`, I added tests in comm_test to verify that the correct messages are sent the expected number of times.

No


message/outbound_msg_builder.go
```diff
@@ -181,6 +181,8 @@ type OutboundMsgBuilder interface {
 		chainID ids.ID,
 		msg []byte,
 	) (OutboundMessage, error)
+
+	SimplexOutboundMessageBuilder
```
samliok: rather than bloating this file up, i thought it would be cleaner to separate the simplex outbound messages into their own file

simplex/test_util.go
```diff
@@ -37,9 +41,31 @@ func newEngineConfig() (*Config, error) {
...
+
 	return &Config{
 		Ctx:        simplexChainContext,
-		Validators: newTestValidatorInfo([]validators.GetValidatorOutput{nodeInfo}),
+		Validators: newTestValidators(append(validators, nodeInfo)),
 		SignBLS:    ls.Sign,
 	}, nil
 }
+
+func generateTestValidators(num uint64) ([]validators.GetValidatorOutput, error) {
```
yacovm: this is a test, so we can simplify the function by panicking or require.NoError-ing if we cannot initialize a new signer.

simplex/test_util.go
```diff
@@ -18,7 +18,11 @@ func newTestValidatorInfo(allVds []validators.GetValidatorOutput) map[ids.NodeID
 	return vds
 }
 
-func newEngineConfig() (*Config, error) {
+func newEngineConfig(numNodes uint64) (*Config, error) {
+	if numNodes == 0 {
+		numNodes = 1 // Ensure at least one node for testing
```
yacovm: I think it's odd to correct input passed to this function. Why not panic instead?

simplex/comm.go
```diff
@@ -0,0 +1,126 @@
...
+	})
+	return nodes
+}
+
+func (c *Comm) ListNodes() []simplex.NodeID {
+	return c.nodes
+}
+
+func (c *Comm) SendMessage(msg *simplex.Message, destination simplex.NodeID) {
+	var outboundMessage message.OutboundMessage
```
yacovm: can we extract out this huge switch into a function?

simplex/test_util.go
samliok: added testing.T, but had to rename the file to end in `_test` so it would pass linting.

yacovm (approved):

message/outbound_msg_builder.go
```diff
@@ -181,6 +181,8 @@ type OutboundMsgBuilder interface {
...
 }
 
 type outMsgBuilder struct {
```
yacovm, line 190: Unrelated to this PR, but I have to say I am not sure I understand why this is an un-exported struct. it Makes it impossible to use anywhere in the code (for tests) and we also have only one concrete implementation of it.

@StephenButtolph do you know why?

message/simplex_outbound_msg_builder.go
```diff
@@ -0,0 +1,347 @@
...
+		qc []byte,
+	) (OutboundMessage, error)
+
+	ReplicationRequest(
+		chainID ids.ID,
+		seqs []uint64,
+		latestRound uint64,
+	) (OutboundMessage, error)
+
+	VerifiedReplicationResponse(
```
yacovm: why is this called `VerifiedReplicationResponse`? the method should be according to the protobuf message we're sending, not the Simplex one.

simplex/comm.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/message"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/snow/networking/sender"
+	"github.com/ava-labs/avalanchego/subnets"
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+var errNodeNotFound = errors.New("node not found in the validator list")
```
yacovm: can we emphasize that it is our node that is not found?

simplex/comm.go
```diff
@@ -0,0 +1,123 @@
...
+
+func NewComm(config *Config) (*Comm, error) {
+	nodes := make([]simplex.NodeID, 0, len(config.Validators))
+
+	// grab all the nodes that are validators for the subnet
+	for _, vd := range config.Validators {
+		nodes = append(nodes, vd.NodeID[:])
+	}
+
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
```
yacovm, line 42: can we log the chain ID and subnet ID when it's not found?

simplex/comm.go
samliok: added to the error where it was being returned, so this err could potentially be reused

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+		msgBuilder: config.OutboundMsgBuilder,
+		chainID:    config.Ctx.ChainID,
+	}
+
+	return c, nil
+}
+
+func sortNodes(nodes []simplex.NodeID) []simplex.NodeID {
+	slices.SortFunc(nodes, func(i, j simplex.NodeID) int {
+		return strings.Compare(i.String(), j.String())
```
StephenButtolph: Is there a reason that we use the strings to sort the nodes? It seems like we should just sort the bytes as that will be faster.
```suggestion
		return bytes.Compare(i, j)
```

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+	return nodes
+}
```
StephenButtolph: Because the slice is sorted in-place, I think the std lib (and our sorting helpers) typically don't return the slice. Otherwise a user may think that the input slice is unmodified.
```suggestion
func sortNodes(nodes []simplex.NodeID) {
	slices.SortFunc(nodes, func(i, j simplex.NodeID) int {
		return strings.Compare(i.String(), j.String())
	})
}
```

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (c *Comm) SendMessage(msg *simplex.Message, destination simplex.NodeID) {
+	outboundMsg, err := c.simplexMessageToOutboundMessage(msg)
+	if err != nil {
+		c.logger.Error("Failed creating message", zap.Error(err))
+		return
+	}
+
+	dest := ids.NodeID(destination)
```
StephenButtolph: Casting from `[]byte` to `[20]byte` will panic if the provided slice doesn't have len==20. It seems like we should be using `ToNodeID` here.
```suggestion
	dest, err := ids.ToNodeID(destination)
```

message/outbound_msg_builder.go
StephenButtolph: We can utilize this in tests by using `message.NewCreator`.

message/simplex_outbound_msg_builder.go
```diff
@@ -0,0 +1,361 @@
...
+package message
+
+import (
+	"github.com/ava-labs/simplex"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/proto/pb/p2p"
+)
+
+type SimplexOutboundMessageBuilder interface {
```
StephenButtolph: I'm a bit confused by the asymmetry here.

We have already added support for inbound simplex messages, but the inbound logic just treats the simplex messages as opaque. Whereas the outbound messages here explicitly handle each possible form of the simplex messages.

Shouldn't the `OutboundMessageBuilder` just take in a `*p2p.Simplex` message to align with the inbound message handler?

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+
+func NewComm(config *Config) (*Comm, error) {
+	nodes := make([]simplex.NodeID, 0, len(config.Validators))
+
+	// grab all the nodes that are validators for the subnet
+	for _, vd := range config.Validators {
+		nodes = append(nodes, vd.NodeID[:])
+	}
+
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
```
StephenButtolph, line 42: What is the plan for non-validator nodes? Is this going to be modified in the future? Or is there going to be some other implementation entirely for non-validators?

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func sortNodes(nodes []simplex.NodeID) []simplex.NodeID {
+	slices.SortFunc(nodes, func(i, j simplex.NodeID) int {
+		return strings.Compare(i.String(), j.String())
+	})
+	return nodes
+}
+
+func (c *Comm) ListNodes() []simplex.NodeID {
```
StephenButtolph: nit: I feel like the return type makes it obvious that this is a list. Should this just be:
```suggestion
func (c *Comm) Nodes() []simplex.NodeID {
```
?

Or perhaps even
```suggestion
func (c *Comm) Validators() []simplex.NodeID {
```

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+			continue
+		}
+
+		c.SendMessage(msg, node)
+	}
+}
+
+func (c *Comm) simplexMessageToOutboundMessage(msg *simplex.Message) (message.OutboundMessage, error) {
+	var outboundMessage message.OutboundMessage
+	var err error
```
StephenButtolph: I don't think there is a good reason we assign these variables just to return them at the end of the function. Can we replace the assignment to these variables just with a `return` in all the cases?

This should catch things like:
- Using the wrong `err` on line 110.
- Returning `nil, nil` when there is not a valid message to construct (should return a non-nil error imo)

simplex/comm.go
```diff
@@ -0,0 +1,132 @@
...
+		return strings.Compare(i.String(), j.String())
+	})
+	return nodes
+}
+
+func (c *Comm) ListNodes() []simplex.NodeID {
+	return c.nodes
+}
+
+func (c *Comm) SendMessage(msg *simplex.Message, destination simplex.NodeID) {
```
StephenButtolph: nit:
```suggestion
func (c *Comm) Send(msg *simplex.Message, destination simplex.NodeID) {
```
`Message` feels redundant.

simplex/comm_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func TestCommSendMessage(t *testing.T) {
+	config := newEngineConfig(t, 1)
+
+	destinationNodeID := ids.GenerateTestNodeID()
+
+	ctrl := gomock.NewController(t)
+	msgCreator := messagemock.NewOutboundMsgBuilder(ctrl)
+	sender := sendermock.NewExternalSender(ctrl)
```
StephenButtolph: I generally hate mocks. While I don't think the usage of the `sender` mock is too problematic, I don't see why we can't use the real message builder here.

`message.NewCreator` should be sufficient no?

simplex/comm.go
yacovm: There is no concrete written plan yet, but I think this interface should not be used for non validators. 
I also don't think that the epoch should handle non validator block retrieval, and that we can handle it at a higher layer and have it decoupled from the logic of the epoch. 

Right now, we're wiring the epoch object into Simplex despite it not being the most high level component.

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+		simplexMsg = newP2PSimplexFinalization(c.chainID, msg.Finalization.Finalization.BlockHeader, msg.Finalization.QC.Bytes())
+	case msg.ReplicationRequest != nil:
+		simplexMsg = newP2PSimplexReplicationRequest(c.chainID, msg.ReplicationRequest.Seqs, msg.ReplicationRequest.LatestRound)
+	case msg.VerifiedReplicationResponse != nil:
+		msg, err := newP2PSimplexReplicationResponse(c.chainID, msg.VerifiedReplicationResponse.Data, msg.VerifiedReplicationResponse.LatestRound)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create replication response: %w", err)
+		}
+		simplexMsg = msg
+	}
```
StephenButtolph, line 135: I think we should add a `default` case to error on. It seems unexpected for us to be sending an empty message.

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+		return bytes.Compare(i, j)
+	})
+}
+
+func (c *Comm) ListNodes() []simplex.NodeID {
+	return c.nodes
+}
+
+func (c *Comm) SendMessage(msg *simplex.Message, destination simplex.NodeID) {
+	outboundMsg, err := c.simplexMessageToOutboundMessage(msg)
```
StephenButtolph, line 79: It feels pretty wasteful to create a new outbound message and call `Send` for each validator when we are `Broadcast`ing... Perhaps we could restructure the code so that we have a `send(msg *simplex.Message, nodeIDs set.Set[ids.ID])` helper which both `SendMessage` and `Broadcast` could use.

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type Comm struct {
+	logger   simplex.Logger
+	subnetID ids.ID
+	chainID  ids.ID
+	// nodeID is this nodes ID
+	nodeID simplex.NodeID
+	// nodes are the IDs of all the nodes in the subnet
+	nodes []simplex.NodeID
```
StephenButtolph: If we stored a copy of these as a `set.Set[ids.ID]` (without our own nodeID) then we wouldn't need to construct that set on every call to `Broadcast`.

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
+		config.Log.Warn("Node is not a validator for the subnet",
+			zap.String("nodeID", config.Ctx.NodeID.String()),
+			zap.String("chainID", config.Ctx.ChainID.String()),
+			zap.String("subnetID", config.Ctx.SubnetID.String()),
+		)
+		return nil, fmt.Errorf("%w could not find our node: %s", errNodeNotFound, config.Ctx.NodeID)
+	}
+
+	sortNodes(nodes)
```
StephenButtolph: Why are these sorted? Is that a requirement for Simplex? (if it is, can we document that?)

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+		nodes = append(nodes, vd.NodeID[:])
+	}
+
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
+		config.Log.Warn("Node is not a validator for the subnet",
+			zap.String("nodeID", config.Ctx.NodeID.String()),
+			zap.String("chainID", config.Ctx.ChainID.String()),
+			zap.String("subnetID", config.Ctx.SubnetID.String()),
+		)
+		return nil, fmt.Errorf("%w could not find our node: %s", errNodeNotFound, config.Ctx.NodeID)
```
StephenButtolph: "node not found in the validator list could not find our node: NodeID-xxx"

It reads a bit weirdly to me. Perhaps:
```suggestion
		return nil, fmt.Errorf("our %w: %s", errNodeNotFound, config.Ctx.NodeID)
```

"our node not found in the validator list: NodeID-xxx"

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// sortNodes sorts the nodes in place by their byte representations.
+func sortNodes(nodes []simplex.NodeID) {
+	slices.SortFunc(nodes, func(i, j simplex.NodeID) int {
+		return bytes.Compare(i, j)
+	})
+}
+
+func (c *Comm) ListNodes() []simplex.NodeID {
```
StephenButtolph: Could we document on the interface that we explicitly expect our own NodeID to be returned here? It feels odd that we would report our own nodeID for the communication layer but not send messages to ourself upon broadcast.

StephenButtolph: I also want to confirm that this is expected to always return all validators, not just the validators we are connected to right? I mainly ask because it seems weird to use the `Communication` interface as a place to expose what validators there are in the network, rather than just who we are currently connected to.

simplex/messages.go
```diff
@@ -0,0 +1,233 @@
...
+package simplex
+
+import (
+	"github.com/ava-labs/simplex"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/proto/pb/p2p"
+)
+
+func newP2PSimplexBlockProposal(
```
StephenButtolph: Everywhere that we have `newP2PSimplex...` should probably be replaced with `new`, we are in the `simplex` package, everything is simplex related.

simplex/messages.go
```diff
@@ -0,0 +1,233 @@
...
+		Message: &p2p.Simplex_ReplicationResponse{
+			ReplicationResponse: &p2p.ReplicationResponse{
+				Data:        qrs,
+				LatestRound: latestQR,
+			},
+		},
+	}, nil
+}
+
+func simplexBlockheaderToP2P(bh simplex.BlockHeader) *p2p.BlockHeader {
```
StephenButtolph:
```suggestion
func simplexBlockHeaderToP2P(bh simplex.BlockHeader) *p2p.BlockHeader {
```

simplex/comm_test.go
```diff
@@ -0,0 +1,106 @@
...
+	)
+	require.NoError(t, err)
+
+	config.OutboundMsgBuilder = mc
+	config.Sender = sender
+
+	comm, err := NewComm(config)
+	require.NoError(t, err)
+
+	sender.EXPECT().Send(gomock.Any(), gomock.Any(), comm.subnetID, gomock.Any())
```
StephenButtolph: Should we be verifying that the message we are sending (along with the recipient) are correct?

simplex/comm_test.go
```diff
@@ -0,0 +1,106 @@
...
+	)
+	require.NoError(t, err)
+
+	config.OutboundMsgBuilder = mc
+	config.Sender = sender
+
+	comm, err := NewComm(config)
+	require.NoError(t, err)
+
+	sender.EXPECT().Send(gomock.Any(), gomock.Any(), comm.subnetID, gomock.Any()).Times(2)
```
StephenButtolph: Should we be verifying that the message we are sending (along with the recipients) are correct?

simplex/comm_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	comm.Broadcast(&testSimplexMessage)
+}
+
+func TestCommFailsWithoutCurrentNode(t *testing.T) {
+	config := newEngineConfig(t, 3)
+
+	ctrl := gomock.NewController(t)
+	msgCreator := messagemock.NewOutboundMsgBuilder(ctrl)
```
StephenButtolph: Do we need to use a mock here?

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+		subnetID:   config.Ctx.SubnetID,
+		nodes:      nodes,
+		nodeID:     config.Ctx.NodeID[:],
+		logger:     config.Log,
+		sender:     config.Sender,
+		msgBuilder: config.OutboundMsgBuilder,
+		chainID:    config.Ctx.ChainID,
+	}
+
+	return c, nil
```
StephenButtolph: Optional nit:
```suggestion
	return &Comm{
		subnetID:   config.Ctx.SubnetID,
		nodes:      nodes,
		nodeID:     config.Ctx.NodeID[:],
		logger:     config.Log,
		sender:     config.Sender,
		msgBuilder: config.OutboundMsgBuilder,
		chainID:    config.Ctx.ChainID,
	}, nil
```
It just looks a tad cleaner to me

simplex/comm.go
```diff
@@ -0,0 +1,143 @@
...
+	// grab all the nodes that are validators for the subnet
+	for _, vd := range config.Validators {
+		nodes = append(nodes, vd.NodeID[:])
+	}
+
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
+		config.Log.Warn("Node is not a validator for the subnet",
+			zap.String("nodeID", config.Ctx.NodeID.String()),
+			zap.String("chainID", config.Ctx.ChainID.String()),
+			zap.String("subnetID", config.Ctx.SubnetID.String()),
```
StephenButtolph: Using a `Stringer` rather than `String` allows the `.String()` function to be evaluated lazily. While this probably isn't important for a `Warn` log, it is a critical performance boost with `Verbo` and `Debug` logs, so I try to be consistent everywhere.
```suggestion
			zap.Stringer("nodeID", config.Ctx.NodeID),
			zap.Stringer("chainID", config.Ctx.ChainID),
			zap.Stringer("subnetID", config.Ctx.SubnetID),
```

simplex/comm.go
samliok: Currently we require `Comm.Nodes()` to return a consistent order of nodes, for all nodes. This is how we [select the leader ](https://github.com/ava-labs/Simplex/blob/d33ae6ae82ea90e8cf0a0d15026694f5506931a4/epoch.go#L2688). We should definitely document this invariant.

simplex/comm.go
samliok: updated on simplex + avago side
https://github.com/ava-labs/Simplex/pull/181

simplex/comm.go
samliok: updated on simplex side as well
https://github.com/ava-labs/Simplex/pull/182#pullrequestreview-3021677751

simplex/comm.go
samliok: i think this would make things awkward with `comm.Nodes()` since it expects `[]simplex.NodeID` and also expects the current nodeID to be included.

simplex/comm.go
samliok: updated on simplex side as well
https://github.com/ava-labs/Simplex/pull/182#pullrequestreview-3021677751

simplex/comm.go
StephenButtolph: I was suggesting to store both on the struct

samliok: actually updated so that `Nodes` does this converting since we call `Nodes` much less frequently than `Broadcast`

simplex/comm.go
samliok: correct, we need to return all the validators that are part of the current epoch

go.mod
```diff
@@ -81,13 +81,14 @@ require (
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 )
 
+require github.com/ava-labs/firewood-go-ethhash/ffi v0.0.8 // indirect
```
StephenButtolph: Why is this moved to a different `require` group?

simplex/comm.go
```diff
@@ -0,0 +1,138 @@
...
+		config.Log.Warn("Node is not a validator for the subnet",
+			zap.Stringer("nodeID", config.Ctx.NodeID),
+			zap.Stringer("chainID", config.Ctx.ChainID),
+			zap.Stringer("subnetID", config.Ctx.SubnetID),
+		)
+		return nil, fmt.Errorf("our %w: %s", errNodeNotFound, config.Ctx.NodeID)
+	}
+
+	broadcastNodes := set.Set[ids.NodeID]{}
+	allNodes := make([]simplex.NodeID, 0, len(config.Validators)-1)
```
StephenButtolph:
```suggestion
	broadcastNodes := set.NewSet[ids.NodeID](len(config.Validators) - 1)
	allNodes := make([]simplex.NodeID, 0, len(config.Validators))
```

simplex/comm.go
```diff
@@ -0,0 +1,138 @@
...
+	case msg.ReplicationRequest != nil:
+		simplexMsg = newReplicationRequest(c.chainID, msg.ReplicationRequest)
+	case msg.VerifiedReplicationResponse != nil:
+		msg, err := newReplicationResponse(c.chainID, msg.VerifiedReplicationResponse)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create replication response: %w", err)
+		}
+		simplexMsg = msg
+	default:
+		return nil, fmt.Errorf("unknown message type: %T", msg)
```
StephenButtolph: This isn't a type switch, won't this always print `unknown message type: *simplex.Message`? Perhaps we should have:
```suggestion
		return nil, fmt.Errorf("unknown simplex message: %+v", msg)
```

This should output:
```
unknown simplex message: &{BlockMessage:<nil> VerifiedBlockMessage:<nil> EmptyNotarization:<nil> VoteMessage:<nil> EmptyVoteMessage:<nil> Notarization:<nil> FinalizeVote:<nil> Finalization:<nil> ReplicationResponse:<nil> VerifiedReplicationResponse:<nil> ReplicationRequest:<nil>}
```

go.mod
samliok: oh this is weird, i think i ran `go mod tidy` and it did this automatically. changed it back

PR #4001, opened 2025-06-09
Move HTTP2 routing information into headers

joshua-kim:
## Why this should be merged

Modifying the http2 `:path` is hacky because it's typically a read-only value, and supporting other tools like postman/grpcurl require more hacks where we have to keep injecting this value

## How this works

Moves the chain id information into an http header, which tooling generally has access to read/write to

## How this was tested

xsvm e2e test

No (this feature wasn't released yet)

### DIFF
--- a/api/server/http2_router.go
+++ b/api/server/http2_router.go
@@ -25,15 +25,9 @@ func newHTTP2Router() *http2Router {
 }
 
 func (h *http2Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// The :path pseudo-header takes the form of /Prefix/Path
-	parsed := strings.Split(r.URL.Path, "/")
-	if len(parsed) < 2 {
-		w.WriteHeader(http.StatusBadRequest)
-		return
-	}
-
-	chainID := parsed[1]
-
+	// the chain-id header must be set to route the request to the correct chain
+	// http2 handler
+	chainID := r.Header.Get("chain-id")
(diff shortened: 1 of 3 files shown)


api/server/http2_router.go
```diff
@@ -25,15 +25,9 @@ func newHTTP2Router() *http2Router {
...
-	if len(parsed) < 2 {
-		w.WriteHeader(http.StatusBadRequest)
-		return
-	}
-
-	chainID := parsed[1]
-
+	// the chain-id header must be set to route the request to the correct chain
+	// http2 handler
+	chainID := r.Header.Get("chain-id")
```
joshua-kim, line 29: Another idea is to return a different error code if this header is not set at all (like bad request) which might help with debugging

mpignatelli12 (approved):

api/server/http2_router.go
```diff
@@ -25,15 +25,14 @@ func newHTTP2Router() *http2Router {
...
+	if len(chainID) == 0 {
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	chainID := parsed[1]
-
```
StephenButtolph, line 36: Lower in this file, we still `r.Clone` the context and strip the `/chainID` prefix. This code isn't needed anymore right?

joshua-kim: Yeah it's not needed anymore, good catch.

PR #4003, opened 2025-06-09
fix: allow for load tests to run in-cluster

RodrigoVillar:
## Why this should be merged

Currently, attempting to run load tests inside a Kube cluster is not possible because of the following:

- We want to use the `inClusterConfig` when running tests inside a cluster: https://github.com/ava-labs/avalanchego/blob/3ee09b94399b34956ccef69025e4e741596b722f/tests/fixture/tmpnet/kube.go#L352
- The `if`-condition of this branch is never triggered because `tmpnet` requires us to specify a file path for the kube config https://github.com/ava-labs/avalanchego/blob/3ee09b94399b34956ccef69025e4e741596b722f/tests/fixture/tmpnet/flags/kubeconfig.go#L48-L56

## How this works

This PR fixes this by defaulting the `kubeconfig` flag to `""`. Furthermore, in `tmpnetctl`, starting a kind cluster without specifying a kube config will default to `$HOME/.kube/config`

## How this was tested

Ran a load test inside a kind cluster.

N/A

### DIFF
--- a/tests/fixture/tmpnet/flags/kubeconfig.go
+++ b/tests/fixture/tmpnet/flags/kubeconfig.go
@@ -44,7 +44,6 @@ func newKubeconfigFlagSetVars(flagSet *pflag.FlagSet, docPrefix string) *Kubecon
 	return v
 }
 
-// internal method registering kubeConfig and kubeconfig context
 func (v *KubeconfigVars) register(stringVar varFunc[string], docPrefix string) {
 	// the default kubeConfig path is set to empty to allow for the use of a projected
 	// token when running in-cluster


tests/fixture/tmpnet/flags/kubeconfig.go
```diff
@@ -48,7 +47,7 @@ func (v *KubeconfigVars) register(stringVar varFunc[string], docPrefix string) {
 	stringVar(
 		&v.Path,
 		"kubeconfig",
-		tmpnet.GetEnvWithDefault(KubeconfigPathEnvVar, os.ExpandEnv("$HOME/.kube/config")),
```
maru-ava, line 51: Maybe update the default value to depend on the value of tmpnet.isRunningInCluster (will need to export) to ensure the previous value outside the cluster and an empty string inside the cluster? That should fix the CI failures.

RodrigoVillar: Done: 6069b0b

tests/fixture/tmpnet/tmpnetctl/main.go
```diff
@@ -283,7 +280,7 @@ func main() {
 			// A valid kubeconfig is required for local kind usage but this is not validated by KubeconfigVars
 			// since unlike kind, tmpnet usage may involve an implicit in-cluster config.
 			if len(kubeconfigVars.Path) == 0 {
-				return errKubeconfigRequired
```
maru-ava: Is this change still required?

RodrigoVillar: Great point - diff reverted 6e4c9aa

tests/fixture/tmpnet/flags/kubeconfig.go
```diff
@@ -45,10 +45,15 @@ func newKubeconfigFlagSetVars(flagSet *pflag.FlagSet, docPrefix string) *Kubecon
 }
 
 func (v *KubeconfigVars) register(stringVar varFunc[string], docPrefix string) {
+	var defaultKubeConfigPath string
```
maru-ava, line 50: Apologies for the piecemeal comments, for some reason my brain is running serial on this PR...

One more thing - please add a comment indicating the rational for varying the kubeconfig default (i.e. use of a projected token when running in-cluster requires an empty path).

RodrigoVillar: Done: 63bbb62

maru-ava: I don't think this needs to be a docstring, just comment setting the value of defaultKubeConfigPath

RodrigoVillar: 63bbb62

tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -842,10 +842,10 @@ func configureExclusiveScheduling(template *corev1.PodTemplateSpec, labelKey str
 	}
 }
 
-// isRunningInCluster detects if this code is running inside a Kubernetes cluster
+// IsRunningInCluster detects if this code is running inside a Kubernetes cluster
 // by checking for the presence of the service account token that's automatically
 // mounted in every pod.
-func isRunningInCluster() bool {
+func IsRunningInCluster() bool {
 	_, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token")
```
Elvis339, line 849: This is not part of the change but I think `IsRunningInCluster` should be updated to use environment variables instead of checking for the service account token file:

```go
func IsRunningInCluster() bool {
    return os.Getenv("KUBERNETES_SERVICE_HOST") != "" && 
           os.Getenv("KUBERNETES_SERVICE_PORT") != ""
}
```

It's more reliable because the current approach fails if I were to set custom service account mount path.

The `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables are always set by kubelet in every pod.

Elvis339 (approved):

tests/fixture/tmpnet/kube_runtime.go
maru-ava: What would be the use case justifying a custom mount path?

Elvis339: Multiple token types where different services require different token audiences.

maru-ava: Sure, that's what would require a custom mount path. A use case, though, would be a description of a requirement we actually have.

PR #4004, opened 2025-06-09
Simplex QuorumCertificate and BLS aggregator

samliok:
## Why this should be merged

Implements the simplex `QuorumCertificate`, `QCDeserializer` and `SignatureAggregator` interfaces. This allows simplex to parse, aggregate and verify quorum certificates(ex. finalizations and notarizations) during execution. 

## How this works

- Builds on top of the BLSVerifier to handle BLS signatures and public keys. 
- The bytes of a QC are serialized with `asn1`

## How this was tested

Added unit tests to `bls_test.go`.

no

### DIFF
--- a/simplex/bls_test.go
+++ b/simplex/bls_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/simplex"
 )
 
 func TestBLSVerifier(t *testing.T) {
--- a/simplex/quorum.go
+++ b/simplex/quorum.go
@@ -46,7 +46,7 @@ func (qc *QC) Verify(msg []byte) error {
 			return fmt.Errorf("%w: %x", errSignerNotFound, signer)
 		}
 
-		pks = append(pks, &pk)
+		pks = append(pks, pk)
 	}
 
 	// aggregate the public keys
@@ -55,7 +55,7 @@ func (qc *QC) Verify(msg []byte) error {
 		return fmt.Errorf("%w: %w", errSignatureAggregation, err)
 	}
 
-	message2Verify, err := encodeMessageToSign(msg, qc.verifier.chainID, qc.verifier.subnetID)
+	message2Verify, err := encodeMessageToSign(msg, qc.verifier.chainID, qc.verifier.networkID)
 	if err != nil {
 		return fmt.Errorf("%w: %w", errEncodingMessageToSign, err)
 	}
--- a/simplex/util_test.go
+++ b/simplex/util_test.go
@@ -36,42 +36,42 @@ func newTestValidators(allVds []*testSigner) map[ids.NodeID]*validators.GetValid
 	return vds
 }
 
-func newEngineConfig(t *testing.T, numNodes uint64) *testConfig {
-	if numNodes == 0 {
-		panic("numNodes must be greater than 0")
-	}
+// func newEngineConfig(t *testing.T, numNodes uint64) *testConfig {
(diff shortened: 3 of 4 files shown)


simplex/quorum.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+)
+
+var (
+	_ simplex.QuorumCertificate   = (*QC)(nil)
+	_ simplex.QCDeserializer      = QCDeserializer{}
+	_ simplex.SignatureAggregator = (*SignatureAggregator)(nil)
+)
+
+type QCDeserializer BLSVerifier
```
yacovm: can we define this closer to `DeserializeQuorumCertificate` ?

simplex/quorum.go
```diff
@@ -0,0 +1,152 @@
...
+// Bytes serializes the quorum certificate into bytes.
+// The serialization format is:
+// [signer1][signer2]...[signerN][signature]
+// where each signer is represented by its NodeID and the signature is the BLS signature.
+func (qc *QC) Bytes() []byte {
+	sigBytes := bls.SignatureToBytes(&qc.sig)
+	buff := make([]byte, len(sigBytes)+len(qc.signers)*ids.NodeIDLen)
+	var pos int
+	for _, signer := range qc.signers {
+		copy(buff[pos:], signer[:ids.NodeIDLen])
```
yacovm: I know this is my fault but I don't like this hacky serialization scheme.

Can we use something implicit ? We have the `QC` struct, we can just add fields and annotations to it and have it serialized / deserialized with ASN1 or Canoto or protobuf.

simplex/quorum.go
```diff
@@ -0,0 +1,152 @@
...
+		pos += ids.NodeIDLen
+	}
+
+	copy(buff[pos:], sigBytes)
+	return buff
+}
+
+// DeserializeQuorumCertificate deserializes a quorum certificate from bytes.
+func (d QCDeserializer) DeserializeQuorumCertificate(bytes []byte) (simplex.QuorumCertificate, error) {
+	quorumSize := simplex.Quorum(len(d.nodeID2PK))
```
yacovm: can we make sure that we have exactly `quorumSize` signers either here or in verify?

samliok: i think `expectedSize` should verify we have exactly a quorum of signers, but i added a check in `verify` anyways.

simplex/util_test.go
```diff
@@ -0,0 +1,92 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
yacovm, line 1: can we try renaming via `git mv` ?

simplex/bls_test.go
```diff
@@ -53,7 +51,210 @@ func TestSignerInvalidMessageEncoding(t *testing.T) {
...
+	signer2, verifier2 := NewBLSAuth(&config.Config)
+	sig2, err := signer2.Sign(msg)
+	require.NoError(t, err)
+	require.NoError(t, verifier2.Verify(msg, sig2, node2ID.NodeID[:]))
+
+	// aggregate the signatures into a quorum certificate
+	signatureAggregator := SignatureAggregator(verifier)
+	qc, err := signatureAggregator.Aggregate(
+		[]simplex.Signature{
+			{Signer: config.Ctx.NodeID[:], Value: sig},
```
yacovm: I'm sorry but I cannot grok this:

```
			{Signer: config.Ctx.NodeID[:], Value: sig},
			{Signer: node2ID.NodeID[:], Value: sig2},
```

Can we make it more uniform?

yacovm (commented):
generally LGTM, but can we use `git mv` ?

PR #4007, opened 2025-06-10
feat: add task for in-cluster load test

RodrigoVillar:
## Why this should be merged

This PR adds a task allowing for running load tests within a local kind cluster. 

## How this works

The task does the following:
- Start a kind cluster if one has not been started yet
- Create an image that runs the load test and deploys it to a local registry
- Create a pod manifest for the load test and deploys the pod to the kind cluster 

## How this was tested

The load test was run with `task test-load-kind-cluster` and passed.

N/A

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -232,10 +232,10 @@ tasks:
       - task: generate-load-contract-bindings
       - task: build
       - cmd: go run ./tests/load/c/main --avalanchego-path=./build/avalanchego {{.CLI_ARGS}}
-    
+  
   test-load-kube:
     desc: Runs load tests against a network deployed to kube
-    cmds: 
+    cmds:
       - task: generate-load-contract-bindings
       - cmd: bash -x ./scripts/tests.load.kube.sh {{.CLI_ARGS}}
 
@@ -243,6 +243,10 @@ tasks:
     desc: Runs load tests against kube with exclusive scheduling
     cmds:
       - cmd: go run ./tests/load/c/main --runtime=kube --kube-use-exclusive-scheduling {{.CLI_ARGS}}
+
+  test-load-kind-cluster:
+    desc: Runs load tests within a kind cluster
+    cmd: bash -x ./scripts/tests.load.kind.sh
 
   test-unit:
     desc: Runs unit tests
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -232,7 +232,7 @@ tasks:
       - task: generate-load-contract-bindings
       - task: build
       - cmd: go run ./tests/load/c/main --avalanchego-path=./build/avalanchego {{.CLI_ARGS}}
-  
+
   test-load-kube:
     desc: Runs load tests against a network deployed to kube
     cmds:
@@ -244,7 +244,7 @@ tasks:
     cmds:
       - cmd: go run ./tests/load/c/main --runtime=kube --kube-use-exclusive-scheduling {{.CLI_ARGS}}
 
-  test-load-kind-cluster:
+  test-load-within-cluster:
     desc: Runs load tests within a kind cluster
     cmd: bash -x ./scripts/tests.load.kind.sh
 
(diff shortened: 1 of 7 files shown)


Taskfile.yml
```diff
@@ -232,10 +232,10 @@ tasks:
       - task: generate-load-contract-bindings
       - task: build
       - cmd: go run ./tests/load/c/main --avalanchego-path=./build/avalanchego {{.CLI_ARGS}}
-    
+  
   test-load-kube:
     desc: Runs load tests against a network deployed to kube
-    cmds:
```
RodrigoVillar, line 238: I think my Neovim formatter caught this - unrelated change but I don't think we should be having trailing spaces in `.yml` files just like we don't have it for `.go` files

Might be worthwhile to add a linter for `.yml` files at some point

maru-ava: Likely a function of Quentin's PR, since my editor is similarly configure to strip trailing newlines on save. Not sure that we need a linter for this though it couldn't hurt.

Taskfile.yml
```diff
@@ -244,6 +244,10 @@ tasks:
     cmds:
       - cmd: go run ./tests/load/c/main --runtime=kube --kube-use-exclusive-scheduling {{.CLI_ARGS}}
 
+  test-load-kind-cluster:
```
maru-ava: Maybe choose a name more reflective of the test running inside the cluster? 'kind cluster' isn't very specific.

scripts/build_image.sh
```diff
@@ -10,6 +10,7 @@ set -euo pipefail
 # DOCKER_IMAGE=avaplatform/avalanchego ./scripts/build_image.sh                       # Build and push multi-arch image to docker hub
 # DOCKER_IMAGE=localhost:5001/avalanchego ./scripts/build_image.sh                    # Build and push multi-arch image to private registry
 # DOCKER_IMAGE=localhost:5001/avalanchego FORCE_TAG_LATEST=1 ./scripts/build_image.sh # Build and push image to private registry with tag `latest`
+# DOCKERFILE="./Dockerfile" ./scripts/build_image.sh                                    # Build image with a custom Dockerfile
```
maru-ava, line 13: (No action required) Why is it desirable to customize this build script instead of following the example of [scripts/build_bootstrap_monitor_image.sh](https://github.com/ava-labs/avalanchego/blob/master/scripts/build_bootstrap_monitor_image.sh)?

Note that a compiled binary is suggested rather than using 'go run' at runtime.

scripts/tests.load.kind.sh
```diff
@@ -0,0 +1,81 @@
...
+
+set -euo pipefail
+
+if ! [[ "$0" =~ scripts/tests.load.kind.sh ]]; then
+    echo "must be run from repository root"
+    exit 255
+fi
+
+# Start kind cluster
+./scripts/start_kind_cluster.sh
```
maru-ava: Suggest passing arguments as per the example of other kind-using scripts.

scripts/tests.load.kind.sh
```diff
@@ -0,0 +1,81 @@
...
+  name: load-test
+  namespace: tmpnet
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: load-test
+  namespace: tmpnet
+rules:
```
maru-ava, line 36: (No action required) How did you arrive at these permissions?

tests/load/prometheus.go
```diff
@@ -101,5 +101,9 @@ func (s *MetricsServer) GenerateMonitoringConfig(monitoringLabels map[string]str
 		return "", err
 	}
 
+	if err := os.MkdirAll(filepath.Dir(collectorFilePath), 0o755); err != nil {
```
maru-ava, line 104: While this addition might avoid an error, the fact that the path does not exist is a symptom of a larger problem: collection is not being configured in the pod. Given the requirement to label test workload metrics with network uuid, which isn't known at the time of pod deployment, I think deployment of local prometheus collector would be suggested so that tmpnet configure it. That would mean setting the collector credentials to the pod - easy enough - but also ensuring the availability of a compatible version of prometheus so that tmpnet could start it.

Maybe coordinate with Elvis to see what the timeline is for getting ARC online? Other than as a learning exercise, I'm less convinced of the wisdom of supporting pod-based workloads if it requires not just publishing an image and that image being complex to build. CI-launched tests won't need to publish images, and don't need extra work to support workload monitoring. Local iteration would likely be easier to support via enabling external access to nodes via a proxy instead of forwarding.

scripts/tests.load.kind.sh
RodrigoVillar: Done: 387e4ef

Taskfile.yml
RodrigoVillar: Done: d9fdb87

PR #4009, opened 2025-06-11
Simplex VerifiedBlock and BlockDeserializer

samliok:
## Why this should be merged

This introduces the `VerifiedBlock` and `BlockDeserializer` components of simplex. VerifiedBlocks are blocks that simplex has as already called `verify()` on. This ensures verify is only called once. 

No


simplex/block.go
```diff
@@ -0,0 +1,88 @@
...
+		ProtocolMetadata: v.metadata,
+		Digest:           v.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block.
+// It concatenates the metadata bytes followed by the inner block bytes.
+func (v *VerifiedBlock) Bytes() []byte {
+	mdBytes := v.metadata.Bytes()
+	buff := make([]byte, len(mdBytes)+len(v.innerBlock))
```
yacovm: Can we serialize this in a protobuf or canoto? I think this serialization should be extendible and not rigid as we have it here. If we want to add a new field in the future, serializing it this way will make it impossible to add a field.

simplex/block.go
```diff
@@ -0,0 +1,88 @@
...
+	copy(buff, mdBytes)
+	copy(buff[len(mdBytes):], v.innerBlock)
+	return buff
+}
+
+// computeDigest computes the digest of the block.
+func (v *VerifiedBlock) computeDigest() {
+	mdBytes := v.metadata.Bytes()
```
yacovm: I don't like the way we hash things by concatenating them without clear separation. 

We should probably hash the ASN1 / codec of the struct containing the inner block and the metadata and not just concatenate them and then hash them, despite the fact that the metadata is in the first place.

simplex/block.go
```diff
@@ -0,0 +1,88 @@
...
+	h := sha256.New()
+	h.Write(mdBytes)
+	h.Write(v.innerBlock)
+	digest := h.Sum(nil)
+	v.digest = simplex.Digest(digest)
+}
+
+type blockDeserializer struct {
+	vm block.ChainVM
```
yacovm: I don't think we need to drag the entire ChainVM interface here just to parse the block. Can't we just define the interface we need?

simplex/block.go
```diff
@@ -0,0 +1,97 @@
...
+	return buff
+}
+
+// computeDigest computes the digest of the block.
+func (v *VerifiedBlock) computeDigest() {
+	h := sha256.New()
+	h.Write(v.Bytes())
+	digest := h.Sum(nil)
+	v.digest = simplex.Digest(digest)
+}
```
StephenButtolph: We should not be relying on the hash of a proto message to be consistent across calls. If we want a proto message with a canonical hash, we should use canoto. Otherwise we should be using some other serialization format.

simplex/block.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return buff
+}
+
+// computeDigest computes the digest of the block.
+func (v *VerifiedBlock) computeDigest() {
+	h := sha256.New()
+	h.Write(v.Bytes())
+	digest := h.Sum(nil)
+	v.digest = simplex.Digest(digest)
```
StephenButtolph: This performs an unnecessary memory allocation here. Simplest fix would be:
```suggestion
	h.Sum(v.digest[:0])
```

Alternatively we could just do:
```golang
	v.digest = sha256.Sum256(v.Bytes())
```

or

```golang
	v.digest = hashing.ComputeHash256Array(v.Bytes())
```

simplex/block.go
```diff
@@ -0,0 +1,97 @@
...
+func (v *VerifiedBlock) BlockHeader() simplex.BlockHeader {
+	v.computeDigestOnce.Do(v.computeDigest)
+	return simplex.BlockHeader{
+		ProtocolMetadata: v.metadata,
+		Digest:           v.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block
+// as the asn1 encoding of `encodedVerifiedBlock`.
```
StephenButtolph: This doesn't use `asn1`

simplex/block.go
```diff
@@ -0,0 +1,97 @@
...
+func (b *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.VerifiedBlock, error) {
+	vb, err := verifiedBlockFromBytes(bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to deserialize verified block: %w", err)
+	}
+
+	_, err = b.parser.ParseBlock(context.Background(), vb.innerBlock)
+	if err != nil {
+		return nil, err
+	}
```
StephenButtolph: Is it expected for us not to keep the parsed block? I would have thought that parsing the block would actually keep the parsed block around.

simplex/block.go
```diff
@@ -0,0 +1,97 @@
...
+// as the asn1 encoding of `encodedVerifiedBlock`.
+func (v *VerifiedBlock) Bytes() []byte {
+	cBlock := pSimplex.VerifiedBlock{
+		Metadata: v.metadata.Bytes(),
+		Block:    v.innerBlock,
+	}
+
+	buff, err := proto.Marshal(&cBlock)
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal verified block: %w", err))
```
StephenButtolph: I generally try to avoid panicking outside of `init` functions... There are a handful of instances where we do panic in avalanchego... But returning an error is almost always preferred to reduce the chances of introducing a panic bug.

Admittedly here I doubt there is a possibility of panicking... But I don't know all the invariants around `proto.Marshal`...

StephenButtolph:
I'm also not super clear on the `VerifiedBlock` naming. It doesn't make sense to me personally to ever parse a "verified" block. If the block is verified, it should be in memory. If the block needs to be parsed, then it must not be in memory.

Talking offline, it seems like "VerifiedBlock" is supposed to imply that a correct _peer_ would have verified the block... But that doesn't seem relevant imo

yacovm:
> Talking offline, it seems like "VerifiedBlock" is supposed to imply that a correct _peer_ would have verified the block... But that doesn't seem relevant imo

Nope, that means that we (this peer) have verified the block.

yacovm:
> I'm also not super clear on the `VerifiedBlock` naming. It doesn't make sense to me personally to ever parse a "verified" block. If the block is verified, it should be in memory. If the block needs to be parsed, then it must not be in memory.

The only place where we parse the bytes of a verified block is when we recover from crash and load it into memory.

simplex/block.go
yacovm: We would need to keep a reference to it by [wiring an accept() functionality](https://github.com/yacovm/avalanchego/blob/simplexIntegration/simplex/block.go#L63). A VerifiedBlock that is deserialized is picked up from the WAL and may be accepted in the future.

yacovm: But I think that in this PR we can leave it like that because the accept functionality isn't there yet.

simplex/block.go
yacovm: I agree, although (just as an anecdote) I have actually tested several versions of Golang protobuf in the past (with simple schemas of byte concatenations) and they were consistent.

However there is indeed no official guarantee that this will be consistent.

yacovm: Also, let's not use ASN1 here because it's slow with large messages.

samliok:
> > Talking offline, it seems like "VerifiedBlock" is supposed to imply that a correct _peer_ would have verified the block... But that doesn't seem relevant imo
> 
> Nope, that means that we (this peer) have verified the block.

So offline, I was referring to the `FromBytes` method of `VerifiedBlock`, which we use to parse an unverified block when we receive it(ex. via bytes from a BlockProposal). Therefore, we do create a `VerifiedBlock` even though we haven't actually verified it yet.

yacovm:
> > > Talking offline, it seems like "VerifiedBlock" is supposed to imply that a correct _peer_ would have verified the block... But that doesn't seem relevant imo
> > 
> > 
> > Nope, that means that we (this peer) have verified the block.
> 
> So offline, I was referring to the `FromBytes` method of `VerifiedBlock`, which we use to parse an unverified block when we receive it(ex. via bytes from a BlockProposal). Therefore, we do create a `VerifiedBlock` even though we haven't actually verified it yet.

It doesn't matter what we can or cannot do in the avalanchego side. `VerifiedBlock` is a Simplex concept and it's an interface:

```
type VerifiedBlock interface {
	// BlockHeader encodes a succinct and collision-free representation of a block.
	BlockHeader() BlockHeader

	// Bytes returns a byte encoding of the block
	Bytes() []byte
}
```

The fact that we have an implementation of it in avalanchego that enables someone to create an instance of it, is irrelevant, as the abstraction is a Simplex one, and in Simplex you cannot create a verified block without you verifying it before.

StephenButtolph:
> The fact that we have an implementation of it in avalanchego that enables someone to create an instance of it, is irrelevant, as the abstraction is a Simplex one, and in Simplex you cannot create a verified block without you verifying it before.

It seems to me that we should never be creating a verified block from bytes. Because of this, I don't think we should have any such support for it in avalanchego.

yacovm:
> > The fact that we have an implementation of it in avalanchego that enables someone to create an instance of it, is irrelevant, as the abstraction is a Simplex one, and in Simplex you cannot create a verified block without you verifying it before.
> 
> It seems to me that we should never be creating a verified block from bytes. Because of this, I don't think we should have any such support for it in avalanchego.

OK then in that case we should just re-verify the blocks we have in the WAL when we restart.

simplex/block.go
```diff
@@ -0,0 +1,80 @@
...
+func (v *VerifiedBlock) BlockHeader() simplex.BlockHeader {
+	return simplex.BlockHeader{
+		ProtocolMetadata: v.metadata,
+		Digest:           v.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block
+// as the proto encoding of `encodedVerifiedBlock`.
+// TODO: we should use a canonical encoding since proto is not guaranteed to be canonical.
```
yacovm: can we remove this TODO now?

simplex/block.go
```diff
@@ -0,0 +1,80 @@
...
+// BlockHeader returns the block header for the verified block.
+func (v *VerifiedBlock) BlockHeader() simplex.BlockHeader {
+	return simplex.BlockHeader{
+		ProtocolMetadata: v.metadata,
+		Digest:           v.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block
+// as the proto encoding of `encodedVerifiedBlock`.
```
yacovm: I don't think we should comment about how we're serializing the VB. That's up to the VB to decide.

simplex/block.go
```diff
@@ -0,0 +1,80 @@
...
+func (v *VerifiedBlock) computeDigest() (simplex.Digest, error) {
+	bytes, err := v.Bytes()
+	if err != nil {
+		return simplex.Digest{}, fmt.Errorf("failed to serialize verified block: %w", err)
+	}
+
+	return hashing.ComputeHash256Array(bytes), nil
+}
+
+func verifiedBlockFromBytes(buff []byte) (*VerifiedBlock, error) {
```
yacovm: why do we have a `VerifiedBlock` from bytes if our deserializer now returns only non verified blocks?

```
type BlockDeserializer interface {
	// DeserializeBlock parses the given bytes and initializes a VerifiedBlock.
	// Returns an error upon failure.
	DeserializeBlock(bytes []byte) (Block, error)
}
```

simplex/block_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Serialize the block
+	blockBytes, err := vb.Bytes()
+	require.NoError(t, err)
+
+	// Deserialize the block
+	vb1, err := verifiedBlockFromBytes(blockBytes)
+	require.NoError(t, err)
+
+	// Check that the inner block data is preserved
```
yacovm: comment doesn't match the code

simplex/block_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"testing"
+
+	"github.com/ava-labs/simplex"
+	"github.com/stretchr/testify/require"
+)
+
+// TestVerifiedBlock tests that a block can be serialized and deserialized correctly,
+// and that the digest is computed correctly after deserialization.
```
yacovm: > // and that the digest is computed correctly after deserialization.

Where do we check this?

simplex/config.go
```diff
@@ -21,6 +22,8 @@ type Config struct {
 
 	// SignBLS is the signing function used for this node to sign messages.
 	SignBLS SignFunc
+
+	Parser block.Parser
```
yacovm: what is this parser for?

simplex/canoto.go
```diff
@@ -0,0 +1,14 @@
...
+package simplex
+
+//go:generate go run github.com/StephenButtolph/canoto/canoto $GOFILE
+
+// CanotoVerifiedBlock is the Canoto representation of a verified block
+type CanotoVerifiedBlock struct {
+	Metadata   []byte `canoto:"bytes,1"`
+	InnerBlock []byte `canoto:"bytes,2"`
+
+	canotoData canotoData_CanotoVerifiedBlock
```
yacovm: why do we need this `canotoData_CanotoVerifiedBlock` ?

simplex/config.go
samliok: for future prs, deleting for now

simplex/canoto.go
samliok: From the [canoto](https://github.com/StephenButtolph/canoto) repo
>All structs must include a field called canotoData that will cache the results of calculating the size of the struct.

simplex/block.go
samliok: In future PRs we use this to construct the `VerifiedBlock` field inside our `Block` struct. It's also nice to have the to(`Bytes()`) and from(`verifiedBlockFromBytes`) methods process the same struct.  

But I can change the function to have something like this signature if you'd prefer
```go
func blockFieldsFromBytes(buff []byte) (simplex.ProtocolMetadata, []byte, error) {
```

yacovm: I had the nested structure you describe in my PoC branch because it was faster to code it like that, I am not sure we really want something like that, but I guess we can decide later.

Let's keep it as it is and worst case we'll change to what you suggested here later.

simplex/block.go
```diff
@@ -0,0 +1,78 @@
...
+// BlockHeader returns the block header for the verified block.
+func (v *VerifiedBlock) BlockHeader() simplex.BlockHeader {
+	return simplex.BlockHeader{
+		ProtocolMetadata: v.metadata,
+		Digest:           v.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block.
+func (v *VerifiedBlock) Bytes() ([]byte, error) {
```
StephenButtolph: Why does this function return an error?

simplex/block.go
```diff
@@ -0,0 +1,78 @@
...
+	cBlock := &CanotoVerifiedBlock{
+		Metadata:   v.metadata.Bytes(),
```
StephenButtolph: Would it make sense for `metadata` to be a canoto struct? If we did that then we could get rid of the indirection between `VerifiedBlock` and `CanotoVerifiedBlock`.

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+
+	buff := cBlock.MarshalCanoto()
+
+	return buff, nil
+}
+
+func (b *Block) Verify(ctx context.Context) (simplex.VerifiedBlock, error) {
+	if b.verified {
+		return b, errBlockAlreadyVerified
+	}
```
StephenButtolph: Does it make sense to track the block verification here? Or should that be done as part of the wider block tracker logic (encompassed in the TODO below)?

I ask because it isn't clear to mean that this boolean is sufficient to guarantee we don't call `Verify` multiple times on the same block.

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// TODO: track blocks that have been verified to ensure they are either rejected or accepted
+	b.verified = true
+	err := b.vmBlock.Verify(ctx)
```
StephenButtolph: Is there any verification done to ensure that the `Prev` referenced in the metadata references the same `vmBlock` as `b.vmBlock.Parent()`?

It seems like that type of verification needs to be done here, no?

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+
+type blockDeserializer struct {
+	parser block.Parser
+}
+
+func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error) {
+	return d.deserializeBlockBytes(bytes)
+}
+
+func (d *blockDeserializer) deserializeBlockBytes(buff []byte) (simplex.Block, error) {
```
StephenButtolph: Why do we have this unexported helper? I think we can just get rid of it:
```suggestion
```

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+	var canotoBlock CanotoSimplexBlock
+
+	if err := canotoBlock.UnmarshalCanoto(buff); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal verified block: %w", err)
```
StephenButtolph:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal block: %w", err)
```

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/ava-labs/avalanchego/utils/hashing"
+)
+
+var _ simplex.BlockDeserializer = (*blockDeserializer)(nil)
+
+var _ simplex.Block = (*Block)(nil)
+
+var _ simplex.VerifiedBlock = (*Block)(nil)
+
+var errBlockAlreadyVerified = errors.New("block has already been verified")
```
StephenButtolph: We typically use just a single var block rather than multiple when we are defining a group of things.
```suggestion
var (
	_ simplex.BlockDeserializer = (*blockDeserializer)(nil)
	_ simplex.Block             = (*Block)(nil)
	_ simplex.VerifiedBlock     = (*Block)(nil)

	errBlockAlreadyVerified = errors.New("block has already been verified")
)
```

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+// Bytes returns the serialized bytes of the verified block.
+func (b *Block) Bytes() ([]byte, error) {
+	cBlock := &CanotoSimplexBlock{
+		Metadata:   b.metadata.Bytes(),
+		InnerBlock: b.vmBlock.Bytes(),
+	}
+
+	buff := cBlock.MarshalCanoto()
+
+	return buff, nil
```
StephenButtolph: Not a meaningful diff, but it seems like we can probably skip the declaration of `buff`.
```suggestion
	return cBlock.MarshalCanoto(), nil
```

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+
+// BlockHeader returns the block header for the verified block.
+func (b *Block) BlockHeader() simplex.BlockHeader {
+	return simplex.BlockHeader{
+		ProtocolMetadata: b.metadata,
+		Digest:           b.digest,
+	}
+}
+
+// Bytes returns the serialized bytes of the verified block.
```
StephenButtolph:
```suggestion
// Bytes returns the serialized bytes of the block.
```

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+	// metadata contains protocol metadata for the block
+	metadata simplex.ProtocolMetadata
+
+	verified bool // whether vmBlock.verify() function has been called
+
+	// the parsed block
+	vmBlock snowman.Block
+}
+
+// BlockHeader returns the block header for the verified block.
```
StephenButtolph:
```suggestion
// BlockHeader returns the block header for the block.
```

simplex/block.go
StephenButtolph: I presume all the additional things that need to be verified in the protocol metadata are handled elsewhere in simplex

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := canotoBlock.UnmarshalCanoto(buff); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal verified block: %w", err)
+	}
+
+	md, err := simplex.ProtocolMetadataFromBytes(canotoBlock.Metadata)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse protocol metadata: %w", err)
+	}
+
+	vmblock, err := d.parser.ParseBlock(context.Background(), canotoBlock.InnerBlock)
```
StephenButtolph: It seems like we should pass a context into `DeserializeBlock`
```suggestion
	vmblock, err := d.parser.ParseBlock(context.TODO(), canotoBlock.InnerBlock)
```

simplex/canoto.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package simplex
+
+//go:generate go run github.com/StephenButtolph/canoto/canoto $GOFILE
+
+// CanotoSimplexBlock is the Canoto representation of a verified block
```
StephenButtolph:
```suggestion
// CanotoSimplexBlock is the Canoto representation of a block
```

simplex/canoto.go
```diff
@@ -0,0 +1,14 @@
...
+type CanotoSimplexBlock struct {
+	Metadata   []byte `canoto:"bytes,1"`
+	InnerBlock []byte `canoto:"bytes,2"`
+
+	canotoData canotoData_CanotoSimplexBlock
+}
```
StephenButtolph: Couple questions:
1. Should this be defined next to `Block`? They feel tightly coupled.
2. Should this be unexported? I'm not against exporting it, but it seems like this is an implementation detail of `Block.Bytes()`. Perhaps it is a useful utility to export just so that other people can consume it... but 🤷.

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+
+	// Serialize the block
+	blockBytes, err := b.Bytes()
+	require.NoError(t, err)
+
+	tests := []struct {
+		name          string
+		expected      []byte
```
StephenButtolph: `expected` doesn't seem to be used. Should this be removed?

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+		parseFunc     func(context.Context, []byte) (snowman.Block, error)
+		expectedError error
+		blockBytes    []byte
+	}{
+		{
+			name:       "block serialization",
+			blockBytes: blockBytes,
+			expected:   blockBytes,
+			parseFunc: func(_ context.Context, b []byte) (snowman.Block, error) {
+				require.Equal(t, b, testBlock.BytesV, "block bytes should match")
```
StephenButtolph: This doesn't seem to be the correct `t` here. This `t` is referencing the outer `TestBlockSerialization` `t` rather than the `t.Run` `t`.

StephenButtolph: `Equal` takes in `(expected, actual)`.
```suggestion
				require.Equal(t, testBlock.BytesV, b, "block bytes should match")
```

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+func TestBlockSerialization(t *testing.T) {
+	unexpectedBlockBytes := errors.New("unexpected block bytes")
+
+	testVM := &blocktest.VM{
+		VM: enginetest.VM{
+			T: t,
+		},
+	}
```
StephenButtolph: `testVM` is defined outside the test loop, but it is not used in setting up the vectors. Can we define this struct inside the test loop?

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	testBlock := snowmantest.Block{
+		HeightV: 1,
+		BytesV:  []byte("test block"),
+	}
```
StephenButtolph:
```suggestion
	testBlock := snowmantest.BuildChild(snowmantest.Genesis)
```

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+			deserializer := &blockDeserializer{
+				parser: testVM,
+			}
+
+			// Deserialize the block
+			deserializedBlock, err := deserializer.DeserializeBlock(tt.blockBytes)
+			require.ErrorIs(t, err, tt.expectedError)
+
+			if tt.expectedError == nil {
+				require.Equal(t, deserializedBlock.BlockHeader().ProtocolMetadata, b.BlockHeader().ProtocolMetadata)
```
StephenButtolph:
```suggestion
				require.Equal(t, b.BlockHeader().ProtocolMetadata, deserializedBlock.BlockHeader().ProtocolMetadata)
```

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+		})
+	}
+}
+
+// TestBlockVerify tests a block cannot be verified more than once.
+func TestBlockVerify(t *testing.T) {
+	testBlock := snowmantest.Block{
+		HeightV: 1,
+		BytesV:  []byte("test block"),
+	}
```
StephenButtolph:
```suggestion
	testBlock := snowmantest.BuildChild(snowmantest.Genesis)
```

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+	require.NoError(t, err, "block should be serializable")
+
+	// Deserialize the block
+	deserializer := &blockDeserializer{
+		parser: &blocktest.VM{
+			VM: enginetest.VM{
+				T: t,
+			},
+			ParseBlockF: func(_ context.Context, b []byte) (snowman.Block, error) {
+				require.Equal(t, b, testBlock.BytesV, "block bytes should match")
```
StephenButtolph:
```suggestion
				require.Equal(t, testBlock.BytesV, b, "block bytes should match")
```

simplex/block_test.go
```diff
@@ -0,0 +1,149 @@
...
+				return &testBlock, nil
+			},
+		},
+	}
+	deserializedBlock, err := deserializer.DeserializeBlock(serializedBlock)
+	require.NoError(t, err, "block should be deserialized successfully")
+	require.Equal(t, b.BlockHeader().ProtocolMetadata, deserializedBlock.BlockHeader().ProtocolMetadata)
```
StephenButtolph: This chunk of code seems a bit misplaced to me. Don't we already test the serialization + parsing in `TestBlockSerialization`?

Is this just to verify that the bytes returned from the unverified block matches the bytes returned by the verified block? If so I feel like we could more easily just verify that `b.Bytes() == block.Bytes()`

simplex/block.go
samliok: I dont think this is needed, but I added it just in case. In `simplex` we have a struct called [oneTimeVerify](https://github.com/ava-labs/Simplex/blob/e4b626ad682ffaaf543646f0d97425e6b46f89fb/util.go#L143) which ensures we only verify the same digest once.

samliok: removed it just to avoid confusion

simplex/block.go
samliok: yes, we handle verifying the metadata in simplex [here](https://github.com/ava-labs/Simplex/blob/e4b626ad682ffaaf543646f0d97425e6b46f89fb/epoch.go#L1824). This also includes making sure the previous digest is expected.

simplex/canoto.go
samliok: yea i separated the file in case we end up having more canoto structs in the future, but i think it makes sense to keep it close to `Block` for now. Also un-exported it.

simplex/block.go
samliok: Created this PR on the simplex side https://github.com/ava-labs/Simplex/pull/178

simplex/block.go
yacovm: The `Prev` is a hash over the entire Simplex parent block, which has inside of it the VM block in its raw form.
I thought that if the the inner child block passes verification when the preference is set to the parent's inner block then it means that the Simplex child and parent blocks have a sibling relation with regard to their inner blocks.

Is that incorrect?

simplex/block.go
```diff
@@ -0,0 +1,102 @@
...
+	return buff, nil
+}
+
+func (b *Block) Verify(ctx context.Context) (simplex.VerifiedBlock, error) {
+	if b.verified {
+		return b, errBlockAlreadyVerified
+	}
+
+	// TODO: track blocks that have been verified to ensure they are either rejected or accepted
+	b.verified = true
```
yacovm: I don't think we should mark anything as verified, valid, ok, good, until we have verified and got nil.

Who knows where this object is moved to later ;-)

simplex/block.go
yacovm: Why does `ParseBlock` takes in a context in the first place? It should be a non-blocking implementation everywhere, so when would we want to abort or set limit to the time it takes to parse?

StephenButtolph: Contexts aren't used just for blocking implementations. For example, contexts are also used to support tracing.

simplex/block.go
samliok: going to leave potentially changing this for a different pr/time as discussed offline.

StephenButtolph (approved):
lgtm

simplex/block.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := canotoBlock.UnmarshalCanoto(bytes); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal block: %w", err)
+	}
+
+	md, err := simplex.ProtocolMetadataFromBytes(canotoBlock.Metadata)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse protocol metadata: %w", err)
+	}
+
+	vmblock, err := d.parser.ParseBlock(context.TODO(), canotoBlock.InnerBlock)
```
yacovm, line 82: we can now add a context now that [this](https://github.com/ava-labs/Simplex/pull/178/commits/c05066f1b7839a7ad1c15ffa2fec78091648dbc2) is merged.

PR #4017, opened 2025-06-17
Connectrpc info api

mpignatelli12:


api/info/connect_handler/connect_handler.go
```diff
@@ -0,0 +1,326 @@
+// connect_handler.go
+// This file implements the ConnectRPC handler for the InfoService
+// It should satisfy the infov1.InfoServiceHandler interface generated from service.proto
+
+package connect_handler
```
joshua-kim, line 5: As discussed offline, it's do this in the `info` package in `service.go`. This should make the diff easier to review, and is more aligned with the goal of this project of migrating the existing api rather than removing one and adding a new one.

api/info/connect_handler/connect_handler.go
```diff
@@ -0,0 +1,326 @@
...
+// that delegates calls to the existing Info implementation
+func NewConnectInfoService(info *info.Info) *ConnectInfoService {
+	return &ConnectInfoService{
+		Info: info,
+	}
+}
+
+type ConnectInfoService struct {
+	*info.Info
+}
```
joshua-kim, line 32: As discussed offline, it's do this in the `info` package in `service.go`. This should make the diff easier to review, and is more aligned with the goal of this project of migrating the existing api rather than removing one and adding a new one. This is kind of implemented in a nice way since we just wrap the old api and convert it into the new response type, but I just don't want the repo in an inconsistent state where we have two sets of apis.

examples/info_connect_client/main.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Create a client
+	client := infov1connect.NewInfoServiceClient(
+		http.DefaultClient,
+		nodeURL,
+	)
+
+	// Request node version
+	fmt.Println("Fetching node version...")
+	versionResp, err := client.GetNodeVersion(
```
joshua-kim: We should remove this file from the PR - since it seems to be used for manual testing purposes. If you want to test this PR you can follow some examples from our docs https://build.avax.network/docs/api-reference/info-api#infogetnodeversion that use `curl` from the command line to issue requests

joshua-kim:
Looks like this was moved to https://github.com/ava-labs/avalanchego/pull/4033

PR #4019, opened 2025-06-17
Add test to re-execute specified range of mainnet C-Chain blocks

aaronbuchwald:
## Why this should be merged

This PR adds a configurable C-Chain benchmark that can be run locally with [Taskfile](https://taskfile.dev/) or via GitHub Action (this PR currently includes triggers for pull requests and manual GH Action, but this should be changed to a cronjob on master and manual GH Action before merge and once done testing).

The test clones a block database and current state database, so that it can execute a range of blocks starting from the state at an arbitrary height N.

The CI workflow copies the data from an S3 Bucket in the Ava Labs Experimental Account in us-east-2 and requires AWS credentials to be provided. It can also be used with local file system directory paths for use on Snoopy/Linus (dedicated machines with large SSDs used for Firewood testing).

## How this works

Constructs an EVM instance mimicking the params passed in from AvalancheGo including passing in a metrics registry with the expected prefix and chain labels attached to it, so that it can be used directly with existing Grafana Dashboards used with tmpnet rather than needing to maintain an alternative without the prefixes/labels.

## How this was tested

This has been tested by running in CI and running through the steps documented in the README. To review this PR, please walk through the steps in the benchmark [README](https://github.com/ava-labs/avalanchego/blob/aaronbuchwald/cchain-reexecute-range-test/tests/bench/c/README.md).

No.

## Recommended Changes Before Merge

- [x] Remove PR trigger in favor of weekly cronjob (currently in use to trigger while testing this PR)
- [x] Update test name to include exact block range

## Recommended Fast Follow Ups to this PR

- [ ] Identify the best place to run this job ie. ARC, self-hosted runner, Blacksmith, etc.
- [ ] Integrate [GitHub Action Benchmark](https://github.com/benchmark-action/github-action-benchmark/) performance tracking on master by setting up `gh-pages` branch or a separate repo (demo of what this will look like on my fork of Canoto [here](https://aaronbuchwald.github.io/canoto/dev/bench/)
## Potential Future Work

- Identify a reasonable micro-benchmark that can be run on every PR and provide a reasonable signal whether the change triggers a performance improvement or degradation
- Implement a job to re-execute the entire C-Chain history for full coverage that uses multiple snapshots and executes multiple subranges in parallel (ie. [0, 5m], [5m, 10m], etc.)
- Implement a task/job to archive all blocks over configured ranges to S3 by fetching directly from the network
- Add an optional feature at the end of the re-execution job that can post the resulting state to S3 (useful for manual GH Action entry point to create new current state snapshots)
- Add support for long-lived feature branches or alternative configurations to run as a cronjob in addition to the default configuration run on the master branch of AvalancheGo
- Support multiple configurations (archive/full) and targets (EBS, SSD, etc) to run the benchmark against
- Support all VMs (P-Chain and X-Chain) for complete test coverage and benchmarking other VMs performance
- Provide a "How-To Guide" for custom VM testing using the same approach


.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -110,4 +111,3 @@ runs:
       with:
         name: ${{ inputs.artifact_prefix }}-kind-logs
         path: /tmp/kind-logs
```
aaronbuchwald: Benchmark does not produce logs in the same location, so we must allow no files to be found to re-use it. If we are going to re-use for this purpose, we should consider how to split out this functionality or just add an input boolean to determine whether to attempt to load artifacts at this location.

.github/workflows/ci.yml
```diff
@@ -61,6 +61,20 @@ jobs:
           prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
           loki_username: ${{ secrets.LOKI_ID || '' }}
           loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
+  c-chain-reexecution:
```
aaronbuchwald: Alternative to modifying `run-monitored-tmpnet-cmd` to support this case is to copy the required setup here. Leaving this for now.

tests/load/prometheus.go
```diff
@@ -23,27 +23,24 @@ import (
 
 type MetricsServer struct {
 	addr     string
-	registry *prometheus.Registry
+	gatherer prometheus.Gatherer
```
aaronbuchwald: Consider moving this to `tests/metrics.go`

Taskfile.yml
```diff
@@ -254,3 +254,64 @@ tasks:
...
+  import-s3-to-dir:
+    desc: Imports an S3 path to a local directory. Unzipping if needed.
+    vars:
+      S3_SRC: '{{.S3_SRC}}'
+      LOCAL_DST: '{{.LOCAL_DST}}'
+    cmds:
+      - cmd: bash -x ./scripts/copy_dir.sh {{.S3_SRC}} {{.LOCAL_DST}}
+
+  import-cchain-reexecute-range:
+    desc: Imports the C-Chain block and state data to re-execute. Defaults to execute the block range (100, 200].
```
aaronbuchwald:
```suggestion
    desc: Imports the C-Chain block and state data to re-execute. Defaults to import the first 200 and the current state created with the default config of the C-Chain (hashdb).
```

Taskfile.yml
```diff
@@ -254,3 +254,64 @@ tasks:
...
+        vars:
+          S3_SRC: '{{.SOURCE_BLOCK_DIR}}'
+          LOCAL_DST: '{{.EXECUTION_DATA_DIR}}/blocks'
+      - task: import-s3-to-dir
+        vars:
+          S3_SRC: '{{.CURRENT_STATE_DIR}}'
+          LOCAL_DST: '{{.EXECUTION_DATA_DIR}}/current-state'
+
+  reexecute-cchain-range:
+    desc: Benchmark re-execute a range of blocks on the C-Chain. Defaults to execute the block range (100, 200].
```
aaronbuchwald:
```suggestion
    desc: Re-execute a range of C-Chain blocks.
```

tests/bench/c/README.md
```diff
@@ -0,0 +1,102 @@
+# C-Chain Benchmark
+
+The C-Chain benchmarks support re-executing a range of mainnet C-Chain blocks against a provided snapshot of the current state as of some starting point.
```
aaronbuchwald:
```suggestion
The C-Chain benchmarks support re-executing a range of mainnet C-Chain blocks against a provided snapshot of the current state as of some initial state.
```

aaronbuchwald:
Need to migrate metrics from DataDog [dashboard](https://app.datadoghq.com/dashboard/6g5-z6k-3qv/avalanche-summary--networking?fromUser=true&refresh_mode=paused&from_ts=1750885734318&to_ts=1751490534318&live=false&tile_focus=6426494357766360) to https://github.com/ava-labs/avalanche-monitoring

aaronbuchwald:
> Need to migrate metrics from DataDog [dashboard](https://app.datadoghq.com/dashboard/6g5-z6k-3qv/avalanche-summary--networking?fromUser=true&refresh_mode=paused&from_ts=1750885734318&to_ts=1751490534318&live=false&tile_focus=6426494357766360) to https://github.com/ava-labs/avalanche-monitoring

Added https://github.com/ava-labs/avalanche-monitoring/pull/29

aaronbuchwald:
Metrics from the last run do not show up when the `gh_job_id` filter is applied for C-Chain reexecution:

https://grafana-poc.avax-dev.network/d/c-chain-processing-custom-v3/avalanche-c-chain-processing-metrics-custom?from=2025-07-08T14:10:45.000Z&to=2025-07-08T14:25:45.000Z&timezone=America%2FNew_York&var-datasource=P1809F7CD0C75ACF3&var-filter=gh_repo%7C%3D%7Cava-labs%2Favalanchego&var-filter=gh_run_id%7C%3D%7C16145632981&var-filter=gh_run_attempt%7C%3D%7C1&var-filter=gh_job_id%7C%3D%7Cc-chain-reexecution&orgId=1&var-chain=C&refresh=10s

With C-Chain reexecution removed, it displays the metrics from the other runs triggered by the same commit (which are not the metrics of interest here and the label should be applied):

https://grafana-poc.avax-dev.network/d/c-chain-processing-custom-v3/avalanche-c-chain-processing-metrics-custom?from=2025-07-08T14:10:45.000Z&to=2025-07-08T14:25:45.000Z&timezone=America%2FNew_York&var-datasource=P1809F7CD0C75ACF3&var-filter=gh_repo%7C%3D%7Cava-labs%2Favalanchego&var-filter=gh_run_id%7C%3D%7C16145632981&var-filter=gh_run_attempt%7C%3D%7C1&orgId=1&var-chain=C&refresh=10s

This seems to indicate the metrics from the reexecution job are not collected correctly.

.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -40,6 +40,13 @@ inputs:
     default: ${{ github.run_attempt }}
   job:
     default: ${{ github.job }}
+  grafana_preview_dashboard_name:
+    default: 'avalanche-main-dashboard'
+  orgId:
```
aaronbuchwald: Revert adding the `orgId` and `refresh` as params

aaronbuchwald: Update c chain benchmark to provide c chain processing dashboard name as an input

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,61 @@
...
+        - name: Run C-Chain Re-Execution
+          uses: ./.github/actions/run-monitored-tmpnet-cmd
+          with:
+            run: EXECUTION_DATA_DIR=$HOME/reexecution-data bash -x ./scripts/run_task.sh reexecute-cchain-range-with-copied-data
+            artifact_prefix: c-chain-reexecute-range
+            prometheus_username: ${{ secrets.PROMETHEUS_ID || '' }}
+            prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
+            loki_username: ${{ secrets.LOKI_ID || '' }}
+            loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
+            runtime: "" # Set runtime input to empty string to disable log collection
```
aaronbuchwald: add new line

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+	executor, err := newVMExecutor(sourceVM, consensusRegistry)
+	r.NoError(err)
+	r.NoError(executor.executeSequence(ctx, blockChan, executionTimeout))
+}
+
+func TestExportBlockRange(t *testing.T) {
+	exportBlockRange(t, sourceBlockDirArg, targetBlockDirArg, startBlockArg, endBlockArg, chanSizeArg)
+}
+
```
aaronbuchwald: move closer to function it's calling

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +276,53 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+func WritePrometheusServiceDiscoveryConfigFile(name string, sdConfigs []SDConfig, withGitHubLabels bool) (string, error) {
+	serviceDiscoveryDir, err := GetPrometheusServiceDiscoveryDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get service discovery dir: %w", err)
+	}
+	sdConfigsJSONPre, err := json.MarshalIndent(sdConfigs, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal sd configs: %w", err)
+	}
+	fmt.Println("sdConfigsJSONPre: ", string(sdConfigsJSONPre))
```
aaronbuchwald: remove prints

scripts/copy_dir.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
aaronbuchwald, line 3: Future improvement: consider replacing with [rclone](https://rclone.org/)

tests/bench/c/README.md
```diff
@@ -0,0 +1,93 @@
+# C-Chain Benchmark
```
aaronbuchwald: Run through again

aaronbuchwald: Add section on using the GitHub Action

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,397 @@
...
+	r.NoError(executor.executeSequence(ctx, blockChan, executionTimeout))
+}
+
+func newMainnetCChainVM(
+	ctx context.Context,
+	vmAndSharedMemoryDB database.Database,
+	chainDataDir string,
+	configBytes []byte,
+	metricsGatherer metrics.MultiGatherer,
+) (block.ChainVM, error) {
```
aaronbuchwald: Future improvement: update to use a VM factory, initialize any VM, and take arguments/config for all of the necessary paramters ie. networkID, X-Chain ID, etc.

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,397 @@
...
+	vm      block.ChainVM
+	metrics *consensusMetrics
+}
+
+type consensusMetrics struct {
+	lastAcceptedHeight prometheus.Gauge
+}
+
+func newConsensusMetrics(registry prometheus.Registerer) (*consensusMetrics, error) {
+	// Mimic the metrics from snowman consensus [engine](../../snow/engine/snowman/metrics.go).
```
aaronbuchwald: Add comment explaining that the registry passed in must be provided with the appropriate matching prefix and chain label as handled in `chains/manager.go`

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+		if blkResult.Height%1000 == 0 {
+			e.log.Info("executing block", zap.Uint64("height", blkResult.Height))
+		}
+		if err := e.execute(ctx, blkResult.BlockBytes); err != nil {
+			return err
+		}
+		if executionTimeout > 0 && time.Since(start) > executionTimeout {
+			e.log.Info("exiting early due to execution timeout", zap.Duration("elapsed", time.Since(start)), zap.Duration("execution-timeout", executionTimeout))
+			return nil
```
aaronbuchwald: Move to a config type for the executor that includes a registry

aaronbuchwald:
Need to update the Grafana link to properly replace the UID + title ie. https://grafana-poc.avax-dev.network/d/c-chain-processing-custom-v8/c-chain-processing?orgId=1&from=now-6h&to=now&timezone=browser&var-datasource=P1809F7CD0C75ACF3&var-filter=&var-chain=C&refresh=10s rather than `/d/<old-uid>/c-chain-processing`

aaronbuchwald:
Need to add configuration so the link displays without `is_ephemeral_node` label

aaronbuchwald:
https://github.com/ava-labs/avalanchego/pull/4019/commits/ec4eecb11c422f7e4766acfdbb5e9e29bc605500 adds [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#charts-on-github-pages-1) tracking, which supports both tracking performance on the main branch and displaying via GitHub Pages (not added in this commit) and comparing performance results

aaronbuchwald:
https://github.com/ava-labs/avalanchego/pull/4019/commits/8bb9ccf17f08155000c721514942b0e16349e9c7 incorporates the signature recovery metrics from https://github.com/ava-labs/coreth/pull/1035

aaronbuchwald:
Note: running locally correctly exports the metrics to Grafana, but when using Linus/Snoopy (or in general), must be careful to ensure that prometheus is running correctly.

I ran into some issues where https://github.com/ava-labs/avalanchego/blob/217c3ce6b85eee9fed64a47ebff88b1e3a141b13/tests/fixture/tmpnet/monitor_processes.go#L442 fails silently to start exporting the metrics because the address was already occupied by Prometheus configured on these nodes.

Since `StartPrometheus` is not currently configurable, it may be desirable to make it configurable and ensure that it fails loudly rather than silently as it was not obvious why metrics were being exported locally, but not when running on Linus/Snoopy.

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,77 @@
...
+          uses: actions/cache@v4
+          with:
+            path: ./cache
+            key: ${{ runner.os }}-reexecute-cchain-range-benchmark.json
+        - name: Compare Benchmark Results
+          uses: benchmark-action/github-action-benchmark@v1
+          with:
+            tool: 'go'
+            output-file-path: $GITHUB_WORKSPACE/reexecution-data/reexecute-cchain-range.txt
+            external-data-json-path: ./cache/${{ runner.os }}-reexecute-cchain-range-benchmark.json
```
aaronbuchwald: Writing to the cache and comparing against the last entry enables a comparison.

This should be changed to write to the cache only on the master branch for comparison and other triggers should compare against the latest baseline set by master

Taskfile.yml
```diff
@@ -266,3 +266,79 @@ tasks:
...
+      TARGET_BLOCK_DIR: '{{.TARGET_BLOCK_DIR}}'
+      START_BLOCK: '{{.START_BLOCK}}'
+      END_BLOCK: '{{.END_BLOCK}}'
+    cmds:
+      - cmd: go test -timeout=0 -run=TestExportBlockRange github.com/ava-labs/avalanchego/tests/bench/c --source-block-dir={{.SOURCE_BLOCK_DIR}} --target-block-dir={{.TARGET_BLOCK_DIR}} --start-block={{.START_BLOCK}} --end-block={{.END_BLOCK}}
+
+  reexecute-cchain-range:
+    desc: Re-execute a range of C-Chain blocks.
+    vars:
+      EXECUTION_DATA_DIR: '{{.EXECUTION_DATA_DIR}}'
```
aaronbuchwald: rename to `CURRENT_STATE_DIR`

aaronbuchwald:
Trying out running on ARC and Blacksmith here:
- https://github.com/ava-labs/avalanchego/pull/4093
- https://github.com/ava-labs/avalanchego/pull/4094

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
+name: C-Chain Benchmark
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      start-block:
+        description: 'The start block for the benchmark.'
+        required: false
+        default: '100'
```
maru-ava: Rather than duplicating the defaults here and in the job, maybe use env vars e.g.

```yaml
env: 
  START_BLOCK: '100'
  ...

on:
  pull_request:
  workflow_dispatch:
    inputs:
      start-block:
        description: 'The start block for the benchmark.'
        required: false
        default: ${{ env.START_BLOCK }}
        ...

File: .github/workflows/c-chain-benchmark.yml [hunk 19]
```diff
@@ -0,0 +1,79 @@
...
+    c-chain-reexecution:
+      permissions:
+        id-token: write
+        contents: write
+      runs-on: ubuntu-latest
+      steps:
+        - name: Configure AWS Credentials
+          uses: aws-actions/configure-aws-credentials@v4
+          with:
+            role-to-assume: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
```
maru-ava: Since this secret won't be available from fork branches, you'll want to make job execution conditional (as per the example of [run-monitored-tmpnet-cmd](https://github.com/ava-labs/avalanchego/blob/master/.github/actions/run-monitored-tmpnet-cmd/action.yml#L68) so that the job doesn't fail on fork branch PRs.

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
...
+            aws-region: us-east-2
+        - name: Set task env via GITHUB_ENV
+          id: set-params
+          run: |
+            if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
```
maru-ava: Since `github.event.inputs.*` will be undefined when the job is not triggered by `workflow_dispatch`, a simpler way of setting each variable would be using `||` to supply a default:

```yaml
echo "START_BLOCK=${{ github.event.inputs.start-block || env.START_BLOCK }}" >> $GITHUB_ENV`
```

Note that the echoed text needs to be appended to `$GITHUB_ENV` to affect the environment - the current form would only result in log output.

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
...
+        - uses: actions/checkout@v4
+        - uses: ./.github/actions/setup-go-for-project
+        - name: Run C-Chain Re-Execution
+          uses: ./.github/actions/run-monitored-tmpnet-cmd
+          with:
+            run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=${{ github.workspace }}/reexecution-data BENCHMARK_OUTPUT_FILE=${{ github.workspace }}/reexecute-cchain-range-benchmark-res.txt
+            artifact_prefix: c-chain-reexecute-range
+            prometheus_username: ${{ secrets.PROMETHEUS_ID || '' }}
+            prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
+            grafana_dashboard_id: 'c-chain-processing-custom-v8/c-chain-processing'
```
maru-ava: Is this the correct dashboard ID now that you've incorporated the target metrics into the main C-Chain dashboard?

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
...
+              echo "SOURCE_BLOCK_DIR=s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb.zip"
+              echo "CURRENT_STATE_DIR=s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-100.zip"
+            fi
+        - uses: actions/checkout@v4
+        - uses: ./.github/actions/setup-go-for-project
+        - name: Run C-Chain Re-Execution
+          uses: ./.github/actions/run-monitored-tmpnet-cmd
+          with:
+            run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=${{ github.workspace }}/reexecution-data BENCHMARK_OUTPUT_FILE=${{ github.workspace }}/reexecute-cchain-range-benchmark-res.txt
+            artifact_prefix: c-chain-reexecute-range
```
maru-ava: This can be omitted since it is only relevant if `inputs.runtime` is set to one of `process` or `kube`. If there is an artifact to collect, the job will need to call `actions/upload-artifact` itself.

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
+name: C-Chain Benchmark
```
maru-ava: Maybe be more specific in both the name of the file and the workflow name? e.g. `C-Chain Re-execution Benchmark`

.github/workflows/c-chain-benchmark.yml
```diff
@@ -0,0 +1,79 @@
...
+          with:
+            run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=${{ github.workspace }}/reexecution-data BENCHMARK_OUTPUT_FILE=${{ github.workspace }}/reexecute-cchain-range-benchmark-res.txt
+            artifact_prefix: c-chain-reexecute-range
+            prometheus_username: ${{ secrets.PROMETHEUS_ID || '' }}
+            prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
+            grafana_dashboard_id: 'c-chain-processing-custom-v8/c-chain-processing'
+            loki_username: ${{ secrets.LOKI_ID || '' }}
+            loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
+            runtime: "" # Set runtime input to empty string to disable log collection
+        - name: Download Previous Benchmark Result
```
maru-ava: I noticed that mixing cron with workflow_dispatch would be problematic, but then I saw your comment. Maybe add a TODO list of work in the PR description that includes this and any other work that should be done pre-merge?

Taskfile.yml
```diff
@@ -266,3 +266,86 @@ tasks:
     cmds:
       - task: build
       - cmd: bash -x ./scripts/tests.upgrade.sh {{.CLI_ARGS}}
+
```
maru-ava: Please add new tasks in sorted order for consistency.

Taskfile.yml
```diff
@@ -266,3 +266,86 @@ tasks:
...
+      SRC: '{{.SRC}}'
+      DST: '{{.DST}}'
+    cmds:
+      - cmd: bash -x ./scripts/copy_dir.sh {{.SRC}} {{.DST}}
+
+  import-cchain-reexecute-range:
+    desc: Imports the C-Chain block and state data to re-execute. Defaults to import the first 200 and the current state created with the default config of the C-Chain (hashdb).
+    vars:
+      EXECUTION_DATA_DIR: '{{.EXECUTION_DATA_DIR}}'
+      SOURCE_BLOCK_DIR: '{{.SOURCE_BLOCK_DIR | default "s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-200.zip"}}'
```
maru-ava: Why does this var have a _DIR suffix if the default is a zip file?

Taskfile.yml
```diff
@@ -266,3 +266,86 @@ tasks:
...
+    desc: Re-execute a range of C-Chain blocks.
+    vars:
+      CURRENT_STATE_DIR: '{{.CURRENT_STATE_DIR}}'
+      SOURCE_BLOCK_DIR: '{{.SOURCE_BLOCK_DIR}}'
+      BENCHMARK_OUTPUT_FILE: '{{.BENCHMARK_OUTPUT_FILE | default ""}}'
+      START_BLOCK: '{{.START_BLOCK}}'
+      END_BLOCK: '{{.END_BLOCK}}'
+      LABELS: '{{.LABELS | default ""}}'
+    cmds:
+      - cmd: |
```
maru-ava: Please move this to a proper script to ensure we get the benefit of `set -euo pipefail`. Task commands are ideally limited to a single trivial line.

scripts/copy_dir.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Usage: ./scripts/copy_dir.sh source_directory destination_directory
+# Sources can be S3 URLs (s3://bucket/path) or a local file path
+# Assumes s5cmd has been installed and is available in the PATH
```
maru-ava: Maybe mention the availability of s5cmd in the nix dev shell?

tests/bench/c/README.md
```diff
@@ -0,0 +1,101 @@
...
```
maru-ava: Same comment as the workflow - maybe be more specific in the path and in this file as to the kind of benchmark? 'reexecution' would seem the relevant keyword.

tests/bench/c/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+The C-Chain benchmarks support re-executing a range of mainnet C-Chain blocks against a provided snapshot of the current state as of some initial state.
+
+AvalancheGo provides a [Taskfile](https://taskfile.dev/) with commands to manage the import/export of data required for re-execution (block range and current state) and triggering a benchmark run.
+
+## Metrics
+
+The C-Chain benchmarks export VM metrics to the same Grafana instance as AvalancheGo CI: https://grafana-poc.avax-dev.network/.
+
+You can view granular C-Chain processing metrics with the label attached to this job (job="c-chain-reexecution") [here](https://grafana-poc.avax-dev.network/d/c-chain-processing-custom-v8/c-chain-processing?orgId=1&from=now-5m&to=now&timezone=browser&var-datasource=P1809F7CD0C75ACF3&var-filter=job%7C%3D%7Cc-chain-reexecution&var-chain=C&refresh=10s).
```
maru-ava: Same question as in the workflow - is this the correct dashboard ID?

tests/bench/c/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Configure Dev Environment
+
+To set up your dev environment to run C-Chain benchmarks, run:
+
+```bash
+nix develop
+```
+
+If using AWS to push/pull S3 buckets, configure your AWS profile with the required access. The instructions here utilize the S3 bucket `s3://avalanchego-bootstrap-testing` in `us-east-2` under the Ava Labs Experimental AWS account.
```
maru-ava: Should we be documenting the use of internal infrastructure in our public repo vs limiting those details to internal wiki/google doc/etc? I've been thinking similar thoughts about documentation around tmpnet monitoring.

tests/bench/c/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Create C-Chain State Snapshot
+
+To execute a range of blocks [N, N+K], we need an initial current state with the last accepted block of N-1. To generate this from scratch, simply execute the range of blocks [1, N-1] (genesis is "executed" vacuously, so do not provide 0 as the start block) locally starting from an empty state:
+
+```bash
+task reexecute-cchain-range CURRENT_STATE_DIR=$HOME/exec-data/current-state SOURCE_BLOCK_DIR=$HOME/exec-data/blocks START_BLOCK=1 END_BLOCK=100
+```
+
+This initailizes a `current-state` subdirectory inside of `$HOME/exec-data`, which will contain two subdirectories `chain-data-dir` and `db`. The `chain-data-dir` is the path passed in via `*snow.Context` to the VM as `snowContext.ChainDataDir`. If the VM does not populate it, it may remain empty after a run. The `db` directory is used to initialize the leveldb instance used to create two nested PrefixDBs the database passed into `vm.Initialize(...)` and the database used by shared memory. These two databases must be built on top of the same base database as documented in the shared memory [README](../../../chains/atomic/README.md#shared-database).
```
maru-ava: Consider adding line breaks to this paragraph to simplify review. Github review works best with shorter line lengths to allow a reviewer to zero in on what they are commenting on.

`initailizes` -> `initializes`
`PrefixDBs` -> `PrefixDBs:`

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	flag.StringVar(&targetBlockDirArg, "target-block-dir", targetBlockDirArg, "DB directory to write blocks into when executing TestExportBlockRange.")
+
+	// Target directory assumes the current-state directory contains a db directory and a chain-data-dir directory.
+	// - db/
+	// - chain-data-dir/
+	flag.StringVar(&targetDirArg, "target-dir", targetDirArg, "Target directory for the current state including VM DB and Chain Data Directory.")
+	flag.Uint64Var(&startBlockArg, "start-block", 101, "Start block to begin execution (exclusive).")
+	flag.Uint64Var(&endBlockArg, "end-block", 200, "End block to end execution (inclusive).")
+	flag.IntVar(&chanSizeArg, "chan-size", 100, "Size of the channel to use for block processing.")
+	flag.DurationVar(&executionTimeout, "execution-timeout", 0, "Benchmark execution timeout. After this timeout has elapsed, terminate the benchmark without error.")
```
maru-ava: (No action required) Maybe make mention that `0` configures no timeout?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+
+	flag.BoolVar(&metricsEnabledArg, "metrics-enabled", true, "Enable metrics collection.")
+	flag.StringVar(&labelsArg, "labels", "", "Comma separated KV list of metric labels to attach to all exported metrics.")
```
maru-ava: (No action required) Maybe provide an example?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	}, true)
+	r.NoError(err)
+}
+
+func parseLabels(labelsStr string) map[string]string {
+	labels := make(map[string]string)
+	for _, label := range strings.Split(labelsStr, ",") {
+		parts := strings.SplitN(label, "=", 2)
+		if len(parts) != 2 {
+			continue
```
maru-ava: (No action required) Why isn't this an error condition?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	r.NoError(prefixGatherer.Register("avalanche_snowman", consensusRegistry))
+
+	if metricsEnabled {
+		labels := map[string]string{
+			"job":               "c-chain-reexecution",
+			"is_ephemeral_node": "false",
+			"chain":             "C",
+		}
+		if labelsArg != "" {
+			customLabels := parseLabels(labelsArg)
```
maru-ava: (No action required) Maybe parse labels before running (i.e. after flag.Parse) to ensure valid input?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	metricsGatherer metrics.MultiGatherer,
+) (block.ChainVM, error) {
+	factory := factory.Factory{}
+	vmIntf, err := factory.New(logging.NoLog{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create VM from factory: %w", err)
+	}
+	vm := vmIntf.(block.ChainVM)
+
+	log := tests.NewDefaultLogger("mainnet-vm-reexecution")
```
maru-ava: (No action required) Maybe initialize this inline since it's not used otherwise? Ditto for warpSigner.

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+			e.config.Log.Info("exiting early due to execution timeout", zap.Duration("elapsed", time.Since(start)), zap.Duration("execution-timeout", executionTimeout))
+			return nil
+		}
+	}
+	e.config.Log.Info("finished executing sequence")
+
+	return nil
+}
+
+func createBlockChanFromLevelDB(tb testing.TB, sourceDir string, startBlock, endBlock uint64, chanSize int) (<-chan blockResult, error) {
```
maru-ava: (No action required) Why does this file use both `testing.B` and `testing.TB`?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	)
+	r.NoError(err)
+	defer func() {
+		r.NoError(sourceVM.Shutdown(ctx))
+	}()
+
+	config := vmExecutorConfig{
+		Log:              tests.NewDefaultLogger("vm-executor"),
+		Registry:         consensusRegistry,
+		ExecutionTimeout: executionTimeout,
```
maru-ava: Given that the timeout is passed directly to `executeSequence`, maybe remove it from the struct?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+			return blkResult.Err
+		}
+
+		if blkResult.Height%1000 == 0 {
+			e.config.Log.Info("executing block", zap.Uint64("height", blkResult.Height))
+		}
+		if err := e.execute(ctx, blkResult.BlockBytes); err != nil {
+			return err
+		}
+		if executionTimeout > 0 && time.Since(start) > executionTimeout {
```
maru-ava: Maybe wrap the context with `context.WithTimeout` if the timeout is non-zero and use a switch?

```golang
  // In for loop:
  select {
    case <-ctx.Done():
      ...
    case blkResult := <-blkChan:
      ...
  }
```

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	server, err := tests.NewPrometheusServer(gatherer)
+	r.NoError(err)
+
+	var sdConfigFilePath string
+	tb.Cleanup(func() {
+		// Ensure a final metrics scrape.
+		// This default delay is set above the default scrape interval used by StartPrometheus.
+		time.Sleep(tmpnet.NetworkShutdownDelay)
+
+		r.NoError(server.Stop())
```
maru-ava: If this fails, it will prevent the removal of sdConfigFilePath. Ensure that both cleanup steps will always execute by registering them separately with `tb.Cleanup`.

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	gasUsed, err := getCounterMetricValue(b, registry, "avalanche_evm_eth_chain_block_gas_used_processed")
+	r.NoError(err)
+	mgasPerSecond := gasUsed / 1_000_000 / elapsed.Seconds()
+	b.ReportMetric(mgasPerSecond, "mgas/s")
+}
+
+func getCounterMetricValue(tb testing.TB, registry prometheus.Gatherer, query string) (float64, error) {
+	metricFamilies, err := registry.Gather()
+	r := require.New(tb)
+	r.NoError(err)
```
maru-ava: In a given function, please avoid mixing testify assertions with a returned error. In this case, maybe just return the error given that the caller is already using testify to check it?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+
+		if sdConfigFilePath != "" {
+			r.NoError(os.Remove(sdConfigFilePath))
+		}
+	})
+
+	sdConfigFilePath, err = tmpnet.WritePrometheusSDConfig(name, tmpnet.SDConfig{
+		Targets: []string{server.Address()},
+		Labels:  labels,
+	}, true)
```
maru-ava: Please try to avoid the use of magic values like this one in favor of either a descriptively-named variable (e.g. `withGithubLabels` or by adding a comment e.g. `true /* withGithubLabels */`.

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+}
+
+// newConsensusMetrics creates a subset of the metrics from snowman consensus
+// [engine](../../snow/engine/snowman/metrics.go).
+//
+// The registry passed in is expected to be registered with the prefix
+// "avalanche_snowman" and the chain label (ex. chain="C") that would be handled
+// by the[chain manager](../../../chains/manager.go).
+func newConsensusMetrics(registry prometheus.Registerer) (*consensusMetrics, error) {
+	errs := wrappers.Errs{}
```
maru-ava: Why is this suggested for a single potential error?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	e.metrics.lastAcceptedHeight.Set(float64(blk.Height()))
+
+	return nil
+}
+
+func (e *vmExecutor) executeSequence(ctx context.Context, blkChan <-chan blockResult, executionTimeout time.Duration) error {
+	blkID, err := e.vm.LastAccepted(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get last accepted block: %w", err)
```
maru-ava: (No action required) Maybe differentiate the error from the subsequent one e.g. `block` -> `block id`?

tests/bench/c/vm_test.go
```diff
@@ -0,0 +1,493 @@
...
+	flag.BoolVar(&metricsEnabledArg, "metrics-enabled", true, "Enable metrics collection.")
+	flag.StringVar(&labelsArg, "labels", "", "Comma separated KV list of metric labels to attach to all exported metrics.")
+
+	flag.Parse()
+	m.Run()
+}
+
+func BenchmarkReexecuteRange(b *testing.B) {
+	benchmarkReexecuteRange(b, sourceBlockDirArg, targetDirArg, startBlockArg, endBlockArg, chanSizeArg, metricsEnabledArg)
+}
```
StephenButtolph: Is this correct usage of golang `Benchmark` tests? I've never seen a benchmark that didn't utilize `b.N` to repeatedly run the benchmark.

.github/workflows/c-chain-benchmark.yml
aaronbuchwald: Updated PR description to reformat a TODO list including removing the PR trigger for this job.

Would you prefer adding that check regardless of the fact it won't run on PRs after being merged?

.github/workflows/c-chain-benchmark.yml
aaronbuchwald: Ah I didn't see the last Grafana PR had been merged. I'll go ahead and update CI grafana and the display link here as well.

.github/workflows/c-chain-benchmark.yml
aaronbuchwald: Updated the PR description, so that the only item listed as TODO pre-merge is to remove the PR trigger.

Taskfile.yml
aaronbuchwald: The script / task automatically unpackages the zip, which is expected to contain a directory.

scripts/copy_dir.sh
aaronbuchwald: Good call, will add a line to mention this.

tests/bench/c/README.md
aaronbuchwald: Updating

tests/bench/c/README.md
aaronbuchwald: I tried to write this documentation, so that it uses the Ava Labs S3 bucket as an example, but so that you could sub an alternative in as well.

I think we already do this with tmpnet a bit. If preferred to actively make it more agnostic, I'd suggest setting the bucket as an environment variable for the bucket and using that throughout the docs.

I'd still prefer to keep the example of re-executing with the defaults, which does require that S3 buckets already exist.

I'd rather keep documentation in one place to reduce sprawl and adopt a style that explains both how to use it as an internal developer and external if we want to ensure external devs can use it.

tests/bench/c/vm_test.go
aaronbuchwald: Updating

tests/bench/c/vm_test.go
aaronbuchwald: Will move this, I prefer to keep `warpSigner` in place, so that it's clustered with the creation of the underlying BLS private and public key

tests/bench/c/vm_test.go
aaronbuchwald: Updated this file now uses `testing.B` where it requires access to benchmark specific functions ie. `b.ReportMetric` and otherwise uses `testing.TB` for helper functions that can take either input.

This runs as a benchmark specifically to report metrics in the golang benchmark format that can be utilized by https://github.com/benchmark-action/github-action-benchmark/

tests/bench/c/vm_test.go
aaronbuchwald: Updated to use it via config rather than pass via `executeSequence`

tests/bench/c/vm_test.go
aaronbuchwald: Adding this slightly differently ie. put the select statement within the loop so that it still loops over the channel directly.

tests/bench/c/vm_test.go
aaronbuchwald: Combining with `errors.Join(...)`. I wanted to ensure that the server shuts down before I wipe the SD Config file, so that the server never picks up the file being removed.

tests/bench/c/vm_test.go
aaronbuchwald: This was simply copied from the link where there were more than one, changing to a single error.

tests/bench/c/vm_test.go
aaronbuchwald: Good call

Taskfile.yml
aaronbuchwald: Moving to `./scripts/benchmark_cchain_range.sh`

tests/bench/c/vm_test.go
aaronbuchwald: This is definitely a bastardization to produce the benchmark output and requires setting `-benchtime=1x`. Will document this and make sure it fails loudly. The expectation here is that it is going to overwrite the `current-state` and does not handle reverting or creating a new copy of the same thing.

aaronbuchwald: Adding:
```
require.Equalf(b, 1, b.N, "BenchmarkReexecuteRange expects to run a single iteration because it overwrites the input current-state, but found (b.N=%d)", b.N)
```

.github/workflows/c-chain-benchmark.yml
maru-ava: No, that's fine.

.github/workflows/c-chain-benchmark.yml
maru-ava: This comment was about cron vs workflow_dispatch though. workflow_dispatch allows targeting any branch, so unless this workflow is updated to e.g. separate result caching by branch, a given run triggered by workflow_dispatch could cache a result that then would become the baseline for a subsequent unrelated run e.g.

   1. workflow_dispatch run #1 → saves results to cache
   2. workflow_dispatch run #2 → compares against run #1's results (not main branch!)

I recall mention in your presentation about being able to save benchmark history in branches though, would that avoid this issue?

Taskfile.yml
maru-ava: If that's the case, why not SOURCE_BLOCK_ZIP? Or a name that reflects the possibility of both zip and dir input?

tests/bench/c/README.md
maru-ava: I guess my point is that I'm not sure we should be doing it with tmpnet either. It's the bad example, not the good one. We're getting into a bad habit of putting everything in this repo for convenience and I'm planning to avoid doing that going forward.

Taskfile.yml
aaronbuchwald: In either case it's a directory, it just automatically unpackages a zip for convenience. What would you suggest?

.github/workflows/c-chain-benchmark.yml
aaronbuchwald: Tried pushing this and realized that `env` is not available here.

Here's the results from `actionlint`:

```
actionlint
.github/workflows/c-chain-reexecution-benchmark.yml:16:22: context "env" is not allowed here. no context is available here. see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details [expression]
   |
16 |         default: ${{ env.START_BLOCK }}
   |                      ^~~~~~~~~~~~~~~
.github/workflows/c-chain-reexecution-benchmark.yml:20:22: context "env" is not allowed here. no context is available here. see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details [expression]
   |
20 |         default: ${{ env.END_BLOCK }}
   |                      ^~~~~~~~~~~~~
.github/workflows/c-chain-reexecution-benchmark.yml:24:22: context "env" is not allowed here. no context is available here. see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details [expression]
   |
24 |         default: ${{ env.SOURCE_BLOCK_DIR }}
   |                      ^~~~~~~~~~~~~~~~~~~~
.github/workflows/c-chain-reexecution-benchmark.yml:28:22: context "env" is not allowed here. no context is available here. see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details [expression]
   |
28 |         default: ${{ env.CURRENT_STATE_DIR }}
```

.github/workflows/c-chain-benchmark.yml
aaronbuchwald: GitHub isolates caches by branch to avoid issues like this: https://docs.github.com/en/actions/reference/dependency-caching-reference#restrictions-for-accessing-a-cache.

A feature branch can access the cache created by its base branch, but sibling branches would not access each other's cache.

Saving the benchmark history to a branch would not avoid this issue (this is currently marked as an immediate follow up to this PR and not included) as it runs into the same issue you called out of conflicts between `workflow_dispatch` and `schedule` where it would write the results from either to the same destination.

To start writing to the branch, we'll need to disable `auto-push` and then make a separate step that decides whether to push the results to the branch iff we are on the master branch.

.github/workflows/c-chain-benchmark.yml
maru-ava: Ugh, my bad. I'm not sure why I assumed this would work.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+	r.NoError(err)
+	tb.Cleanup(func() {
+		r.NoError(db.Close())
+	})
+
+	batch := db.NewBatch()
+	for blkResult := range blockChan {
+		r.NoError(batch.Put(binary.BigEndian.AppendUint64(nil, blkResult.Height), blkResult.BlockBytes))
+
+		if batch.Size() > 10*units.MiB {
```
RodrigoVillar, line 458: Q: why limit the batch size to 10mb?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+	targetBlockDirArg string
+	targetDirArg      string
+	startBlockArg     uint64
+	endBlockArg       uint64
+	chanSizeArg       int
+	metricsEnabledArg bool
+	executionTimeout  time.Duration
+	labelsArg         string
+
+	labels map[string]string = map[string]string{
```
RodrigoVillar: (nit): why the type constraint here? I would think that

```golang
labels = map[string]string{
    // ...
}
```

is sufficient here.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+type vmExecutorConfig struct {
+	Log logging.Logger
+	// Registry is the registry to register the metrics with.
+	Registry prometheus.Registerer
+	// ExecutionTimeout is the maximum timeout to continue executing blocks.
+	// If 0, no timeout is applied. If non-zero, the executor will exit early
+	// WITHOUT error after hitting the timeout.
+	// This is useful to provide consistent duration benchmarks.
+	ExecutionTimeout time.Duration
+}
```
RodrigoVillar: Q: should we get rid of this type? I would say yes for the following reasons:

1. Passing the individual fields as arguments to `newVMExecutor()` would only add two additional parameters (two new + 1 replacement) + two additional fields to `vmExecutor` - I think this is fine IMO.
2. We use the `registry` field only once - if we were to make this an argument, then I don't see the utility of having a config type with two fields.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+			return err
+		}
+
+		select {
+		case <-ctx.Done():
+			e.config.Log.Info("exiting early due to context timeout", zap.Duration("elapsed", time.Since(start)), zap.Duration("execution-timeout", e.config.ExecutionTimeout))
+			return nil
+		default:
+		}
+	}
```
RodrigoVillar: (nit / feel free to ignore): I think the following is cleaner to read:

```golang
for {
    select {
        case blkResult, ok := <-blkChan:
            if !ok { // blkChan is closed
                return nil
            }
            // ...
        case <-ctx.Done():
            // ...
            return nil
    }
}
```

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: I'd prefer to add the config now for ease of use and to make it easier to add further params as needed.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Going to keep the current style

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Good catch

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: 10mb should be a reasonable middle ground. Batching will perform better than writing individual key-value pair updates and we also do not want to write the entire thing in a single batch.

We could experiment with different batch sizes, but it should be sufficient for this change to pick a reasonable middle ground.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Addressed https://github.com/ava-labs/avalanchego/pull/4019/commits/24d5720ae5ddd0b236ee72e58477bffc630fd853

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Thanks for the explanation! This seemed like a magic number to me and I would prefer to see some documentation above the if-statement (e.g. "limiting the batch size to 10mb prevents writing the entire block range in a single batch"), but this is just a nit.

RodrigoVillar (approved):
LGTM

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -0,0 +1,73 @@
+name: C-Chain Re-Execution Benchmark
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      start-block:
+        description: 'The start block for the benchmark.'
+        required: false
+        default: 101
```
StephenButtolph, line 9: Is there a reason why `101` is used rather than `1`? Not important at all, just seems like an odd default.

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -0,0 +1,73 @@
...
+        required: false
+        default: 250000
+      source-block-dir:
+        description: 'The source block directory. Supports S3 directory/zip and local directories.'
+        required: false
+        default: s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb.zip
+      current-state-dir:
+        description: 'The current state directory. Supports S3 directory/zip and local directories.'
+        required: false
+        default: s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100.zip
```
StephenButtolph, line 21: This seems to align with the default start height of `101`, but is there a reason this isn't just the genesis state?

I don't really care to actually change this fwiw, just interested in the rationale

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+		if err := e.execute(ctx, blkResult.BlockBytes); err != nil {
+			return err
+		}
+
+		select {
+		case <-ctx.Done():
+			e.config.Log.Info("exiting early due to context timeout", zap.Duration("elapsed", time.Since(start)), zap.Duration("execution-timeout", e.config.ExecutionTimeout))
+			return nil
+		default:
+		}
```
StephenButtolph: I think
```golang
select {
case <-ctx.Done():
	XXX
default:
}
```
is a bit clunky. It should be replaced with:
```golang
if err := ctx.Err(); err != nil {
	XXX
}
```

```suggestion
		if err := ctx.Err(); err != nil {
			e.config.Log.Info("exiting early due to context timeout",
				zap.Duration("elapsed", time.Since(start)),
				zap.Duration("execution-timeout", e.config.ExecutionTimeout),
				zap.Error(err),
			)
			return nil
		}
```

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get last accepted block: %w", err)
+	}
+	blk, err := e.vm.GetBlock(ctx, blkID)
+	if err != nil {
+		return fmt.Errorf("failed to get last accepted block by blkID %s: %w", blkID, err)
+	}
+
+	start := time.Now()
+	e.config.Log.Info("last accepted block", zap.String("blkID", blkID.String()), zap.Uint64("height", blk.Height()))
```
StephenButtolph:
```suggestion
	e.config.Log.Info("last accepted block",
		zap.Stringer("blkID", blkID),
		zap.Uint64("height", blk.Height()),
	)
```

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+		defer cancel()
+	}
+
+	for blkResult := range blkChan {
+		if blkResult.Err != nil {
+			return blkResult.Err
+		}
+
+		if blkResult.Height%1000 == 0 {
+			e.config.Log.Info("executing block", zap.Uint64("height", blkResult.Height))
```
StephenButtolph: Would it also make sense to include an ETA here? Definitely not required, but could be nice (could also just include a TODO).
```suggestion
			e.config.Log.Info("executing block",
				zap.Uint64("height", blkResult.Height),
			)
```

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+	tb.Cleanup(func() {
+		r.NoError(db.Close())
+	})
+
+	go func() {
+		defer close(ch)
+
+		for i := startBlock; i <= endBlock; i++ {
+			blockBytes, err := db.Get(binary.BigEndian.AppendUint64(nil, i))
+			if err != nil {
```
StephenButtolph: Is this done this way to specifically avoid grabbing an iterator? If so can we document why we don't grab an iterator?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -0,0 +1,517 @@
...
+
+	db, err := leveldb.New(targetDir, nil, logging.NoLog{}, prometheus.NewRegistry())
+	r.NoError(err)
+	tb.Cleanup(func() {
+		r.NoError(db.Close())
+	})
+
+	batch := db.NewBatch()
+	for blkResult := range blockChan {
+		r.NoError(batch.Put(binary.BigEndian.AppendUint64(nil, blkResult.Height), blkResult.BlockBytes))
```
StephenButtolph: A bit torn on this, so feel free to ignore, but it might make more sense to write a function:
```golang
func blockKey(height uint64) []byte {
	return binary.BigEndian.AppendUint64(nil, height)
}
```
I think we only have this in two places, and I doubt we'll ever modify the format of the keys... But just seems like a nice readability improvement.

.github/workflows/c-chain-reexecution-benchmark.yml
aaronbuchwald: Fair question - this is because I wanted to test a range with a non-empty database and therefore it starts with a current state at height 100.

.github/workflows/c-chain-reexecution-benchmark.yml
aaronbuchwald: Yup, this is just because I wanted to test with a non-empty starting state to make sure copying was working as expected.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Sure, will use `timer.EstimateETA(...)`

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Not really, just to make it explicit and I wasn't at all concerned about performance here. Will switch to using an iterator and adding a check that the height key is an exact match for the next height to ensure we fail loudly if there's an unexpected entry or missing entry in the database.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Sure, will add this.

PR #4020, opened 2025-06-18
Remove unnecessary chan from metrics server start

aaronbuchwald:

### DIFF
--- a/tests/load/prometheus.go
+++ b/tests/load/prometheus.go
@@ -38,7 +38,7 @@ func (*MetricsServer) String() string {
 	return "metrics server"
 }
 
-func (s *MetricsServer) Start() (runError <-chan error, err error) {
+func (s *MetricsServer) Start() (<-chan error, error) {
 	const metricsPattern = "/ext/metrics"
 
 	mux := http.NewServeMux()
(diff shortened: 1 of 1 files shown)


tests/load/prometheus.go
```diff
@@ -58,23 +58,21 @@ func (s *MetricsServer) Start() (runError <-chan error, err error) {
 		ReadTimeout:       time.Second,
 	}
 
-	runErrorBiDirectional := make(chan error)
```
aaronbuchwald, line 61: If we just return the original channel, there's no need to create a second variable for it

tests/load/prometheus.go
```diff
@@ -58,23 +58,21 @@ func (s *MetricsServer) Start() (runError <-chan error, err error) {
...
-	runError = runErrorBiDirectional
-	ready := make(chan struct{})
+	errChan := make(chan error)
 	go func() {
-		close(ready)
```
aaronbuchwald, line 65: It's not clear to me why we would create a channel just to wait for the goroutine to close it. It seems this is to ensure the goroutine has started before we return? Maybe this is trying to guarantee that we can reach the server as soon as this function returns, but that would still be race-y given we close the channel before starting the server in the goroutine.

Issue #3935, opened 2025-05-02
Log Antithesis docker compose during generation


StephenButtolph:
The Antithesis docker compose is generated, this means that is isn't trivial to look at what is actually being run in the Antithesis environment.

We should dump the generated docker compose when updating the images.

PR #4021, opened 2025-06-19
Dump generated docker compose during Antithesis image builds

bajpainaman:
#3935 

Previously, the Antithesis docker compose configuration was generated and embedded in the config image, making it difficult to inspect what was actually being run in the Antithesis environment without extracting it from the built image.

This change enhances the `gen_antithesis_compose_config` function to dump the generated docker-compose.yml file during the build process, providing:

- Immediate visibility into the generated compose configuration
- Console output of the complete compose file during builds
- Persistent storage of the compose file in build/antithesis/{setup}/
- Better debugging capabilities for Antithesis test environments

The compose file is now accessible at:
- build/antithesis/avalanchego/docker-compose.yml (for avalanchego setup)
- build/antithesis/xsvm/docker-compose.yml (for xsvm setup)

This improvement maintains backward compatibility while significantly improving the developer experience when working with Antithesis test configurations.


rhkrohan:
@a1k0n I'm having the same issue, and this can be a possible fix. Is there smth wrong with this pr?

maru-ava:
> @a1k0n I'm having the same issue, and this can be a possible fix. Is there smth wrong with this pr?

Sorry, you're having _what_ issue exactly?

PR #4027, opened 2025-06-23
Block Database

DracoLi:
## Why this should be merged

This PR introduces BlockDB, a specialized database optimized for block storage.

Avalanche VMs currently store blocks in a key-value database (LevelDB or PebbleDB). This method is not ideal for block storage because large blocks lead to frequent compactions, resulting in write amplification that degrades performance as the database expands. Additionally, key-value databases are designed for random key-value access, not the sequential patterns typical of blockchain operations.

For how BlockDB works see [README.md](https://github.com/ava-labs/avalanchego/tree/dl/blockdb/x/blockdb).

## Changes

- Added `blockdb` to `x/`.
- ~~Updated our lru cache to support `onEvict`. This is used by the blockdb for storing opened file descriptors for the data files.~~ Moved into separate PR

## How this was tested

Units tests for now

## Todos

- [x] Split data across multiple files when `MaxDataFileSize` is reached


x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+│ - Data Size     │  │      │ Block 1         │
+│ - ...           │  │      │ - Header        │
+├─────────────────┤  │  ┌──>│ - Data          │
+│ Entry[0]        │  │  │   ├─────────────────┤
+│ - Offset ───────┼──┘  │   │     ...         │
+│ - Size          │     │   └─────────────────┘
+│ - Header Size   │     │
+├─────────────────┤     │   ┌─────────────────┐
+│ Entry[1]        │     │   │  Data File 2    │
+│ - Offset ───────┼─────┘   │   (.dat)        │
```
rkuris: This offset points to the header, not to the data, like block 0 did.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+Index File Header (72 bytes):
+┌────────────────────────────────┬─────────┐
+│ Field                          │ Size    │
+├────────────────────────────────┼─────────┤
+│ Version                        │ 8 bytes │
+│ Max Data File Size             │ 8 bytes │
+│ Max Block Height               │ 8 bytes │
+│ Min Block Height               │ 8 bytes │
+│ Max Contiguous Height          │ 8 bytes │
```
rkuris: These 3 are in a different order in the rust implementation, can we swap them here or enter an issue to fix it in rust?

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+│ Data File Size                 │ 8 bytes │
+│ Reserved                       │ 24 bytes│
```
rkuris: Why do we need a reserved area here?

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+└────────────────────────────────┴─────────┘
+
+Index Entry (18 bytes):
+┌────────────────────────────────┬─────────┐
+│ Field                          │ Size    │
+├────────────────────────────────┼─────────┤
+│ Data File Offset               │ 8 bytes │
+│ Block Data Size                │ 8 bytes │
+│ Header Size                    │ 2 bytes │
```
rkuris: We don't need 8 bytes for the block data size, and this should really be a power of 2. Recommend 8/4/4.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+```
+Block Header (26 bytes):
+┌────────────────────────────────┬─────────┐
+│ Field                          │ Size    │
+├────────────────────────────────┼─────────┤
+│ Height                         │ 8 bytes │
+│ Size                           │ 8 bytes │
+│ Header Size                    │ 2 bytes │
+│ Checksum                       │ 8 bytes │
```
rkuris: Recommend height: 8, checksum: 8, size: 4, header_size: 4. This will result in better packing and performance internally.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+└────────────────────────────────┴─────────┘
+```
+
+### Block Overwrites
+
+BlockDB allows overwriting blocks at existing heights. When a block is overwritten, the new block is appended to the data file and the index entry is updated to point to the new location, leaving the old block data as unreferenced "dead" space. However, since blocks are immutable and rarely overwritten (e.g., during reorgs), this trade-off should have minimal impact in practice.
+
+### Fixed-Size Index Entries
+
+Each index entry is exactly 18 bytes on disk, containing the offset, size, and header size. This fixed size enables direct calculation of where each block's index entry is located, providing O(1) lookups. For blockchains with high block heights, the index remains efficient, even at height 1 billion, the index file would only be ~18GB.
```
rkuris: Making this 16 would result in a guarantee that all index entries will fit into a single disk block. With this scheme, some index entries will require two disk reads. See comments above as to how to fix this.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Durability and Fsync Behavior
+
+BlockDB provides configurable durability through the `syncToDisk` parameter:
+
+- When enabled, the data file is fsync'd after every block write, guaranteeing immediate durability
+- The index file is fsync'd periodically (every `CheckpointInterval` blocks) to balance performance and recovery time
+- When disabled, writes rely on OS buffering, trading durability for significantly better performance
```
rkuris: "durability guarantees on kernel or machine failure, but not compromising durability guarantees on process failure"

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+### Fixed-Size Index Entries
+
+Each index entry is exactly 18 bytes on disk, containing the offset, size, and header size. This fixed size enables direct calculation of where each block's index entry is located, providing O(1) lookups. For blockchains with high block heights, the index remains efficient, even at height 1 billion, the index file would only be ~18GB.
+
+### Durability and Fsync Behavior
+
+BlockDB provides configurable durability through the `syncToDisk` parameter:
+
+- When enabled, the data file is fsync'd after every block write, guaranteeing immediate durability
+- The index file is fsync'd periodically (every `CheckpointInterval` blocks) to balance performance and recovery time
```
rkuris: Can you comment on whether the index file is using buffered writes or unbuffered writes? Using buffered writes is probably fine and will result in much better performance, but we should make sure the CheckpointInterval is used to decide how often to flush the index file in all cases.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+- When disabled, writes rely on OS buffering, trading durability for significantly better performance
+
+### Recovery Mechanism
+
+On startup, BlockDB checks for signs of an unclean shutdown. If detected, it performs recovery:
```
rkuris: Can you describe the "signs of an unclean shutdown?"

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+```go
+import "github.com/ava-labs/avalanchego/x/blockdb"
+
+config := blockdb.DefaultDatabaseOptions()
+db, err := blockdb.New(
+    "/path/to/index",  // Index directory
+    "/path/to/data",   // Data directory
+    true,              // Sync to disk
+    false,             // Don't truncate existing data
+    config,
```
rkuris: Why aren't "sync to disk" and "don't truncate" part of config?

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+### Writing and Reading Blocks
+
+```go
+// Write a block with header size
+height := uint64(100)
+blockData := []byte("block data...")
+headerSize := uint16(500) // First 500 bytes are the header
+err := db.WriteBlock(height, blockData, headerSize)
```
rkuris: Shouldn't this fail? The headerSize should never be larger than the size of the blockData.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+// Read a complete block
+blockData, err := db.ReadBlock(height)
+if blockData == nil {
+    // Block doesn't exist at this height
```
rkuris: Or an error occurred. Generally, one should check err first before looking at blockData.

x/blockdb/block.go
```diff
@@ -0,0 +1,359 @@
...
+
+// MaxBlockDataSize is the maximum size of a block in bytes (16 MB).
+const MaxBlockDataSize = 1 << 24
+
+// blockHeader is prepended to each block in the data file.
+type blockHeader struct {
+	Height     BlockHeight
+	Size       uint64
+	HeaderSize BlockHeaderSize
+	Checksum   uint64
```
rkuris: comments above should reorder this, recommend height, checksum, size, header_size.

x/blockdb/block.go
```diff
@@ -0,0 +1,359 @@
...
+	if blockDataLen == 0 {
+		return ErrBlockEmpty
+	}
+
+	if blockDataLen > MaxBlockDataSize {
+		return ErrBlockTooLarge
+	}
+
+	if uint64(headerSize) >= uint64(len(block)) {
+		return ErrHeaderSizeTooLarge
```
rkuris: commented earlier that your example will fail, here is the check for it.

x/blockdb/index.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (s *Database) indexEntryOffset(height BlockHeight) (uint64, error) {
+	if height < s.header.MinHeight {
+		return 0, fmt.Errorf("%w: height %d is less than minimum block height %d", ErrInvalidBlockHeight, height, s.header.MinHeight)
+	}
+	relativeHeight := height - s.header.MinHeight
+
+	// Check for overflow before calculating the final offset.
+	if relativeHeight > (math.MaxUint64-sizeOfIndexFileHeader)/sizeOfIndexEntry {
```
rkuris: This check is not quite sufficient, as it doesn't consider that we both multiply and add. The add can overflow.

I heard we have some checked math operations in golang in avalanchego somewhere we can use.

x/blockdb/block.go
```diff
@@ -0,0 +1,359 @@
...
+	return nil
+}
+
+func (s *Database) allocateBlockSpace(totalSize uint64) (writeDataOffset uint64, err error) {
+	maxDataFileSize := s.header.MaxDataFileSize
+
+	// Check if a single block would exceed the max data file size
+	if maxDataFileSize > 0 && totalSize > maxDataFileSize {
+		return 0, ErrBlockTooLarge
+	}
```
rkuris: This shouldn't be needed, since you shouldn't allow for a maxDataFileSize less than the maximum size of a block.

x/blockdb/block.go
```diff
@@ -0,0 +1,359 @@
...
+			fileIndex := int(currentOffset / maxDataFileSize)
+			localOffset := currentOffset % maxDataFileSize
+
+			if localOffset+totalSize > maxDataFileSize {
+				writeOffset = (uint64(fileIndex) + 1) * maxDataFileSize
+				newOffset = writeOffset + totalSize
+			}
+		}
+
+		if s.nextDataWriteOffset.CompareAndSwap(currentOffset, newOffset) {
```
rkuris: Nice way of doing this! Presumably this is faster in the non-contention case than a mutex?

x/blockdb/block.go
```diff
@@ -0,0 +1,359 @@
...
+		newOffset := currentOffset + totalSize
+
+		if maxDataFileSize > 0 {
+			fileIndex := int(currentOffset / maxDataFileSize)
+			localOffset := currentOffset % maxDataFileSize
+
+			if localOffset+totalSize > maxDataFileSize {
+				writeOffset = (uint64(fileIndex) + 1) * maxDataFileSize
+				newOffset = writeOffset + totalSize
+			}
```
rkuris: I think this means that files other than the first one will not contain a header. Is this intentional? If so, it means the first file is always going to be opened and can never be deleted which should be mentioned in the README.

x/blockdb/database.go
```diff
@@ -0,0 +1,264 @@
...
+		return true
+	})
+}
+
+func (s *Database) dataFilePath(index int) string {
+	return filepath.Join(s.dataDir, fmt.Sprintf(dataFileNameFormat, index))
+}
+
+func (s *Database) getOrOpenDataFile(fileIndex int) (*os.File, error) {
+	if handle, ok := s.fileCache.Load(fileIndex); ok {
```
rkuris: I think we need some limit on the fileCache size, otherwise we could run out of file descriptors if the maxFileSize is pretty small and/or blocks are really big.

x/blockdb/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+- **O(1) Performance**: Both reads and writes complete in constant time
+- **Parallel Operations**: Multiple threads can read and write blocks concurrently without blocking
+- **Flexible Write Ordering**: Supports out-of-order block writes for bootstrapping
+- **Configurable Durability**: Optional `syncToDisk` mode guarantees immediate recoverability
+- **Automatic Recovery**: Detects and recovers unindexed blocks after unclean shutdowns
+
+## Design
+
+BlockDB uses two file types: index files and data files. The index file maps block heights to locations in data files, while data files store the actual block content. Data storage can be split across multiple files based on the maximum data file size.
```
rkuris: There is only one index file and multiple data files. Can you call this out?

x/blockdb/database.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+// Database stores blockchain blocks on disk and provides methods to read, and write blocks.
+type Database struct {
+	indexFile *os.File
+	dataDir   string
+	options   DatabaseConfig
+	header    indexFileHeader
+	log       logging.Logger
+	mu        sync.RWMutex
```
rkuris: A better name or a comment that this is only for ensuring close operates on a database that is not being accessed by other processes would be helpful here.

x/blockdb/README.md
DracoLi: 👍 , i'll fix the order here

x/blockdb/README.md
DracoLi: I wanted to account for we might add features that will require us to store more data in the header in future versions. If this happens, we can add it here without needing to reindex.

x/blockdb/block.go
DracoLi: Yeah, agreed. In practice, setting `maxDataFileSize` smaller than `MaxBlockDataSize` wouldn’t make sense. But is it worth enforcing? One current use case is I'm setting `maxDataFileSize` to a small value in tests to trigger splitting logic ([see here](https://github.com/ava-labs/avalanchego/blob/dl/blockdb/x/blockdb/writeblock_test.go#L325C1-L333C4)).

x/blockdb/block.go
DracoLi: yeah, this should be more lightweight

x/blockdb/block.go
DracoLi: not every block will contain the header (blockSize includes the metadata header + block). We are only splitting the data files here, not the index file. This is just calculating the global next write offset if the current data file cannot fit the block.

x/blockdb/database.go
DracoLi: good idea. i can set a 10k limit

x/blockdb/index.go
DracoLi: for sure, looks like we have `safemath`. I'll update the package to use `safemath` throughout

x/blockdb/database.go
DracoLi: good point, i'll do both

x/blockdb/block.go
DracoLi: The variable names could have been better. I updated this method to be a clearer in terms of what its doing.

x/blockdb/README.md
DracoLi: i think good idea to combine them - will update

x/blockdb/README.md
DracoLi: updated the examples to handle errors

x/blockdb/README.md
```diff
@@ -0,0 +1,190 @@
...
+On startup, BlockDB checks for signs of an unclean shutdown by comparing the data file size on disk with the indexed data size stored in the index file header. If the data files are larger than what the index claims, it indicates that blocks were written but the index wasn't properly updated before shutdown.
+
+**Recovery Process:**
+
+1. Starts scanning from where the index left off (`NextWriteOffset`)
+2. For each unindexed block found:
+   - Validates the block header and checksum
+   - Writes the corresponding index entry
+3. Updates the max contiguous height and max block height
+4. Persists the updated index header
```
yacovm: >    - Writes the corresponding index entry
> 4. Persists the updated index header

This isn't clear. Do you mean you add an index entry in the index file in the former and in the latter you update the header of the index file? Can you clarify?

x/blockdb/database.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+const (
+	indexFileName          = "blockdb.idx"
+	dataFileNameFormat     = "blockdb_%d.dat"
+	defaultFilePermissions = 0o666
+
+	// Since 0 is a valid height, math.MaxUint64 is used to indicate unset height.
+	// It is not be possible for block height to be max uint64 as it would overflow the index entry offset
```
yacovm: s/It is not be possible/It is not possible/

x/blockdb/database.go
DracoLi: also updated to use lru cache

x/blockdb/README.md
```diff
@@ -0,0 +1,190 @@
...
+## Usage
+
+### Creating a Database
+
+```go
+import "github.com/ava-labs/avalanchego/x/blockdb"
+
+config := blockdb.DefaultDatabaseConfig()
+db, err := blockdb.New(
+    "/path/to/index",  // Index directory
```
yacovm: I think that two parameters of the same type is usually confusing. 

Why not merge them into the config struct? Less parameters this way and less chance to conflate the order of the index and data.

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	if err := s.writeIndexEntryAt(indexFileOffset, writeDataOffset, blockDataLen, headerSize); err != nil {
+		return err
+	}
+
+	return s.updateBlockHeights(height)
+}
+
+// ReadBlock retrieves a block by its height.
```
yacovm: I think we should put all instance methods of (Database) in the same file.  It's rather uncommon in Golang to split the methods of a struct across several files.

x/blockdb/database.go
```diff
@@ -0,0 +1,276 @@
...
+	unsetHeight = math.MaxUint64
+)
+
+// Database stores blockchain blocks on disk and provides methods to read, and write blocks.
+type Database struct {
+	indexFile *os.File
+	dataDir   string
+	options   DatabaseConfig
+	header    indexFileHeader
+	log       logging.Logger
```
yacovm, line 188: This is a general comment but I've noticed there is very little logging in this package. 

Is there a reason to not log what we're doing in debug? I think it can be useful if we ever need to debug an issue.

Also I think it makes sense to log errors that are critical such as a corruption we cannot recover from or an error that isn't an EOF.

x/blockdb/index.go
```diff
@@ -0,0 +1,191 @@
...
+	Offset uint64
+	// Size is the length in bytes of the block's data (excluding the blockHeader).
+	Size uint32
+	// HeaderSize is the size in bytes of the block's header portion within the data.
+	HeaderSize BlockHeaderSize
+}
+
+// IsEmpty returns true if this entry is uninitialized.
+// This indicates a slot where no block has been written.
+func (e indexEntry) IsEmpty() bool {
```
yacovm: I'm curious to know why this and the below methods don't have a pointer receiver. 

Is there a reason for that? When you don't have a pointer receiver, Golang creates a shallow copy of the object instead of just referencing it.

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+// Returns nil if the block is not found.
```
yacovm: Why return nil if the block isn't found instead of (nil, [ErrNotFound](https://github.com/syndtr/goleveldb/blob/master/leveldb/db.go#L855-L856)) ?

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+				if actualBlockEndOffset, err = safemath.Add(actualWriteOffset, uint64(totalSize)); err != nil {
+					return 0, fmt.Errorf(
+						"adding block of size %d to new file offset %d would overflow: %w",
+						totalSize, actualWriteOffset, err,
+					)
+				}
+			}
+		}
+
+		if s.nextDataWriteOffset.CompareAndSwap(currentOffset, actualBlockEndOffset) {
```
yacovm: nice trick

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+// file splitting by advancing the nextWriteOffset when a data file would be exceeded.
+//
+// Parameters:
+//   - totalSize: The total size in bytes needed for the block
+//
+// Returns:
+//   - writeDataOffset: The data file offset where the block should be written
+//   - err: Error if allocation fails (e.g., block too large, overflow, etc.)
+func (s *Database) allocateBlockSpace(totalSize uint32) (writeDataOffset uint64, err error) {
+	maxDataFileSize := s.header.MaxDataFileSize
```
yacovm: Another way of supporting splitting the data across files is just setting `maxDataFileSize to maxuint64 if `s.header.MaxDataFileSize` is 0. 

In that case, then `currentFileIndex := int(currentOffset / maxDataFileSize)` is always 0 and `currentOffset % maxDataFileSize` is always `currentOffset`.

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+			// Check if this block would span across file boundaries
+			blockEndWithinFile, err := safemath.Add(offsetWithinCurrentFile, uint64(totalSize))
+			if err != nil {
+				return 0, fmt.Errorf(
+					"calculating block end within file would overflow: %w",
+					err,
+				)
+			}
+			if blockEndWithinFile > maxDataFileSize {
+				// Advance the current write offset to the start of the next file since
```
yacovm: So basically if we cannot fit this block within the space left in the current data file, we start from the next file and leave the remaining space empty, right?

x/blockdb/index.go
```diff
@@ -0,0 +1,191 @@
...
+func (s *Database) persistIndexHeader() error {
+	// The index file must be fsync'd before the header is written to prevent
+	// a state where the header is persisted but the index entries it refers to
+	// are not. This could lead to data inconsistency on recovery.
+	if s.options.SyncToDisk {
+		if err := s.indexFile.Sync(); err != nil {
+			return fmt.Errorf("failed to sync index file before writing header state: %w", err)
+		}
+	}
+
```
yacovm: below you do three atomic reads and then persist the header. Is it not possible that you will read an inconsistent view of the atomic variables because some other goroutine will update one of them in parallel? Since we're not relying on locking, what guarantees us that all three atomic variables are consistent to an execution that makes sense?

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+
+	if s.maxContiguousHeight.CompareAndSwap(prevContiguousCandidate, writtenBlockHeight) {
+		currentMax := writtenBlockHeight
+		for {
+			nextHeightToVerify, err := safemath.Add(currentMax, 1)
+			if err != nil {
+				s.log.Error("overflow in height calculation when updating max contiguous height")
+				break
+			}
+			entry, err := s.readIndexEntry(nextHeightToVerify)
```
yacovm: can you add comments regarding why we are reading the index entry here? I guess we're trying to see if we've indexed this height in the past.

Is this loop to repair gaps in the contiguous height?

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+	// Read the complete block data
+	blockData := make(BlockData, indexEntry.Size)
+	dataFile, localOffset, err := s.getDataFileAndOffset(indexEntry.Offset)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get data file for block at height %d: %w", height, err)
+	}
+	_, err = dataFile.ReadAt(blockData, int64(localOffset+uint64(sizeOfBlockHeader)))
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return nil, nil
```
yacovm: why are we returning `nil, nil` in case of an EOF?

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+
+	return blockData, nil
+}
+
+// ReadHeader retrieves only the header portion of a block by its height.
+// Returns nil if the block is not found or no header.
+func (s *Database) ReadHeader(height BlockHeight) (BlockData, error) {
+	s.closeMu.RLock()
+	defer s.closeMu.RUnlock()
+
```
yacovm: would it make sense to extract:

```
	if s.closed {
		return nil, ErrDatabaseClosed
	}

	indexEntry, err := s.readIndexEntry(height)
	if err != nil {
		return nil, err
	}
	if indexEntry.IsEmpty() {
		return nil, nil
	}
```

to a separate function and just call it in ReadBlock, ReadHeader, ReadBody ?

x/blockdb/database.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
+
+	s := &Database{
+		options:   config,
+		log:       log,
+		fileCache: lru.NewCache[int, *os.File](MaxDataFiles),
```
yacovm: The default size of a data file is 0.5TB and we have an LRU of `MaxDataFiles` which is 10,000.

It's nice that we're supporting a ledger of up to 5 Petabyte bytes, however - why do we need to hold file descriptors to all 5 Peta bytes at the same time? I think we should have at most 10 items in the LRU by default, which is 5TB of data. 

Thoughts?

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+// BlockData defines the type for block data.
+type BlockData = []byte
+
+// BlockHeaderSize is the size of the header in the block data.
+type BlockHeaderSize = uint32
+
+// MaxBlockDataSize is the maximum size of a block in bytes (16 MB).
+const MaxBlockDataSize = 1 << 24
+
+// blockHeader is prepended to each block in the data file.
```
yacovm: maybe call it blockEntryHeader? Because block header refers to the header of the blockchain block, which we already have an API method to fetch. 

This is confusing, for example: 

```
dataFile.ReadAt(headerData, int64(localOffset+uint64(sizeOfBlockHeader)))
```

Is it about the header of the block entry or the block header of the blockchain block? A bit confusing.

x/blockdb/block.go
```diff
@@ -0,0 +1,419 @@
...
+// BlockHeight defines the type for block heights.
+type BlockHeight = uint64
+
+// BlockData defines the type for block data.
+type BlockData = []byte
+
+// BlockHeaderSize is the size of the header in the block data.
+type BlockHeaderSize = uint32
+
+// MaxBlockDataSize is the maximum size of a block in bytes (16 MB).
```
yacovm: I understand why a blockchain system would have a maximum block size.

I don't understand why this block store implementation needs a maximum block size. What if the rest of the system allows a block of 20MB and then we would agree (consensus) on a block we cannot put into the block store? (I know this is a theoretic question as I think we support blocks of up to roughly 2MB, but asking for the sake of correctness) 

In my opinion, if we don't need a maximum block size for the implementation, we can do without it. 

In the place where we check if the block is too large, we can just use the block data file size.

x/blockdb/README.md
```diff
@@ -0,0 +1,190 @@
...
+│ Data File Offset               │ 8 bytes │
+│ Block Data Size                │ 4 bytes │
+│ Header Size                    │ 4 bytes │
+└────────────────────────────────┴─────────┘
+```
+
+#### Data File Structure
+
+Each block in the data file is stored with a header followed by the raw block data:
+
```
yacovm, line 76: would it not make sense to add a version field? Imagine that in the future we'll need to change the structure of the data file entry. 

If you have a node with the old version, they'll want to migrate their data, but it's not clear where the header of the block entry starts and where the data (block header and body) parts start. Maybe spare 2 bytes for a version? 
Alternatively we can put 2 bytes to say where the header of the block entry ends.

x/blockdb/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+Block Header (24 bytes):
```
yacovm: Block entry header?

x/blockdb/database.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	dataFiles := make(map[int]string)
+	maxIndex := -1
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+		var index int
+		if n, err := fmt.Sscanf(file.Name(), dataFileNameFormat, &index); n == 1 && err == nil {
```
yacovm: what if `err != nil` ? should we log this?

x/blockdb/database.go
```diff
@@ -0,0 +1,276 @@
...
+		return fmt.Errorf("failed to create data directory %s: %w", dataDir, err)
+	}
+
+	if truncate {
+		dataFiles, _, err := s.listDataFiles()
+		if err != nil {
+			return fmt.Errorf("failed to list data files for truncation: %w", err)
+		}
+		for _, filePath := range dataFiles {
+			if err := os.Remove(filePath); err != nil {
```
yacovm: I'm not sure I understand why we're removing data files. Can you explain the rational?

x/blockdb/recovery.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}
+	} else if len(dataFiles) > 1 || maxIndex > 1 {
+		return fmt.Errorf("%w: expect only 1 data file at index 0, got %d files with max index %d", ErrCorrupted, len(dataFiles), maxIndex)
+	}
+
+	// Calculate the expected next write offset based on the data on disk.
+	var calculatedNextDataWriteOffset uint64
+	if s.header.MaxDataFileSize > 0 {
+		// All data files before the last should be full.
```
yacovm: > // All data files before the last should be full.

What if we have written a block that didn't fit into a data file and opened a new one?

x/blockdb/recovery.go
```diff
@@ -0,0 +1,214 @@
...
+	// Write index entry for this block
+	indexFileOffset, idxErr := s.indexEntryOffset(bh.Height)
+	if idxErr != nil {
+		return bh, fmt.Errorf("cannot get index offset for recovered block %d: %w", bh.Height, idxErr)
+	}
+	if err := s.writeIndexEntryAt(indexFileOffset, offset, bh.Size, bh.HeaderSize); err != nil {
+		return bh, fmt.Errorf("failed to update index for recovered block %d: %w", bh.Height, err)
+	}
+
+	if err := s.updateBlockHeights(bh.Height); err != nil {
```
yacovm: do we really need to do this each time for each block? Can't we do it after we scan all block entries?

yacovm (commented):
Made a first pass, great work.

x/blockdb/recovery.go
DracoLi: offset should still be correct in this case. We are not calculating the total size but the next write offset. I will update the comment here so its clearer.

x/blockdb/database.go
DracoLi: yeah truncate means we are resetting all data. So removing index and existing data files.

x/blockdb/block.go
DracoLi: not exactly leaving the remaining space empty since the data file will just be however large the data is. So it will usually be less than `maxDataFileSize`. 
The offset is tracked as a global offset and adds `MaxDataFileSize` if a data file is done/full, so we know the location of the next file and offset.

x/blockdb/README.md
DracoLi: 👍 , i think that works well. i can move it to the config.

x/blockdb/database.go
DracoLi: yeah logging could be greatly improved in this package. I'll make a pass to add more logs.

x/blockdb/index.go
DracoLi: the struct is pretty small so i didn't think the copy performance should be impactful. I didn't opt to use pointer here just to be clear we are not modifying the struct. I don't feel strongly either way - do you think its preferred to just use pointers for all the methods?

x/blockdb/block.go
DracoLi: Thinking through again, I think we should probably make this consistent with how `GetBlock` works in snowman - it returns `database.ErrNotFound` when the block isn’t found.

Originally, I didn’t do that because I didn't see a missing block as an actual error, compared to things like corruption or I/O issues during read. Clients will need to check for `NotFound` via `errors.Is` in many cases to handle the errors differently.

But I think that returning an explicit not found error makes the API clearer, I can update this.

x/blockdb/block.go
DracoLi: great point. I think thats better versus needing to handle the case where 0 is unlimited. I'll update.

x/blockdb/index.go
DracoLi: yeah for sure - the three atomics can be inconsistent. But I think that might be okay in this case.
We read `NextWriteOffset` first, so any inconsistency means `MaxContiguousHeight` or `MaxBlockHeight` is larger than what `NextWriteOffset` implies. in the worst case, we crash with a inconsistent header, recovery should reindexes everything, so the atomic states should be consistent on next startup.
As for block reads and writes, they use the atomic vars directly, so those accesses aren’t affected by the potential header mismatch.

x/blockdb/block.go
DracoLi: yep thats exactly right. I'll add some more comments to the code here so its clearer for readers

x/blockdb/block.go
DracoLi: you are right, we don't need the EOF check when reading both data and index file as the offset should be always within file size. I'll take these out.

x/blockdb/block.go
DracoLi: agreed, will clean this up a bit

x/blockdb/recovery.go
DracoLi: I wanted to reuse `updateBlockHeights` here for simplicity but yeah I agree, we should do it after recovery scan and we don't need the concurrency handling in `updateBlockHeights`

x/blockdb/database.go
DracoLi: its expected to have non data files in the folder so we are expected to errors here. maybe a debug log?

x/blockdb/README.md
DracoLi: 👍  - I think storing a version here makes sense. I can added that in.

x/blockdb/block.go
DracoLi: yeah i agree, i think the store can be flexible and not have a constraint on this. We are also using uint32 to store the block size so that naturally will give it a 4GB max limit per block. Plus we are also already validating that the block size, including block entry header, can fit into one data file. 
i'll remove this.

x/blockdb/block.go
DracoLi: 👍, i think `blockEntryHeader` is a better name. I'll refactor and also improve the comment.

x/blockdb/database.go
DracoLi: this is to accommodate for when MaxDataFileSize is set to a small value for some reason. 
Hows making this part of config and set the default to 10 to match the default max data file size.

x/blockdb/index.go
yacovm: If we really want to use a lock-free method, can't we create a struct that represents all three fields and store it in an `atomic.Pointer`, and then use compare&swap to update all three values atomically? This way we never write an inconsistent value to the disk. 

```
headerInfo atomic.Pointer[headerInfo]

...

type headerInfo struct {
	// maxBlockHeight tracks the highest block height that has been written to the db, even if there are gaps in the sequence.
	maxBlockHeight uint64
	// nextDataWriteOffset tracks the next position to write new data in the data file.
	nextDataWriteOffset uint64
	// maxContiguousHeight tracks the highest block height known to be contiguously stored.
	maxContiguousHeight uint64
}

```

DracoLi: discussed offline. `nextDataWriteOffset` is not expected to be consistent with the heights and the header on disk will at worst be behind so recovery should fix the issues.
Will move `maxBlockHeight` and `maxContiguousHeight` into a struct so they are consistent with each other

x/blockdb/database.go
yacovm: Why would we want to do that?

x/blockdb/database.go
```diff
@@ -0,0 +1,1337 @@
...
+			MaxContiguousHeight: unsetHeight,
+			NextWriteOffset:     0,
+		}
+		s.setBlockHeights(unsetHeight, unsetHeight)
+
+		headerBytes, err := s.header.MarshalBinary()
+		if err != nil {
+			return fmt.Errorf("failed to serialize new header: %w", err)
+		}
+		if uint64(len(headerBytes)) != sizeOfIndexFileHeader {
```
yacovm, line 1023: We initialzie a buffer of `sizeOfIndexFileHeader` inside of `MarshalBinary` and then outside we check if it's equal to the size. How can it not be?

x/blockdb/database.go
```diff
@@ -0,0 +1,1337 @@
...
+	return nil
+}
+
+func (s *Database) loadOrInitializeHeader() error {
+	fileInfo, err := s.indexFile.Stat()
+	if err != nil {
+		return fmt.Errorf("failed to get index file stats: %w", err)
+	}
+
+	// reset index file if its empty or we are truncating
```
yacovm: > or we are truncating

can you explain when we would want to truncate?

x/blockdb/config.go
```diff
@@ -0,0 +1,128 @@
...
+	// MaxDataFiles is the maximum number of data files descriptors cached.
+	MaxDataFiles int
+
+	// CheckpointInterval defines how frequently (in blocks) the index file header is updated (default: 1024).
+	CheckpointInterval uint64
+
+	// SyncToDisk determines if fsync is called after each write for durability.
+	SyncToDisk bool
+
+	// Truncate determines if existing data should be truncated when opening the database.
```
yacovm: I don't think this option is something we would ever use in a production setting, so probably we should remove it.

x/blockdb/database.go
```diff
@@ -0,0 +1,1337 @@
...
+
+	return nil
+}
+
+func (s *Database) closeFiles() {
+	if s.indexFile != nil {
+		s.indexFile.Close()
+	}
+	if s.fileCache != nil {
+		s.fileCache.Flush()
```
yacovm, line 1072: // closes the data files

x/blockdb/database.go
```diff
@@ -0,0 +1,1337 @@
...
+		if err := dataFile.Sync(); err != nil {
+			return fmt.Errorf("failed to sync data file after writing block %d: %w", bh.Height, err)
+		}
+	}
+	return nil
+}
+
+func (s *Database) updateBlockHeights(writtenBlockHeight BlockHeight) error {
+	s.updateBlockHeightsAtomically(func(current *blockHeights) *blockHeights {
+		if current == nil {
```
yacovm: can `current` be nil? I thought `s.getBlockHeights()` never returns nil.

x/blockdb/database.go
```diff
@@ -0,0 +1,1337 @@
...
+			heights := &blockHeights{
+				maxBlockHeight:      writtenBlockHeight,
+				maxContiguousHeight: unsetHeight,
+			}
+			if writtenBlockHeight == s.header.MinHeight {
```
yacovm: I don't understand why this is true. Can you elaborate? 

If we're writing the minimum height then the max contiguous height is what we write?

What if we wrote blocks 0, 2, 3, ..., 100 and then we write block 1. 
The min height is 1, since the genesis block always exists, but once we write block 1 our max contiguous height is 101, no?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+			updated.maxBlockHeight = writtenBlockHeight
+		}
+
+		// Update max contiguous height logic
+		prevContiguousCandidate := uint64(unsetHeight)
+		if writtenBlockHeight > s.header.MinHeight {
+			prevContiguousCandidate = writtenBlockHeight - 1
+		}
+
+		if current.maxContiguousHeight == prevContiguousCandidate {
```
yacovm, line 1167: if we could get rid of `maxContiguousHeight` we can save this while big `if` statement.

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+
+		// Update max block height if needed
+		if writtenBlockHeight > current.maxBlockHeight || current.maxBlockHeight == unsetHeight {
+			updated.maxBlockHeight = writtenBlockHeight
+		}
+
+		// Update max contiguous height logic
+		prevContiguousCandidate := uint64(unsetHeight)
+		if writtenBlockHeight > s.header.MinHeight {
+			prevContiguousCandidate = writtenBlockHeight - 1
```
yacovm, line 1164: can you explain this?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+				nextHeightToVerify, err := safemath.Add(currentMax, 1)
+				if err != nil {
+					s.log.Error("Failed to update block heights: overflow in height calculation",
+						zap.Uint64("currentMax", currentMax),
+						zap.Error(err),
+					)
+					break
+				}
+				// Check if we have indexed a block at the next height, which would extend our contiguous sequence
+				_, err = s.readIndexEntry(nextHeightToVerify)
```
yacovm, line 1181: This method is used in each insert. We're running a loop here and iterating over blocks until we hit a gap. It seems to me this is quite the overhead to just maintain the `maxContiguousHeight` which as far as I understood is not essential.

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+			}
+			updated.maxContiguousHeight = currentMax
+		}
+
+		return updated
+	})
+
+	// Check if we need to persist header on checkpoint interval
+	if writtenBlockHeight%s.config.CheckpointInterval == 0 {
+		if err := s.persistIndexHeader(); err != nil {
```
yacovm, line 1205: can this be called in parallel if we get a write of height `h` and a write of height `h'` such as `h=checkpointerInterval * k` and `h'=checkpointinterval * m` for some `k != m` ? 

If so, then I'm not sure that the above atomic compare-and-swap helps us, because the writes can be preempted in such a way that the write to the disk will write stale data, no?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+		return nil, err
+	}
+
+	// Return nil if there's no header data
+	if indexEntry.HeaderSize == 0 {
+		return nil, nil
+	}
+
+	// Validate header size doesn't exceed total block size
+	if indexEntry.HeaderSize > indexEntry.Size {
```
yacovm, line 523: Seems like the negative/error paths below are not covered in tests, is it possible to add a test for them?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+	}
+
+	bh := blockEntryHeader{
+		Height:     height,
+		Size:       blockDataLen,
+		HeaderSize: headerSize,
+		Checksum:   calculateChecksum(block),
+		Version:    BlockEntryVersion,
+	}
+	if err := s.writeBlockAt(writeDataOffset, bh, block); err != nil {
```
yacovm, line 384: Seems the below error paths aren't covered in tests. Is it possible to test them?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+	if err != nil {
+		return fmt.Errorf("calculating file size contribution would overflow: %w", err)
+	}
+	calculatedNextDataWriteOffset = fileSizeContribution
+
+	lastFileInfo, err := os.Stat(dataFiles[maxIndex])
+	if err != nil {
+		return fmt.Errorf("failed to get stats for last data file %s: %w", dataFiles[maxIndex], err)
+	}
+	calculatedNextDataWriteOffset, err = safemath.Add(calculatedNextDataWriteOffset, uint64(lastFileInfo.Size()))
```
yacovm, line 786: I don't understand why we're doing it like this instead of just summing all the file sizes. 
If we have a file in the middle that we can't write to its full extent and we open a new file, the file size will be smaller than the maximum file size, and this computation won't be accurate anymore, or am I missing something?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+		return fmt.Errorf("failed to write index entry: %w", err)
+	}
+	return nil
+}
+
+func (s *Database) persistIndexHeader() error {
+	// The index file must be fsync'd before the header is written to prevent
+	// a state where the header is persisted but the index entries it refers to
+	// are not. This could lead to data inconsistency on recovery.
+	if s.config.SyncToDisk {
```
yacovm, line 711: why do we fsync before updating the index file and not after?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+				currentMax = nextHeightToVerify
+			}
+			updated.maxContiguousHeight = currentMax
+		}
+
+		return updated
+	})
+
+	// Check if we need to persist header on checkpoint interval
+	if writtenBlockHeight%s.config.CheckpointInterval == 0 {
```
yacovm, line 1204: wondering if it makes more sense to persist not based on the block height but instead based on iteration? Something like:

```
	var checkpointIteration atomic.Uint64
	
	if checkpointIteration.Add(1) % s.config.CheckpointInterval == 0 {
		
	}
```

This way we are independent of the input and do it more uniformly.

x/blockdb/database.go
DracoLi: discussed separately, will be removing this

x/blockdb/database.go
DracoLi: this should never happen, i'll remove this check

x/blockdb/database.go
DracoLi: removed

x/blockdb/database.go
DracoLi: no, we can remove this check

x/blockdb/database.go
DracoLi: There is no concept of the genesis block here. If min height is set to 1, you cannot write block 0. We can change that behavior but I don't see how is this needed right now.

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+
+	nextDataWriteOffset := s.nextDataWriteOffset.Load()
+	switch {
+	case calculatedNextDataWriteOffset == nextDataWriteOffset:
+		s.log.Debug("Recovery: data files match index header, no recovery needed.")
+		return nil
+
+	case calculatedNextDataWriteOffset < nextDataWriteOffset:
+		// this happens when the index claims to have more data than is actually on disk
+		return fmt.Errorf("%w: index header claims to have more data than is actually on disk "+
```
yacovm, line 799: does this actually imply that when we write the index header to disk it's atomic?

x/blockdb/database.go
DracoLi: if we write block 10, we check if the current maxContiguousHeight is written height - 1 (so 9). If it is we can extend the height. 
However if we write the min height, the current maxContiguousHeight cannot be min height - 1 since we don't allow writing a block below min height, so unset is used.

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+			"(calculated: %d bytes, index header: %d bytes)",
+			ErrCorrupted, calculatedNextDataWriteOffset, nextDataWriteOffset)
+	default:
```
yacovm, line 802: seems like this default case can be extracted to a new function as it's pretty big

x/blockdb/database.go
DracoLi: yes, i think important to note this. Taking  `maxContiguousHeight` out will make implementation simpler and reduce some resource usage.
I think we can evaluate and take `maxContiguousHeight` out if we have no plans to use it. But I don't think it needs to be in this PR. We can test the blockdb with an vm first before making a decision to see if we should remove it as a follow up.

x/blockdb/database.go
DracoLi: yes, checkpoint on height is not very good during bootstrapping. I've considered this as well but I think it can be added as a future follow up. wdyt?

x/blockdb/database.go
DracoLi: @rkuris had a good explaination on this https://github.com/ava-labs/blockstore/blob/main/src/store.rs#L478

x/blockdb/database.go
DracoLi: it will be accurate because the offset does not equal to the total file size. The offset is calculated by number of files - 1 and the data offset of the last file. There are a couple of tests in recovery_test that tests for this scenario.

x/blockdb/database.go
DracoLi: yes, interval checkpoint should resolve this though. the height based interval is not very good during bootstraping when heights written is not sequential. 
I don't think the CAS is related to the checkpoint? Its for handling concurrent updates to height

x/blockdb/database.go
DracoLi: not sure what you mean by that, can you clarify?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+			zap.Uint64("maxContiguousBlockHeight", heights.maxContiguousHeight),
+			zap.Uint64("maxBlockHeight", heights.maxBlockHeight),
+		)
+	}
+	return nil
+}
+
+func (s *Database) recoverBlockAtOffset(offset, totalDataSize uint64) (blockEntryHeader, error) {
+	var bh blockEntryHeader
+	if totalDataSize-offset < uint64(sizeOfBlockEntryHeader) {
```
yacovm, line 878: This probably happens when we opened a new file and crashed, right?

So why return an error here instead of just truncating the file and continuing?

DracoLi: For this error, I think it can only happen if we wrote a partial block due to crashing during write.
I think we can likely just truncate the data file if at any point we failed the `recoverBlockAtOffset` call. we will just stop if we cannot recover anymore and truncate the rest of the data files. thoughts?

x/blockdb/database.go
```diff
@@ -0,0 +1,1319 @@
...
+	blockData := make([]byte, bh.Size)
+	blockDataOffset, err := safemath.Add(localOffset, uint64(sizeOfBlockEntryHeader))
+	if err != nil {
+		return bh, fmt.Errorf("calculating block data offset would overflow at offset %d: %w", offset, err)
+	}
+	if _, err := dataFile.ReadAt(blockData, int64(blockDataOffset)); err != nil {
+		return bh, fmt.Errorf("%w: failed to read block data at offset %d: %w", ErrCorrupted, offset, err)
+	}
+	calculatedChecksum := calculateChecksum(blockData)
+	if calculatedChecksum != bh.Checksum {
```
yacovm, line 928: If the checksum is wrong, why do we return error and fail the entire startup instead of truncating the block records from here after?

x/blockdb/database.go
yacovm: I think it really depends if it's the last file or not. If it's not the last file then something is messed up with the storage, and then we should fail.

However if this is the last file, and all other files and blocks before this block are fine, then it makes sense to truncate.

x/blockdb/database.go
yacovm: I'm personally fine if you want to add it in the future, maybe it makes sense to open an issue and assign it to you so it won't get forgotten, because I will sure forget this in the next morning

x/blockdb/database.go
yacovm: can it be that we somehow crashed mid write and as a result the header now contains data that does not make sense?

x/blockdb/database.go
DracoLi: mainly for simplicity but if any data in the data file is wrong it could mean the data files are corrupted somehow by external factors. In this case, is it better to just fail the node versus assuming the indexed data files are still fine?

x/blockdb/database.go
yacovm: > yes, interval checkpoint should resolve this though.

Can you elaborate? I'm not sure I follow

x/blockdb/database.go
yacovm: Can you open an issue for tracking this?

x/blockdb/database.go
DracoLi: hmm, i think that makes sense. If we failed to recover data from the last file it could mean the os crashed and we can truncate and continue. But if any other data file is corrupted, then its likely it got corrupted from external factors and we error? 
Can't multiple file writes be buffered and not written before the os crashes?

x/blockdb/database.go
DracoLi: yeah, that can happen. right now we just error but one other thing we can do here is just truncate the index file and reindex.

x/blockdb/database.go
yacovm: ah you're right, it stems from the implementation of `getDataFileAndOffset`.

x/blockdb/database.go
DracoLi: i meant if we change checkpoint logic to checkpoint on blocks written instead of height, then we shouldn't encounter two checkpoints in parallel

x/blockdb/database.go
yacovm: Yeah that's a good point. I think for now, let's keep it as is, and open an issue to revisit this?

x/blockdb/database.go
yacovm: yeah, I was referring to the case if that was the last record written.

x/blockdb/database.go
DracoLi: https://github.com/ava-labs/avalanchego/issues/4074

x/blockdb/database.go
yacovm: yes

x/blockdb/database.go
DracoLi: Good idea, created an issue - https://github.com/ava-labs/avalanchego/issues/4075

x/blockdb/database.go
yacovm: Yeah that's a good idea to have re-indexing ability. Let's open an issue? 😉

x/blockdb/database.go
yacovm: I would say it depends on the checkpoint interval and on the amount of concurrent writes...

x/blockdb/database.go
yacovm: makes sense, let's do it.

x/blockdb/database.go
DracoLi: https://github.com/ava-labs/avalanchego/issues/4076

x/blockdb/database.go
DracoLi: yeah, it doesn't prevent it but just make it less likely during normal operations. Any issues you can see if we have concurrent writes of the header? Considering the index header itself is fine to be stale.

x/blockdb/database.go
DracoLi: :+1: https://github.com/ava-labs/avalanchego/issues/4077

x/blockdb/database.go
yacovm: I don't know what happens if we have concurrent writes of the header. I would suggest we don't test it. However since we scan the header when we recover and we don't have to write the actual most updated value, we can just restrict only one right happening at a time:

```
	if occupied.CompareAndSwap(false, true) {
		s.persistIndexHeader()
		occupied.Store(false)
	}
```

x/blockdb/block.go
DracoLi: update: maximum block size is no longer enforced. Instead we are using uint32 so it will be around the max of data file size or `math.MaxUint16` (4GB).
See https://github.com/ava-labs/avalanchego/pull/4027#discussion_r2186122053

x/blockdb/database.go
DracoLi: yeah i like this solution. I'll update

x/blockdb/database_test.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+func TestIndexEntrySizePowerOfTwo(t *testing.T) {
+	// Check that sizeOfIndexEntry is a power of 2
+	// This is important for memory alignment and performance
+	require.Equal(t, uint64(0), sizeOfIndexEntry&(sizeOfIndexEntry-1),
+		"sizeOfIndexEntry (%d) is not a power of 2", sizeOfIndexEntry)
+}
+
+func TestNew_IndexFileConfigPrecedence(t *testing.T) {
```
yacovm, line 185: This test raises a good point - shouldn't we log an info in case the file config differs from the software config? Otherwise we will get weird behavior: We first log the software config and then proceed to overwrite it silently.

DracoLi: yeah good idea, we should log a ~~warning~~ info

yacovm:
I think we can merge it but we still should address and implement the issues opened before we actually let users activate this.

PR #4029, opened 2025-06-23
Proposervm API

cam-schultz:
## Why this should be merged
Adds a path `/proposervm` to all ProposerVM chains (C, P, subnet-evm) that exposes the following methods:
- `proposervm.getProposedHeight`
- `proposervm.getProposerBlockWrapper`

This API will also be useful for exposing P-Chain epoch views if [ACP-181](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/181-p-chain-epoched-views) is adopted.

## How this works

Overrides `CreateHandlers` in `proposervm.VM` to call the inner VM's `CreateHandlers` method, and register the `/proposervm` endpoint.

## How this was tested
Manual testing on a local network. Meaningful unit tests would likely require refactoring `package platformvm` to expose internal structs that are used to [setup a concrete VM](https://github.com/ava-labs/avalanchego/blob/366868c34bc4bced103354c6368dde2a9a02a084/vms/platformvm/vm_test.go#L126). Happy to take suggestions on alternatives.

Yes


vms/proposervm/service.go
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
cam-schultz: The lint job did not like including `2025` in the copyright date, nor did I find any instances of `2025` in the codebase, so I assume it is alright to add this file with a terminal date of `2024`.

vms/proposervm/block/block.go
```diff
@@ -43,16 +43,16 @@ type SignedBlock interface {
...
+	Timestamp    int64  `serialize:"true" json:"timestamp"`
+	PChainHeight uint64 `serialize:"true" json:"pChainHeight"`
+	Certificate  []byte `serialize:"true" json:"certificate"`
+	Block        []byte `serialize:"true" json:"block"`
 }
 
 type statelessBlock struct {
-	StatelessBlock statelessUnsignedBlock `serialize:"true"`
-	Signature      []byte                 `serialize:"true"`
+	StatelessBlock statelessUnsignedBlock `serialize:"true" json:"block"`
```
cam-schultz: `statelessBlock` as the JSON tag seemed more confusing than `block`, but happy to update this if others disagree.

vms/proposervm/service.go
```diff
@@ -0,0 +1,66 @@
...
+		zap.String("method", "getProposedHeight"),
+	)
+
+	reply.Height = avajson.Uint64(p.vm.lastAcceptedHeight)
+	return nil
+}
+
+// GetProposerBlockArgs is the parameters supplied to the GetProposerBlockWrapper API
+type GetProposerBlockArgs struct {
+	ProposerID ids.ID              `json:"proposerID"`
```
cam-schultz: I added this type rather than using the existing `api.GetBlockArgs` to eliminate confusion over what exactly this parameter is. It is the "Proposer ID" listed in the subnets explorer, e.g. https://subnets.avax.network/c-chain/block/64357262, _not_ the block hash/ID as seen by the inner VM.

That said, I'm not sure what the best way to actually source the "Proposer ID" in the general case, since it's not exposed by the innerVM. Similarly, an endpoint to map Block hash -> Proposer ID seems like it would break encapsulation.

vms/proposervm/service.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting"
+
+	avajson "github.com/ava-labs/avalanchego/utils/json"
+)
+
+type ProposerAPI struct {
+	vm *VM
+}
+
```
yacovm, line 20: I think we need to lock the VM prior to accessing its state

cam-schultz: Thanks, fixed

vms/proposervm/block/block.go
ceyonur: IMO we should use `statelessBlock` otherwise there will be 2 `block` tags in the same JSON element.

vms/proposervm/service.go
ceyonur: The word `Proposer` sounds like the actual proposer node. But if that's the common usage, I think it's fine. Maybe `ProposerBlockID` etc can make it cleaner. I think the actual word for that is `OuterBlockID` but sounds even more complicated.

cam-schultz: The explorer uses `Proposer ID`/`Proposer Node ID` to disambiguate, but yes I agree the naming is awkward. Happy to go with whatever the consensus here is.

vms/proposervm/vm.go
```diff
@@ -251,6 +254,25 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+	server.RegisterCodec(json.NewCodec(), "application/json;charset=UTF-8")
+	err = server.RegisterService(&ProposerAPI{vm}, "proposervm")
+	if err != nil {
+		return nil, fmt.Errorf("failed to register proposervm service: %w", err)
+	}
+	handlers["/proposervm"] = server
+
+	return handlers, nil
+}
+
```
yacovm, line 279: what about a unit test that invokes this API?

vms/proposervm/vm.go
```diff
@@ -251,6 +254,25 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+func (vm *VM) CreateHandlers(ctx context.Context) (map[string]http.Handler, error) {
+	handlers, err := vm.ChainVM.CreateHandlers(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create inner VM handlers: %w", err)
+	}
+
+	server := rpc.NewServer()
+	server.RegisterCodec(json.NewCodec(), "application/json")
+	server.RegisterCodec(json.NewCodec(), "application/json;charset=UTF-8")
+	err = server.RegisterService(&ProposerAPI{vm}, "proposervm")
```
yacovm: Why don't we have [these two](https://github.com/ava-labs/avalanchego/blob/507e6bbb5e7d166749fa0b5633e26fd36b753064/vms/platformvm/vm.go#L443-L444) also similar to the platformVM?

```
	server.RegisterInterceptFunc(vm.metrics.InterceptRequest)
	server.RegisterAfterFunc(vm.metrics.AfterRequest)
```

Any thoughts about this?

vms/proposervm/client.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting"
+	"github.com/ava-labs/avalanchego/utils/rpc"
+)
```
yacovm, line 14: Can we add unit tests for this file?

vms/proposervm/service.go
```diff
@@ -0,0 +1,70 @@
...
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting"
+
+	avajson "github.com/ava-labs/avalanchego/utils/json"
+)
+
+type ProposerAPI struct {
```
yacovm, line 17: What is the reason for having the proposerVM as a field instead of just adding the functions to the proposerVM VM? 

If we wish to isolate the VM object then it makes sense to add two functions or (an) interface(s) with:

1. getLastAcceptedHeight
2. GetBlock 

and then it'll be easier to test.

vms/proposervm/client.go
```diff
@@ -0,0 +1,52 @@
...
+
+var _ Client = (*client)(nil)
+
+type Client interface {
```
yacovm: who/what is consuming this interface? Usually in Golang, interfaces are declared where they're used, not where they're implemented.

vms/proposervm/service.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
yacovm: can we add tests to the code in this file?

vms/proposervm/service.go
```diff
@@ -0,0 +1,70 @@
...
+	p.vm.ctx.Log.Debug("API called",
+		zap.String("service", "proposervm"),
+		zap.String("method", "getProposerBlockWrapper"),
+		zap.String("proposerID", args.ProposerID.String()),
+		zap.String("encoding", args.Encoding.String()),
+	)
+	p.vm.ctx.Lock.Lock()
+	defer p.vm.ctx.Lock.Unlock()
+
+	block, err := p.vm.GetBlock(r.Context(), args.ProposerID)
```
yacovm: am I reading this wrong and we're actually sending back the entire proposerVM block which also contains the inner VM block? If so, is it possible to only distill the proposerVM bytes?

vms/proposervm/client.go
ylg-avalabs: Added unit tests.

vms/proposervm/service.go
ylg-avalabs: Added unit tests

vms/proposervm/service.go
ylg-avalabs: Thank you for the suggestion. Created the `ProposerVMServer` interface above and mocked it in the unit tests.

vms/proposervm/service.go
ylg-avalabs: The proposerVM fields will be add in this [PR](https://github.com/ava-labs/avalanchego/pull/3746/files#diff-61f08fbc8163cf9db1277f5b073f80903a39d328762a0d75e055b1ff7eef8873R136-R147), and we will only return these useful fields in the PR.

vms/proposervm/client.go
ylg-avalabs: The `Client` interface will be used outside of the repo.

vms/proposervm/vm.go
ylg-avalabs: Added metrics for the proposer vm. Currently, it only contains [metrics.APIInterceptor](https://github.com/ava-labs/avalanchego/pull/4029/files#diff-3981d70b6932134dee1165f3be7ebaa1528c1596ffd1adff155e512beec73c76R15). We will add more metrics when we need later.

vms/proposervm/service.go
yacovm: Would it be possible to rebase this PR on top of that PR (3746) ? 

I am not a fan of leaving incomplete yet accessible APIs that are pending on other PRs.

Also, rebasing this PR on top of that PR would make it possible to properly change the output of the API in this PR and not need to wait for a new PR.

PR #4031, opened 2025-06-24
chore: move windows from tier 3 to unsupported

aaronbuchwald:
## Why this should be merged

This PR moves Windows from Tier 3 support to Unsupported according to the support tiers outlined in the README.

## How this was tested

CI

Yes.

## Notes on Retained Windows Specific Support

Opted to leave windows specific support in place rather than remove it in this PR because the intent is to downgrade the guarantee of support, but this does not mean we need to remove the existing windows specific support. If preferred, we can remove all windows specific handling as well.

The following files contain windows specific handling or references based on a search of the codebase:

- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/vms/rpcchainvm/runtime/subprocess/linux_stopper.go#L7
- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/utils/ulimit/ulimit_windows.go#L13
- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/utils/storage/storage_windows.go#L4
- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/utils/perms/write_file.go#L15
- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/tests/fixture/tmpnet/detached_process_windows.go#L11
- https://github.com/ava-labs/avalanchego/blob/092cf182dd92b66a8ab882b657d3ae52b174c913/tests/fixture/tmpnet/README.md#package-details


maru-ava:
~Do we also want to remove usage of `go:build windows`?~

I see that's handled separately by #4032

StephenButtolph (approved):
lgtm. Will wait to see if there is any pushback

aaronbuchwald:
> ~Do we also want to remove usage of `go:build windows`?~
> 
> I see that's handled separately by #4032

Ya I split these out into two as I thought this was the minimum changes needed to deprecate official windows support ie. testing and generating releases. The follow on is if we also want to delete the work we've done, so that it can run on windows, which I don't think is strictly necessary.

PR #4033, opened 2025-06-24
Connectrpc health api

mpignatelli12:


api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/api/admin"
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	adminv1 "github.com/ava-labs/avalanchego/proto/pb/admin/v1"
+	adminv1connect "github.com/ava-labs/avalanchego/proto/pb/admin/v1/adminv1connect"
+)
+
+// NewConnectAdminService returns a ConnectRPC-compatible AdminServiceHandler
+// that delegates calls to the existing Admin implementation
```
joshua-kim: Wording of this comment is slightly awkward + we'll have to update it when we stop wrapping the existing api

```
// NewConnectAdminService returns a ConnectRPC handler for the Admin API
```

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+func NewConnectAdminService(admin *admin.Admin) *ConnectAdminService {
+	return &ConnectAdminService{
+		Admin: admin,
+	}
+}
+
+// ConnectAdminService implements the admin.v1.AdminService using the existing Admin service
```
joshua-kim: We don't need to specify that this implements an interface - that's implicitly known in Go when an interface type check is done + you do a compile time check later.

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+type ConnectAdminService struct {
+	*admin.Admin
+}
+
+// Compile-time check: ensure ConnectAdminService implements AdminServiceHandler
+var _ adminv1connect.AdminServiceHandler = (*ConnectAdminService)(nil)
```
joshua-kim: This check is typically done at the top of the file, after the imports. This is also idiomatic in Go so you don't need a comment to explain it

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+
+// StartCPUProfiler starts a CPU profile writing to the specified file
+func (s *ConnectAdminService) StartCPUProfiler(
```
joshua-kim: nit: update receiver names to `c` (IDE can do this in one click, ping me if you don't know how to do this). Typically in Go you name receivers the first letter of the struct. The Goland refacting tool has a shortcut to instantly rename all usages of this so you don't have to do it manually each time.

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+// that delegates calls to the existing Admin implementation
+func NewConnectAdminService(admin *admin.Admin) *ConnectAdminService {
+	return &ConnectAdminService{
+		Admin: admin,
+	}
+}
+
+// ConnectAdminService implements the admin.v1.AdminService using the existing Admin service
+type ConnectAdminService struct {
+	*admin.Admin
```
joshua-kim: Avoid embedding types - every field/function gets promoted into this type's signature and can introduce subtle bugs/weird behavior. It's useful for some cases but I find it to be an overly blunt tool and I generally don't embed unless I need to.

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+
+// StartCPUProfiler starts a CPU profile writing to the specified file
+func (s *ConnectAdminService) StartCPUProfiler(
+	_ context.Context,
+	_ *connect.Request[emptypb.Empty],
+) (*connect.Response[emptypb.Empty], error) {
+	if err := s.Admin.StartCPUProfiler(nil, nil, &api.EmptyReply{}); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+	return connect.NewResponse(&emptypb.Empty{}), nil
```
joshua-kim: nit: add a space here (same w/ rest of functions)

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// StopCPUProfiler stops the CPU profile
+func (s *ConnectAdminService) StopCPUProfiler(
+	_ context.Context,
+	_ *connect.Request[emptypb.Empty],
+) (*connect.Response[emptypb.Empty], error) {
+	if err := s.Admin.StopCPUProfiler(nil, nil, &api.EmptyReply{}); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
```
joshua-kim: Why are we setting the error code to `connect.CodeInternal`? The docs ([ref](https://github.com/connectrpc/connect-go/blob/84c176191838c0d5b4704840b6aa4a7f38dce0b9/code.go#L89-L91)) suggest that this is reserved for an invariant break, and it doesn't seem like we have enough information to make that decision here. I'm worried this might lead to incorrect client retry behavior/misinterpreted error signal. Could we just return the error so it defaults to the unknown response code?

> Connect handlers attach status codes to errors using the NewError function. Handlers should return coded errors; if they don't, Connect will use the deadline_exceeded code for context.DeadlineExceeded, canceled for context.Canceled, and unknown for all other errors. For example:

([ref](https://connectrpc.com/docs/go/errors/?utm_source=chatgpt.com#working-with-errors))

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	return connect.NewResponse(&emptypb.Empty{}), nil
+}
+
+// SetLoggerLevel sets the log level and/or display level for loggers
+func (s *ConnectAdminService) SetLoggerLevel(
+	_ context.Context,
+	req *connect.Request[adminv1.SetLoggerLevelArgs],
+) (*connect.Response[adminv1.SetLoggerLevelReply], error) {
+	// Convert the request to the format used by the JSON-RPC API
```
joshua-kim: This is overly verbose, the code here is self-documenting so we can avoid the comment here

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+	jsonArgs := &admin.SetLoggerLevelArgs{
+		LoggerName: req.Msg.LoggerName,
+	}
+
+	// Handle optional log level parameters
```
joshua-kim: Same nit

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+	if err := s.Admin.LockProfile(nil, nil, &api.EmptyReply{}); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+	return connect.NewResponse(&emptypb.Empty{}), nil
+}
+
+// Alias attempts to alias an HTTP endpoint to a new name
+func (s *ConnectAdminService) Alias(
+	_ context.Context,
+	req *connect.Request[adminv1.AliasArgs],
```
joshua-kim: nit: name this to `request`. I think old people in the 80s or something used tons of abbreviations in their code when monitors were limited to 80 chars but with modern-day monitors we have no such restrictions so we can optimize for clarity instead of screen width

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	reply := &adminv1.GetConfigReply{
+		ConfigJson: string(configJSON),
+	}
+	return connect.NewResponse(reply), nil
+}
+
+// LoadVMs loads any new VMs available to the node
+func (s *ConnectAdminService) LoadVMs(
```
joshua-kim: I would be very down to deprecate this api... it literally doesn't work currently. Can we just remove it from the new api?

api/admin/connecthandler/admin_connect.go
```diff
@@ -0,0 +1,305 @@
...
+		protoLoggerLevels[name] = &adminv1.LogAndDisplayLevels{
+			LogLevel:     levels.LogLevel.String(),
+			DisplayLevel: levels.DisplayLevel.String(),
+		}
+	}
+
+	return &adminv1.GetLoggerLevelReply{
+		LoggerLevels: protoLoggerLevels,
+	}
+}
```
joshua-kim: These helpers are both pretty trivlal and are only used in one place so I would just in-line them in the functions. I think breaking them apart makes it harder to read the code since I have to jump down to see the implementation.

api/admin/service.go
```diff
@@ -64,16 +64,17 @@ type Admin struct {
 
 // NewService returns a new admin API service.
 // All of the fields in [config] must be set.
-func NewService(config Config) (http.Handler, error) {
+func NewService(config Config) (http.Handler, *Admin, error) {
```
joshua-kim: Instead of returning `Admin` (3 return values), can we just take in the `Admin` as an argument to this function (so the caller creates it)?

api/admin/service.go
```diff
@@ -391,8 +392,8 @@ type DBGetReply struct {
 	ErrorCode rpcdbpb.Error `json:"errorCode"`
 }
 
-//nolint:stylecheck // renaming this method to DBGet would change the API method from "dbGet" to "dBGet"
-func (a *Admin) DbGet(_ *http.Request, args *DBGetArgs, reply *DBGetReply) error {
+// DBGet returns the value of a database entry
+func (a *Admin) DBGet(_ *http.Request, args *DBGetArgs, reply *DBGetReply) error {
```
joshua-kim: I don't think this should be a part of this PR - can we clean the diff here?

api/admin/service_test.go
```diff
@@ -153,7 +153,7 @@ func TestServiceDBGet(t *testing.T) {
 			require := require.New(t)
 
 			reply := &DBGetReply{}
-			require.NoError(a.DbGet(
+			require.NoError(a.DBGet(
```
joshua-kim: Same comment

api/health/connecthandler/health_connect.go
```diff
@@ -0,0 +1,137 @@
...
+// NewConnectHealthService returns a ConnectRPC-compatible ConnectHealthService
+// that delegates calls to the existing health service implementation.
+func NewConnectHealthService(healthService health.Health) *ConnectHealthService {
+	return &ConnectHealthService{
+		Service: healthService,
+	}
+}
+
+type ConnectHealthService struct {
+	Service health.Health
```
joshua-kim: I think this can be unexported

api/health/connecthandler/health_connect.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (s *ConnectHealthService) Readiness(
+	_ context.Context,
+	req *connect.Request[healthv1.ReadinessArgs],
+) (*connect.Response[healthv1.ReadinessReply], error) {
+	// The health.Health interface has different methods than the service.
+	// We need to use the methods of the Health interface.
```
joshua-kim: I'm not sure what this comment is for... can we remove it? It's clear that this is a wrapper/adapter pattern so we don't need to document this.

api/health/connecthandler/health_connect.go
```diff
@@ -0,0 +1,137 @@
...
+	checks, healthy := s.Service.Readiness(req.Msg.Tags...)
+
+	// Convert the health.Result map to protobuf format
```
joshua-kim: same nit: For self documenting code - comments make readability worse. Can we remove?

api/health/connecthandler/health_connect.go
```diff
@@ -0,0 +1,137 @@
...
+	checksProto := make(map[string]*healthv1.Result, len(checks))
```
joshua-kim: preallocating the map size is a premature optimization unless we have profiler data that suggests otherwise - can we remove?

api/health/connecthandler/health_connect.go
```diff
@@ -0,0 +1,137 @@
...
+) (*connect.Response[healthv1.HealthReply], error) {
+	// The health.Health interface has different methods than the service.
+	// We need to use the methods of the Health interface.
+	checks, healthy := s.Service.Health(req.Msg.Tags...)
+
+	// Convert the health.Result map to protobuf format
+	checksProto := make(map[string]*healthv1.Result, len(checks))
+	for name, check := range checks {
+		checksProto[name] = convertResult(check)
+	}
```
joshua-kim: It seems like since `Readiness` and `Health` and `Liveness` all return the same type we can just make a single helper does `convertResult`, and it also handles the creation of the map + converting internally.

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,327 @@
...
+	if err := s.Info.GetNodeID(nil, nil, &jsonReply); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+
+	nodePOP := []byte{}
+	if jsonReply.NodePOP != nil {
+		// Use Marshal to serialize ProofOfPossession to bytes
+		// MarshalJSON is not ideal here. Ideally, we would use a binary serialization method (MarshalBinary, Marshal, etc.)
+		var err error
+		nodePOP, err = jsonReply.NodePOP.MarshalJSON()
```
joshua-kim: Why are we using json serialization here? `ProofOfPossession` is just bytes so we could just send encoded bytes over the wire. Currently this will output:
```
{
  "nodeId": "NodeID-7Xhw2mDxuDS44j42TCB6U5579esbSt3Lg",
  "nodePop": "eyJwdWJsaWNLZXkiOiIweDkwMGM5YjExOWI1YzgyZDc4MWQ0YjQ5YmU3OGMzZmM3YWU2NWYyYjQzNWI3ZWQ5ZTNhOGI5YTAzZTQ3NWVkZmY4NmQ4YTY0ODI3ZmVjOGRiMjNhNmYyMzZhZmJmMTI3ZCIsInByb29mT2ZQb3NzZXNzaW9uIjoiMHg4YmZkNmQ0ZDIwODZiMmI4MTE1ZDhmNzJmOTQwOTVmZWZlNWE2YzA3ODc2YjJhY2NmNTFhODExYWRmNTIwZjM4OWU3NGEzZDIxNTJhNmQ5MGI1MjFlMmJlNThmZmU0NjgwNDNkYzVlYTY4YjRjNDQ0MTBlYjY3ZjhkYzI0ZjEzZWQ0ZjE5NDAwMDc2NGMwZTkyMmNkMjU0YTM1ODhhNDk2MmIxY2I0ZGI3ZGU0YmI5Y2RhOWQ5ZDRkNmIwM2YzZDIifQ=="
}
```

instead of displaying the nested struct fields

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,327 @@
...
+
+	reply := &v1.PeersReply{
+		NumPeers: uint32(jsonReply.NumPeers),
+		Peers:    peers,
+	}
+
+	return connect.NewResponse(reply), nil
+}
+
+// Helper function to format time
```
joshua-kim: This comment is also not needed, but would be helpful to document that this formats stuff in "YYYY-MM-DDThh:mm:ssZ"

api/info/service.go
```diff
@@ -98,22 +98,23 @@ func NewService(
 	myIP *utils.Atomic[netip.AddrPort],
 	network network.Network,
 	benchlist benchlist.Manager,
-) (http.Handler, error) {
+) (http.Handler, *Info, error) {
```
joshua-kim: Similar comment w/ Admin creation

api/metrics/connecthandler/metrics_connect.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"connectrpc.com/connect"
+	"github.com/prometheus/client_golang/prometheus"
+
+	metricsv1 "github.com/ava-labs/avalanchego/proto/pb/metrics/v1"
+	dto "github.com/prometheus/client_model/go"
+)
+
+type ConnectMetricsService struct {
+	Service prometheus.Gatherer
```
joshua-kim: Same comment w.r.t exporting

api/metrics/connecthandler/metrics_connect.go
```diff
@@ -0,0 +1,72 @@
...
+			case dto.MetricType_COUNTER:
+				if m.Counter != nil {
+					value = m.Counter.GetValue()
+				}
+			case dto.MetricType_GAUGE:
+				if m.Gauge != nil {
+					value = m.Gauge.GetValue()
+				}
+			default:
+				// Skip histograms, summaries, and untyped metrics
```
joshua-kim: You don't need to document that you're skipping stuff, but it would be more helpful to document why you are skipping these types (it's not obvious to me)

node/node.go
```diff
@@ -391,6 +401,12 @@ type Node struct {
 
 	// Closed when a sufficient amount of bootstrap nodes are connected to
 	onSufficientlyConnected chan struct{}
+
+	// Info API
+	info *info.Info
+
+	// Admin API
```
joshua-kim: Don't need comments here

node/node.go
```diff
@@ -778,7 +794,7 @@ func (n *Node) initDatabase() error {
 	rawExpectedGenesisHash := hashing.ComputeHash256(n.Config.GenesisBytes)
 
 	rawGenesisHash, err := n.DB.Get(genesisHashKey)
-	if err == database.ErrNotFound {
+	if errors.Is(err, database.ErrNotFound) {
```
joshua-kim: Clean diff

proto/admin/v1/service.proto
```diff
@@ -0,0 +1,93 @@
...
+
+// AdminService is the API service for node admin management
+service AdminService {
+  rpc StartCPUProfiler(google.protobuf.Empty) returns (google.protobuf.Empty);
+  rpc StopCPUProfiler(google.protobuf.Empty) returns (google.protobuf.Empty);
+  rpc MemoryProfile(google.protobuf.Empty) returns (google.protobuf.Empty);
+  rpc LockProfile(google.protobuf.Empty) returns (google.protobuf.Empty);
+  rpc Alias(AliasArgs) returns (google.protobuf.Empty);
+  rpc AliasChain(AliasChainArgs) returns (google.protobuf.Empty);
+  rpc GetChainAliases(GetChainAliasesArgs) returns (GetChainAliasesReply);
```
joshua-kim: RPC naming convention is typically `*Request/*Response`. Can we use that naming convention in our proto types? You should be able to rename if you install the proto plugin in your ide + use the aforementioned refactor tool

proto/admin/v1/service.proto
```diff
@@ -0,0 +1,93 @@
...
+
+message DBGetArgs {
+  string key = 1;
+}
+
+enum ErrorCode {
+  ERROR_CODE_UNSPECIFIED = 0;
+  ERROR_CODE_NOT_FOUND = 1;
+  ERROR_CODE_DATABASE_ERROR = 2;
+  // Add more if needed
```
joshua-kim: same nit

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+var _ infov1connect.InfoServiceHandler = (*connectInfoService)(nil)
+
+type connectInfoService struct {
+	*info.Info
+}
+
+// NewConnectInfoService returns a ConnectRPC handler for the Info API
+func NewConnectInfoService(info *info.Info) infov1connect.InfoServiceHandler {
```
joshua-kim: 1. This is more of a matter of preference, but I typically don't make constructor functions for trivial initialization like this (I would just use struct initialization) since it's just wiring some fields. It can be useful when you want to make sure that if a new field is added, all consumers of this api break at compile time instead of runtime, but I'm not too concerned about it here since it's an api we own and it's only used in avalanchego.
2. If we did keep this, we would want to return a `*ConnectInfoService` type so that the caller of this API could use it either as an interface or as a struct instead of being forced to use the interface. If you're interested in learning more about the rationale you can search "accept interfaces, return structs" which is the Go mantra for this.

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/ava-labs/avalanchego/proto/pb/info/v1/infov1connect"
+	"github.com/ava-labs/avalanchego/upgrade"
+
+	v1 "github.com/ava-labs/avalanchego/proto/pb/info/v1"
+)
+
+var _ infov1connect.InfoServiceHandler = (*connectInfoService)(nil)
+
+type connectInfoService struct {
+	*info.Info
```
joshua-kim: Can we avoid embedding this type and make this a field instead? Embedding auto-promotes all fields/methods which can result in subtle side-effects so I generally avoid using it unless I need it.

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+
+// NewConnectInfoService returns a ConnectRPC handler for the Info API
+func NewConnectInfoService(info *info.Info) infov1connect.InfoServiceHandler {
+	return &connectInfoService{
+		Info: info,
+	}
+}
+
+// GetNodeVersion returns the semantic version, database version, RPC protocol version,
+// Git commit hash, and the list of VM versions this node is running
```
joshua-kim: This comment is too coupled to the data being returned (similar comment on the rest of the api functions), if we ever update the response type we'll have to update this comment. Would it be reasonable for the user to interpret the version response since it labels fields like databaseVersion/rpcProtocolVersion/etc? I think for this + the other api functions we could just copy the comments from the existing api ([ref](https://github.com/ava-labs/avalanchego/blob/5cb8b4363dfda46a3490e31c16a9f4c454f6acfe/api/info/service.go#L156)).
```suggestion
// GetNodeVersion returns the version this node is running
```

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+		DatabaseVersion:    jsonResponse.DatabaseVersion,
+		RpcProtocolVersion: uint32(jsonResponse.RPCProtocolVersion),
+		GitCommit:          jsonResponse.GitCommit,
+		VmVersions:         vmVersions,
+	}
+
+	return connect.NewResponse(response), nil
+}
+
+// GetNodeID returns this node's unique identifier and proof-of-possession bytes
```
joshua-kim: Similar comment (gonna stop leaving this comment on the rest of this file since it'll be repetitive)
```suggestion
// GetNodeID returns the node ID of this node
```

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+	if jsonResponse.NodePOP != nil {
+		p := jsonResponse.NodePOP
+		nodePOP = make([]byte, len(p.PublicKey)+len(p.ProofOfPossession))
+		copy(nodePOP[0:len(p.PublicKey)], p.PublicKey[:])
+		copy(nodePOP[len(p.PublicKey):], p.ProofOfPossession[:])
+	}
+
+	response := &v1.GetNodeIDResponse{
+		NodeId:  jsonResponse.NodeID.String(),
+		NodePop: nodePOP,
```
joshua-kim: This is going to be raw bytes, so it might be hard to read. As an example, this command outputs:

```
➜  avalanchego git:(connectrpc-health-api) grpcurl \                                                                    <<<
  -H "Avalanche-API-Route: info" \
  -plaintext \
  -d '{}' \
  127.0.0.1:9650 \
  info.v1.InfoService/GetNodeID
{
  "nodeId": "NodeID-7Xhw2mDxuDS44j42TCB6U5579esbSt3Lg",
  "nodePop": "kAybEZtcgteB1LSb54w/x65l8rQ1t+2eOouaA+R17f+G2KZIJ/7I2yOm8javvxJ9i/1tTSCGsrgRXY9y+UCV/v5abAeHayrM9RqBGt9SDziedKPSFSptkLUh4r5Y/+RoBD3F6mi0xEQQ62f43CTxPtTxlAAHZMDpIs0lSjWIpJYrHLTbfeS7nNqdnU1rA/PS"
}
```

But master outputs:

```
➜  avalanchego git:(master) curl -s -POST --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"info.getNodeID",
    "params" :{}
}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/info | jq
{
  "jsonrpc": "2.0",
  "result": {
    "nodeID": "NodeID-7Xhw2mDxuDS44j42TCB6U5579esbSt3Lg",
    "nodePOP": {
      "publicKey": "0x900c9b119b5c82d781d4b49be78c3fc7ae65f2b435b7ed9e3a8b9a03e475edff86d8a64827fec8db23a6f236afbf127d",
      "proofOfPossession": "0x8bfd6d4d2086b2b8115d8f72f94095fefe5a6c07876b2accf51a811adf520f389e74a3d2152a6d90b521e2be58ffe468043dc5ea68b4c44410eb67f8dc24f13ed4f194000764c0e922cd254a3588a4962b1cb4db7de4bb9cda9d9d4d6b03f3d2"
    }
  },
  "id": 1
}
```

We might need some helper, it looks like we have a type implemented that gives us this type in an encoded format that we could modify/do something similar to help us here: https://github.com/ava-labs/avalanchego/blob/a6df03e9092d1f16a7400f854fa24c03ccdc9da0/vms/platformvm/signer/proof_of_possession.go#L70

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+		benched := make([]string, len(peer.Benched))
+		copy(benched, peer.Benched)
+
+		peers = append(peers, &v1.PeerInfo{
+			Ip:             peer.IP.String(),
+			PublicIp:       peer.PublicIP.String(),
+			NodeId:         peer.ID.String(),
+			Version:        peer.Version,
+			LastSent:       formatTime(peer.LastSent),
+			LastReceived:   formatTime(peer.LastReceived),
```
joshua-kim: It looks like if we used https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp we can pass in time and not have to worry about timestamp formatting hell

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+			Benched:        benched,
+			ObservedUptime: uint32(peer.ObservedUptime),
```
joshua-kim: Similarly it looks like we could use a duration type so we don't have to have confusion about what this time actually means (nanos/millis/seconds?) https://protobuf.dev/reference/protobuf/google.protobuf/#duration

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+			TrackedSubnets: trackedSubnets,
+		})
+	}
+
+	response := &v1.PeersResponse{
+		NumPeers: uint32(jsonResponse.NumPeers),
```
joshua-kim, line 224: This is a part of the existing api so we could keep it, but I'm so confused as to why it is a part of the existing api since this is literally just `len(peers)` (which is already a part of this response, so it's not conveying any additional information). I kind of want to drop this from the connect api eventually...

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+// Peers returns metadata (IP, nodeID, version, uptimes, subnets, etc.) for the given peer node IDs
+func (c *connectInfoService) Peers(
+	_ context.Context,
+	request *connect.Request[v1.PeersRequest],
+) (*connect.Response[v1.PeersResponse], error) {
+	nodeIDs := make([]ids.NodeID, 0, len(request.Msg.NodeIds))
+	for _, nodeIDStr := range request.Msg.NodeIds {
+		nodeID, err := ids.NodeIDFromString(nodeIDStr)
+		if err != nil {
+			return nil, connect.NewError(
```
joshua-kim: Can we just return this error and not worry about error codes (so connect will just use the default unspecified code)? I'm open to introducing new error codes as they're needed but for now without a clear reason for them I think it's a bit premature.

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	var jsonResponse info.PeersReply
+	if err := c.Info.Peers(nil, &jsonRequest, &jsonResponse); err != nil {
+		return nil, err
+	}
+
+	peers := make([]*v1.PeerInfo, 0, len(jsonResponse.Peers))
+	for _, peer := range jsonResponse.Peers {
+		// Convert TrackedSubnets (set.Set[ids.ID]) to []string
```
joshua-kim: This is reasonably self-documenting, so we could drop this comment

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+		trackedSubnetsIDs := peer.TrackedSubnets.List()
+		trackedSubnets := make([]string, len(trackedSubnetsIDs))
+		for i, id := range trackedSubnetsIDs {
+			trackedSubnets[i] = id.String()
+		}
```
joshua-kim: In Go it's more idiomatic to use `append` instead of indexing to copy stuff over
```suggestion
		trackedSubnets := make([]string, len(trackedSubnetsIDs))
		for _, subnetID := range peer.TrackedSubnets.List() {
			trackedSubnets = append(trackedSubnets, subnetID.String())
		}
```

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+		Peers:    peers,
+	}
+
+	return connect.NewResponse(response), nil
+}
+
+// formatTime formats a time.Time object to a string in "YYYY-MM-DDThh:mm:ssZ" format
+func formatTime(t time.Time) string {
+	return t.Format(time.RFC3339)
+}
```
joshua-kim: Related comment but I think we can remove this if we use the timestamp library

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+
+// IsBootstrapped returns whether the named chain has finished its bootstrap process on this node
+func (c *connectInfoService) IsBootstrapped(
+	_ context.Context,
+	request *connect.Request[v1.IsBootstrappedRequest],
+) (*connect.Response[v1.IsBootstrappedResponse], error) {
+	// Use the chain from the request
```
joshua-kim: Can remove this comment

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+func (c *connectInfoService) GetVMs(
+	_ context.Context,
+	_ *connect.Request[v1.GetVMsRequest],
+) (*connect.Response[v1.GetVMsResponse], error) {
+	var jsonResponse info.GetVMsReply
+	if err := c.Info.GetVMs(nil, nil, &jsonResponse); err != nil {
+		return nil, err
+	}
+
+	// Convert the VM map from JSON-RPC format to protobuf format
```
joshua-kim: Can remove this comment

api/info/connecthandler/info_connect.go
```diff
@@ -0,0 +1,325 @@
...
+	vms := make(map[string]*v1.VMAliases)
+	for vmID, aliases := range jsonResponse.VMs {
+		vms[vmID.String()] = &v1.VMAliases{
+			Aliases: aliases,
+		}
+	}
+
+	// Convert the FXs map from JSON-RPC format to protobuf format
```
joshua-kim: Can remove this comment

node/node.go
```diff
@@ -1349,33 +1392,47 @@ func (n *Node) initInfoAPI() error {
...
+	mux := http.NewServeMux()
+	mux.Handle(reflectionPattern, reflectionHandler)
+	mux.Handle(infoPattern, infoConnHandler)
+	n.APIServer.AddHeaderRoute("info", mux)
+
 	return n.APIServer.AddRoute(
 		service,
 		"info",
-		"",
+		"info",
```
joshua-kim: This looks like a bug from a bad merge/rebase, we probably need to undo this diff (there's a similar diff on the other apis when we call `AddRoute`).

tests/e2e/connect/info_test.go
```diff
@@ -0,0 +1,244 @@
...
+
+	"github.com/ava-labs/avalanchego/proto/pb/info/v1/infov1connect"
+
+	infov1 "github.com/ava-labs/avalanchego/proto/pb/info/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+type stubInfoService struct{}
```
joshua-kim: As discussed offline let's try to use `tmpnet` so we can get rid of this

tests/e2e/connect/info_test.go
```diff
@@ -0,0 +1,244 @@
...
+) (*connect.Response[infov1.GetVMsResponse], error) {
+	return connect.NewResponse(&infov1.GetVMsResponse{
+		Vms: map[string]*infov1.VMAliases{
+			"avm": {Aliases: []string{"avm"}},
+		},
+		Fxs: map[string]string{},
+	}), nil
+}
+
+var _ = Describe("InfoService ConnectRPC E2E (integration, no stubs)", func() {
```
joshua-kim: Here's an example of a test that calls the info api, maybe we could try to see if we can get a similar setup where we setup tmpnet nodes and call their apis by hitting their urls: https://github.com/ava-labs/avalanchego/blob/0554b91f73e40bdab3fb1eefde2093ded0b99b05/tests/e2e/p/interchain_workflow.go#L112

utils/resource/usage.go
joshua-kim, line 1: This diff is probably from a bad merge, we should undo this diff (`git checkout master -- utils/resource/usage.go`)

vms/rpcchainvm/vm_test.go
joshua-kim, line 1: Same comment w.r.t bad merge

api/admin/service.go
```diff
@@ -391,7 +388,7 @@ type DBGetReply struct {
 	ErrorCode rpcdbpb.Error `json:"errorCode"`
 }
 
-//nolint:staticcheck // renaming this method to DBGet would change the API method from "dbGet" to "dBGet"
+//nolint:stylecheck // renaming this method to DBGet would change the API method from "dbGet" to "dBGet"
```
joshua-kim: This diff looks like it needs to be undone/is unrelated

proto/info/v1/service.proto
```diff
@@ -0,0 +1,126 @@
...
+  string database_version = 2;
+  uint32 rpc_protocol_version = 3;
+  string git_commit = 4;
+  map<string, string> vm_versions = 5;
+}
+
+message GetNodeIDRequest {}
+message GetNodeIDResponse {
+  string node_id = 1;
+  bytes node_pop = 2;
```
joshua-kim: Same comment w.r.t this reading weirdly - I think we want this to be a nested struct instead of raw bytes (e.g the public key and pop signature)

proto/info/v1/service.proto
```diff
@@ -0,0 +1,126 @@
...
+  rpc GetNodeID(GetNodeIDRequest) returns (GetNodeIDResponse);
+  rpc GetNodeIP(GetNodeIPRequest) returns (GetNodeIPResponse);
+  rpc GetNetworkID(GetNetworkIDRequest) returns (GetNetworkIDResponse);
+  rpc GetNetworkName(GetNetworkNameRequest) returns (GetNetworkNameResponse);
+  rpc GetBlockchainID(GetBlockchainIDRequest) returns (GetBlockchainIDResponse);
+  rpc Peers(PeersRequest) returns (PeersResponse);
+  rpc IsBootstrapped(IsBootstrappedRequest) returns (IsBootstrappedResponse);
+  rpc Upgrades(UpgradesRequest) returns (UpgradesResponse);
+  rpc Uptime(UptimeRequest) returns (UptimeResponse);
+  rpc GetVMs(GetVMsRequest) returns (GetVMsResponse);
```
joshua-kim: Some of these are inconsistent with others in terms of naming. It seems like the convention for rpcs is to just have rpcs called `Foo` instead of `GetFoo` unless we are disambiguating like `GetFooByBar` or `GetFooByBaz` (i.e the Get is implied and is otherwise redundant). Could we drop the `Get*` prefix?

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
+	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
+
+	infov1 "github.com/ava-labs/avalanchego/proto/pb/info/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestInfoE2E(t *testing.T) {
```
maru-ava: Please remove this test. Ginkgo is intended to be declarative, and execution of a this file should be via the ginkgo test runner (e.g. `ginkgo -v --focus-file=info_e2e_test ./tests/e2e`) rather than by direct execution to ensure that the shared setup in `tests/e2e/e2e_test.go` gets executed.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/onsi/ginkgo"
+
+	"github.com/ava-labs/avalanchego/config"
+	"github.com/ava-labs/avalanchego/proto/pb/info/v1/infov1connect"
+	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
+	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
+
+	infov1 "github.com/ava-labs/avalanchego/proto/pb/info/v1"
+
+	. "github.com/onsi/ginkgo/v2"
```
maru-ava: We don't use dot imports, please remove.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,151 @@
...
+	. "github.com/onsi/gomega"
```
maru-ava: As per the example of the existing e2e tests, we don't use gomega. testify (require.*) is the standard across all types of testing.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+var _ = ginkgo.Describe("[Connect Info API]", func() {
+	var (
+		tc     = e2e.NewTestContext()
+		env    = e2e.GetEnv(tc)
+		client infov1connect.InfoServiceClient
+		ctx    = tc.DefaultContext()
+	)
+
+	BeforeSuite(func() {
```
maru-ava: Since this file should be defining a test rather than a suite, you'll want to use `BeforeAll` which is localized to the `Describe` closure it is defined in.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+		rootDir := env.RootNetworkDir
+		network := tmpnet.NewDefaultNetwork("connect-info-e2e")
+
+		err := tmpnet.BootstrapNewNetwork(ctx, tc.Log(), network, rootDir)
+		Expect(err).ToNot(HaveOccurred())
+
+		node, err := network.GetNode(ids.BuildTestNodeID([]byte("node-0")))
```
maru-ava, line 37: Although, if this test only needs a single node rather than a network (i.e. no consensus), maybe this test would be a good candidate for an integration test (i.e. start a single node network and test). Would be faster because the shared network deploys a subnet that this test doesn't require. Feel free to poke me on slack if you'd like to explore this.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+	})
+
+	It("IsBootstrapped returns true for X", func() {
+		Eventually(func() bool {
+			req := connect.NewRequest(&infov1.IsBootstrappedRequest{Chain: "X"})
+			req.Header().Set("Avalanche-API-Route", "info")
+			resp, err := client.IsBootstrapped(ctx, req)
+			return err == nil && resp.Msg.IsBootstrapped
+		}, 60*time.Second, 2*time.Second).Should(BeTrue())
+	})
```
maru-ava, line 107: Please use the tests/e2e version of Eventually (compatible with testify) and use the same constants as other tests.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -5,134 +5,128 @@ package connect_test
...
+		require.NoError(tc, err)
 
 		node, err := network.GetNode(ids.BuildTestNodeID([]byte("node-0")))
 		e2e.WaitForHealthy(tc, node)
 		url := node.GetAccessibleURI()
 		client = infov1connect.NewInfoServiceClient(http.DefaultClient, url)
 	})
 
-	It("NodeVersion returns version info", func() {
+	tc.By("NodeVersion returns version info", func() {
```
maru-ava: `It` defines a test, `By` adds logging around a block 'by doing x'. `By` is intended for use internal to `It`.

tests/e2e/connect/info_e2e_test.go
```diff
@@ -5,134 +5,128 @@ package connect_test
...
 	)
 
-	BeforeSuite(func() {
+	ginkgo.BeforeAll(func() {
 		rootDir := env.RootNetworkDir
 		network := tmpnet.NewDefaultNetwork("connect-info-e2e")
 
 		err := tmpnet.BootstrapNewNetwork(ctx, tc.Log(), network, rootDir)
-		Expect(err).ToNot(HaveOccurred())
+		require.NoError(tc, err)
```
maru-ava: Common practice when `require` is used pervasively in a test is to redefine/shadow it early e.g. `require := testify.New(tc)` and thus avoid having to pass `tc` on every invocation.

PR #4034, opened 2025-06-24
Add block timeskew metric

yacovm:
## Why this should be merged

We have a metric that measures the time between building the block to accepting it. However it uses the block's timestamp. This value includes the time the VM builds a block as we need to compute the time and insert the time into the block before we build it.

To better compute the time between right before the block builder releases the block to the network and the block's acceptance it can be useful if we can estimate the time skew between the local time and the block's timestamp. 

## How this works

After the snowman engine builds a block, compute the difference between the current real time to the block's timestamp and publish it via a metrics gauge. Then we can use a monitor that divides this accumulated number (because it's a gauge that only adds) to the number of blocks built to get the average time skew. 

## How this was tested

I ran a modified local network and sent some Ethereum transactions to the C chain and then probed the metrics endpoint on the HTTP endpoint:

```
➜  ethcli git:(main) ✗ curl  -s "http://localhost:9650/ext/metrics" | grep skew | grep 'chain="C"'
avalanche_snowman_block_time_skew{chain="C"} 4.92657e+08
➜  ethcli git:(main) ✗ curl  -s "http://localhost:9660/ext/metrics" | grep skew | grep 'chain="C"'
avalanche_snowman_block_time_skew{chain="C"} 0
➜  ethcli git:(main) ✗ curl  -s "http://localhost:9670/ext/metrics" | grep skew | grep 'chain="C"'
avalanche_snowman_block_time_skew{chain="C"} 5.2301e+08

```


snow/engine/snowman/metrics.go
```diff
@@ -112,6 +113,10 @@ func newMetrics(reg prometheus.Registerer) (*metrics, error) {
 			Name: "blks_issued",
 			Help: "number of blocks that have been issued into consensus by discovery mechanism",
 		}, []string{"source"}),
+		blockTimeSkew: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "block_time_skew",
+			Help: "The differences between the current time and the block's timestamp",
+		}),
```
StephenButtolph: Should we include (either in the name or the help text) that this is specifically for blocks _built_ by the node? Perhaps making the name `blks_built_time_skew` to align more the with the existing `blks_built` metric (which is used to actually get the time skew per block)

yacovm: Changed the name

yacovm: and changed the description too

StephenButtolph (approved):

PR #4035, opened 2025-06-24
Fix bug in rpcchainvm where `io.EOF` is not propagated correctly

joshua-kim:
## Why this should be merged

The `io.EOF` type is an important sentinel value that is used to signal that there is no more data left to read. Currently, our code just copies the error string which loses the type information which is needed to detect `io.EOF`.

ref: https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/io/io.go;l=38-39

> // EOF is the error returned by Read when no more input is available.
// (Read must return EOF itself, not an error wrapping EOF,
// because callers will test for EOF using ==.)
// Functions should return EOF only to signal a graceful end of input.
// If the EOF occurs unexpectedly in a structured data stream,
// the appropriate error is either [ErrUnexpectedEOF] or some other error
// giving more detail.
var EOF = errors.New("EOF")
## How this works

Adds a check to see if the error string is `"EOF"`.

## How this was tested

Added a regression test

No


vms/rpcchainvm/ghttp/greader/greader_test.go
```diff
@@ -0,0 +1,48 @@
...
+		grpc.WithInsecure(),
+	)
+	require.NoError(err)
+
+	client := NewClient(reader.NewReaderClient(conn))
+
+	buf := make([]byte, 1)
+	n, err := client.Read(buf)
+	require.Zero(n)
+	require.ErrorIs(err, io.EOF)
```
joshua-kim: Should probably do an equality check instead of using errors.Is

mpignatelli12 (approved):

version/constants.go
```diff
@@ -15,15 +15,15 @@ const (
 	// RPCChainVMProtocol should be bumped anytime changes are made which
 	// require the plugin vm to upgrade to latest avalanchego release to be
 	// compatible.
-	RPCChainVMProtocol uint = 41
+	RPCChainVMProtocol uint = 42
```
StephenButtolph, line 18: When we change the protocol version we also update the buf readme (`proto/README.md`) accordingly.

proto/io/reader/reader.proto
```diff
@@ -18,5 +18,16 @@ message ReadResponse {
   // read is the payload in bytes
   bytes read = 1;
   // error is an error message
-  optional string error = 2;
+  optional Error error = 2;
```
StephenButtolph:
```suggestion
  Error error = 2;
```
`optional` doesn't do anything for message types.

vms/rpcchainvm/ghttp/greader/greader_test.go
```diff
@@ -0,0 +1,51 @@
...
+	)
+	require.NoError(err)
+
+	client := NewClient(reader.NewReaderClient(conn))
+
+	buf := make([]byte, 1)
+	n, err := client.Read(buf)
+	require.Zero(n)
+	// Do not use require.ErrorIs because callers use equality checks on io.EOF
+	require.Equal(err, io.EOF)
```
StephenButtolph:
```suggestion
	require.Equal(io.EOF, err)
```

`require.Equal` takes in `(expected, actual)`

vms/rpcchainvm/ghttp/greader/reader_client.go
```diff
@@ -31,8 +31,11 @@ func (c *Client) Read(p []byte) (int, error) {
...
 	copy(p, resp.Read)
 
-	if resp.Error != nil {
-		err = errors.New(*resp.Error)
+	// Sentinel errors must be special-cased through an error code
+	if resp.Error != nil && resp.Error.ErrorCode == readerpb.ErrorCode_ERROR_CODE_EOF {
+		err = io.EOF
+	} else if resp.Error != nil {
+		err = errors.New(resp.Error.Message)
 	}
```
StephenButtolph:
```suggestion
	if resp.Error != nil {
		// Sentinel errors must be special-cased through an error code
		switch resp.Error.ErrorCode {
		case readerpb.ErrorCode_ERROR_CODE_EOF:
			err = io.EOF
		default:
			err = errors.New(resp.Error.Message)
		}
	}
```
Super pedantic, but I just dislike checking the same condition multiple times when not required (`resp.Error != nil`).

vms/rpcchainvm/ghttp/greader/reader_client.go
```diff
@@ -31,8 +31,11 @@ func (c *Client) Read(p []byte) (int, error) {
 
 	copy(p, resp.Read)
 
-	if resp.Error != nil {
-		err = errors.New(*resp.Error)
```
StephenButtolph, line 35: We have this same pattern in: `vms/rpcchainvm/ghttp/gconn/conn_client.go` for `Read` and `Write`. `Read` almost certainly should have this same `io.EOF` handling. Additionally they are documented to report an error wrapping `os.ErrDeadlineExceeded` if the deadlines are exceeded.

StephenButtolph: We also have this pattern in `vms/rpcchainvm/ghttp/gwriter/writer_client.go` for `Write`, but I don't see any special errors that are documented as being expected to be reported.

joshua-kim: I added some handling for this in `gconn` + a regression test

version/constants.go
joshua-kim: Updated the readme

PR #4036, opened 2025-06-25
Support header-based routing

joshua-kim:
## Why this should be merged

We currently decouple our routers across protocols, but this doesn't support connectrpc which is agnostic to protocols, and wants to handle h1 + h2. 

## How this works

1. Supports routing based on a routing header
2. Refactors our handlers on our vm interface to be against a single http handler, agnostic of protocol
3. handlers must implement their own routing

## How this was tested

CI passes

Yes

### DIFF
--- a/utils/resource/usage.go
+++ b/utils/resource/usage.go
@@ -65,6 +65,7 @@ type Manager interface {
 	// Shutdown allocated resources and stop tracking all processes.
 	Shutdown()
 }
+
 type manager struct {
 	log            logging.Logger
 	processMetrics *metrics


api/server/router.go
```diff
@@ -9,99 +9,127 @@ import (
 	"net/http"
 	"sync"
 
-	"github.com/gorilla/mux"
```
joshua-kim: As part of this PR I decided to just gut our dependency on `gorila/mux`... I can remove it from the diff but we had to touch stuff with the router anyways... so it seemed forgivable....

api/server/server.go
```diff
@@ -163,7 +150,7 @@ func (s *server) Dispatch() error {
 
 func (s *server) RegisterChain(chainName string, ctx *snow.ConsensusContext, vm common.VM) {
 	ctx.Lock.Lock()
-	handlers, err := vm.CreateHandlers(context.TODO())
+	handler, err := vm.NewHTTPHandler(context.TODO())
```
joshua-kim: We also did a rename, which I can also undo. But since we're already messing with this interface signature it seemed forgivable to try to sneak in.

connectproto/buf.gen.yaml
```diff
@@ -0,0 +1,11 @@
+version: v1
+plugins:
+  - name: go
+    out: pb
+    opt: paths=source_relative
+  - name: go-grpc
+    out: pb
+    opt: paths=source_relative
+  - plugin: connect-go
```
joshua-kim, line 6: We need to compile our connectrpc stuff with this plugin - so for now I think separate buf modules are the cleanest way to do this. I think ideally it's all under `proto`, so we'd have a dir structure like `proto/connect` and `proto/everythingelse`, but if we do a move in this PR it's going to make the diff huge so I made them both top-level dirs for this PR and I think we should do the move under a single `proto` dir in a follow-up.

vms/example/xsvm/api/ping.go
```diff
@@ -9,33 +9,32 @@ import (
...
+	"github.com/ava-labs/avalanchego/connectproto/pb/xsvm"
+	"github.com/ava-labs/avalanchego/connectproto/pb/xsvm/xsvmconnect"
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
-var _ xsvm.PingServer = (*PingService)(nil)
+var _ xsvmconnect.PingHandler = (*PingService)(nil)
 
 type PingService struct {
-	xsvm.UnsafePingServer
```
joshua-kim, line 24: Connect doesn't have an `Unsafe` server... so we either have to embed the `UnimplementedPingServer` or just not embed anything at all. I forgot the context on why we did this originally... but sans the context I feel like relying on the interface check is reasonable since it will let us know if we didn't implement a method at compile time.

api/server/router.go
```diff
@@ -9,99 +9,125 @@ import (
...
-
 	"github.com/ava-labs/avalanchego/utils/set"
 )
 
 var (
-	errUnknownBaseURL  = errors.New("unknown base url")
+	HTTPHeaderRoute = "Avalanche-API-Route"
```
StephenButtolph: nit: This should be a `const` rather than a `var`.

api/server/router.go
```diff
@@ -9,99 +9,125 @@ import (
...
+		routes:         make(map[string]http.Handler),
 	}
 }
 
 func (r *router) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
-	r.router.ServeHTTP(writer, request)
+	route, ok := request.Header[http.CanonicalHeaderKey(HTTPHeaderRoute)]
```
StephenButtolph: It feels a bit odd to call `http.CanonicalHeaderKey` on every `ServeHTTP`. Shouldn't `HTTPHeaderRoute` just _be_ the canonical header key?

We could add a test that the string is canonical.

api/server/router.go
```diff
@@ -9,99 +9,125 @@ import (
...
+	if !ok {
+		// If there is no routing header, fall-back to the legacy path-based
+		// routing
+		r.mux.ServeHTTP(writer, request)
+		return
+	}
+
+	// Request specified the routing header key but did not provide a
+	// corresponding value
+	if len(route) == 0 {
```
StephenButtolph: Should we expect there to be exactly one here?
```suggestion
	if len(route) != 1 {
```

api/server/router.go
```diff
@@ -9,99 +9,125 @@ import (
...
 		aliases:        make(map[string][]string),
-		routes:         make(map[string]map[string]http.Handler),
+		headerRoutes:   make(map[string]http.Handler),
+		routes:         make(map[string]http.Handler),
 	}
 }
 
 func (r *router) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
```
StephenButtolph, line 51: The read lock was previously required when using gorilla's mux. Do we still need to hold this read lock during the `ServeHTTP` calls here? It seems like we don't need to hold it at the very least with the header based routing

proto/vm/vm.proto
```diff
@@ -20,8 +20,7 @@ service VM {
   // Shutdown is called when the node is shutting down.
   rpc Shutdown(google.protobuf.Empty) returns (google.protobuf.Empty);
   // Creates the HTTP handlers for custom chain network calls.
```
StephenButtolph:
```suggestion
  // Creates the HTTP handler for custom chain network calls.
```

snow/engine/common/vm.go
```diff
@@ -65,20 +65,8 @@ type VM interface {
...
-	// it have an extension called `accounts`, where clients could get
-	// information about their accounts.
-	CreateHandlers(context.Context) (map[string]http.Handler, error)
-
-	// CreateHTTP2Handler returns the http/2 handler to register into the
-	// avalanchego api server.
-	CreateHTTP2Handler(ctx context.Context) (http.Handler, error)
+	// NewHTTPHandler returns the handler to register into the avalanchego http
+	// server. The "Avalanche-API-Route" header must be specified with this VM's
+	// corresponding chain id by clients to route requests to this handler.
```
StephenButtolph: I don't think this comment is accurate. We still support path-based routing to this handler

tests/e2e/vms/xsvm.go
```diff
@@ -196,32 +201,40 @@ var _ = ginkgo.Describe("[XSVM]", ginkgo.Label("xsvm"), func() {
...
 		require.NoError(err)
-		require.Equal(msg, reply.Message)
+		require.Equal(msg, reply.Msg.Message)
 
 		tc.By("serving bidirectional streaming rpc")
-		stream, err := client.StreamPing(tc.DefaultContext())
-		require.NoError(err)
 
+		stream := client.StreamPing(tc.DefaultContext())
+		stream.RequestHeader().Set(server.HTTPHeaderRoute, chainID.String())
```
StephenButtolph: This initialization seems pretty clunky.

Can we implement a helper (similarly to how we did with the grpc library)?

Specifically, I think we should be able to have a single helper that takes in the `node.URI` and `chainID` and it should return a `xsvmconnect.PingClient` which automatically sets the `server.HTTPHeaderRoute` by using the `connect.Interceptors` that we can pass into `xsvmconnect.NewPingClient`.

vms/avm/vm.go
```diff
@@ -311,11 +312,22 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
...
+	mux := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		path := r.URL.Path
+
+		if strings.HasSuffix(path, "/wallet") {
+			walletServer.ServeHTTP(w, r)
+			return
+		}
+
+		rpcServer.ServeHTTP(w, r)
+	})
```
StephenButtolph: How much do we care about being accurate for the pathing here (and in other VMs?).

For example this is now a valid query:
```
curl --location 'http://localhost:9650/ext/bc/X/asdkcjhbasdc' \
--data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"avm.getHeight"
}'
```

While exposing additional paths won't break anyone... It does seem like some unexpected behavior... For example:
```
curl --location 'http://localhost:9650/ext/bc/X/walllet' \
--data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"avm.getHeight"
}'
```
could report
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32000,
        "message": "rpc: can't find service \"avm.GetHeight\"",
        "data": null
    },
    "id": 1
}
```
when it should _really_ be reporting 404 (because there was a typo of `walllet` rather than `wallet`)

It feels like we should enforce proper routing if possible.

vms/example/xsvm/api/ping.go
StephenButtolph: The `xsvm.UnsafePingServer` was only required because gRPC requires embedding either the safe or unsafe version. It seems like connect doesn't require that.

vms/rpcchainvm/ghttp/greader/greader_test.go
```diff
@@ -45,6 +45,5 @@ func TestErrIOEOF(t *testing.T) {
 	buf := make([]byte, 1)
 	n, err := client.Read(buf)
 	require.Zero(n)
-	// Do not use require.ErrorIs because callers use equality checks on io.EOF
```
StephenButtolph: This diff should probably be removed.

vms/rpcchainvm/vm_client.go
```diff
@@ -370,42 +370,21 @@ func (vm *VMClient) Shutdown(ctx context.Context) error {
...
 	clientConn, err := grpcutils.Dial(resp.ServerAddr)
 	if err != nil {
 		return nil, err
 	}
 
 	vm.conns = append(vm.conns, clientConn)
-	return ghttp.NewClient(httppb.NewHTTPClient(clientConn)), nil
+	handler := ghttp.NewClient(httppb.NewHTTPClient(clientConn))
+
+	return handler, nil
```
StephenButtolph: reduce diff:
```suggestion
	return ghttp.NewClient(httppb.NewHTTPClient(clientConn)), nil
```

vms/rpcchainvm/vm_server.go
```diff
@@ -333,56 +333,24 @@ func (vm *VMServer) Shutdown(ctx context.Context, _ *emptypb.Empty) (*emptypb.Em
 	return &emptypb.Empty{}, errs.Err
 }
 
-func (vm *VMServer) CreateHandlers(ctx context.Context, _ *emptypb.Empty) (*vmpb.CreateHandlersResponse, error) {
-	handlers, err := vm.vm.CreateHandlers(ctx)
+func (vm *VMServer) NewHTTPHandler(ctx context.Context, _ *emptypb.Empty) (*vmpb.NewHTTPHandlerResponse, error) {
+	handler, err := vm.vm.NewHTTPHandler(ctx)
```
StephenButtolph, line 349: Do we require all VMs to return an http.Handler now? We used to support not specifying any.

vms/rpcchainvm/ghttp/greader/greader_test.go
joshua-kim: Ugh bad rebase. Will remove.

vms/rpcchainvm/vm_server.go
joshua-kim: fixed + added a regression test

snow/engine/common/vm.go
joshua-kim: Fixed comment

api/server/router.go
joshua-kim: It looks like this is safe since there's a mutex internally in the router and we can register these paths dynamically... I removed this lock + the related hacky locked interfaces

connectproto/buf.gen.yaml
joshua-kim: Follow-up issue: https://github.com/ava-labs/avalanchego/issues/4066

vms/avm/vm.go
joshua-kim: Updated to support old interface signature for path-based routing

tests/e2e/vms/xsvm.go
joshua-kim: Added a helper

api/server/router.go
joshua-kim: Updated + added a regression test

api/connectclient/client.go
```diff
@@ -0,0 +1,41 @@
...
+		conn := next(ctx, spec)
+		conn.RequestHeader().Set(server.HTTPHeaderRoute, s.Route)
+		return conn
+	}
+}
+
+func (SetRouteHeaderInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+	return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
+		return next(ctx, conn)
+	}
```
StephenButtolph:
```suggestion
	return next
```

api/server/router.go
```diff
@@ -61,6 +89,16 @@ func (r *router) GetHandler(base, endpoint string) (http.Handler, error) {
 	return handler, nil
 }
 
+func (r *router) AddHeaderRoute(route string, handler http.Handler) bool {
```
StephenButtolph, line 92: Doesn't this need locking?

go.mod
```diff
@@ -79,6 +81,8 @@ require (
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 )
 
+require github.com/ava-labs/firewood-go-ethhash/ffi v0.0.8 // indirect
```
StephenButtolph: Why is this in a separate `require` statement? Shouldn't this be with the other indirect dependencies?

utils/resource/usage.go
```diff
@@ -66,6 +67,26 @@ type Manager interface {
 	Shutdown()
 }
 
+type NoManager struct{}
```
StephenButtolph: Why is this added in this PR?

snow/engine/common/vm.go
```diff
@@ -64,6 +64,8 @@ type VM interface {
 	// Version returns the version of the VM.
 	Version(context.Context) (string, error)
 
+	// Deprecated: implementations should expose apis through NewHTTPHandler.
+	//
```
StephenButtolph: I don't think we can realistically deprecate this.
```suggestion
```

go.mod
joshua-kim: For some reason my go.mod automatically put this in here... reformatted the file

snow/engine/common/vm.go
joshua-kim: My intention was to say that future APIs should be exposed through `NewHTTPHandler`... not to say that we would eventually remove this. If this is confusing I can remove it.

api/connectclient/client.go
joshua-kim: 😓

utils/resource/usage.go
joshua-kim: I added this to test `NewHTTPHandler`'s implementation in `rpcchainvm` since it needs a `resource.Manager` to initialize... but it seems like I can actually just set this to `nil` since it's not used in that code path and it works fine. Removed.

api/server/router.go
joshua-kim: Whoops, yeah this does.

api/server/router.go
```diff
@@ -61,6 +89,19 @@ func (r *router) GetHandler(base, endpoint string) (http.Handler, error) {
...
+	r.routeLock.Lock()
+	defer r.routeLock.Unlock()
```
StephenButtolph: When we are reading from `headerRoutes`, we hold `r.lock`. When we are writing to `headerRoutes` we are holding `r.routeLock`. I think we need to hold `r.lock` here right?
```suggestion
	r.lock.Lock()
	defer r.lock.Unlock()
```

utils/resource/usage.go
```diff
@@ -65,7 +65,6 @@ type Manager interface {
 	// Shutdown allocated resources and stop tracking all processes.
 	Shutdown()
 }
-
 type manager struct {
```
StephenButtolph: Seems like an unrelated diff now
```suggestion

type manager struct {
```

PR #4037, opened 2025-06-26
Update golangci-lint to v2 

joshua-kim:
## Why this should be merged

The latest version of golangci-lint (which is on homebrew/nix) uses v2, which is not compatible with the v1 schema

## How this works

updates our schema using `golangci-lint migrate`.

## How this was tested

Ran `golangci-lint --fix`

Yes

### DIFF
--- a/vms/rpcchainvm/ghttp/gconn/gconn_test.go
+++ b/vms/rpcchainvm/ghttp/gconn/gconn_test.go
@@ -37,7 +37,7 @@ func TestErrIOEOF(t *testing.T) {
 	connpb.RegisterConnServer(server, connServer)
 
 	go func() {
-		require.NoError(server.Serve(listener))
+		_ = server.Serve(listener)
 	}()
 
 	grpcConn, err := grpc.DialContext(context.Background(), "bufnet",
@@ -79,7 +79,7 @@ func TestOSErrDeadlineExceeded(t *testing.T) {
 	connpb.RegisterConnServer(server, connServer)
 
 	go func() {
-		require.NoError(server.Serve(listener))
+		_ = server.Serve(listener)
 	}()
 
 	grpcConn, err := grpc.DialContext(context.Background(), "bufnet",
--- a/vms/rpcchainvm/ghttp/greader/greader_test.go
+++ b/vms/rpcchainvm/ghttp/greader/greader_test.go
@@ -29,7 +29,7 @@ func TestErrIOEOF(t *testing.T) {
 	reader.RegisterReaderServer(server, readerServer)
 
 	go func() {
-		require.NoError(server.Serve(listener))
+		_ = server.Serve(listener)
 	}()
 
 	conn, err := grpc.DialContext(context.Background(), "bufnet",


.golangci.yml
```diff
@@ -1,47 +1,10 @@
-# https://golangci-lint.run/usage/configuration/
+version: "2"
 run:
-  timeout: 10m
```
joshua-kim, line 3: v1 had a default timeout of 1m so we had to bump it up since our linting would sometimes take a while. V2 has no timeout by default, so this is no longer an issue. You can configure the tiemout via a parameter when running `golangci-lint` now

.golangci.yml
```diff
@@ -76,150 +32,145 @@ linters:
...
-            desc: github.com/stretchr/testify/require should be used instead.
-          - pkg: "io/ioutil"
-            desc: io/ioutil is deprecated. Use package io or os instead.
-  errorlint:
-    # Check for plain type assertions and type switches.
-    asserts: false
-    # Check for plain error comparisons.
-    comparison: false
-  forbidigo:
-    # Forbid the following identifiers (list of regexp).
```
joshua-kim, line 108: Comments aren't migrated by `migrate`... so I could migrate the comments by-hand if someone requests me to.

.golangci.yml
```diff
@@ -51,20 +14,13 @@ linters:
...
-    - gci
     - goconst
     - gocritic
-    # - goerr113
-    - gofmt
-    - gofumpt
-    # - gomnd
     - goprintffuncname
     - gosec
-    - gosimple
```
joshua-kim, line 63: A lot of these linters (`gosimple`, `stylecheck`, etc) got merged into `staticcheck` as part of v2

maru-ava:
Good to see this update, but installing things from homebrew is not ideal. We want to enable local reproducibility of CI, and that requires ensuring the same versions in both places and we don't use homebrew in CI.

Any interest in porting over the work I did in hypersdk to enable versioned install of golang tools?

https://github.com/ava-labs/hypersdk/pull/2030

joshua-kim:
Yeah I realized that this was also breaking in nix since it seems like the version of golangci-lint on nix is also on v2. Our linter script works because it uses go install on the v1 version. I do find this inconsistent behavior annoying so I do think it would be nice if we could have versioned installs.

maru-ava:
Would it make sense to break this PR up so that a given linter and its changes could be reviewed one-by-one (i.e. by upgrading in the initial PR but disabling failing linters)? I worry that the lack of traceability for any given change makes it all but impossible for a reviewer to do a good job.

joshua-kim:
Something like that could be possible. I could probably do something where we have a initial PR to migrate + any failing linters disabled, and have subsequent stacked PRs per-linter that we fix and merge down into this branch. Is that what you had in mind?

.github/workflows/ci.yml
```diff
@@ -146,7 +146,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: umbrelladocs/action-linkspector@a0567ce1c7c13de4a2358587492ed43cab5d0102 #v1.3.4
+      - uses: umbrelladocs/action-linkspector@v1.2.5
```
joshua-kim: So this line looks pretty weird. This seems to be an issue on v1.3 of this action ([ref](https://github.com/UmbrellaDocs/action-linkspector/issues/30)). Seems like the work-around it to downgrade in the meantime. It seems like this failed on this PR because this action only runs when *.yml and *.md files are changed by default. I'm not entirely sure why it broke on this specific PR, since we have made changes to other *.md and *.yml files before.

I re-ran the job MANY times and it did not pass until this was downgraded... so it does seem like something is actually wrong here.

.golangci.yml
```diff
@@ -75,151 +31,146 @@ linters:
     - predeclared
     - revive
     - spancheck
-    - staticcheck
-    - stylecheck
+#    - staticcheck
```
joshua-kim: PR: https://github.com/ava-labs/avalanchego/pull/4039

.golangci.yml
```diff
@@ -75,151 +31,146 @@ linters:
...
     - tagalign
-    - testifylint
-    - typecheck
+#    - testifylint
```
joshua-kim: PR: https://github.com/ava-labs/avalanchego/pull/4038

.github/workflows/ci.yml
maru-ava: Please update to use the same `SHA #version` format used for the previous version.

.golangci.yml
```diff
@@ -1,47 +1,10 @@
-# https://golangci-lint.run/usage/configuration/
```
maru-ava, line 1: Why is this link removed? It seems generally useful.

StephenButtolph (approved):

PR #4038, opened 2025-06-26
fix testifylint

joshua-kim:
## Why this should be merged

Fixes `testifylint` for golangci v2.

## How this works

Fixes the lint errors

## How this was tested

lint passes


network/p2p/p2ptest/client.go
joshua-kim, line 1: If anyone has better ideas on how to keep these error checks... let me know... but I felt like it wasn't worth the hassle.

network/network_test.go
```diff
@@ -288,42 +288,37 @@ func newFullyConnectedTestNetwork(t *testing.T, handlers []router.InboundHandler
...
+	require.NoError(eg.Wait())
 }
 
 func TestIngressConnCount(t *testing.T) {
 	require := require.New(t)
 
 	emptyHandler := func(context.Context, message.InboundMessage) {}
 
-	_, networks, wg := newFullyConnectedTestNetwork(
+	_, networks, _ := newFullyConnectedTestNetwork(
```
StephenButtolph: The prior code here was definitely wrong, rather than deleting the `eg` entirely, I think we should add:
```go
for _, net := range networks {
	net.StartClose()
}
require.NoError(eg.Wait())
```
to the end of this test.

network/peer/message_queue_test.go
```diff
@@ -56,23 +56,19 @@ func TestMessageQueue(t *testing.T) {
...
-		require.False(ok)
-		close(done)
+		gotOk <- q.Push(ctx, msgs[0])
 	}()
-	<-done
+	require.False(<-gotOk)
 
 	// Assert that Push returns false when the queue is closed
-	done = make(chan struct{})
+	gotOk = make(chan bool)
```
StephenButtolph: I think the existing `gotOk` chan is sufficient here.
```suggestion
```

tests/e2e/vms/xsvm.go
```diff
@@ -226,40 +226,46 @@ var _ = ginkgo.Describe("[XSVM]", ginkgo.Label("xsvm"), func() {
...
-		go func() {
-			defer func() {
-				ginkgo.GinkgoRecover()
-				wg.Done()
-			}()
+			return nil
+		})
+
+		eg.Go(func() error {
+			defer ginkgo.GinkgoRecover()
```
StephenButtolph: Do we still need to call `ginkgo.GinkgoRecover()` since we are no longer calling `require` in this goroutine?

PR #4039, opened 2025-06-26
Golangci migrate fix staticcheck

joshua-kim:
## Why this should be merged

Fixes staticcheck on golangci-lint v2

## How this works

Fixes linter errors

## How this was tested

Ran linter

No


api/admin/service.go
```diff
@@ -391,7 +391,7 @@ type DBGetReply struct {
 	ErrorCode rpcdbpb.Error `json:"errorCode"`
 }
 
-//nolint:stylecheck // renaming this method to DBGet would change the API method from "dbGet" to "dBGet"
```
joshua-kim, line 394: stylecheck no longer exists and has now been merged into staticcheck

app/app.go
```diff
@@ -47,7 +47,7 @@ type App interface {
 
 func New(config nodeconfig.Config) (App, error) {
 	// Set the data directory permissions to be read write.
-	if err := perms.ChmodR(config.DatabaseConfig.Path, true, perms.ReadWriteExecute); err != nil {
```
maru-ava: Was it an accident that there are so many instances of explicit references to embedded types? Or are we just assuming that the golangci-lint check is somehow 'right'?

vms/propertyfx/mint_operation.go
```diff
@@ -36,8 +36,8 @@ func (op *MintOperation) Outs() []verify.State {
 }
 
 func (op *MintOperation) Verify() error {
-	switch {
-	case op == nil:
+	switch op {
```
maru-ava: Is this actually a requirement we want/need to enforce? As with the use of embedded types, this is functionally equivalent as far as golang is concerned.

vms/avm/txs/executor/syntactic_verifier.go
```diff
@@ -104,7 +104,7 @@ func (v *SyntacticVerifier) CreateAssetTx(tx *txs.CreateAssetTx) error {
 	}
 
 	for _, r := range tx.Name {
-		if r > unicode.MaxASCII || !(unicode.IsLetter(r) || unicode.IsNumber(r) || r == ' ') {
+		if r > unicode.MaxASCII || (!unicode.IsLetter(r) && !unicode.IsNumber(r) && r != ' ') {
```
maru-ava, line 107: Why do we want to enforce this style if a) both forms are functionality equivalent and b) we should be testing the result anyway.

vms/propertyfx/mint_operation.go
joshua-kim: We don't have to require it because it is functionally equivalent as you mentioned. This is probably a bad example in the diff, but the [docs](https://staticcheck.dev/docs/checks/#QF1002) include a better one where it does clean up the switch more.

Before:
```
switch {
case x == 1 || x == 2, x == 3:
    ...
case x == 4:
    ...
default:
    ...
}
```
After:
```
switch x {
case 1, 2, 3:
    ...
case 4:
    ...
default:
    ...
}
```

app/app.go
joshua-kim: I'm not sure what the historical reason is... maybe we had conflicts in the past and removed them - or maybe the simple answer is that we just explicitly referenced the embedded type for no reason. 

w.r.t Verifying its correctness, I think we would be concerned if there was a conflict in the type + embedded type and behavior was different for this. I generally trust golangci-lint so I didn't double-check every change.

maru-ava: I'm less accepting of seemingly arbitrary linting rules. I'd like to know there's a good reason for forcing people to write code a specific way, especially given how simple golang is as a language. I don't see why it's worth requiring this specific rule.

But this isn't really a technical issue, so I'll defer to Stephen.

maru-ava:
My comments aren't intended to block merge. For all 3 of them, it comes down to what style rules we want to have enforced rather than any technical criteria, so I defer to Stephen's judgement.

app/app.go
StephenButtolph: This is definitely an opinion question... So I can say mine, but not sure where that leaves us.

For my self, I find embedding pretty subtle. I generally like explicitly referencing the embedded field to make it clear that the call is not being made on ourself, but on an inner type. I especially find this nice in places like the proposervm/tracevm/metervm. It feels like those places are ripe for an accidental infinite recursion.

vms/propertyfx/mint_operation.go
StephenButtolph: These switch case changes are pretty dubious... The reason they were done the way they were before was to be consistent with the other `Verify` functions which have many verification conditions... If we don't care about consistency here, then I think basically all the cases where the linter suggested changes should be converted into `if else` checks.

vms/avm/txs/executor/syntactic_verifier.go
joshua-kim: My guess is that the linter thinks that Demorgan's law is either idiomatic, is commonly understood, or is more readable. I'm down to disable it... but if there isn't a strong reason for/against I generally default to whatever comes out-of-the-box because I just dislike having custom configurations to maintain.

vms/propertyfx/mint_operation.go
joshua-kim: Swapped out w/ if-else checks where it made sense

vms/avm/txs/executor/syntactic_verifier.go
joshua-kim: For now going to keep the Demorgan's law linter since it's the default unless anyone has a strong opinion on it

app/app.go
joshua-kim: Summary of an offline discussion but I think we should just use embedding less. We use it liberally as poor-man's inheritance but when we're frequently changing the embedded method behavior it feels like a code smell. In these cases we should probably just use a field and make passthroughs to satisfy the interface.

vms/nftfx/transfer_operation.go
```diff
@@ -31,8 +31,8 @@ func (op *TransferOperation) Outs() []verify.State {
 }
 
 func (op *TransferOperation) Verify() error {
-	switch {
-	case op == nil:
+	switch op {
```
StephenButtolph: This should also be an `if else` imo

StephenButtolph (approved):

PR #4043, opened 2025-06-28
chore: add reference to fake exponential formula

lechpzn:


StephenButtolph:
I don't think this link really adds anything. The entire pseudocode is already in the comment (and there is already the attribution as well).

PR #4044, opened 2025-06-30
feat: add default flag options

RodrigoVillar:

### DIFF
--- a/tests/fixture/e2e/flags.go
+++ b/tests/fixture/e2e/flags.go
@@ -124,11 +124,38 @@ func (v *FlagVars) ActivateGranite() bool {
 	return v.activateGranite
 }
 
-func RegisterFlags() *FlagVars {
-	return RegisterFlagsWithDefaultOwner("")
+type DefaultOptions struct {
+	owner     string
+	nodeCount int
 }
 
-func RegisterFlagsWithDefaultOwner(defaultOwner string) *FlagVars {
+type DefaultOption func(*DefaultOptions)
+
+func NewDefaultOptions(options []DefaultOption) *DefaultOptions {
+	d := &DefaultOptions{}
+	for _, do := range options {
+		do(d)
+	}
+	return d
+}
+
+func WithDefaultOwner(owner string) DefaultOption {
+	return func(d *DefaultOptions) {
+		d.owner = owner
+	}
+}
+
+func WithDefaultNodeCount(nodeCount int) DefaultOption {
+	return func(d *DefaultOptions) {
+		d.nodeCount = nodeCount
+	}
+}
+
+func RegisterFlags() *FlagVars {
+	return RegisterFlagsWithOptions(WithDefaultNodeCount(tmpnet.DefaultNodeCount))
+}
+
+func RegisterFlagsWithOptions(ops ...DefaultOption) *FlagVars {
 	vars := FlagVars{}
 
 	flag.BoolVar(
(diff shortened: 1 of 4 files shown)


tests/fixture/e2e/flags.go
```diff
@@ -124,11 +124,50 @@ func (v *FlagVars) ActivateGranite() bool {
...
+	nodeCount int
+}
+
+func NewDefaultOptions(ops []DefaultOption) *DefaultOptions {
+	o := &DefaultOptions{}
+	o.applyOptions(ops)
+	return o
+}
+
+func (d *DefaultOptions) applyOptions(ops []DefaultOption) {
```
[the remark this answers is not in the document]
RodrigoVillar: Unexporting `NewDefaultOptions()` since only `RegisterFlagsWithOptions()` is meant to call it. Keeping the slice of `Option` since `RegisterFlagsWithOptions()` is variadic: 649bf44

tests/fixture/e2e/flags.go
```diff
@@ -124,11 +124,46 @@ func (v *FlagVars) ActivateGranite() bool {
...
+	}
+}
+
+func WithDefaultNodeCount(nodeCount int) DefaultOption {
+	return func(d *DefaultOptions) {
+		d.nodeCount = nodeCount
+	}
+}
+
+func RegisterFlagsWithOptions(ops ...DefaultOption) *FlagVars {
```
maru-ava: If there is no RegisterFlags alternative, maybe it's unnecessary to suffix with `WithOptions`?

tests/fixture/e2e/flags.go
```diff
@@ -124,11 +124,46 @@ func (v *FlagVars) ActivateGranite() bool {
 	return v.activateGranite
 }
 
-func RegisterFlags() *FlagVars {
-	return RegisterFlagsWithDefaultOwner("")
+type DefaultOption func(*DefaultOptions)
```
maru-ava, line 127: Does the option machinery need to be exported?

tests/fixture/e2e/flags.go
```diff
@@ -124,11 +124,46 @@ func (v *FlagVars) ActivateGranite() bool {
...
+	}
+	return o
+}
+
+func (d *DefaultOptions) Owner() string {
+	return d.owner
+}
+
+func (d *DefaultOptions) NodeCount() int {
+	if d.nodeCount == 0 {
```
maru-ava: Maybe `<=` would make more sense?

RodrigoVillar: Done: 9cbbc5d

tests/fixture/e2e/flags.go
RodrigoVillar: Done: 79d9d60

tests/fixture/e2e/flags.go
RodrigoVillar: I would prefer to keep `DefaultOption` exported. We export instances of this type via:
- `WithX()` functions which return a `DefaultOption`
- `RegisterFlags()` which takes several `DefaultOption`

This also follows the convention for options in the AvalancheGo repo - as an example:
https://github.com/ava-labs/avalanchego/blob/66a9447132fcc059d4c02901944b971ba07c8f99/wallet/subnet/primary/common/options.go#L44

maru-ava: I'm not sure the wallet implementation is relevant here. It's not necessary to export a type to return it from an exported function which is how clients are expected to create options. Exporting only permits something other than this package to create instances of the type which is arguably not needed.

tests/fixture/e2e/flags.go
```diff
@@ -124,11 +124,46 @@ func (v *FlagVars) ActivateGranite() bool {
 	return v.activateGranite
 }
 
-func RegisterFlags() *FlagVars {
-	return RegisterFlagsWithDefaultOwner("")
+type DefaultOption func(*DefaultOptions)
+
+type DefaultOptions struct {
```
maru-ava, line 129: (No action required) I still think this a pretty complicated alternative to using a struct.

```golang
type Defaults struct {
  Owner     string
  NodeCount int
}
...
flags.RegisterFlags(flags.Defaults{})
```

PR #4050, opened 2025-07-02
[nix] Update to 25.05

maru-ava:
## Why this should be merged

A given nix release only receives security updates for 7 months, and though security updates aren't critical for our tooling, there's no reason not to upgrade regularly.

https://nixos.org/blog/announcements/2025/nixos-2505/

## How this was tested

CI

N/A

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -199,11 +199,12 @@ tasks:
       - cmd: bash -x ./scripts/tests.e2e.kube.sh {{.CLI_ARGS}}
 
   test-e2e-kube-ci:
-    desc: Runs e2e tests against a network deployed to kube [serially]
+    # Free github action runners do not have sufficient resources to reliably run a full e2e run against a kube-hosted network
+    desc: Runs the xsvm e2e tests in serial against a network deployed to kube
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -203,7 +203,7 @@ tasks:
     env:
       E2E_SERIAL: 1
     cmds:
-      - task: test-e2e-kube
+      - cmd: bash -x ./scripts/tests.e2e.kube.sh {{.CLI_ARGS}}
```
maru-ava: I noticed this needed to be changed when the job flaked in CI on this PR.  Our e2e jobs can't be reliably run in parallel on github action runners, and E2E_SERIAL is how we ensure serial execution. But since tasks don't inherit the environment of a calling task, this wasn't previously configured correctly. Calling the script directly ensures that E2E_SERIAL will be respected.

maru-ava:
Updated to limit the scope of the e2e-kube job to just xsvm (to ensure subnet coverage) to avoid the flakiness caused by the resource-constrained github runners. In any case, the kube job is now disabled.

PR #4051, opened 2025-07-03
ci(load-tests): add GitHub Actions workflow for load testing on k8s

Elvis339:
## Why this should be merged
Enables CI execution on platform-chaos-k8s addressing three critical EVM performance testing requirements: 
1. Access to larger nodes than GitHub provides
2. Controlled latency simulation between nodes using chaos mesh for geographic distribution testing
3. Secure cluster access without exposing credentials to GitHub Actions. 

This foundation supports the EVM performance testing initiative and future regression analysis capabilities.

## How this works
GitHub workflows can now leverage platform-chaos-k8s infrastructure for EVM load testing. Workflows automatically provision test environments. This provides a self-service testing environment without requiring direct cluster access or manual infrastructure setup.

## How this was tested
- https://github.com/ava-labs/avalanchego/actions/runs/15978698247/job/45067717343

No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -87,7 +87,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.1 // indirect
 	github.com/VictoriaMetrics/fastcache v1.12.1 // indirect
 	github.com/ava-labs/firewood-go-ethhash/ffi v0.0.8 // indirect
-	github.com/ava-labs/simplex v0.0.0-20250626192006-220e6aeacdc1
+	github.com/ava-labs/simplex v0.0.0-20250715173145-e4fe035cb9b2
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bits-and-blooms/bitset v1.10.0 // indirect
 	github.com/btcsuite/btcd/btcec/v2 v2.3.2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -78,8 +78,8 @@ github.com/ava-labs/ledger-avalanche/go v0.0.0-20241009183145-e6f90a8a1a60 h1:EL
 github.com/ava-labs/ledger-avalanche/go v0.0.0-20241009183145-e6f90a8a1a60/go.mod h1:/7qKobTfbzBu7eSTVaXMTr56yTYk4j2Px6/8G+idxHo=
 github.com/ava-labs/libevm v1.13.14-0.3.0.rc.1 h1:vBMYo+Iazw0rGTr+cwjkBdh5eadLPlv4ywI4lKye3CA=
 github.com/ava-labs/libevm v1.13.14-0.3.0.rc.1/go.mod h1:+Iol+sVQ1KyoBsHf3veyrBmHCXr3xXRWq6ZXkgVfNLU=
-github.com/ava-labs/simplex v0.0.0-20250626192006-220e6aeacdc1 h1:ipeWExRrhYF7DZ/bcigoQrzo3vZWNZrFx8W+Yg2sJ2Q=
-github.com/ava-labs/simplex v0.0.0-20250626192006-220e6aeacdc1/go.mod h1:GVzumIo3zR23/qGRN2AdnVkIPHcKMq/D89EGWZfMGQ0=
+github.com/ava-labs/simplex v0.0.0-20250715173145-e4fe035cb9b2 h1:PZ5PMEDkTbd6NLNiwKWV8nz7QvAM+QC9Rj3/NrL9ICA=
+github.com/ava-labs/simplex v0.0.0-20250715173145-e4fe035cb9b2/go.mod h1:GVzumIo3zR23/qGRN2AdnVkIPHcKMq/D89EGWZfMGQ0=
 github.com/aymerick/raymond v2.0.3-0.20180322193309-b565731e1464+incompatible/go.mod h1:osfaiScAUVup+UC9Nfq76eWqDhXlp+4UYaA8uhTBO6g=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
--- a/message/messagemock/outbound_message_builder.go
+++ b/message/messagemock/outbound_message_builder.go
@@ -390,6 +390,21 @@ func (mr *OutboundMsgBuilderMockRecorder) Put(chainID, requestID, container any)
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Put", reflect.TypeOf((*OutboundMsgBuilder)(nil).Put), chainID, requestID, container)
 }
 
+// SimplexMessage mocks base method.
+func (m *OutboundMsgBuilder) SimplexMessage(msg *p2p.Simplex) (message.OutboundMessage, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "SimplexMessage", msg)
+	ret0, _ := ret[0].(message.OutboundMessage)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// SimplexMessage indicates an expected call of SimplexMessage.
+func (mr *OutboundMsgBuilderMockRecorder) SimplexMessage(msg any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SimplexMessage", reflect.TypeOf((*OutboundMsgBuilder)(nil).SimplexMessage), msg)
+}
+
 // StateSummaryFrontier mocks base method.
 func (m *OutboundMsgBuilder) StateSummaryFrontier(chainID ids.ID, requestID uint32, summary []byte) (message.OutboundMessage, error) {
 	m.ctrl.T.Helper()
--- a/message/outbound_msg_builder.go
+++ b/message/outbound_msg_builder.go
@@ -181,6 +181,10 @@ type OutboundMsgBuilder interface {
 		chainID ids.ID,
 		msg []byte,
 	) (OutboundMessage, error)
+
+	SimplexMessage(
+		msg *p2p.Simplex,
+	) (OutboundMessage, error)
 }
 
 type outMsgBuilder struct {
@@ -719,6 +723,18 @@ func (b *outMsgBuilder) AppGossip(chainID ids.ID, msg []byte) (OutboundMessage, 
 					ChainId:  chainID[:],
 					AppBytes: msg,
 				},
+			},
+		},
+		b.compressionType,
+		false,
+	)
+}
+
+func (b *outMsgBuilder) SimplexMessage(msg *p2p.Simplex) (OutboundMessage, error) {
+	return b.builder.createOutbound(
+		&p2p.Message{
+			Message: &p2p.Message_Simplex{
+				Simplex: msg,
 			},
 		},
 		b.compressionType,
--- a/simplex/block.go
+++ b/simplex/block.go
@@ -67,7 +67,7 @@ type blockDeserializer struct {
 	parser block.Parser
 }
 
-func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error) {
+func (d *blockDeserializer) DeserializeBlock(ctx context.Context, bytes []byte) (simplex.Block, error) {
 	var canotoBlock canotoSimplexBlock
 
 	if err := canotoBlock.UnmarshalCanoto(bytes); err != nil {
@@ -79,7 +79,7 @@ func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error
 		return nil, fmt.Errorf("failed to parse protocol metadata: %w", err)
 	}
 
-	vmblock, err := d.parser.ParseBlock(context.TODO(), canotoBlock.InnerBlock)
+	vmblock, err := d.parser.ParseBlock(ctx, canotoBlock.InnerBlock)
 	if err != nil {
 		return nil, err
 	}
--- a/simplex/block_test.go
+++ b/simplex/block_test.go
@@ -21,7 +21,7 @@ import (
 
 func TestBlockSerialization(t *testing.T) {
 	unexpectedBlockBytes := errors.New("unexpected block bytes")
-
+	ctx := context.Background()
 	testBlock := snowmantest.BuildChild(snowmantest.Genesis)
 
 	b := &Block{
@@ -88,7 +88,7 @@ func TestBlockSerialization(t *testing.T) {
 			}
 
 			// Deserialize the block
-			deserializedBlock, err := deserializer.DeserializeBlock(tt.blockBytes)
+			deserializedBlock, err := deserializer.DeserializeBlock(ctx, tt.blockBytes)
 			require.ErrorIs(t, err, tt.expectedError)
 
 			if tt.expectedError == nil {
--- a/simplex/bls_test.go
+++ b/simplex/bls_test.go
@@ -13,8 +13,7 @@ import (
 )
 
 func TestBLSVerifier(t *testing.T) {
-	config, err := newEngineConfig()
-	require.NoError(t, err)
+	config := newEngineConfig(t, 1)
 	signer, verifier := NewBLSAuth(config)
 	otherNodeID := ids.GenerateTestNodeID()
 
@@ -81,7 +80,7 @@ func TestBLSVerifier(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err = verifier.Verify(msg, tt.sig, tt.nodeID)
+			err := verifier.Verify(msg, tt.sig, tt.nodeID)
 			require.ErrorIs(t, err, tt.expectErr)
 		})
 	}
--- a/simplex/comm.go
+++ b/simplex/comm.go
@@ -1,0 +1,138 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package simplex
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/simplex"
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/message"
+	"github.com/ava-labs/avalanchego/proto/pb/p2p"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/snow/networking/sender"
+	"github.com/ava-labs/avalanchego/subnets"
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+var (
+	_               simplex.Communication = (*Comm)(nil)
+	errNodeNotFound                       = errors.New("node not found in the validator list")
+)
+
+type Comm struct {
+	logger   simplex.Logger
+	subnetID ids.ID
+	chainID  ids.ID
+	// broadcastNodes are the nodes that should receive broadcast messages
+	broadcastNodes set.Set[ids.NodeID]
+	// allNodes are the IDs of all the nodes in the subnet
+	allNodes []simplex.NodeID
+
+	// sender is used to send messages to other nodes
+	sender     sender.ExternalSender
+	msgBuilder message.OutboundMsgBuilder
+}
+
+func NewComm(config *Config) (*Comm, error) {
+	if _, ok := config.Validators[config.Ctx.NodeID]; !ok {
+		config.Log.Warn("Node is not a validator for the subnet",
+			zap.Stringer("nodeID", config.Ctx.NodeID),
+			zap.Stringer("chainID", config.Ctx.ChainID),
+			zap.Stringer("subnetID", config.Ctx.SubnetID),
+		)
+		return nil, fmt.Errorf("our %w: %s", errNodeNotFound, config.Ctx.NodeID)
+	}
+
+	broadcastNodes := set.NewSet[ids.NodeID](len(config.Validators) - 1)
+	allNodes := make([]simplex.NodeID, 0, len(config.Validators))
+	// grab all the nodes that are validators for the subnet
+	for _, vd := range config.Validators {
+		allNodes = append(allNodes, vd.NodeID[:])
+		if vd.NodeID == config.Ctx.NodeID {
+			continue // skip our own node ID
+		}
+
+		broadcastNodes.Add(vd.NodeID)
+	}
+
+	return &Comm{
+		subnetID:       config.Ctx.SubnetID,
+		broadcastNodes: broadcastNodes,
+		allNodes:       allNodes,
+		logger:         config.Log,
+		sender:         config.Sender,
+		msgBuilder:     config.OutboundMsgBuilder,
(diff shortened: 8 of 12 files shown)


.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+# This workflow runs comprehensive load tests against our Kubernetes cluster
+on:
+  workflow_dispatch:
+
+jobs:
+  load_tests:
+    name: Load Tests
+    runs-on: avalanche-avalanchego-runner
+    container:
+      image: ghcr.io/actions/actions-runner:2.325.0
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/avalanchego/pull/4051/commits/2654ccc92e1f706f3e5e09311c43550091f4ee68

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,34 @@
+name: Load Tests
+
+# This workflow runs comprehensive load tests against our Kubernetes cluster
+on:
```
maru-ava: Maybe add inputs for the load test duration, the image to use, and how many nodes to deploy?

The antithesis trigger workflow shows examples of usage: https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/trigger-antithesis-avalanchego.yml#L6

@RodrigoVillar Anything else that should be configurable?

Elvis339: I can open a new PR to add more inputs once [#4044](https://github.com/ava-labs/avalanchego/pull/4044) merges.
Regarding the image tag I'd make it optional since manual workflow dispatch already lets us select the branch..

If we add `image_tag` the command becomes:
`go run ./tests/load/c/main --runtime=kube --kube-image <tag>`
This would require an additional step to build the image if it doesn't exist. I think choosing the branch might be more intuitive since people can just select the branch they want to test. Happy to change if there's a strong need for testing against different AvalancheGo versions.

RodrigoVillar: I agree that there should be some configuration for the number of nodes to deploy and whether the load test should run indefinitely (the load test doesn't support a test duration without failing). 

I would also like to see configuration for the test params (e.g. `minTPS`, `maxTPS`, etc.) but this can be deferred to a future PR.

.github/workflows/load-tests.yaml
RodrigoVillar, line 1: (no action necessary)  At some point, it would be great to add documentation for how to run the action.

.github/workflows/load-tests.yaml
Elvis339: Once we have these options, I'll add them here.

RodrigoVillar (approved):

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Setup Go for project
+        uses: ./.github/actions/setup-go-for-project
+
+      - name: Install Nix
+        uses: ./.github/actions/install-nix
+
+      - name: Run load tests
+        shell: nix develop --command bash -x {0}
+        run: |
+          go run ./tests/load/c/main --runtime=kube
```
maru-ava: Seeing this again, I think this invocation should be in a task to ensure local reproducibility. But, the current task (test-load-kube) is typesquatting the name we'd want to use. Maybe rename the existing task to `test-load-kube-kind` and make it clear in the `desc` fields that `test-load-kube` is for targeting an already-running kube cluster and `test-load-kube-kind` is for targeting a local kind cluster?

.github/workflows/load-tests.yaml
maru-ava: Maybe to tests/load/README.md?

.github/workflows/load-tests.yaml
Elvis339: In this [commit](https://github.com/ava-labs/avalanchego/pull/4051/commits/422d8363f982b5260698fc87b263ff94f264d537) you can find updates to the Kubernetes load testing setup.

- Single script now handles both Kind and remote clusters via `SKIP_KIND_CLUSTER` environment variable
- Automatic image selection: builds locally for Kind, uses Docker Hub for remote
- Supports `--kube-image` override for both cluster types

maru-ava: Why is this preferable to just adding a task that does what appears on line 34?

Elvis339: We already use scripts throughout the codebase: test-e2e, test-upgrade, and others all call scripts rather than Go directly. This keeps things consistent with our existing patterns.

Even with context, it's hard to follow when to use which approach, what each script does, what cmd line argument to pass.. The complications get hidden in the script rather than exposed to every developer who needs to run load tests.
The script provides clear usage documentation and examples right at the top. 

The goal is to make this frictionless for developers if they have to know when to use which command line arguments, then we've failed imo.

maru-ava: go-task was introduced to be the primary entrypoint to the tooling in this repo. If there's something we can execute, exposing it via a task is the first priority. How it is implemented - golang, external script, or inline script - is not relevant.

Edit: To your point, there is a problem of a lack of documentation of tmpnet flags (something I am actually working on). But I don't think the way to solve that is piecemeal additions on a script-by-script basis. tmpnet is the common thread here, and I'd rather be able to link to its updated documentation in the taskfile than sprinkle bits and pieces here and there.

My main concern with the change you've proposed in https://github.com/ava-labs/avalanchego/pull/4051/commits/422d8363f982b5260698fc87b263ff94f264d537 is that it adds a lot of complexity to the script rather than just duplicating the invocation of the load test in the task. I think the function of the script should be limited to 'test with kind' since everything it does before starting the load test is kind-specific. 

I'm not blocking merge on this, but I will warn you that I am likely to revert the script and taskfile changes once updated tmpnet docs land.

Elvis339: I can split this into two separate scripts: tests.load.kube.kind.sh and tests.load.kube.sh. This would reduce cognitive load, make the purpose of each script clear, and eliminate the need for environment variable conditionals.

Each script would have a single, focused responsibility instead of branching logic based on SKIP_KIND_CLUSTER.

maru-ava: Why would we need a script to wrap a single golang command that could be inlined in the task?

maru-ava: Again, _tasks_ are the intended primary entrypoint into our tooling. There is no reason to add a script except where the complexity exceeds what can reasonably be defined for a task. There is also nothing preventing expanding documentation of a given task. There is even a field, `summary`, that enables expanding on what appears in `desc`.

Elvis339: Actually you're right. It doesn't make sense to optimize for local development since running load tests on platform-chaos-k8s is possible via the GitHub Actions workflow. No one will manually run load tests against different clusters locally.

But, the `tests.load.kube.sh` name is misleading since it actually creates and targets Kind clusters, not generic Kubernetes clusters.

We should rename it to `tests.load.kube.kind.sh` to be explicit about what it does. What do you think?

maru-ava: > We should rename it to tests.load.kube.kind.sh to be explicit about what it does. What do you think?

Sounds good!

.github/workflows/load-tests.yaml
Elvis339: Should REAMDE.md be moved into load2 package?

RodrigoVillar: The GH actions use the code in `tests/load` - the `README` should go in `tests/load` until we switch support to `tests/load2`

.github/workflows/ci.yml
```diff
@@ -233,7 +233,7 @@ jobs:
         shell: bash
         run: nix develop --command ./scripts/run_task.sh test-bootstrap-monitor-e2e
   load:
-    name: Run load test
+    name: Run process based load test
```
maru-ava: nit: process-based

Taskfile.yml
```diff
@@ -233,11 +233,19 @@ tasks:
...
-  test-load-kube:
-    desc: Runs load tests against a network deployed to kube
+  test-load-kube-kind:
+    desc: Runs load tests against a kind cluster
     cmds:
       - task: generate-load-contract-bindings
-      - cmd: bash -x ./scripts/tests.load.kube.sh {{.CLI_ARGS}}
+      - cmd: bash -x ./scripts/tests.load.kube.kind.sh {{.CLI_ARGS}}
+
+  test-load-kube-remote:
```
maru-ava: I don't think 'remote' is an accurate description - it could be any cluster other than the kind cluster we deploy for local and CI testing.

tests/load/README.md
```diff
@@ -114,3 +114,29 @@ for {
...
+```shell
+./scripts/run_task.sh test-load-kube-kind -- --duration=10m --nodes=3
+```
+
+### With custom image (skips building from local code)
+```shell
+./scripts/run_task.sh test-load-kube-kind -- --kube-image=my-registry/avalanchego:v1.2.3
+```
+
+## Against remote platform-chaos-k8s
```
maru-ava: Suggest 'Trigger CI job on self-hosted runner` i.e. no need to reference where it runs.

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,39 @@
+name: Load Tests on k8s
```
maru-ava: Suggest naming 'Load test on self-hosted runners' since k8s is just an implementation detail.

.github/workflows/ci.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4051/commits/d324366414f16dc4b9311ad33393b349094c71e2

tests/load/README.md
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4051/commits/451645e91dc2289aec6991f96935d08a5574d41c

.github/workflows/load-tests.yaml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4051/commits/3f174f4a83de1aeb0067a53242e999cbaeeefcb0

Taskfile.yml
Elvis339: Renamed to `test-load-kube` https://github.com/ava-labs/avalanchego/pull/4051/commits/af3b3432ee14c839508127515b38fb8bd513d0f5

Taskfile.yml
```diff
@@ -233,15 +233,23 @@ tasks:
...
+    desc: Runs load tests against a kind cluster
+    cmds:
+      - task: generate-load-contract-bindings
+      - cmd: bash -x ./scripts/tests.load.kube.kind.sh {{.CLI_ARGS}}
+
   test-load-kube:
-    desc: Runs load tests against a network deployed to kube
+    desc: Runs load tests against a kubernetes cluster
+    vars:
+      KUBE_IMAGE: '{{.KUBE_IMAGE | default "avaplatform/avalanchego:latest"}}'
```
maru-ava: I don't think it should be necessary to explicitly configure vars to supply arguments.  `task test-load-kube -- kube-image=[image]` enables the same thing.

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,39 @@
...
+    inputs:
+      image:
+        description: 'AvalancheGo image to test'
+        required: false
+        default: 'avaplatform/avalanchego:latest'
+        type: string
+
+jobs:
+  load_tests:
+    name: Run load test on platform-chaos-k8s
```
maru-ava: platform-chaos-k8s -> self-hosted runners

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go for project
+        uses: ./.github/actions/setup-go-for-project
+
+      - name: Install Nix
+        uses: ./.github/actions/install-nix
+
+      - name: Run load tests
```
maru-ava: Sorry I missed this before, but you'll want to remove nix installation above and run this step with the run-monitored-tmpnet-cmd custom action to enable metrics to be collected from the load test itself.

Taskfile.yml
Elvis339: Right, wanted to be explicit. Changed it anyway.

.github/workflows/load-tests.yaml
Elvis339: Ahh, missed it, thanks for pointing it out.

.github/workflows/load-tests.yaml
Elvis339: I had to update the [action](https://github.com/ava-labs/avalanchego/blob/be89c473b13941d43afe316556496b67b4335192/.github/actions/run-monitored-tmpnet-cmd/action.yml#L14).

https://github.com/ava-labs/avalanchego/actions/runs/16124719363/job/45498806843

Taskfile.yml
```diff
@@ -233,15 +233,21 @@ tasks:
       - task: build
       - cmd: go run ./tests/load/c/main --avalanchego-path=./build/avalanchego {{.CLI_ARGS}}
 
+  test-load-kube-kind:
```
maru-ava: Please maintain the sorted order

.github/workflows/load-tests.yaml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: avalanche-avalanchego-runner
+    container:
+      image: ghcr.io/actions/actions-runner:2.325.0
+
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y xz-utils
+
```
maru-ava: Please maintain consistency with the style of existing workflows (no blank lines between steps)

.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -51,12 +54,12 @@ runs:
     - uses: cachix/install-nix-action@02a151ada4993995686f9ed4f1be7cfbb229e56f #v31
       with:
         github_access_token: ${{ inputs.github_token }}
-    - run: ${{ github.action_path }}/nix-develop.sh --command echo "dependencies installed"
+    - run: ${{ inputs.action_path }}/nix-develop.sh --command echo "dependencies installed"
```
maru-ava: Nice work working around the path being incorrect. What is github.action_path set to without this change? 

Is there a way to ensure the runner is setting the correct action path instead of having to work around it in the action?

Elvis339: Without this change the `github.action_path` results in `/__w/_temp/1bb118b0-d006-4f54-ad28-e72e69c46333.sh: line 1: /home/runner/_work/avalanchego/avalanchego/./.github/actions/run-monitored-tmpnet-cmd/nix-develop.sh: No such file or directory` [reference](https://github.com/ava-labs/avalanchego/actions/runs/16124211189/job/45497116418).

Updating the `uses` key which is setting the `github.action_path` to `.github/actions/run-monitored-tmpnet-cmd/nix-develop.sh` ([commit](https://github.com/ava-labs/avalanchego/pull/4051/commits/80ed6fcc09d1e7411360cab686b1a75d6a462632)) results in job not even starting: https://github.com/ava-labs/avalanchego/pull/4051/checks

Taskfile.yml
Elvis339: https://github.com/ava-labs/avalanchego/pull/4051/commits/eaff4eebff3e8d3e2bd70fde047f543222f8e9fe

.github/workflows/load-tests.yaml
Elvis339: https://github.com/ava-labs/avalanchego/pull/4051/commits/3952ee94342e7d93b74cfec329238a1a190eea63

tests/load/README.md
```diff
@@ -114,3 +114,29 @@ for {
...
+./scripts/run_task.sh test-load-kube-kind -- --kube-image=my-registry/avalanchego:v1.2.3
+```
+
+## Trigger CI job on self-hosted runner
+Use GitHub Actions workflow "Load Tests on k8s"
+- Go to Actions tab in GitHub
+- Select "Load Tests on k8s" workflow
+- Click "Run workflow"
+- Specify image (defaults to avaplatform/avalanchego:latest)
+- Additional parameters can be passed via workflow inputs (duration, connections, node count)
```
RodrigoVillar: This should live in `tests/load/c` as that's where we already detail instructions for how to run the C-Chain load tests.

Elvis339: https://github.com/ava-labs/avalanchego/pull/4051/commits/48d2aabc85430715dc1c470710e88ebae53615ea

tests/load/c/README.md
```diff
@@ -20,50 +20,67 @@ From the load test perspective, only the TPS (transactions per second) is logged
...
+
+Initiate nix shell: `nix develop`
+
+## Local tmpnet
+
+To run the load test locally against tmpnet nodes:
 
 ```bash
-# Install nix (skip this step if nix is already installed)
 ./scripts/run_task.sh install-nix
```
RodrigoVillar: (nit): this should be

```
./scripts/run_task.sh test-load
```

Elvis339: https://github.com/ava-labs/avalanchego/pull/4051/commits/028b178c5405ae92d45afdb848df6f66e8b3d312

RodrigoVillar (approved):

.github/actions/run-monitored-tmpnet-cmd/action.yml
maru-ava: Why it isn't possible to configure the self-hosted runner to have the same behavior as a github runner for custom actions in the same repo?

Elvis339: It is possible if we were to maintain our own image.

maru-ava: Why does that require our own image? Is there no way to set the environment for the existing image?

Elvis339: We can modify the workspace [mountPath](https://github.com/actions/actions-runner-controller/blob/469a0faec420d358832f3f5561f9813d530683e3/charts/gha-runner-scale-set/values.yaml#L379) but, that wouldn't solve our problem we'd still have {{ customMountPath }}/{owner}/{repo}/ structure.

The existing ARC image doesn't provide a way to configure the workspace structure that would fix the issue, unless we manage our own image.

maru-ava: Maybe try replacing ${{ github._action_path }} with $GITHUB_ACTION_PATH? It [promises](https://github.com/actions/runner/issues/716#issuecomment-795238933
) to work with self-hosted runner containers and I've [tested it to work](https://github.com/ava-labs/avalanchego/pull/4061) with github runners.

I'd like to avoid requiring that every invocation of the custom action has to provide the action path.

Elvis339: Works 🥳 

Thank you! 

https://github.com/ava-labs/avalanchego/actions/runs/16151409885/job/45583343263?pr=4051

tests/load/README.md
```diff
@@ -113,4 +113,3 @@ for {
         iters += 1
 }
 ```
-
```
maru-ava: Please revert unrelated change

tests/load/c/README.md
```diff
@@ -20,50 +20,68 @@ From the load test perspective, only the TPS (transactions per second) is logged
...
+
+## CI job on self-hosted runner
 
-2. Set your monitoring credentials using the credentials you can find in your password manager
+Use GitHub Actions workflow "Load Tests on k8s":
+- Go to Actions tab in GitHub
+- Select "Load Tests on k8s" workflow
+- Click "Run workflow"
+- Specify image (defaults to avaplatform/avalanchego:latest)
+- Additional parameters can be passed via workflow inputs (duration, connections, node count)
```
maru-ava, line 61: (No action required) Are you intending to add these inputs before merge? If not, maybe remove this line.

tests/load/c/README.md
```diff
@@ -20,50 +20,68 @@ From the load test perspective, only the TPS (transactions per second) is logged
...
+
+## Visualize metrics in Grafana
```
maru-ava, line 63: (No action required) Maybe replace this section with a link to the relevant sections in the tmpnet README?

tests/load/c/README.md
Elvis339: Added duration, node count should be merged soon so this is fine to stay I think.

https://github.com/ava-labs/avalanchego/blob/6eadb2370acf6c07944d0956a4ef7365aec25b92/.github/actions/self-hosted-load-test/action.yml#L13

tests/load/c/README.md
Elvis339: There is a link at the bottom: https://github.com/ava-labs/avalanchego/blob/4b123817483729494c6a1caa5ebc493ecdd5c008/tests/load/c/README.md#visualize-metrics-in-grafana

.github/workflows/ci.yml
```diff
@@ -274,4 +274,22 @@ jobs:
             prometheus_password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
             loki_username: ${{ secrets.LOKI_ID || '' }}
             loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
-
+  load_self_hosted:
```
maru-ava: As per offline conversation, my apologies for suggesting we want to run these jobs on PRs based on labels. Even without exclusive scheduling we'd still be wasting resources on our test cluster with every push. What value would that provide vs having to trigger a job manually as required?

.github/actions/self-hosted-load-test/action.yml
```diff
@@ -0,0 +1,63 @@
...
+  steps:
+    - name: Install dependencies
+      shell: bash
+      # The xz-utils might be present on some containers
+      run: |
+        if ! command -v xz &> /dev/null; then
+          sudo apt-get update
+          sudo apt-get install -y xz-utils
+        fi
+    - name: Run exclusive load test
```
maru-ava: Why would it be desirable to duplicate the step for exclusive and non-exclusive use rather than providing exclusive scheduling as a flag to the test invocation?

I'm not entirely sure why I approved the addition of test-load-exclusive as a task given that the only difference is the flag.

.github/workflows/self-hosted-load-tests.yaml
```diff
@@ -0,0 +1,39 @@
...
+        required: false
+        default: 'avaplatform/avalanchego:latest'
+        type: string
+      exclusive_scheduling:
+        description: 'Enable exclusive scheduling'
+        required: false
+        default: false
+        type: boolean
+      duration:
+        description: "Load test duration"
```
maru-ava: Maybe provide some examples to avoid ambiguity as to the values that could be used (e.g. 10m, 1h, etc)

.github/workflows/ci.yml
Elvis339: The rationale was to run self-hosted load tests against the current branch changes. However, this doesn't work as expected because:
- We don't maintain branch-specific Docker images - only the master branch gets published to Docker Hub
- Tests run against master, not the current branch so we're not actually testing our changes

No real value without proper image building. The tests would need branch-specific images to be meaningful.

.github/actions/self-hosted-load-test/action.yml
Elvis339: Probably to be specific in the absence of documentation.

[Updated.](https://github.com/ava-labs/avalanchego/pull/4051/commits/d47e753b85e65e6c88d935c4d2ba6ae03593b041)

.github/workflows/ci.yml
maru-ava: Even with 'proper image building', I don't think it would be desirable to risk triggering tests on expensive and limited resources on every PR push.

PR #4052, opened 2025-07-03
Add Verification + Tracking Logic To Simplex Blocks

samliok:
## Why this should be merged

In addition to calling `Verify` on the blocks inner vmBlock, we need to do some additional checks to ensure a simplex block can be verified. Furthermore, we need to ensure that any verified blocks are eventually [accepted or rejected](https://github.com/ava-labs/avalanchego/blob/master/snow/consensus/snowman/block.go#L38). 

## How this works

In simplex we do some [pre-verification logic](https://github.com/ava-labs/Simplex/blob/main/epoch.go#L1824). However this is not enough, as we do not have access to the inner `vmBlock` pointed to by the `prev` field in the `ProtocolMetadata`. This means our verification logic in avalanchego must verify that `prev` for the block we are currently verifying points to that blocks parent. To understand more about why this important, see [this test](https://github.com/ava-labs/avalanchego/pull/4052/files#diff-d91108ebba60f252a541424f2c5f299b4a22fb63e4b4461ac607b97e4ce29923R230).

This PR also introduces a `blockTracker` struct to keep track of blocks that have been verified. When `index` is called on a digest, the `blockTracker` calls `Accept` on that block and `Reject` on any competing blocks.


simplex/block.go
```diff
@@ -7,19 +7,25 @@ package simplex
...
+	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/ava-labs/simplex"
 
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
 	"github.com/ava-labs/avalanchego/utils/hashing"
+	"github.com/ava-labs/avalanchego/vms/proposervm/tree"
```
yacovm: Can we move vms/proposervm/tree to somewhere like `snow/tree` ? It's a bit insanitary that Simplex uses something that is in proposervm

simplex/block.go
```diff
@@ -58,9 +66,53 @@ func (b *Block) Bytes() ([]byte, error) {
...
+	if prevBlock != nil {
+		if b.vmBlock.Parent() != prevBlock.vmBlock.ID() {
+			return fmt.Errorf("%w: parentID %s, prevID %s", errMismatchedPrevDigest, b.vmBlock.Parent(), prevBlock.vmBlock.ID())
+		}
+
+		return nil
+	}
+
+	// if we do not have it in the map, it's possible for it to be the last accepted block
+	lastID, err := b.blockTracker.vm.LastAccepted(ctx)
```
yacovm: Instead of doing that can't we load the last accepted to `simplexDigestsToBlock` upon creation? This way we also don't depend on the implementation of the VM's `LastAccepted` in this method, for instance if the VM doesn't implement this efficiently. 

But the bigger benefit is that we should *always* find the previous block this way (unless it's the genesis block) and we have less special handling.

simplex/block.go
```diff
@@ -58,9 +66,53 @@ func (b *Block) Bytes() ([]byte, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to get last accepted block: %w", err)
+	}
+
+	if lastID != b.vmBlock.Parent() {
```
yacovm: I would very much like us to explicitly make sure that the chain induced by Simplex is aligned with the chain induced by the VM, and not only the inner VMs are aligned:

```
prevBlock.BlockHeader().Digest == b.metadata.Prev && prevBlock.vmBlock.ID() == b.vmBlock.Parent()
```

simplex/block.go
```diff
@@ -84,11 +136,66 @@ func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error
...
+}
+
+// indexBlock calls accept on the block with the given digest, and reject on competing blocks.
+func (bt *blockTracker) indexBlock(ctx context.Context, digest simplex.Digest) error {
+	bt.lock.Lock()
+	defer bt.lock.Unlock()
+
+	bd, exists := bt.simplexDigestsToBlock[digest]
+	if !exists {
+		return fmt.Errorf("%w: %s", errDigestNotFound, digest)
```
yacovm, line 192: this case isn't covered by tests. Can we add a test for this?

simplex/block.go
samliok: [added](https://github.com/ava-labs/avalanchego/pull/4052/commits/566a549a5060dbad516f551413fff99c815aaaa1)

simplex/block.go
samliok: moved to utils

simplex/block.go
```diff
@@ -84,11 +135,70 @@ func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error
...
+
+// blockTracker is used to ensure that blocks are properly rejected, if competing blocks are accepted.
+type blockTracker struct {
+	lock sync.Mutex
+
+	// tracks the simplex digests to the blocks that have been verified
+	simplexDigestsToBlock map[simplex.Digest]*Block
+
+	// lastAcceptedID is the ID of the last accepted block at the time of the block tracker creation.
+	lastAcceptedID ids.ID
```
yacovm: shouldn't we update these two values once we we call `indexBlock`?

Can we make a test that is like `TestVerifyInnerBlockBreaksHashChain` but goes through several blocks?

samliok: i dont think we need to update them, since we will have always have the last accepted id/digest in the block trackers memory. I do think the name is misleading though. Added to this [test](https://github.com/ava-labs/avalanchego/pull/4052/commits/46e2bdd379ceee1ad38fef9b9908f905afdfddea)

simplex/block.go
```diff
@@ -58,9 +68,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+		return nil, errGenesisVerification
+	}
+
+	err := b.verifyParentMatchesPrevBlock()
+	if err != nil {
+		return nil, err
+	}
+
+	if b.blockTracker.isBlockAlreadyVerified(b.vmBlock) {
+		return nil, errBlockAlreadyVerified
```
yacovm: I'm wondering why are we returning an error here instead of `nil`?

yacovm (dismissed):

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+
+	err := b.verifyParentMatchesPrevBlock()
+	if err != nil {
+		return nil, err
+	}
+
+	if b.blockTracker.isBlockAlreadyVerified(b.vmBlock) {
+		return b, nil
+	}
+
```
yacovm: I think we need to grab the context lock here, because we're executing this function asynchronously.

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+// verifyParentMatchesPrevBlock verifies that the previous block referenced in the current block's metadata
+// matches the parent of the current block's vmBlock.
+func (b *Block) verifyParentMatchesPrevBlock() error {
+	prevBlock := b.blockTracker.getBlockByDigest(b.metadata.Prev)
+	if prevBlock == nil {
+		return fmt.Errorf("%w: %s", errDigestNotFound, b.metadata.Prev)
+	}
+
+	if b.vmBlock.Parent() != prevBlock.vmBlock.ID() || b.metadata.Prev != prevBlock.digest {
```
StephenButtolph: When can `b.metadata.Prev != prevBlock.digest` ever be true? Removing this check doesn't break any tests.

simplex/block.go
```diff
@@ -84,11 +133,74 @@ func (d *blockDeserializer) DeserializeBlock(bytes []byte) (simplex.Block, error
...
+}
+
+func newBlockTracker(latestBlock *Block) *blockTracker {
+	simplexDigestsToBlock := make(map[simplex.Digest]*Block)
+	simplexDigestsToBlock[latestBlock.digest] = latestBlock
+
+	return &blockTracker{
+		tree:                  tree.New(),
+		simplexDigestsToBlock: simplexDigestsToBlock,
+	}
```
StephenButtolph:
```suggestion
	return &blockTracker{
		tree: tree.New(),
		simplexDigestsToBlock: map[simplex.Digest]*Block{
			latestBlock.digest: latestBlock,
		},
	}
```

simplex/util_test.go
```diff
@@ -4,12 +4,51 @@
...
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman/snowmantest"
 	"github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
 )
 
+// newBlockWithDigest is a helper function that creates a new block and sets its digest.
+// This is helpful since otherwise we would need the blockDeserializer to create the block.
+func newBlockWithDigest(t *testing.T, vmBlock snowman.Block, tracker *blockTracker, round, seq uint64, prev simplex.Digest) *Block {
```
StephenButtolph: I have found in other tests that having various `newBlockWithXXX(`, `newBlock(` ect scales really annoyingly.

If we don't foresee adding more things here, then feel free to ignore... But it might make sense to make a single `newBlock` function which takes in a config with reasonable zero values.

That has proven to be much more flexible in my experience.

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+
+	err := b.verifyParentMatchesPrevBlock()
+	if err != nil {
+		return nil, err
+	}
+
+	if b.blockTracker.isBlockAlreadyVerified(b.vmBlock) {
+		return b, nil
+	}
```
StephenButtolph: Don't we still need to add `b` to the `simplexDigestsToBlock` map here?

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
 func (b *Block) Verify(ctx context.Context) (simplex.VerifiedBlock, error) {
-	// TODO: track blocks that have been verified to ensure they are either rejected or accepted
-	return b, b.vmBlock.Verify(ctx)
+	// we should not verify the genesis block
+	if b.metadata.Seq == 0 {
+		return nil, errGenesisVerification
+	}
+
+	err := b.verifyParentMatchesPrevBlock()
+	if err != nil {
```
StephenButtolph: For simple checks like these (where `err` shouldn't be referenced later) I think we typically try to use inline `if` statements. The only exception to this is with very long (multi-line) statements, which are hard to read imo.
```suggestion
	if err := b.verifyParentMatchesPrevBlock(); err != nil {
```

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+		return nil, err
+	}
+
+	if b.blockTracker.isBlockAlreadyVerified(b.vmBlock) {
+		return b, nil
+	}
+
+	err = b.vmBlock.Verify(ctx)
+	if err != nil {
```
StephenButtolph: See above comment
```suggestion
	if err := b.vmBlock.Verify(ctx); err != nil {
```

simplex/block.go
```diff
@@ -58,9 +67,49 @@ func (b *Block) Bytes() ([]byte, error) {
...
+	// once verified, the block tracker will eventually either accept or reject the block
+	b.blockTracker.trackBlock(b)
+	return b, nil
+}
+
+// verifyParentMatchesPrevBlock verifies that the previous block referenced in the current block's metadata
+// matches the parent of the current block's vmBlock.
+func (b *Block) verifyParentMatchesPrevBlock() error {
+	prevBlock := b.blockTracker.getBlockByDigest(b.metadata.Prev)
+	if prevBlock == nil {
```
StephenButtolph: I know that maps return the zero value when looking up something fails, and that we never expect to actually place a `nil` value in the map. But in avalanchego we typically do explicit success checks (so `prevBlock, ok :=` ...)

simplex/block.go
samliok: i think if it's in the tree it should already be in the digests map, but added anyways

simplex/block.go
```diff
@@ -84,11 +125,80 @@ func (d *blockDeserializer) DeserializeBlock(ctx context.Context, bytes []byte)
...
 		metadata: *md,
 		vmBlock:  vmblock,
-		digest:   digest,
+		digest:   computeDigest(bytes),
 	}, nil
 }
+
+// blockTracker is used to ensure that blocks are properly rejected, if competing blocks are accepted.
+type blockTracker struct {
+	lock sync.Mutex
```
StephenButtolph, line 137: I think yacov's comment around needing to hold the context lock here was more around how every `VM` function expects the `*snow.Context.Lock` to be held when it is called. Not that we needed to add a new lock.

simplex/block.go
StephenButtolph: It isn't guaranteed to be in the digests map (because the simplex `digest` is not 1<->1 with the inner VM's `ID()`).

simplex/block.go
samliok: as talked about offline, we are going to grab this lock where simplex implements the VM interface

yacovm: So, in Simplex we have two main ways in which a VM can be invoked:

1. You receive a message and respond to it. This is mostly fetching blocks from the VM or indexing them. 
2. Asynchronously via the scheduler which verifies blocks and builds new blocks.

All (1) are synchronized using a lock, but (2) are not synchronized, as we don't want to block Simplex from responding to messages or doing network I/O or writing to the WAL while building or verifying a block. 

However, we should take care of that in the avalanchego side. 

I think the simplest solution would be to make a VM wrapper that synchronizes all VM methods.

yacovm (approved):

PR #4057, opened 2025-07-08
[tmpnet] Source tmpnet defaults from a configmap

maru-ava:
## Why this should be merged

Previously the scheduling label and value required to enable exclusive scheduling were defined as flag defaults. To enable the cluster to define these defaults, the defaults are now sourced from a configmap in the target namespace.

## How this was tested

CI, manually

N/A

### DIFF
--- a/tests/fixture/tmpnet/kube_runtime.go
+++ b/tests/fixture/tmpnet/kube_runtime.go
@@ -52,7 +52,7 @@ const (
 	antiAffinityLabelValue = "exclusive"
 
 	// Name of config map containing tmpnet defaults
-	kubeRuntimeConfigMapName = "tmpnet"
+	defaultsConfigMapName = "tmpnet-defaults"
 )
 
 var errMissingSchedulingLabels = errors.New("--kube-scheduling-label-key and --kube-scheduling-label-value are required when exclusive scheduling is enabled")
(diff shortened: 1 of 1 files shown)


tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -72,6 +78,53 @@ type KubeRuntimeConfig struct {
...
+	)
+
+	configMap, err := clientset.CoreV1().ConfigMaps(c.Namespace).Get(ctx, kubeRuntimeConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get ConfigMap: %w", err)
+	}
+
+	var (
+		schedulingLabelKey   = configMap.Data["defaultSchedulingLabelKey"]
+		schedulingLabelValue = configMap.Data["defaultSchedulingLabelValue"]
```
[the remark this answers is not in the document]
maru-ava: No

tests/fixture/tmpnet/kube_runtime.go
```diff
@@ -72,6 +78,53 @@ type KubeRuntimeConfig struct {
...
+			zap.String("schedulingLabelKey", schedulingLabelKey),
+		)
+		c.SchedulingLabelKey = schedulingLabelKey
+	}
+	if len(c.SchedulingLabelValue) == 0 && len(schedulingLabelValue) > 0 {
+		log.Info("setting default value for SchedulingLabelValue",
+			zap.String("schedulingLabelValue", schedulingLabelValue),
+		)
+		c.SchedulingLabelValue = schedulingLabelValue
+	}
```
RodrigoVillar, line 118: Is there a benefit to mix-and-matching between flag values and ConfigMap values here? I would think that the `KubeRuntimeConfig` would go with exclusively with flag values or exclusively with ConfigMap values.

maru-ava: Configmap values are _defaults_ and only used when a flag has an empty value (as per the logging communicating their usage). The intention is to ensure that a cluster can provide appropriate defaults rather than to prevent a user from supplying values via flags.

JuanLeon2: Rodrigo, can you Resolve this if you agree/understand, or follow up?

maru-ava:
Rebased

PR #4059, opened 2025-07-08
chore: replace polling with subscription-based acceptance

RodrigoVillar:
## Why this should be merged

As mentioned in https://github.com/ava-labs/avalanchego/pull/3996#discussion_r2172992203, we should not be polling API nodes for a transaction's status in the context of load testing. This PR modifies the `awaitTx()` to use block-subscriptions instead of excessive polling for determining a transaction's status.

## How this works

For each block `awaitTx()` hears about, `awaitTx()` queries the nonce of the sender at the given height to determine if the sender's nonce has incremented by one (i.e. if the sender's outstanding transaction was included in a block). If the nonce has incremented, `awaitTx()` then makes a single call to `client.TransactionReceipt()` to determine if the transaction successfully executed or not.

## How this was tested

CI

N/A

### DIFF
--- a/tests/load2/wallet.go
+++ b/tests/load2/wallet.go
@@ -94,6 +94,7 @@ func (w Wallet) awaitTx(
 	errs <-chan error,
 	txHash common.Hash,
 ) error {
+	account := crypto.PubkeyToAddress(w.privKey.PublicKey)
 	for {
 		select {
 		case <-ctx.Done():
@@ -103,7 +104,7 @@ func (w Wallet) awaitTx(
 		case h := <-headers:
 			latestNonce, err := w.client.NonceAt(
 				ctx,
-				crypto.PubkeyToAddress(w.privKey.PublicKey),
+				account,
 				h.Number,
 			)
 			if err != nil {


tests/load2/wallet.go
```diff
@@ -70,32 +87,40 @@ func (w *Wallet) SendTx(
 	return nil
 }
 
-func awaitTx(
+func (w Wallet) awaitTx(
```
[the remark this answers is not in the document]
RodrigoVillar: `awaitTx()` doesn't modify any of the values of `w` and shouldn't need a pointer

tests/load2/wallet.go
```diff
@@ -47,8 +45,22 @@ func newWallet(
...
+	sub, err := w.client.SubscribeNewHead(ctx, headers)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		sub.Unsubscribe()
+
+		// wait for err chan to close before safely closing headers
+		<-sub.Err()
```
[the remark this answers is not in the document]
RodrigoVillar: Adding the `ctx` branch could lead to the subscription sending to a closed channel if the context is closed.

tests/load2/wallet.go
```diff
@@ -57,7 +69,12 @@ func (w *Wallet) SendTx(
 	issuanceDuration := time.Since(startTime)
 	w.metrics.issue(issuanceDuration)
 
-	if err := awaitTx(ctx, w.client, tx.Hash(), pingFrequency); err != nil {
+	if err := w.awaitTx(
```
maru-ava: Please limit use of `if with short statement` to _single line_ statements to avoid compromising legibility. As soon as a statement spans multiple lines, fall back to separating the error assignment from the check e.g.

```golang
err := myfunc(
  arg1,
  arg2,
  ...
)
if err != nil { ... }
```

tests/load2/wallet.go
```diff
@@ -70,32 +87,40 @@ func (w *Wallet) SendTx(
...
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-ticker.C:
+		case err := <-errs:
+			return err
+		case h := <-headers:
+			latestNonce, err := w.client.NonceAt(
+				ctx,
+				crypto.PubkeyToAddress(w.privKey.PublicKey),
```
maru-ava: (No action required) Maybe compute this outside the loop?

tests/load2/wallet.go
```diff
@@ -70,32 +87,40 @@ func (w *Wallet) SendTx(
...
+				h.Number,
+			)
+			if err != nil {
+				return err
+			}
+
+			if latestNonce == w.nonce+1 {
```
maru-ava, line 114: (No action required) Would there ever be a case where the nonce that the wallet sees is greater than one more than the wallet nonce?

tests/load2/wallet.go
```diff
@@ -47,8 +45,22 @@ func newWallet(
 func (w *Wallet) SendTx(
 	ctx context.Context,
 	tx *types.Transaction,
-	pingFrequency time.Duration,
 ) error {
+	// start listening for blocks
+	headers := make(chan *types.Header)
+	sub, err := w.client.SubscribeNewHead(ctx, headers)
```
maru-ava, line 51: (No action required) Why is it desirable for each wallet to subscribe separately?

tests/load2/wallet.go
RodrigoVillar: Yes - if the client sends a tx without using the wallet and then sends another tx _with_ the wallet, then we'll have that `latestNonce > w.nonce+1`

tests/load2/wallet.go
RodrigoVillar: Done: e5c6871

tests/load2/wallet.go
RodrigoVillar: Done: bc4de3c

tests/load2/wallet.go
RodrigoVillar: Each wallet consumes the header from `headers` to get the account's latest nonce. Since a value can only be read from a channel once, having multiple wallets use a single subscription would lead to multiple wallets missing out on headers.

tests/load2/wallet.go
maru-ava: Ah, ok. So no chance of it being more than one (i.e. suggesting `>=` would be more appropriate) in the context of a load test.

tests/load2/wallet.go
maru-ava: Sounds doable but more complicated to minimize the number of subscriptions required. Something we can revisit if the number of subscriptions required becomes an issue.

Elvis339 (approved):

PR #4060, opened 2025-07-08
Ensure that headers added by the server are propagated

joshua-kim:
## Why this should be merged

Headers added by middleware can be currently dropped by rpcchainvm for calls handled by `HandleHTTPSimple`

## How this works

HandleHTTPSimple does not propagate response headers when sending the request to the plugin, so when avalanchego reconstructs the response back the client it loses the response headers because the plugin was never given them.

## How this was tested

Added a regression test

Confirmed regression:
```
➜  avalanchego git:(master) curl -s -v --location --request POST 'http://localhost:9650/ext/bc/Q3fReCSX3QSXxGXkcFC4wyU4ZmPsc22HAGjoge8oxWCzQ8bQD' \
-H 'Origin: https://develop.avacloud-app.pages.dev' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "xsvm.network",
    "params":{},
    "id": 1
}'

...

>
* upload completely sent off: 84 bytes
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Tue, 08 Jul 2025 19:25:09 GMT
< Content-Length: 180
<

...
```

Manually tested that the headers are received:
```
➜  avalanchego git:(header-fix) curl -s -v --location --request POST 'http://localhost:9650/ext/bc/Q3fReCSX3QSXxGXkcFC4wyU4ZmPsc22HAGjoge8oxWCzQ8bQD' \
-H 'Origin: https://develop.avacloud-app.pages.dev' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "xsvm.network",
    "params":{},
    "id": 1
}'

...

< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: *
< Content-Type: application/json; charset=utf-8
< Node-Id: NodeID-7Xhw2mDxuDS44j42TCB6U5579esbSt3Lg
< Vary: Origin
< X-Content-Type-Options: nosniff
< Date: Tue, 08 Jul 2025 19:23:28 GMT
< Content-Length: 180
<
{ [180 bytes data]
* Connection #0 to host localhost left intact

...
```

Yes (rpcchainvm bump)


proto/README.md
```diff
@@ -1,6 +1,6 @@
 # Avalanche gRPC
 
-Now Serving: **Protocol Version 42**
+Now Serving: **Protocol Version 43**
```
StephenButtolph: We haven't done a release with protocol version 42 yet. I don't think this PR should update the version again.

proto/http/http.proto
```diff
@@ -155,11 +155,13 @@ message HandleSimpleHTTPRequest {
   string method = 1;
   // url specifies either the URI being requested
   string url = 2;
-  // headers contains the request header fields either received
+  // request_headers contains the request header fields either received
   // by the server or to be sent by the client
```
StephenButtolph: These are just the headers received by the server right?
```suggestion
  // request_headers contains the request header fields received by the server
```

joshua-kim: Not my comment 😿 but we should still improve it. Yeah these are the headers we receive, i.e the request headers.

StephenButtolph (approved):

PR #4064, opened 2025-07-09
refactor(x/sync): remove need for passing proof nodes to completeWorkItem

alarso16:
## Why this should be merged

As a part of the refactor to allow Firewood to use x/sync, we should remove any proof logic from sync manager. This moves finding the next key to request to the proof commit implementation, as well as parsing the proof from bytes.

## How this works

The `completeWorkItem` function does not need to know the largest handled key, but rather the next key that needs retrieved. This is handled through a hook, since even on some non-error returns, it shouldn't be called (see implementation). Instead of passing in the `DB` to the Sync Manager, it will take this `ProofClient`

## How this was tested

Unit tests.

No.

### DIFF
--- a/x/sync/db.go
+++ b/x/sync/db.go
@@ -6,23 +6,14 @@ package sync
 import (
 	"context"
 
+	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/maybe"
-	"github.com/ava-labs/avalanchego/x/merkledb"
 
 	pb "github.com/ava-labs/avalanchego/proto/pb/sync"
 )
 
-type DB interface {
-	merkledb.Clearer
-	merkledb.MerkleRootGetter
-	merkledb.ProofGetter
-	merkledb.ChangeProofer
-	merkledb.RangeProofer
-}
-
-type ProofClient interface {
-	merkledb.Clearer
-	merkledb.MerkleRootGetter
+type DBSyncClient interface {
+	GetRootHash(ctx context.Context) (ids.ID, error)
 	HandleRangeProofResponse(ctx context.Context, request *pb.SyncGetRangeProofRequest, responseBytes []byte) (maybe.Maybe[[]byte], error)
 	HandleChangeProofResponse(
 		ctx context.Context,
@@ -30,4 +21,5 @@ type ProofClient interface {
 		responseBytes []byte,
 	) (maybe.Maybe[[]byte], error)
 	Error() error
+	Clear() error
 }
--- a/x/sync/db.go
+++ b/x/sync/db.go
@@ -3,7 +3,14 @@
 
 package sync
 
-import "github.com/ava-labs/avalanchego/x/merkledb"
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/utils/maybe"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+
+	pb "github.com/ava-labs/avalanchego/proto/pb/sync"
+)
 
 type DB interface {
 	merkledb.Clearer
@@ -11,4 +18,17 @@ type DB interface {
 	merkledb.ProofGetter
 	merkledb.ChangeProofer
 	merkledb.RangeProofer
+}
+
+type ProofClient interface {
+	merkledb.Clearer
+	merkledb.MerkleRootGetter
+	HandleRangeProofResponse(ctx context.Context, request *pb.SyncGetRangeProofRequest, responseBytes []byte, onFinish func(maybe.Maybe[[]byte])) error
+	HandleChangeProofResponse(
+		ctx context.Context,
+		request *pb.SyncGetChangeProofRequest,
+		responseBytes []byte,
+		onFinish func(maybe.Maybe[[]byte]),
+	) error
+	RegisterErrorHandler(handler func(error))
 }
--- a/x/sync/db.go
+++ b/x/sync/db.go
@@ -23,12 +23,11 @@ type DB interface {
 type ProofClient interface {
 	merkledb.Clearer
 	merkledb.MerkleRootGetter
-	HandleRangeProofResponse(ctx context.Context, request *pb.SyncGetRangeProofRequest, responseBytes []byte, onFinish func(maybe.Maybe[[]byte])) error
+	HandleRangeProofResponse(ctx context.Context, request *pb.SyncGetRangeProofRequest, responseBytes []byte) (maybe.Maybe[[]byte], error)
 	HandleChangeProofResponse(
 		ctx context.Context,
 		request *pb.SyncGetChangeProofRequest,
 		responseBytes []byte,
-		onFinish func(maybe.Maybe[[]byte]),
-	) error
-	RegisterErrorHandler(handler func(error))
+	) (maybe.Maybe[[]byte], error)
+	Error() error
 }
--- a/x/sync/manager.go
+++ b/x/sync/manager.go
@@ -9,14 +9,12 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
-	"golang.org/x/exp/maps"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/ava-labs/avalanchego/ids"
@@ -24,7 +22,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/maybe"
 	"github.com/ava-labs/avalanchego/utils/set"
-	"github.com/ava-labs/avalanchego/x/merkledb"
 
 	pb "github.com/ava-labs/avalanchego/proto/pb/sync"
 )
@@ -134,7 +131,6 @@ type Manager struct {
 	// Set to true when StartSyncing is called.
 	syncing   bool
 	closeOnce sync.Once
-	tokenSize int
 
 	stateSyncNodeIdx uint32
 	metrics          SyncMetrics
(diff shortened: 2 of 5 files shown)


x/sync/manager.go
```diff
@@ -642,14 +656,25 @@ func (m *Manager) handleChangeProofResponse(
...
-		m.completeWorkItem(ctx, work, largestHandledKey, targetRootID, rangeProof.EndProof)
 	default:
 		return fmt.Errorf(
 			"%w: %T",
 			errUnexpectedChangeProofResponse, changeProofResp,
 		)
 	}
+	// Find the next key to fetch.
+	// If this is empty, then we have no more keys to fetch.
+	if !largestHandledKey.IsNothing() {
```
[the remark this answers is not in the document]
alarso16: I will be pushing this down a layer in a later PR

x/sync/sync_test.go
```diff
@@ -422,6 +422,55 @@ func Test_Sync_FindNextKey_ExtraValues(t *testing.T) {
...
+		ChangeProofClient:     p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, NewGetChangeProofHandler(db)),
+		TargetRoot:            targetRoot,
+		SimultaneousWorkLimit: 5,
+		Log:                   logging.NoLog{},
+		BranchFactor:          merkledb.BranchFactor16,
+	}, prometheus.NewRegistry())
+	require.NoError(err)
+
+	// Since both keys are identical, the next key should be nothing, since the range is complete
+	nextKey, err := syncer.findNextKey(context.Background(), testKey, maybe.Some([]byte{0x11, 0x11}), proof.EndProof)
```
joshua-kim: Why are we calling unexported code to test this? Is it not possible for us to test this w/ the exported api? Testing package implementation details tends to be brittle + not representative of how callers use this package

alarso16: This is just ensuring an edge case in the `findNextKey` function is handled correctly. As you point out, if this function is going to be tested thoroughly, all the tests should be redesigned to ensure that these edge cases are hit and handled correctly internally

x/merkledb/db.go
```diff
@@ -1390,3 +1429,217 @@ func cacheEntrySize(key Key, n *node) int {
...
+	// or have the same ID in the other proof.
+	proofKeyPath := ToKey(lastReceivedKey)
+
+	// If the received proof is an exclusion proof, the last node may be for a
+	// key that is after the [lastReceivedKey].
+	// If the last received node's key is after the [lastReceivedKey], it can
+	// be removed to obtain a valid proof for a prefix of the [lastReceivedKey].
+	if !proofKeyPath.HasPrefix(endProof[len(endProof)-1].Key) {
+		endProof = endProof[:len(endProof)-1]
+		if len(endProof) == 0 {
```
alarso16, line 1473: This was a missed edge case prior to these tests

x/merkledb/generator.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
alarso16, line 1: Moved from `x/sync/sync_test.go`

x/sync/merkledb_client.go
```diff
@@ -0,0 +1,479 @@
...
+//
+// [rangeEnd] is the end of the range that we want to fetch.
+//
+// Returns Nothing if there are no more keys to fetch in [lastReceivedKey, rangeEnd].
+//
+// [endProof] is the end proof of the last proof received.
+//
+// Invariant: [lastReceivedKey] < [rangeEnd].
+// If [rangeEnd] is Nothing it's considered > [lastReceivedKey].
+func (c *xSyncClient) findNextKey(
```
alarso16: Everything this function and below was completely unmodified.

x/sync/proof_test.go
```diff
@@ -0,0 +1,657 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p/p2ptest"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/maybe"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Sync_FindNextKey_InSync(t *testing.T) {
```
alarso16, line 26: These tests are unmodified besides initialization

x/sync/db.go
```diff
@@ -12,3 +19,18 @@ type DB interface {
 	merkledb.ChangeProofer
 	merkledb.RangeProofer
 }
+
+type ProofClient interface {
```
alarso16: Any ideas for a better name?

rkuris: SyncClient?

x/sync/db.go
```diff
@@ -12,3 +19,16 @@ type DB interface {
...
+	merkledb.Clearer
+	merkledb.MerkleRootGetter
```
rkuris: nit: naming: These are no longer merkledb interfaces, correct? Sure, they are implemented by merkledb but IIRC golang just lets you declare interfaces. If so, interfaces used by ProofClient (or SyncClient) should be maybe in a package named xsync or something.

x/sync/manager.go
```diff
@@ -134,24 +131,20 @@ type Manager struct {
...
-	tokenSize int
 
 	stateSyncNodeIdx uint32
 	metrics          SyncMetrics
 }
 
 // TODO remove non-config values out of this struct
 type ManagerConfig struct {
-	DB                    DB
+	ProofClient           ProofClient
```
rkuris: nit: Does it make sense to just use `ProofClient` here so you aren't stuttering? Not sure if that makes sense.

x/sync/manager.go
```diff
@@ -134,24 +131,20 @@ type Manager struct {
...
 	RangeProofClient      *p2p.Client
 	ChangeProofClient     *p2p.Client
 	SimultaneousWorkLimit int
 	Log                   logging.Logger
 	TargetRoot            ids.ID
-	BranchFactor          merkledb.BranchFactor
```
rkuris, line 151: What happened to this?

x/sync/manager.go
```diff
@@ -327,7 +315,7 @@ func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
 
 	if work.localRootID == targetRootID {
 		// Start root is the same as the end root, so we're done.
-		m.completeWorkItem(ctx, work, work.end, targetRootID, nil)
+		m.completeWorkItem(work, work.end, targetRootID)
```
rkuris, line 319: 🏆 seems like not having context means we aren't doing any database work down there, which I think is good!

x/sync/proof_test.go
rkuris: These comments significantly reduce review time, ty for doing that!

x/sync/db.go
joshua-kim: What do we define this interface for?

x/sync/merkledb_client.go
```diff
@@ -0,0 +1,478 @@
...
+)
+
+var _ ProofClient = (*xSyncClient)(nil)
+
+type ClientConfig struct {
+	Hasher       merkledb.Hasher
+	BranchFactor merkledb.BranchFactor
+}
+
+type xSyncClient struct {
```
joshua-kim: This abstraction feels very difficult to reason about and there doesn't seem to be a clear abstraction boundary defined - what is this type intended to represent and what is it's responsibility? It's not really a client - it doesn't make any requests, but seems to define handlers for how to interpret responses + you can interact with the underlying database it's coupled to.

x/sync/db.go
alarso16: The x/sync manager should be able to work with other database implementations besides `MerkleDB` (specifically, [Firewood](https://github.com/ava-labs/firewood)). This is the smallest reasonable interface to allow other databases to hook in.

x/sync/merkledb_client.go
alarso16: It isn't really a client, but I didn't know what to call it. The goal was to remove the dependence on database and proof implementation (specifically `merkledb`) from the sync manager. The x/sync manager in particular shouldn't depend on any specific database/proof implementation

x/sync/manager.go
alarso16: Moved to `merkleDBSyncClient`. It was only used for db ops

x/sync/manager.go
alarso16: This actually does something different - it embeds the struct.
Here I made a little [demo](https://go.dev/play/p/2zC1MBa-0rg)

x/sync/db.go
alarso16: How about `DBSyncClient`?

x/sync/db.go
```diff
@@ -3,12 +3,23 @@
...
+	GetRootHash(ctx context.Context) (ids.ID, error)
+	HandleRangeProofResponse(ctx context.Context, request *pb.SyncGetRangeProofRequest, responseBytes []byte) (maybe.Maybe[[]byte], error)
+	HandleChangeProofResponse(
+		ctx context.Context,
+		request *pb.SyncGetChangeProofRequest,
+		responseBytes []byte,
+	) (maybe.Maybe[[]byte], error)
+	Error() error
+	Clear() error
 }
```
joshua-kim, line 25: Thoughts on this interface are:
1. The abstraction boundary isn't clear, we couple the database w/ how responses are handled
2. `Error` is strange IMO, we're expecting our sync package to just keep checking this return value. We could probably get rid of this from the interface if we moved this information somewhere else... we would only fatal during response handling so does it seem reasonable to see if it's possible to expose that information to that API?

x/sync/proof_test.go
joshua-kim: I would avoid unnecessary moves if possible in this PR, it makes it difficult to review - it's okay (and common) in Go to just stuff things in one file.

alarso16:
Closed in favor of smaller PRs

PR #4065, opened 2025-07-09
Add default validator set functions to snowtest.context

JonathanOppenheimer:
## Why this should be merged
The old `ValidatorState` in `NewContext` was set by default to be `nil.` This behavior was not replicated in `snowtest.Context` and these empty default needs to be available to many tests in subnet-evm.  

## How this works
Returns `nil` for the two validator functions. 

No


StephenButtolph (approved):
lgtm

PR #4070, opened 2025-07-10
Make run-monitored-tmpnet-cmd configurable

aaronbuchwald:
This PR updates the run-monitored-tmpnet-cmd GitHub Action to allow a caller to specify the dashboard name to display in CI. It also removes the usage of `bash -x` as a prefix to the command that gets passed in, so that it can be used to execute a go command or binary rather than being restricted to specify a bash script.

Broken out from https://github.com/ava-labs/avalanchego/pull/4019 which displays a different dashboard than the main dashboard for a bootstrap test and runs a go test rather than a bash script.


aaronbuchwald:
TODO: need to handle `is_ephemeral_node=false` in the display link as well https://github.com/ava-labs/avalanchego/pull/4019#issuecomment-3058901400

.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -68,7 +70,7 @@ runs:
     - name: Run command
       shell: bash
       # --impure ensures the env vars are accessible to the command
-      run: ${{ inputs.run_env }} ${{ github.action_path }}/nix-develop.sh --impure --command bash -x ${{ inputs.run }}
```
maru-ava, line 71: What suggested this change?

aaronbuchwald: I made this change so that I can run a golang benchmark directly without using a bash script as an intermediary and broke it out from https://github.com/ava-labs/avalanchego/pull/4019. I've switched to using a task since then, but still thought it was a good change.

maru-ava: If you're not using it, YAGNI.

aaronbuchwald: Closing this PR in favor of https://github.com/ava-labs/avalanchego/pull/4089

PR #4071, opened 2025-07-10
feat(load2): add contract tests

RodrigoVillar:

### DIFF
--- a/tests/load2/main/main.go
+++ b/tests/load2/main/main.go
@@ -5,9 +5,11 @@ package main
 
 import (
 	"flag"
+	"math/big"
 	"os"
 	"time"
 
+	"github.com/ava-labs/libevm/accounts/abi/bind"
 	"github.com/ava-labs/libevm/ethclient"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
@@ -16,6 +18,7 @@ import (
 	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
 	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
 	"github.com/ava-labs/avalanchego/tests/load"
+	"github.com/ava-labs/avalanchego/tests/load/c/contracts"
 	"github.com/ava-labs/avalanchego/tests/load2"
 )
 
@@ -110,14 +113,103 @@ func main() {
 	chainID, err := workers[0].Client.ChainID(ctx)
 	require.NoError(err)
 
+	txOpts, err := bind.NewKeyedTransactorWithChainID(workers[0].PrivKey, chainID)
+	require.NoError(err)
+
+	_, tx, contract, err := contracts.DeployEVMLoadSimulator(txOpts, workers[0].Client)
+	require.NoError(err)
+
+	_, err = bind.WaitDeployed(ctx, workers[0].Client, tx)
+	require.NoError(err)
+	workers[0].Nonce++
+
+	randomTest, err := createRandomTest(contract)
+	require.NoError(err)
+
 	generator, err := load2.NewLoadGenerator(
 		workers,
 		chainID,
 		metricsNamespace,
 		registry,
-		load2.ZeroTransferTest{PollFrequency: pollFrequency},
+		randomTest,
 	)
 	require.NoError(err)
 
 	generator.Run(tc, ctx, loadTimeout, testTimeout)
+}
+
+func createRandomTest(contract *contracts.EVMLoadSimulator) (load2.RandomTest, error) {
+	count := big.NewInt(5)
+	weightedTests := []load2.WeightedTest{
+		{
+			Test:   load2.ZeroTransferTest{},
+			Weight: 100,
+		},
+		{
+			Test: load2.ReadTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.WriteTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.StateModificationTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.HashingTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.MemoryTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.CallDepthTest{
+				Contract: contract,
+				Count:    count,
+			},
+			Weight: 100,
+		},
+		{
+			Test:   load2.ContractCreationTest{Contract: contract},
+			Weight: 100,
+		},
+		{
+			Test: load2.PureComputeTest{
+				Contract:      contract,
+				NumIterations: count,
+			},
+			Weight: 100,
+		},
+		{
+			Test: load2.LargeEventTest{
+				Contract:  contract,
+				NumEvents: count,
+			},
+			Weight: 100,
+		},
+		{
+			Test:   load2.ExternalCallTest{Contract: contract},
+			Weight: 100,
+		},
+	}
+
+	return load2.NewRandomTest(weightedTests)
 }
--- a/tests/load2/tests.go
+++ b/tests/load2/tests.go
@@ -5,9 +5,13 @@ package load2
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"fmt"
 	"math/big"
+	"math/rand/v2"
 	"time"
 
+	"github.com/ava-labs/libevm/accounts/abi/bind"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/crypto"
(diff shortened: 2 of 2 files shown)


tests/load2/tests.go
```diff
@@ -48,5 +114,231 @@ func (z ZeroTransferTest) Run(
 	})
 	require.NoError(err)
 
-	require.NoError(wallet.SendTx(ctx, tx, z.PollFrequency))
+	require.NoError(wallet.SendTx(ctx, tx, time.Millisecond))
```
RodrigoVillar: Waiting on #4059 to be merged in to remove `PollFrequency` parameter.

tests/load2/main/main.go
```diff
@@ -110,14 +113,103 @@ func main() {
...
-		load2.ZeroTransferTest{PollFrequency: pollFrequency},
+		randomTest,
 	)
 	require.NoError(err)
 
 	generator.Run(tc, ctx, loadTimeout, testTimeout)
 }
+
+func createRandomTest(contract *contracts.EVMLoadSimulator) (load2.RandomTest, error) {
+	count := big.NewInt(5)
```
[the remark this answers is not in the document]
RodrigoVillar: `count` is not a magic number in this context.

tests/load2/tests.go
```diff
@@ -5,25 +5,91 @@ package load2
...
+}
+
+func (r RandomTest) Run(
+	tc tests.TestContext,
+	ctx context.Context,
+	wallet *Wallet,
+) {
+	require := require.New(tc)
+
+	index, ok := r.weighted.Sample(rand.Uint64N(r.totalWeight)) //#nosec G404
```
Elvis339: How does this reflect load testing reproducibility, should we consider using a seeded random source?

tests/load2/tests.go
```diff
@@ -5,25 +5,91 @@ package load2
...
 	"github.com/stretchr/testify/require"
 
 	"github.com/ava-labs/avalanchego/tests"
+	"github.com/ava-labs/avalanchego/tests/load/c/contracts"
+	"github.com/ava-labs/avalanchego/utils/sampler"
 )
 
-var _ Test = (*ZeroTransferTest)(nil)
+var (
+	maxFeeCap = big.NewInt(300000000000)
```
Elvis339: Should we have a config structure for example
```
type LoadTestConfig struct {
    MaxFeeCap     *big.Int
    DefaultTimeout time.Duration
    GasTipMultiplier int64
}
```

tests/load2/tests.go
```diff
@@ -48,5 +114,231 @@ func (z ZeroTransferTest) Run(
...
+	require.NoError(err)
+
+	tx, err := r.Contract.SimulateReads(txOpts, r.Count)
+	require.NoError(err)
+
+	require.NoError(wallet.SendTx(ctx, tx, time.Millisecond))
+}
+
+type WriteTest struct {
+	Contract *contracts.EVMLoadSimulator
```
Elvis339, line 258: Noticed we're repeating the same setup in pretty much every test:

```go
Contract *contracts.EVMLoadSimulator  // in every struct

// and this exact block everywhere:
txOpts, err := newTxOpts(wallet.privKey, wallet.chainID, maxFeeCap, wallet.nonce)
require.NoError(err)
```

Could probably clean this up with a base struct or something. Here's one way that might work:

```go
type BaseContractTest struct {
    Contract *contracts.EVMLoadSimulator
}

func (b *BaseContractTest) executeContractTx(
    tc tests.TestContext,
    ctx context.Context, 
    wallet *Wallet,
    txFunc func(*bind.TransactOpts) (*types.Transaction, error),
) {
    require := require.New(tc)
    
    txOpts, err := newTxOpts(wallet.privKey, wallet.chainID, maxFeeCap, wallet.nonce)
    require.NoError(err)
    
    tx, err := txFunc(txOpts)
    require.NoError(err)
    
    require.NoError(wallet.SendTx(ctx, tx, time.Millisecond))
}

// Then tests become much simpler:
type ReadTest struct {
    BaseContractTest
    Count *big.Int
}

func (r ReadTest) Run(tc tests.TestContext, ctx context.Context, wallet *Wallet) {
    r.executeContractTx(tc, ctx, wallet, func(txOpts *bind.TransactOpts) (*types.Transaction, error) {
        return r.Contract.SimulateReads(txOpts, r.Count)
    })
}
```

Would cut down the boilerplate quite a bit. Though there's probably other ways to approach this too this is just what came to mind first. What do you think?

tests/load2/tests.go
```diff
@@ -5,18 +5,86 @@ package load2
...
 	"github.com/ava-labs/avalanchego/tests"
+	"github.com/ava-labs/avalanchego/tests/load/c/contracts"
+	"github.com/ava-labs/avalanchego/utils/sampler"
 )
 
-var _ Test = (*ZeroTransferTest)(nil)
+var (
+	maxFeeCap = big.NewInt(300000000000)
+
+	_ Test = (*RandomTest)(nil)
```
maru-ava: Maybe remove these interface compliance checks in favor of adding test collection to a constructor in this file, in which case the compiler will be checking compliance already?

tests/load2/main/main.go
```diff
@@ -107,14 +110,106 @@ func main() {
...
 	)
 	require.NoError(err)
 
 	generator.Run(tc, ctx, loadTimeout, testTimeout)
 }
+
+func createRandomTest(contract *contracts.EVMLoadSimulator) (load2.RandomTest, error) {
+	// test params
+	count := big.NewInt(5)
+	weight := uint64(100)
```
maru-ava: Why is weight supplied if doesn't vary?

tests/load2/main/main.go
```diff
@@ -107,14 +110,106 @@ func main() {
...
+		randomTest,
 	)
 	require.NoError(err)
 
 	generator.Run(tc, ctx, loadTimeout, testTimeout)
 }
+
+func createRandomTest(contract *contracts.EVMLoadSimulator) (load2.RandomTest, error) {
+	// test params
+	count := big.NewInt(5)
```
maru-ava: I would second copilot's comment - what does this value represent and why should it be 5?

tests/load2/main/main.go
```diff
@@ -107,14 +110,106 @@ func main() {
 	chainID, err := workers[0].Client.ChainID(ctx)
 	require.NoError(err)
 
+	txOpts, err := bind.NewKeyedTransactorWithChainID(workers[0].PrivKey, chainID)
```
maru-ava: Given that this block (L113-125) is effectively test setup, maybe encapsulate in a constructor for RandomTest for the sake of maintainability?

tests/load2/main/main.go
```diff
@@ -107,14 +110,106 @@ func main() {
...
 		registry,
-		load2.ZeroTransferTest{},
+		randomTest,
 	)
 	require.NoError(err)
 
 	generator.Run(tc, ctx, loadTimeout, testTimeout)
 }
+
+func createRandomTest(contract *contracts.EVMLoadSimulator) (load2.RandomTest, error) {
```
maru-ava: Why does this belong here rather than in tests.go?

tests/load2/tests.go
```diff
@@ -5,18 +5,86 @@ package load2
...
+	_ Test = (*HashingTest)(nil)
+	_ Test = (*MemoryTest)(nil)
+	_ Test = (*CallDepthTest)(nil)
+	_ Test = (*ContractCreationTest)(nil)
+	_ Test = (*PureComputeTest)(nil)
+	_ Test = (*LargeEventTest)(nil)
+	_ Test = (*ExternalCallTest)(nil)
+)
+
+type RandomTest struct {
```
maru-ava: (No action required) Given the apparent centrality of weighting to this type, maybe `RandomWeightedTest`?

tests/load2/tests.go
```diff
@@ -5,18 +5,86 @@ package load2
...
+	tests       []Test
+	weighted    sampler.Weighted
+	totalWeight uint64
+}
+
+func NewRandomTest(weightedTests []WeightedTest) (RandomTest, error) {
+	weighted := sampler.NewWeighted()
+
+	// Initialize weighted set
+	tests := make([]Test, len(weightedTests))
```
maru-ava, line 160: (No action required) Maybe encapsulate this conversion in a method of the `WeightedTest` struct?

tests/load2/tests.go
maru-ava: I think this is a good direction, and given that executeContractTx doesn't need anything that isn't provided as arguments, maybe just make it a function?

tests/load2/main/main.go
RodrigoVillar: `count` indicates how many times to repeat an action if the test permits it - e.g.
https://github.com/ava-labs/avalanchego/blob/2bd9aa0327f72990b1fcfda45ea5f645177f4f54/tests/load/c/contracts/EVMLoadSimulator.sol#L46

RodrigoVillar: `count` indicates how many times to repeat an action if the test permits it - e.g.

https://github.com/ava-labs/avalanchego/blob/2bd9aa0327f72990b1fcfda45ea5f645177f4f54/tests/load/c/contracts/EVMLoadSimulator.sol#L46

The decision to set `count = 5` is arbitrary but constant - load 1.0 used random counts which is not ideal if we want the individual tests (`ReadTest`, `WriteTest`, etc.) to be reproducible.

tests/load2/tests.go
RodrigoVillar: Done: d9f52b0

tests/load2/tests.go
RodrigoVillar: Done: 5b409b7

tests/load2/tests.go
RodrigoVillar: Done: c2ccfd4

tests/load2/main/main.go
RodrigoVillar: Moved to `tests.go`: c2ccfd4

tests/load2/main/main.go
RodrigoVillar: Done: c2ccfd4

tests/load2/main/main.go
maru-ava: Maybe summarize that in a comment?

tests/load2/main/main.go
```diff
@@ -107,12 +107,15 @@ func main() {
 	chainID, err := workers[0].Client.ChainID(ctx)
 	require.NoError(err)
 
+	randomTest, err := load2.NewRandomTests(ctx, chainID, &workers[0])
```
maru-ava: Why randomTest but NewRandomTests?

tests/load2/tests.go
maru-ava: I'd second that suggestion. It wouldn't have to be the default behavior, but I think it would make sense to accept an optional seed.

tests/load2/tests.go
```diff
@@ -5,18 +5,169 @@ package load2
...
 
-var _ Test = (*ZeroTransferTest)(nil)
+var maxFeeCap = big.NewInt(300000000000)
+
+// NewRandomTests creates a RandomWeightedTest containing a collection of EVM
+// load testing scenarios.
+//
+// This function handles the setup of the tests and also assigns each test
+// an equal weight, making them equally likely to be selected during random test execution.
+func NewRandomTests(ctx context.Context, chainID *big.Int, worker *Worker) (RandomWeightedTest, error) {
```
maru-ava: Why is this `Tests` plural despite returning `RandomWeightedTest`? Or maybe, why isn't it RandomWeightedTests given that it is a wrapper around a set of tests?

tests/load2/main/main.go
RodrigoVillar: Renamed to `NewRandomTest`: 2253934

tests/load2/main/main.go
RodrigoVillar: Done: 49119b9

tests/load2/tests.go
RodrigoVillar: Added source here: 7ebe451

tests/load2/main/main.go
RodrigoVillar: The tests need some weight if we want to use `sampler.Weighted`. As to why not use something like: 

https://github.com/ava-labs/avalanchego/blob/28d549171f1594bfd9c8645a7d261f16f9362760/utils/sampler/uniform.go#L7-L16

I'm more than certain that these values will be tweaked in the future such as the tests will no longer have a uniform probability.

Elvis339 (approved):

tests/load2/main/main.go
maru-ava: Maybe add a TODO indicating future intent so that the current static value makes sense?

RodrigoVillar: Done: d44a62d

PR #4072, opened 2025-07-10
Add WritePrometheusServiceDiscoveryConfigFile helper

aaronbuchwald:
This PR adds a helper to write a prometheus service discovery config file into the expected location within `~/.tmpnet`.

This is used within https://github.com/ava-labs/avalanchego/pull/4019 to write the new config file using the existing tmpnet <-> prometheus integration.


tests/load/prometheus.go
```diff
@@ -1,118 +0,0 @@
...
-	ready := make(chan struct{})
-	go func() {
-		close(ready)
-		err = s.server.Serve(listener)
-		if errors.Is(err, http.ErrServerClosed) {
-			return
-		}
-		runErrorBiDirectional <- err
-	}()
-	<-ready
```
aaronbuchwald, line 74: I was not sure the purpose of this, but have not seen any related issues after removing, so simplified this and moved it to another file where it can be shared.

tests/prometheus_server.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (s *PrometheusServer) Stop() error {
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	return s.server.Shutdown(shutdownCtx)
+}
+
+// Address returns the address the server is listening on.
```
RodrigoVillar, line 83: (nit) Shouldn't this be

```
// Address returns the address the server is exporting to.
```

tests/prometheus_server.go
```diff
@@ -0,0 +1,79 @@
...
+//
+// If the server has not started, the address will be empty.
+func (s *PrometheusServer) Address() string {
+	return s.addr
+}
+
+// URL returns the URL to request the prometheus metrics from the server.
+func (s *PrometheusServer) URL() string {
+	return "http://" + s.addr + s.path
+}
```
RodrigoVillar: This is unused, should we keep it?

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +276,43 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+}
+
+func applyGitHubLabels(sdConfigs []SDConfig) []SDConfig {
+	for _, sdConfig := range sdConfigs {
+		for label, value := range GetGitHubLabels() {
+			sdConfig.Labels[label] = value
+		}
+	}
+	return sdConfigs
+}
```
RodrigoVillar: (nit) Could this go after  `WritePrometheusServiceDiscoveryConfigFile()`?

tests/prometheus_server.go
aaronbuchwald: These are the same thing, the server is listening on a port. Exporting is the wrong word here. I think you're thinking of where it is exporting metrics to, which is the overall process, but this is the address that the server is listening on ie. see `net.Listen(...)` used in the body of `Start(...)`

tests/prometheus_server.go
aaronbuchwald: Removing

tests/fixture/tmpnet/monitor_processes.go
aaronbuchwald: Could you explain why? I assume you're asking because this function tries to place helper functions below where they are used?

aaronbuchwald: Addressed

RodrigoVillar: My original thought process was that exported functions *should generally* take precedence over un-exported functions.

Similar: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

tests/prometheus_server.go
RodrigoVillar: Ah I see what you mean, thanks for the clarification!

RodrigoVillar (approved):

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +276,43 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+	if err := os.MkdirAll(serviceDiscoveryDir, perms.ReadWriteExecute); err != nil {
+		return "", fmt.Errorf("failed to create service discovery dir: %w", err)
+	}
+
+	if withGitHubLabels {
+		sdConfigs = applyGitHubLabels(sdConfigs)
+	}
+
+	configPath := filepath.Join(serviceDiscoveryDir, name+".json")
+	configData, err := json.MarshalIndent(sdConfigs, "", "  ")
```
maru-ava: Prefer DefautlJSONMarshal

tests/prometheus_server.go
```diff
@@ -0,0 +1,74 @@
...
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	return s.server.Shutdown(shutdownCtx)
+}
+
+// Address returns the address the server is listening on.
+//
+// If the server has not started, the address will be empty.
+func (s *PrometheusServer) Address() string {
```
maru-ava, line 85: Maybe return s.server.Addr instead instead of storing the value twice?

tests/prometheus_server.go
```diff
@@ -0,0 +1,74 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
maru-ava, line 1: Maybe add docstrings to functions and document how the type is intended to be used?

tests/prometheus_server.go
```diff
@@ -0,0 +1,74 @@
...
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+type PrometheusServer struct {
+	addr     string
```
maru-ava: Maybe hard-code the address and intended path rather than using parameters given that all uses in this PR and #4019 use the same values? Or is there a reason to vary these values that I'm missing?

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +276,43 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
 	return filepath.Join(tmpnetDir, cmdName, "file_sd_configs"), nil
 }
 
+type SDConfig struct {
+	Targets []string          `json:"targets"`
+	Labels  map[string]string `json:"labels"`
+}
+
+func WritePrometheusServiceDiscoveryConfigFile(name string, sdConfigs []SDConfig, withGitHubLabels bool) (string, error) {
```
maru-ava: Why `[]SDConfig`? afaict there is only ever one.

Also, this name is out of control and triggering my java flashbacks. I'd much prefer `WritePrometheusSDConfig`

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +276,43 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+		sdConfigs = applyGitHubLabels(sdConfigs)
+	}
+
+	configPath := filepath.Join(serviceDiscoveryDir, name+".json")
+	configData, err := json.MarshalIndent(sdConfigs, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal config: %w", err)
+	}
+
+	return configPath, os.WriteFile(configPath, configData, perms.ReadWrite)
```
maru-ava: Maybe this is the norm for other parts of avalanchego, but I'd much prefer not to return an error with anything but empty values.

tests/prometheus_server.go
```diff
@@ -0,0 +1,74 @@
...
+	go func() {
+		defer close(errChan)
+		err := s.server.Serve(listener)
+		if errors.Is(err, http.ErrServerClosed) {
+			return
+		}
+		errChan <- err
+	}()
+
+	return errChan, nil
```
maru-ava: Given that this err channel is only relevant to shutdown, how about internalizing it and have it be checked by Stop() (i.e. and returned with errors.Join)? That way clients don't have to be exposed to this detail - they can just worry about Stop'ing the server.

tests/load2/main/main.go
```diff
@@ -71,28 +70,28 @@ func main() {
...
+	monitoringConfigFilePath, err := tmpnet.WritePrometheusServiceDiscoveryConfigFile("load-test", []tmpnet.SDConfig{
+		{
+			Targets: []string{metricsServer.Address()},
+			Labels:  network.GetMonitoringLabels(),
+		},
+	}, false)
+	require.NoError(err, "failed to generate monitoring config file")
 
-		require.NoError(metricsServer.Stop(), "failed to stop metrics server")
+	tc.DeferCleanup(func() {
```
maru-ava, line 78: This cleanup registration belongs immediately after the server start has completed without error so that a failure to write the SD file doesn't prevent the cleanup from being registered.

tests/load/c/main/main.go
```diff
@@ -69,20 +69,21 @@ func main() {
...
-	monitoringConfigFilePath, err := metricsServer.GenerateMonitoringConfig(log, network.GetMonitoringLabels())
+	monitoringConfigFilePath, err := tmpnet.WritePrometheusServiceDiscoveryConfigFile("load-test", []tmpnet.SDConfig{
+		{
+			Targets: []string{metricsServer.Address()},
+			Labels:  network.GetMonitoringLabels(),
+		},
+	}, false)
 	require.NoError(err, "failed to generate monitoring config file")
 
 	tc.DeferCleanup(func() {
```
maru-ava, line 74: This cleanup registration belongs immediately after the server start has completed without error so that a failure to write the SD file doesn't prevent the cleanup from being registered.

maru-ava:
> We should brainstorm a bit and consider exposing monitoring service (Prometheus) rather than using it from tmpnet.

Why?

tests/fixture/tmpnet/monitor_processes.go
aaronbuchwald: Following the style from https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config, but there is only ever one, so will update the function signature.

Was aiming to follow the style above in `GetPrometheusServiceDiscoveryDir`, but this has gotten a bit longer, so will update it.

tests/prometheus_server.go
aaronbuchwald: Updating to hardcode it

tests/prometheus_server.go
aaronbuchwald: This is just migrating from another file and cleaning up what I thought was some unnecessary complexity, so didn't think documenting the existing type should be in scope for this PR, but happy to do so if you prefer

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +275,54 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+	Targets []string          `json:"targets"`
+	Labels  map[string]string `json:"labels"`
+}
+
+// WritePrometheusSDConfig writes a Prometheus SD config file
+// with the provided sdConfig to the location expected by tmpnet.
+// If withGitHubLabels is true, it attempts to append GitHub-specific labels
+// to the SDConfig before writing.
+//
+// Returns the path to the written configuration file.
```
aaronbuchwald:
```suggestion
// WritePrometheusSDConfig writes the SDConfig with the provided name
// to the location expected by tmpnet.
// If withGitHubLabels is true, it attempts to append GitHub-specific labels
// to the SDConfig before writing.
//
// Returns the path to the written configuration file.
```

tests/prometheus_server.go
maru-ava: Please do. tests/load was merged with less oversight than normal with the explicit expectation that it would be replaced or rewritten in the near term, so anything copied from that path can't be considered 'done'.

tests/prometheus_server.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return nil
+}
+
+func (s *PrometheusServer) Stop() error {
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	err := s.server.Shutdown(shutdownCtx)
+	if err != nil {
```
maru-ava: I was expecting to see the use of errors.Join so that both the error from Shutdown and from the error channel could be returned. Is that not desirable?

tests/load2/main/main.go
```diff
@@ -74,28 +73,23 @@ func main() {
...
 	})
 
-	monitoringConfigFilePath, err := metricsServer.GenerateMonitoringConfig(
-		log,
-		network.GetMonitoringLabels(),
-	)
-	require.NoError(err)
+	monitoringConfigFilePath, err := tmpnet.WritePrometheusSDConfig("load-test", tmpnet.SDConfig{
+		Targets: []string{metricsServer.Address()},
+		Labels:  network.GetMonitoringLabels(),
```
maru-ava, line 84: Is there an expectation of varying the monitoring labels? If not, maybe update WritePrometheusSDConfig to accept the server address and internalize the construction of an SDConfig?

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +275,54 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+// SDConfig represents a Prometheus service discovery config entry.
+//
+// file_sd_config docs: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config
+type SDConfig struct {
+	Targets []string          `json:"targets"`
+	Labels  map[string]string `json:"labels"`
+}
+
+// WritePrometheusSDConfig writes the SDConfig with the provided name
+// to the location expected by tmpnet.
```
maru-ava: by tmpnet -> by the prometheus instance configured by tmpnet

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -273,6 +275,54 @@ func getServiceDiscoveryDir(cmdName string) (string, error) {
...
+// If withGitHubLabels is true, it attempts to append GitHub-specific labels
```
maru-ava: Maybe remove 'attempts' given that there is no error condition associated with application of github labels?

maru-ava: Also, where is this going to be true? Usage in this PR only supplies false.

tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -261,6 +261,8 @@ func getWorkingDir(cmdName string) (string, error) {
 
 // GetPrometheusServiceDiscoveryDir returns the path for prometheus file-based
 // service discovery configuration.
+//
+// file_sd_config docs: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config
```
maru-ava: Maybe include a link here or for SDConfig but not for both?

tests/load2/main/main.go
aaronbuchwald: Yes, #4019 allows devs to specify custom labels to attach to their metrics. This is useful to attach my own metrics to a local run outside of CI, such as `owner=aaron`

tests/fixture/tmpnet/monitor_processes.go
aaronbuchwald: 'attempts' here is meant to convey that it looks for them in the environment, but they are not guaranteed to be populated, so even though there is no error handling, it is still an 'attempt'.

This will be true in #4019, where it's used to attach the GitHub env variables to the metrics.

tests/fixture/tmpnet/monitor_processes.go
aaronbuchwald: Removing here and leaving on the `SDConfig` type

tests/prometheus_server.go
aaronbuchwald: Will switch to `errors.Join(...)`

tests/fixture/tmpnet/monitor_processes.go
aaronbuchwald: Updating to be more specific:

```
// WritePrometheusSDConfig writes the SDConfig with the provided name
// to the location expected by the prometheus instance start by tmpnet.
//
// If withGitHubLabels is true, checks env vars for GitHub-specific labels
// and adds them as labels if present before writing the SDConfig.
//
// Returns the path to the written configuration file.
```

aaronbuchwald: The alternative here is to just leave to the caller to gather the GitHub labels before calling this function.

aaronbuchwald: Draft here if removing is preferred https://github.com/ava-labs/avalanchego/pull/4090

tests/load2/main/main.go
maru-ava: So maybe add 2 methods - one that uses default labels implicitly and one that accepts labesl? I don't think it makes sense to add complexity for the majority of callpoints just because one needs arbitrary labeling.

tests/prometheus_server.go
aaronbuchwald: Collapsed start into the constructor and documented it here: https://github.com/ava-labs/avalanchego/pull/4072/commits/27f20b0c69e83a019c39a4c773976264f8b6a141

PR #4083, opened 2025-07-14
Refactor buf modules

joshua-kim:
## Why this should be merged

We compile our buf modules seprately since the `proto` module doesn't want to compile stuff with the `connect` plugin. This PR puts our buf modules under a common directory instead of separate `proto` and `connectproto` directories.

## How this works

Moves files

## How this was tested

CI

No


.github/workflows/ci.yml
```diff
@@ -131,7 +131,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: bufbuild/buf-action@dfda68eacb65895184c76b9ae522b977636a2c47 #v1.1.4
         with:
-          input: "proto"
```
joshua-kim, line 134: Upon making changes to our CI I think there was a miss and we need another github action to run the linter on the connect module as a follow-up to this.

joshua-kim: Issue: https://github.com/ava-labs/avalanchego/issues/4084

joshua-kim:
Coreth branch: https://github.com/ava-labs/coreth/pull/1047

.github/workflows/ci.yml
maru-ava: Is there a way to _not_ use a github action for a new linter? Ideally this one should go away in favor of a task that can be executed both locally and in CI.

mpignatelli12 (approved):

.github/workflows/ci.yml
joshua-kim: Hm - I'm being dumb... our `check-generate-protobuf` nix task which is already in ci runs linting under the `connectproto` directory already (which also does run locally).

It's actually unclear to me what this CI step is for - since it doesn't push the buf schema and just runs the linter.

maru-ava: Suggest removing then.

joshua-kim: PR: https://github.com/ava-labs/avalanchego/pull/4189

PR #4085, opened 2025-07-14
Measure bloom filter AppGossip hit rate

yacovm:
## Why this should be merged

It can be insightful to measure how effective and efficient is our gossip pull mechanism that is based on bloom filters.

## How this works

This commit adds a metric to measure the bloom filter hit rate % for the gossip pull queries.

## How this was tested

- Updated `TestGossiperGossip` accordingly to reflect the metrics
- Ran a modified avalanchego node and installed prometheus and grafana and observed the following:

image

No.

### DIFF
--- a/snow/engine/common/notifier.go
+++ b/snow/engine/common/notifier.go
@@ -6,11 +6,14 @@ package common
 import (
 	"context"
 	"sync"
+	"time"
 
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
+
+const errThrottleTime = 100 * time.Millisecond
 
 // Subscription is a function that blocks until either the given context is cancelled, or a message is returned.
 // It is used to receive messages from a VM such as Pending transactions, state sync completion, etc.
@@ -73,6 +76,11 @@ func (nf *NotificationForwarder) forwardNotification() {
 	msg, err := nf.Subscribe(ctx)
 	if err != nil {
 		nf.Log.Debug("Failed subscribing to notifications", zap.Error(err))
+		// Wait to retry
+		select {
+		case <-time.After(errThrottleTime):
+		case <-ctx.Done():
+		}
 		return
 	}
 
@@ -80,6 +88,11 @@ func (nf *NotificationForwarder) forwardNotification() {
 
 	if err := nf.Engine.Notify(ctx, msg); err != nil {
 		nf.Log.Debug("Failed notifying engine", zap.Error(err))
+		// Wait to retry
+		select {
+		case <-time.After(errThrottleTime):
+		case <-ctx.Done():
+		}
 		return
 	}
 
--- a/tests/fixture/tmpnet/flags/kube_runtime.go
+++ b/tests/fixture/tmpnet/flags/kube_runtime.go
@@ -20,10 +20,9 @@ const (
 )
 
 var (
-	errKubeNamespaceRequired       = errors.New("--kube-namespace is required")
-	errKubeImageRequired           = errors.New("--kube-image is required")
-	errKubeMinVolumeSizeRequired   = fmt.Errorf("--kube-volume-size must be >= %d", tmpnet.MinimumVolumeSizeGB)
-	errKubeSchedulingLabelRequired = errors.New("--kube-scheduling-label-key and --kube-scheduling-label-value are required when --kube-use-exclusive-scheduling is enabled")
+	errKubeNamespaceRequired     = errors.New("--kube-namespace is required")
+	errKubeImageRequired         = errors.New("--kube-image is required")
+	errKubeMinVolumeSizeRequired = fmt.Errorf("--kube-volume-size must be >= %d", tmpnet.MinimumVolumeSizeGB)
 )
 
 type kubeRuntimeVars struct {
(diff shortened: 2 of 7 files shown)


network/p2p/gossip/handler.go
```diff
@@ -124,3 +135,31 @@ func (h Handler[_]) AppGossip(_ context.Context, nodeID ids.NodeID, gossipBytes
...
+
+	hitsOneHundred, err := safemath.Mul(hits, 100)
+	if err != nil {
+		log.Warn("failed to calculate hit ratio",
+			zap.Uint64("hits", hits),
+			zap.Uint64("misses", misses),
+			zap.Error(err),
+		)
+		return 0, false
+	}
```
joshua-kim: I wonder if we can get rid of these overflow checks... since the bloom filter is `[]byte` (which can be at most of size max int) could we just emit a metric on both the hits + length of the filter? We could figure out the hit/miss % in dashboards with metric math.

network/p2p/gossip/gossip_test.go
```diff
@@ -123,6 +131,12 @@ func TestGossiperGossip(t *testing.T) {
 
 			metrics, err := NewMetrics(prometheus.NewRegistry(), "")
 			require.NoError(err)
+
+			testHistogram := &testHistogram{
+				Histogram: metrics.bloomFilterHitRate,
+			}
+			metrics.bloomFilterHitRate = testHistogram
+
```
joshua-kim, line 136: This looks a bit hacky to me, if we want to do an assert that the metrics are the values we're expecting would it make sense for us to consider using [prometheus/testutil](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil) here?

network/p2p/gossip/gossip_test.go
```diff
@@ -611,3 +627,58 @@ type testValidatorSet struct {
 func (t testValidatorSet) Has(_ context.Context, nodeID ids.NodeID) bool {
 	return t.validators.Contains(nodeID)
 }
+
+func TestComputeBloomFilterHitPercentage(t *testing.T) {
```
joshua-kim: I feel like non-exported code is an implementation detail of the package api so where reasonable I try to test non-exported code through the public api. Since we're already able to test this through AppRequest in the above test, should we just remove this test?

network/p2p/gossip/gossip_test.go
yacovm: `testutil.ToFloat64` doesn't work with histograms. Do you have a suggestion?

network/p2p/gossip/handler.go
yacovm: > I wonder if we can get rid of these overflow checks... since the bloom filter is []byte (which can be at most of size max int) could we just emit a metric on both the hits + length of the filter? We could figure out the hit/miss % in dashboards with metric math.

I am not sure what you mean by the length of the filter. The bloom filter is constant size and its length is always the same, but yet independent of the elements that it can "contain". 

I guess you meant the size of the mempool? Since the maximum number of hits we can get is the size of the mempool, not the bloom filter. Sure I can do that, that's a good point.

yacovm: Unfortunately we don't have a Size() on the set interface so I will try to code it without the safemath.

network/p2p/gossip/gossip_test.go
yacovm: Sure will remove it.

yacovm: [Removed](https://github.com/ava-labs/avalanchego/pull/4085/commits/b21c5a553095c4a7feb11523272b1dc8f3ca336f#diff-ec31b0b63659ae05816344564ef472d1dea4c2e60d8b6ba344339e37eaebab81) the test.

network/p2p/gossip/handler.go
yacovm: [Changed](https://github.com/ava-labs/avalanchego/pull/4085/commits/b21c5a553095c4a7feb11523272b1dc8f3ca336f#diff-64f640fd349509626d5061afbee4d706338a2f04f12a0bca5cf2a96447d614d5) the code to not use safemath.

network/p2p/gossip/handler.go
```diff
@@ -77,6 +81,11 @@ func (h Handler[T]) AppRequest(_ context.Context, _ ids.NodeID, _ time.Time, req
 		return nil, p2p.ErrUnexpected
 	}
 
+	hitsPercentage, ok := computeBloomFilterHitPercentage(hits, total, h.log)
+	if ok {
+		h.metrics.bloomFilterHitRate.Observe(float64(hitsPercentage))
+	}
```
StephenButtolph: `computeBloomFilterHitPercentage` seems a bit over-engineered imo.

1. We currently cast `float64 -> uint64 -> float64`, we should just keep the values as `float64`.
2. I don't think we really get anything out of handling `hits > total`, it is realistically impossible for `uint64` to overflow due to incrementing (it would take over 30k years incrementing once per nanosecond to overflow here)..

```suggestion
	if total != 0 {
		h.metrics.bloomFilterHitRate.Observe(100*(float64(hits)/float64(total)))
	}
```

yacovm: https://github.com/ava-labs/avalanchego/pull/4085/commits/088c90868f79e393d888ddb0a2aec06d4545b155

StephenButtolph (approved):

joshua-kim (approved):

PR #4088, opened 2025-07-15
feat: add eviction callback in LRU cache

DracoLi:
## Why this should be merged

Adds an `onEvict` callback to the LRU cache to allow resource cleanup when entries are evicted.

`x/blockdb` (https://github.com/ava-labs/avalanchego/pull/4027) uses the LRU cache to store file descriptors, so this allows `blockdb` to properly close the file descriptors when an entry is evicted.

## How this works

- Adds `NewCacheWithOnEvict` to set an optional cleanup function
- The callback is called whenever an entry is evicted (during `Put`, `Evict`, or `Flush`)
- Existing code continues to work unchanged

## How this was tested

Unit tests added

Don't think so


cache/lru/cache_test.go
```diff
@@ -19,3 +21,30 @@ func TestCacheEviction(t *testing.T) {
 	c := NewCache[ids.ID, int64](2)
 	cachetest.Eviction(t, c)
 }
+
```
yacovm, line 24: why not put the OnEvict test logic in the `cachetest.Eviction` and `cachetest.Basic` ? Then we can do more fine grained testing and not test everything in the end of the test.

DracoLi: `cachetest` is used to test all implementations of `Cacher`. Which is used by the `SizedCache` as well. Here I only added onEvict to the `Cache` struct.
I didn't think we need `onEvict` on the interface, thoughts?

yacovm: Oh I missed that part. Maybe then just write your own test then? 

Because the current test just runs through a bunch of operations and then tests a final result. For example the eviction test, does both puts and evicts, which is noisy if you test only the eviction function in the end of the test.

DracoLi: yeah we don't need to reuse it those tests. I can write some custom tests for this.

DracoLi: https://github.com/ava-labs/avalanchego/pull/4088/commits/236e6fd2e31c82b2572246c6f78a9895554f7971

cache/lru/cache.go
```diff
@@ -20,6 +20,20 @@ type Cache[K comparable, V any] struct {
...
+	// onEvict is called with the key and value of an entry before eviction, if set.
+	onEvict func(K, V)
+}
+
+// SetOnEvict sets a callback to be called with the key and value of an entry before eviction.
+// The onEvict callback is called while holding the cache lock.
+// Do not call any cache methods (Get, Put, Evict, Flush) from within the callback
+// as this will cause a deadlock. The callback should only be used for cleanup
+// operations like closing files or releasing resources.
+func (c *Cache[K, V]) SetOnEvict(cb func(K, V)) {
```
joshua-kim: Would it be reasonable for us to avoid adding this `Set*` function + callback approach? I generally find callbacks difficult/confusing to debug and reason about. If we refactored the `Cacher` interface to just return the evicted value to something like:

```
// Cacher acts as a best effort key value store.
type Cacher[K comparable, V any] interface {
	// Put inserts an element into the cache. If space is required, elements will
	// be evicted.
	Put(key K, value V) (V, bool) // Expose evicted value

...

	// Evict removes the specified entry from the cache
	Evict(key K) (V, bool) // Expose evicted value

...
}
```

Downstream consumers of the api could get the evicted information from just checking the return value.

DracoLi: I think this approach makes sense and definitely looks simpler. My only concern is that we’d need to update `SizedCache` to also return evicted values, even though it’s not needed atm. That might be fine, but it does add some complications compared to just adding `OnEvict` to `Cache`.

- `SizedCache` can evict multiple entries on `Put`, so the interface would probably need to return all evicted values, not just a single one. It also flushes everything if the new entry size exceeds `maxSize`, which means we’d need to loop and return all evicted elements even if the caller doesn’t care about them.
- Similarly for Flush, we’d have to iterate through and return all elements even if consumer might not need it. Alternatively, we can add an iterator over the cache elements instead and not have Flush return evicted elements.

joshua-kim: Ah I didn't know `SizedCache` existed. If it's overkill we can ignore my idea - we can always throw stuff in the backlog.

joshua-kim: Why does this need to exist as a setter? Can we get away with just initializing the map with a callback instead? I suspect we don't need to support dynamically updating the on eviction behavior

cache/lru/cache.go
```diff
@@ -55,14 +72,26 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
 func (c *Cache[K, _]) Evict(key K) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-
+	if c.onEvict != nil {
```
joshua-kim: Instead of doing this nil check we could just pass in a callback that does nothing on initialize

cache/lru/cache_test.go
```diff
@@ -19,3 +21,89 @@ func TestCacheEviction(t *testing.T) {
...
+func TestCacheOnEvict(t *testing.T) {
+	tests := []struct {
+		name            string
+		cacheSize       int
+		operations      func(*Cache[ids.ID, int64])
+		expectedEvicted map[ids.ID]int64
```
joshua-kim: Can we use a slice or something to make sure that the eviction ordering matters?

cache/lru/cache_test.go
```diff
@@ -19,3 +21,89 @@ func TestCacheEviction(t *testing.T) {
...
+				{1}: int64(1), // evicted by Put
+				{2}: int64(2), // evicted by Evict
+				{3}: int64(3), // evicted by Flush
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := NewCache[ids.ID, int64](tt.cacheSize)
```
joshua-kim: nit: is there a reason we have to use `ids.ID` and `int64`? Can we keep it simple and just primitives like a map of `map[int]int` or something?

cache/lru/cache.go
DracoLi: yeah that would make things cleaner. one benefit of `nil` is that we can avoid iteration over elements on flush if we don't care about the callback, which is most cases right now.

cache/lru/cache_test.go
DracoLi: don't believe so, i followed the type for existing cache tests. I think two ints works fine.

cache/lru/cache.go
DracoLi: good idea, setting this on Initialization is better.

cache/lru/cache.go
```diff
@@ -55,14 +77,26 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
...
+		if value, found := c.elements.Get(key); found {
```
joshua-kim: nit: Go typically uses `ok` instead of `found` for the var name for these existence checks

cache/lru/cache.go
```diff
@@ -20,12 +20,31 @@ type Cache[K comparable, V any] struct {
...
+	}
+}
+
+// NewCacheWithCallback creates a new LRU cache with the given size and eviction callback.
+// The onEvict callback is called with the key and value of an entry before eviction.
+// The onEvict callback is called while holding the cache lock.
+// Do not call any cache methods (Get, Put, Evict, Flush) from within the callback
+// as this will cause a deadlock. The callback should only be used for cleanup
+// operations like closing files or releasing resources.
+func NewCacheWithCallback[K comparable, V any](size int, onEvict func(K, V)) *Cache[K, V] {
```
joshua-kim: nit: Can we call this `NewCacheWithOnEvict` or `NewCacheWithOnEvictCallback` or something along those lines to make it clear what the callback does?

cache/lru/cache.go
```diff
@@ -20,12 +20,31 @@ type Cache[K comparable, V any] struct {
...
+		onEvict:  func(K, V) {}, // no-op by default
+	}
+}
+
+// NewCacheWithOnEvict creates a new LRU cache with the given size and eviction callback.
+// The onEvict callback is called with the key and value of an entry before eviction.
+// The onEvict callback is called while holding the cache lock.
+// Do not call any cache methods (Get, Put, Evict, Flush) from within the callback
+// as this will cause a deadlock. The callback should only be used for cleanup
+// operations like closing files or releasing resources.
```
joshua-kim: "The callback should only be used for cleanup operations like closing files or releasing resources." this seems a bit misleading since the caller can do anything as long as it doesn't require holding the lock. Should we just remove this sentence?

cache/lru/cache.go
```diff
@@ -34,7 +53,10 @@ func (c *Cache[K, V]) Put(key K, value V) {
 	defer c.lock.Unlock()
 
 	if c.elements.Len() == c.size {
-		oldestKey, _, _ := c.elements.Oldest()
+		oldestKey, oldestValue, found := c.elements.Oldest()
+		if found {
```
joshua-kim: nit: use `ok`

cache/lru/cache.go
```diff
@@ -55,15 +77,22 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
 func (c *Cache[K, _]) Evict(key K) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-
+	if value, ok := c.elements.Get(key); ok {
+		c.onEvict(key, value)
+	}
 	c.elements.Delete(key)
```
joshua-kim: nit: early return for readability

```suggestion
	value, ok := c.elements.Get(key);
	if !ok {
	    return 
	}
	
	c.elements.Delete(key)
	c.onEvict(key, value)
```

cache/lru/cache.go
```diff
@@ -55,15 +77,22 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
...
 func (c *Cache[_, _]) Flush() {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	c.elements.Clear()
+	// Call onEvict for each element before clearing
+	iter := c.elements.NewIterator()
+	for iter.Next() {
+		c.onEvict(iter.Key(), iter.Value())
+		c.elements.Delete(iter.Key())
```
joshua-kim: Since we need to call the `onEvict` callback... it feels like `Flush` is now becoming an implementation of `Evict` over the entire `Cache`. Would it make sense to refactor the locked part of `Evict` into something like `evict` so we can just call `evict` over each element in this loop so we can avoid re-implementing `Evict` here?

cache/lru/cache.go
DracoLi: yeah i can remove

cache/lru/cache.go
DracoLi: yeah, we can just use Evict here.

cache/lru/cache.go
```diff
@@ -34,7 +48,10 @@ func (c *Cache[K, V]) Put(key K, value V) {
 	defer c.lock.Unlock()
 
 	if c.elements.Len() == c.size {
-		oldestKey, _, _ := c.elements.Oldest()
+		oldestKey, oldestValue, ok := c.elements.Oldest()
+		if ok {
+			c.onEvict(oldestKey, oldestValue)
+		}
 		c.elements.Delete(oldestKey)
```
joshua-kim: nit: can we call `evict` here as well?

cache/lru/cache.go
```diff
@@ -55,15 +72,19 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
...
+	if !ok {
+		return
+	}
+	c.onEvict(key, value)
 	c.elements.Delete(key)
 }
 
 func (c *Cache[_, _]) Flush() {
-	c.lock.Lock()
-	defer c.lock.Unlock()
```
joshua-kim: I think we still want to hold the lock here to guarantee that `Flush` returns with an empty cache

cache/lru/cache.go
```diff
@@ -55,15 +72,19 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
 func (c *Cache[K, _]) Evict(key K) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-
+	value, ok := c.elements.Get(key)
+	if !ok {
+		return
+	}
+	c.onEvict(key, value)
 	c.elements.Delete(key)
```
joshua-kim: Sorry if my previous comment was vague - i meant to refactor this blob of code into a non-exported `evict` so that the rest of the cache code could call `evict` when it needed to remove an element + have the `onEvict` callback called.

cache/lru/cache.go
```diff
@@ -20,12 +20,26 @@ type Cache[K comparable, V any] struct {
...
+	onEvict func(K, V)
 }
 
+// NewCache creates a new LRU cache with the given size.
 func NewCache[K comparable, V any](size int) *Cache[K, V] {
 	return &Cache[K, V]{
 		elements: linked.NewHashmap[K, V](),
 		size:     max(size, 1),
+		onEvict:  func(K, V) {}, // no-op by default
+	}
```
joshua-kim: Could we update this constructor to call `NewCacheWithOnEvict` and pass in the no-op callback?

cache/lru/cache.go
```diff
@@ -34,8 +48,10 @@ func (c *Cache[K, V]) Put(key K, value V) {
 	defer c.lock.Unlock()
 
 	if c.elements.Len() == c.size {
-		oldestKey, _, _ := c.elements.Oldest()
-		c.elements.Delete(oldestKey)
+		oldestKey, _, ok := c.elements.Oldest()
+		if ok {
+			c.evict(oldestKey)
+		}
```
joshua-kim: nit: I generally in-line var initialization to if-scopes where possible to make it clear that the var isn't used anywhere else... it's such a preference and it's unimportant so you can ignore this suggestion if you want.
```suggestion
		if oldestKey, _, ok := c.elements.Oldest(); ok {
			c.evict(oldestKey)
		}
```

cache/lru/cache.go
```diff
@@ -55,15 +71,26 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
 func (c *Cache[K, _]) Evict(key K) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
+	c.evict(key)
```
joshua-kim: nit: spacing
```suggestion

	c.evict(key)
```

cache/lru/cache.go
```diff
@@ -55,15 +71,26 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
...
+}
 
+func (c *Cache[K, V]) evict(key K) {
+	value, ok := c.elements.Get(key)
+	if !ok {
+		return
+	}
+	c.onEvict(key, value)
```
joshua-kim: nit: spacing
```suggestion
	}
	
	c.onEvict(key, value)
```

cache/lru/cache.go
```diff
@@ -55,15 +71,26 @@ func (c *Cache[K, V]) Get(key K) (V, bool) {
...
 	c.elements.Delete(key)
 }
 
 func (c *Cache[_, _]) Flush() {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	c.elements.Clear()
+	iter := c.elements.NewIterator()
+	for iter.Next() {
```
joshua-kim: nit: more pedantic nits w.r.t in-lining scopes where possible
```suggestion
	for iter := c.elements.NewIterator(); iter.Next();  {
```

cache/lru/cache.go
```diff
@@ -56,14 +67,23 @@ func (c *Cache[K, _]) Evict(key K) {
...
 
-	c.elements.Delete(key)
+	c.evict(key)
+}
+
+func (c *Cache[K, V]) evict(key K) {
+	if value, ok := c.elements.Get(key); ok {
+		c.onEvict(key, value)
+		c.elements.Delete(key)
+	}
```
joshua-kim: nit: early return where possible for readability
```suggestion
	value, ok := c.elements.Get(key)
	if !ok {
		return
	}
	
	c.onEvict(key, value)
	c.elements.Delete(key)
```

joshua-kim (approved):

cache/lru/cache.go
```diff
@@ -56,14 +67,26 @@ func (c *Cache[K, _]) Evict(key K) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	c.evict(key)
+}
+
+func (c *Cache[K, V]) evict(key K) {
+	value, ok := c.elements.Get(key)
```
yacovm, line 70: When do we check and get `!ok` ?  I would guess only upon `Evict()`, right? If so, wouldn't it make sense to move the check there?

DracoLi: yeah, we can also remove the `Get` here since `Flush` and `Put` have the value already.

PR #4091, opened 2025-07-16
Simplex QuorumCertificate and BLS aggregator

samliok:
## Why this should be merged

Implements the simplex `QuorumCertificate`, `QCDeserializer` and `SignatureAggregator` interfaces. This allows simplex to parse, aggregate and verify quorum certificates(ex. finalizations and notarizations) during execution. 

## How this works

- Builds on top of the BLSVerifier to handle BLS signatures and public keys. 
- The bytes of a QC are serialized with the `Codec`

## How this was tested

Added unit tests to `qc_test.go`.

no


simplex/quorum.go
```diff
@@ -0,0 +1,183 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
yacovm: maybe call the file `qc.go` ?

simplex/quorum.go
```diff
@@ -0,0 +1,183 @@
...
+	// QC errors
+	errFailedToParseQC       = errors.New("failed to parse quorum certificate")
+	errUnexpectedSigners     = errors.New("unexpected number of signers in quorum certificate")
+	errSignatureAggregation  = errors.New("signature aggregation failed")
+	errEncodingMessageToSign = errors.New("failed to encode message to sign")
+)
+
+// QC represents a quorum certificate in the Simplex consensus protocol.
+type QC struct {
+	verifier BLSVerifier
```
yacovm: why not a pointer to `BLSVerifier` ?

simplex/quorum.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Signers returns the list of signers for the quorum certificate.
+func (qc *QC) Signers() []simplex.NodeID {
+	return qc.signers
+}
+
+// Verify checks if the quorum certificate is valid by verifying the aggregated signature against the signers' public keys.
+func (qc *QC) Verify(msg []byte) error {
+	pks := make([]*bls.PublicKey, 0, len(qc.signers))
+	if len(qc.signers) != simplex.Quorum(len(qc.verifier.nodeID2PK)) {
```
yacovm:
```
quorum := simplex.Quorum(len(qc.verifier.nodeID2PK))
	if len(qc.signers) != quorum {
		return fmt.Errorf("%w: expected %d signers but got %d", errUnexpectedSigners, quorum, len(qc.signers))
	}
```

simplex/quorum.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if !bls.Verify(aggPK, &qc.sig, message2Verify) {
+		return errSignatureVerificationFailed
+	}
+
+	return nil
+}
+
+// asn1QC is the ASN.1 structure for the quorum certificate.
```
yacovm: The QC bytes play no cryptographic role, so we can just use the regular serialization way with Codec.

simplex/quorum.go
```diff
@@ -0,0 +1,183 @@
...
+
+var (
+	_ simplex.QuorumCertificate   = (*QC)(nil)
+	_ simplex.QCDeserializer      = QCDeserializer{}
+	_ simplex.SignatureAggregator = (*SignatureAggregator)(nil)
+
+	// QC errors
+	errFailedToParseQC       = errors.New("failed to parse quorum certificate")
+	errUnexpectedSigners     = errors.New("unexpected number of signers in quorum certificate")
+	errSignatureAggregation  = errors.New("signature aggregation failed")
```
yacovm: `errSignatureAggregation` - seems like this isn't tested... can we add a test to it?

simplex/qc.go
```diff
@@ -0,0 +1,160 @@
...
+// Bytes serializes the quorum certificate into bytes.
+func (qc *QC) Bytes() []byte {
+	serializedQC := &SerializedQC{
+		Sig:     bls.SignatureToBytes(qc.sig),
+		Signers: qc.signers,
+	}
+
+	bytes, err := Codec.Marshal(CodecVersion, serializedQC)
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal QC: %w", err))
```
samliok: todo: don't panic, update the simplex interface to allow an error return value

simplex/quorum.go
samliok: not sure if its worth creating a test for this. we return this error when `bls.AggregateSignatures(signatures)` fails and we already make sure all the `signatures` are valid before calling.

yacovm (approved):

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+type QC struct {
+	verifier *BLSVerifier
+	sig      *bls.Signature
+	signers  []simplex.NodeID
+}
+
+type SerializedQC struct {
+	Sig     []byte           `serialize:"true"`
+	Signers []simplex.NodeID `serialize:"true"`
+}
```
StephenButtolph: I know we use the avalanchego codec for formatting the BLS messages to partition the message space from Warp messages... But I feel like we should be using `canoto` wherever possible in the simplex code.

simplex/qc.go
StephenButtolph: If we replace the codec usage with canoto, then we won't need to change the interface

simplex/qc.go
StephenButtolph: Why do we serialize the full list of nodeIDs? Shouldn't we be using a bitset to indicate participation similarly to warp?

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+func (qc *QC) Verify(msg []byte) error {
+	pks := make([]*bls.PublicKey, 0, len(qc.signers))
+	quorum := simplex.Quorum(len(qc.verifier.nodeID2PK))
+	if len(qc.signers) != quorum {
+		return fmt.Errorf("%w: expected %d signers but got %d", errUnexpectedSigners, quorum, len(qc.signers))
+	}
+
+	// ensure all signers are in the membership set
+	for _, signer := range qc.signers {
+		pk, exists := qc.verifier.nodeID2PK[ids.NodeID(signer)]
```
StephenButtolph: In `DeserializeQuorumCertificate`, we never verify that `signer` is at least `20` bytes long.

Here, we cast `signer` (with type `simplex.NodeID`) into `ids.NodeID`. If `signer` is less than `20` bytes, this will panic.

We need to be very careful when handling untrusted and unverified inputs.

Conversions from `[]byte` to `ids.NodeID` must always use `ids.ToNodeID` to ensure that no panics can happen.

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+	return qc.signers
+}
+
+// Verify checks if the quorum certificate is valid by verifying the aggregated signature against the signers' public keys.
+func (qc *QC) Verify(msg []byte) error {
+	pks := make([]*bls.PublicKey, 0, len(qc.signers))
+	quorum := simplex.Quorum(len(qc.verifier.nodeID2PK))
+	if len(qc.signers) != quorum {
+		return fmt.Errorf("%w: expected %d signers but got %d", errUnexpectedSigners, quorum, len(qc.signers))
+	}
```
StephenButtolph: Are there checks somewhere else that verify that the same signer isn't included multiple times?

If so we should document that assumption here. If not we need to add that.

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+
+// QC represents a quorum certificate in the Simplex consensus protocol.
+type QC struct {
+	verifier *BLSVerifier
+	sig      *bls.Signature
+	signers  []simplex.NodeID
+}
+
+type SerializedQC struct {
+	Sig     []byte           `serialize:"true"`
```
StephenButtolph: By using `[]byte` for the signature, the serialized format will need to include the length of the signature. (or if we switch this to canoto, will at least require us to verify the length).

This is why in warp we specify the length here explicitly:
```suggestion
	Sig     [bls.SignatureLen]byte           `serialize:"true"`
```

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, fmt.Errorf("%w: %w", errFailedToParseSignature, err)
+	}
+
+	qc := QC{
+		sig:     sig,
+		signers: serializedQC.Signers,
+	}
+
+	verifier := BLSVerifier(d)
+	qc.verifier = &verifier
```
StephenButtolph: Are we intentionally copying the `BLSVerifier` struct here? I would have probably expected this to be referencing the same `BLSVerifier` struct.

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		sigs = append(sigs, sig)
+	}
+
+	aggregatedSig, err := bls.AggregateSignatures(sigs)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errSignatureAggregation, err)
+	}
+
+	verifier := BLSVerifier(a)
```
StephenButtolph: Similar question, is this intentionally a copy?

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+	return &qc, nil
+}
+
+// SignatureAggregator aggregates signatures into a quorum certificate.
+type SignatureAggregator BLSVerifier
+
+// Aggregate aggregates the provided signatures into a quorum certificate.
+// It requires at least a quorum of signatures to succeed.
+// If any signature is from a signer not in the membership set, it returns an error.
+func (a SignatureAggregator) Aggregate(signatures []simplex.Signature) (simplex.QuorumCertificate, error) {
```
StephenButtolph: This probably doesn't need to include any handling of duplicates, as it should be unexpected for simplex to provide duplicates here (imo).

simplex/qc.go
```diff
@@ -0,0 +1,159 @@
...
+	quorumSize := simplex.Quorum(len(a.nodeID2PK))
+	if len(signatures) < quorumSize {
+		return nil, fmt.Errorf("%w: expected %d signatures but got %d", errUnexpectedSigners, quorumSize, len(signatures))
+	}
```
StephenButtolph, line 166: Is this defence-in-depth? I wouldn't expect simplex to ever call this function without sufficient signers.

StephenButtolph (commented):
I didn't review the tests yet, will look though the tests after these are addressed.

simplex/qc.go
samliok: it weird though `bls.SignatureToBytes(*bls.Signature) []byte` doesn't return a fixed length byte array. Shouldn't it always return an array of `bls.SignatureLen` 🤷

[added](https://github.com/ava-labs/avalanchego/pull/4091/commits/4a423ba9d02bcfcab3aad232ebbd682fa57e26a0)

simplex/qc.go
yacovm: Simplex checks for:

1. [Finalization](https://github.com/ava-labs/Simplex/blob/63c644082d61f641ea960cd87455aca08d6972b1/util.go#L44)
2. [Notarization](https://github.com/ava-labs/Simplex/blob/d33ae6ae82ea90e8cf0a0d15026694f5506931a4/epoch.go#L1361)
3. [Empty notarization](https://github.com/ava-labs/Simplex/blob/d33ae6ae82ea90e8cf0a0d15026694f5506931a4/epoch.go#L1292)

During the regular Simplex admission path.

Though now when I quickly skimmed through the code, I think we can process an un-verified notarization through the replication path, oops... @samliok can you confirm?

I think we should be prudent and double check this here in the avalanchego side as well.

samliok: So a couple notes after looking a bit deeper.

- We do check if multiple nodes have signed twice in simplex [see here](https://github.com/ava-labs/Simplex/blob/6e87c4b242d78ca09e048003f503ee7fcdb566a8/util.go#L57). Although it's probably a better idea to do it here in Verify(and potentially remove the check in simplex)?
-  We don't verify notarizations through the replication path. Created an [issue](https://github.com/ava-labs/Simplex/issues/188)
- Noticed we don't add re-add replication task if block verification fails. [Issue](https://github.com/ava-labs/Simplex/issues/187)
- We may possible add a signature for the wrong digest when creating a notarization. [Issue](https://github.com/ava-labs/Simplex/issues/185)


I'll tackle these issues on this simplex side, plus [added](https://github.com/ava-labs/avalanchego/pull/4091/commits/4a423ba9d02bcfcab3aad232ebbd682fa57e26a0) checking for double signers in avalanchego.

simplex/qc.go
samliok: yep, simplex checks the signers before calling aggregate

simplex/qc.go
samliok: using a bitset would be nice, but the implementation feels kind of gross. I'd need to store a canonical ordering of nodes in the `blsVerifier` to deserialize the bitset, as well as a mapping of nodeIDs to indices in the `QC` in order to serialize. 

I noticed warp has a `IndexedValidator` for creating the bitset, and some helper functions to create a canonical ordering to read the bitset. Is all this complexity needed just to reduce the size of the serializedQC? Even if we have 100 nodes thats only 2000 bytes or  ~.002 mbs potentially saved

simplex/qc.go
yacovm: But it seems to me that we call [verifyQuorumRound](https://github.com/ava-labs/Simplex/blob/main/epoch.go#L2483) whenever we process a replication response, which calls [IsFinalizationValid](https://github.com/ava-labs/Simplex/blob/main/epoch.go#L2508) and it [performs](https://github.com/ava-labs/Simplex/blob/main/util.go#L44-L77) the duplicity checks.

simplex/qc.go
StephenButtolph: It seems to me like reducing the size of every block by 4 KiB (2K for the notarization and 2K for the finalization) would be a good idea.

While we do need a canonical validator set ordering, we already do that to specify the leader rotation.

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+)
+
+var (
+	_ simplex.QuorumCertificate   = (*QC)(nil)
+	_ simplex.QCDeserializer      = (*QCDeserializer)(nil)
+	_ simplex.SignatureAggregator = (*SignatureAggregator)(nil)
+
+	// QC errors
```
StephenButtolph: We are in the `qc` file, it seems obvious to me that these are `qc` related errors.
```suggestion
```

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+
+// Verify checks if the quorum certificate is valid by verifying the aggregated signature against the signers' public keys.
+func (qc *QC) Verify(msg []byte) error {
+	pks := make([]*bls.PublicKey, 0, len(qc.signers))
+	quorum := simplex.Quorum(len(qc.verifier.nodeID2PK))
+	if len(qc.signers) != quorum {
+		return fmt.Errorf("%w: expected %d signers but got %d", errUnexpectedSigners, quorum, len(qc.signers))
+	}
+
+	uniqueSigners := make(map[ids.NodeID]struct{}, len(qc.signers))
```
StephenButtolph:
```suggestion
	uniqueSigners := set.NewSet[ids.NodeID](len(qc.signers))
```

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// ensure all signers are in the membership set
+	for _, signer := range qc.signers {
+		id, err := ids.ToNodeID(signer)
+		if err != nil {
+			return fmt.Errorf("error converting signer to NodeID: %w", err)
+		}
+
+		if _, exists := uniqueSigners[id]; exists {
```
StephenButtolph: Assuming above change
```suggestion
		if uniqueSigners.Contains(id) {
```

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+			return fmt.Errorf("%w: %x", errDuplicateSigner, signer)
+		}
+		uniqueSigners[id] = struct{}{}
+
+		pk, exists := qc.verifier.nodeID2PK[id]
+		if !exists {
+			return fmt.Errorf("%w: %x", errSignerNotFound, signer)
+		}
+
+		pks = append(pks, pk)
```
StephenButtolph: It seems weird to add to `uniqueSigners` prior to erroring, also if we use a set we should use `.Add(`
```suggestion

		pk, exists := qc.verifier.nodeID2PK[id]
		if !exists {
			return fmt.Errorf("%w: %x", errSignerNotFound, signer)
		}

		uniqueSigners.Add(id)
		pks = append(pks, pk)
```

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Signers returns the list of signers for the quorum certificate.
+func (qc *QC) Signers() []simplex.NodeID {
+	return qc.signers
+}
+
+// Verify checks if the quorum certificate is valid by verifying the aggregated signature against the signers' public keys.
+func (qc *QC) Verify(msg []byte) error {
+	pks := make([]*bls.PublicKey, 0, len(qc.signers))
```
StephenButtolph: nit: can we move this closer to where we use it? (just prior to the `qc.signers` loop)

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	qc := QC{
+		sig:     sig,
+		signers: canotoQC.Signers,
+	}
+
+	qc.verifier = d.verifier
+
+	return &qc, nil
```
StephenButtolph:
```suggestion
	return &QC{
		verifier: d.verifier,
		sig:      sig,
		signers:  canotoQC.Signers,
	}, nil
```

simplex/qc.go
```diff
@@ -0,0 +1,181 @@
...
+
+		if _, exists := a.verifier.nodeID2PK[id]; !exists {
+			return nil, fmt.Errorf("%w: %x", errSignerNotFound, signer)
+		}
+		signers = append(signers, signer)
+		sig, err := bls.SignatureFromBytes(signature.Value)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", errFailedToParseSignature, err)
+		}
+		sigs = append(sigs, sig)
```
StephenButtolph: Being super pedantic, but it seems slightly more clear if we append together at the end.
```suggestion
		sig, err := bls.SignatureFromBytes(signature.Value)
		if err != nil {
			return nil, fmt.Errorf("%w: %w", errFailedToParseSignature, err)
		}
		
		signers = append(signers, signer)
		sigs = append(sigs, sig)
```

simplex/qc_test.go
```diff
@@ -0,0 +1,261 @@
...
+	configs := newNetworkConfigs(t, 2)
+	quorum := simplex.Quorum(len(configs))
+
+	msg := []byte("Begin at the beginning, and go on till you come to the end: then stop")
+
+	signatures := make([]simplex.Signature, 0, quorum)
+	var signer BLSSigner
+	var verifier BLSVerifier
+
+	for i, config := range configs {
```
StephenButtolph: This loop is a bit odd... We allocate `signatures` to have a capacity of `2`, we add `3` signatures to it (index [0] and [1] are duplicates) and then inside of `Aggregate` we drop the third signature. While this does do what the comment states, the code doesn't seem straightforward.. Why do we add a third signature at all? (And therefore why do we have this loop at all?)

simplex/qc_test.go
```diff
@@ -0,0 +1,261 @@
...
+		{
+			name:  "too short input",
+			input: make([]byte, 10),
+			err:   errFailedToParseQC,
+		},
+		{
+			name:  "invalid signature bytes",
+			input: make([]byte, simplex.Quorum(len(verifier.nodeID2PK))*ids.NodeIDLen+bls.SignatureLen),
+			err:   errFailedToParseQC,
+		},
```
StephenButtolph: What is special about these inputs? They don't seem to align with the name of the vectors.

simplex/qc_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// Aggregate should only use the first 3 signatures
+	signatureAggregator := SignatureAggregator{verifier: &verifier}
+	qc, err := signatureAggregator.Aggregate(signatures)
+	require.NoError(t, err)
+
+	// Should only have 3 signers, not 4
+	require.Len(t, qc.Signers(), simplex.Quorum(len(configs)))
+	require.NoError(t, qc.Verify(msg))
+}
```
StephenButtolph: Would it make sense to have a single test vector against `Aggregate`? tbh I feel like almost all of the tests in this file could be combined into a single test vector (and it would probably reduce a lot of the repetitive setup).

simplex/bls.go
```diff
@@ -116,8 +120,18 @@ func createVerifier(config *Config) BLSVerifier {
...
 	}
 
+	nodeIDs := make([]ids.NodeID, 0, len(config.Validators))
 	for _, node := range config.Validators {
 		verifier.nodeID2PK[node.NodeID] = node.PublicKey
+		nodeIDs = append(nodeIDs, node.NodeID)
+	}
+
+	utils.Sort(nodeIDs)
+	verifier.canonicalNodeIDs = make([]ids.NodeID, len(nodeIDs))
```
yacovm: `verifier.canonicalNodeIDs = make([]ids.NodeID, len(nodeIDs))` - seems like we don't really need to allocate a new slice, as we already have the `nodeIDs`, which is sorted and is the same.

simplex/qc.go
```diff
@@ -0,0 +1,242 @@
...
+
+	canotoData canotoData_canotoQC
+}
+
+// Signers returns the list of signers for the quorum certificate.
+func (qc *QC) Signers() []simplex.NodeID {
+	signers := make([]simplex.NodeID, len(qc.signers))
+	for i, signer := range qc.signers {
+		signers[i] = simplex.NodeID(signer[:])
+	}
```
yacovm, line 54: `simplex.NodeID` is redundant

simplex/qc.go
```diff
@@ -0,0 +1,242 @@
...
+// CanotoQC is the Canoto representation of a quorum certificate
+type canotoQC struct {
+	Sig     [bls.SignatureLen]byte `canoto:"fixed bytes,1"`
+	Signers []byte                 `canoto:"bytes,2"`
+
+	canotoData canotoData_canotoQC
+}
+
+// Signers returns the list of signers for the quorum certificate.
+func (qc *QC) Signers() []simplex.NodeID {
```
yacovm, line 50: why are we returning a copy of the signers and not just `qc.signers` ?

simplex/qc.go
```diff
@@ -0,0 +1,242 @@
...
+	if len(signatures) < quorumSize {
+		return nil, fmt.Errorf("%w: expected %d signatures but got %d", errUnexpectedSigners, quorumSize, len(signatures))
+	}
+	signatures = signatures[:quorumSize]
+
+	signers := make([]ids.NodeID, 0, quorumSize)
+	sigs := make([]*bls.Signature, 0, quorumSize)
+	for _, signature := range signatures {
+		signer := signature.Signer
+		id, err := ids.ToNodeID(signer)
```
yacovm, line 173: why do we do this `ids.ToNodeID` ?

simplex/qc.go
```diff
@@ -0,0 +1,242 @@
...
+	return &QC{
+		verifier: a.verifier,
+		signers:  signers,
+		sig:      aggregatedSig,
+	}, nil
+}
+
+func (d *QCDeserializer) signersFromBytes(signerBytes []byte) ([]ids.NodeID, error) {
+	signerIndices := set.BitsFromBytes(signerBytes)
+	if len(signerIndices.Bytes()) != len(signerBytes) {
```
yacovm: what is the purpose of doing: `if len(signerIndices.Bytes()) != len(signerBytes) {`? why do that instead of, suppose - `bytes.Equal(signerIndices.Bytes(), signerBytes)` ?

simplex/qc.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("%w: %w", errFailedToFilterSigners, err)
+	}
+	return signers, nil
+}
+
+// filterNodeIDS returns the nodeIDs in nodeIDs whose
+// bit is set to 1 in indices.
+//
+// Returns an error if indices references an unknown node.
+func filterNodes(
```
yacovm, line 219: can we add a UT for this?

simplex/qc_test.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/ava-labs/simplex"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+)
+
+// TestQCDuplicateSigners tests verification fails if the
+// same signer signs multiple times.
+func TestQCDuplicateSigners(t *testing.T) {
+	configs := newNetworkConfigs(t, 2)
```
yacovm: can we have a more realistic setting - 4 nodes instead of 2?

simplex/qc_test.go
```diff
@@ -0,0 +1,233 @@
...
+	signatureAggregator := SignatureAggregator{verifier: &verifier}
+	qc, err := signatureAggregator.Aggregate(signatures)
+
+	require.NoError(t, err)
+	err = qc.Verify(msg)
+	require.ErrorIs(t, err, errDuplicateSigner)
+}
+
+func TestQCDeserializerInvalidBytes(t *testing.T) {
+	config := newEngineConfig(t, 2)
```
yacovm, line 47: same - can we have a more realistic setting - 4 nodes instead of 2 ?

simplex/qc.go
samliok: `qc.signers` are of type `[]ids.Node`. Unless we want to store both `simplex.NodeID` and `ids.NodeID` in the `qc`  I feel like theres no clean way to avoid converting from one to another. I chose to store `signers` as `ids.NodeID` since there would be less total conversions to do

simplex/qc.go
samliok: to ensure no panics could happen when converting from a `simplex.NodeID` to `ids.NodeID`

see https://github.com/ava-labs/avalanchego/pull/4091#discussion_r2220207400

yacovm (approved):

simplex/qc.go
```diff
@@ -0,0 +1,243 @@
...
+func (qc *QC) createSignersBitSet() []byte {
+	bitset := set.NewBits()
+	for _, signer := range qc.signers {
+		// index should always exist, since we deserialized the signers from the same verifier
+		index := qc.verifier.canonicalNodeIDIndices[signer]
+		bitset.Add(index)
+	}
+
+	signerBytes := bitset.Bytes()
+	return signerBytes
```
StephenButtolph:
```suggestion
	return bitset.Bytes()
```

StephenButtolph (approved):
trivial nit and lgtm

PR #4095, opened 2025-07-16
Simplify test block creation

StephenButtolph:
## Why this should be merged

Reduces the number of options required to provide to the `newBlock` test helper.

### DIFF
--- a/simplex/block_test.go
+++ b/simplex/block_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ava-labs/simplex"
 	"github.com/stretchr/testify/require"
 
+	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman/snowmantest"
 	"github.com/ava-labs/avalanchego/snow/engine/enginetest"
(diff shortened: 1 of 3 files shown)


simplex/block_test.go
```diff
@@ -157,139 +163,87 @@ func TestVerify(t *testing.T) {
 	require.Equal(t, blockBytes, vBlockBytes, "block bytes should match after verification")
 }
 
-// TestVerifyPrevIsLatest tests that a block with a prev digest that is not found in the block tracker
-// successfully verifies the block if it is the latest accepted block.
```
StephenButtolph, line 161: This test was no longer relevant. It seems like this test was written for a previous version of the code.

simplex/util_test.go
```diff
@@ -10,72 +10,53 @@ import (
...
+				Seq:     0,
+			},
 		}
+		bytes, err := block.Bytes()
+		require.NoError(t, err)
+
+		digest := computeDigest(bytes)
+		block.digest = digest
+
+		block.blockTracker = newBlockTracker(block)
```
samliok, line 43: it feels weird to set the `blockTracker` with the block we are returning. This means `newBlock.index()` would succeed. 

```go
genesis := newBlock(t, newBlockConfig{})
err := genesis.blockTracker.indexBlock(ctx, genesis.digest)
require.NoError(t, err) // this should error
```

StephenButtolph: Talked offline and decided that calling `indexBlock` with a previously accepted hash is unexpected (and therefore undefined behavior).

PR #4097, opened 2025-07-17
Throttle platformVM block building until normal ops

yacovm:
## Why this should be merged

When the platformVM is bootstrapping, WaitForEvent() is called on it by the notification forwarder, which forwards the events to the handler, which in turn routes them to the appropriate engine according to the state.

The platformVM cannot know when it should propose a block until it has finished bootstrapping, and as a result it will fail in attempting.

If the proposerVM selects the platformVM for block building before it has finished bootstrapping, it will result in the notifier in a spinning loop until it has finished bootstrapping.

## How this works

This fix makes the platformVM throttle its WaitForEvent() call until it has finished bootstrapping.

## How this was tested

unit test

No

### DIFF
--- a/vms/platformvm/vm_test.go
+++ b/vms/platformvm/vm_test.go
@@ -2252,6 +2252,7 @@ func TestThrottleBlockBuildingUntilNormalOperationsStart(t *testing.T) {
 
 	impatientContext, cancel = context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
+
 	msg, err = vm.WaitForEvent(impatientContext)
 	require.NoError(err)
 	require.Equal(common.PendingTxs, msg)


vms/platformvm/block/builder/builder.go
```diff
@@ -105,6 +113,8 @@ func (b *builder) WaitForEvent(ctx context.Context) (common.Message, error) {
 			return common.PendingTxs, nil
 		}
 
+		fmt.Println(duration)
```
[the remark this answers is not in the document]
yacovm: oops, what would I do without you, copilot?

vms/platformvm/block/builder/builder.go
```diff
@@ -92,6 +94,12 @@ func (b *builder) WaitForEvent(ctx context.Context) (common.Message, error) {
 			return 0, err
 		}
 
+		select {
```
StephenButtolph: Would it be simpler to just put:
```golang
// We shouldn't call durationToSleep until the chain is marked as no longer
// bootstrapping.
if !b.txExecutorBackend.Bootstrapped.Get() {
	<-ctx.Done()
}
```
at the top of `WaitForEvent` here rather than introducing a new channel?

alarso16: That seems sufficient (and I checked, it does still pass the UT)

yacovm: Done

vms/platformvm/block/builder/builder.go
```diff
@@ -92,6 +92,13 @@ func (b *builder) WaitForEvent(ctx context.Context) (common.Message, error) {
 			return 0, err
 		}
 
+		// We shouldn't call durationToSleep until the chain is marked as no longer
+		// bootstrapping.
+		if !b.txExecutorBackend.Bootstrapped.Get() {
+			<-ctx.Done()
+			return 0, ctx.Err()
+		}
```
StephenButtolph: Is there a reason this is inside the `for` loop? It seems like we only want/need to check this once.

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+		RewardConfig:           defaultRewardConfig,
+		UpgradeConfig:          upgradetest.GetConfigWithUpgradeTime(upgradetest.Latest, latestForkTime),
+	}}
+
+	defer func() {
+		vm.ctx.Lock.Lock()
+		defer vm.ctx.Lock.Unlock()
+
+		require.NoError(vm.Shutdown(context.Background()))
+	}()
```
StephenButtolph: nit: can we move the deferring of shutdown after `Initialize`? It looks a bit misplaced here imo.

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+
+	vm.state.SetTimestamp(vm.clock.Time())
+	vm.state.SetFeeState(gas.State{
+		Capacity: defaultDynamicFeeConfig.MaxCapacity,
+	})
+
+	newTime := latestForkTime.Add(genesistest.DefaultValidatorDuration)
+
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+	require.NoError(vm.onBootstrapStarted())
```
StephenButtolph: I think `vm.SetState` handles the call to `onBootstrapStarted`.
```suggestion
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+
+	impatientContext, cancel := context.WithTimeout(context.Background(), time.Millisecond*100)
+
+	msg, err := vm.WaitForEvent(impatientContext)
+	cancel()
+	require.ErrorIs(context.DeadlineExceeded, err)
+	require.Equal(msg, common.Message(0))
+
+	ctx.Lock.Lock()
+	require.NoError(vm.onNormalOperationsStarted())
```
StephenButtolph:
```suggestion
	require.NoError(vm.SetState(context.Background(), snow.NormalOp))
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+	atomicDB := prefixdb.New([]byte{1}, db)
+
+	vm.clock.Set(latestForkTime)
+	ctx := snowtest.Context(t, snowtest.PChainID)
+
+	m := atomic.NewMemory(atomicDB)
+	msm := &mutableSharedMemory{
+		SharedMemory: m.NewSharedMemory(ctx.ChainID),
+	}
+	ctx.SharedMemory = msm
```
StephenButtolph: We aren't issuing any atomic txs in this test, so we don't need to override the shared memory module. The default shared memory implementation provided by `snowtest.Context` should be sufficient
```suggestion
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+	appSender := &enginetest.Sender{}
+	appSender.CantSendAppGossip = true
+	appSender.SendAppGossipF = func(context.Context, common.SendConfig, []byte) error {
+		return nil
+	}
+	appSender.SendAppErrorF = func(context.Context, ids.NodeID, uint32, int32, string) error {
+		return nil
+	}
+
+	dynamicConfigBytes := []byte(`{"network":{"max-validator-set-staleness":0}}`)
```
StephenButtolph: I don't think this test needs to modify the network config (we can just pass in `nil` config bytes)
```suggestion
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+	vm.state.SetFeeState(gas.State{
+		Capacity: defaultDynamicFeeConfig.MaxCapacity,
+	})
+
+	newTime := latestForkTime.Add(genesistest.DefaultValidatorDuration)
+
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+	require.NoError(vm.onBootstrapStarted())
+
+	vm.clock.Set(newTime)
```
StephenButtolph, line 2238: Can we put a comment here explaining why this is needed?
```suggestion
	// Advance the time so that the block builder would be willing to remove the
	// genesis validators.
	vm.clock.Set(newTime)
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+		nil,
+		appSender,
+	))
+
+	vm.state.SetTimestamp(vm.clock.Time())
+	vm.state.SetFeeState(gas.State{
+		Capacity: defaultDynamicFeeConfig.MaxCapacity,
+	})
+
+	newTime := latestForkTime.Add(genesistest.DefaultValidatorDuration)
```
StephenButtolph: trivial nit, but can we move this calculation to right before it is used? (after the `SetState` call).

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+		nil,
+		dynamicConfigBytes,
+		nil,
+		appSender,
+	))
+
+	vm.state.SetTimestamp(vm.clock.Time())
+	vm.state.SetFeeState(gas.State{
+		Capacity: defaultDynamicFeeConfig.MaxCapacity,
+	})
```
StephenButtolph: Why are we manually modifying the state here?
```suggestion
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+
+	vm.clock.Set(newTime)
+
+	ctx.Lock.Unlock()
+
+	impatientContext, cancel := context.WithTimeout(context.Background(), time.Millisecond*100)
+
+	msg, err := vm.WaitForEvent(impatientContext)
+	cancel()
+	require.ErrorIs(context.DeadlineExceeded, err)
```
StephenButtolph:
```suggestion
	require.ErrorIs(err, context.DeadlineExceeded)
```

vms/platformvm/vm_test.go
```diff
@@ -2180,3 +2180,98 @@ func TestPruneMempool(t *testing.T) {
...
+	require.Equal(msg, common.Message(0))
```
StephenButtolph:
```suggestion
	require.Zero(msg)
```

StephenButtolph (approved):

PR #4106, opened 2025-07-22
ci: remove load 1.0

RodrigoVillar:
## Why this should be merged

With #4071 merged in, the features of the original load testing framework are now supported by load 2.0, which means that we can now deprecate load 1.0.

This PR will be followed by additional PRs to remove load 1.0 altogether and to add documentation for load 2.0.

## How this works

Removes load 1.0 tests from CI + Taskfiles; if an existing load 2.0 counterpart job didn't exist, the load 1.0 job was rewritten for load 2.0.

## How this was tested

CI

N/A

### DIFF
--- a/.github/workflows/c-chain-reexecution-benchmark.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark.yml
@@ -20,7 +20,7 @@ on:
         required: false
         default: s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100.zip
   schedule:
-    - cron: '0 6 * * 0' # Runs every Sunday at 06:00 UTC
+    - cron: '0 9 * * *' # Runs every day at 09:00 UTC (04:00 EST)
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -232,46 +232,33 @@ jobs:
       - name: Run e2e tests
         shell: bash
         run: nix develop --command ./scripts/run_task.sh test-bootstrap-monitor-e2e
-  load:
-    name: Run process-based load test
+  load2:
```
maru-ava: Maybe load2 jobs are removed in favor of load jobs that target load2 internally in this PR, and when load goes away load2 replaced it? Ideally we end up with 'load' as a singular thing.

RodrigoVillar: Yep was going to remove the `load2` naming in the following PR, but doing it in this PR also makes sense; done: 4d3b579

PR #4112, opened 2025-07-22
refactor: remove load 1.0 

RodrigoVillar:
## Why this should be merged

Since there's now feature parity between load 1.0 and 2.0, there's no longer a need for load 1.0 🪓

## How this works

This PR deletes the existing `load` package and renames `load2` to just `load`.

## How this was tested

CI

N/A


RodrigoVillar:
> Why is the 'c' path being removed? Not critical, just curious as to your thinking. At some point there will need to be load testing for the other primary chains.

Since we only have one use case for load testing (C-Chain), I didn't think it was necessary to add a `c` directory. I agree though that we'll eventually be load testing more chains and creating separate paths for each chain will make sense then.

PR #4118, opened 2025-07-22
Added Juan wherever Maru - expect discussion

JuanLeon2:
## Why this should be merged

Maru wants to take a step back from responsibility.  Juan would *very carefully* look at PRs for which currently Maru is a required reviewer (other then Stephen).    Half of the argument is that we need someone to look at those PRs.  The other half is that other than action workflows, and those are debatable, there is little damage these subtrees can do directly to AvalancheGo. 

This is temporary, until someone steps up, or is hired, and comes to be trusted (this does not imply Juan is trusted)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,21 +8,21 @@
 * @StephenButtolph
 *.md @meaghanfitzgerald @StephenButtolph
 /.dockerignore @maru-ava
-/.envrc @maru-ava
-/.github/ @maru-ava
-/.github/*.md @maru-ava @meaghanfitzgerald
+/.envrc @maru-ava @JuanLeon2
+/.github/ @maru-ava @JuanLeon2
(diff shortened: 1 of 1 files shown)


JuanLeon2 (commented):
> ## Pull Request Overview
> 
> This pull request adds @joshua-kim as a code owner alongside @maru-ava for various configuration and infrastructure-related files and directories. The change is intended as a temporary measure to provide review coverage while Maru steps back from some responsibilities.
> 
> - Adds @JuanLeon2 as an additional reviewer for infrastructure and configuration files
> - Maintains existing code ownership structure while expanding review capacity
> - Focuses on lower-risk areas like scripts, tests, and configuration files
> 
> 
> 
>

joshua-kim (approved):

StephenButtolph (approved):

PR #4122, opened 2025-07-25
Storage Component For Simplex

samliok:
## Why this should be merged

This PR introduces the storage component for Simplex. The three functions of the `simplex.Storage` interface are 

- `Index` which indexes a block and finalization into the storage. Because indexing a block is already handled when a block is accepted, all this component needs to do is index the finalization. 
- `Retrieve` grabs the block and finalization from storage
- `Height` tells simplex how many blocks have been indexed. The next sequence to commit is always equal to the height.


simplex/storage.go
```diff
@@ -0,0 +1,236 @@
...
+	Round:   0,
+	Seq:     0,
+}
+
+type Storage struct {
+	// height represents the number of blocks indexed in storage.
+	height atomic.Uint64
+
+	// db is the underlying database used to store finalizations.
+	db database.Database
```
yacovm: Is it possible that we just define an interface of what we need and use it, and then inject the interface? I also don't see why we need to do the `prefixdb.New(simplexPrefix, config.DB)` in this file. Can't we do it at a higher level where we instantiate the storage?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+// Retrieve returns the block and finalization at [seq].
+// If [seq] is not found, returns false.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, bool) {
+	if seq == 0 {
+		return s.genesisBlock, simplex.Finalization{}, true
+	}
+
+	block, err := getBlockAtHeight(context.TODO(), s.vm, seq)
+	if err != nil {
+		if err == database.ErrNotFound {
```
yacovm, line 115: aren't we returning here the same result no matter what err is?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+var (
+	_               simplex.Storage = (*Storage)(nil)
+	genesisMetadata                 = simplex.ProtocolMetadata{
+		Version: 1,
```
yacovm: why version 1? I don't think I've seen we use version 1 in the Simplex library anywhere, but I could be wrong.

Shouldn't this be 0?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+// If [seq] is not found, returns false.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, bool) {
+	if seq == 0 {
+		return s.genesisBlock, simplex.Finalization{}, true
+	}
+
+	block, err := getBlockAtHeight(context.TODO(), s.vm, seq)
+	if err != nil {
+		if err == database.ErrNotFound {
+			return nil, simplex.Finalization{}, false
```
yacovm: can we log an error in places where the error is not ErrNotFound?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+// retrieveFinalization retrieves the finalization at [seq].
+// If the finalization is not found, it returns false.
+func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, bool, error) {
+	seqBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBuff, seq)
+	finalizationBytes, err := s.db.Get(seqBuff)
+	if err != nil {
+		if err == database.ErrNotFound {
+			return simplex.Finalization{}, false, nil
+		}
```
yacovm, line 221: can we log an error in places where the error is not ErrNotFound?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+	buff := make([]byte, len(blockHeaderBytes)+len(qcBytes))
+	copy(buff, blockHeaderBytes)
+	copy(buff[len(blockHeaderBytes):], qcBytes)
+	return buff
+}
+
+// finalizationFromBytes deserialized the bytes into a simplex.Finalization.
+func finalizationFromBytes(bytes []byte, d *QCDeserializer) (simplex.Finalization, error) {
+	var finalization simplex.Finalization
+	if err := finalization.Finalization.FromBytes(bytes[:simplex.BlockHeaderLen]); err != nil {
```
yacovm: This `finalization.Finalization.FromBytes(bytes[:simplex.BlockHeaderLen]` technique implicitly assumes that the finalization size will never change throughout the versions. 

If we ever need to change the finalization size, this will break backward compatibility.
Can we use a more flexible encoding method such as canoto or proto to encode the `bytes`?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+		s.log.Warn("attempted to index genesis block, which should not be indexed")
+		return errGenesisIndexed
+	}
+
+	currentHeight := s.height.Load()
+	if currentHeight != bh.Seq {
+		return fmt.Errorf("%w: expected %d, got %d", errUnexpectedSeq, currentHeight, bh.Seq)
+	}
+
+	if !bytes.Equal(bh.Digest[:], finalization.Finalization.Digest[:]) {
```
yacovm: That's a nice sanity check.

I think another very important sanity check we should do is to ensure that the previous block of the block header points to the block hash we have committed. 

Can we add such a check?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+		return fmt.Errorf("failed to index block: %w", err)
+	}
+
+	s.height.Add(1) // only increment height after successful indexing
+	return nil
+}
+
+// getGenesisBlock returns the genesis block wrapped as a Block instance.
+func getGenesisBlock(ctx context.Context, config *Config) (*Block, error) {
+	genesis := &Block{
```
yacovm: This comment is unrelated to this line, but, I noticed that we only populate the block tracker in the test. Yet we use this field in production code.... Isn't its initialization missing somewhere?

simplex/storage.go
```diff
@@ -0,0 +1,239 @@
...
+	errInvalidQC        = errors.New("invalid quorum certificate")
+	errMismatchedDigest = errors.New("mismatched digest in finalization")
+)
+
+type Storage struct {
+	// height represents the number of blocks indexed in storage.
+	height atomic.Uint64
+
+	// db is the underlying database used to store finalizations.
+	db database.KeyValueReaderWriterDeleter
```
yacovm: can't we just use an interface that defines the methods we need - Put and Get?

simplex/storage.go
samliok: i think you are right. we should probably inject the block tracker in the `blockDeserializer`and pass it to the block in `DeserializeBlock`.  I added it here and in `Retrieve` as well so we would error rather than panic if a `verify` or `index` was called on the block.

I'll add the `blockDeserializer` changes in another PR

simplex/storage.go
samliok: sounds good, i added canoto. I think im going to follow this PR by moving all the canoto structs into their own file to reduce the simplex file bloat.

simplex/storage.go
samliok: are you saying check `bh.Prev == prevCommited.digest`? Is calling `s.blockTracker.getBlockByDigest(bh.Prev)` sufficient or are you suggesting to grab the previous block from storage?

yacovm: > are you saying check bh.Prev == prevCommited.digest?

yeah exactly.

> Is calling s.blockTracker.getBlockByDigest(bh.Prev) sufficient or are you suggesting to grab the previous block from storage?

I wasn't suggesting a specific implementation, but I think what's best is to simply hold a reference to the last committed block hash and just update it upon each commit.

yacovm (approved):

simplex/config.go
```diff
@@ -24,6 +30,9 @@ type Config struct {
 	// across all nodes in the subnet.
 	Validators map[ids.NodeID]*validators.GetValidatorOutput
 
+	VM block.ChainVM
+
+	DB KeyValueReaderWriter
```
StephenButtolph: Is there a reason we make a new interface here?
```suggestion
	DB database.KeyValueReaderWriter
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+
+	errUnexpectedSeq    = errors.New("unexpected sequence number")
+	errGenesisIndexed   = errors.New("genesis block should not be indexed")
+	errInvalidQC        = errors.New("invalid quorum certificate")
+	errMismatchedDigest = errors.New("mismatched digest in finalization")
+)
+
+type Storage struct {
+	// height represents the number of blocks indexed in storage.
+	height atomic.Uint64
```
StephenButtolph, line 40: This is a pretty confusing name. `height` being equal to the `lastAcceptedHeight+1`. Could we either:
- align the two (have it equal `lastAcceptedHeight` and perhaps name it that)
- change the name of this variable to better convey what it is?

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, bool) {
+	if seq == 0 {
+		return s.genesisBlock, simplex.Finalization{}, true
+	}
+
+	block, err := getBlockAtHeight(context.TODO(), s.vm, seq)
+	if err != nil {
+		if err == database.ErrNotFound {
+			s.log.Error("block not found for sequence", zap.Uint64("seq", seq), zap.Error(err))
+		}
```
StephenButtolph: Why do we log specifically for `ErrNotFound`?

I'm asking mainly because `ErrNotFound` is traditionally a fairly "expected" error, but any other error is almost always unexpected.

It seems to me like if we are logging an error for `ErrNotFound` we should probably be logging for all errors.

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	return s, nil
+}
+
+func (s *Storage) Height() uint64 {
+	return s.height.Load()
+}
+
+// Retrieve returns the block and finalization at [seq].
+// If [seq] is not found, returns false.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, bool) {
```
StephenButtolph: I feel like this should return an error. Not being able to find a block because the process is out of file descriptors is pretty different than the block not being on disk.

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+
+	return vb, finalization, true
+}
+
+// Index indexes the finalization in the storage.
+// It stores the finalization bytes at the current height and increments the height.
+func (s *Storage) Index(ctx context.Context, block simplex.VerifiedBlock, finalization simplex.Finalization) error {
+	bh := block.BlockHeader()
+	if bh.Seq == 0 {
+		s.log.Warn("attempted to index genesis block, which should not be indexed")
```
StephenButtolph: Since we are logging it as a warn/error, it is implicit that it shouldn't be done.
```suggestion
		s.log.Error("attempted to index genesis block")
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+
+	currentHeight := s.height.Load()
+	if currentHeight != bh.Seq {
+		return fmt.Errorf("%w: expected %d, got %d", errUnexpectedSeq, currentHeight, bh.Seq)
+	}
+
+	if s.blockTracker.lastIndexed != bh.Prev {
+		s.log.Warn("attempted to index block with mismatched previous digest",
+			zap.String("expected", s.blockTracker.lastIndexed.String()),
+			zap.String("got", bh.Prev.String()))
```
StephenButtolph:
```suggestion
		s.log.Error("attempted to index block with mismatched previous digest",
			zap.Stringer("expected", s.blockTracker.lastIndexed),
			zap.Stringer("got", bh.Prev),
		)
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+		blockTracker: blockTracker,
+	}
+
+	snowmanGenesis, err := getBlockAtHeight(ctx, config.VM, 0)
+	if err != nil {
+		return nil, err
+	}
+	genesis.vmBlock = snowmanGenesis
+
+	// set the digest
```
StephenButtolph: Prefer full initialization over partial initialization as much as possible.
```suggestion
	snowmanGenesis, err := getBlockAtHeight(ctx, config.VM, 0)
	if err != nil {
		return nil, err
	}

	genesis := &Block{
		vmBlock:      snowmanGenesis,
		metadata:     genesisMetadata,
		blockTracker: blockTracker,
	}
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	block, err := vm.GetBlock(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	return block, nil
```
StephenButtolph:
```suggestion
	return vm.GetBlock(ctx, id)
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+
+type canotoFinalization struct {
+	Finalization []byte `canoto:"bytes,1"`
+	QC           []byte `canoto:"bytes,2"`
+
+	canotoData canotoData_canotoFinalization
+}
+
+// finalizationFromBytes deserialized the bytes into a simplex.Finalization.
+func (c *canotoFinalization) finalization(d *QCDeserializer) (simplex.Finalization, error) {
```
StephenButtolph: I feel like `canotoFinalization.toFinalization(s.deserializer)` reads better than `canotoFinalization.finalization(s.deserializer)`
```suggestion
func (c *canotoFinalization) toFinalization(d *QCDeserializer) (simplex.Finalization, error) {
```

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+		s.log.Warn("attempted to index genesis block, which should not be indexed")
+		return errGenesisIndexed
+	}
+
+	currentHeight := s.height.Load()
+	if currentHeight != bh.Seq {
+		return fmt.Errorf("%w: expected %d, got %d", errUnexpectedSeq, currentHeight, bh.Seq)
+	}
+
+	if s.blockTracker.lastIndexed != bh.Prev {
```
StephenButtolph: During review I initially found it odd that we reference `lastIndexed` without the lock held, but I assume that the expectation is that only this thread modifies `lastIndexed`, so it is safe to read this without the lock held.

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+		return fmt.Errorf("%w: expected %s, got %s", errMismatchedPrevDigest, s.blockTracker.lastIndexed, bh.Prev)
+	}
+
+	if !bytes.Equal(bh.Digest[:], finalization.Finalization.Digest[:]) {
+		return fmt.Errorf("%w: expected %d, got %d", errMismatchedDigest, bh.Digest, finalization.Finalization.Digest)
+	}
+
+	if finalization.QC == nil {
+		return errInvalidQC
+	}
```
StephenButtolph: Would it be possible to make it more clear as to what checks are sanity checks vs required checks?

Perhaps we always `log.Error` on sanity checks?

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+		return nil, err
+	}
+	genesis.digest = computeDigest(bytes)
+
+	return genesis, nil
+}
+
+// retrieveFinalization retrieves the finalization at [seq].
+// If the finalization is not found, it returns false.
+func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, bool, error) {
```
StephenButtolph: The `bool` is never really used from this function. It is only handled as `if !found || err != nil {`, so I think we should remove the `bool` and just report the `error`.
```suggestion
func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, bool, error) {
```

simplex/config.go
samliok: yacov wanted me to make a new one with just the `Put` and `Get` methods. I'd rather just use the existing interface  like you suggested even if it is more general than we need

https://github.com/ava-labs/avalanchego/pull/4122#discussion_r2240904076

simplex/storage.go
yacovm: It's simply the height of the chain. It's fairly a common term, we use it extensively in Simplex and it's always the next sequence to commit (e.g if the genesis block is sequence 0, then the first block to commit will have the index of 1, as it is committed when the chain is at height of 1). 

It's not equal to the last accepted height + 1, but equal to the last accepted sequence + 1.

simplex/storage.go
samliok: updated the simplex api & the pr

simplex/storage.go
samliok: yea we should not be indexing concurrently, but i can add a lock as a sanity check

simplex/storage.go
StephenButtolph: > It's not equal to the last accepted height + 1, but equal to the last accepted sequence + 1.

Well - it is currently set to the last accepted height + 1 not the last accepted sequence + 1.
```golang
s.height.Store(lastAcceptedBlock.Height() + 1)
```
no?

samliok: i thought block heights should always match to sequences

simplex/storage.go
StephenButtolph: I don't think we should add this locking. I think a comment would make more sense.

yacovm: The only not concurrent thing in Simplex is block building and block verification, but that needs to be synchronized in the avalanchego side since the VM is not thread safe.

simplex/storage.go
```diff
@@ -0,0 +1,275 @@
...
+
+// retrieveFinalization retrieves the finalization at [seq].
+// If the finalization is not found, it returns false.
+func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, error) {
+	seqBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBuff, seq)
+	finalizationBytes, err := s.db.Get(seqBuff)
+	if err != nil {
+		if err == database.ErrNotFound {
+			return simplex.Finalization{}, simplex.ErrBlockNotFound
```
StephenButtolph, line 220: Does simplex have special handling for `ErrBlockNotFound`?

I don't have a problem with returning this sentinel error, just mainly interested on if this is something simplex actually uses.

simplex/storage.go
```diff
@@ -0,0 +1,275 @@
...
+	if err := canotoFinalization.UnmarshalCanoto(finalizationBytes); err != nil {
+		return simplex.Finalization{}, err
+	}
+
+	finalization, err := canotoFinalization.toFinalization(s.deserializer)
+	if err != nil {
+		return simplex.Finalization{}, err
+	}
+
+	return finalization, nil
```
StephenButtolph:
```suggestion
	return canotoFinalization.toFinalization(s.deserializer)
```

simplex/storage.go
yacovm: > Well - it is currently set to the last accepted height + 1 not the last accepted sequence + 1.

If we only have the genesis block, then the height of the chain should be 1, no?

simplex/storage.go
yacovm:
```
Does simplex have special handling for ErrBlockNotFound?
```

It has special handling for anything but that - we [put](https://github.com/ava-labs/Simplex/blob/main/epoch.go#L100) [the](https://github.com/ava-labs/Simplex/blob/main/epoch.go#L2747) epoch in a state of "I am halted". In the future we can cleverly shut down the chain.

simplex/util_test.go
```diff
@@ -48,9 +51,12 @@ func newBlock(t *testing.T, config newBlockConfig) *Block {
 		config.round = config.prev.metadata.Round + 1
 	}
 
-	vmBlock := snowmantest.BuildChild(config.prev.vmBlock.(*snowmantest.Block))
+	vmBlock := snowmantest.BuildChild(config.prev.vmBlock.(*wrappedBlock).Block)
 	block := &Block{
-		vmBlock:      vmBlock,
+		vmBlock: &wrappedBlock{
+			Block: vmBlock,
+			vm:    config.prev.vmBlock.(*wrappedBlock).vm,
```
yacovm, line 58: would it make sense to put `vm` into `config` so this line will be less verbose?

simplex/storage.go
StephenButtolph: Perhaps to just make it very clear, we can call this variable `nextHeight`?
```suggestion
	// nextHeight is the height of the next block to index.
	nextHeight atomic.Uint64
```
Or we could name it `blocksIndexed`? I think `length` could also make sense.

From everything I have seen, the `height` of a blockchain is equal to the monotonically increasing number included in the last block. (For ethereum they call this the block number, in avalanche we call it the block height).

So, the height (from what I can see) is defined as the number of blocks _since genesis_. The height of a chain with only the genesis block would be 0.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (s *Storage) Height() uint64 {
+	return s.height.Load()
+}
+
+// Retrieve returns the block and finalization at [seq].
+// If [seq] is not found, returns simplex.ErrBlockNotFound.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, error) {
+	if seq == 0 {
```
StephenButtolph, line 109:
```suggestion
	// The genesis block doesn't have a finalization, so we need to handle it
	// specifically.
	if seq == 0 {
```

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// Index indexes the finalization in the storage.
+// It stores the finalization bytes at the current height and increments the height.
+func (s *Storage) Index(ctx context.Context, block simplex.VerifiedBlock, finalization simplex.Finalization) error {
+	bh := block.BlockHeader()
+	if bh.Seq == 0 {
+		s.log.Error("attempted to index genesis block")
+		return errGenesisIndexed
+	}
```
StephenButtolph: Since we always initialize `s.height` to be `lastAcceptedHeight + 1`, this check is a subset of the immediately following check. I feel like we probably don't need to include this check specifically and we should instead just have the following check.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	// no need to lock the blockTracker, since Index should not be called concurrently
+	if s.blockTracker.lastIndexed != bh.Prev {
+		s.log.Error("attempted to index block with mismatched previous digest",
+			zap.Stringer("expected", s.blockTracker.lastIndexed),
+			zap.Stringer("got", bh.Prev))
+
+		return fmt.Errorf("%w: expected %s, got %s", errMismatchedPrevDigest, s.blockTracker.lastIndexed, bh.Prev)
+	}
+
+	if !bytes.Equal(bh.Digest[:], finalization.Finalization.Digest[:]) {
```
StephenButtolph:
```suggestion
	if bh.Digest != finalization.Finalization.Digest {
```

simplex/storage.go
yacovm: To me, `blocksIndexed` is the definition of `height`. 

> From everything I have seen, the height of a blockchain is equal to the monotonically increasing number included in the last block. (For ethereum they call this the block number, in avalanche we call it the block height).

There is the problem - the number in the block cannot be the height, but should be the number (what in Simplex we call the sequence).

In my opinion, a height is something that is defined outside of the block, based on the blocks in the ledger:

https://www.investopedia.com/terms/b/block-height.asp

> The block height of a particular [block](https://www.investopedia.com/terms/b/block.asp) is defined as the number of blocks preceding it in the [blockchain](https://www.investopedia.com/terms/b/blockchain.asp).


> So, the height (from what I can see) is defined as the number of blocks since genesis. The height of a chain with only the genesis block would be 0.

I think we should treat the genesis block as a regular block when counting the number of blocks, and therefore a chain with only the genesis block, is of height 1.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+// If [seq] is not found, returns simplex.ErrBlockNotFound.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, error) {
+	if seq == 0 {
+		return s.genesisBlock, simplex.Finalization{}, nil
+	}
+
+	block, err := getBlockAtHeight(context.TODO(), s.vm, seq)
+	if err != nil {
+		if err == database.ErrNotFound {
+			s.log.Error("block not found for sequence", zap.Uint64("seq", seq), zap.Error(err))
```
StephenButtolph: Just wanted to confirm that the simplex code prevents this code from getting hit (otherwise we'll get notifications/pages when someone hits it)

yacovm: It does not, and we should totally remove this.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// Retrieve returns the block and finalization at [seq].
+// If [seq] is not found, returns simplex.ErrBlockNotFound.
+func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalization, error) {
+	if seq == 0 {
+		return s.genesisBlock, simplex.Finalization{}, nil
+	}
+
+	block, err := getBlockAtHeight(context.TODO(), s.vm, seq)
```
yacovm: can we call it `getBlock` instead? Since we're retrieving the sequence, not the height

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+// If the finalization is not found, it returns false.
+func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, error) {
+	seqBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBuff, seq)
+	finalizationBytes, err := s.db.Get(seqBuff)
+	if err != nil {
+		if err == database.ErrNotFound {
+			return simplex.Finalization{}, simplex.ErrBlockNotFound
+		}
+		s.log.Error("failed to retrieve finalization", zap.Uint64("seq", seq), zap.Error(err))
```
yacovm: So I would say we should log it with debug, not Error. I think in general we shouldn't log these things with error because this struct doesn't have enough context to know when an error should be logged.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+
+	finalization, err := s.retrieveFinalization(seq)
+	if err != nil {
+		return nil, simplex.Finalization{}, err
+	}
+
+	vb := &Block{vmBlock: block, metadata: finalization.Finalization.ProtocolMetadata, blockTracker: s.blockTracker}
+	bytes, err := vb.Bytes()
+	if err != nil {
+		s.log.Error("failed to serialize block", zap.Error(err))
```
yacovm: I think here we can log an error though!

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return vb, finalization, nil
+}
+
+// Index indexes the finalization in the storage.
+// It stores the finalization bytes at the current height and increments the height.
+func (s *Storage) Index(ctx context.Context, block simplex.VerifiedBlock, finalization simplex.Finalization) error {
+	bh := block.BlockHeader()
+	if bh.Seq == 0 {
+		s.log.Error("attempted to index genesis block")
```
yacovm: A bit of OCD from my side but if that's not too hard -  can we please start each error log with a capital letter? Here and in 140, 148, 156 163.

simplex/storage.go
yacovm: I don't follow, can you explain? The next check is a digest check. This one is a sequence one.

simplex/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	seqBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBuff, currentHeight)
+
+	finalizationBytes := finalizationToBytes(finalization)
+	if err := s.db.Put(seqBuff, finalizationBytes); err != nil {
+		return fmt.Errorf("failed to store finalization: %w", err)
+	}
+
```
yacovm, line 174: If we crash in between, is there any recovery logic needed? Do we need to roll-forward the VM and call indexBlock?

```
finalizationBytes := finalizationToBytes(finalization)
	if err := s.db.Put(seqBuff, finalizationBytes); err != nil {
		return fmt.Errorf("failed to store finalization: %w", err)
	}
    <-------------------------
	err := s.blockTracker.indexBlock(ctx, bh.Digest)
	if err != nil {
		return fmt.Errorf("failed to index block: %w", err)
	}
```

simplex/storage.go
samliok: but we are calling the `vm.GetBlockAtHeight` with the sequence passed in?

simplex/util_test.go
samliok: i think that would cause more headache, since the tester would need to remember to pass the vm into each  block `config`.  Handling it in the test helper feels like better ux, even though this may be a verbose line

simplex/storage.go
samliok: basically if we try to index `seq` 0, we will always catch that error on the following check since current height is always greater than 0.

simplex/storage.go
yacovm: but the parameter we're inserting is `seq` and I see there is disagreement on the semantics of height so the function name I am proposing is more neutral 😁

simplex/storage.go
samliok: so if we put seq `x` in the database and don't call `accept`, then when we startup we will set the height to the last accepted block(which is x-1). The next time we index, we will just overwrite the finalization `x` with a new `x'` 

However, i think we need to have a sanity check in retrieve to ensure we do not try and retrieve a block >= height

simplex/storage.go
samliok: I am fine with either definition, I will keep it as `height` for now

simplex/storage.go
yacovm: >  then when we startup we will set the height to the last accepted block(which is x-1)

The reason is that we rely on the VM to find out what is the latest block and not on the finalization DB, right?

> However, i think we need to have a sanity check in retrieve to ensure we do not try and retrieve a block >= height

Why? won't we just return `database.ErrNotFound` ?

simplex/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	bh := block.BlockHeader()
+	currentHeight := s.height.Load()
+	if currentHeight != bh.Seq {
+		s.log.Error("Attempted to index block with mismatched sequence number",
+			zap.Uint64("expected", currentHeight),
+			zap.Uint64("got", bh.Seq))
+		return fmt.Errorf("%w: expected %d, got %d", errUnexpectedSeq, currentHeight, bh.Seq)
+	}
+
+	// no need to lock the blockTracker, since Index should not be called concurrently
```
yacovm: I am not sure that the block tracker's last indexed should even be in the `block.go` given that we use it only in `storage.go`. 

Shouldn't we move it to be a field of `Storage` and just move `bt.lastIndexed = digest` to `Index()` ? 

image

simplex/storage.go
samliok: Yep correct on both questions

simplex/storage.go
samliok: yea I think that makes sense

simplex/storage_test.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// TestIndexMismatchedChild tests that the previously indexed digest matches the
+// previous digest of the block being indexed.
+func TestIndexMismatchedChild(t *testing.T) {
+	ctx := context.Background()
+	genesis := newBlock(t, newBlockConfig{})
+	child1 := newBlock(t, newBlockConfig{prev: genesis})
+	child1Sibling := newBlock(t, newBlockConfig{prev: genesis})
+	child2Cousin := newBlock(t, newBlockConfig{prev: child1Sibling})
```
yacovm: I think it's not the cousin but nephew, no?

```
	//        GB 
    //      /     \
	//  child1    child1Sibling
    //                   |
	//              child2Cousin
```

simplex/storage_test.go
```diff
@@ -0,0 +1,305 @@
...
+	_, err = child1Sibling.Verify(ctx)
+	require.NoError(t, err)
+
+	// Index child1
+	require.NoError(t, s.Index(ctx, child1, newTestFinalization(t, configs, child1.BlockHeader())))
+
+	_, err = child2Cousin.Verify(ctx)
+	require.NoError(t, err)
+
+	// Attempt to index child2 before child1 is verified
```
yacovm: I don't understand this comment. isn't `child1` verified in line 235? And also `child1Sibling` is verified in line 237

simplex/storage.go
```diff
@@ -0,0 +1,275 @@
...
+	genesis.digest = computeDigest(bytes)
+
+	return genesis, nil
+}
+
+// retrieveFinalization retrieves the finalization at [seq].
+// If the finalization is not found, it returns false.
+func (s *Storage) retrieveFinalization(seq uint64) (simplex.Finalization, error) {
+	seqBuff := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBuff, seq)
```
StephenButtolph: Should we be using `finalizationKey` helper here?

simplex/storage.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	errUnexpectedSeq    = errors.New("unexpected sequence number")
+	errInvalidQC        = errors.New("invalid quorum certificate")
+	errMismatchedDigest = errors.New("mismatched digest in finalization")
+)
+
+type Storage struct {
+	// height represents the number of blocks indexed in storage.
+	height atomic.Uint64
```
StephenButtolph, line 40: after talking offline I think we agreed to call this `numBlocks`

StephenButtolph (commented):
2 trivial comments then lgtm

PR #4123, opened 2025-07-25
refactor: Remove DB implementation from x/sync manager

alarso16:
## Why this should be merged

Currently, the sync manager depends on the merkledb proof implementation, which should be isolated from the request scheduling duties of the manager. This abstracts the proofs and all database operations into a separate object, with its methods close to the original implementations of these operations to reduce the diff.

To complete this refactor (in a separate PR) would move this new struct into a different file, and remove the unnecessary syncer creations.

## How this works

Isolates merkledb imports into a separate struct.

## How this was tested

Existing tests, plus one extra for ensuring database errors are returned as expected to end the sync.

Yes - TBD


x/sync/manager.go
```diff
@@ -63,6 +65,71 @@ const (
 	retryPriority
 )
 
+var _ DBClient = (*merkleDBSyncClient)(nil)
```
alarso16, line 67: This all should be moved to a different file, but postponed to a future PR to reduce diff

x/sync/manager.go
```diff
@@ -231,6 +286,10 @@ func (m *Manager) sync(ctx context.Context) {
 		switch {
 		case ctx.Err() != nil:
 			return // [m.workLock] released by defer.
+		case m.config.DBClient.Error() != nil:
```
alarso16, line 288: Any ideas how to get rid of this?

x/sync/manager.go
```diff
@@ -327,22 +386,16 @@ func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
...
 	if work.localRootID == targetRootID {
 		// Start root is the same as the end root, so we're done.
-		m.completeWorkItem(ctx, work, work.end, targetRootID, nil)
+		m.completeWorkItem(work, work.end, targetRootID)
 		m.finishWorkItem()
 		return
 	}
 
+	// If the target root is empty, the `FinalizeSync` method can short-circuit the sync process.
 	if targetRootID == ids.Empty {
```
alarso16, line 394: I feel like this check should really be owned by `DBClient`, but I kept it here to avoid changes in messaging (This would finish after one request, but intercepting here requires no requests). In the EVM case, this doesn't make a lot of sense anyway, since the root hash of an empty trie is not the same as `ids.Empty`. Any thoughts?

alarso16 (commented):
Some notes to help reviewers

x/sync/db.go
```diff
@@ -12,3 +20,26 @@ type DB interface {
...
+	) (maybe.Maybe[[]byte], error)
+
+	// Finalizes the sync process by ensuring the provided root is the current root of the database, and any
+	// additional cleanup is performed.
+	// If the root is empty, it clears the database.
+	FinalizeSync(ctx context.Context, root ids.ID) error
+
+	// If any unrecoverable error occurs during the sync process within the database,
+	// the error is available in this method.
+	Error() error
```
joshua-kim, line 44: One idea to get rid of this would be to have a sentinel error value this package defines like `ErrFoobar` that implementations could return/wrap their errors with so that we don't need this, we could just have x/sync check the return values of the implementation's interface to see if it should die or not.

x/sync/manager.go
joshua-kim: It's fine (and idiomatic) to have a single file in a package in Go. It's not necessary IMO to split your package across files and I find that it generally makes the package harder to maintain since any changes across files aren't easily tracable w/ Git.

This is a habit a lot of developers from other languages do because the language enforces it (Java for example), but Go doesn't require it so we're not required to do this practice.

x/sync/manager.go
```diff
@@ -63,6 +64,71 @@ const (
...
+
+type merkleDBSyncClient struct {
+	db        DB
+	config    *ClientConfig
+	tokenSize int
+	err       *utils.Atomic[error]
+	errorOnce sync.Once
+}
+
+func NewClient(db DB, config *ClientConfig) (*merkleDBSyncClient, error) {
```
joshua-kim, line 82: Try to avoid returning unexported types from exported code - it makes the return value incredibly awkward to use by external callers since the type isn't available to you.

x/sync/manager.go
```diff
@@ -63,6 +64,71 @@ const (
...
+// FinalizeSync implements DBClient.
+func (c *merkleDBSyncClient) FinalizeSync(ctx context.Context, root ids.ID) error {
+	// If the root is empty, we can clear the database and finish immediately.
+	if root == ids.Empty {
+		if err := c.db.Clear(); err != nil {
+			c.setError(err)
+			return fmt.Errorf("failed to clear database: %w", err)
+		}
+		return nil
+	}
```
joshua-kim, line 119: The caller knows the root id... so why do we couple the behavior of clearing to the verification of the sync being done? We could just have two separate i/f signatures to check that the root is correct vs to skip syncing.

x/sync/manager.go
alarso16: Ah interesting. The struct's method's are currently scattered throughout the file to reduce the diff for review, so I should probably at minimum group them together in a separate PR, right?

joshua-kim: Yeah we could do this as a subsequent PR/follow-up.

alarso16:
Using #4155 instead

PR #4124, opened 2025-07-27
Simplex reconfiguration mechanism specification

yacovm:
## Why this should be merged

Specifies the Simplex reconfiguration mechanism

## How this works

Just documentation

## How this was tested

CI

No


simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+Whenever a block proposer builds a block $B_{i+1}$ built on top of block $B_i$, it adopts the `PChainReferenceHeight` of block $B_i$, since $B_{i+1}$ is in the same epoch.
+It then samples its P-chain height and if it detects that the validator set
+is different from the validator set that is derived from the P-chain height of `PChainReferenceHeight`,
+it sets `NextPChainReferenceHeight` to the P-chain height it sampled.
+
+When the block $B_{i+1}$ is verified by nodes other than the block proposer, they check whether:
+
+- The `PChainReferenceHeight` hasn't changed from the previous block $B_i$.
+- `NextPChainReferenceHeight > PChainReferenceHeight`.
+- They have locally the P-chain height corresponding to `NextPChainReferenceHeight`.
```
samliok: could it often happen that the node has just received a new P-Chain height and sends a block before a quorum of nodes have synced that P-Chain height? I'm concerned this could lead to a lot of nodes saying blocks are not valid, then timing out only to repeat this process with the next block.

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+### Motivation and introduction
+
+Consensus systems may occasionally need to change the state of the consensus protocol uniformly and have the change appear as an atomic change.
+The simplest way of achieving a uniform and atomic state change is to encode the state change in a block and then have the block be finalized.
+Some consensus protocols (e.g. non-pipelined variants of PBFT) may support such a manner of state change, but others require a more sophisticated approach.
+
+Consider for example, the Simplex consensus protocol, where a block may be notarized in some round but not finalized in that round,
+but only in later rounds.
+In such a case, not enough nodes will finalize the block in that round, and in order for that block to be finalized, additional blocks need to be built on top of it,
+and finalized. It is only when the additional blocks are finalized, that the aforementioned block can be considered as finalized.
```
samliok: I think this is weirdly worded as it implies blocks of round `i + 1` must be finalized before blocks of round `i`. If i get this point, what your saying is we may see `i + 1` was finalized, but not `i` meaning we will initiate the replication process to finalize `i`.

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+
+Hereafter, we will refer to an MSM block as a block made by the MSM and that wraps either a VM block or a Telock.
+
+The role of telocks is to collect and aggregate information about the state of the consensus protocol in a consensus agnostic manner.
+Telocks are treated by the Simplex consensus protocol as regular blocks, yet they only get stored in the WAL and are purged when the first block of the next epoch is accepted.
+Telocks (as their name implies) can only exist at the end of an epoch, and never at the beginning or in the middle of an epoch.
+
+Since Telocks are purged once an epoch changes, their block sequence numbers are also available for reuse in the next epoch.
+For example, if in epoch `e` the sealing block is $B_k$, there can be several telocks $B_{k+1}$, $B_{k+2}, ..., B_{k+l}$ belonging to epoch $e$.
+In the successive epoch $k$, the first block will have a sequence of $B_{k+1}$, the second block will have a sequence of $B_{k+2}$ and so on and so forth.
```
samliok: what about the rounds of the block?

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+
+As mentioned before, the MSM advances its state by building blocks and looking at the previous block and the sealing block's finalization certificate if such exists.
+The MSM uses the following rules to determine how to update the metadata:
+
+1. If the current block (the block being built) is the sealing block of the current epoch, `SealingBlockSeq` is set to the current block sequence
+   and `NextPChainReferenceHeight` is set to be the P-chain height sampled.
+2. If the current block isn't the last block of the current epoch, it is either in the epoch of the sealing block or in the successive epoch:
+- 2.1 If `SealingBlockSeq` of the previous block and is smaller than the current block sequence but is greater than `0`, then:
```
samliok:
```suggestion
- 2.1 If the `SealingBlockSeq` of the previous block is smaller than the current block sequence but is greater than `0`, then:
```

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+  - 2.1.1 If the finalization certificate of the sealing block is available, then the MSM sets `SealingBlockSeq` to be 0,
```
samliok: I'm assuming we plan on updating the `MSM` state of finalizations on `Storage.Index`?

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+
+Even if the rest of the nodes finalized block 30, they will still agree to accept block 31 with metadata as above,
+as it is always possible that the block builder of block 31 did not have the finalization certificate of the sealing block.
+
+Finally, if the block builder of block 32 finalized the sealing block, then its MSM will ask the VM to build a block, and will encode the following epoch information:
+
+```
+{
+   PChainReferenceHeight: 151
+   EpochNumber: 31
```
samliok: should the epoch be `30` since it is set to the sequence of the sealing block?

simplex/reconfiguration.md
samliok, line 1: would it make sense to move this to `simplex/docs/reconfiguration.md` just to avoid file bloat in the simplex package? or `simplex/specs/..`?

simplex/reconfiguration.md
yacovm: Yeah. What will happen as a result, is the block proposal rejected and we immediately agree on the empty block (this is the reason for making [this](https://github.com/ava-labs/Simplex/pull/190) optimization) and the next leader will propose the block. Eventually we will get to a point where we have a correct leader and a quorum of nodes with that P-chain height. 

I did consider this problem and while I think there is a heuristic that may minimize the chance we encounter this (only advance the epoch once enough time has passed from the timestamp of the P-chain block) I don't think it's really a problem, as this has zero effect on user experience if we do the empty block agreement right away and don't time out. 

By the way, I think we have a similar problem for the [proposerVM](https://github.com/ava-labs/avalanchego/tree/master/vms/proposervm#snowman-validations): 

> A block must have a PChainHeight that is less or equal to current P-Chain height.

simplex/reconfiguration.md
yacovm: yeah I can move it in a follow-up PR, sure.

simplex/reconfiguration.md
yacovm: It is only weirdly worded because you know the internals of how our Simplex implementation works. 
We collect finalizations recursively because we commit blocks along with the finalization certificate in a single API call.

We only collect finalization certificates on each each block because it gives builtin validity for block replication and also we could built a threshold VRF for this in the future.

However, there could be other implementations of Simplex, where you only collect finalization certificates for blocks in rounds for which an empty block was not notarized, and the application above the consensus layer doesn't care for finalization certificate to be there for each block sequence. 

Consider for instance, an SQL database replicated via Simplex. It only needs total order among its transactions so it matters not if each block has a finalization certificate or not.

simplex/reconfiguration.md
yacovm: No, I think the state will be updated not only upon Index but also as we build blocks on top of other blocks.

simplex/reconfiguration.md
yacovm: yeah, oops...

simplex/reconfiguration.md
yacovm: Good point, so I think the same holds for round numbers. I will add it in the next revision.

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+If we assume that nodes cannot send transactions to a mem-pool like clients do, then the only way of nodes contributing self generated data is by having each node propose a block in its turn,
+and then accumulating the data generated by each node's block.
+
+To that end, we introduce the Metadata State-Machine (MSM) which lies between the consensus instance and the VM.
+The role of the MSM is to mask away the complexity of such state changes from the consensus protocol, let the consensus protocol
+be able to keep the existing application abstraction, and to prevent it from deviating from its usual standard operation.
+More specifically, the MSM detects when the state of the consensus protocol needs to change, and then it hijacks the block building and verification mechanisms of the VM,
+until the state change can be atomically applied to the consensus protocol instance.
+
+Similarly to the proposerVM, the MSM wraps the inner VM block with its own block wrapper and encodes the state transition data in the wrapped block.
```
samliok, line 148: by `innerVM block` this would mean the `SImplex` block(if its a simplex chain)?

samliok: by `innerVM block` this would mean the `SImplex` block(if its a simplex chain)?

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+We therefore add auxiliary information to the epoch information, the block validation descriptor:
+
+```
+{
+   PChainReferenceHeight uint64
+   EpochNumber uint64
+   NextPChainReferenceHeight uint64
+   SealingBlockSeq uint64
+   BlockValidationDescriptor {
+      Type: uint8
```
samliok, line 294: should this be an `enum`?

simplex/reconfiguration.md
```diff
@@ -0,0 +1,318 @@
...
+```
+{
+   PChainReferenceHeight uint64
+   EpochNumber uint64
+   NextPChainReferenceHeight uint64
+   SealingBlockSeq uint64
+}
+```
+
+We next describe how the MSM uses the above metadata to manage the Simplex epoch change.
```
samliok: Maybe a good section to include could be how the `MSM` stores this additional encoded information, and how it would allow other nodes to read from it. For example, in replication I think the non-validators would need to about some of these fields in order to verify the finalizations received from replication.

simplex/reconfiguration.md
yacovm: The inner block is the VM block, like the subnet-EVM, etc.

simplex/reconfiguration.md
yacovm: maybe 🤷

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+```
+message ICMEpochInfo {
+   uint64 epoch_start_time = 1; // The start time
+   uint64 epoch_number = 2; // The epoch number
+   uint64 p_chain_epoch_height = 3; // The P-chain height of the epoch
+```
+
+The Simplex epoch information is a protobuf encoded message with the following schema:
+
+```
```
samliok:
```suggestion
```proto
```
just a side note, but if you add proto like above it will add some color 😀

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+message OuterBlock {
+  ICMEpochInfo icm_epoch_info = 1; // The ICM epoch information.
+  SimplexEpochInfo simplex_epoch_info = 2; // The Simplex epoch information.
+  AuxiliaryInfo auxiliary_info = 3; // The auxiliary information.
+}
+```
+
+The digest of the block is computed as follows:
+
+Let $h_i$ be the hash of the inner block, $*$ be the raw bytes of the simplex block after the inner block.
```
samliok: > the raw bytes of the simplex block after the inner block

is this just the simplex metadata, or how we currently get the simplex bytes with `simplexBytes = { metadata.bytes, innerblock.bytes }` ?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+
+```
+message ObservedPChainHeight {
+   bytes node_index = 1; // The index of the node that proposed the block, always two bytes long (uint16)
+   uint64 p_chain_height = 2; // The P-chain height sampled by the node
+}
+
+message NodeBLSMapping {
+    bytes node_id = 1; // The nodeID
+    bytes bls_key = 2; // The BLS key of the node
```
samliok: why do we need the bls_keys in this message? shouldn't we be able to get them from the validator manager?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+
+- The `next_p_chain_reference_height` is the P-chain height of the next epoch and is only encoded in the last block of epoch `epoch_number` (the sealing block), otherwise it is set to `0`.
+
+- The `prev_vm_block_seq` is the sequence number of the previous VM block, and it is used to efficiently find the last VM block upon startup.
+
+- The `observed_p_chain_height` is a deterministically encoded map that contains the P-chain heights that were observed by the nodes of the current epoch.
+  The key of the map is the node index that proposed the block, and the value is the P-chain height that was sampled by that node.
+  It's used to determine what would be the `next_p_chain_reference_height` of the next epoch, based on the P-chain heights sampled by the nodes so far.
+
+- The `sealing_block_seq` is the sequence number of the sealing block of the epoch, and it is set to `0` in all blocks that are not sealing blocks.
```
samliok: it seams that we have a lot of fields that are only populated if the block is a sealing block. Can we have just have a `sealing_metadata` field in the simplex epoch info that contains all of these fields? This way we can easily organize the specific sealing block fields and omit the `sealing_metadata` if the block is not a sealing block.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+where `OuterBlock` is a protobuf message that contains the ICM epoch information, the Simplex epoch information, and auxiliary information:
+
+```
+message OuterBlock {
+  ICMEpochInfo icm_epoch_info = 1; // The ICM epoch information.
+  SimplexEpochInfo simplex_epoch_info = 2; // The Simplex epoch information.
+  AuxiliaryInfo auxiliary_info = 3; // The auxiliary information.
```
samliok: should `auxiliary_info ` be an optional field?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+    bytes bls_key = 2; // The BLS key of the node
+  }
+
+message BlockValidationDescriptor {
+  oneof BlockValidationType {
+      repeated NodeBLSMapping aggregated_membership = 1; // The BLS keys of the nodes in the next epoch
+  }
+}
+
+message NextEpochApprovals {
```
samliok: just a thought 💭

but since we have the nodes of the next epoch stating they are ready, can that epoch optimistically start producing blocks before finalization? I guess we would have two epochs running which could be gross but while the previous epoch is waiting for a finalization for the sealing block, this new epoch can start proposing and notarizing blocks. If the previous epoch doesn't get a finalization for that epoch we could just delete the new epoch.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+
+In some cases, there might not be any user activity, so blocks would not be naturally built by the underlying VM.
+However, the protocol may still need to extend the chain due to external reasons such as a change in the validator set,
+or auxiliary information that needs to be collected by a protocol independent to Simplex.
+
+To that end, the MSM will build blocks that do not contain any transactions. Such blocks are called $Metablocks$, and they are built by the MSM without any interaction with the VM.
+A Metablock can carry auxiliary information that is needed by the Simplex consensus protocol, and can be a sealing block of an epoch.
+An epoch is not expected to contain a lot of metablocks, as the MSM will always prefer building a VM block over a metablock.
+The only way for a metablock to be built is for the MSM to determine that the current epoch should be sealed or the MSM has been
+given auxiliary information as input to be included in the next block, but there isn't any user activity that would trigger the VM to build a block.
```
samliok, line 192: what is an example of `auxiliary` info that needs to be built, but can't because of inactivity(excluding sealing blocks)?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+#### Forcing chain progress
+
+In some cases, there might not be any user activity, so blocks would not be naturally built by the underlying VM.
+However, the protocol may still need to extend the chain due to external reasons such as a change in the validator set,
+or auxiliary information that needs to be collected by a protocol independent to Simplex.
+
+To that end, the MSM will build blocks that do not contain any transactions. Such blocks are called $Metablocks$, and they are built by the MSM without any interaction with the VM.
+A Metablock can carry auxiliary information that is needed by the Simplex consensus protocol, and can be a sealing block of an epoch.
+An epoch is not expected to contain a lot of metablocks, as the MSM will always prefer building a VM block over a metablock.
+The only way for a metablock to be built is for the MSM to determine that the current epoch should be sealed or the MSM has been
```
samliok, line 191: these blocks can only be built by the leader?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+is different from the validator set that is derived from the P-chain height of `p_chain_reference_height`,
+it sets `observed_p_chain_height[i]` to the sampled P-chain height.
+
+When the block $B_{k+1}$ built by the `i`'th node is verified by nodes other than the block proposer, they ensure that:
+
+- The `p_chain_reference_height` hasn't changed from the previous block $B_k$.
+- `next_p_chain_reference_height > p_chain_reference_height` or `next_p_chain_reference_height == 0`.
+- The validator set derived from the P-chain height corresponding to `next_p_chain_reference_height` is different from the validator set derived by the P-chain height corresponding to `p_chain_reference_height`.
+- If `next_p_chain_reference_height > 0`, then `observed_p_chain_height` contains at least `n-f` entries where `n` is the number of nodes in the epoch and `f` is the maximum number of faulty nodes that can be tolerated by the Simplex consensus protocol for `n` nodes.
+- If `next_p_chain_reference_height > 0`, then `next_p_chain_reference_height` is the median among the `observed_p_chain_height` values.
```
samliok: so the block builder only sets `next_p_chain_reference_height` once `n-f` entries have observed a validator set change. What happens in the case where we have multiple view changes during this process? 

Take for example a 4 node network where observed p chain heights are 2,3,4. The median would be 3 but its possible `4` was a view change noted by an adversary. Therefore only one honest node that voted for `3` would be able to verify the epoch?

simplex/docs/reconfiguration.md
yacovm: It's everything that is not the inner block. So, everything in the "outer block" section.

simplex/docs/reconfiguration.md
yacovm: We could but that's in O(n) time complexity, since we don't have an index anywhere. The only way to compute the node mapping in the P-chain is to take the current stake and apply inverse diffs iteratively until you reach the desired block height.

simplex/docs/reconfiguration.md
yacovm: yeah, good idea. This way zero values won't be taking space.

yacovm: Then we'll be left with only two fields - the epoch and the P-chain reference height.

The rest of the fields actually [shouldn't be serialized over the wire](https://protobuf.dev/programming-guides/proto3/#field-labels) if they're 0:

> the field is set to the default (zero) value. It will not be serialized to the wire. In fact, you cannot determine whether the default (zero) value was set or parsed from the wire or not provided at all.

I'm not sure it's worth it having an extra message with 6 items and a message with only 2. We might as well combine the two 🤷

simplex/docs/reconfiguration.md
yacovm: It doesn't really matter. In protobuf a field that is a non primitive type is implicitly optional.

simplex/docs/reconfiguration.md
yacovm: In theory it's possible because you know the previous block's hash once you know the sealing block. However, until the sealing block is finalized you don't know if the epoch change will succeed or not. So you might end up needing to prune also blocks from the prefix of the next epoch. This can be detrimental to applications and confuse them.

simplex/docs/reconfiguration.md
yacovm: for example, contribution of randomness for a DKG process. We want to split the process into as many blocks as possible because it's usually linear in the number of validators and we have a block size limit, so we can't just wait for the sealing block and then "dump" all auxiliary information we have in memory as it might be too big.

simplex/docs/reconfiguration.md
yacovm: yes. and also it's worth noting that auxiliary information can always come inside a VM block if a VM block needs to be built.

simplex/docs/reconfiguration.md
yacovm: If we don't have enough correct nodes that have replicated block 3 in the P-chain by the time the block with the next_p_chain_ref_height is proposed then it won't be notarized. Eventually though, enough correct nodes will replicate p-chain block 3, and then some correct node will propose this block.

simplex/docs/reconfiguration.md
yacovm: so, everything here: 
image

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
+# Reconfiguring a Simplex L1
+
+
+In this document, we describe the reconfiguration technique of Simplex.
+We start by briefly recalling the Simplex consensus protocol, following by describing the challenges of reconfiguring a consensus system,
+after which we describe how epochs address these challenges.
+Specifically, we distinguish between ICM epochs and Simplex epochs and then explain how Simplex epochs are implemented.
```
StephenButtolph:
```suggestion
Specifically, we distinguish between [ICM epochs](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/181-p-chain-epoched-views/README.md) and Simplex epochs and then explain how Simplex epochs are implemented.
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+We then introduce the MSM (Metadata State-Machine) which is a consensus agnostic functionality that is used to manage metadata for consensus systems.
+Once we have explained how epochs and metadata are maintained, we describe the Simplex reconfiguration mechanism and how it is accomplished by using the MSM.
+Lastly, we explain how a new node that onboards the system can validate all blocks in the chain from genesis despite the fact that the validator set changed over time.
+
+
```
StephenButtolph, line 11:
```suggestion
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+Then, the next epoch is numbered $k$ and not $e+1$.
+
+We denote such a block $B_k$ with sequence number $k$ which is part of epoch $e$ as the $sealing~block$ of epoch $e$.
+The validator set for epoch $k$ is derived from the P-chain height of the sealing block $B_k$.
+
+In order to support ICM epochs, the Simplex protocol facilitates the encoding of both ICM epochs and Simplex epochs.
+Whenever a block is built, the block builder encodes the ICM epoch information (`EpochStartTime`, `epoch_number`, `PChainEpochHeight`) in the same manner as the proposerVM encodes it,
+and also encodes the Simplex epoch information, as well as some additional auxiliary information that may be used by other protocols, as explained in the next section:
+
+### Simplex block encoding
```
StephenButtolph: Would it make sense to move the serialization spec / description to the end of the readme (similar to an appendix). I feel like placing it first is pretty jarring (as it really isn't too relevant for actually conveying how the epochs are managed)

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+that may be used by other protocols that require their information to be collected via the process of block building by the protocol participants.
+A common way for such information collection is to have each participant supply its input as part of the auxiliary information in the block, when it is its turn to build a block.
+
+The auxiliary information is encoded as a protobuf message with the following schema:
+
+```proto
+message AuxiliaryInfo {
+   bytes info = 1; // The auxiliary information. Can be empty if no auxiliary information is provided.
+   uint64 prev_aux_info_seq = 2; // The sequence number of the previous block containing auxiliary information.
+   bytes application_id = 3; // The application ID of the auxiliary information (if applicable) encoded as a two byte long uint16.
```
StephenButtolph:
```suggestion
   uint32 application_id = 3; // The application ID of the auxiliary information (if applicable).
```
Is is strictly more efficient to specify this as a `uint32` (or a `uint64`) rather than a byte slice representing a `uint16`.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+Apart from the ICM epoch information and the Simplex epoch information, the block also contains auxiliary information
+that may be used by other protocols that require their information to be collected via the process of block building by the protocol participants.
+A common way for such information collection is to have each participant supply its input as part of the auxiliary information in the block, when it is its turn to build a block.
+
+The auxiliary information is encoded as a protobuf message with the following schema:
+
+```proto
+message AuxiliaryInfo {
+   bytes info = 1; // The auxiliary information. Can be empty if no auxiliary information is provided.
+   uint64 prev_aux_info_seq = 2; // The sequence number of the previous block containing auxiliary information.
```
StephenButtolph, line 118: Is this the previous info _for this application id_? Or is it regardless of the application id?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+where `OuterBlock` is a protobuf message that contains the ICM epoch information, the Simplex epoch information, and auxiliary information:
+
+```proto
+message OuterBlock {
+  ICMEpochInfo icm_epoch_info = 1; // The ICM epoch information.
+  SimplexEpochInfo simplex_epoch_info = 2; // The Simplex epoch information.
+  AuxiliaryInfo auxiliary_info = 3; // The auxiliary information.
```
StephenButtolph: We only seem to allow a single `AuxiliaryInfo`, but in the `AuxiliaryInfo` we have different application IDs. Are we trying to support multiple different types of `AuxiliaryInfo`?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+The digest of the block is computed as follows:
```
StephenButtolph:
```suggestion
The digest of the simplex block is computed as follows:
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+
+Let $h_i$ be the hash of the inner block, $*$ be the raw bytes of the simplex block after the inner block.
+Then, the digest of the Simplex block is the hash of: `h_i || *` where `||` denotes concatenation.
+This way of hashing the block allows any holder of a finalization certificate for the block to authenticate the block while hiding the content of the inner block.
```
StephenButtolph: Would it be easier for us to just define the preimage as:
```proto
message SimplexBlock {
   bytes inner_block_hash = 1; // The hash of the inner block built by the VM, opaque to Simplex.
   OuterBlock outer_block = 2; // The outer block that wraps the inner block without the inner block.
   bytes protocol_metadata = 3 // The Simplex protocol metadata, set by the Simplex consensus protocol.
}
```
rather than the hybrid preimage we have right now?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+
+The `ICM epoch info` is encoded as a protobuf message with the following schema:
```
StephenButtolph: If we want to leave this as protobuf for readability, I think that makes sense. But if we want this to act as a binding spec, this should probably specify `canoto`. Otherwise the hash of these messages really isn't well defined

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+   uint64 epoch_start_time = 1; // The start time
+   uint64 epoch_number = 2; // The epoch number
+   uint64 p_chain_epoch_height = 3; // The P-chain height of the epoch
+```
+
+The Simplex epoch information is a protobuf encoded message with the following schema:
+
+```proto
+message ObservedPChainHeight {
+   bytes node_index = 1; // The index of the node that proposed the block, always two bytes long (uint16)
```
StephenButtolph: Similar comment as before, we shouldn't be using `bytes` for any integers. It is more efficient (both space and time) to use either `uint32` or `uint64` for these.

Also, canoto (not proto) has explicit support for `uint16`

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+- The validator set of the epoch numbered `epoch_number` is therefore derived from `p_chain_reference_height`.
+  Even though it is sufficient to know `epoch_number` to calculate `p_chain_reference_height`, it is encoded in the block for convenience.
+
+- The `next_p_chain_reference_height` is the P-chain height of the next epoch and is only encoded in the last block of epoch `epoch_number` (the sealing block), otherwise it is set to `0`.
+
+- The `prev_vm_block_seq` is the sequence number of the previous VM block, and it is used to efficiently find the last VM block upon startup.
+
+- The `observed_p_chain_height` is a deterministically encoded map that contains the P-chain heights that were observed by the nodes of the current epoch.
+  The key of the map is the node index that proposed the block, and the value is the P-chain height that was sampled by that node.
+  It's used to determine what would be the `next_p_chain_reference_height` of the next epoch, based on the P-chain heights sampled by the nodes so far.
```
StephenButtolph: Do we need to encode this full map in every block? Shouldn't each block only contain the leader's observed P-chain height?

simplex/docs/reconfiguration.md
StephenButtolph: It is not currently clear to me which fields are only expected to be populated by the last block.

It seems like 6 of them are only in the last block, but the description only says that for `next_p_chain_reference_height`, `sealing_block_seq`, and `block_validation_descriptor` afaict

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,473 @@
...
+It then samples its P-chain height and if it detects that the validator set
+is different from the validator set that is derived from the P-chain height of `p_chain_reference_height`,
+it sets `observed_p_chain_height[i]` to the sampled P-chain height.
+
+When the block $B_{k+1}$ built by the `i`'th node is verified by nodes other than the block proposer, they ensure that:
+
+- The `p_chain_reference_height` hasn't changed from the previous block $B_k$.
+- `next_p_chain_reference_height > p_chain_reference_height` or `next_p_chain_reference_height == 0`.
+- The validator set derived from the P-chain height corresponding to `next_p_chain_reference_height` is different from the validator set derived by the P-chain height corresponding to `p_chain_reference_height`.
+- If `next_p_chain_reference_height > 0`, then `observed_p_chain_height` contains at least `n-f` entries where `n` is the number of nodes in the epoch and `f` is the maximum number of faulty nodes that can be tolerated by the Simplex consensus protocol for `n` nodes.
```
StephenButtolph: How are nodes expected to communicate their `p_chain_height` for the next epoch if they are not also validators for the current epoch?

It seems that this was addressed / modified later on. Some part of me feels like it make more sense to describe both the leader's proposal mechanism at the same time as non-leader proposals... But this also could just be looked at with an eye towards what the leader is doing... So maybe just ignore this comment

simplex/docs/reconfiguration.md
yacovm: That's a good question. I don't think it makes sense to have several applications coexisting. The reason we have an application id is mostly for backward compatibility. I envision VMs having a single option for an application_id at a given time, and just using it to know which encoding to expect in the `bytes info` field. When the chain upgrades the application, the application id changes to reflect that.

This also means that if the chain actually has two (or more) different applications it wants their state to be encoded as part of the `bytes info` field, it will just create a wrapper encoding to wrap around both of them, and designate accordingly an application id. 

Thoughts?

simplex/docs/reconfiguration.md
yacovm: I expect that if we have different applications, they will be encoded together into the same blob of `bytes info` and the application id is just there to signal to the VM what it needs to expect in the info byte blob.

simplex/docs/reconfiguration.md
yacovm: Yeah we can use your definition. I was just trying to convey here that we want the inner block to be hashed in a standalone and independent manner.

simplex/docs/reconfiguration.md
yacovm: Yes it's a cumulative observed p-chain height. We encode it in each block for simplicity of processing. Otherwise we would need to search for the encoded p-chain height of other nodes in previous blocks and that's more complex.

However, keep in mind that in most blocks in the epoch, this map is empty, as we only start encoding this map once the validator set sampled by the nodes is different than the validator set that defines the epoch.

simplex/docs/reconfiguration.md
yacovm: > How are nodes expected to communicate their p_chain_height for the next epoch if they are not also validators for the current epoch?

They're not expected to do that. The assumption is that in the current epoch, we have up to a third nodes that can lie. Therefore, once we have `n-f` distinct nodes that observed a p-chain height, we can deduce the p-chain height for the next epoch and we know such p-chain height exists.

simplex/docs/reconfiguration.md
yacovm: > It is not currently clear to me which fields are only expected to be populated by the last block.

> It seems like 6 of them are only in the last block, but the description only says that for next_p_chain_reference_height, sealing_block_seq, and block_validation_descriptor afaict

I am not sure where the description says that.

In the example it says:

> Once enough approvals are gathered, the sealing block can be built:

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R7

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1L11

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R392-R497

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R125

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R437

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R442-R447

simplex/docs/reconfiguration.md
yacovm: changed to canoto

simplex/docs/reconfiguration.md
yacovm: https://github.com/ava-labs/avalanchego/pull/4124/commits/11cf86e0c34617736c814f0cd993277bb317be82#diff-ce6a09bcc0fb5ee66f8aacc5697bf0b18495301aae4dabaa7e7e615f91ea7de1R466

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+  This behavior makes reconfiguration challenging as in a reconfiguration, the protocol needs to change its voting logic for descendant blocks
+  based on a predecessor block which may not end up part of the chain.
+
+## 2. Epoch management
+
+### ICM epochs vs Simplex epochs
+
+The epoch management for ICM is defined as follows:
+Let $E$ be the current epoch with a start time of $T_{E}$ and end time of $T_{E} + D$.
+The first block whose build time is later than $T_{E} + D$ and is denoted B<sup>E</sup><sub>end</sub>, is the last block of epoch $E$.
```
StephenButtolph: Pedantic:
```suggestion
The first block whose build time is at least $T_{E} + D$ and is denoted B<sup>E</sup><sub>end</sub>, is the last block of epoch $E$.
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+message AuxiliaryInfo {
+   bytes info = 1; // The auxiliary information. Can be empty if no auxiliary information is provided.
+   uint64 prev_aux_info_seq = 2; // The sequence number of the previous block containing auxiliary information.
+   uint32 application_id = 3; // The application ID of the auxiliary information (if applicable).
+}
+```
+
+The `prev_aux_info_seq` field points to the previous block that contained auxiliary information, or `0` if the chain has no such block.
+As most blocks are not expected to contain auxiliary information, this provides a way for the application that uses the auxiliary information to
+efficiently find all blocks that contain auxiliary information without having to iterate over all blocks in the chain.
```
StephenButtolph: We seem to have two different workarounds to avoid maintaining state:
1. Include O(n) duplicate information in the block header
2. Expect block verification to iterate backwards over O(n) blocks

Why don't we just maintain a current state? It seems more aligned with the rest of our implementations.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+   NextEpochApprovals next_epoch_approvals = 8; // The epoch change approvals of the next epoch by at least n-f nodes.
+}
+```
+
+- The validator set of the epoch numbered `epoch_number` is derived from `p_chain_reference_height`.
+  Even though it is sufficient to know `epoch_number` to calculate `p_chain_reference_height`, it is encoded in the block for convenience.
+
+- The `next_p_chain_reference_height` is the P-chain height of the next epoch and is only encoded in the last block of epoch `epoch_number` (the sealing block), otherwise it is set to `0`.
+
+- The `prev_vm_block_seq` is the sequence number of the previous VM block, and it is used to efficiently find the last VM block upon startup.
```
StephenButtolph, line 93: I don't really follow the rationale that this is needed to efficiently find the last VM block during startup. It isn't _wrong_ to include this information... But why wouldn't we just have this written somewhere locally if it is just for startup?

I guess (after reading later) that we are attempting to avoid any usage of storage outside the block storage. I guess we wouldn't want to include this as a flag _outside_ of the hash which we could still persist with every block but avoid it needing to be distributed as part of consensus?

I suppose it doesn't really matter... Just interested on the thought process

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+- If `next_p_chain_reference_height > 0`, then `observed_p_chain_height` contains at least `n-f` entries where `n` is the number of nodes in the epoch and `f` is the maximum number of faulty nodes that can be tolerated by the Simplex consensus protocol for `n` nodes.
+- If `next_p_chain_reference_height > 0`, then `next_p_chain_reference_height` is the median among the `observed_p_chain_height` values.
+- If `observed_p_chain_height` changed from the previous block, then the only entry that changed is `observed_p_chain_height[i]`.
+
+If block $B_{k+1}$ is the sealing block of the current epoch, then `next_p_chain_reference_height > 0`, otherwise it is set to `0`.
+If block $B_{k+1}$ is considered to be the sealing block its epoch, then the next block - block $B_{k+2}$ belongs to epoch `k+1`.
+When block $B_{k+2}$ is built, it will have its `epoch_number` set to $k+1$ and its `p_chain_reference_height` set to the `next_p_chain_reference_height` of $B_{k+1}$ and
+its `next_p_chain_reference_height` set to `0`.
```
StephenButtolph: It feels like we are doing quite a bit of work here to:
1. Ensure we don't advance the pChainHeight too high (malicious node advances the pChainHeight too high)
2. Ensure we timely advance the pChainHeight (malicious node advances the pChainHeight but very slowly)

Why can't we:
1. Ensure we don't advance the pChainHeight too high by just verifying that the pChain has finalized that block (now no one can advance the pChainHeight too high)
2. Assume that a majority of the time, a virtuous node will be the block producer and advance the pChainHeight accordingly (while someone may try to slowly advance the pChainHeight, it only works for an (in expectation) constant number of proposals)

Perhaps the concern is that the verification for (1) is non-deterministic. But isn't that still a problem when we go to verify the `NextEpochApprovals` signature?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+is different from the validator set that is derived from the P-chain height of `p_chain_reference_height`,
+it sets `observed_p_chain_height[i]` to the sampled P-chain height.
+
+When the block $B_{k+1}$ built by the `i`'th node is verified by nodes other than the block proposer, they ensure that:
+
+- The `p_chain_reference_height` hasn't changed from the previous block $B_k$.
+- `next_p_chain_reference_height > p_chain_reference_height` or `next_p_chain_reference_height == 0`.
+- The validator set derived from the P-chain height corresponding to `next_p_chain_reference_height` is different from the validator set derived by the P-chain height corresponding to `p_chain_reference_height`.
+- If `next_p_chain_reference_height > 0`, then `observed_p_chain_height` contains at least `n-f` entries where `n` is the number of nodes in the epoch and `f` is the maximum number of faulty nodes that can be tolerated by the Simplex consensus protocol for `n` nodes.
+- If `next_p_chain_reference_height > 0`, then `next_p_chain_reference_height` is the median among the `observed_p_chain_height` values.
```
StephenButtolph: I feel like `median` isn't sufficiently defined when `n-f` is even (ex `n = 5` `f = 1`). What should we do in that case?

StephenButtolph: Pedantic, but do we care that the median value here might not actually change the validator set? I suppose as long as we are advancing the pChainHeight, we are happy.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+The role of the MSM is to mask away the complexity of such state changes from the consensus protocol, let the consensus protocol
+be able to keep the existing application abstraction, and to prevent it from deviating from its usual standard operation.
+More specifically, the MSM detects when the state of the consensus protocol needs to change, and then it hijacks the block building and verification mechanisms of the VM,
+until the state change can be safely applied to the consensus protocol instance.
+
+Similarly to the proposerVM, the MSM wraps the inner VM block with its own block wrapper and encodes the state transition data in the wrapped block.
+
+However, it is important to distinguish between the MSM and a VM. The MSM is a very simple construction and in most cases it is simple enough
+to be implemented as a state transition function that receives as input the previous state along with auxiliary information, and outputs the new state.
+Unlike the VM, it has no transaction memory-pool, and it cannot perform any DB lookup, nor it can do any disk or network I/O.
```
StephenButtolph: > it has no transaction memory-pool ... nor it can do any disk or network I/O

Is that true? How will validators of the next epoch provide the DKG / signatures to the validators of the current epoch?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+message BlockValidationDescriptor {
+  oneof BlockValidationType {
+      repeated NodeBLSMapping aggregated_membership = 1; // The BLS keys of the nodes in the next epoch
+  }
+}
+
+message NextEpochApprovals {
+  repeated bytes node_ids = 1; // The nodeIDs of nodes belonging to the next epoch that approve the epoch change.
```
StephenButtolph: When I first looked at this, I thought this was the nodeIDs for the next epoch. But this seems to just be the nodeIDs that _signed_ this epoch approval.

Can this just be a bitset?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+}
+
+message NodeBLSMapping {
+    bytes node_id = 1; // The nodeID
+    bytes bls_key = 2; // The BLS key of the node
+  }
+
+message BlockValidationDescriptor {
+  oneof BlockValidationType {
+      repeated NodeBLSMapping aggregated_membership = 1; // The BLS keys of the nodes in the next epoch
```
StephenButtolph: It isn't valid to include a `repeated` field in a `OneOf`

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,497 @@
...
+Apart from the ICM epoch information and the Simplex epoch information, the block also contains auxiliary information
+that may be used by other protocols that require their information to be collected via the process of block building by the protocol participants.
+A common way for such information collection is to have each participant supply its input as part of the auxiliary information in the block, when it is its turn to build a block.
+
+The auxiliary information is encoded as a canoto message with the following schema:
+
+```proto
+message AuxiliaryInfo {
+   bytes info = 1; // The auxiliary information. Can be empty if no auxiliary information is provided.
+   uint64 prev_aux_info_seq = 2; // The sequence number of the previous block containing auxiliary information.
```
StephenButtolph, line 118: I guess the expectation here is that during validation we will recursively load all the blocks that contain this information for the DKG.

I do wish we could more efficiently manage state to avoid doing a bunch of O(n) lookups... For example, when validators are posting their shares, we'll probably need to either verify that the same person hasn't contributed multiple times, or check whether a sufficient number of people have contributed... Both of which could be tracked in a single read, but would require an O(n) scan with this mechanism.

Is there a reason this seq needs to actually be included in the block hash? If we move it out of the block hash, we could at least make this a "temporary" decision on the node storage side (even if we maintain that everything is stored in a single append-only store for now)

simplex/docs/reconfiguration.md
yacovm: > I guess (after reading later) that we are attempting to avoid any usage of storage outside the block storage. I guess we wouldn't want to include this as a flag outside of the hash which we could still persist with every block but avoid it needing to be distributed as part of consensus? 

Right, the MSM should be simple and be stateless, it's a simple state transition function and all information is encoded in the blocks. 

Including this as a flag outside the hash would be possible, but then this piece of information will no longer be transferrable, and each node will need to compute it for itself for each block.

However, if we include it as part of the data that consensus agrees on, then we can replicate this information easily and by doing that, replication will be simple, there would not be a need to execute the MSM for each block.

simplex/docs/reconfiguration.md
yacovm: The backward pointers are meant to efficiently reconstruct the data from scratch after a reboot. 
I am really trying to avoid here to use yet another goLevelDB table that will end up being stale after the epoch ends, and will make testing harder.

In practice, we will never need to do any kind of O(n) lookup because the MSM (and anything that is implemented on top of it) is a state transition function. 

So, if we starting from scratch after a restart, we look at the last block of the auxiliary information and traverse backwards and build our in-memory state snapshot. 

After we have constructed our state, we receive a block and parse it, and either accept it or reject it in constant time, because our state object is already supposed to be in a form of an efficient data structure. 

The reason this is included in the hash is that we can efficiently and easily replicate it without the need of computing this ourselves. 

Consider a node that joins the L1 and now begins replicating blocks. If the signature is over everything, we never need to think about backwards compatibility, and changing logic over versions. We can just blindly replicate the entire chain even if the validation logic changed throughout the blocks.

simplex/docs/reconfiguration.md
yacovm: I guess the median isn't the best term to describe what I was thinking. 

What I really want to accomplish is to have a way of disregarding "extreme" opinions about the observed p-chain height.

To do that, I imagine at least `n-f` nodes stating their observed p-chain height, and then arranging all values in a sorted order: `v1, v2, ..., v_k`.  

If we have at least `n-f` of them, then we can have `f` nodes that are the lowest value, and `f` nodes that are the highest value, and this amounts to `2f` nodes.

Now, we are left with some amount of nodes in between, and these values are guaranteed to not harm us.

Why? Because up to `f` nodes can set us back (be among the lower values to try and stall us) and up to `f` nodes can claim an abnormally observed p-chain height. In such a case, the values between them will correspond to a value that some correct node has observed.

Does that make sense?

yacovm: I think that it makes sense to say that if the "median" observed p-chain height does not correspond to a p-chain block that changes the validator set, then we have no epoch change.

simplex/docs/reconfiguration.md
yacovm: > Ensure we don't advance the pChainHeight too high (malicious node advances the pChainHeight too high)
Ensure we timely advance the pChainHeight (malicious node advances the pChainHeight but very slowly)

That's correct but also we need a deterministic way to converge on a singe value. 

> Perhaps the concern is that the verification for (1) is non-deterministic. But isn't that still a problem when we go to verify the NextEpochApprovals signature?

Are you saying that a node that has its P-chain height too far back cannot verify a signature of a node in the next epoch because it hasn't seen it yet? I guess you're right, good point. 

So are you essentially suggesting to just set the p-chain height by the block proposer and if it's malicious, then the nodes will reject that block? 

I guess that works and simplifies things quite a lot.

Will add it tomorrow.

simplex/docs/reconfiguration.md
yacovm: The MSM is just a primitive object that can build a block based on information from the previous blocks, and can verify a block state transition.

In practice, the validators will receive information from the validators of the next epoch and will need to verify this information and insert it into the block if they are the block proposer. 

I haven't inserted this information into the spec. Do you think it's highly incomplete without it?

simplex/docs/reconfiguration.md
yacovm: Yes it can.

simplex/docs/reconfiguration.md
yacovm: thanks, fixed

simplex/docs/reconfiguration.md
yacovm: fixed

simplex/docs/reconfiguration.md
yacovm: Updated accordingly.

simplex/docs/reconfiguration.md
StephenButtolph: It isn't highly incomplete. I guess the point is just that the MSM is mainly the "verification" side, and not the building side.

simplex/docs/reconfiguration.md
StephenButtolph: Got it, so this is only intended to be done on startup (not during verification) and it is attempting to avoid iteration over the full epoch rather than just the last O(n) blocks that include any DKG information.

simplex/docs/reconfiguration.md
StephenButtolph: Yeah to clarify, the idea made sense to me. I was just commenting that there was an edge case that didn't seem well defined.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,474 @@
+# Reconfiguring a Simplex L1
+
+
+In this document, we describe the reconfiguration technique of Simplex.
```
StephenButtolph, line 3:
```suggestion
# Reconfiguring a Simplex L1

In this document, we describe the reconfiguration technique of Simplex.
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,474 @@
...
+
+Since Telocks are purged once an epoch changes, their block sequence numbers are also available for reuse in the next epoch.
+For example, if in epoch `e` the sealing block is $B_k$, there can be several telocks $B_{k+1}$, $B_{k+2}, ..., B_{k+l}$ belonging to epoch $e$.
+In the successive epoch $k$, the first block will have a sequence of $B_{k+1}$, the second block will have a sequence of $B_{k+2}$ and so on and so forth.
+
+In order to know when the current epoch can be terminated and the next epoch can be instantiated, the MSM uses the `sealing_block_seq` field of the Simplex epoch information.
+The `sealing_block_seq` is set to `0` in all blocks that are built except from when the block builder detects that the sealing block of the current epoch has been finalized.
+A telock with a `sealing_block_seq > 0` that is notarized, indicates that at least `f+1` correct nodes have finalized the sealing block of the current epoch.
+This means that under the assumption of up to `f` failures, the sealing block and its corresponding finalization certificate can always be replicated by nodes that haven't finalized it
+and by doing so, move to the next epoch.
```
StephenButtolph: I'm a bit confused by this paragraph.

It isn't clear to me why we need to add a new field `sealing_block_seq` to identify the sealing block. Couldn't we just use something like `next_p_chain_reference_height` to see if a block is the sealing block?

Also, why does seeing a notarized `telock` indicate that `f+1` correct nodes have finalized the sealing block?

yacovm: > It isn't clear to me why we need to add a new field sealing_block_seq to identify the sealing block. Couldn't we just use something like next_p_chain_reference_height to see if a block is the sealing block?

No, after we have determined the next P-chain height, we still need to approve the epoch transition because we need confirmations from `n-f` nodes in the next epoch:


```
### Criteria for creation of the sealing block of an epoch

In order to create the sealing block, the following criteria need to be met:

1. There is agreement upon which P-chain height will be used to select the validator set of the next epoch.

2. At least `n-f` nodes belonging to the next epoch have approved the epoch change.
```

Therefore the sealing block is the first block that achieves these criteria. 

You could ask "why do we need to explicitly mark what is the sealing block? can't we just look at these criteria?"

We could in theory, however it's not easy to distinguish between a block that has `n-f-k` approvals for some `k>0` and a block that has `n-f` approvals and is the sealing block. As a bootstrapping node, you would need to replicate the P-chain and figure out what is `n-f` first. However if we set it explicitly, you can replicate the L1 without replicating the P-chain, which gives you speed and also fault tolerance. 

> Also, why does seeing a notarized telock indicate that f+1 correct nodes have finalized the sealing block?

It does not, it's a mistake. Fixed, thanks for noticing!

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+It then samples its P-chain height and if it detects that the validator set
+is different from the validator set that is derived from the P-chain height of `p_chain_reference_height`,
+it sets `next_p_chain_reference_height` to the sampled P-chain height.
+
+When the block $B_{k+1}$ built by the `i`'th node is verified by nodes other than the block proposer, they ensure that:
+
+- The `p_chain_reference_height` hasn't changed from the previous block $B_k$.
+- `next_p_chain_reference_height > p_chain_reference_height` or `next_p_chain_reference_height == 0`.
+- If `next_p_chain_reference_height > 0`, the node has observed the P-chain height exists in the P-chain.
+- If `next_p_chain_reference_height > 0`, the validator set derived from the P-chain height corresponding to is different from the validator set derived by the P-chain height corresponding to `p_chain_reference_height`.
```
StephenButtolph, line 142: Presumably if the parent's `next_p_chain_reference_height != 0` then the child's `next_p_chain_reference_height` must equal the parent's value right?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+- The `next_p_chain_reference_height` is the P-chain height of the next epoch, otherwise it is set to `0`.
+
+- The `prev_vm_block_seq` is the sequence number of the previous VM block, and it is used to efficiently find the last VM block upon startup.
+
+- The `sealing_block_seq` is the sequence number of the sealing block of the epoch, and it is set to `0` before the sealing block is created.
+  It is used to determine when the Simplex instance can transition to the next epoch. It is safe to transition to the next epoch once the sealing block is finalized.
+
+- The `block_validation_descriptor` describes how to validate the blocks of the next epoch.
+  It is encoded in blocks starting from the block that has `next_p_chain_reference_height > 0` until and including the sealing block.
+  It is used to verify the quorum certificates of blocks that are built in the next epoch.
```
StephenButtolph, line 97: Can we make it so that we only include this in the sealing block? It seems like a bunch of duplicate information (which we will need to verify the correctness of in every block)... Additionally, if we ever wanted to support threshold simplex, we might not even know this until after the DKG has finished.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+A common way for such information collection is to have each participant supply its input as part of the auxiliary information in the block, when it is its turn to build a block.
+
+The auxiliary information is encoded as a canoto message with the following schema:
+
+```proto
+message AuxiliaryInfo {
+   bytes info = 1; // The auxiliary information. Can be empty if no auxiliary information is provided.
+   uint64 prev_aux_info_seq = 2; // The sequence number of the previous block containing auxiliary information.
+   uint32 application_id = 3; // The application ID of the auxiliary information (if applicable).
+}
```
StephenButtolph, line 120: I don't _really_ have a strong preference here. But I do generally find version tags in proto as an antipattern (as the message structure itself is typically preferred to handle the versioning in proto).

Would it make sense for `info` to be a OneOf rather than coupling `info` and `application_id`? Or would it make sense for `info` to handle its own versioning?

Technically we don't need to actually decide on this until we have a non-zero `application_id`... But I figured it made sense to just finalize this now.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+```
+
+- The validator set of the epoch numbered `epoch_number` is derived from `p_chain_reference_height`.
+
+- The `next_p_chain_reference_height` is the P-chain height of the next epoch, otherwise it is set to `0`.
+
+- The `prev_vm_block_seq` is the sequence number of the previous VM block, and it is used to efficiently find the last VM block upon startup.
+
+- The `sealing_block_seq` is the sequence number of the sealing block of the epoch, and it is set to `0` before the sealing block is created.
+  It is used to determine when the Simplex instance can transition to the next epoch. It is safe to transition to the next epoch once the sealing block is finalized.
```
StephenButtolph: Is this included in the sealing block as well? Or is this just to provide an O(1) operation to look-back while "flushing the pipeline" to accept the sealing block?

It reads like it is included in the sealing block, but it seems like it could just be included in the "Telock"s. Including the same seq multiple times in the block feels weird to me.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+When the block $B_{k+1}$ built by the `i`'th node is verified by nodes other than the block proposer, they ensure that:
+
+- The `p_chain_reference_height` hasn't changed from the previous block $B_k$.
+- `next_p_chain_reference_height > p_chain_reference_height` or `next_p_chain_reference_height == 0`.
+- If `next_p_chain_reference_height > 0`, the node has observed the P-chain height exists in the P-chain.
+- If `next_p_chain_reference_height > 0`, the validator set derived from the P-chain height corresponding to is different from the validator set derived by the P-chain height corresponding to `p_chain_reference_height`.
+
+If block $B_{k+1}$ is considered to be the sealing block of its epoch, then the next block - block $B_{k+2}$ belongs to epoch `k+1`.
+When block $B_{k+2}$ is built, it will have its `epoch_number` set to $k+1$ and its `p_chain_reference_height` set to the `next_p_chain_reference_height` of $B_{k+1}$ and
+its `next_p_chain_reference_height` set to `0`.
```
StephenButtolph: nit: Presumably it should be possible for the first block of an epoch to set `next_p_chain_reference_height` to a non-zero value.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+In order to create the sealing block, the following criteria need to be met:
+
+1. There is agreement upon which P-chain height will be used to select the validator set of the next epoch.
+
+2. At least `n-f` nodes belonging to the next epoch have approved the epoch change.
+
+
+The first criterion is satisfied by the `next_p_chain_reference_height` being greater than `0`.
+Since nodes check their P-chain possess the P-chain height corresponding to `next_p_chain_reference_height`,
+a block containing a specific `next_p_chain_reference_height` indicates that at least `f+1` correct nodes have observed that P-chain height in the P-chain.
```
StephenButtolph: The first sentence reads a bit weirdly to me (as `next_p_chain_reference_height > 0` doesn't really guarantee anything. It is the following check that actually does the verification.
```suggestion
The first criterion is satisfied by nodes ensuring their P-chain possess the P-chain height corresponding to `next_p_chain_reference_height`.
A block containing a specific `next_p_chain_reference_height` indicates that at least `f+1` correct nodes have observed that P-chain height in the P-chain.
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+
+The second criterion is satisfied by the `next_epoch_approvals` field containing an aggregated BLS signature from at least `n-f` nodes
+belonging to the next epoch. Unlike the previous field, this field also needs to be updated by nodes in the next epoch that might not be in the current epoch.
+To that end, nodes in the next epoch create a `NextEpochApprovals` message with `node_ids` belonging only to themselves, and broadcast it to the nodes in the current epoch.
+Then, when a node in the current epoch builds a block, it sets the bits in the `node_ids` field corresponding to the union of the existing and new approving nodes, and aggregates the signatures of the nodes.
+The signature is over the `aux_info_digest` and the `next_p_chain_reference_height` fields, to ensure that the nodes in the next epoch agree not only on the validator set,
```
StephenButtolph: Presumably we are using the BLS keys registered on the P-chain. If that's the case, we'll need to separate the message spaces between the simplex block signing and warp signing here.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+but also on any kind of application specific auxiliary information that might be needed by the next epoch. One example for such auxiliary information
+can be messages in a cryptographic protocol that are to be used in the next epoch.
+
+The `NextEpochApprovals` may only change from block to block as long as the following rules are followed:
+
+- Additional nodes may be added to the `node_ids` field, but a node may only be removed as long as the total number of node ids doesn't decrease.
+
+- The `aux_info_digest` and the `next_p_chain_reference_height` fields may only change if the number of `node_ids` increases.
+
+- The `signature` field should always match the aggregated signature of the nodes corresponding to the `node_ids` field over the `aux_info_digest` and the `next_p_chain_reference_height` fields.
```
StephenButtolph: Is there a good reason we don't just enforce that this can only be populated when there are `n-f` signatures?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+
+The aforementioned rules imply that in order to ensure liveness, the MSM needs to keep in its memory various combinations of different `aux_info_digest` and `next_p_chain_reference_height` values,
+and always try to propose the one that has the most signatures supporting it. A node may sign multiple different combinations of `aux_info_digest` and `next_p_chain_reference_height` values.
+Therefore, if an application that relies on the auxiliary information is susceptible to malicious nodes picking the combinations,
+it needs to have measures put in place.
```
StephenButtolph: I'm confused, shouldn't the node know what `aux_info_digest` and `next_p_chain_reference_height` are getting signed? I thought they were just the chain's `next_p_chain_reference_height` value and whatever the output of the `AuxiliaryInfo` was.

I feel like an application shouldn't allow a node to sign multiple of these messages in a single epoch.

Then we don't need any additional out-of-spec DoS protections

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+
+Since Telocks are purged once an epoch changes, their block sequence numbers are also available for reuse in the next epoch.
+For example, if in epoch `e` the sealing block is $B_k$, there can be several telocks $B_{k+1}$, $B_{k+2}, ..., B_{k+l}$ belonging to epoch $e$.
+In the successive epoch $k$, the first block will have a sequence of $B_{k+1}$, the second block will have a sequence of $B_{k+2}$ and so on and so forth.
+
+In order to know when the current epoch can be terminated and the next epoch can be instantiated, the MSM uses the `sealing_block_seq` field of the Simplex epoch information.
+The `sealing_block_seq` is set to `0` in all blocks that are built except from when the block builder detects that the criteria for the creation of the sealing block of the current epoch have been met.
+A telock with a `sealing_block_seq > 0` that is finalized, indicates that at least `f+1` correct nodes have finalized the sealing block of the current epoch.
+This means that under the assumption of up to `f` failures, the sealing block and its corresponding finalization certificate can always be replicated by nodes that haven't finalized it
+and by doing so, move to the next epoch.
```
StephenButtolph: I think I'm probably missing something subtle here.

I would have expected something like:
> Once we have finalized the sealing block, we can exit the epoch as we can always provide the finalization certificate to any other peer to also have them advance to the new epoch.

What am I missing here that requires finalizing a telock block? (And what guarantee does finalizing a telock block provide that finalizing the sealing block _doesn't_ provide?) It seems like we're introducing some `f+1` finalization requirement here, but it isn't clear to me why that is needed.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,471 @@
...
+defined, as a preparation for the creation of the epoch sealing block, and in the rest of the blocks it is empty.
+
+### Replicating a Simplex chain
+
+When replicating a chain driven by Simplex, the onboarding node first replicates the P-chain and acquires the latest validator set of the chain.
+It then proceeds to replicate only the latest sealing blocks by querying a majority of nodes for the sealing blocks, and dropping candidate responses
+that aren't found in the responses of enough nodes.
+
+Thanks to the fact that a sealing block $B_l$ will have the epoch number $k$ of its previous sealing block $B_k$, it is possible to recursively find out the sealing blocks
+of all previous epochs once the latest sealing block has been verified by a majority poll.
```
StephenButtolph: How do we verify the correctness of a sealing block that we are recursively fetching? Do we include the hash of the previous sealing block in the next sealing block? I don't think knowing it's seq is sufficient to trust any block provided with that seq.

simplex/docs/reconfiguration.md
StephenButtolph: I definitely agree that it should be obvious (without having the P-chain) which block is the sealing block. I just don't think that `sealing_block_seq` needs to exist for that.

simplex/docs/reconfiguration.md
StephenButtolph: After finishing the read through this time, it seems clear that this was planned to be in the sealing block... But my question still stands I think on whether or not it actually needs to be

simplex/docs/reconfiguration.md
yacovm: Yes, unless it's a new epoch and then it's 0.

simplex/docs/reconfiguration.md
yacovm: We could but then it would be very heavy/complex to validate `next_epoch_approvals`.

It's there since `next_p_chain_reference_height` is defined because you validate the `next_epoch_approvals` mutation by looking at the `block_validation_descriptor` of the previous block and using it. 

I guess we could add some kind of caching mechanism so we don't need to always re-compute `block_validation_descriptor`. 

> Additionally, if we ever wanted to support threshold simplex, we might not even know this until after the DKG has finished.

Yes that's a good point.

simplex/docs/reconfiguration.md
yacovm: I am not in favor of having `bytes` fields handling their own versioning. I think it's a slippery slope that leads to abuse and incorrect usage.

Having a one-of replaces the application_id quite elegantly, however then how would one integrate its own custom L1 into avalanchego? He/She would need to recompile the protobufs to change the application.

That's why I have the application_id as a separate field - to solidify the protobuf structure and make it permanent yet flexible, so that if in the future someone wants to extend its own L1 he/she could do it without the need to extend the protobuf schema.

simplex/docs/reconfiguration.md
yacovm: Yeah, so - it is included in the sealing block as a means to say "this is a sealing block" but also is included in the telocks (though it is only hinted, not explicitly stated) in order to know how far backwards to flush them.

simplex/docs/reconfiguration.md
StephenButtolph: I guess the question below around including multiple `next_epoch_approvals` is relevant here as well. Since if we only included `next_epoch_approvals` and `block_validation_descriptor` once, then we wouldn't eat any additional complexity.

simplex/docs/reconfiguration.md
yacovm: I agree but I don't think this should be written in the spec. 

We should also make sure that we encode into the message space things like the chain ID, subnet ID, etc. so I think domain separation with other protocols is too much detail for this specification.

simplex/docs/reconfiguration.md
yacovm: It doesn't require it to be finalized, it's just a misleading wording.

simplex/docs/reconfiguration.md
yacovm: Yeah I think we can make it as you suggest, since this is a behavior we anyway need to support for the next epoch's validators that aren't in the current epoch's validator set.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,460 @@
...
+belonging to the next epoch. Unlike the previous field, this field also needs to be updated by nodes in the next epoch that might not be in the current epoch.
+To that end, nodes in the next epoch create a `NextEpochApprovals` message with `node_ids` belonging only to themselves, and broadcast it to the nodes in the current epoch.
+Then, when a node in the current epoch builds a block, it sets the bits in the `node_ids` field corresponding to the union of the existing and new approving nodes, and aggregates the signatures of the nodes.
+The signature is over the `aux_info_digest` and the `next_p_chain_reference_height` fields, to ensure that the nodes in the next epoch agree not only on the validator set,
+but also on any kind of application specific auxiliary information that might be needed by the next epoch. One example for such auxiliary information
+can be messages in a cryptographic protocol that are to be used in the next epoch.
+
+When validating the block, the MSM ensures that the `aux_info_digest` corresponds to the digest of the `info` field in the `AuxiliaryInfo` encoded in the block.
```
StephenButtolph: This line confused me a bit. `aux_info_digest` is only included in `NextEpochApprovals`. But `AuxiliaryInfo` can technically be included whenever.

Which `info` field are we using to verify the `aux_info_digest`?

yacovm: image

I'm just saying that the info field from the `AuxiliaryInfo` message needs to be consistent with the `aux_info_digest`. The `NextEpochApprovals` are signing over both `aux_info_digest` and the `next_p_chain_reference_height` fields.

yacovm: I removed it since I think it's more succinct this way.

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,457 @@
...
+
+In order to support ICM epochs, the Simplex protocol facilitates the encoding of both ICM epochs and Simplex epochs.
+Whenever a block is built, the block builder encodes the ICM epoch information (`EpochStartTime`, `epoch_number`, `PChainEpochHeight`) in the same manner as the proposerVM encodes it,
+and also encodes the Simplex epoch information, as well as some additional auxiliary information that may be used by other protocols, as explained below:
+
+```proto
+message SimplexEpochInfo {
+   uint64 p_chain_reference_height = 1;
+   uint64 epoch_number = 2;
+   bytes prev_sealing_block_hash = 3; // The hash of the sealing block of the previous epoch
```
samliok, line 77: should we use `Digest` instead of `Hash` to keep it consistent with our naming conventions?
```suggestion
   bytes prev_sealing_block_digest = 3; // The digest of the sealing block of the previous epoch
```

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,457 @@
...
+   uint64 next_p_chain_reference_height = 4;
+   uint64 prev_vm_block_seq = 5; // The sequence of the previous VM block
+   BlockValidationDescriptor block_validation_descriptor = 6; // Describes how to validate the blocks of the next epoch
+   NextEpochApprovals next_epoch_approvals = 7; // The epoch change approvals of the next epoch by at least n-f nodes.
```
samliok, line 81: If a node sends a finalized vote for a sealing block, doesn't that count as an approval? Or are we requiring `n-f` approvals by the next epoch's nodes?

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,457 @@
...
+given auxiliary information as input to be included in the next block, but there isn't any user activity that would trigger the VM to build a block.
+
+
+### Criteria for creation of the sealing block of an epoch
+
+In order to create the sealing block, the following criteria need to be met:
+
+1. There is agreement upon which P-chain height will be used to select the validator set of the next epoch.
+
+2. At least `n-f` nodes belonging to the next epoch have approved the epoch change.
```
samliok, line 201: i think this answers my comment above 👍

simplex/docs/reconfiguration.md
```diff
@@ -0,0 +1,457 @@
...
+    next_p_chain_reference_height: 151
+    block_validation_descriptor: [TGVhcm4gdG8g==, dmFsdWUgeW91cnNlbGY==, ... , XBwaW5lc3MuCg==]
+    next_epoch_approvals: {
+        node_ids: 10111101
+        signature: "erpcyBqaGUg05z=="
+    }
+}
+```
+
+In case the sealing block is not finalized yet, the MSM will build descendent Telocks until it is finalized:
```
samliok, line 295: I get that the MSM builds descendent Telocks if it is the block proposer, but what happens in this scenario?

we get a sealing block(SB) for seq x `SB_x` but we don't finalize it. Lets say the other nodes in the network get an empty notarization for that round, and the next block producer creates a normal block for seq x. How does the block builder + MSM know which block to build off of? Would it create a telock ontop of `SB_x` or does it create its own sealing block `SB_x+1`?

Does this change if the non sealing block `B_x` gets finalized? And what if it gets finalized by a descendant?

simplex/docs/reconfiguration.md
yacovm: We care about approvals from the next epoch's validator set, not from this epoch's validator set. That's the idea behind the next epoch approvals. Otherwise we would not have that.

simplex/docs/reconfiguration.md
yacovm: > How does the block builder + MSM know which block to build off of?

It knows which block to build off according to the block Simplex builds on. 
If it builds on top of a sealing block, it's a Telock and the VM is throttled.
Otherwise it just builds a VM block. 

If the sealing block doesn't end up finalized then it's fine, we will eventually build another sealing block, won't we?

Let me know if i'm missing something.

PR #4133, opened 2025-07-28
Add HeightIndex interface to database

DracoLi:
## Why this should be merged

Add an `BlockDatabase` interface to `database` package. This will be used by the VMs to integrate x/blockdb, starting with [coreth](https://github.com/ava-labs/coreth) (see [POC PR](https://github.com/ava-labs/coreth/pull/1079)).

## How this works

Interface added to `database/database.go`

No


database/database.go
```diff
@@ -94,3 +94,24 @@ type Database interface {
...
+
+// BlockDatabase defines the interface for storing and retrieving blockchain blocks.
+type BlockDatabase interface {
+	// WriteBlock inserts a block into the store at the given height with the specified header size.
+	WriteBlock(height uint64, block []byte, headerSize uint32) error
+
+	// ReadBlock retrieves a block by its height.
+	ReadBlock(height uint64) ([]byte, error)
+
+	// ReadHeader retrieves only the header portion of a block by its height.
```
yacovm: In avalanchego we also need to [just read the block ID](https://github.com/ava-labs/avalanchego/blob/c202c5aed1b04441012089c4fa411c4cdc19578a/snow/engine/snowman/block/vm.go#L59), which is used to service block dissemination.

I guess this header is only used for C-chain, but probably we should have yet another API call for the block ID...?

DracoLi: I think we can keep this mapping data (id -> height) still stored in the key-value db. The blockdb can just store the bytes.

DracoLi: yeah `ReadHeader` is specifically added for the c-chain since their interface allows reading header & body separately. However, I'm also exploring whether we can just not to support this, as the benefits appear to be minimal.

DracoLi: removed read header and body

maru-ava:
Does this have a target consumer (i.e. simplex)?

DracoLi:
@maru-ava yes, coreth will be using this first. I'm working on a PR to switch it to use blockdb.

maru-ava:
> @maru-ava yes, coreth will be using this first. I'm working on a PR to switch it to use blockdb.

Maybe link the coreth PR that depends on this PR so that the intention is clear?

database/database.go
```diff
@@ -94,3 +94,21 @@ type Database interface {
...
+	WriteBlock(height uint64, block []byte) error
+
+	// ReadBlock retrieves a block by its height.
+	ReadBlock(height uint64) ([]byte, error)
+
+	// HasBlock checks if a block exists at the given height.
+	HasBlock(height uint64) (bool, error)
+
+	// Inspect returns a detailed report of the block database.
+	Inspect() (string, error)
```
joshua-kim: What is this for?

database/database.go
```diff
@@ -94,3 +94,21 @@ type Database interface {
...
 	health.Checker
 }
+
+// BlockDatabase defines the interface for storing and retrieving blockchain blocks.
+type BlockDatabase interface {
+	// WriteBlock inserts a block into the store at the given height.
+	WriteBlock(height uint64, block []byte) error
+
+	// ReadBlock retrieves a block by its height.
+	ReadBlock(height uint64) ([]byte, error)
```
joshua-kim: 1. Should we use `Put/Get/Has` instead of `Write/Read/HasBlock`? 
2. Do we not support deletion?

database/database.go
DracoLi: provide report of the database. It will be used here https://github.com/ava-labs/coreth/blob/master/plugin/evm/vm_database.go#L54
For the block database, I plan to output data such as total data file size, number of data files, index file size, number of blocks, etc.

database/database.go
DracoLi: 1. I think we can, but I think it also makes sense to not match exactly with the key-value database methods to avoid confusion. I'm open to change this though if you feel strongly.
2. No deletion support. Only accepted blocks should be stored.

database/database.go
joshua-kim: It just seems weird to put onto the interface to me... would it make sense for us to expose an [iterator](https://pkg.go.dev/iter) api so that we could consume that to still be able to build these reports?

DracoLi: i don't like putting the iterator api on this since the reports cannot be generated by iterating the blocks.
We don't necessarily need the inspect so I'm fine with omitting it. We can always do type assertion if needed to call inspect.

joshua-kim: Yeah upon reading the related code I don't even know why we have these database reports in the C-Chain

database/database.go
joshua-kim: Another question is whether we want to return an error on the interface `ErrNotFound` if the block isn't found... we could also return something like
```
WriteBlock(...) (bool, error)
```

and return an error if we have some issue with i/o or something, and return false if the block isn't there. I generally hate having to reason about sentinel errors, especially when there's only one possible one.

I also wonder if we should define our own `Close` method instead of embedding `io.Close`. One one hand it is consistent w/ the rest of the style (not a reason), but on the other hand, I don't think we explicitly use the `io.Close` interface at all. 

I'm going to approve anyways since I feel like these naming/style/sentinel error comments are probably something for Stephen to decide.

joshua-kim (approved):

database/database.go
DracoLi: Yeah I'm very open to changing this. I don't feel strongly over the Close interface. As for the `ReadBlock` return type, at least in coreth, it currently doesn't need to check for not found errors as it handles all errors the same way, but could be different for other vms.

database/database.go
```diff
@@ -94,3 +94,18 @@ type Database interface {
...
+
+	// HasBlock checks if a block exists at the given height.
+	HasBlock(height uint64) (bool, error)
+
+	// Close closes the block database.
+	io.Closer
+}
```
StephenButtolph:
```suggestion
// HeightIndex defines the interface for storing and retrieving entries by height.
type HeightIndex interface {
	// Put inserts the entry into the store at the given height.
	Put(height uint64, bytes []byte) error

	// Get retrieves an entry by its height.
	Get(height uint64) ([]byte, error)

	// Has checks if an entry exists at the given height.
	Has(height uint64) (bool, error)

	// Close closes the database.
	io.Closer
}
```

StephenButtolph (approved):

PR #4135, opened 2025-07-29
uplift: Add combined metrics package from evm repositories

### LINKED ISSUE #4127: Uplift metrics package

JonathanOppenheimer:
## Why this should be merged
There are packages in subnet-evm and coreth that share a lot of functionality / are mirrored. These packages should be uplifted to AvalancheGo to prevent the duplicative maintenance of two sets of the same code. This is part of that effort. This would close #4127. 

## How this works
This code represents the 'merged' difference of `coreth/metrics` and `subnet-evm/metrics.` The two packages diverged slightly, and deference was given to additions.  

Key merges: 
- Takes `metricsLock` locking from `subnet-evm` 
- Takes and supports `GaugeInfo` metrics from `subnet-evm` 
- Takes `enabled_test.go` from `coreth` - this test does not exist in `subnet-evm` 

Linting changes can be viewed standalone in 45f25b0d25d40d8d8dd7347a85e7cacb7fbc717d

## How this was tested
This is currently 'dead' code. From my understanding, this would need to be included in a release of AvalancheGo to then be used in the evm repositories. 

No


vms/evm/metrics/prometheus/interfaces.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package prometheus
+
+import "github.com/ava-labs/libevm/metrics"
+
+var _ Registry = metrics.Registry(nil)
+
```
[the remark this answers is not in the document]
JonathanOppenheimer: This is unnecessary

vms/evm/metrics/prometheus/interfaces.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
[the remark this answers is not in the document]
JonathanOppenheimer: This is unnecessary

alarso16 (approved):
Looks good from the coreth/subnet-evm perspective

vms/evm/metrics/metricstest/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/ava-labs/libevm/metrics"
+)
+
+var metricsLock sync.Mutex
+
+// WithMetrics enables go-ethereum metrics globally for the test.
+// If the [metrics.Enabled] is already true, nothing is done.
+// Otherwise, it is set to true and is reverted to false when the test finishes.
```
StephenButtolph: Should this function document that it prevents any other tests from running `WithMetrics` concurrently?
```suggestion
// WithMetrics enables [metrics.Enabled] for the test and prevents any other
// tests with metrics from running concurrently.
//
// If [metrics.Enabled] is modified, its original value is restored as part of
// the testing cleanup.
```

vms/evm/metrics/metricstest/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+func WithMetrics(t *testing.T) {
+	metricsLock.Lock()
+	t.Cleanup(func() {
+		metricsLock.Unlock()
+	})
```
StephenButtolph:
```suggestion
	t.Cleanup(metricsLock.Unlock)
```

vms/evm/metrics/metricstest/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/ava-labs/libevm/metrics"
+)
+
+var metricsLock sync.Mutex
+
+// WithMetrics enables go-ethereum metrics globally for the test.
+// If the [metrics.Enabled] is already true, nothing is done.
+// Otherwise, it is set to true and is reverted to false when the test finishes.
+func WithMetrics(t *testing.T) {
```
StephenButtolph: For helpers, ideally `testing.TB` should be used so that this helper can be called from tests/benchmarks/fuzzs rather than just tests.
```suggestion
func WithMetrics(t testing.TB) {
```

vms/evm/metrics/prometheus/enabled_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestMetricsEnabledByDefault(t *testing.T) {
+	require.True(t, metrics.Enabled, "libevm/metrics.Enabled")
+
+	switch m := metrics.NewCounter().(type) {
+	case metrics.NilCounter:
+		require.Failf(t, "unexpected type", "metrics.NewCounter() got %T; want %T", m, new(metrics.StandardCounter))
+	case *metrics.StandardCounter:
+	default:
+		require.Failf(t, "unexpected type", "metrics.NewCounter() got unknown type %T", m)
+	}
```
StephenButtolph:
```suggestion
	require.IsType(t, (*metrics.StandardCounter)(nil), metrics.NewCounter(), "metrics.NewCounter() returned wrong type")
```

vms/evm/metrics/prometheus/enabled_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"testing"
+
+	"github.com/ava-labs/libevm/metrics"
+	// NOTE: This test assumes that there are no imported packages that might
+	// change the default value of [metrics.Enabled]. It is therefore in package
+	// `prometheus_test` in case any other tests modify the variable. If any
+	// imports here or in the implementation do actually do so then this test
+	// may have false negatives.
```
StephenButtolph: It seems like this comment is misplaced, can we move this comment to be on the actual `TestMetricsEnabledByDefault` function? Also the `// Note: ` seems unnecessary.

vms/evm/metrics/prometheus/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+
+package prometheus
+
+import "github.com/ava-labs/libevm/metrics"
+
+var _ Registry = metrics.Registry(nil)
+
+type Registry interface {
+	// Call the given function for each registered metric.
+	Each(func(string, any))
```
StephenButtolph: Nit: while most people could probably guess `name`, it seems like `any` is probably not super descriptive.
```suggestion
	Each(func(name string, metric any))
```

vms/evm/metrics/prometheus/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+	// Get the metric by the given name or nil if none is registered.
+	Get(string) any
```
StephenButtolph:
```suggestion
	Get(name string) any
```

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+					Value: ptrTo(m.Snapshot().Value()),
+				},
+			}},
+		}, nil
+	case metrics.GaugeInfo:
+		return nil, fmt.Errorf("%w: %q is a %T", errMetricSkip, name, m)
+	case metrics.Histogram:
+		snapshot := m.Snapshot()
+
+		quantiles := []float64{.5, .75, .95, .99, .999, .9999}
```
StephenButtolph: `quantiles` seems to be used here and with `metrics.Timer`. Can we move this out to the global `var` block?

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+					Quantile:    dtoQuantiles,
+				},
+			}},
+		}, nil
+	case metrics.ResettingTimer:
+		snapshot := m.Snapshot()
+		count := snapshot.Count()
+		if count == 0 {
+			return nil, fmt.Errorf("%w: %q resetting timer metric count is zero", errMetricSkip, name)
+		}
```
StephenButtolph: At the risk of asking a question about code I may have previously written, why do we disallow `count == 0` for this metric?

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+
+		pvShortPercent := []float64{50, 95, 99}
```
StephenButtolph: Similarly to the other constant, I think we should probably define this in a global var block (so that we can avoid populating this slice every time this case is hit).

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+		mfs = append(mfs, mf)
+	}
+
+	return mfs, nil
+}
+
+var (
+	errMetricSkip             = errors.New("metric skipped")
+	errMetricTypeNotSupported = errors.New("metric type is not supported")
+)
```
StephenButtolph: in avalanchego we typically have a single `var` block at the top of the file to specify all the globals for the file.

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Gatherer implements [prometheus.Gatherer] interface by
+// gathering all metrics from the given Prometheus registry.
+type Gatherer struct {
+	registry Registry
+}
+
+var _ prometheus.Gatherer = (*Gatherer)(nil)
```
StephenButtolph: nit: Can we move this up to the top of the file (aligns with all the other type assertions in avalanchego)

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+	"strings"
+
+	"github.com/ava-labs/libevm/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Gatherer implements [prometheus.Gatherer] interface by
+// gathering all metrics from the given Prometheus registry.
```
StephenButtolph:
```suggestion
// Gatherer implements the [prometheus.Gatherer] interface by gathering all
// metrics from a [Registry].
```

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+	slices.Sort(names)
+
+	mfs = make([]*dto.MetricFamily, 0, len(names))
+	for _, name := range names {
+		mf, err := metricFamily(g.registry, name)
+		switch {
+		case errors.Is(err, errMetricSkip):
+			continue
+		case err != nil:
+			return nil, err
```
StephenButtolph: From the spec on `prometheus.Gatherer#Gather`:

> // Even if an error occurs, Gather attempts to gather as many metrics as
    // possible. Hence, if a non-nil error is returned, the returned
    // MetricFamily slice could be nil (in case of a fatal error that
    // prevented any meaningful metric collection) or contain a number of
    // MetricFamily protobufs, some of which might be incomplete, and some
    // might be missing altogether.

Rather than erroring here, it seems like we should be accumulating an `[]error` and then returning `errors.Join` of the resulting (potentially empty) slice.

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,168 @@
...
+	nilHistogram := metrics.NewHistogram(nil)
+	register(t, "nil/histogram", nilHistogram)
+	nilMeter := metrics.NewMeter()
+	register(t, "nil/meter", nilMeter)
+	nilResettingTimer := metrics.NewResettingTimer()
+	register(t, "nil/resetting_timer", nilResettingTimer)
+	nilSample := metrics.NewUniformSample(1028)
+	register(t, "nil/sample", nilSample)
+	nilTimer := metrics.NewTimer()
+	register(t, "nil/timer", nilTimer)
```
StephenButtolph: I feel like putting these on a single line significantly improves the readability of this function.
```suggestion

	register(t, "nil/counter", metrics.NewCounter())
	register(t, "nil/counter_float64", metrics.NewCounterFloat64())
	register(t, "nil/ewma", &metrics.NilEWMA{})
	register(t, "nil/gauge", metrics.NewGauge())
	register(t, "nil/gauge_float64", metrics.NewGaugeFloat64())
	register(t, "nil/gauge_info", metrics.NewGaugeInfo())
	register(t, "nil/healthcheck", metrics.NewHealthcheck(nil))
	register(t, "nil/histogram", metrics.NewHistogram(nil))
	register(t, "nil/meter", metrics.NewMeter())
	register(t, "nil/resetting_timer", metrics.NewResettingTimer())
	register(t, "nil/sample", metrics.NewUniformSample(1028))
	register(t, "nil/timer", metrics.NewTimer())
```

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	register(t, "unsupported", metrics.NewHealthcheck(nil))
+	families, err = gatherer.Gather()
+	require.ErrorIs(t, err, errMetricTypeNotSupported)
+	require.Empty(t, families)
+}
+
+func registerNilMetrics(t *testing.T, register func(t *testing.T, name string, collector any)) {
+	metrics.Enabled = false
```
StephenButtolph, line 148:
```suggestion
	// The NewXXX metrics functions return nil metrics types when the metrics
	// are disabled.
	metrics.Enabled = false
```

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	resettingTimer := metrics.NewResettingTimer()
+	register(t, "test/resetting_timer", resettingTimer)
+	resettingTimer.Update(time.Second) // must be after register call
+
+	emptyResettingTimer := metrics.NewResettingTimer()
+	register(t, "test/empty_resetting_timer", emptyResettingTimer)
+
+	emptyResettingTimer.Update(time.Second) // no effect because of snapshot below
+	register(t, "test/empty_resetting_timer_snapshot", emptyResettingTimer.Snapshot())
```
StephenButtolph, line 142: Would it make sense to move this out to a helper function `registerRealMetrics` like how we do with `registerNilMetrics`? I feel like it is a bit hard to actually see what this function is testing currently.

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,168 @@
...
+test_timer{quantile="0.9999"} 1.2e+08
+test_timer_sum 2.3e+08
+test_timer_count 6
+`
+	var (
+		stringReader = strings.NewReader(expectedString)
+		parser       expfmt.TextParser
+	)
+	expectedMetrics, err := parser.TextToMetricFamilies(stringReader)
+	require.NoError(t, err)
```
StephenButtolph: I know I just added this, but I wonder if it would be possible to separate this from this test definition. It looks a lot like we are testing the parsing here, but we are really just generating the expected metrics. Perhaps we could expose this as another helper function `func expectedMetrics(t *testing.T) map[string]*dto.MetricFamily {`

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	require.Len(t, families, len(expectedMetrics))
+	for i, got := range families {
+		require.NotNil(t, *got.Name)
```
StephenButtolph: Whoops:
```suggestion
		require.NotNil(t, got.Name)
```

vms/evm/metrics/metricstest/metrics.go
StephenButtolph: I also wonder whether or not it would actually make more sense if we _always_ restored the metrics value with the initial value.

Currently, we only restore `metrics.Enabled = false` at the end of the test if it was initially set to `false`, but perhaps we should just be restoring it to whatever value it had at the beginning of the test. That way if it is initially set to `true`, and we in the test set it to `false`, then we would ensure that it is restored to `true` when the test finishes.

JonathanOppenheimer: This makes a lot of sense - I've made the change. Out of curiosity why would a test test that modify `metrics.Enabled` say halfway through itself?

vms/evm/metrics/prometheus/enabled_test.go
JonathanOppenheimer: Oops!

vms/evm/metrics/prometheus/prometheus.go
JonathanOppenheimer: This is conjecture but! 

- For the `metrics.Timer` case, `SampleSum` is calculated as `ptrTo(float64(snapshot.Sum()))`
- For the `metrics.ResettingTimer` case,   `SampleSum` is calculated as `ptrTo(float64(count) * snapshot.Mean())` to reconstruct the total sum. I assume this was done as  `metrics.ResettingTime`'s `snapshot` implementation does not provide a `sum()` function. 

Thus if `count == 0,` `SampleSum` will just be 0 and thus meaningless / wholly incorrect? This checks ensures only meaningful metrics are sent.

vms/evm/metrics/prometheus/prometheus_test.go
JonathanOppenheimer: Yeah that's significantly cleaner - delineates the code better too.

vms/evm/metrics/prometheus/prometheus_test.go
JonathanOppenheimer: After both of the function changes I like `TestGatherer_Gather` a lot more.

vms/evm/metrics/metricstest/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/ava-labs/libevm/metrics"
+)
+
+var metricsLock sync.Mutex
+
+// WithMetrics enables [metrics.Enabled] for the test and prevents any other
+// tests with metrics from running concurrently.
+//
+// If [metrics.Enabled] is modified, its original value is restored as part of
+// the testing cleanup.
```
StephenButtolph:
```suggestion
// [metrics.Enabled] is restored to its original value during testing cleanup.
```

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,205 @@
...
+			errs = append(errs, err)
+			continue
+		}
+		mfs = append(mfs, mf)
+	}
+
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
+	}
+	return mfs, nil
```
StephenButtolph: I think the comment on the interface for `Gather` suggests returning errors and data.
```suggestion
	return mfs, errors.Join(errs...)
```

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,205 @@
...
+		mf, err := metricFamily(g.registry, name)
+
+		switch {
+		case errors.Is(err, errMetricSkip):
+			continue
+		case err != nil:
+			errs = append(errs, err)
+			continue
+		}
+		mfs = append(mfs, mf)
```
StephenButtolph: I think removing the control flow logic makes this a bit more readable
```suggestion
		mf, err := metricFamily(g.registry, name)
		switch {
		case err == nil:
			mfs = append(mfs, mf)
		case !errors.Is(err, errMetricSkip):
			errs = append(errs, err)
		}
```

vms/evm/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,205 @@
...
+func (g *Gatherer) Gather() (mfs []*dto.MetricFamily, err error) {
+	// Gather and pre-sort the metrics to avoid random listings
+	var names []string
+	var errs []error
+	g.registry.Each(func(name string, _ any) {
+		names = append(names, name)
+	})
+	slices.Sort(names)
+
+	mfs = make([]*dto.MetricFamily, 0, len(names))
```
StephenButtolph: nit: can we define `errs` closer to where it is used?
```suggestion
	var names []string
	g.registry.Each(func(name string, _ any) {
		names = append(names, name)
	})
	slices.Sort(names)

	var (
		mfs = make([]*dto.MetricFamily, 0, len(names))
		errs []error
	)
```

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,170 @@
...
+		require.NotNil(t, got.Name)
+
+		want := expectedMetrics[*got.Name]
+		require.Equal(t, want, got, i)
+	}
+
+	register(t, "unsupported", metrics.NewHealthcheck(nil))
+	families, err = gatherer.Gather()
+	require.ErrorIs(t, err, errMetricTypeNotSupported)
+	require.Empty(t, families)
```
StephenButtolph: When we adopt the expected behavior for `Gather` (to return as many metrics as possible), `families` should equal the previous value of `families`.

StephenButtolph (commented):
final comments then lgtm

vms/evm/metrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,170 @@
...
+	require.NoError(t, err)
+
+	expectedMetrics := expectedMetrics(t)
+	require.Len(t, families, len(expectedMetrics))
+	for i, got := range families {
+		require.NotNil(t, got.Name)
+
+		want := expectedMetrics[*got.Name]
+		require.Equal(t, want, got, i)
+	}
```
StephenButtolph: Actually, @joshua-kim pointed out that we should probably be using [`testutil.GatherAndCompare`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil#GatherAndCompare)

```golang
require.NoError(testutil.GatherAndCompare(
	gatherer,
	strings.NewReader(wantMetrics),
))

JonathanOppenheimer: This test case is actually really weird when you return both the `mfs` and the errors. Check my other comment about how the `resetting timer` behaves differently. I wrote a kind of jank fix to get the test passing, but it's definitely not what you had in mind. curious as to your thoughts.

vms/evm/metrics/prometheus/prometheus_test.go
JonathanOppenheimer: Resolving this in favor of discussion in other comment.

StephenButtolph (approved):
lgtm

vms/evm/metrics/prometheus/prometheus.go
JonathanOppenheimer: This was fixed by @StephenButtolph in 22ea450

PR #4144, opened 2025-07-29
feat(load): non-zero random transfers

RodrigoVillar:
## How this was tested

CI

No


tests/load/tests.go
```diff
@@ -65,11 +65,13 @@ func NewRandomTest(
 		// writes, or computes) for a test that supports repeated operations.
 		// This value is arbitrary but kept constant to ensure test reproducibility.
 		count = big.NewInt(5)
+		// value specifies the amount to send in a transfer test
+		value = big.NewInt(1)
```
maru-ava, line 69: (No action required) Is this the smallest amount that can be transferred?

tests/load/tests.go
```diff
@@ -223,16 +227,21 @@ func (ZeroTransferTest) Run(
 	bigGwei := big.NewInt(params.GWei)
 	gasTipCap := new(big.Int).Mul(bigGwei, big.NewInt(1))
 	gasFeeCap := new(big.Int).Mul(bigGwei, maxFeeCap)
-	senderAddress := crypto.PubkeyToAddress(wallet.privKey.PublicKey)
+
+	// Generate non-existent account address
+	pk, err := crypto.GenerateKey()
+	require.NoError(err)
+	recipient := crypto.PubkeyToAddress(pk.PublicKey)
```
maru-ava, line 234: (No action required) No change required in this PR, but iirc there was a capability in the hypersdk load testing that would collect transferred funds in the background to avoid funds exhaustion. Is funds exhaustion not a possibility for this testing or just not yet?

tests/load/tests.go
RodrigoVillar: Yes, `value` is denominated in wei which is the smallest denomination of AVAX.

tests/load/tests.go
RodrigoVillar: The general workflow for HyperSDK was as follows:

- Distribute funds from a funder account to `n` account
- `n` accounts run the load test
- At the end, the `n` accounts send funds back to the original funder

Since we're not using a funder here, there's no need to send back funds. Furthermore, since we're sending only 1 Wei per transaction, funds exhaustion is much more likely to occur from high fees than by sending 1 Wei.

joshua-kim (approved):

PR #4145, opened 2025-07-30
Info connectrpc service

joshua-kim:
## Why this should be merged

Migrates the existing `info` api to the connectrpc protocol.

## How this works

Factored out of https://github.com/ava-labs/avalanchego/pull/4033

Supports a wider range of clients, supports streaming use-cases, and provides type-safety out-of-the-box with proto files. Also, our documentation kind of lives in code now since it comes as part of the proto file.

## How this was tested

Was mostly tested manually. Added a e2e test, but it's mostly a smoke test at this point.

No


api/info/service.go
```diff
@@ -68,13 +68,12 @@ var (
 // Info is the API service for unprivileged info on a node
 type Info struct {
 	Parameters
-	log          logging.Logger
-	validators   validators.Manager
-	myIP         *utils.Atomic[netip.AddrPort]
-	networking   network.Network
-	chainManager chains.Manager
-	vmManager    vms.Manager
```
joshua-kim: Redundant parameter

tests/e2e/api/api.go
```diff
@@ -0,0 +1,158 @@
...
+		require.NotZero(response.Msg)
+	})
+
+	ginkgo.It("serves GetUptime", func() {
+		request := connect.NewRequest(&infopb.GetUptimeRequest{})
+		response, err := client.GetUptime(tc.DefaultContext(), request)
+		require.NoError(err)
+
+		require.NotZero(response.Msg.RewardingStakePercentage)
+		require.NotZero(response.Msg.WeightedAveragePercentage)
```
joshua-kim: A lot of these tests could definitely be better... but it's hard to get something that doesn't flake + doesn't require changes to tmpnet to get better coverage. FWIW, we don't have any api tests right now.

api/info/connect/service.go
```diff
@@ -0,0 +1,349 @@
...
+		trackedSubnetsIDs := peer.TrackedSubnets.List()
+		trackedSubnets := make([]string, len(trackedSubnetsIDs))
+		for i, id := range trackedSubnetsIDs {
+			trackedSubnets[i] = id.String()
+		}
+
+		benched := make([]string, len(peer.Benched))
+		copy(benched, peer.Benched)
+
+		peers = append(peers, &infopb.Peer{
```
joshua-kim, line 196: NumPeers was removed since that information is implicitly available in the response, as well as ACPs since that's available over GetACPs

connectproto/info/service.proto
```diff
@@ -0,0 +1,213 @@
...
+}
+message GetBootstrappedResponse {
+  // True if this node is running this chain and has finished bootstrapping it.
+  bool bootstrapped = 1;
+}
+
+message GetUpgradesRequest {}
+message GetUpgradesResponse {
+  // Starting time for the Apricot Phase 1 upgrade.
+  google.protobuf.Timestamp apricot_phase_1_time = 1;
```
joshua-kim, line 138: I wasn't sure if it made sense to have the `_time` suffix since it's clear from the type that it is... but we could also just be explicit if we wanted to.

api/info/service_test.go
```diff
@@ -1,94 +0,0 @@
...
-	resources := initGetVMsTest(t)
-
-	id1 := ids.GenerateTestID()
-	id2 := ids.GenerateTestID()
-	vmIDs := []ids.ID{id1, id2}
-	alias1 := []string{id1.String(), "vm1-alias-1", "vm1-alias-2"}
-
-	resources.mockVMManager.EXPECT().ListFactories().Times(1).Return(vmIDs, nil)
-	resources.mockVMManager.EXPECT().Aliases(id1).Times(1).Return(alias1, nil)
-	resources.mockVMManager.EXPECT().Aliases(id2).Times(1).Return(nil, errTest)
```
joshua-kim: My PR broke these tests... we could fix them/rewrite them to not use mocks but I felt like they had questionable value to begin with + I was horrified by the code I wrote 3 years ago...

tests/e2e/api/api.go
maru-ava: Are these tests that benefit from being tested with consensus (i.e. more than one node)? If not, maybe they would be better tested in a lighter-weight manner (e.g. a single node integration test)?

joshua-kim: Yeah these do benefit from multi-node setups, but for this test specifically we can't guarantee what the exact expected uptime is since we don't know in advance what exact range of time nodes will be up for.

PR #4149, opened 2025-08-01
[monitoring] Clean up monitoring configuration in preparation for grafana cloud migration

maru-ava:
## Why this should be merged

Our current monitoring stack was only ever intended to be a stop-gap until a managed alternative became available. Grafana Cloud is that alternative. This PR cleans up the configuration of monitoring (endpoints and urls) to simplify the pending migration. 

## How this was tested

CI

N/A

### DIFF
--- a/.github/actions/run-monitored-tmpnet-cmd/action.yml
+++ b/.github/actions/run-monitored-tmpnet-cmd/action.yml
@@ -50,7 +50,7 @@ inputs:
     default: ${{ github.job }}
   grafana_dashboard_id:
     description: 'The identifier of the Grafana dashboard to use, in the format <UID>/<dashboard-name>.'
-    default: 'mabpvtq/avalanche-main-dashboard'
+    default: 'kBQpRdWnk/avalanche-main-dashboard'
 
 runs:
   using: composite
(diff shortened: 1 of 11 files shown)


tests/fixture/tmpnet/monitor_processes.go
```diff
@@ -26,20 +26,18 @@ import (
 )
 
 const (
-	collectorTickerInterval = 100 * time.Millisecond
+	collectorTickerInterval = 1 * time.Second
```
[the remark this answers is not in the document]
maru-ava: Thanks but no thanks. This is just increasing the interval between queries for collected metrics or logs.

joshua-kim (approved):

joshua-kim (commented):
LGTM pending a fix for the rate-limiting.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -20,6 +20,14 @@ inputs:
   aws-region:
     description: 'AWS region to use for S3 access.'
     required: true
+  prometheus-url:
```
maru-ava: tmpnet expects this value for use in verifying metrics collection, but it's up to the job to actually call the check when `TMPNET_CHECK_METRICS_COLLECTED` is set.

aaronbuchwald: Clarifying this comment - it looks like this flag is set to true by default if the prometheus username is populated. Reading this comment, it sounds like you are saying it's up to this or another job using it to check that metrics are collected, but it seems like that's taken care of for us. Am I misunderstanding here?

aaronbuchwald (approved):
LGTM - left one question on a comment I did not understand

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: - The job itself needs to respond to `TMPNET_CHECK_METRICS_COLLECTED` being set. 
- [e2e.NewTestEnvironment](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/e2e/env.go#L99) adds a cleanup function to check for collected metrics if `TMPNET_CHECK_METRICS_COLLECTED` is set. This cleanup function requires `PROMETHEUS_URL` to query prometheus.
- The re-execution jobs do not currently respond to `TMPNET_CHECK_METRICS_COLLECTED`
- ~But the re-execution jobs still need to provide `PROMETHEUS_URL` (despite not using it - metrics collection only requires `PROMETHEUS_PUSH_URL`) because I have chosen not to special-case the tmpnet code that collects the environment variables~ 

Edit: PROMETHEUS_URL is no longer required if checks are not performed.

maru-ava:
CI is passing. Need to add the open source dashboards back to grafana and update the metrics link to the main dashboard.

PR #4151, opened 2025-08-01
Separate re-execution job params for PR from schedule

aaronbuchwald:
This PR updates the params used on `pull_request` events to only execute the range [101, 100k], so that it runs in ~4mins. This is small enough to make sure it's not slowing down CI while also making sure that it runs, so we catch any potential break.

### DIFF
--- a/.github/workflows/c-chain-reexecution-benchmark.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark.yml
@@ -30,6 +30,7 @@ on:
 
 jobs:
     c-chain-reexecution:
+      if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'ava-labs/avalanchego'
       permissions:
         id-token: write
         contents: write


.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -43,12 +43,28 @@ jobs:
...
+                echo "END_BLOCK=100000"
+                echo "SOURCE_BLOCK_DIR=s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb.zip"
+                echo "CURRENT_STATE_DIR=s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100.zip"
+              } >> "$GITHUB_ENV"
+            else
+              {
+                echo "START_BLOCK=${{ github.event.inputs.start-block }}"
+                echo "END_BLOCK=${{ github.event.inputs.end-block }}"
+                echo "SOURCE_BLOCK_DIR=${{ github.event.inputs.source-block-dir }}"
+                echo "CURRENT_STATE_DIR=${{ github.event.inputs.current-state-dir }}"
```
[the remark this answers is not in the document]
aaronbuchwald: There's no need for a default here as this is handled by the other block block rather than in-line. I think this is much clearer and easy to follow.

The manual workflow already contains defaults for each parameter, so a user should know not to set them to empty and it's fine imo for this to error if a user does.

maru-ava:
Please make sure the job can run against fork branches (i.e. without access to repo secrets). Currently the job fails if repo secrets are not present (e.g. https://github.com/ava-labs/avalanchego/actions/runs/16667501218/job/47176600599).

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -35,23 +35,51 @@ jobs:
...
+              } >> "$GITHUB_ENV"
+            fi
+
+            # Determine if this is an external fork and set SHOULD_RUN accordingly
+            if [[ "${{ github.event.pull_request.head.repo.full_name }}" != "" && "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]]; then
+              echo "SHOULD_RUN=false" >> "$GITHUB_ENV"
```
maru-ava: Should this job run at all if repo secrets are not available? I'm assuming not, because that would mean that aws credentials wouldn't be available and then how would the test get the data it requires.

So, maybe add a conditional checking for any of the secrets at the job level as per https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions

aaronbuchwald: Thanks! Missed this, this is much cleaner than the previous version of skipping it

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -30,6 +30,7 @@ on:
 
 jobs:
     c-chain-reexecution:
+      if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'ava-labs/avalanchego'
```
maru-ava, line 33: Maybe instead perform an explicit check for one of the secrets being non-empty? I recall trying to conditionally detect fork branches in a similar manner in the past and it not always working reliably, and the important detail anyway is whether the secrets are available.

aaronbuchwald: Secrets are not available in the if condition evaluation stage, so to do that I think that I'd need to create a separate job that checks for it, create an output, and then read it here. I prefer this solution over creating a separate job just to do it, but did consider it.

Do you know what conditions caused this type of condition to perform unexpectedly? Open to a better solution, but don't think we can check if the secrets are available in the if condition directly unfortunately.

maru-ava: All I remember is that I went back to checking the secret value in the run-monitored-tmpnet-cmd action  for emitting a warning about fork branches. But as you say, secrets aren't supported for job if conditionals so what you have is fine and hopefully whatever issues I ran into are unrelated.

joshua-kim (approved):

PR #4155, opened 2025-08-04
refactor: create proof interface

alarso16:
## Why this should be merged

Alternative to #4123 

I think it has quite a few benefits:
- More clearly separates responsibilities between the DB and the network
- The proof directly references the backing database, and decouples `findNextKey` for more proper testing
- A lot easier to read the manager code (once the proof logic and parser is moved to a different location)

Here's the things I'm not sure about:
- Right now the database is still needed for `Clear` and `GetMerkleRoot`. I think `Clear` is easily removed if you allow a single request to send the superfluous proof, but not sure about the `GetMerkleRoot` check
- The DB is closely coupled to `ProofParser`, so if we do need to supply even just one method for the database, that seems prone to misuse
- This applies a lot of the specific logic (to protect against DoS, for example) into the proof handling logic, so it probably needs doc-ed pretty well in the interface.

Next steps (i.e. different PR):
- Instantiate the `ProofParser` separately, and provide it instead of the `DB` to the `Manager`.
- Clean up the `findNextKey` tests, since they should only test that function, decoupled from the syncer
- Move these structs into a separate file/package (TBD where, will consult code owners)

## How this works

Creates `rangeProof` and `changeProof` structs using a shared interface generated from a `ProofParser`.

## How this was tested

Existing UT

No

### DIFF
--- a/x/sync/db.go
+++ b/x/sync/db.go
@@ -18,21 +18,9 @@ type DB interface {
 	merkledb.RangeProofer
 }
 
-// A `Proof` is a stateful object that represents a proof returned from a remote peer.
-// It must be verified before being committed to the underlying database.
-// The `Proof` is backed by a database, which may be altered during Commit.
-// Since the `Proof` must be returned by a `ProofParser`, it has knowledge of the request
-// that generated it, which will be used when finding the next key to request.
-type Proof interface {
-	// Commit applies the changes in the proof to the underlying database.
-	// An error is returned if the commit fails, in which case the DB may be
-	// corrupted. The byte slice returned is the next key that should be
-	// requested. If Nothing is returned, there are no more keys to request
-	// before the end key of the request.
-	Commit(context.Context) (maybe.Maybe[[]byte], error)
-}
-
-type ProofParser interface {
-	ParseRangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
-	ParseChangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
+type ProofHandler[T, U any] interface {
+	ParseRangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (T, error)
+	ParseChangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (U, error)
+	CommitRangeProof(ctx context.Context, proof T) (maybe.Maybe[[]byte], error)
+	CommitChangeProof(ctx context.Context, proof U) (maybe.Maybe[[]byte], error)
 }
--- a/x/sync/manager.go
+++ b/x/sync/manager.go
@@ -38,9 +38,7 @@ const (
 )
 
 var (
-	_ Proof       = (*rangeProof)(nil)
-	_ Proof       = (*changeProof)(nil)
-	_ ProofParser = (*proofParser)(nil)
+	_ ProofHandler[*rangeProofResponse, *changeProofResponse] = (*proofHandler)(nil)
 
 	ErrAlreadyStarted                = errors.New("cannot start a Manager that has already been started")
 	ErrAlreadyClosed                 = errors.New("Manager is closed")
@@ -52,6 +50,7 @@ var (
 	ErrZeroWorkLimit                 = errors.New("simultaneous work limit must be greater than 0")
 	errInvalidRangeProof             = errors.New("failed to verify range proof")
 	errInvalidChangeProof            = errors.New("failed to verify change proof")
+	errNoProofFound                  = errors.New("no proof found in change proof response")
 	errTooManyKeys                   = errors.New("response contains more than requested keys")
 	errTooManyBytes                  = errors.New("response contains more than requested bytes")
 	errUnexpectedChangeProofResponse = errors.New("unexpected response type")
@@ -67,13 +66,13 @@ const (
 	retryPriority
 )
 
-type proofParser struct {
+type proofHandler struct {
 	db        DB
 	hasher    merkledb.Hasher
 	tokenSize int
 }
 
-func NewParser(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor) (ProofParser, error) {
+func NewHandler(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor) (ProofHandler[*rangeProofResponse, *changeProofResponse], error) {
 	if db == nil {
 		return nil, ErrNoDBProvided
 	}
@@ -85,7 +84,7 @@ func NewParser(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor
 		hasher = merkledb.DefaultHasher
 	}
 
-	return &proofParser{
+	return &proofHandler{
 		db:        db,
 		hasher:    hasher,
 		tokenSize: merkledb.BranchFactorToTokenSize[branchFactor],
@@ -125,7 +124,8 @@ func newWorkItem(localRootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[
 	}
 }
 
-type Manager struct {
+type Manager[T, U any] struct {
+	proofHandler ProofHandler[T, U]
 	// Must be held when accessing [config.TargetRoot].
 	syncTargetLock sync.RWMutex
 	config         ManagerConfig
@@ -170,7 +170,6 @@ type Manager struct {
 
 // TODO remove non-config values out of this struct
 type ManagerConfig struct {
-	Parser                ProofParser
 	RangeProofClient      *p2p.Client
 	ChangeProofClient     *p2p.Client
 	SimultaneousWorkLimit int
@@ -179,14 +178,14 @@ type ManagerConfig struct {
 	StateSyncNodes        []ids.NodeID
 }
 
-func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
+func NewManager[T, U any](proofHandler ProofHandler[T, U], config ManagerConfig, registerer prometheus.Registerer) (*Manager[T, U], error) {
 	switch {
+	case proofHandler == nil:
+		return nil, ErrNoParserProvided
 	case config.RangeProofClient == nil:
 		return nil, ErrNoRangeProofClientProvided
 	case config.ChangeProofClient == nil:
 		return nil, ErrNoChangeProofClientProvided
-	case config.Parser == nil:
-		return nil, ErrNoParserProvided
 	case config.Log == nil:
 		return nil, ErrNoLogProvided
 	case config.SimultaneousWorkLimit == 0:
@@ -198,7 +197,8 @@ func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manage
 		return nil, err
 	}
 
-	m := &Manager{
+	m := &Manager[T, U]{
+		proofHandler:    proofHandler,
 		config:          config,
 		doneChan:        make(chan struct{}),
 		unprocessedWork: newWorkHeap(),
@@ -210,7 +210,7 @@ func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manage
 	return m, nil
 }
 
-func (m *Manager) Start(ctx context.Context) error {
+func (m *Manager[_, _]) Start(ctx context.Context) error {
 	m.workLock.Lock()
 	defer m.workLock.Unlock()
 
@@ -234,7 +234,7 @@ func (m *Manager) Start(ctx context.Context) error {
 // sync awaits signal on [m.unprocessedWorkCond], which indicates that there
 // is work to do or syncing completes.  If there is work, sync will dispatch a goroutine to do
 // the work.
-func (m *Manager) sync(ctx context.Context) {
+func (m *Manager[_, _]) sync(ctx context.Context) {
 	defer func() {
 		// Invariant: [m.workLock] is held when this goroutine begins.
 		m.close()
@@ -272,7 +272,7 @@ func (m *Manager) sync(ctx context.Context) {
 }
 
 // Close will stop the syncing process
-func (m *Manager) Close() {
+func (m *Manager[_, _]) Close() {
 	m.workLock.Lock()
 	defer m.workLock.Unlock()
 
@@ -281,7 +281,7 @@ func (m *Manager) Close() {
 
 // close is called when there is a fatal error or sync is complete.
 // [workLock] must be held
-func (m *Manager) close() {
+func (m *Manager[_, _]) close() {
 	m.closeOnce.Do(func() {
 		// Don't process any more work items.
 		// Drop currently processing work items.
@@ -299,7 +299,7 @@ func (m *Manager) close() {
 	})
 }
 
-func (m *Manager) finishWorkItem() {
+func (m *Manager[_, _]) finishWorkItem() {
 	m.workLock.Lock()
 	defer m.workLock.Unlock()
 
@@ -308,7 +308,7 @@ func (m *Manager) finishWorkItem() {
 }
 
 // Processes [item] by fetching a change or range proof.
-func (m *Manager) doWork(ctx context.Context, work *workItem) {
+func (m *Manager[_, _]) doWork(ctx context.Context, work *workItem) {
 	// Backoff for failed requests accounting for time this job has already
 	// spent waiting in the unprocessed queue
 	now := time.Now()
@@ -339,7 +339,7 @@ func (m *Manager) doWork(ctx context.Context, work *workItem) {
 
 // Fetch and apply the change proof given by [work].
 // Assumes [m.workLock] is not held.
-func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
+func (m *Manager[_, _]) requestChangeProof(ctx context.Context, work *workItem) {
 	targetRootID := m.getTargetRoot()
 
 	if work.localRootID == targetRootID {
@@ -393,7 +393,7 @@ func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
 
 // Fetch and apply the range proof given by [work].
 // Assumes [m.workLock] is not held.
-func (m *Manager) requestRangeProof(ctx context.Context, work *workItem) {
+func (m *Manager[_, _]) requestRangeProof(ctx context.Context, work *workItem) {
 	targetRootID := m.getTargetRoot()
 
 	request := &pb.SyncGetRangeProofRequest{
@@ -437,7 +437,7 @@ func (m *Manager) requestRangeProof(ctx context.Context, work *workItem) {
 	m.metrics.RequestMade()
 }
 
-func (m *Manager) sendRequest(ctx context.Context, client *p2p.Client, requestBytes []byte, onResponse p2p.AppResponseCallback) error {
+func (m *Manager[_, _]) sendRequest(ctx context.Context, client *p2p.Client, requestBytes []byte, onResponse p2p.AppResponseCallback) error {
 	if len(m.config.StateSyncNodes) == 0 {
 		return client.AppRequestAny(ctx, requestBytes, onResponse)
 	}
@@ -450,7 +450,7 @@ func (m *Manager) sendRequest(ctx context.Context, client *p2p.Client, requestBy
 	return client.AppRequest(ctx, set.Of(nodeID), requestBytes, onResponse)
 }
 
-func (m *Manager) retryWork(work *workItem) {
+func (m *Manager[_, _]) retryWork(work *workItem) {
 	work.priority = retryPriority
 	work.queueTime = time.Now()
 	work.requestFailed()
@@ -462,7 +462,7 @@ func (m *Manager) retryWork(work *workItem) {
 }
 
 // Returns an error if we should drop the response
-func (m *Manager) shouldHandleResponse(
+func (m *Manager[_, _]) shouldHandleResponse(
 	bytesLimit uint32,
 	responseBytes []byte,
 	err error,
@@ -489,7 +489,7 @@ func (m *Manager) shouldHandleResponse(
 	return nil
 }
 
-func (m *Manager) handleRangeProofResponse(
+func (m *Manager[_, _]) handleRangeProofResponse(
 	ctx context.Context,
 	targetRootID ids.ID,
 	work *workItem,
@@ -501,7 +501,7 @@ func (m *Manager) handleRangeProofResponse(
 		return err
 	}
 
-	rangeProof, err := m.config.Parser.ParseRangeProof(
+	rangeProof, err := m.proofHandler.ParseRangeProof(
 		ctx,
 		responseBytes,
 		request.RootHash,
@@ -513,7 +513,7 @@ func (m *Manager) handleRangeProofResponse(
 		return err
 	}
 	// Replace all the key-value pairs in the DB from start to end with values from the response.
-	nextKey, err := rangeProof.Commit(ctx)
+	nextKey, err := m.proofHandler.CommitRangeProof(ctx, rangeProof)
 	if err != nil {
 		m.setError(err)
 		return nil
@@ -531,7 +531,7 @@ type proofRequest struct {
 	keyLimit int
 }
 
-func (p *proofParser) ParseRangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error) {
+func (p *proofHandler) ParseRangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (*rangeProofResponse, error) {
 	var rangeProofProto pb.RangeProof
 	if err := proto.Unmarshal(responseBytes, &rangeProofProto); err != nil {
 		return nil, err
@@ -542,7 +542,7 @@ func (p *proofParser) ParseRangeProof(ctx context.Context, responseBytes, rootHa
 		return nil, err
 	}
 
-	proof := &rangeProof{
+	proof := &rangeProofResponse{
 		merkleProof: &merkleProof,
 		request: &proofRequest{
 			rootHash: rootHash,
@@ -550,58 +550,53 @@ func (p *proofParser) ParseRangeProof(ctx context.Context, responseBytes, rootHa
 			endKey:   endKey,
 			keyLimit: int(keyLimit),
 		},
-		db:        p.db,
-		tokenSize: p.tokenSize,
-		hasher:    p.hasher,
 	}
 
-	if err := proof.verify(ctx); err != nil {
+	if err := p.verifyRangeProof(ctx, proof); err != nil {
 		return nil, err
 	}
 	return proof, nil
 }
 
-type rangeProof struct {
+type rangeProofResponse struct {
 	merkleProof *merkledb.RangeProof
 	request     *proofRequest
-	db          DB
-	tokenSize   int
-	hasher      merkledb.Hasher
 }
 
-func (r *rangeProof) verify(ctx context.Context) error {
+func (p *proofHandler) verifyRangeProof(ctx context.Context, proof *rangeProofResponse) error {
 	// Database is empty, no proof is needed.
-	if bytes.Equal(r.request.rootHash, ids.Empty[:]) {
+	if bytes.Equal(proof.request.rootHash, ids.Empty[:]) {
 		return nil
 	}
 
 	return verifyRangeProof(
 		ctx,
-		r.merkleProof,
-		r.request.keyLimit,
-		r.request.startKey,
-		r.request.endKey,
-		r.request.rootHash,
-		r.tokenSize,
-		r.hasher,
+		proof.merkleProof,
+		proof.request.keyLimit,
+		proof.request.startKey,
+		proof.request.endKey,
+		proof.request.rootHash,
+		p.tokenSize,
+		p.hasher,
 	)
 }
 
-func (r *rangeProof) Commit(ctx context.Context) (maybe.Maybe[[]byte], error) {
+func (p *proofHandler) CommitRangeProof(ctx context.Context, proof *rangeProofResponse) (maybe.Maybe[[]byte], error) {
 	// If the root is empty, we clear the database of all values.
-	if bytes.Equal(r.request.rootHash, ids.Empty[:]) {
-		return maybe.Nothing[[]byte](), r.db.Clear()
+	if bytes.Equal(proof.request.rootHash, ids.Empty[:]) {
+		return maybe.Nothing[[]byte](), p.db.Clear()
 	}
 
-	// Range proofs must always be committed, even if no key changes are present.
-	// This is because it may have been provided instead of a change proof, and the keys were deleted.
-	if err := r.db.CommitRangeProof(ctx, r.request.startKey, r.request.endKey, r.merkleProof); err != nil {
-		return maybe.Nothing[[]byte](), err
+	// Since this is a response to a range proof, we don't actually have to make any changes if no keys were provided
+	if len(proof.merkleProof.KeyChanges) > 0 {
+		if err := p.db.CommitRangeProof(ctx, proof.request.startKey, proof.request.endKey, proof.merkleProof); err != nil {
+			return maybe.Nothing[[]byte](), err
+		}
 	}
 
 	// Find the next key to fetch.
 	// This will return Nothing if there are no more keys to fetch in the range `largestHandledKey` to `r.request.endKey`.
-	nextKey, err := findNextKey(ctx, r.db, r.merkleProof.KeyChanges, r.request.endKey, r.merkleProof.EndProof, r.tokenSize)
+	nextKey, err := findNextKey(ctx, p.db, proof.merkleProof.KeyChanges, proof.request.endKey, proof.merkleProof.EndProof, p.tokenSize)
 	if err != nil {
 		return maybe.Nothing[[]byte](), err
 	}
@@ -609,7 +604,7 @@ func (r *rangeProof) Commit(ctx context.Context) (maybe.Maybe[[]byte], error) {
 	return nextKey, nil
 }
 
-func (m *Manager) handleChangeProofResponse(
+func (m *Manager[_, _]) handleChangeProofResponse(
 	ctx context.Context,
 	targetRootID ids.ID,
 	work *workItem,
@@ -621,7 +616,7 @@ func (m *Manager) handleChangeProofResponse(
 		return err
 	}
 
-	proof, err := m.config.Parser.ParseChangeProof(
+	proof, err := m.proofHandler.ParseChangeProof(
 		ctx,
 		responseBytes,
 		request.EndRootHash,
@@ -633,7 +628,7 @@ func (m *Manager) handleChangeProofResponse(
 		return fmt.Errorf("%w: %w", errInvalidChangeProof, err)
 	}
 
-	nextKey, err := proof.Commit(ctx)
+	nextKey, err := m.proofHandler.CommitChangeProof(ctx, proof)
 	if err != nil {
 		m.setError(err)
 		return nil
@@ -643,12 +638,19 @@ func (m *Manager) handleChangeProofResponse(
 	return nil
 }
 
-func (p *proofParser) ParseChangeProof(ctx context.Context, responseBytes, endRootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error) {
+type changeProofResponse struct {
+	rangeProof  *merkledb.RangeProof
+	changeProof *merkledb.ChangeProof
+	request     *proofRequest
+}
+
+func (p *proofHandler) ParseChangeProof(ctx context.Context, responseBytes, endRootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (*changeProofResponse, error) {
 	var changeProofResp pb.SyncGetChangeProofResponse
 	if err := proto.Unmarshal(responseBytes, &changeProofResp); err != nil {
 		return nil, err
 	}
 
+	var proof *changeProofResponse
 	switch changeProofResp := changeProofResp.Response.(type) {
 	case *pb.SyncGetChangeProofResponse_ChangeProof:
 		// The server had enough history to send us a change proof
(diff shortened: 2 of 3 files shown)


x/sync/manager.go
```diff
@@ -510,43 +541,113 @@ func (m *Manager) handleRangeProofResponse(
...
+		p.request.startKey,
+		p.request.endKey,
+		p.request.rootHash,
+		p.tokenSize,
+		p.hasher,
+	)
+}
+
+func (p *rangeProof) Commit(ctx context.Context) error {
+	if len(p.merkleProof.KeyChanges) == 0 {
```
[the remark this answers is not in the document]
alarso16: This feels really silly, I'm ignoring this

x/sync/manager.go
```diff
@@ -510,43 +541,113 @@ func (m *Manager) handleRangeProofResponse(
...
+	// If this is empty, then we have no more keys to fetch.
+	if !largestHandledKey.IsNothing() {
+		nextKey, err := findNextKey(ctx, p.db, largestHandledKey.Value(), p.request.endKey, p.merkleProof.EndProof, p.tokenSize)
+		if err != nil {
+			return maybe.Nothing[[]byte](), err
+		}
+		largestHandledKey = nextKey
+	}
+
+	return largestHandledKey, nil
```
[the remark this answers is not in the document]
alarso16: This is a good idea, I will do this (well, a version of this that makes sense)

x/sync/manager.go
```diff
@@ -63,6 +63,39 @@ const (
...
+	BranchFactor merkledb.BranchFactor
+}
+
+type merkleDBProofParser struct {
+	db        DB
+	config    *ProofParserConfig
+	tokenSize int
+}
+
+func NewClient(db DB, config *ProofParserConfig) (ProofParser, error) {
```
joshua-kim: nit:
1. can be unexported
2. Maybe re-name this to `newParser`?
3. We could just pass in the hasher/branch factor as parameters and get rid of the `ProofParserConfig` type entirely I think

x/sync/manager.go
```diff
@@ -63,6 +63,39 @@ const (
 	retryPriority
 )
 
+var (
+	_ Proof       = (*rangeProof)(nil)
+	_ Proof       = (*changeProof)(nil)
+	_ ProofParser = (*merkleDBProofParser)(nil)
+)
```
joshua-kim: nit: merge package-level vars into the top-level var block

x/sync/manager.go
```diff
@@ -167,12 +201,13 @@ func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manage
 	case config.SimultaneousWorkLimit == 0:
 		return nil, ErrZeroWorkLimit
 	}
-	if err := config.BranchFactor.Valid(); err != nil {
-		return nil, err
-	}
 
-	if config.Hasher == nil {
-		config.Hasher = merkledb.DefaultHasher
```
joshua-kim, line 175: Seems like we should keep this default override for the syncer and have the proof parser consume whatever the syncer uses

x/sync/manager.go
```diff
@@ -63,6 +63,39 @@ const (
...
+}
+
+func NewClient(db DB, config *ProofParserConfig) (ProofParser, error) {
+	if err := config.BranchFactor.Valid(); err != nil {
+		return nil, err
+	}
+
+	if config.Hasher == nil {
+		config.Hasher = merkledb.DefaultHasher
+	}
```
joshua-kim: nit: move this back to `New`

x/sync/manager.go
```diff
@@ -97,6 +130,8 @@ func newWorkItem(localRootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[
 }
 
 type Manager struct {
+	proofParser ProofParser
+
```
joshua-kim: nit: I don't know what's idiomatic here but I generally observe this ordering:
1. Embedded stuff at the top of the struct (but also don't embed anyways, because in general it leads to weird behavior)
2. Things the caller passed in to you + exported stuff
3. Things you own + unexported stuff

Generally whitespace separates these, so I'd probably put this somewhere towards the bottom. In our codebase we also generally put a lock over the fields it synchronizes in a block but I think that's something specific to our repo style. We're also not consistent about whitespacing, but the general ordering of 1->2->3 generally is maintained.

x/sync/manager.go
```diff
@@ -510,43 +541,113 @@ func (m *Manager) handleRangeProofResponse(
...
+		p.request.keyLimit,
+		p.request.startKey,
+		p.request.endKey,
+		p.request.rootHash,
+		p.tokenSize,
+		p.hasher,
+	)
+}
+
+func (p *rangeProof) Commit(ctx context.Context) error {
```
joshua-kim: nit: idiomatic in go to have a receiver starting with the first letter of the type, so `r` would be idiomatic here.

x/sync/manager.go
```diff
@@ -510,43 +541,113 @@ func (m *Manager) handleRangeProofResponse(
...
+			endKey:   endKey,
+			keyLimit: int(keyLimit),
+		},
+		db:        m.db,
+		tokenSize: m.tokenSize,
+		hasher:    m.config.Hasher,
+	}, nil
+}
+
+type rangeProofContext struct {
```
joshua-kim: nit: Should we replace the `*Context` suffix with `*Request` since that's where this data came from? I generally find that context is an overloaded term and doesn't indicate much information (like *backend.... and *manager...).

x/sync/db.go
```diff
@@ -12,3 +17,14 @@ type DB interface {
...
+
+type Proof interface {
+	Verify(context.Context) error
+	Commit(context.Context) error
+	FindNextKey(context.Context) (maybe.Maybe[[]byte], error)
+}
+
+type ProofParser interface {
+	RangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
+	ChangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
```
joshua-kim: I learned this recently but typically Go will prefer to have a verb prefix for actions that have some side effect. In this case we're returning a `Proof` to the caller, so something like `ParseRangeProof` might make sense here. Typically nouns are reserved for returning data or querying state.

x/sync/manager.go
```diff
@@ -561,97 +662,148 @@ func (m *Manager) handleChangeProofResponse(
...
-		return fmt.Errorf(
+		return nil, fmt.Errorf(
 			"%w: %T",
 			errUnexpectedChangeProofResponse, changeProofResp,
 		)
 	}
+}
 
-	return nil
+type changeProofContext struct {
```
joshua-kim: Same comments as from the range proof types on these types as well.

x/sync/manager.go
```diff
@@ -510,43 +541,113 @@ func (m *Manager) handleRangeProofResponse(
...
+}
+
+type rangeProofContext struct {
+	rootHash []byte
+	startKey maybe.Maybe[[]byte]
+	endKey   maybe.Maybe[[]byte]
+	keyLimit int
+}
+
+type rangeProof struct {
```
joshua-kim: nit: Can you move these types closer to where their functions are defined i.e above where Verify/Commit/FindNextKey are?

x/sync/manager.go
```diff
@@ -561,97 +662,148 @@ func (m *Manager) handleChangeProofResponse(
...
+	// If this is empty, then we have no more keys to fetch.
+	if !largestHandledKey.IsNothing() {
+		nextKey, err := findNextKey(ctx, p.db, largestHandledKey.Value(), p.request.endKey, p.merkleProof.EndProof, p.tokenSize)
+		if err != nil {
+			return maybe.Nothing[[]byte](), err
+		}
+		largestHandledKey = nextKey
+	}
+
+	return largestHandledKey, nil
```
joshua-kim: Since range proofs are a case of change proofs it seems like we should be able to refactor this implementation w/ the new range proof code we're introducing in this PR.

x/sync/sync_test.go
```diff
@@ -187,6 +187,9 @@ func Test_Sync_FindNextKey_InSync(t *testing.T) {
 	require.NoError(err)
 	require.NotNil(syncer)
 
+	proofParser, ok := syncer.proofParser.(*merkleDBProofParser)
+	require.True(ok)
+
```
joshua-kim: Remove diff - test is introspecting into implementation details. As long as the sync occurs correctly the caller of this package shouldn't care about its type - especially because the type isn't exported anyways.

x/sync/manager.go
alarso16: Good to know! I noticed some of these, so I'll try and do that in the future too

x/sync/manager.go
alarso16: The syncer doesn't use this - it's only used in the proof logic

x/sync/manager.go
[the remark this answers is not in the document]
alarso16: Actually, without refactoring `merkleProof` this isn't very easy...

x/sync/sync_test.go
alarso16: These tests were already very coupled with the unexported code. The options are:
- Modify the tests to specifically test the proofParser/Proof interface, without the syncer
- Reconstruct the proof using the unexported proofParser and recreating the requests

x/sync/manager.go
```diff
@@ -63,6 +67,29 @@ const (
...
 )
 
+type proofParser struct {
+	db           DB
+	hasher       merkledb.Hasher
+	branchFactor merkledb.BranchFactor
+	tokenSize    int
+}
+
+func newParser(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor) (ProofParser, error) {
```
rrazvan1: maybe replace **DB** type with **merkledb.MerkleDB**?
So its more explicitly a parser implementation for merkleDB.
And also, we can add a getter for Hasher and BranchFactor, because with the current approach you can provide a different hasher/branchFactor than the actual one from the DB.

x/sync/manager.go
```diff
@@ -167,12 +196,10 @@ func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manage
 	case config.SimultaneousWorkLimit == 0:
 		return nil, ErrZeroWorkLimit
 	}
-	if err := config.BranchFactor.Valid(); err != nil {
-		return nil, err
-	}
 
-	if config.Hasher == nil {
-		config.Hasher = merkledb.DefaultHasher
+	parser, err := newParser(config.DB, config.Hasher, config.BranchFactor)
```
rrazvan1: **NewManager** function shouldn't be linked to a specific parser implementation.
What about moving the **Parser** inside **ManagerConfig**?

x/sync/manager.go
```diff
@@ -561,97 +654,148 @@ func (m *Manager) handleChangeProofResponse(
...
-		return fmt.Errorf(
+		return nil, fmt.Errorf(
 			"%w: %T",
 			errUnexpectedChangeProofResponse, changeProofResp,
 		)
 	}
+}
 
-	return nil
+type changeProofRequest struct {
```
rrazvan1: This is the same as **rangeProofRequest**.
Keep only one, **proofRequest**?

x/sync/manager.go
```diff
@@ -510,43 +533,113 @@ func (m *Manager) handleRangeProofResponse(
...
+func (r *rangeProof) Commit(ctx context.Context) error {
+	if len(r.merkleProof.KeyChanges) == 0 {
+		return nil
+	}
+	return r.db.CommitRangeProof(ctx, r.request.startKey, r.request.endKey, r.merkleProof)
+}
 
-	if len(rangeProof.KeyChanges) > 0 {
-		largestHandledKey = maybe.Some(rangeProof.KeyChanges[len(rangeProof.KeyChanges)-1].Key)
+func (r *rangeProof) FindNextKey(ctx context.Context) (maybe.Maybe[[]byte], error) {
```
rrazvan1: Duplicate function as **changeProof.FindNextKey**

x/sync/manager.go
```diff
@@ -561,97 +654,148 @@ func (m *Manager) handleChangeProofResponse(
...
+	}
+	nextKey, err := proof.FindNextKey(ctx)
+	if err != nil {
+		m.setError(err)
 		return err
 	}
 
-	startKey := maybeBytesToMaybe(request.StartKey)
-	endKey := maybeBytesToMaybe(request.EndKey)
+	m.completeWorkItem(work, nextKey, targetRootID)
```
rrazvan1, line 637: **completeWorkItem** is expecting **largestHandledKey** as second parameter, not **nextKey**.

In some scenarios like **targetRoot** being changed while syncer is running, we could requeue the **workItem**, with a different **end**, than the one which was previously used. https://github.com/ava-labs/avalanchego/blob/50f0e507c03e8ef401b45d39f6d78fe005ad4202/x/sync/manager.go#L1101

x/sync/manager.go
alarso16: Yeah it is, but it's because this part references fields in the struct. This requires upstream (i.e. merkledb changes) to refacotr, or passing even more values into the `findNextKey` function

x/sync/manager.go
alarso16: `largestHandledKey <= nextKey`, so this works as well. Additionally, just in case there is later a key `x` added in the range `largestHandledKey < x < nextKey`, we want to be able to query it later

x/sync/manager.go
alarso16: This is probably out of scope for the PR, right?. I'm already planning to have a follow-up completely removing merkledb as a dependency for the package, so maybe it makes more sense there

x/sync/manager.go
alarso16: Ok done

x/sync/manager.go
alarso16: Done

x/sync/manager.go
alarso16: It's still in the parser intialization

x/sync/manager.go
```diff
@@ -327,25 +341,11 @@ func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
 
 	if work.localRootID == targetRootID {
 		// Start root is the same as the end root, so we're done.
-		m.completeWorkItem(ctx, work, work.end, targetRootID, nil)
+		m.completeWorkItem(work, work.end, targetRootID)
 		m.finishWorkItem()
 		return
 	}
 
-	if targetRootID == ids.Empty {
```
alarso16, line 335: This was the only functional thing that required the database. To resolve this, it will send one request to the network, the network will return an empty proof, and the proof will handle the empty root case.

x/sync/manager.go
```diff
@@ -848,17 +975,7 @@ func (m *Manager) Wait(ctx context.Context) error {
 		return err
 	}
 
-	root, err := m.config.DB.GetMerkleRoot(ctx)
-	if err != nil {
-		return err
-	}
-
-	if targetRootID := m.getTargetRoot(); targetRootID != root {
```
alarso16, line 856: This can be handled by the consumer - I don't think we should provide the database to the manager

x/sync/manager.go
```diff
@@ -926,36 +1042,14 @@ func (m *Manager) setError(err error) {
 // If [workItem.start] is Nothing, then we've fetched all the key-value
 // pairs up to and including [largestHandledKey].
 //
-// If [largestHandledKey] is Nothing, then we've fetched all the key-value
-// pairs at and after [workItem.start].
-//
-// [proofOfLargestKey] is the end proof for the range/change proof
-// that gave us the range up to and including [largestHandledKey].
+// If [nextKey] is Nothing, then we've fetched all the key-value
+// pairs in the trie with root [rootID].
```
rrazvan1:
```suggestion
// If [nextKey] is Nothing, then we've fetched all the key-value
// pairs in the trie with root [rootID] with maximum key [work.end].
```

x/sync/manager.go
```diff
@@ -666,12 +786,18 @@ func (m *Manager) handleChangeProofResponse(
 //
 // Invariant: [lastReceivedKey] < [rangeEnd].
 // If [rangeEnd] is Nothing it's considered > [lastReceivedKey].
-func (m *Manager) findNextKey(
+func findNextKey(
```
rrazvan1, line 797: this function says:
`// Invariant: [lastReceivedKey] < [rangeEnd].`

With the new code, I think this isnt actually true anymore.
We should check this before calling findNextKey, or remove the invariant and, if last key of [keyChanges], is the same as [rangeEnd], we can early return **Nothing**.

x/sync/manager.go
alarso16: This isn't true. I changed this :)

`findNextKey` won't return larger than `work.end`, but `proof.FindNextKey(ctx)` could! So with this implementation, your comment is true. However, once Firewood is used, it will not be true

x/sync/manager.go
alarso16: Done

x/sync/manager.go
alarso16: There is a small amount intricacy here about whether `nextKey` should be included in the completed work queue. IMO it's ok, even though the key is not yet found, we can treat the interval as open when queuing, which should handle every case. Additionally, the work queue will still properly merge these sets once the next proof is returned

x/sync/manager.go
```diff
@@ -664,14 +784,25 @@ func (m *Manager) handleChangeProofResponse(
...
 	rangeEnd maybe.Maybe[[]byte],
 	endProof []merkledb.ProofNode,
+	tokenSize int,
 ) (maybe.Maybe[[]byte], error) {
+	lastReceivedKey := rangeEnd.Value()
+	if len(keyChanges) > 0 {
+		lastReceivedKey = keyChanges[len(keyChanges)-1].Key
+
+		// If the proof provides keys after the range end, we can return that the range was complete.
+		if compareKeys(maybe.Some(lastReceivedKey), rangeEnd) > 0 {
```
rrazvan1: Nothing value should also be returned for **lastReceivedKey** being equal to **rangeEnd**
**lastReceivedKey** being the same as **rangeEnd** happens more often.
Doesn't make sense remote nodes sending **keyChanges** outside of interval.

alarso16: It doesn't make sense, but they could, right?

alarso16: yeah that seems reasonable - I fixed it

x/sync/manager.go
```diff
@@ -926,36 +1061,14 @@ func (m *Manager) setError(err error) {
...
-		// nextStartKey being Nothing indicates that the entire range has been completed
-		if nextStartKey.IsNothing() {
-			largestHandledKey = work.end
-		} else {
-			// the full range wasn't completed, so enqueue a new work item for the range [nextStartKey, workItem.end]
-			m.enqueueWork(newWorkItem(work.localRootID, nextStartKey, work.end, work.priority, time.Now()))
-			largestHandledKey = nextStartKey
-		}
+func (m *Manager) completeWorkItem(work *workItem, nextKey maybe.Maybe[[]byte], rootID ids.ID) {
+	if compareKeys(work.end, nextKey) == 1 && !nextKey.IsNothing() {
```
rrazvan1:
```suggestion
	if compareKeys(work.end, nextKey) >= 0 && !nextKey.IsNothing() {
```
**nextKey** is the next key with different/missing value. So even if it is the same as **work.end**, we need to process it, in order for the current work range to be considered completed.

x/sync/manager.go
rrazvan1: I am not happy with this, I would still be trying to get the **largestKey** from the proof.
By using the **nextKey** instead of **largestKey** we are making the ranges bigger: both for requeuing and marking the work processed.

For example, if you continuously pivot (with the right timing), you can enlarge the work range forever, because you are going to get a larger and larger **nextKey**, based on the **largestKey** (which is the previous **nextKey**).

What do you think about this?

x/sync/network_server.go
```diff
@@ -111,6 +111,22 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 		}
 	}
 
+	// If the end root is empty, the proof will delete everything.
+	if endRoot == ids.Empty {
+		proofBytes, err := proto.Marshal(&pb.SyncGetChangeProofResponse{
+			Response: &pb.SyncGetChangeProofResponse_ChangeProof{
+				ChangeProof: (&merkledb.ChangeProof{}).ToProto(),
+			},
+		})
```
rrazvan1: If someone is requesting a **changeProof** from: `rootA -> ids.Empty`, shouldn't that **changeProof** actually contains ALL the keys from **rootA** with **Nothing** value?

To avoid doing this, maybe we should have your approach (of returning an empty response), but returning an empty **RangeProof**.

```suggestion
		proofBytes, err := proto.Marshal(&pb.SyncGetChangeProofResponse{
			Response: &pb.SyncGetChangeProofResponse_RangeProof{
				RangeProof: (&merkledb.RangeProof{}).ToProto(),
			},
		})
```

alarso16: We can do that

x/sync/manager.go
```diff
@@ -561,97 +630,160 @@ func (m *Manager) handleChangeProofResponse(
...
+	}
+
+	// Find the next key to fetch.
+	// If we wanted the empty root, we don't need to fetch any keys.
+	if bytes.Equal(c.request.rootHash, ids.Empty[:]) {
+		return maybe.Nothing[[]byte](), nil
+	}
+
+	// This will return Nothing if there are no more keys to fetch in the range `largestHandledKey` to `c.request.endKey`.
+	nextKey, err := findNextKey(ctx, c.db, c.merkleProof.KeyChanges, c.request.endKey, c.merkleProof.EndProof, c.tokenSize)
```
rrazvan1: if **c.request.endKey** is **Nothing** and **c.merkleProof.KeyChanges** is empty (which implies **c.merkleProof.EndProof** being also empty) **findNextKey** will return **0x00** bcs:

```
    lastReceivedKey := rangeEnd.Value()
	if len(keyChanges) > 0 {
		lastReceivedKey = keyChanges[len(keyChanges)-1].Key
	}

	if len(endProof) == 0 {
		// We try to find the next key to fetch by looking at the end proof.
		// If the end proof is empty, we have no information to use.
		// Start fetching from the next key after [lastReceivedKey].
		nextKey := lastReceivedKey
		nextKey = append(nextKey, 0)
		return maybe.Some(nextKey), nil
	}
```
Maybe add a check before calling **findNextKey**, and early return **Nothing** (both for rangeproof and changeproof).

alarso16: This is really weird actually, but I don't think that's possible. Check this snippet below, where `db` is empty
```
	rangeProof, err := db.GetRangeProof(ctx, maybe.Some([]byte{0xff}), maybe.Nothing[[]byte](), 10)
	require.NoError(err)
```
This fails with `proof is empty`. So we wouldn't ever receive a proof structured like this

rrazvan1: I dont think x/sync should assume anything about remote nodes.
But we can do that later in other PR.

x/sync/manager.go
```diff
@@ -471,11 +459,7 @@ func (m *Manager) retryWork(work *workItem) {
 }
 
 // Returns an error if we should drop the response
-func (m *Manager) shouldHandleResponse(
-	bytesLimit uint32,
-	responseBytes []byte,
-	err error,
-) error {
+func (m *Manager) shouldHandleResponse(bytesLimit uint32, responseBytes []byte, err error) error {
```
joshua-kim: 1. Undo unrelated diff
2. Line width we try to enforce (although it's not very consistent) is 80 chars, so even if we kept this diff it would diverge from what we try to keep the repo style at

x/sync/network_server_test.go
```diff
@@ -108,8 +108,8 @@ func Test_Server_GetRangeProof(t *testing.T) {
 				KeyLimit:   defaultRequestKeyLimit,
 				BytesLimit: defaultRequestByteSizeLimit,
 			},
-			proofNil:    true,
-			expectedErr: p2p.ErrUnexpected,
+			proofNil:    false,
+			expectedErr: nil,
```
joshua-kim: nit: these are not needed since the zero values will be `false` and `nil` respectively

x/sync/manager.go
```diff
@@ -561,97 +630,160 @@ func (m *Manager) handleChangeProofResponse(
...
+		c.request.startKey,
+		c.request.endKey,
+		endRoot,
+	)
+}
+
+func (c *changeProof) Commit(ctx context.Context) error {
+	if c.db == nil {
+		return errNeedDBForOperation
+	}
```
joshua-kim: Is this even possible to hit + in `rangeProof`? Should we remove these `nil` checks and just panic if the provided db is `nil`?

x/sync/db.go
```diff
@@ -12,3 +17,31 @@ type DB interface {
...
+	// is applied. If there are no more keys to request in the range provided,
+	// it returns Nothing.
+	// The proof must be committed first.
+	FindNextKey(context.Context) (maybe.Maybe[[]byte], error)
+}
+
+type ProofParser interface {
+	ParseRangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
+	ParseChangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
+}
```
joshua-kim: I kind of wonder if we can simplify some of these interfaces... I think it's good for us to have separate signatures to signal fatal/non-fatal errors like we're doing in the current PR, but I'm wondering if we need to explicitly define things like `Verify/Commit/FindNextKey`. To me:

1. Parse* seems to overlap somewhat with the idea of `Verify`, since we're taking a request and dropping it if it's invalid (parse cares about the bytes being valid or not, verify cares about the actual data being valid or not, but if we just broaden the boundary to be on the request being valid or not it seems reasonable. 
2. Commit/FindNextKey seem coupled, we can't proceed syncing unless we know the next key so returning the next key to sync as part of `Commit` seems reasonable to me.
3. I _think_ the implementation applying proofs just cares about the key changes... so I think we could probably simplify the `Proof` interface as well.

```suggestion
type Proof interface {
    KeyChanges() []KeyChange
}

type ProofHandler[T Proof, U Proof] interface {
    ParseRangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (T, error)
    ParseChangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (U, error)
	
    Apply(context.Context, keyChanges []KeyChange) (maybe.Maybe[[]byte], error)
}
```

x/sync/manager.go
```diff
@@ -63,6 +67,28 @@ const (
 	retryPriority
 )
 
+type proofParser struct {
+	db        DB
+	hasher    merkledb.Hasher
+	tokenSize int
+}
+
+func newParser(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor) (ProofParser, error) {
```
joshua-kim: 1. nit: return the concrete type - we lose type information if we return the interface (although it makes no difference in this context). 
2. Is this expected? We're not calling this anywhere other than tests - so I'm not sure how we would initialize this parser

x/sync/manager.go
```diff
@@ -63,6 +67,28 @@ const (
...
+	if err := branchFactor.Valid(); err != nil {
+		return nil, err
+	}
+
+	if hasher == nil {
+		hasher = merkledb.DefaultHasher
+	}
```
joshua-kim, line 85: This config belongs to the syncer - so I think this config validation/override belongs there instead. I think if we use a different parser implementation we would get unexpected behavior since we wouldn't do this validity check/override.

alarso16: I think this information really belongs to the DB - it's not used anywhere in the syncer code, so it doesn't make sense to me that it would belong to the syncer

x/sync/manager.go
alarso16: The argument for this is that you can verify range proofs without a DB, but I agree - I'll move this check into parser initialization

x/sync/db.go
alarso16: We can do something like this. In merkledb, the implementation of applying a proof does only depend on the key changes, but it may be impractical for other use cases (e.g. Firewood). I'll change this so the only operation on a `Proof` is `Commit`

x/sync/manager.go
alarso16: Is what expected? It's only called in tests because x/sync is only used in tests...

x/sync/manager.go
alarso16: Moved check that db is nil to parser creation

x/sync/manager.go
joshua-kim: How would you create an instance of `Manager`? You can't create an instance of `ProofParser` if you're using this package because `newParser` isn't exposed to you right?

alarso16: Hmm I guess... I'll export it

x/sync/db.go
joshua-kim: 1. I don't know about how firewood is using it - but we shouldn't prematurely abstract if we can get away with less. Is firewood actually using anything other than the key changes?
2. If `Commit` lives on `Proof`, then the implementation of `Proof` needs to know about the db and manage a reference to it - which I find to be a confusing abstraction (why does a proof hold a reference to the db)?. I would recommend having the `ProofHandler` know how to apply proofs since a proof is just some data. If we wanted the implementation of `Proof` to implement `Commit` - an improvement might be to make it look something like:
```
Commit(ctx context.Context, db DB) (maybe.Maybe[[]byte], error)
```
Since now you don't have to worry about the db reference throughout the implementation's lifetime, but this api looks worse to me since the proof still has to know about the db.
3. If we don't use generics, the i/f signature of 
```
    ParseRangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
    ParseChangeProof(responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (Proof, error)
```
weirds weirdly to me - since it's not clear to me how the two functions differ at all. I think we can still use generics despite the fact that a change proof returns a range or change proof, since the `Proof` interface just defines what operations to perform on the db so I believe this interface still works.

alarso16: 1. Firewood doesn't use anything other than the key changes to commit, but the issue relies mostly in FFI complexity. We pass to Rust the bytes representing the proof, it's parsed into some internal Rust item, and C-style pointer is passed back to Go. Now, to get the keys and values from this struct, we need to return two arrays of byte slices in Rust managed memory, copy each into a Go slice, free the Rust arrays, just to pass these back into Rust. I think this would be quite blocking for Firewood.
2. Firewood shouldn't have to follow the same schema as merkledb - so what would `DB` even be? I agree that the abstraction is not necessarily the most intuitive, but I have no idea how something like this could be constructed. What would the implementation of `Commit` look like?
3. This feels more like a flaw in merkledb handling. I agree it's unnecessary and I would love to fix it (since in Firewood these functions will be the same I think). For some reason unknown to me, the response types of range proof request and change proof request are different, even though they can both be range proofs. Should I fix that in this PR too? I can simplify this in a different PR

joshua-kim: 2. Isn't that implemented by firewood? The difference would be moving where `Commit` lives off of the `Proof` type
3. This is an unrelated change and I don't want us to scope creep - the suggestion I'm making is to just define the concrete type w/ a generic and it should be possible to do. If we define T as a range proof and U as a change proof (which internally is a range or change proof), I think this still works

x/sync/manager.go
joshua-kim: Ah - I see we moved it out of the syncer config. I think this is fine now.

x/sync/manager.go
joshua-kim: We're still returning the interface instead of the concrete type here it looks like.

demosdemon:
I'm not sure I understand the full intent here. You're passing in a grpc response bytes to a generic parse. But, grpc is pretty strict that the return type of <https://github.com/ava-labs/avalanchego/blob/a6ee47b191efbc723df00b93713314a97ac6e38b/x/sync/manager.go#L385-L389> will be a serialized protobuf GetChangeProofResponse <https://github.com/ava-labs/avalanchego/blob/e03af8404fb056826f1909f13492c4184ec26c06/proto/sync/sync.proto#L20>.

This interface should likely be generic over the grpc methods themselves otherwise these changes are irrelevant.

---

edit: realized we're not using the grpc interface and the response structure isn't required to be a protobuf value

x/sync/manager.go
```diff
@@ -63,6 +66,31 @@ const (
 	retryPriority
 )
 
+type proofHandler struct {
+	db        DB
+	hasher    merkledb.Hasher
+	tokenSize int
+}
+
+func NewHandler(db DB, hasher merkledb.Hasher, branchFactor merkledb.BranchFactor) (ProofHandler[*rangeProofResponse, *changeProofResponse], error) {
```
demosdemon, line 75: This doesn't feel very generic. And I don't understand how users outside of this package are supposed to use _this_ ProofHandler or any manager associated with it if they cannot refer to the generic types for range and change proofs. 

I can create the new handler and I can create the manager but I cannot store the manager on [another type](https://github.com/ava-labs/hypersdk/blob/d6a972662fba5dc2c0213f9067796b21c9fccc0b/statesync/merkledb_adapter.go#L36).

x/sync/db.go
```diff
@@ -12,3 +17,10 @@ type DB interface {
 	merkledb.ChangeProofer
 	merkledb.RangeProofer
 }
+
+type ProofHandler[T, U any] interface {
+	ParseRangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (T, error)
+	ParseChangeProof(ctx context.Context, responseBytes, rootHash []byte, startKey, endKey maybe.Maybe[[]byte], keyLimit uint32) (U, error)
+	CommitRangeProof(ctx context.Context, proof T) (maybe.Maybe[[]byte], error)
+	CommitChangeProof(ctx context.Context, proof U) (maybe.Maybe[[]byte], error)
+}
```
demosdemon, line 26: looking at the `DB` interface. it feels like `DB` should be made generic over the types of proofs it returns instead of this additional interface that does the same thing:

https://github.com/ava-labs/avalanchego/blob/66ca7dc0ce354ff4c4ebdc8b377e099ba91f2e2b/x/merkledb/db.go#L67-L130

firewood is unlikely to implement the three proof interfaces if we also need this additional interface

demosdemon: What if we made DB look like this


```go
type ProofMarshaller interface {
	MarshalProofBinary() ([]byte, error)
	UnmarshalProofBinary(data []byte) error
}

type MerkleDB = DB[*merkledb.Proof, *merkledb.ChangeProof, *merkledb.RangeProof]

type DB[TKeyProof, TChangeProof, TRangeProof ProofMarshaller] interface {
	merkledb.Clearer
	merkledb.MerkleRootGetter
	ProofGetter[TKeyProof]
	ChangeProofer[TChangeProof]
	RangeProofer[TRangeProof]
}

type ProofGetter[T any] interface {
	GetProof(ctx context.Context, keyBytes []byte) (T, error)
}

type ChangeProofer[T any] interface {
	GetChangeProof(
		ctx context.Context,
		startRootID ids.ID,
		endRootID ids.ID,
		start maybe.Maybe[[]byte],
		end maybe.Maybe[[]byte],
		maxLength int,
	) (T, error)

	VerifyChangeProof(
		ctx context.Context,
		proof T,
		start maybe.Maybe[[]byte],
		end maybe.Maybe[[]byte],
		expectedEndRootID ids.ID,
	) error

	CommitChangeProof(ctx context.Context, proof T) error
}

type RangeProofer[T any] interface {
	GetRangeProofAtRoot(
		ctx context.Context,
		rootID ids.ID,
		start maybe.Maybe[[]byte],
		end maybe.Maybe[[]byte],
		maxLength int,
	) (T, error)

	CommitRangeProof(ctx context.Context, start, end maybe.Maybe[[]byte], proof T) error
}
```

With this change, merkledb as is doesn't need any changes except to implement ProofMarshaller.

```patch
diff --git a/x/merkledb/proof.go b/x/merkledb/proof.go
index 23f7e520d..2b1854af7 100644
--- a/x/merkledb/proof.go
+++ b/x/merkledb/proof.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/avalanchego/trace"
 	"github.com/ava-labs/avalanchego/utils/maybe"
+	"github.com/gogo/protobuf/proto"
 
 	pb "github.com/ava-labs/avalanchego/proto/pb/sync"
 )
@@ -236,6 +237,18 @@ func (proof *Proof) UnmarshalProto(pbProof *pb.Proof) error {
 	return nil
 }
 
+func (proof *Proof) MarshalProofBinary() ([]byte, error) {
+	return proto.Marshal(proof.ToProto())
+}
+
+func (proof *Proof) UnmarshalProofBinary(data []byte) error {
+	var pbProof pb.Proof
+	if err := proto.Unmarshal(data, &pbProof); err != nil {
+		return err
+	}
+	return proof.UnmarshalProto(&pbProof)
+}
+
 type RangeProof ChangeProof
 
 func (proof *RangeProof) ToProto() *pb.RangeProof {
@@ -294,6 +307,18 @@ func (proof *RangeProof) UnmarshalProto(pbProof *pb.RangeProof) error {
 	return nil
 }
 
+func (proof *RangeProof) MarshalProofBinary() ([]byte, error) {
+	return proto.Marshal(proof.ToProto())
+}
+
+func (proof *RangeProof) UnmarshalProofBinary(data []byte) error {
+	var pbProof pb.RangeProof
+	if err := proto.Unmarshal(data, &pbProof); err != nil {
+		return err
+	}
+	return proof.UnmarshalProto(&pbProof)
+}
+
 // Validate received data from change/range proof requests
 // using the requested range.
 func validateChangeProof(
@@ -516,6 +541,18 @@ func (proof *ChangeProof) UnmarshalProto(pbProof *pb.ChangeProof) error {
 	return nil
 }
 
+func (proof *ChangeProof) MarshalProofBinary() ([]byte, error) {
+	return proto.Marshal(proof.ToProto())
+}
+
+func (proof *ChangeProof) UnmarshalProofBinary(data []byte) error {
+	var pbProof pb.ChangeProof
+	if err := proto.Unmarshal(data, &pbProof); err != nil {
+		return err
+	}
+	return proof.UnmarshalProto(&pbProof)
+}
+
 // Verifies that the given [proofNodes]:
 // - if the node's key is within the key range, that has a value that matches the value passed in the change list or in the db.
 func verifyChangeProofKeyValues(ctx context.Context, db *merkleDB, keyChanges []KeyChange, proofNodes []ProofNode, start maybe.Maybe[Key], end maybe.Maybe[Key], hasher Hasher) error {
```

x/sync/manager.go
alarso16: They will NOT use this proof handler. They will create their own. A separate PR will move everything related merkledb into a different package. Any consumer will create their own `ProofHandler`

x/sync/db.go
alarso16: One small intricacy here is around `FindNextKey`. That is the only consumer of `GetProof`, but requires tracing the proof. Likely, if we took this route, we would remove the `GetProof` API and require `CommitXXXProof` to return `(maybe.Maybe[[]byte], error)`, so the proof implementation doesn't have to be exposed

alarso16: Additionally, we would still need some weird handling for when range proofs are returned from a change proof request. Because of this, we will have to wrap the `*ChangeProof` type with a `*RangeOrChangeProof`

alarso16:
Closing in favor of a different design. Hopefully third time's the charm...

PR #4158, opened 2025-08-05
refactor: use ProofCreator

alarso16:
## Why this should be merged

This follows the DB decoupling established in #4155, separating request validity from proof formation.

Next steps are similar to the other PR as well:
- Move creator to separate package and provide instead of `DB`
- Separate testing implementations

## How this works

Moves all proof creation and retries to a separate interface, wrapped in creation.

## How this was tested

Existing UT

No

### DIFF
--- a/x/sync/client_test.go
+++ b/x/sync/client_test.go
@@ -39,10 +39,10 @@ func newDefaultDBConfig() merkledb.Config {
 
 func newFlakyRangeProofHandler(
 	t *testing.T,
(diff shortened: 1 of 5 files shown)


alarso16:
This is unnecessary. If more refactoring in the server side is necessary, a more creative method will be taken

alarso16:
This is unnecessary. If more refactoring in the server side is necessary, a more creative method will be taken

PR #4159, opened 2025-08-06
Simplex Block Builder Component

samliok:
## Why this should be merged

This PR introduces the `BlockBuilder` component for Simplex. The `BlockBuilder` is responsible for building blocks when requested by consensus as well as notifying consensus a block is ready to be built. 

## How this works

The `BlockBuilder`struct has two methods
1. `BuildBlock` builds a block when requested by a simplex node.
2. `IncomingBlock` notifies consensus that a block is ready to be built. If no blocks are proposed in Simplex, either the Simplex node is faulty(or offline), or there are no blocks to propose. Therefore when IncomingBlock notifies consensus, our node should expect a block proposal. If no proposals arrive within a certain time, our node sends an empty vote and attempts to proceed to the next round.


simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+	blockTracker *blockTracker
+}
+
+// BuildBlock continuously tries to build a block until the context is cancelled. If there are no blocks to be built, it will wait for an event from the VM.
+// It returns false if the context was cancelled, otherwise it returns the built block and true.
+func (b *BlockBuilder) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata) (simplex.VerifiedBlock, bool) {
+	for {
+		select {
+		case <-ctx.Done():
+			b.log.Info("Context cancelled, stopping block building")
```
yacovm: Can we adjust this to debug level?

simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+			return nil, false
+		default:
+			err := b.incomingBlock(ctx)
+			if err != nil {
+				b.log.Error("Error waiting for incoming block:", zap.Error(err))
+				continue
+			}
+			vmBlock, err := b.vm.BuildBlock(ctx)
```
yacovm: I think we should have some kind of exponential backoff here, otherwise we're going to do a busy-wait if the VM is misbehaving.

simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+			if err != nil {
+				b.log.Error("Error building block:", zap.Error(err))
+				continue
+			}
+			simplexBlock, err := newBlock(metadata, vmBlock, b.blockTracker)
```
yacovm: if we fail here it's because we have failed serializing the block. I think we should return `_, false` here.

simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+			if err != nil {
+				b.log.Error("Error creating simplex block:", zap.Error(err))
+				continue
+			}
+
+			verifiedBlock, err := simplexBlock.Verify(ctx)
```
yacovm: we should probably add an exponential backoff here too in case we cannot verify the block we've built.

simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+func (b *BlockBuilder) incomingBlock(ctx context.Context) error {
+	for {
+		msg, err := b.vm.WaitForEvent(ctx)
+		if err != nil {
+			return err
+		}
+		if msg == common.PendingTxs {
+			return nil
+		}
+		b.log.Warn("Received unexpected message", zap.String("message", msg.String()))
```
yacovm: Let's add an exponential backoff sleep here, so we won't hog the CPU.

simplex/block_builder.go
```diff
@@ -0,0 +1,79 @@
...
+func (b *BlockBuilder) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata) (simplex.VerifiedBlock, bool) {
+	for {
+		select {
+		case <-ctx.Done():
+			b.log.Info("Context cancelled, stopping block building")
+			return nil, false
+		default:
+			err := b.incomingBlock(ctx)
+			if err != nil {
+				b.log.Error("Error waiting for incoming block:", zap.Error(err))
```
yacovm: if Simplex has cancelled the context, then it's not something we should be ashamed of. 
So I don't think we should print an error.

simplex/block_builder_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			count := 0
+			vm := newTestVM()
+
+			vm.WaitForEventF = func(_ context.Context) (common.Message, error) {
+				count++
+				time.Sleep(20 * time.Millisecond)
```
yacovm: why do we sleep here? is this essential?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			timeoutCtx, cancelCtx := context.WithTimeout(ctx, 100*time.Millisecond)
+			defer cancelCtx()
+
+			block, built := bb.BuildBlock(timeoutCtx, child.BlockHeader().ProtocolMetadata)
+			require.Equal(t, tt.shouldBuild, built)
+			if tt.expectedBlock == nil {
+				require.Nil(t, block, "Block should be nil when not built")
+				require.GreaterOrEqual(t, count, 1)
```
yacovm: are we checking here that we tried to build a block?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			vm.BuildBlockF = tt.vmBlockBuildF
+
+			bb := &BlockBuilder{
+				log:          logging.NoLog{},
+				vm:           vm,
+				blockTracker: genesis.blockTracker,
+			}
+
+			timeoutCtx, cancelCtx := context.WithTimeout(ctx, 100*time.Millisecond)
```
yacovm, line 82: why do we need the timeout context here? Can't we use a background context?

samliok: no we need `WithContext` since `BuildBlock` only returns if the context is cancelled or a block is built. Some of our tests above fail to build a block so we need to cancel the context to return.

simplex/block_builder_test.go
samliok: this check should ensure we call `WaitForEvent` at least twice. Since when we fail we want to check if we try again. 

i need to update the require to be `Greater` not `GreaterOrEqual`

simplex/block_builder_test.go
samliok: initially did this so we wouldn't over call `WaitForEvent`, but this isn't needed with the exponential backoff stuff

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+				curWait = backoff(ctx, curWait, maxBackoff)
+				continue
+			}
+			vmBlock, err := b.vm.BuildBlock(ctx)
+			if err != nil {
+				b.log.Error("Error building block:", zap.Error(err))
+				curWait = backoff(ctx, curWait, maxBackoff)
+				continue
+			}
+			simplexBlock, err := newBlock(metadata, vmBlock, b.blockTracker)
```
yacovm: if we successfully build the block, we need to reset the `curWait` to 10 millis, no?

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				b.log.Error("Error creating simplex block:", zap.Error(err))
+				return nil, false
+			}
+			verifiedBlock, err := simplexBlock.Verify(ctx)
+			if err != nil {
+				b.log.Error("Error verifying block we built ourselves: %s", zap.Error(err))
```
yacovm: I am not sure this should be error... @StephenButtolph - how often is it that we build a block that we fail in verifying? 

Can't we have some transaction that depends on P-chain advancing (or not advancing) and right after we build the block the P-chain advanced and now the transaction is invalid?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
```
yacovm, line 19: can we please add a test for the exponential backoff?

simplex/block_builder_test.go
```diff
@@ -140,3 +140,39 @@ func TestIncomingBlock(t *testing.T) {
...
+	timeoutCtx, cancelCtx := context.WithTimeout(ctx, time.Second)
+	defer cancelCtx()
+
+	block, built := bb.BuildBlock(timeoutCtx, child.BlockHeader().ProtocolMetadata)
+	require.False(t, built)
+	require.Nil(t, block)
+
+	// expected number of count increases
+	// 10, 20, 40, 80, 160, 320, 640 = 7 attempts
```
yacovm, line 176: I guess these attempts are due to the timeout? In that case, I think it's better to just control the amount of retries in the BuildBlockF itself (based on the count), otherwise if the CI runner is overloaded we might have less retries because of lag.... thoughts?

yacovm (approved):

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+func (b *BlockBuilder) incomingBlock(ctx context.Context) error {
+	for {
+		msg, err := b.vm.WaitForEvent(ctx)
+		if err != nil {
+			return err
+		}
+		if msg == common.PendingTxs {
+			return nil
+		}
+		b.log.Info("Received unexpected message", zap.String("message", msg.String()))
```
StephenButtolph:
```suggestion
		b.log.Info("Received unexpected message", zap.Stringer("message", msg))
```

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+// IncomingBlock blocks until a new block is ready to be built from the VM, or until the
+// context is cancelled.
+func (b *BlockBuilder) IncomingBlock(ctx context.Context) {
+	err := b.incomingBlock(ctx)
+	if err != nil {
+		b.log.Debug("Error waiting for incoming block", zap.Error(err))
+	}
+}
+
+func (b *BlockBuilder) incomingBlock(ctx context.Context) error {
```
StephenButtolph:
```suggestion
func (b *BlockBuilder) waitForPendingTxs(ctx context.Context) error {
```

simplex/block_builder.go
StephenButtolph: Should this be a warn?
```suggestion
		b.log.Warn("Received unexpected message", zap.String("message", msg.String()))
```

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			return verifiedBlock, true
+		}
+	}
+}
+
+// IncomingBlock blocks until a new block is ready to be built from the VM, or until the
+// context is cancelled.
+func (b *BlockBuilder) IncomingBlock(ctx context.Context) {
```
StephenButtolph: While I don't necessarily think this function should be called `WaitForPendingTxs`, I don't like `IncomingBlock`. imo this name is not descriptive (or even really accurate) for what it is doing.

Would it make sense in the simplex context to be called `WaitForBuildBlockAvailable`? Open to less verbose suggestions.

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+func (b *BlockBuilder) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata) (simplex.VerifiedBlock, bool) {
+	curWait := 10 * time.Millisecond
+	maxBackoff := 5 * time.Second
+
+	for {
+		select {
+		case <-ctx.Done():
+			b.log.Debug("Context cancelled, stopping block building")
+			return nil, false
+		default:
```
StephenButtolph:
```suggestion
		if ctx.Err() != nil {
			b.log.Debug("Context cancelled, stopping block building")
			return nil, false
		}
```

Then the `default` indentation can be reduced

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+	blockTracker *blockTracker
+}
+
+// BuildBlock continuously tries to build a block until the context is cancelled. If there are no blocks to be built, it will wait for an event from the VM.
+// It returns false if the context was cancelled, otherwise it returns the built block and true.
+func (b *BlockBuilder) BuildBlock(ctx context.Context, metadata simplex.ProtocolMetadata) (simplex.VerifiedBlock, bool) {
+	curWait := 10 * time.Millisecond
+	maxBackoff := 5 * time.Second
+
+	for {
```
StephenButtolph: We should also use `initBackoff` below where we reset the backoff and remove the other calls to `backoff`, but didn't include it in this suggestion because github suggestions aren't smart enough to only show the relevant diff.
```suggestion
	const (
		initBackoff = 10 * time.Millisecond
		maxBackoff  = 5 * time.Second
	)
	for curWait := initBackoff; ; curWait = backoff(ctx, curWait, maxBackoff) {
```

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+			vmBlock, err := b.vm.BuildBlock(ctx)
+			if err != nil {
+				b.log.Info("Error building block", zap.Error(err))
+				curWait = backoff(ctx, curWait, maxBackoff)
+				continue
+			}
+			simplexBlock, err := newBlock(metadata, vmBlock, b.blockTracker)
+			if err != nil {
+				return nil, false
```
StephenButtolph: Can we log this as an error?

simplex/block_builder.go
```diff
@@ -0,0 +1,101 @@
...
+	case <-ctx.Done():
+		return 0
+	case <-time.After(backoff):
+	}
+
+	backoff *= 2 // Exponential backoff
+	if backoff > maxBackoff {
+		backoff = maxBackoff
+	}
+	return backoff
```
StephenButtolph:
```suggestion
	return min(maxBackoff, 2*backoff) // Exponential backoff
```

simplex/block_builder.go
StephenButtolph: Should we be using the backoff logic here as well?

simplex/config.go
```diff
@@ -13,6 +13,11 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
+type KeyValueReaderWriter interface {
+	Get(key []byte) ([]byte, error)
+	Put(key []byte, value []byte) error
+}
+
```
StephenButtolph: This is unused.
```suggestion
```

simplex/storage.go
```diff
@@ -123,13 +123,10 @@ func (s *Storage) Retrieve(seq uint64) (simplex.VerifiedBlock, simplex.Finalizat
 		return nil, simplex.Finalization{}, err
 	}
 
-	vb := &Block{vmBlock: block, metadata: finalization.Finalization.ProtocolMetadata, blockTracker: s.blockTracker}
-	bytes, err := vb.Bytes()
+	vb, err := newBlock(finalization.Finalization.ProtocolMetadata, block, s.blockTracker)
 	if err != nil {
-		s.log.Error("Failed to serialize block", zap.Error(err))
```
StephenButtolph, line 129: Is there a reason we are removing this log?

StephenButtolph (commented):
Haven't reviewed the tests yet

simplex/storage.go
samliok: no, re-added

simplex/block_builder.go
yacovm: `WaitForPendingTxs` sounds better to me than `IncomingBlock` but I don't like involving the term transaction, because Simplex operates at a block scope level and transactions are opaque to it. 

Perhaps `WaitForPendingBlock` ? `WaitForBuildBlockAvailable` seems a bit too verbose. Maybe `WaitForBlock` ?

simplex/block_builder.go
samliok: no, i dont think we need an exponential backoff after receiving an event. Wouldn't we want to return right away if we got a `PendingTxs` message, and we shouldn't ever get multiple `StateSyncDone` messages right?

simplex/block_builder.go
samliok: what about `WaitForAvailableBlock`?

simplex/block_builder.go
StephenButtolph: > we shouldn't ever get multiple StateSyncDone messages right?

VMs _must_ be willing to return multiple `StateSyncDone` messages.

simplex/block_builder.go
StephenButtolph: I'm happy with any of these suggestions my (light) favorite would probably be `WaitForPendingBlock`.

simplex/block_builder_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.Equal(t, 7, count, "Should have retried multiple times due to backoff")
```
StephenButtolph: This test seems racy to me. This seems to be assuming (which is probably correct normally) that the timeout goroutine will be scheduled before the main testing goroutine is rescheduled.

I feel like it would make more sense measure the amount of time it takes to call `WaitForEvent` `7` times and require that it is at least 630ms. (0+ 10 + 20 + 40 + 80 + 160 + 320).

Done this way it isn't racy because `time.After` is guaranteed to advance time _at least_ as much as requested.

simplex/block_builder.go
yacovm: Is this for the dynamic state sync?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,178 @@
...
+			defer cancelCtx()
+
+			block, built := bb.BuildBlock(timeoutCtx, child.BlockHeader().ProtocolMetadata)
+			require.Equal(t, tt.shouldBuild, built)
+			if tt.expectedBlock == nil {
+				require.Nil(t, block)
+				require.Greater(t, count, 1)
+			} else {
+				require.Equal(t, tt.expectedBlock, block)
+			}
```
StephenButtolph:
```suggestion
			require.Equal(t, tt.expectedBlock, block)
			if tt.expectedBlock == nil {
				require.Greater(t, count, 1)
			}
```

simplex/block_builder.go
StephenButtolph: Even with the current static state sync VMs can return multiple `StateSyncDone` messages. For example: https://github.com/ava-labs/coreth/blob/master/plugin/evm/vm.go#L867. This can be triggered multiple times if the engine hasn't marked bootstrapping as finished.

simplex/__debug_bin2891051211
StephenButtolph, line 1: This should be deleted. Could you make and issue/pr to add this pattern to the `.gitignore`?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+	timeoutTime := 10 * time.Second
+	timeoutCtx, cancelCtx := context.WithTimeout(ctx, timeoutTime)
+	defer cancelCtx()
+
+	start := time.Now()
+	bb.WaitForPendingBlock(timeoutCtx)
+	endTime := time.Since(start)
+
+	// 10, 20, 40, 80, 160, 320, 640 = 1270ms
+	require.GreaterOrEqual(t, endTime.Milliseconds(), int64(1270))
```
StephenButtolph:
```suggestion
	require.GreaterOrEqual(t, endTime, 1270 * time.Millisecond)
```

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+	})
+
+	count := 0
+	vm.WaitForEventF = func(_ context.Context) (common.Message, error) {
+		count++
+		return common.PendingTxs, nil
+	}
+
+	vm.BuildBlockF = func(_ context.Context) (snowman.Block, error) {
+		if count > 7 {
```
StephenButtolph: Since we are rate limiting `BuildBlock` I feel like this is a bit more natural
```suggestion
	vm.WaitForEventF = func(_ context.Context) (common.Message, error) {
		return common.PendingTxs, nil
	}

	count := 0
	vm.BuildBlockF = func(_ context.Context) (snowman.Block, error) {
		count++
		if count > 7 {
```

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	bb := &BlockBuilder{
+		log:          logging.NoLog{},
+		vm:           vm,
+		blockTracker: genesis.blockTracker,
+	}
+
+	timeoutTime := 10 * time.Second
+	timeoutCtx, cancelCtx := context.WithTimeout(ctx, timeoutTime)
+	defer cancelCtx()
```
StephenButtolph: Why do we add this timeout?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	bb := &BlockBuilder{
+		log:          logging.NoLog{},
+		vm:           vm,
+		blockTracker: nil,
+	}
+
+	timeoutTime := 10 * time.Second
+	timeoutCtx, cancelCtx := context.WithTimeout(ctx, timeoutTime)
+	defer cancelCtx()
```
StephenButtolph: Why do we add this timeout?

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+	ctx := context.Background()
+	vm := newTestVM()
+
+	count := 0
+	vm.WaitForEventF = func(_ context.Context) (common.Message, error) {
+		if count < 7 {
+			count++
+			return common.StateSyncDone, nil
+		}
+		return common.PendingTxs, nil
```
StephenButtolph: This is definitely being a bit pedantic, but this `WaitForEvent` count implementation is very similar (but different) to the `BuildBlock` count implementation above. Can we make it so that they follow the same logic? (increment the same way, ect). Bonus points if we can actually share the code in a reasonable way... But I had to understand twice why this is expected to call the functions 8 times

simplex/block_builder_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	start := time.Now()
+	block, built := bb.BuildBlock(timeoutCtx, child.BlockHeader().ProtocolMetadata)
+	endTime := time.Since(start)
+
+	require.True(t, built)
+	require.Equal(t, child.BlockHeader(), block.BlockHeader())
+
+	// 10, 20, 40, 80, 160, 320, 640 = 1270ms
+	require.GreaterOrEqual(t, endTime.Milliseconds(), int64(1270))
```
StephenButtolph:
```suggestion
	require.GreaterOrEqual(t, endTime, 1270 * time.Millisecond)
```

StephenButtolph: We could technically define:
```
	const (
		failedAttempts       = 7
		minimumExpectedDelay = initBackoff * (1<<failedAttempts - 1)
	)
```
but I'm not sure if this is just over-engineering stuff or if it would actually cleanup the code.

Feel free to adopt or ignore.

PR #4160, opened 2025-08-06
Design for Replicating Accepted Blocks for Non Validators

samliok:

### DIFF
--- a/simplex/NonValidators.MD
+++ b/simplex/NonValidators.MD
@@ -1,0 +1,47 @@
+### Replicating Blocks for Non-Validator Nodes
+
+**Motive**
+The current implementation of Simplex sends messages only to validator nodes. Messages can be sent directly to a specific node or broadcast to all nodes validating the chain. This limitation means Simplex cannot support non-validating nodes listening for updates to the chain. Namely, non-validating nodes such as API nodes and light clients have no way of querying the blockchain for accepted blocks. This document outlines the method for allowing non-validators to replicate accepted blocks.
+
+**Goal**: Allow nodes not participating in consensus (non-validators) to be able to replicate and verify accepted blocks.
+
+**Standard Replication Path**: Requesting Missing Sequences
+When a validating node falls behind it will eventually receive a finalization for a `seq x` that is greater than its current known seq. A finalization for a higher sequence means the node has fallen behind, and blocks for sequences up to and including `seq x` are guaranteed to be accepted into the chain. In this case, the node will begin the replication process which involves requesting the sequences it is missing (all sequences from its last accepted seq to `x`). Eventually, the nodes will respond and the lagging node will have caught up.
+Simplex’s `HandleMessage` function will process the replication request and respond with a response.
+
+```go
+func (e *Epoch) HandleMessage(msg *Message, from NodeID) error {
+```
+
+`ReplicationRequests` have the following structure:
+
+```go
+type ReplicationRequest struct {
+   Seqs        []uint64 // sequences we are requesting
+   LatestRound uint64   // latest round that we are aware of
+}
+
+```
+
+A `ReplicationResponse` has the following structure:
+
+```go
+type ReplicationResponse struct {
+   Data        []QuorumRound
+   LatestRound *QuorumRound
+}
+// QuorumRound represents a round that has achieved quorum on either
+// (empty notarization), (block & notarization), or (block, finalization)
+type QuorumRound struct {
+   Block             Block
+   Notarization      *Notarization
+   Finalization      *Finalization
+   EmptyNotarization *EmptyNotarization
+}
+```
+
+Non-validators should re-use this logic and can ignore any `QuorumRound` that do not have finalizations. This main benefit of reusing the same logic is that we can take advantage of the in-memory `rounds` map which will reduce the number of times we read from storage. Secondarily, this would be a minimalistic change that won't require maintaining a separate code path.
+
+**Broadcasting Finalizations**
+
+In order to learn about blocks that were accepted, non-validators need to receive finalizations. To accommodate this, we should expand the `Communication` implementation to also broadcast to non-validator nodes whenever the message being broadcasted is a `FinalizationMessage`. Once a non-validator receives a finalization message for a higher sequence, it can then request the full block via the standard replication path.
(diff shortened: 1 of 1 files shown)


simplex/NonValidators.MD
```diff
@@ -0,0 +1,54 @@
...
+
+**Goal**: Allow nodes not participating in consensus (non-validators) to be able to replicate and verify accepted blocks.
+
+**Broadcasting Finalizations**
+
+In order to learn about blocks that were accepted, non-validators need to receive finalizations. To accommodate this, we should expand the `Communication` implementation to also broadcast to non-validator nodes whenever the message being broadcasted is a `FinalizationMessage`. This does not require any changes to the `Simplex` protocol itself, as the `Communication` component is already an isolated component. It would be up to the `Communication` struct and `Engine` that wraps `Simplex` to maintain a list of non-validator nodes and broadcast finalization messages to them.
```
yacovm: `FinalizationMessage` -> `Finalization`.

simplex/NonValidators.MD
```diff
@@ -0,0 +1,54 @@
...
+// (empty notarization), (block & notarization), or (block, finalization)
+type QuorumRound struct {
+   Block             Block
+   Notarization      *Notarization
+   Finalization      *Finalization
+   EmptyNotarization *EmptyNotarization
+}
+```
+
+Non-validators should re-use this logic and can ignore any `QuorumRound` that do not have finalizations. The main benefit of reusing the same logic is that we can take advantage of the in-memory `rounds` map which will reduce the number of times we read from storage. Secondarily, this would be a minimalistic change that won't require maintaining a separate code path. In order to allow this, `Simplex` needs to be modified to accept `ReplicationRequests` from non-validator nodes.
```
yacovm: > In order to allow this, `Simplex` needs to be modified to accept `ReplicationRequests` from non-validator nodes.

I think the right approach would be to have a structure that will contain the epoch object and handle dispatching the requests instead of the epoch. That structure that contains the epoch inside of it, will reach into the rounds map of the epoch object if needed, to search for the missing block.

simplex/NonValidators.MD
```diff
@@ -0,0 +1,54 @@
...
+
+**Processing Replication Responses**
+
+When a non-validator receives a `ReplicationResponse`, it needs to verify the `QuorumRounds` by verifying the aggregated BLS signature amongst the corresponding validator set. Since different epochs can be [configured](https://github.com/ava-labs/avalanchego/pull/4124) with different validator sets, the non-validator needs to ensure it has the correct validator set for the epoch in which the `Finalization` was created. In the proposed implementation, the `Epoch` number represents the sequence of the sealing block the `Finalization` was created for. The non-validator node can follow [this proposed logic](https://github.com/ava-labs/avalanchego/pull/4124/files#diff-b9b433e0fcc085f654c562e9eaadb1e4de29b7fe6ee4bdc970280b895b3c078bR305) to verify the missing `Finalizations`:
```
yacovm: > The non-validator node can follow [this proposed logic](https://github.com/ava-labs/avalanchego/pull/4124/files#diff-b9b433e0fcc085f654c562e9eaadb1e4de29b7fe6ee4bdc970280b895b3c078bR305) to verify the missing

Can we please briefly summarize what is outlined in the link, for the sake of completeness?

simplex/NonValidators.MD
```diff
@@ -0,0 +1,54 @@
...
+
+**Bringing It All Together**
+
+Non-validators do not need to participate in consensus, so they can run a simplified version of `Simplex` that only processes `FinalizationMessages` and `ReplicationRequests/Responses`. The non-validator can maintain its own list of known validator nodes to send `ReplicationRequests` to. Whenever it receives a `FinalizationMessage` for a higher sequence than it currently knows, it can initiate the replication process by sending `ReplicationRequests` to the validators, or first syncing the correct validator set if necessary. Once it receives the `ReplicationResponses`, it can verify and store the accepted blocks.
```
yacovm: Can this functionality be re-used by a Simplex validator for replication of blocks if it has been offline long enough and is in a past epoch?

samliok: yea, the logic would be mostly the same. Although a validator would also want to replicate the notarizations, which non-validators can discard.

yacovm: If we replicate past epochs we don't need to replicate non finalized blocks

simplex/NonValidators.MD
samliok: what would be the benefit of creating a structure to separate the `ReplicationRequests` from the main `HandleMessage` flow? I can see it being useful to increase parallelization since we won't need to grab epoch lock, but were you thinking of something else?

yacovm: I don't think that the Epoch object should dispatch requests from non-validators. As the code matures we would probably need to add some extra logic for nodes, such as throttling, or metrics, etc. and I think we should keep the responsibilities of the epoch object to the minimum to remove complexity.

yacovm: >  I can see it being useful to increase parallelization since we won't need to grab epoch lock, but were you thinking of something else?

If we would want to access the rounds map we would probably need to lock (maybe we'll want to make it an RWLock but would still need to lock)

simplex/NonValidators.MD
```diff
@@ -0,0 +1,65 @@
...
+
+### Higher-Level Replication Coordinator
+
+To address these issues, we should implement a higher-level component that intercepts and routes replication messages based on the epoch associated with the requested sequences. This coordinator would handle the logic of determining where and how to fetch the requested data.
+
+For sequences in the current epoch, we could utilize the current epoch struct its in-memory cache for potential performance improvements. However, the behavior would differ based on node type: validating nodes would retrieve notarizations and empty notarizations in addition to finalizations, while non-validating nodes would only retrieve blocks and finalizations since they only care about accepted blocks.
```
yacovm, line 53: > However, the behavior would differ based on node type: validating nodes would retrieve notarizations and empty notarizations in addition to finalizations, while non-validating nodes would only retrieve blocks and finalizations since they only care about accepted blocks.

So basically a validator would look at a request's origin and would respond differently based on whether the requester is a validator or not, right?

PR #4162, opened 2025-08-07
Dynamically update mempool gossip request rate limit

joshua-kim:
## Why this should be merged

Prevents us from being spammed by gossip requests

## How this works

Dynamically updates the throttling limit based on the number of connected validators such that larger networks expect fewer samples against each node.

## How this was tested

Added unit tests

Yes


utils/set/set.go
```diff
@@ -192,3 +192,17 @@ func (s *Set[T]) Peek() (T, bool) {
 	}
 	return utils.Zero[T](), false
 }
+
+// Intersect returns the set intersection of s1 and s2
+func Intersect[T comparable](s1 Set[T], s2 Set[T]) Set[T] {
```
joshua-kim, line 197: I thought this looked better than putting it on a receiver... but admittedly it is inconsistent w/ the rest of the code.

vms/avm/network/config.go
```diff
@@ -21,9 +21,9 @@ var DefaultConfig = Config{
 	PushGossipMaxRegossipFrequency:              30 * time.Second,
 	PushGossipFrequency:                         500 * time.Millisecond,
 	PullGossipPollSize:                          1,
-	PullGossipFrequency:                         1500 * time.Millisecond,
```
joshua-kim, line 24: These constants were updated so we could conveniently use 3600 - but we could undo this diff.

In retrospect I'm also wondering if it makes sense to just decide the frequency we make requests based on the requests-per-validator throttling limit we configure...

utils/set/set.go
StephenButtolph, line 197: Just wanted to make sure that it was intentional not to use `Difference` here:
(A - B) == (A ∩ B)

joshua-kim: I thought those two operations weren't equal? I'm using this to get the connected validator set, which is the intersection of validators + connected peers. Wouldn't using difference would give us the non-connected validator set or am I misunderstanding your comment?

StephenButtolph: Sorry, I'm an idiot. lol

A - (A - B) == (A ∩ B)

(Which is super inefficient)

network/p2p/handler.go
```diff
@@ -57,6 +58,79 @@ func (NoOpHandler) AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([
 	return nil, nil
 }
 
+type DynamicThrottlerHandler struct {
+	handler         *ThrottlerHandler
+	throttler       *SlidingWindowThrottler
+	validatorSet    ValidatorSet
+	requestsPerPeer float64
```
samliok, line 71: i dont think it makes sense to have this as a float, unless we can send .x of a request?

network/p2p/handler.go
```diff
@@ -57,6 +58,79 @@ func (NoOpHandler) AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([
...
+	d.prevNumConnectedValidators = numValidators
+
+	n := float64(numValidators)
+
+	// guaranteed to not overflow an int
+	expectedSamples := d.requestsPerPeer / n
+	variance := d.requestsPerPeer * (n - 1) / (n * n)
+	stdDeviation := math.Sqrt(variance)
+
+	limit := int(expectedSamples + 4*stdDeviation)
```
samliok: seems like this limit should be lower. 4 std deviations means the node is in 99.9+th percentile of expectedSamples sent. wouldn't we want this to be between 2~3 sd's?

vms/example/xsvm/vm.go
```diff
@@ -74,6 +75,9 @@ func (vm *VM) Initialize(
 	vm.Network, err = p2p.NewNetwork(
 		chainContext.Log,
 		appSender,
+		chainContext.ValidatorState,
+		ids.Empty,
```
samliok: we can grab the subnet id from `chainContext.SubnetID`

network/p2p/network.go
```diff
@@ -59,6 +59,9 @@ type clientOptions struct {
 func NewNetwork(
 	log logging.Logger,
 	sender common.AppSender,
+	validatorState validators.State,
+	subnetID ids.ID,
+	maxValidatorSetStaleness time.Duration,
```
samliok: would it be better to pass in some sort of validator config? saying this because these three parameters are not used outside of creating the inner `Validators` struct, and I'm personally not a fan of methods with too many parameters. 

feel free to ignore if this is just not preferred golang style

network/p2p/network.go
```diff
@@ -126,11 +134,15 @@ func (n *Network) AppGossip(ctx context.Context, nodeID ids.NodeID, msg []byte)
 
 func (n *Network) Connected(_ context.Context, nodeID ids.NodeID, _ *version.Application) error {
 	n.Peers.add(nodeID)
+	n.Validators.connected(nodeID)
```
samliok: im slightly confused on why it is ok to add to both structs here. Is it true that all `validators` are `peers` but not all `peers` are `validators`? If so then don't we have to do some sort of check to make sure the node is a validator

samliok: ah i see we are adding peers, but are not considering them validators yet. I think this method should be named a bit more clear. Another thought, why can't we just pass in the `*Peers` object rather than creating a duplicate copy of it?

network/p2p/handler.go
joshua-kim: It's exposed to the caller as an `int`, but within the package we're using it as a float so that we don't have to deal with integer math/having to cast it to a float where necessary since the throttling limit is a floating point value.

This comment did however realize that I'm casting `limit` to an int... which was not intended...

network/p2p/handler.go
joshua-kim: It's just for outlier detection... each validator makes to a uniformly random selected validator, so we're modeling the distribution of requests we can get in expectation as a normal distribution. We just want to throttle people trying to spam us with requests while not throttling non-byzantine validators who are just happening to re-sample us... so this would throttle a non-byzantine validator 1-99.995% = 0.005% of the time. If we're doing 2 std devs it'd be closer to 2% (seems like a big false positive rate), and 3 would be 0.2% (still seems big). 4 std devs was chosen somewhat arbitrarily in an offline conversation... so it is a bit of a magic number. I could add a comment if that helps explain the reasoning?

vms/example/xsvm/vm.go
joshua-kim: Yeah that's not what I intended. Nice catch.

network/p2p/network.go
joshua-kim: I considered this, but then `Peers` and `Validators` would need to share a reference to a lock. We could do something like that - but I thought sharing a lock reference looked worse than just copying the data. Admittedly `Validators` is kind of in a bad spot right now... it owns both data for connected and non-connected validators which I find confusing.

network/p2p/network.go
joshua-kim: I don't want us to just slap "preferred go style" as an answer (since that doesn't really give us insight into the reasoning behind a decision...). It's definitely an idea... my thoughts are:
1. The number of things that need to be specified do not change if we put it into a config/as explicit parameters. One difference about params vs configs is that you break your callers when you modify it... which can be good or bad.
2. Configs are nice when you have zero-safe/reasonable defaults that you can use - since the zero value of a config will recursively initialize its fields with their zero values + you can set defaults lazily.
    * Some defaults we could use:
        * Reasonably I think we could have default value for `maxValidatorSetStaleness`
        * ... Maybe we could default to not logging or logging using stdout? 
    * But introducing a config to just wrap one value (`maxValidatorSetStaleness`) is a bit odd to add a layer of abstraction over a single value, and I don't know when we would ever want the default behavior for logging since we only use avalanchego's logger except for in tests.

network/p2p/handler_test.go
```diff
@@ -112,3 +116,166 @@ func TestValidatorHandlerAppRequest(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require := require.New(t)
+
+			if len(tt.validatorSets) != len(tt.requests) {
+				require.Fail("test vectors must be of same length")
+			}
```
samliok:
```suggestion
			require.Equal(len(tt.validatorSets), len(tt.requests))
```
is a one liner better here?

samliok (approved):

network/p2p/handler_test.go
joshua-kim: I can use `require.Len()`... not sure what I was doing here...

network/p2p/validators.go
```diff
@@ -110,6 +96,8 @@ func (v *Validators) refresh(ctx context.Context) {
 	}
 	utils.Sort(v.validatorList)
 
+	v.connectedValidators = set.Intersect(v.validatorSet, v.peers)
```
StephenButtolph, line 134: It seems that now we are only updating the connected validator set during this refresh, whereas previously it was updated on-the-fly.

Was that intended?

network/p2p/network.go
StephenButtolph: These arguments feel pretty unusual to me. We previously seemed to be trying to keep this package more modular in how things hook together... This seems to go against that.

Would it make sense to be able to register onConnected / onDisconnected notifications rather than having the p2p network struct own the validators entirely?

vms/platformvm/config/network.go
```diff
@@ -10,20 +10,22 @@ import (
...
+	PushRegossipNumPeers:           0,
+	PushGossipDiscardedCacheSize:   16384,
+	PushGossipMaxRegossipFrequency: 30 * time.Second,
+	PushGossipFrequency:            500 * time.Millisecond,
+	PullGossipPollSize:             1,
+	PullGossipFrequency:            1500 * time.Millisecond,
+	PullGossipThrottlingPeriod:     10 * time.Second,
+	// PullGossipRequestsPerValidator = PullGossipThrottlingPeriod / PullGossipFrequency =
+	// 10 seconds/period / 1.5 requests/second ~= 6.67 requests
+	PullGossipRequestsPerValidator:              7,
```
StephenButtolph: Does this make sense being a config? It is derived from `PullGossipFrequency` and `PullGossipThrottlingPeriod`. Or is this a config because it is based on _other_ peoples values of those provided configs?

vms/platformvm/config/network.go
```diff
@@ -10,20 +10,22 @@ import (
...
+	PushGossipNumValidators:        100,
+	PushGossipNumPeers:             0,
+	PushRegossipNumValidators:      10,
+	PushRegossipNumPeers:           0,
+	PushGossipDiscardedCacheSize:   16384,
+	PushGossipMaxRegossipFrequency: 30 * time.Second,
+	PushGossipFrequency:            500 * time.Millisecond,
+	PullGossipPollSize:             1,
+	PullGossipFrequency:            1500 * time.Millisecond,
+	PullGossipThrottlingPeriod:     10 * time.Second,
```
StephenButtolph: I had expected us to substantially increase the throttling period here. Isn't that needed to provide stronger bounds?

network/p2p/handler.go
```diff
@@ -57,6 +58,81 @@ func (NoOpHandler) AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([
...
+	// Throttling limit will be initialized when a request is handled
+	throttler := NewSlidingWindowThrottler(period, 0)
+
+	return &DynamicThrottlerHandler{
+		handler:         NewThrottlerHandler(handler, throttler, log),
+		throttler:       throttler,
+		validatorSet:    validatorSet,
+		requestsPerPeer: float64(requestsPerPeer),
+	}
+}
```
StephenButtolph: **TL;DR: We need to increase the period.**

Currently, the number of validators on mainnet is ~1k. The period is configured to be equal to `10s`, with a sample frequency of `1.5s`. This means we expect a single node to perform `6.6` requests per period (which we currently round up to `7`).

This means that:
```
expectedSamples = 7 / 1000 = 0.007
variance = 7 * (1000 - 1) / (1000 * 1000) ~= 0.007
stdDeviation ~= 0.08
limit = 0.007 + 4 * 0.08 ~= 0.327
```

The same logic on fuji with 100 validators
```
expectedSamples = 7 / 100 = 0.07
variance = 7 * (100 - 1) / (100 * 100) ~= 0.07
stdDeviation ~= 0.26
limit = 0.07 + 4 * 0.26 ~= 1.11
```

I think we need to significantly increase our window size for this logic to be reasonable. Having the limit be `< 1` doesn't make sense to me (as the actual rate limit it rounded up to the nearest integer). When we had talked about this internally we had discussed a period of one hour.

If we do the math with a period over an hour on mainnet:
```
expectedSamples = 3600 / 1.5 / 1000 = 2.4
variance = (3600 / 1.5) * (1000 - 1) / (1000 * 1000) ~= 2.4
stdDeviation ~= 1.54
limit = 2.4 + 4 * 1.54 ~= 8.56
```

This means we will effectively allow 9 requests in an hour, whereas the current logic would allow 1 req every 10s for 360 requests per hour.

network/p2p/throttler.go
StephenButtolph, line 1: Are we planning on changing the `currentHits` logic in a separate PR?

Recall that we want rate limited requests to increase `currentHits` as well as non-rate limited requests.

network/p2p/handler.go
```diff
@@ -57,6 +58,81 @@ func (NoOpHandler) AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([
...
+
+	// guaranteed to not overflow an int
+	expectedSamples := d.requestsPerPeer / n
+	variance := d.requestsPerPeer * (n - 1) / (n * n)
+	stdDeviation := math.Sqrt(variance)
+
+	// Throttle anything beyond 4 standard deviations which should throttle
+	// anything beyond the 99.994 percentile of expected requests.
+	limit := expectedSamples + 4*stdDeviation
+	d.throttler.setLimit(limit)
```
StephenButtolph, line 131: Would it make sense to make this a metric? I feel like this is something we should be interested in watching.

vms/platformvm/config/network.go
StephenButtolph: Is there a good reason that this can't be a float? We don't need the configs to be consensus critical.

network/p2p/validators.go
joshua-kim: Yeah this is done on purpose since if we hook this into connected/disconnected we'll hit a deadlock w/ the context lock. We could avoid this, see related thread https://github.com/ava-labs/avalanchego/pull/4162#discussion_r2271372882

network/p2p/throttler.go
joshua-kim: I was thinking we would adjust it in a separate PR... if you want us to couple it to this I can but it seems like an independent change to me

network/p2p/handler.go
joshua-kim: Sure - we can add a metric here

network/p2p/handler.go
joshua-kim: For some reason I wanted to keep the existing gossip values which is why I didn't adjust the values in this PR... but I think in hindsight it makes more sense to adjust it since I agree that the current values look pretty unusual w/ the validator set-based dynamic throttling.

vms/platformvm/config/network.go
joshua-kim: > ... Or is this a config because it is based on other peoples values of those provided configs?

Yeah that's correct

> Is there a good reason that this can't be a float? We don't need the configs to be consensus critical.

There isn't a reason - I use a float here which does play more nicely w/ the limit which is also a float.

network/p2p/network.go
joshua-kim: I wanted this originally but ended up not doing it... and upon looking back, the reason I coupled it was because I wanted `Validators` to know about `Peers` but since I ended up decoupling them anyways the original reason for this is now gone... I think this should be possible to do.

network/p2p/throttler.go
StephenButtolph: I'm happy with it as a separate change.

network/p2p/validators.go
joshua-kim: I did update this to update on-the-fly - but we can't do refreshes on the validator set due to the aforementioned reason during connect/disconnect

network/p2p/handler.go
joshua-kim: Added a metric + test

network/p2p/handler.go
joshua-kim: Updated throttling limit/period

vms/platformvm/config/network.go
joshua-kim: Updated as per the related comment

network/p2p/handler_test.go
```diff
@@ -112,3 +119,275 @@ func TestValidatorHandlerAppRequest(t *testing.T) {
...
+				require.GreaterOrEqual(
+					testutil.ToFloat64(handler.throttleLimitMetric),
+					wantSamples,
+				)
+			}
+		})
+	}
+}
+
+func TestNewDynamicThrottlerHandler(t *testing.T) {
```
joshua-kim, line 330: These tests are almost certainly overkill...

network/p2p/handler.go
```diff
@@ -57,6 +64,113 @@ func (NoOpHandler) AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([
...
+}
+
+func (d *DynamicThrottlerHandler) checkUpdateThrottlingLimit(ctx context.Context) {
+	numValidators := d.validatorSet.Len(ctx)
+
+	if numValidators == d.prevNumConnectedValidators {
+		return
+	}
+
+	d.prevNumConnectedValidators = numValidators
```
StephenButtolph, line 110: Isn't `checkUpdateThrottlingLimit` able to be called multiple times concurrently?

network/p2p/network.go
```diff
@@ -143,8 +167,11 @@ func (n *Network) NewClient(handlerID uint64, options ...ClientOption) *Client {
 		sender:        n.sender,
 		router:        n.router,
 		options: &clientOptions{
+			// Default to sending requests to ourselves
 			nodeSampler: &PeerSampler{
-				Peers: n.Peers,
+				Peers: &Peers{
+					set: set.OfSampleable[ids.NodeID](n.NodeID),
+				},
```
StephenButtolph: I feel like the default client behavior of sampling only ourself is pretty unusual.

Should we just require providing a node sampler? It would also remove the need to pass in the `NodeID` to `NewNetwork`

network/p2p/network.go
```diff
@@ -167,27 +194,20 @@ type Peers struct {
 	set  set.SampleableSet[ids.NodeID]
 }
 
-func (p *Peers) add(nodeID ids.NodeID) {
+func (p *Peers) Connected(nodeID ids.NodeID) {
```
StephenButtolph, line 149: Do we ever use the peers struct in production code anymore?

network/p2p/network.go
joshua-kim: Yeah I think that makes sense. In hindsight... this is pretty weird.

network/p2p/network.go
joshua-kim: No - it's only used for testing.

network/p2p/handler.go
joshua-kim: 😟 Yes it is. Added a lock around this.

StephenButtolph (approved):
Code lgtm - we should canary it out to make sure it is working as intended before merging.

StephenButtolph:
deadlock fix seems to be working as expected. Continuing to QA to verify the rate limiting is working as expected.

PR #4163, opened 2025-08-08
Fix noctx linter

joshua-kim:
## Why this should be merged

It's not a part of our CI, but I have a local version of golangci which is newer and this fires for me locally. This linter enforces the usage of newer context function signatures.

## How this works

Fixes linter by adding contexts where appropriate.

## How this was tested

CI

No


database/rpcdb/db_test.go
```diff
@@ -24,14 +24,14 @@ type testDatabase struct {
...
-func setupDB(t testing.TB) *testDatabase {
+func setupDB(t testing.TB, ctx context.Context) *testDatabase {
 	require := require.New(t)
 
 	db := &testDatabase{
 		server: memdb.New(),
 	}
 
-	listener, err := grpcutils.NewListener()
+	listener, err := grpcutils.NewListener(ctx)
```
StephenButtolph: We only ever pass in `ctx = context.Background()`, and I don't see why we would ever care to pass in anything else.
```suggestion
	listener, err := grpcutils.NewListener(context.Background())
```

vms/rpcchainvm/factory.go
```diff
@@ -49,15 +49,16 @@ func (f *factory) New(log logging.Logger) (interface{}, error) {
...
+	listener, err := grpcutils.NewListener(context.TODO())
 	if err != nil {
 		return nil, fmt.Errorf("failed to create listener: %w", err)
 	}
 
 	status, stopper, err := subprocess.Bootstrap(
 		context.TODO(),
 		listener,
-		subprocess.NewCmd(f.path),
+		subprocess.NewCmd(context.TODO(), f.path),
```
StephenButtolph: I'm not sure that `context.TODO()` is correct here. If we were to take a context in to `New`, we _must not_ provide that context to this command, as then the caller's context getting cancelled would kill the sub-process, even after this function has returned successfully.

The prior behavior where this was `context.Background` seems to be more correct.

vms/rpcchainvm/grpcutils/server.go
```diff
@@ -88,8 +89,8 @@ func WithStreamInterceptor(streamInterceptor grpc.StreamServerInterceptor) Serve
 
 // NewListener returns a TCP listener listening against the next available port
 // on the system bound to localhost.
-func NewListener() (net.Listener, error) {
-	return net.Listen("tcp", "127.0.0.1:")
+func NewListener(ctx context.Context) (net.Listener, error) {
+	return (&net.ListenConfig{}).Listen(ctx, "tcp", "127.0.0.1:")
```
StephenButtolph: Because this function hardcodes localhost, it is unexpected for this `Listen` to ever block. I would prefer the prior behavior of passing in `context.Background()`.

utils/ips/lookup.go
```diff
@@ -16,8 +17,8 @@ var errNoIPsFound = errors.New("no IPs found")
 // pick any of the IPs.
 //
 // Note: IPv4 is preferred because `net.Listen` prefers IPv4.
-func Lookup(hostname string) (netip.Addr, error) {
-	ips, err := net.LookupIP(hostname)
+func Lookup(ctx context.Context, hostname string) (netip.Addr, error) {
+	ips, err := (&net.Resolver{}).LookupIPAddr(ctx, hostname)
```
StephenButtolph, line 21: This is a good change :+1:

tests/fixture/tmpnet/process_runtime.go
```diff
@@ -120,9 +125,16 @@ func (p *ProcessRuntime) Start(ctx context.Context) error {
 		return stacktrace.Errorf("writing node flags: %w", err)
 	}
 
-	// All arguments are provided in the flags file
-	cmd := exec.Command(runtimeConfig.AvalancheGoPath, "--config-file", p.node.GetFlagsPath()) // #nosec G204
-	// Ensure process is detached from the parent process so that an error in the parent will not affect the child
+	// All arguments are provided in the flags file.
```
maru-ava: (No action required) Maybe localize this comment by moving it to the line above `--config-file`?

main/main.go
```diff
@@ -61,12 +62,12 @@ func main() {
 		fmt.Println(app.Header)
 	}
 
-	nodeApp, err := app.New(nodeConfig)
+	nodeApp, err := app.New(context.Background(), nodeConfig)
```
maru-ava, line 68: (No action required) Would it make sense to replace the current use of `signal.Notify` with a context created by [signal.NotifyContext](https://pkg.go.dev/os/signal#NotifyContext) and then use it both here and on line 71? Doing so would ensure clean shutdown when signals are received.

Not saying this PR needs to tackle it, but I think it would be a good cleanup to tackle at some point.

cc: @StephenButtolph

vms/rpcchainvm/grpcutils/server.go
```diff
@@ -89,7 +90,7 @@ func WithStreamInterceptor(streamInterceptor grpc.StreamServerInterceptor) Serve
 // NewListener returns a TCP listener listening against the next available port
 // on the system bound to localhost.
 func NewListener() (net.Listener, error) {
-	return net.Listen("tcp", "127.0.0.1:")
+	return (&net.ListenConfig{}).Listen(context.Background(), "tcp", "127.0.0.1:")
```
maru-ava, line 93: (No action required) This could also be a good place to use a notify context.

vms/rpcchainvm/runtime/subprocess/non_linux_stopper.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func NewCmd(path string, args ...string) *exec.Cmd {
-	return exec.Command(path, args...)
+	return exec.CommandContext(context.Background(), path, args...)
```
maru-ava, line 19: (No action required) This could also be a good place to use a notify context.

network/peer/peer.go
```diff
@@ -205,6 +205,7 @@ type peer struct {
 // Invariant: There must only be one peer running at a time with a reference to
 // the same [config.InboundMsgThrottler].
 func Start(
+	ctx context.Context,
```
aaronbuchwald, line 208: Since these functions are previously expected to run until there's an error or network shutdown adding a context to the signature means we need to be careful about what context gets passed in and whether cancelling it could cancel what are supposed to be long-running functions. Looking through the networking package for any case where this could break assumptions.

network/peer/peer.go
```diff
@@ -755,7 +756,7 @@ func (p *peer) handle(msg message.InboundMessage) {
 		msg.OnFinishedHandling()
 		return
 	case *p2p.PeerList:
-		p.handlePeerList(m)
+		p.handlePeerList(ctx, m)
```
aaronbuchwald, line 759: Should we add a context to all of the handler functions rather than just for peer list?

aaronbuchwald: I assume we only added it here because it was the only one that passed down the context another level

network/peer/peer.go
```diff
@@ -1188,7 +1189,7 @@ func (p *peer) handlePeerList(msg *p2p.PeerList) {
 		)
 	}
 
-	if err := p.Network.Track(discoveredIPs); err != nil {
+	if err := p.Network.Track(ctx, discoveredIPs); err != nil {
```
aaronbuchwald, line 1192: The `ctx` that gets passed in here comes all the way from `Start` along the following path:

Start: https://github.com/ava-labs/avalanchego/blob/75efe4e93aa49a07c3bd4e6f50ec96e6601d49aa/network/peer/peer.go#L208
go p.readMessages: https://github.com/ava-labs/avalanchego/blob/75efe4e93aa49a07c3bd4e6f50ec96e6601d49aa/network/peer/peer.go#L236
handle: https://github.com/ava-labs/avalanchego/blob/75efe4e93aa49a07c3bd4e6f50ec96e6601d49aa/network/peer/peer.go#L499
handlePeerList: https://github.com/ava-labs/avalanchego/blob/75efe4e93aa49a07c3bd4e6f50ec96e6601d49aa/network/peer/peer.go#L759
Track: https://github.com/ava-labs/avalanchego/blob/75efe4e93aa49a07c3bd4e6f50ec96e6601d49aa/network/peer/peer.go#L1192

Passing an expired context into `Track` seems like it could be problematic because we use it in an async context to establish/upgrade peer connections, which may error if that context is cancelled.

At the same time, the caller may pass in a context and defer cancellation until that function returns, which could incorrectly cancel any of these goroutines to create a connection.

I thought this may create a subtle change in invariants for `StartTestPeer` (which to my knowledge is still used by the relayer in production), when first looking at this, but it seems that uses `TestNetwork` which is a no-op for `Track` and wouldn't be the case.

Still this code seems fairly sensitive to an accidental, subtle change in invariants by switching from a context that is scoped to the duration of a function to being used in a long-lived async manner.

Would it make sense to exclude this package from this linter for a first pass?

aaronbuchwald:
We may want to exclude then network package from this change or split it out to a follow up as there's some undocumented behavior and potential for accidentally promoting a context scoped to the duration of a function to using a context for long-lived goroutines.

network/peer/peer.go
joshua-kim: > Passing an expired context into Track seems like it could be problematic because we use it in an async context to establish/upgrade peer connections, which may error if that context is cancelled.

Why would an error here be concerning? If a peer errors we would close our connection and contexts are only canceled on shutdown anyway.

> At the same time, the caller may pass in a context and defer cancellation until that function returns, which could incorrectly cancel any of these goroutines to create a connection.

This would be improper usage of the context. If you have a long-running goroutine, you shouldn't be canceling the context. You would probably want to hold onto the cancel function and cancel it later.

FWIW, the root parent context passed into avalanchego is never canceled.

> Would it make sense to exclude this package from this linter for a first pass?

Wouldn't we still have to resolve this linter in this package anyways? Either we would have a bunch of nolint overrides, or we would have to apply this diff anyways.

network/peer/peer.go
joshua-kim: > Should we add a context to all of the handler functions rather than just for peer list?

I don't think we should pre-emptively add a context parameter to the other functions because it would be unused

> I assume we only added it here because it was the only one that passed down the context another level

That is correct

main/main.go
joshua-kim: It seems like it would simplify shutdown - I think it makes sense to do. I think we could probably tackle it in a follow-up to this.

vms/rpcchainvm/grpcutils/server.go
joshua-kim: In this + your other comment, we don't use `signal.Notify`/care about signal handling here... am I missing something?

tests/fixture/tmpnet/process_runtime.go
joshua-kim: Moved

vms/rpcchainvm/grpcutils/server.go
maru-ava: Apologies if I wasn't clear, it's related to my earlier comment about using a NotifyContext for the whole app and have every subsequent context use that as a parent so that things like this - a potentially persistent tcp listener - would be shut down cleanly on app shutdown. That implies that NewListener should be accepting a context and a long-running listener could be initialized with a context with a NotifyContext in its lineage.

joshua-kim: Ah, that makes sense.

network/peer/peer.go
aaronbuchwald: Isn't this package and the new test peer used externally as well? It's less than ideal, but my understanding is that `StartTestPeer` has external consumers that depend on this and this code change would subtly change the behavior such that the previous code will handle it fine if you cancel a context after the function returns and this code may change that behavior. So if we make this change we need to double check that those external consumers do not misuse this API.

joshua-kim: 1. `StartTestPeer` is not used by the relayer - `TestNetwork` is ([ref](https://github.com/ava-labs/icm-services/blob/b42b638a27c17569d7652f9ff0e9d64eb0f7d21e/peers/app_request_network.go#L191)).
2. This PR adds the contexts to the `Network` type. If there is misuse of the contexts, it'll be when the relayer upgrades to the next version of avalanchego. It's not possible for the relayer to have existing misuse of the context since it's not on the `Network` api yet.
3. Additionally, this PR doesn't expose contexts on the `TestNetwork` type anyways - we hardcode `context.Background`.

JonathanOppenheimer:
Closing this to refresh it, after consultation with @joshua-kim

JonathanOppenheimer:
See: https://github.com/ava-labs/avalanchego/pull/4975

PR #4164, opened 2025-08-11
Add network config to connect to all validators

iansuvak:
## Why this should be merged

Connecting to validators of all networks beyond the current max of 16 tracked subnets without being a primary network validator is useful for ICM message signing.

## How this works

Adds a new `connectToAllValidators` network config option but doesn't expose it on the avalanchego config level. 
Propagates it down to `peer` and `iptracker` levels 

## How this was tested

Tested using E2E on a branch of `icm-services` repo. 
Added new unit tests primarily on the `ip_tracker` side. 

No

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -544,7 +544,7 @@ func (n *network) KnownPeers() ([]byte, []byte) {
 // There are 3 types of responses:
 //
 // - Respond with subnet IPs tracked by both ourselves and the peer
-//   - We do not consider ourself to be a primary network validator
+//   - We do not consider ourself to be a primary network validator and do not track all subnets
 //
 // - Respond with all subnet IPs
 //   - The peer requests all peers
@@ -552,7 +552,7 @@ func (n *network) KnownPeers() ([]byte, []byte) {
 //
 // - Respond with subnet IPs tracked by the peer
 //   - The peer does not request all peers
-//   - We believe ourself to be a primary network validator
+//   - We believe ourself to be a primary network validator or we track all subnets
 //
 // The reason we allow the peer to request all peers is so that we can avoid
 // sending unnecessary data in the case that we consider them a primary network


network/ip_tracker.go
```diff
@@ -64,11 +65,12 @@ func newIPTracker(
...
-		tracked:        make(map[ids.NodeID]*trackedNode),
-		bloomAdditions: make(map[ids.NodeID]int),
-		connected:      make(map[ids.NodeID]*connectedNode),
-		subnet:         make(map[ids.ID]*gossipableSubnet),
+		bloomMetrics:           bloomMetrics,
+		tracked:                make(map[ids.NodeID]*trackedNode),
+		bloomAdditions:         make(map[ids.NodeID]int),
+		connected:              make(map[ids.NodeID]*connectedNode),
+		subnet:                 make(map[ids.ID]*gossipableSubnet),
+		connectToAllValidators: connectToAllValidators,
```
joshua-kim: nit: move this to before `bloomMetrics`, because this is something that the caller provided for construction (we typically order fields in stuff the caller defined -> stuff we created for our own state).

network/ip_tracker.go
```diff
@@ -272,7 +276,7 @@ func (i *ipTracker) WantsConnection(nodeID ids.NodeID) bool {
 	defer i.lock.RUnlock()
 
 	node, ok := i.tracked[nodeID]
-	return ok && node.wantsConnection()
+	return ok && (node.wantsConnection() || (i.connectToAllValidators && node.validatedSubnets.Len() > 0))
```
joshua-kim: nit: we need to update the comment on this function

network/ip_tracker.go
```diff
@@ -428,7 +433,8 @@ func (i *ipTracker) addTrackableID(nodeID ids.NodeID, subnetID *ids.ID) {
 		nodeTracker.manuallyTracked = true
 	} else {
 		nodeTracker.validatedSubnets.Add(*subnetID)
-		if *subnetID == constants.PrimaryNetworkID || i.trackedSubnets.Contains(*subnetID) {
+		if *subnetID == constants.PrimaryNetworkID || i.trackedSubnets.Contains(*subnetID) ||
+			(i.connectToAllValidators && nodeTracker.validatedSubnets.Len() > 0) {
```
joshua-kim: Why do we do this `.Len()` check? Isn't this always guaranteed to be greater than zero since we add to it in the previous line?

network/network.go
```diff
@@ -503,13 +504,14 @@ func (n *network) AllowConnection(nodeID ids.NodeID) bool {
 		return true
 	}
 	_, areWeAPrimaryNetworkAValidator := n.config.Validators.GetValidator(constants.PrimaryNetworkID, n.config.MyNodeID)
-	return areWeAPrimaryNetworkAValidator || n.ipTracker.WantsConnection(nodeID)
+	return areWeAPrimaryNetworkAValidator || n.ipTracker.WantsConnection(nodeID) || n.config.ConnectToAllValidators
```
joshua-kim: Why do we check the `ConnectToAllValidators` config here? I thought this was checked in `WantsConnection` and if this is a non-validator we're connecting to i think we'll blindly accept the connection even if we're also a non-validator.

network/network_test.go
```diff
@@ -160,6 +160,10 @@ func newDefaultResourceTracker() tracker.ResourceTracker {
 }
 
 func newTestNetwork(t *testing.T, count int) (*testDialer, []*testListener, []ids.NodeID, []*Config) {
+	return newTestNetworkInner(t, count, false)
+}
+
+func newTestNetworkInner(t *testing.T, count int, connectToAllValidators bool) (*testDialer, []*testListener, []ids.NodeID, []*Config) {
```
joshua-kim: Instead of having this function maybe we could pass in the config to use so we could specify the `connectToAllValidators` flag... `*Inner` is a bit rough to read.

network/peer/peer_test.go
```diff
@@ -649,7 +649,6 @@ func TestShouldDisconnect(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			require := require.New(t)
-
```
joshua-kim, line 652: nit: clean out unrelated diff

network/ip_tracker.go
```diff
@@ -267,12 +271,13 @@ func (i *ipTracker) ManuallyGossip(subnetID ids.ID, nodeID ids.NodeID) {
...
 //  2. The node has been manually gossiped on a tracked subnet.
 //  3. The node is currently a validator on a tracked subnet.
+//  4. The node is currently a validator on any subnet and connectToAllValidators is true.
 func (i *ipTracker) WantsConnection(nodeID ids.NodeID) bool {
 	i.lock.RLock()
 	defer i.lock.RUnlock()
 
 	node, ok := i.tracked[nodeID]
-	return ok && node.wantsConnection()
+	return ok && (node.wantsConnection() || (i.connectToAllValidators && node.validatedSubnets.Len() > 0))
```
StephenButtolph:
```suggestion
	return ok && node.wantsConnection()
```

network/ip_tracker.go
```diff
@@ -320,7 +326,7 @@ func (i *ipTracker) AddIP(ip *ips.ClaimedIPPort) bool {
 	if connectedNode, ok := i.connected[ip.NodeID]; ok {
 		i.setGossipableIP(trackedNode.ip, connectedNode.trackedSubnets)
 	}
-	return trackedNode.wantsConnection()
+	return trackedNode.wantsConnection() || (i.connectToAllValidators && trackedNode.validatedSubnets.Len() > 0)
```
StephenButtolph:
```suggestion
	return trackedNode.wantsConnection()
```

network/ip_tracker.go
```diff
@@ -428,7 +434,7 @@ func (i *ipTracker) addTrackableID(nodeID ids.NodeID, subnetID *ids.ID) {
 		nodeTracker.manuallyTracked = true
 	} else {
 		nodeTracker.validatedSubnets.Add(*subnetID)
-		if *subnetID == constants.PrimaryNetworkID || i.trackedSubnets.Contains(*subnetID) {
+		if *subnetID == constants.PrimaryNetworkID || i.trackedSubnets.Contains(*subnetID) || i.connectToAllValidators {
```
StephenButtolph, line 437: This line here removes the requirement for checking the flag in `AddIP` or `WantsConnection`.

network/network.go
StephenButtolph, line 1: Should we also be updating the `Peers` function here?

network/peer/peer.go
```diff
@@ -641,10 +641,11 @@ func (p *peer) sendNetworkMessages() {
 		case <-p.getPeerListChan:
 			knownPeersFilter, knownPeersSalt := p.Config.Network.KnownPeers()
 			_, areWeAPrimaryNetworkValidator := p.Validators.GetValidator(constants.PrimaryNetworkID, p.MyNodeID)
+			requestAllSubnetIPs := areWeAPrimaryNetworkValidator || p.Config.ConnectToAllValidators
 			msg, err := p.Config.MessageCreator.GetPeerList(
 				knownPeersFilter,
 				knownPeersSalt,
-				areWeAPrimaryNetworkValidator,
+				requestAllSubnetIPs,
```
StephenButtolph, line 649: I think we should be updating this in `writeMessages()` with `Handshake` as well.

network/network.go
iansuvak: I made the change but this is just for the node with `connectToAllValidators` set to true to be more useful in responding to `GetPeerList` messages from it's peers rather than it's ability to discover others right?

network/ip_tracker_test.go
```diff
@@ -63,12 +107,15 @@ func TestIPTracker_ManuallyTrack(t *testing.T) {
 	subnetID := ids.GenerateTestID()
 	tests := []struct {
 		name           string
-		initialState   func(t *testing.T) *ipTracker
+		initialState   func(t *testing.T, config *testConfig) *ipTracker
 		expectedChange func(*ipTracker)
+		skipConfigs    []*testConfig
```
joshua-kim: Is this used? Seems to always be empty so we never skip any tests

network/ip_tracker_test.go
```diff
@@ -308,16 +370,24 @@ func TestIPTracker_ManuallyGossip(t *testing.T) {
...
-			expectedState := test.initialState(t)
 
-			testState.ManuallyGossip(test.subnetID, ip.NodeID)
-			test.expectedChange(expectedState)
+	for _, config := range testConfigs {
+		t.Run(config.name, func(t *testing.T) {
+			for _, test := range tests {
+				if shouldSkipTest(test.skipConfigs, config) {
+					continue
+				}
```
joshua-kim: I left a few comments previously on getting rid of the skip configs, but if we kept them would we want to call `t.Skip`? This will silently pass the test otherwise since we've already entered `t.Run`.

network/ip_tracker_test.go
```diff
@@ -198,11 +258,12 @@ func TestIPTracker_ManuallyGossip(t *testing.T) {
 					gossipableIndices:  make(map[ids.NodeID]int),
 				}
 			},
+			skipConfigs: []*testConfig{&connectToAllValidatorsConfig},
```
joshua-kim: Why do we support test skipping in this PR? It seems like some of these are valid configurations (connect to all validators specified, but a non-connected untracked non-validator tries to connect to us, I think we'd expect a no-op).

network/ip_tracker_test.go
```diff
@@ -18,11 +18,55 @@ import (
...
+var connectToAllValidatorsConfig = testConfig{
+	name:       "connect_to_all_validators",
+	newTracker: newTestIPTrackerConnectToAll,
+}
+
+// testConfigs defines all the test configurations we want to run
+var testConfigs = []*testConfig{
+	&normalConfig,
+	&connectToAllValidatorsConfig,
+}
```
joshua-kim: I think these test helpers are being added to try to not copy-paste too much code, but I think that copy-pasting the existing test-cases and changing one param (enabling connect to all validators) would actually be simpler to read than some of the boilerplate which is adding more complexity (setup functions in test tables, config passing, more test-helpers, etc).

Since we're trying to re-use the existing test coverage across two configs, it's causing some friction where we have to introduce this "skip config" abstraction since the connect-to-all and normal config when behavior expected differs.

iansuvak: I've gone ahead and made the change to copy pasta with comments instead for any case where the behavior differs. I'm not sure if this is cleaner yet but let me know what you think.

network/ip_tracker_test.go
iansuvak: So the behavior here actually does differ since the node is  indeed a validator just not on the subnet we are explicitly tracking so we do expect to add the subnetID to the `trackedNode.trackedSubnets` field since we implicitly track all subnets.

network/ip_tracker_test.go
iansuvak: I agree-- I've made the change to remove skip configs but if we choose to put them back I will make the change.

network/ip_tracker_test.go
```diff
@@ -129,7 +132,82 @@ func TestIPTracker_ManuallyTrack(t *testing.T) {
...
+				tracker := newTestIPTracker(t, false)
+				tracker.Connected(ip, set.Of(constants.PrimaryNetworkID))
+				tracker.OnValidatorAdded(subnetID, ip.NodeID, nil, ids.Empty, 0)
+				return tracker
+			},
+			expectedChange: func(tracker *ipTracker) {
+				tracker.tracked[ip.NodeID].manuallyTracked = true
+			},
+		},
+		// Section below is same as above with connectToAllValidators set to true
```
joshua-kim: nit: I would remove these section comments throughout this PR - at some point we're going to add more tests and it's going to go out of sync... even if we add an start-and-end-section comment we're probably going to mess it up somehow by putting tests in the middle.

iansuvak: Removed the section comments but left the individual case comments that differ in behavior when `connectToAllValidators` is set to false.

network/network.go
```diff
@@ -574,7 +576,7 @@ func (n *network) Peers(
 		}
 	}
 
-	if areWeAPrimaryNetworkValidator && requestAllPeers {
+	if (areWeAPrimaryNetworkValidator || n.config.ConnectToAllValidators) && requestAllPeers {
```
StephenButtolph: I feel like this should be:
```go
	_, areWeAPrimaryNetworkValidator := n.config.Validators.GetValidator(constants.PrimaryNetworkID, n.config.MyNodeID)
	weHaveAllSubnetIPs := areWeAPrimaryNetworkValidator || n.config.ConnectToAllValidators

	// Only return IPs for subnets that we are tracking.
	var allowedSubnets func(ids.ID) bool
	if weHaveAllSubnetIPs {
		allowedSubnets = func(ids.ID) bool { return true }
	} else {
		allowedSubnets = func(subnetID ids.ID) bool {
			return subnetID == constants.PrimaryNetworkID || n.ipTracker.trackedSubnets.Contains(subnetID)
		}
	}

	if weHaveAllSubnetIPs && requestAllPeers {
```
If we are keeping up-to-date with all IPs, we should be able to serve all IPs upon request.

As it is right now we iterate over all subnets but we filter them to only be subnets that we track.

network/network.go
StephenButtolph: All of these cases should be getting modified.

Essentially, where we were previously checking if we are a primary network validator.

We should now be checking (if we are a primary network validator) or (if we are connecting to all validators).

joshua-kim:
Thought that I already approved this...

PR #4165, opened 2025-08-11
Move C-Chain benchmark to custom action and add ARC + GH runner triggers

aaronbuchwald:
Unfortunately, GitHub Actions can be restrictive in unexpected and tedious ways.

This PR updates the C-Chain execution benchmark workflow to run on both Actions Runner Controller and a GitHub hosted runner.

To run on ARC, we need to specify the image parameter in the runs-on section, but do not want to add this extra layer when running on a GitHub runner. Unfortunately, if you populate this parameter and even conditionally evaluate it to an empty string, it's still interpreted differently from not being there in the first place

ie.

```
    runs-on: ${{ github.event.inputs.runner || 'avalanche-avalanchego-runner-2ti' }}
    container:
      image: ''
```

is invalid as once `container.image` is specified, it must provide an image parameter.

As a result, this PR migrates the bulk of the logic for the benchmark to a custom action and sets up two separate jobs so that the `runs-on` and `container.image` params can be set separately for ARC vs. GitHub runner.

Now that ARC is supported and we have current state snapshots in the S3 bucket every 10m blocks and at height 33m, this changes the scheduled benchmark run to execute the range [33000001, 34000000].

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,7 +12,8 @@
 /.github/ @joshua-kim @maru-ava
 /.github/*.md @joshua-kim @maru-ava @meaghanfitzgerald
 /.github/CODEOWNERS @StephenButtolph
-/.github/workflows/c-chain-reexecution-benchmark.yml @aaronbuchwald
+/.github/actions/c-chain-reexecution-benchmark/ @aaronbuchwald
+/.github/workflows/c-chain-reexecution-benchmark* @aaronbuchwald
 /.gitignore @joshua-kim @maru-ava @StephenButtolph
 /.golangci.yml @joshua-kim @maru-ava @StephenButtolph
 /Dockerfile @joshua-kim @maru-ava
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -43,7 +43,7 @@ inputs:
   github-token:
     description: 'GitHub token provided to GitHub Action Benchmark.'
     required: true
-  push:
+  push-github-action-benchmark:
     description: 'Whether to push the benchmark result to GitHub.'
     required: false
     default: false
@@ -95,10 +95,10 @@ runs:
         auto-push: false
 
     - name: Push Benchmark Result
-      if: ${{ inputs.push }}
+      if: ${{ inputs.push-github-action-benchmark }}
       uses: benchmark-action/github-action-benchmark@v1
       with:
         tool: 'go'
         output-file-path: ${{ env.BENCHMARK_OUTPUT_FILE }}
         github-token: ${{ inputs.github-token }}
-        auto-push: true
+        auto-push: ${{ inputs.push-github-action-benchmark }}
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -21,11 +21,9 @@ inputs:
   aws-role:
     description: 'AWS role to assume for S3 access.'
     required: true
-    default: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
   aws-region:
     description: 'AWS region to use for S3 access.'
     required: true
-    default: 'us-east-2'
   prometheus-username:
     description: 'The username for the Prometheus instance.'
     required: false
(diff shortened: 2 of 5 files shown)


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -0,0 +1,105 @@
...
+    description: 'The current state directory. Supports S3 directory/zip and local directories.'
+    required: false
+    default: 's3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100.zip'
+  aws-role:
+    description: 'AWS role to assume for S3 access.'
+    required: true
+  aws-region:
+    description: 'AWS region to use for S3 access.'
+    required: true
+  prometheus-username:
```
maru-ava, line 23: (No action required) Once https://github.com/ava-labs/avalanchego/pull/4149 merges, it will be necessary to provide prometheus URLs in addition to the auth credentials.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -0,0 +1,105 @@
+name: 'C-Chain Re-Execution Benchmark'
+description: 'Run C-Chain re-execution benchmark'
+
+inputs:
+  start-block:
+    description: 'The start block for the benchmark.'
+    required: false
```
maru-ava: (No action required) Since the zero value for a boolean field like `required` is `false`, it doesn't have to be specified explicitly.

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -29,76 +29,80 @@ on:
...
+          source-block-dir: ${{ steps.set-params.outputs.source-block-dir }}
+          current-state-dir: ${{ steps.set-params.outputs.current-state-dir }}
+          prometheus-username: ${{ secrets.PROMETHEUS_ID || '' }}
+          prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
+          push-github-action-benchmark: ${{ github.event_name == 'schedule' }}
+          aws-role: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-region: 'us-east-2'
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  c-chain-reexecution-pr-gh-runner: # Smoke test on GitHub hosted runner
```
maru-ava: (No action required) Maybe put this job in a separate workflow file to provide a clean separation of triggering conditions? i.e. the workflow for this job would only have `on: pull_request`, and the workflow for the arc job would have `on: workflow` and `on: schedule`?

JuanLeon2 (approved):
Did you mean to add actions/c-chain-reexecution-benchmark to CODEOWNERS, to match the workflow?  In

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: Noted: I suppose that should be included here if that merges first and there if vice versa. Unless it's valid to pass the param even when it doesn't exist (which I believe fails linting)

.github/workflows/c-chain-reexecution-benchmark.yml
aaronbuchwald: Ya that's a good point. I was thinking the same, but slightly preferred to keep it in the same file as it's easier to read in one place. Will switch to two files with a shared prefix in their name.

aaronbuchwald:
> Did you mean to add actions/c-chain-reexecution-benchmark to CODEOWNERS, to match the workflow? In

Good call out, I'll make this change

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: Removing and doing a pass over the default values

.github/workflows/c-chain-reexecution-benchmark.yml
aaronbuchwald: I am keeping the pull request label for both to ensure it's working on both, but running a small enough range that it's reasonable to run on every PR.

.github/CODEOWNERS
```diff
@@ -12,7 +12,8 @@
 /.github/ @joshua-kim @maru-ava
 /.github/*.md @joshua-kim @maru-ava @meaghanfitzgerald
 /.github/CODEOWNERS @StephenButtolph
-/.github/workflows/c-chain-reexecution-benchmark.yml @aaronbuchwald
+/.github/actions/c-chain-reexecution-benchmark/ @aaronbuchwald
+/.github/workflows/c-chain-reexecution-benchmark* @aaronbuchwald
```
aaronbuchwald, line 16: Cover both workflows with their respective triggers

PR #4166, opened 2025-08-11
uplift: (validators) - refactor uptime tracking from evm repos

### LINKED ISSUE #4207: Uplift validators package
See https://github.com/ava-labs/avalanchego/pull/4166

JonathanOppenheimer:
## Why this should be merged
There are packages in subnet-evm and coreth that share a lot of functionality / are mirrored. These packages should be uplifted to AvalancheGo to prevent the duplicative maintenance of two sets of the same code. This is part of that effort, specifically in assisting the uplifting of the warp package. Closes #4207 

Validators has no external imports but there are a couple things worth noting: 

- `validators/locked_reader.go` is similar to `snow/uptime.LockedCalculator` (thread-safe reads) but not a duplicate: it also returns validator metadata atomically with uptime. 
- There is a minor name overlap with `validatorData` it exists in `proposervm` but with different fields and in a different package

## How this was tested

No

### DIFF
--- a/vms/evm/uptimetracker/state.go
+++ b/vms/evm/uptimetracker/state.go
@@ -138,9 +138,9 @@ func (s *state) addNewValidator(vdr *validator) {
 	s.updatedValidators.Add(vdr.validationID)
 }
 
-// updateValidator updates the validator with the given validationID to the
-// given isActive state -- this function does assume that a validator with t
-// he given validationID exists.
+// updateValidator sets the isActive state of the validator with the given
+// validationID -- this function assumes that a validator with the given
+// validationID exists.
 func (s *state) updateValidator(validationID ids.ID, isActive bool) bool {
 	v := s.validators[validationID]
 


vms/evm/plugin/validators/interfaces/interfaces.go
joshua-kim, line 1: 1. `interfaces` files/packages are considered a cardinal sin in Go - interfaces should be defined near where they are consumed/used.
2. I'm not sure if we even need any of these interfaces - I think we could probably just delete them all and use the concrete type instead.

vms/evm/plugin/validators/locked_reader.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"time"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/evm/plugin/validators/interfaces"
+
+	stateinterfaces "github.com/ava-labs/avalanchego/vms/evm/plugin/validators/state/interfaces"
+)
+
+type RLocker interface {
```
joshua-kim: Do we need this interface? If we're only ever using `sync.RWLock`, it feels like a premature abstraction that just adds package bloat/more layers of indirection.

vms/evm/plugin/validators/locked_reader.go
```diff
@@ -0,0 +1,54 @@
...
+	RLock()
+	RUnlock()
+}
+
+type lockedReader struct {
```
joshua-kim: Why do we have this locked reader? The existing `manager` type already know about the lock anyways - so I'm wondering why we care to decouple it.

vms/evm/plugin/validators/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	avalancheuptime "github.com/ava-labs/avalanchego/snow/uptime"
+	stateinterfaces "github.com/ava-labs/avalanchego/vms/evm/plugin/validators/state/interfaces"
+	uptimeinterfaces "github.com/ava-labs/avalanchego/vms/evm/plugin/validators/uptime/interfaces"
+)
+
+const (
+	SyncFrequency = 1 * time.Minute
+)
+
+type manager struct {
```
joshua-kim: Can we avoid the naming pattern of  `*Manager` - it feels like a slippery slope into a god-object that has no clear abstraction boundary? It seems like something like `ValidatorState` might make more sense.

vms/evm/plugin/validators/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	chainCtx *snow.Context
+	stateinterfaces.State
+	uptimeinterfaces.PausableManager
```
joshua-kim: Can we get rid of all of the interface embedding (use fields instead)? It adds more layers of indirection and can lead to subtle behavior where we can silently call the wrong function in the embedded type.

vms/evm/plugin/validators/README.md
joshua-kim, line 1: Who's the intended audience of this doc? It doesn't seem like something a person building on us would use... so I'm assuming it's for us? If so - I would prefer for this documentation to live in godoc, not as a markdown file since `.md` docs tend to be more difficult to maintain/easy to forget to update as they are decoupled from code.

vms/evm/plugin/validators/state/interfaces/mock_listener.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	reflect "reflect"
+
+	ids "github.com/ava-labs/avalanchego/ids"
+	gomock "go.uber.org/mock/gomock"
+)
+
+// MockStateCallbackListener is a mock of StateCallbackListener interface.
+type MockStateCallbackListener struct {
```
joshua-kim: 1. Mocks tend to create brittle tests + doo a poor job of testing our actual production code, so I want us to avoid using them where possible. Can we delete this type + the related mock generation code?
2. Our only listener implementation we use is for uptime tracking, so I'm wondering if it would make more sense for us to test uptimes by mutating the validator state on the manager type and checking for events afterwards.

vms/evm/plugin/validators/state/interfaces/state.go
```diff
@@ -0,0 +1,55 @@
...
+	AddValidator(vdr Validator) error
+	// UpdateValidator updates an existing validator in the state
+	UpdateValidator(vdr Validator) error
+	// DeleteValidator removes a validator from the state
+	DeleteValidator(vID ids.ID) error
+	// WriteState writes the validator state to the disk
+	WriteState() error
+	// RegisterListener registers a callback listener for state changes
+	RegisterListener(StateCallbackListener)
+}
```
joshua-kim: These interfaces don't seem to be used - can we just export the `manager` type and use that instead? The interfaces seem to just add more layers of indirection without any clear benefit.

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,361 @@
...
+
+var (
+	ErrAlreadyExists  = errors.New("validator already exists")
+	ErrImmutableField = errors.New("immutable field cannot be updated")
+)
+
+const (
+	updatedStatus dbUpdateStatus = true
+	deletedStatus dbUpdateStatus = false
+)
```
joshua-kim: 1. Why do we have this type alias?
2. Do we really need to create two aliases for the only two possible values of a bool (true/false)? This seems to add more layers of indirection for no clear benefit. I recall @StephenButtolph mentioned using a enum instead of this alias which is another idea that sounds reasonable.

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	if data.Weight != vdr.Weight {
+		data.Weight = vdr.Weight
+		updated = true
+	}
+
+	if updated {
+		s.updatedData[vdr.ValidationID] = updatedStatus
+	}
```
joshua-kim:
```suggestion
	s.updatedData[vdr.ValidationID] = updated
```

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,361 @@
...
+	for _, listener := range s.listeners {
+		listener.OnValidatorAdded(vdr.ValidationID, vdr.NodeID, vdr.StartTimestamp, vdr.IsActive)
+	}
+	return nil
+}
+
+// UpdateValidator updates the validator in the state
+// returns an error if the validator does not exist or if the immutable fields are modified
+func (s *state) UpdateValidator(vdr interfaces.Validator) error {
+	data, exists := s.data[vdr.ValidationID]
```
joshua-kim: This is used throughout the PR - but typically Go style uses `ok` instead of `exists` for this var name.
```suggestion
	data, ok := s.data[vdr.ValidationID]
```

vms/evm/plugin/validators/manager.go
```diff
@@ -0,0 +1,178 @@
...
+	if err := m.State.WriteState(); err != nil {
+		return fmt.Errorf("failed to write validator state: %w", err)
+	}
+
+	return nil
+}
+
+// DispatchSync starts the sync process
+// DispatchSync holds the given lock while performing the sync.
+func (m *manager) DispatchSync(ctx context.Context, lock sync.Locker) {
```
joshua-kim: 1. This feels like a smell... we already have a `lockedReader`, but `DispatchSync` takes in a lock as well. I'm wondering why we even have a `lockedReader` at all.

2. `DispatchSync` also feels really similar to `Initialize`. One idea we could consider is to just expect the caller (`vm`) to just implement the for loop/select body, and this api is just `Sync`. If we do this - we can get rid of `Initialize` as well since the caller can call `Sync` after setting the validator ids to update uptimes for.

3. I'm wondering if we need to use the VM lock (it's not clear to me yet). If not, we could define the lock on this type instead.

vms/evm/plugin/validators/manager.go
joshua-kim, line 1: This package (everything under `validators/`) has a ton of sprawl - I think we could reasonably put this under one file. We need to figure out a package name/type name.... maybe something like `evm/state/validators.go`? I'm not sure what `plugin` means in this context.

vms/evm/plugin/validators/locked_reader.go
JonathanOppenheimer: We can remove the locked reader type entirely, and just pass the lock in when needed.

vms/evm/plugin/validators/README.md
JonathanOppenheimer: I am going to split the comments from this document into all the relevant places - comments in line.

vms/evm/plugin/validators/state/state.go
JonathanOppenheimer: 1. Idk 
2. TIL that Go does not have enums, but instead,`iota`. Curious as to why no enums! I have made the change.

vms/evm/plugin/validators/manager.go
JonathanOppenheimer: 1. `lockedReader` is removed entirely in favor of the passed in lock. 
2. Implemented 
3. I will leave the VM lock for now

vms/evm/plugin/validators/state/state.go
joshua-kim: Enum and iota can be thought  of as pretty the same thing - just different names. If someone refers to an enum they mean iota.

vms/evm/plugin/validators/state/interfaces/mock_listener.go
JonathanOppenheimer: I removed all the mocked tests - let me know if you're a fan of it now.

vms/evm/plugin/validators/manager.go
JonathanOppenheimer: Let's discuss the structure after the code clean up first - think it'll make more sense then as I'm sure a lot of stuff will change / get deleted etc. 

`plugin` is the name from the EVM repositories.

vms/evm/plugin/validators/uptime/pausable_manager.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/ava-labs/avalanchego/vms/evm/plugin/validators/state"
+)
+
+var errPausedDisconnect = errors.New("paused node cannot be disconnected")
+
+type PausableManager interface {
+	uptime.Manager
+	state.StateCallbackListener
+	IsPaused(nodeID ids.NodeID) bool
+}
```
joshua-kim: This interface looks premature - we only have one implementation of it. Should we remove?

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,389 @@
...
+		updated = updatedStatus
+	}
+
+	s.updatedData[vdr.ValidationID] = updated
+	return nil
+}
+
+// DeleteValidator marks the validator as deleted
+// marked validator will be deleted from disk when WriteState is called
+func (s *State) DeleteValidator(vID ids.ID) error {
```
joshua-kim: This return value is never used in the calling code... should we remove it?

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,389 @@
...
+	data, ok := s.data[vID]
+	if !ok {
+		return database.ErrNotFound
```
joshua-kim: Commenting here but this is applicable to all of our CRUD apis on this type - we return this sentinel value which is weird to me because:
1. We depend on the `database` package, but these are all in-memory operations
2. We never care about this error type in the calling code - we just use it as a flag to show that the key being looked up wasn't there. We could just use a `bool` for this instead, which is more idiomatic w/ other crud operation apis (not even just in Go) where you might have a signature like `Delete(k Key) bool` that tells you if the key was actually present.

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,389 @@
...
+	clear(s.updatedData)
+	return nil
+}
+
+// SetStatus sets the active status of the validator with the given vID
+// For L1 validators, an `active` status implies the validator balance on the P-Chain is
+// sufficient to cover the continuous validation fee. When an L1 validator balance is
+// depleted, it is marked as `inactive` on the P-Chain and this information is passed to the
+// Subnet-EVM's state.
+func (s *State) SetStatus(vID ids.ID, isActive bool) error {
```
joshua-kim: Is this dead code? For some reason I can't find where this is used...

vms/evm/plugin/validators/uptime/pausable_manager.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/vms/evm/plugin/validators/state"
+)
+
+var errPausedDisconnect = errors.New("paused node cannot be disconnected")
+
+type PausableManager interface {
+	uptime.Manager
+	state.StateCallbackListener
+	IsPaused(nodeID ids.NodeID) bool
```
joshua-kim: Why do we have this function? It doesn't have usages outside of testing. I think we can still get coverage over what it's intended to test by checking the uptime.

vms/evm/plugin/validators/manager_test.go
```diff
@@ -0,0 +1,272 @@
...
+			wantAddedValidators: map[ids.ID]ids.NodeID{
+				testValidationIDs[0]: testNodeIDs[0], // Only initial validator
+			},
+			wantRemovedValidators: map[ids.ID]ids.NodeID{},
+			wantStatusUpdates:     map[ids.ID]bool{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(tt *testing.T) {
```
joshua-kim: Some naming conventions:
1. Not really a nit - in Go `*testing.T` is pretty much always named `t` (I've never seen it named anything other than this)
2. nit: `tc` is typically `tt`... I've seen `tc` but it's less common
3. in-between a nit and not a nit: `testCases` is typically `tests`

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// GetNodeIDs returns the node IDs of validators in the state
+func (s *State) GetNodeIDs() set.Set[ids.NodeID] {
+	ids := set.NewSet[ids.NodeID](len(s.index))
+	for nodeID := range s.index {
+		ids.Add(nodeID)
+	}
+	return ids
+}
```
joshua-kim: I wonder if it just makes sense to have some single api like `GetFoo` that returns all of the staking metadata together instead of having signatures for each of these.

vms/evm/plugin/validators/state/state.go
```diff
@@ -0,0 +1,389 @@
...
+
+// StateCallbackListener is a listener for the validator state
+type StateCallbackListener interface {
+	// OnValidatorAdded is called when a new validator is added
+	OnValidatorAdded(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool)
+	// OnValidatorRemoved is called when a validator is removed
+	OnValidatorRemoved(vID ids.ID, nodeID ids.NodeID)
+	// OnValidatorStatusUpdated is called when a validator's active status changes
+	OnValidatorStatusUpdated(vID ids.ID, nodeID ids.NodeID, isActive bool)
+}
```
joshua-kim: I think this interface exists because of the package structure - we need some way to pass in our uptime tracking so it can know about validator add/removal events.

I think if we re-structure our packages, we can probably get rid of a lot of complexity. My current understanding of this PR Is that it has:
* `validators` - defines an abstraction that keeps the EVM's view of the validator set sync'd with the P-Chain. Knows about the `State` since it needs to write that info somewhere, and `PausableManager` since it needs to be able to update uptimes.
* `state` - Defines on-disk representation of validator metadata - exposes listeners so that you can hook into events of when validators are added/deleted/updated so that `PausableManager` can get updates
* `uptime` - Defines uptime calculations

From what I can tell, the abstraction that the VM actually needs to know about is something that lets us stay in sync with the P-Chain, and some way of getting the uptime for a validator to figure out if we should reward it or not. I think therefore, we should merge all of these packages into one package and just call it something like `uptimetracker` or something. We can expose apis to sync it + get the uptime for a validator, and i think all other code should be able to be unexported.

vms/evm/plugin/validators/uptime/pausable_manager.go
JonathanOppenheimer: I've turned it into a struct - would you like something else?

vms/evm/plugin/validators/state/state.go
JonathanOppenheimer: The `error`? It is used. 

e.g. 
```
    // Remove validators no longer in the current set
	for vID := range currentValidationIDs {
		if _, exists := newValidators[vID]; !exists {
			if err := m.State.DeleteValidator(vID); err != nil {
				return fmt.Errorf("failed to delete validator %s: %w", vID, err)
			}
		}
	}
```

vms/evm/plugin/validators/uptime/pausable_manager.go
JonathanOppenheimer: I removed it.

vms/evm/plugin/validators/state/state.go
JonathanOppenheimer: I am not sure how I feel about that assuming you're proposing something like 

```func (s *State) GetMetadata() (set.Set[ids.ID], set.Set[ids.NodeID]) ```

as the uses of these two functions are always just looking for one of the information types, so you'd end up doing a lot of `_` as the other part isn't needed. Passing more information than is needed seems bad.

Comment (joshua-kim) [reply vms/evm/plugin/validators/state/state.go]: I don't mean returning multiple return values, I mean returning a struct like `validatorData` or something of a similar shape.

File: vms/evm/plugin/validators/state/state.go (hunk 12 above)
Comment (joshua-kim) [reply vms/evm/plugin/validators/state/state.go]: When I say it's not used I mean that the error itself is not used - this is equivalent to:
``` 
   // Remove validators no longer in the current set
	for vID := range currentValidationIDs {
		if _, exists := newValidators[vID]; !exists {
			if !m.State.DeleteValidator(vID) {
				return fmt.Errorf("failed to delete validator %s", vID)
			}
		}
	}
```

The error is used to indicate that the validator wasn't there to delete - which can be signaled with a bool since only return two possible errors (`database.ErrNotFound`/`nil`)

File: vms/evm/uptimetracker/manager_test.go
Comment (joshua-kim) [inline vms/evm/uptimetracker/manager_test.go:1]: Now that we've created a single package (and in Go, the package is the abstraction, not a unit of organization), can we try to get something into a single `UptimeTracker` struct? I think in-line with my previous comment, we can probably get something that just has apis required for syncing + uptime calculations and we should be able to unexport a lot of the code here to slim down the package api.

File: vms/evm/uptimetracker/state.go [hunk 19]
```diff
@@ -0,0 +1,393 @@
...
+
+	// Notify the callbacks of current state (matches old RegisterListener behavior)
+	if s.onAdded != nil {
+		for vID, data := range s.data {
+			s.onAdded(vID, data.NodeID, data.StartTime, data.IsActive)
+		}
+	}
+}
+
+// SetCallbacks sets custom callbacks. Intended for tests.
```
joshua-kim: Test-specific code should be avoided - if our api isn't testable we should refactor. Mutable state (`Set*`) is also something we should avoid anyways.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,393 @@
...
+func (s *state) SetCallbacks(
+	onAdded func(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool),
+	onRemoved func(vID ids.ID, nodeID ids.NodeID),
+	onStatusUpdated func(vID ids.ID, nodeID ids.NodeID, isActive bool),
+) {
+	s.onAdded = onAdded
+	s.onRemoved = onRemoved
+	s.onStatusUpdated = onStatusUpdated
+
+	// Notify the callbacks of current state (matches old RegisterListener behavior)
```
joshua-kim: Why do we have these comments? It's not relevant to the maintainer what the old behavior was.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,393 @@
...
+	data  map[ids.ID]*validatorData // vID -> validatorData
+	index map[ids.NodeID]ids.ID     // nodeID -> vID
+	// updatedData tracks the updates since WriteValidator was last called
+	updatedData map[ids.ID]dbUpdateStatus // vID -> updated status
+	db          database.Database
+
+	// Callback hooks for validator state changes
+	onAdded         func(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool)
+	onRemoved       func(vID ids.ID, nodeID ids.NodeID)
+	onStatusUpdated func(vID ids.ID, nodeID ids.NodeID, isActive bool)
```
joshua-kim: If we put things under a single `UptimeTracker` type we should be able to avoid callback complexity

vms/evm/plugin/validators/state/state.go
JonathanOppenheimer: Even so, isn't that still unnecessary? - you're only ever going to use one of the struct's fields - I don't see the benefit of doing so - I think I'm missing something

vms/evm/uptimetracker/manager_test.go
JonathanOppenheimer: Yup - I think this is a lot cleaner now!

vms/evm/uptimetracker/state.go
JonathanOppenheimer: There's now two tests, for this one - let me know what you think.

snow/uptime/manager.go
```diff
@@ -114,6 +114,14 @@ func (m *manager) Disconnect(nodeID ids.NodeID) error {
...
 }
 
+// The `CalculateUptime` method calculates a node's uptime based on its connection status,
+// connected time, and the current time. It first retrieves the node's current uptime and
+// last update time from the state, returning an error if retrieval fails. If tracking hasn’t
+// started, it assumes the node has been online since the last update, adding this duration
+// to its uptime. If the node is not connected and tracking is `active`, uptime remains
+// unchanged and returned. For connected nodes, the method ensures the connection time does
+// not predate the last update to avoid double counting. Finally, it adds the duration since
+// the last connection time to the node's uptime and returns the updated values.
```
joshua-kim: Unrelated diff hunk - should we pull this out of the diff?

vms/evm/uptimetracker/codec.go
joshua-kim, line 1: Do we need a separate file for this? We could keep the code closer to where it's used. The codec handles marshaling/unmarshaling, so at the top of the file where `state` is defined might make sense.

vms/evm/uptimetracker/pausable_manager.go
```diff
@@ -0,0 +1,149 @@
...
+type pausableManager struct {
+	manager    uptime.Manager
+	pausedVdrs set.Set[ids.NodeID]
+	// connectedVdrs is a set of nodes that are connected to the manager.
+	// This is used to immediately connect nodes when they are unpaused.
+	connectedVdrs set.Set[ids.NodeID]
+}
+
+// NewPausableManager takes an uptime.Manager and returns a PausableManager
+func newPausableManager(manager uptime.Manager) *pausableManager {
```
joshua-kim: w/ the new api refactor, this type doesn't need to exist anymore and it's just adding more layers of indirection. It should be possible to just merge this into the `UptimeTracker` + this helps get rid of some file sprawl.

vms/evm/uptimetracker/pausable_manager_test.go
```diff
@@ -0,0 +1,323 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/utils/timer/mockable"
+)
+
+func TestPausableManager(t *testing.T) {
```
joshua-kim: These are now testing implementation details - `pausableManager` doesn't exist from the perspective of the api consumer. We want to rewrite all of these tests (applicable to `state_test` as well) to go through the `UptimeManager` public api.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/snow/validators"
+	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/utils/timer/mockable"
+)
+
+const (
+	// SyncFrequency is the recommended frequency for calling Sync()
+	// VMs can use this as a default but are free to choose their own timing
+	SyncFrequency = 1 * time.Minute
```
joshua-kim: Should this be defined by the consumer/deleted from this package? This default isn't going to be re-used anyways - we only have one consumer using it (evm). Our api technically doesn't know about frequencies of calling sync - it just defines how to sync so explicitly documenting how frequently to call it feels like it's leaking implementation details of the `evm` event loop of re-syncing the P-Chain into this package which just cares about uptimes.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+// The UptimeTracker struct is responsible for managing the state of validators by fetching
+// information from P-Chain state (via GetCurrentValidatorSet in chain context) and updating
+// the local state accordingly. The caller is responsible for calling Sync() periodically to
+// keep the state up-to-date. The sync operation first removes validators that are no longer
+// in the P-Chain validator set, then adds new validators and updates existing validators.
+// This order of operations ensures that uptimes of validators being removed and re-added
+// under the same nodeIDs are updated in the same sync operation despite having different
+// validationIDs.
```
joshua-kim: This comment is so verbose + leaky. Why do we care about things like:
> The sync operation first removes validators that are no longer
// in the P-Chain validator set, then adds new validators and updates existing validators.
// This order of operations ensures that uptimes of validators being removed and re-added
// under the same nodeIDs are updated in the same sync operation despite having different
// validationIDs.

This comment is just reciting the implementation of `Sync` (so it doesn't even belong here, it belongs on `Sync` - and even there it would still be leaky).

I think a struct-level comment is appropriate here, but we should aim to have something that tells us what this type's responsibility is - not what the code does at a line-by-line-level.

vms/evm/plugin/validators/state/state.go
joshua-kim: The calling patterns that both of these apis are being used for are for iterations over each validation. We expose two different apis since some care about the validation id, while others care about the node id. If we're letting callers iterate over our state, at that point I'm asking if it just makes more sense for us to export a type that has that information (like `validatorData`) so we we can have a single `GetValidatorDatas` api instead of having both `GetNodeIDs` and `GetValidationIDs` (both of which are just subsets of the `validatorData` structs).

vms/evm/uptimetracker/pausable_manager.go
```diff
@@ -0,0 +1,149 @@
...
+// The AvalancheGo uptime manager records the time when a peer is connected to the tracker node.
+// When a paused/`inactive` validator is connected, the pausable uptime manager does not directly
+// invoke the `Connected` method on the AvalancheGo uptime manager, thus the connection time is not
+// directly recorded. Instead, the pausable uptime manager waits for the validator to increase its
+// continuous validation fee balance and resume operation. When the validator resumes, the tracker
+// node records the resumed time and starts tracking the uptime of the validator.
+//
+// Note: The uptime manager does not check if the connected peer is a validator or not. It records
+// the connection time assuming that a non-validator peer can become a validator whilst they're
+// connected to the uptime manager.
```
joshua-kim: More comments that are leaky/verbose/irrelevant to the reader

vms/evm/uptimetracker/pausable_manager.go
```diff
@@ -0,0 +1,149 @@
...
+func (p *pausableManager) connect(nodeID ids.NodeID) error {
+	p.connectedVdrs.Add(nodeID)
+	if !p.isPaused(nodeID) && !p.manager.IsConnected(nodeID) {
+		return p.manager.Connect(nodeID)
+	}
+	return nil
+}
+
+// IsConnected returns true if the node with the given ID is connected to this manager
+// Note: Inner manager may have a different view of the connection status due to pausing
```
joshua-kim: Comment needs to be updated

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,291 @@
...
+			return err
+		}
+	}
+	return it.Error()
+}
+
+// addData adds the data to the state
+// returns an error if the data already exists
+func (s *state) addData(vID ids.ID, data *validatorData) error {
+	if _, exists := s.data[vID]; exists {
```
joshua-kim: nit: `ok` naming - we should do a pass over the PR for this since it's used in a lot of places (`exists`, and `f`)

vms/evm/uptimetracker/manager_test.go
joshua-kim, line 1: It also looks like all of these filenames need to be updated

snow/uptime/manager.go
JonathanOppenheimer: This comes from a readme file that will be deleted from subnet-evm. I just moved it to where it should be. If it's not moved now, it will cease to exist anywhere (which maybe is okay)

joshua-kim: I think it's fine to remove... the godoc is _really_ wordy (it's literally a translation of code into a comment). These types of docs aren't useful because if a reader cared about the information at this level, they could just read the code. It also adds maintenance burden because we have to keep it in sync w/ the code, which is difficult at this granularity.

The existing code is also reasonably documented + is simple enough such that it's understandable without extensive documentation. If we wanted a function-level godoc (which I don't think is a bad idea) we could add one - but it might make more sense at an interface-level instead of on this implementation because subnet evm implements the interface.

JonathanOppenheimer: Sounds good! I'll remove it.

vms/evm/plugin/validators/state/state.go
JonathanOppenheimer: Gotcha - updated.

vms/evm/uptimetracker/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+package uptimetracker
+
+import (
+	"time"
+
+	"github.com/ava-labs/libevm/metrics"
+)
+
+// Metrics defines the interface for uptime tracker metrics
+type Metrics interface {
```
joshua-kim: Why are we adding metrics as part of this PR? It seems like this file should be removed from this diff

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,315 @@
...
+		return nil, fmt.Errorf("failed to load data from disk: %w", err)
+	}
+	return s, nil
+}
+
+// GetUptime returns the uptime of the validator with the given nodeID
+func (s *state) GetUptime(
+	nodeID ids.NodeID,
+) (time.Duration, time.Time, error) {
+	data, f := s.getData(nodeID)
```
joshua-kim: ref: https://github.com/ava-labs/avalanchego/pull/4166#discussion_r2334143676

vms/evm/uptimetracker/state_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+)
+
+func TestState(t *testing.T) {
+	require := require.New(t)
+	db := memdb.New()
+	state, err := newState(db)
```
joshua-kim: ref: https://github.com/ava-labs/avalanchego/pull/4166#discussion_r2334098939

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,315 @@
...
+const (
+	codecVersion  uint16         = 0
+	updatedStatus dbUpdateStatus = iota
+	deletedStatus
+)
+
+var (
+	ErrAlreadyExists  = errors.New("validator already exists")
+	ErrImmutableField = errors.New("immutable field cannot be updated")
+)
```
joshua-kim: Can we put stuff in a single const + var block? It makes it hard to track global state if we have multiple var/const scopes around the package.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,315 @@
...
+
+type state struct {
+	data  map[ids.ID]*validatorData // vID -> validatorData
+	index map[ids.NodeID]ids.ID     // nodeID -> vID
+	// updatedData tracks the updates since WriteValidator was last called
+	updatedData map[ids.ID]dbUpdateStatus // vID -> updated status
+	db          database.Database
+}
+
+func init() {
```
joshua-kim: init typically lives at the top of the file

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,315 @@
...
+	IsActive       bool       `json:"isActive"`       // Whether validator is currently active
+	IsL1Validator  bool       `json:"isL1Validator"`  // Whether this is an L1 validator
+}
+
+// The state implementation only allows existing validator's `weight` and `IsActive`
+// fields to be updated; all other fields should be constant and if any other field
+// changes, the state manager errors and does not update the validator.
+//
+// The state implementation also assumes NodeIDs are unique in the tracked set.
+type validatorData struct {
```
joshua-kim: This should live closer to the functions that have receivers defined on it

vms/evm/uptimetracker/validator_test.go
```diff
@@ -0,0 +1,624 @@
...
+				require.Equal(newVdr.IsL1Validator, vdrData.IsL1Validator)
+			}
+		})
+	}
+}
+
+func TestSync_ValidatorLifecycle(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
joshua-kim: We should avoid using any mocks in this package

vms/evm/uptimetracker/validator_test.go
```diff
@@ -0,0 +1,624 @@
...
+				// We should be optimistic and assume the node was online and active until we start tracking
+				checkUptime(t, up, nodeID0, wantUptime, currentTime)
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			u, clk, _ := setupTestEnv(t, nodeID0, startTime)
+			test.testFunc(t, u, clk)
```
joshua-kim: If the test-table is just a function, we're basically writing tests in each test table definition. At that point - we should just write separate tests.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,329 @@
...
+	// This is used to immediately connect nodes when they are unpaused.
+	connectedVdrs set.Set[ids.NodeID]
+}
+
+// NewUptimeTracker returns a new validator state
+// that manages the validator state and the uptime manager.
+// ValidatorState is not thread safe and should be used with the VM locked.
+func NewUptimeTracker(
+	ctx *snow.Context,
+	manager uptime.Manager,
```
joshua-kim: It feels weird for us to know about `snow.Context` (overkill since we only care about the P-Chain), and `uptime.Manager` (the `UptimeTracker` is the source of truth for uptimes, but this api exposes another source of truth for uptimes). I think we can just accept a `validators.State` here and create the `uptime.Manager` type within this function's scope.

vms/evm/uptimetracker/validator.go
joshua-kim, line 1: Should this be called `uptime_tracker.go`? `validator.go` seems out-of-place since the abstraction being exposed is the `UptimeTracker`.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,329 @@
...
+		return errors.New("failed to write validator state")
+	}
+
+	return nil
+}
+
+// Sync synchronizes the validator state with the current validator set and writes the state to the database.
+// Sync is not safe to call concurrently and should be called with the VM locked.
+func (u *UptimeTracker) Sync(ctx context.Context) error {
+	start := time.Now()
```
joshua-kim: I don't love the usage of `mockable.Clock` - but I'd argue that it makes sense to use here because this type `u` already depends on `mockable.Clock` from the constructor - so suddenly deciding to use `time.Now()` creates inconsistent behavior.

The alternative to using mock-time is to pass timestamps in as parameters (which I do think looks better), but since `uptime.Manager` (existing code) already uses the mock-clock we don't have a choice but to use it in this code.

vms/evm/uptimetracker/state.go
joshua-kim, line 1: Would it make sense for us to remove the `state` type entirely and just in-line it into `UptimeTracker`?  I feel like this state does just belong to the `UptimeTracker` and breaking it into its own struct doesn't actually simplify anything. It seems like in most/all cases the functions on this struct are only called in one non-test location, so we could just in-line usage of it. It also forces us to write tests against the public api.

vms/evm/uptimetracker/metrics.go
JonathanOppenheimer: Ah I just saw the TODO and thought we might as well, while we're going through everything here.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,329 @@
...
+		err := u.resume(nodeID)
+		if err != nil {
+			log.Error("failed to handle validator removed %s: %s", nodeID, err)
+		}
+	}
+}
+
+// OnValidatorStatusUpdated is called when the status of a validator is updated.
+// If the node is active, it will be resumed. If the node is inactive, it will be paused.
+func (u *UptimeTracker) onValidatorStatusUpdated(_ ids.ID, nodeID ids.NodeID, isActive bool) {
```
joshua-kim: This looks like vestigial code from callback hell we previously had - could we just in-line usage of it? Generally I dislike these types of functions when they wrap trivial logic + are used in only one place because they add more layers of indirection.

vms/evm/uptimetracker/validator.go
```diff
@@ -0,0 +1,329 @@
...
+	} else {
+		err = u.pause(nodeID)
+	}
+	if err != nil {
+		log.Error("failed to update status for node %s: %s", nodeID, err)
+	}
+}
+
+// updateValidatorState updates the local validator state to match the current validator set.
+func (u *UptimeTracker) updateValidatorState(newValidators map[ids.ID]*validators.GetCurrentValidatorOutput) error {
```
joshua-kim: This could also be in-lined into `Sync`

vms/evm/uptimetracker/state.go
JonathanOppenheimer: It'll get pretty bulky, but sure!

vms/evm/uptimetracker/metrics.go
joshua-kim: I think we can just leave the TODO for now - I think the current scope of this PR should be to get use-case parity with the existing `validators` package. I'm happy to add metrics - but we should scope it to a follow-up.

vms/evm/uptimetracker/validator.go
JonathanOppenheimer: We also need to pass a subnetID variable as well.

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -0,0 +1,1057 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer, line 1: breaking up the big table tests made this file very long.

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -0,0 +1,1057 @@
...
+}
+
+func (t *testValidatorState) GetCurrentValidatorSet(ctx context.Context, subnetID ids.ID) (map[ids.ID]*validators.GetCurrentValidatorOutput, uint64, error) {
+	return t.validators, uint64(len(t.validators)), nil
+}
+
+// setupUptimeTracker is a helper function to setup a UptimeTracker for testing
+// if validatorSet is nil, it will be initialized to an empty map.
+// if db is nil, it will be initialized to a new memory database.
+func setupUptimeTracker(t *testing.T, validatorSet map[ids.ID]*validators.GetCurrentValidatorOutput, db database.Database) *UptimeTracker {
```
JonathanOppenheimer: I tried to do the go ... optional arguments, but I couldn't get it nicely, so I just passed in `nil` to activate the default values.

vms/evm/uptimetracker/uptime_tracker_test.go
JonathanOppenheimer: if there's any tests you want to remove, lmk - I feel like 22 something tests for this is kinda absurd.

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -0,0 +1,1045 @@
...
+// testValidatorState is a minimal implementation that only provides what we actually use
+type testValidatorState struct {
+	validators map[ids.ID]*validators.GetCurrentValidatorOutput
+}
+
+func (t *testValidatorState) GetCurrentValidatorSet(_ context.Context, _ ids.ID) (map[ids.ID]*validators.GetCurrentValidatorOutput, uint64, error) {
+	return t.validators, uint64(len(t.validators)), nil
+}
+
+// Stub implementations for the rest of the interface (unused in our tests)
```
JonathanOppenheimer: taking suggestions on how to get rid of this / `testValidatorState` if possible.

vms/evm/uptimetracker/metrics.go
JonathanOppenheimer: Sounds good - I put a TODO at the top of the UptimeTracker.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,522 @@
...
+		data.IsActive = vdr.IsActive
+		updated = updatedStatus
+	}
+
+	if data.Weight != vdr.Weight {
+		data.Weight = vdr.Weight
+		updated = updatedStatus
+	}
+
+	u.updatedData[vdr.ValidationID] = updated
```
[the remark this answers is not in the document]
JonathanOppenheimer: Wait is this correct? Shouldn't we update `updatedData` regardless of whether or not the data was updated?

joshua-kim (approved):

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type dbUpdateStatus int
+
+type validator struct {
+	UpDuration    time.Duration `serialize:"true"`
+	LastUpdated   uint64        `serialize:"true"`
+	NodeID        ids.NodeID    `serialize:"true"`
+	Weight        uint64        `serialize:"true"`
+	StartTime     uint64        `serialize:"true"`
```
joshua-kim, line 50: We have these fields because we used to use this to serve the `GetCurrentValidators` api on the evm, but since the new api for this package just cares about uptimes and doesn't serve this data we could consider removing it. I'm not sure how this behaves with our codec/if it's safe to do though @StephenButtolph .

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+}
+
+type dbUpdateStatus int
+
+type validator struct {
+	UpDuration    time.Duration `serialize:"true"`
+	LastUpdated   uint64        `serialize:"true"`
+	NodeID        ids.NodeID    `serialize:"true"`
+	Weight        uint64        `serialize:"true"`
```
joshua-kim, line 45: We have these fields because we used to use this to serve the `GetCurrentValidators` api on the evm, but since the new api for this package just cares about uptimes and doesn't serve this data we could consider removing it. I'm not sure how this behaves with our codec/if it's safe to do though @StephenButtolph .

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,245 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse validation ID: %w", err)
+		}
+
+		vdr := &validator{
+			validationID: validationID,
+		}
+
+		if _, err := codecManager.Unmarshal(it.Value(), vdr); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal validator validator: %w", err)
```
joshua-kim: nit: typo in in error message

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,245 @@
...
+func (s *state) getValidatorByNodeID(nodeID ids.NodeID) (*validator, bool) {
+	validationID, ok := s.nodeIDsToValidationIDs[nodeID]
+	if !ok {
+		return nil, false
+	}
+
+	v, ok := s.validators[validationID]
+	if !ok {
+		return nil, false
+	}
```
joshua-kim: This `ok` check is redundant since `s.nodeIDsToValidationIDs` is guaranteed to have the same data as in `s.validators` - we should be able to just ignore this return value.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,267 @@
...
+			return fmt.Errorf("failed to update validator set: %w", err)
+		}
+	}
+
+	// Initialize uptimes if this is the first time Sync has been called
+	if !u.synced {
+		validationIDs := make([]ids.NodeID, 0)
+		validationIDs = append(validationIDs, u.state.getNodeIDs()...)
+
+		if err := u.manager.StartTracking(validationIDs); err != nil {
```
joshua-kim: I -think- I wrote this (it looks like code I would write)? Not sure why though.
```suggestion
		if err := u.manager.StartTracking(u.state.getNodeIDs()); err != nil {
```

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,267 @@
...
+		// This validator is still in the latest update
+		if _, ok := newValidators[validationID]; ok {
+			continue
+		}
+
+		// This validator was removed in the lastest update
+		nodeID, ok := u.state.getNodeID(validationID)
+		if !ok {
+			return fmt.Errorf("failed to fetch validator %s", validationID)
+		}
```
joshua-kim: This looks scary but we're guaranteed to have this information so we can skip the check
```suggestion
                 // we are guaranteed to have this node id
		nodeID, _ := u.state.getNodeID(validationID)
```

vms/evm/uptimetracker/state.go
joshua-kim: I looked over the codec impl and this doesn't look safe to remove.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,267 @@
...
+
+			// This validator is no longer active and is treated as offline
+			if err := u.deactivate(newValidator.NodeID); err != nil {
+				return fmt.Errorf(
+					"failed to deactivate validator %s: %w",
+					newValidator.NodeID,
+					err,
+				)
+			}
+		} else {
```
joshua-kim: I think we could just add a `continue` here to do an early-exit? if/else is usually harder to read than early-exit code.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,245 @@
...
+	v := s.validators[validationID]
+
+	updated := false
+	if v.IsActive != isActive {
+		v.IsActive = isActive
+		s.updatedValidators.Add(validationID)
+		updated = true
+	}
+
+	return updated
```
joshua-kim: We can remove a var here. Also going to suggest inverting the if to make the early-exit code simpler.
```suggestion
	if v.IsActive == isActive {
	  return false
	}

	v.IsActive = isActive
	s.updatedValidators.Add(validationID)
        return true
```

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,245 @@
...
+	*validator,
+	bool,
+) {
+	v, ok := s.validators[validationID]
+	if !ok {
+		return &validator{}, false
+	}
+
+	return v, true
+}
```
joshua-kim: We never use this return value - we could remove it:
```suggestion
func (s *state) hasValidationID(validationID ids.ID) bool {
	_, ok := s.validators[validationID]
	return ok
}
```

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,234 @@
...
+	s.nodeIDsToValidationIDs[validator.NodeID] = validationID
+}
+
+func (s *state) getValidatorByNodeID(nodeID ids.NodeID) (*validator, bool) {
+	validationID, ok := s.nodeIDsToValidationIDs[nodeID]
+	if !ok {
+		return nil, false
+	}
+
+	// s.nodeIDsToValidationIDs is guaranteed to have the same data as in s.validators
```
joshua-kim: Comments have a maintenance cost associated with them so we should avoid coupling code identifiers to comments - this stuff isn't indexed by the IDE and even if it was it's hard to keep in-sync w/ the code. 
```suggestion
	// we are guaranteed to have this validator
```

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,234 @@
...
+
+	delete(s.validators, v.validationID)
+	delete(s.nodeIDsToValidationIDs, v.NodeID)
+
+	s.deletedValidators.Add(v.validationID)
+
+	return true
+}
+
+func (s *state) writeState() error {
```
StephenButtolph: `state.writeState` seems weird. Should this be:
```suggestion
func (s *state) writeModifications() error {
```
or `flush` or something?

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,234 @@
...
+		return time.Time{}, database.ErrNotFound
+	}
+
+	return time.Unix(int64(v.LastUpdated), 0), nil
+}
+
+func (s *state) addNewValidator(vdr *validator) {
+	s.addValidator(vdr.validationID, vdr)
+	s.updatedValidators.Add(vdr.validationID)
+}
```
StephenButtolph, line 139: I feel like `addValidator` and `addNewValidator` are not self documenting. What are the different usages of these functions?

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (s *state) updateValidator(validationID ids.ID, isActive bool) bool {
```
StephenButtolph, line 144: Should we document that this function assumes the validator currently exists?

If this is called when the validator doesn't exist it will panic... Which is different than `deleteValidator`

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,263 @@
...
+)
+
+// UptimeTracker tracks uptime information for validators
+type UptimeTracker struct {
+	validatorState validators.State
+	subnetID       ids.ID
+	manager        uptime.Manager
+
+	lock                sync.Mutex
+	height              int
```
StephenButtolph: It seems like we use `height` as an `int` rather than a `uint64` so that we can signal `-1` as not populated, but we already have a flag `synced` to signal this... Can we just use the flag?

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf("failed to write state: %w", err)
+	}
+
+	return nil
+}
+
+func (u *UptimeTracker) update(
+	height uint64,
+	currentValidatorSet map[ids.ID]*validators.GetCurrentValidatorOutput,
+) error {
```
StephenButtolph, line 149: Would it make sense not to pass in `height` into this function and update it in the if statement where we check the height?

It feels like we should either include the `if` statement in this function or we should move the height management entirely out of this function.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -0,0 +1,263 @@
...
+		// This validator was removed in the lastest update
+		// we are guaranteed to have this node id
+		nodeID, _ := u.state.getNodeID(validationID)
+
+		if !u.state.deleteValidator(validationID) {
+			return fmt.Errorf("failed to delete validator %s", validationID)
+		}
+
+		u.deactivatedValidators.Remove(nodeID)
+	}
```
StephenButtolph: Why do we have a `getValidationIDs` helper? Can't we just iterate over the map directly?
```suggestion
	newValidators := currentValidatorSet

	for validationID, validator := range u.state.validators {
		// This validator is still in the latest update
		if _, ok := newValidators[validationID]; ok {
			continue
		}

		if !u.state.deleteValidator(validationID) {
			return fmt.Errorf("failed to delete validator %s", validationID)
		}

		u.deactivatedValidators.Remove(validator.NodeID)
	}
```

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -0,0 +1,619 @@
...
+)
+
+func TestUptimeTracker_GetUptime(t *testing.T) {
+	tests := []struct {
+		name string
+		// invariant: arrays must all be of the same length
+		timestamps             []time.Time
+		validators             [][]validators.GetCurrentValidatorOutput
+		connectedValidators    [][]ids.NodeID
+		disconnectedValidators [][]ids.NodeID
```
StephenButtolph: I feel like this invariant is a bit of a smell, can we make a struct which includes the `timestamp` + `validators` + `connectedValidators` + `disconnectedValidators` and then just have a slice of that struct type? It should allow us to remove a decent number of empty slice initializations as well.

JonathanOppenheimer: yeah this is a big clean up

vms/evm/uptimetracker/uptime_tracker.go
JonathanOppenheimer: Shouldn't we actually export this function?

https://github.com/ava-labs/subnet-evm/blob/87c8cfca01e6fc58d76cefd44dde873f4a3fecc6/plugin/evm/service.go#L36

I'm not sure what I would replace it with if we delete it.

vms/evm/uptimetracker/uptime_tracker.go
JonathanOppenheimer: I think it simplifies the sync function if we do the if statement check here -- let me know if you'd prefer the other proposed option.

vms/evm/uptimetracker/uptime_tracker.go
StephenButtolph: I think the expectation is that you use the existing validators.State and call `GetCurrentValidators`

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,242 @@
...
+// for persistence to the database. This should be used when adding validators during
+// runtime operations.
+func (s *state) addNewValidator(vdr *validator) {
+	s.addValidatorToMemory(vdr.validationID, vdr)
+	s.updatedValidators.Add(vdr.validationID)
+}
+
+// updateValidator updates the validator with the given validationID to the
+// given isActive state -- this function does assume that a validator with t
+// he given validationID exists.
```
StephenButtolph: The comment doesn't read correctly.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,242 @@
...
+	if !ok {
+		return time.Time{}, database.ErrNotFound
+	}
+
+	return time.Unix(int64(v.LastUpdated), 0), nil
+}
+
+// addNewValidator adds a new validator to the state and marks it
+// for persistence to the database. This should be used when adding validators during
+// runtime operations.
```
StephenButtolph: The newlines here look a bit weird. Why is the first line so short and the second line so long?
```suggestion
// addNewValidator adds a new validator to the state and marks it for
// persistence to the database. This should be used when adding validators
// during runtime operations.
```
This isn't something that is enforced throughout the codebase... So I don't usually comment on it... But this just looked particularly weird to me.

When I write code I try to stick to the 80 character style... But again we don't enforce this in the codebase.

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	v.IsActive = isActive
+	s.updatedValidators.Add(validationID)
+	return true
+}
+
+// deleteValidator deletes the validator with the given validationID -- this
+// function does assume that a validator with the given validationID exists.
+func (s *state) deleteValidator(validationID ids.ID) bool {
```
StephenButtolph: This function always returns `true`... Seems like a simplification can be made

vms/evm/uptimetracker/state.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	return v.NodeID, true
+}
+
+func (s *state) getNodeIDs() []ids.NodeID {
+	return maps.Keys(s.nodeIDsToValidationIDs)
+}
+
+func (s *state) getValidationIDs() []ids.ID {
```
StephenButtolph: This is dead code (again, surprised this isn't caught by a linter)

vms/evm/uptimetracker/state.go
JonathanOppenheimer: We should enforce this!

vms/evm/uptimetracker/state.go
JonathanOppenheimer: removed - saw your other comment.

vms/evm/uptimetracker/state.go
JonathanOppenheimer: Is this better?

> // updateValidator sets the isActive state of the validator with the given
// validationID -- this function assumes that a validator with the given
// validationID exists.

PR #4170, opened 2025-08-13
Cache validator sets retrieved by client

geoff-vball:


snow/validators/state.go
```diff
@@ -114,3 +145,64 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+		return nil, err
+	}
+
+	// cache the validator set
+	c.cache.Put(height, validatorSets)
+
+	return validatorSets, nil
+}
+
+func (c *CachedState) GetValidatorSet(
```
geoff-vball: In the case that the `State` is a wrapped state like `tracedState`, this doesn't actually call the underlying `GetValidatorSet` method - will that cause any problems for us?

iansuvak: `tracedState` is a wrapper that needs access to a different state implementation which would presumably make the relevant calls itself at a level no?

geoff-vball: I think in the case you wrap `CachedState -> TracedState -> State`,  if you call `GetValidatorSet` the trace would be tagged with `GetAllValidatorSets` because that would be what you'd be calling under the hood.

If you wrapped `TracedState -> CachedState -> State` then the trace would be tagged (correctly?) with `GetValidatorSet`

Don't think it's a big deal either way though

chains/manager.go
```diff
@@ -1115,6 +1115,9 @@ func (m *manager) createSnowmanChain(
 			return nil, fmt.Errorf("expected validators.State but got %T", vm)
 		}
 
+		// Wrap the validator state with a cached state
+		m.validatorState = validators.NewCachedState(m.validatorState)
```
michaelkaplan13: This needs to be `valState` or else it just gets overridden below it seems.

snow/validators/gvalidators/validator_state_client.go
```diff
@@ -55,6 +56,47 @@ func (c *Client) GetSubnetID(ctx context.Context, chainID ids.ID) (ids.ID, error
 	return ids.ToID(resp.SubnetId)
 }
 
+func (c *Client) GetAllValidatorSets(
+	ctx context.Context,
+	height uint64,
+) (map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput, error) {
```
yacovm: would it make sense to create a type for `map[ids.NodeID]*validators.GetValidatorOutput`?  

Perhaps `SubnetValidators` and then a type for `map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput` such as `SubnetsValidators` ?

Then things like the above, and `cache cache.Cacher[uint64, map[ids.ID]map[ids.NodeID]*GetValidatorOutput]` would be less verbose.

snow/validators/gvalidators/validator_state_client.go
```diff
@@ -103,36 +129,24 @@ func (c *Client) GetCurrentValidatorSet(
...
-			// PublicKeyFromValidUncompressedBytes is used rather than
-			// PublicKeyFromCompressedBytes because it is significantly faster
-			// due to the avoidance of decompression and key re-verification. We
-			// can safely assume that the BLS Public Keys are verified before
-			// being added to the P-Chain and served by the gRPC server.
-			publicKey = bls.PublicKeyFromValidUncompressedBytes(validator.PublicKey)
-			if publicKey == nil {
-				return nil, 0, errFailedPublicKeyDeserialize
-			}
+			return nil, 0, fmt.Errorf("failed to parse validator: %w", err)
```
yacovm: `fmt.Errorf("failed to parse validator (NodeID: %s, PublicKey: %s): %w",
					hex.EncodeToString(validator.NodeId), hex.EncodeToString(validator.PublicKey), err)` ?

snow/validators/gvalidators/validator_state_client.go
```diff
@@ -103,36 +129,24 @@ func (c *Client) GetCurrentValidatorSet(
 		SubnetId: subnetID[:],
 	})
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, fmt.Errorf("failed to get current validator set for subnet %s: %w", subnetID.String(), err)
 	}
 
 	vdrs := make(map[ids.ID]*validators.GetCurrentValidatorOutput, len(resp.Validators))
 	for _, validator := range resp.Validators {
```
yacovm, line 178: Just me being usually paranoid, but - validator is a pointer, can it be nil?

snow/validators/gvalidators/validator_state_test.go
```diff
@@ -236,3 +236,211 @@ func setupValidatorSet(b *testing.B, size int) map[ids.NodeID]*validators.GetVal
...
+			vdr0.NodeID: vdr0,
+			vdr1.NodeID: vdr1,
+			vdr2.NodeID: vdr2,
+		},
+		subnetID2: {
+			vdr2.NodeID: vdr2,
+			vdr3.NodeID: vdr3,
+		},
+	}
+	height := uint64(1337)
```
yacovm: leet?

snow/validators/state.go
```diff
@@ -114,3 +147,64 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+		cache: lru.NewCache[uint64, map[ids.ID]map[ids.NodeID]*GetValidatorOutput](validatorSetsCacheSize),
+	}
+}
+
+func (c *CachedState) GetAllValidatorSets(
+	ctx context.Context,
+	height uint64,
+) (map[ids.ID]map[ids.NodeID]*GetValidatorOutput, error) {
+	if validatorSet, ok := c.cache.Get(height); ok {
+		return maps.Clone(validatorSet), nil
```
yacovm: wondering why are we cloning this? The caller can anyway mutate the returned objects since the PK is a pointer. Are we concerned the validator set will be somehow mutated?

snow/validators/state.go
```diff
@@ -114,3 +147,64 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+	height uint64,
+	subnetID ids.ID,
+) (map[ids.NodeID]*GetValidatorOutput, error) {
+	if validatorSets, ok := c.cache.Get(height); ok {
+		validatorSet, exists := validatorSets[subnetID]
+		if !exists {
+			return nil, ErrValidatorSetForSubnetNotFound
+		}
+		return maps.Clone(validatorSet), nil
+	}
```
yacovm, line 241: so here we lookup but don't put in the cache on a miss. Can you explain the difference in behavior with the GetAllValidatorSets? 

Are we trying to prevent the all validator sets from being evicted from the cache?

snow/validators/state.go
```diff
@@ -114,3 +147,64 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
 }
+
+type CachedState struct {
+	State
+
+	// Caches validators for all subnets at various heights.
+	// Key: height
+	// Value: mapping height -> subnet ID -> validator set
+	cache cache.Cacher[uint64, map[ids.ID]map[ids.NodeID]*GetValidatorOutput]
+}
```
StephenButtolph: I'm not sure that this is the right place to perform the caching for warp verification. We will want to cache the canonical (sorted) validator set corresponding to the bitset expected for warp signatures... So we will need to implement a cache at a higher level than this implementation anyways.

With that in mind, do we want to add this caching at all?

yacovm: > We will want to cache the canonical (sorted) validator set corresponding to the bitset expected for warp signatures

Why? I thought our goal is to eliminate P-chain state lookups, not PK aggregation

geoff-vball: So, I think it might make sense to keep this layer (all validator sets from a given set may be used for warp verification for the duration of the epoch), and then the top caching layer of the canonical validator set could be kept on a per-subnet basis, since the sorting takes extra computation that we probably don't want to do for all subnets if not necessary?

snow/validators/state.go
```diff
@@ -165,3 +175,74 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+func NewCachedState(
+	state State,
+	graniteActivation time.Time,
+) State {
+	return &cachedState{
+		State:      state,
+		activation: graniteActivation,
+		cache:      lru.NewCache[uint64, map[ids.ID]WarpSet](validatorSetsCacheSize),
+	}
+}
```
StephenButtolph, line 209: Should add tests for this

chains/manager.go
```diff
@@ -1124,16 +1128,21 @@ func (m *manager) createSnowmanChain(
 		// P-chain.
 		ctx.ValidatorState = valState
 
-		// Initialize the validator state for future chains.
-		m.validatorState = validators.NewLockedState(&ctx.Lock, valState)
+		// Initialize validatorState for future chains.
+		valState = validators.NewLockedState(&ctx.Lock, valState)
 		if m.TracingEnabled {
-			m.validatorState = validators.Trace(m.validatorState, "lockedState", m.Tracer)
```
StephenButtolph, line 1130: While not explicitly needed in this PR, I felt like using `valState` until we assigned to `m.validatorState` at the end made a bit more sense.

snow/validators/state.go
```diff
@@ -165,3 +175,74 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+			return WarpSet{}, err
+		}
+		return FlattenValidatorSet(vdrSet)
+	}
+
+	s, err := c.GetWarpValidatorSets(ctx, height)
+	if err != nil {
+		return WarpSet{}, err
+	}
+	return s[subnetID], nil
```
StephenButtolph, line 247: Note: the change in the interface specification of `GetWarpValidatorSet` is needed here because this method can not distinguish between a subnet with no validators and a subnet with an invalid validator set.

It is critical that both of these cases will always mark the warp messages as invalid.

snow/validators/state.go
```diff
@@ -6,11 +6,19 @@ package validators
...
 	"sync"
+	"time"
 
+	"github.com/ava-labs/avalanchego/cache"
+	"github.com/ava-labs/avalanchego/cache/lru"
 	"github.com/ava-labs/avalanchego/ids"
 )
 
-var _ State = (*lockedState)(nil)
+const validatorSetsCacheSize = 8
```
yacovm, line 16: how was this `8` chosen?

vms/platformvm/warp/validator.go
```diff
@@ -39,6 +39,8 @@ var FlattenValidatorSet = validators.FlattenValidatorSet
 // GetCanonicalValidatorSetFromSubnetID returns the CanonicalValidatorSet of [subnetID] at
 // [pChainHeight]. The returned CanonicalValidatorSet includes the validator set in a canonical ordering
 // and the total weight.
+//
+// Deprecated: Use [validators.State.GetWarpValidatorSet] instead.
```
yacovm, line 43: But this is still what's used by the warp verifier. Are you planning to change the warp verifier too later in a separate PR?

snow/validators/state.go
```diff
@@ -165,3 +175,74 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+		if err != nil {
+			return WarpSet{}, err
+		}
+		return FlattenValidatorSet(vdrSet)
+	}
+
+	s, err := c.GetWarpValidatorSets(ctx, height)
+	if err != nil {
+		return WarpSet{}, err
+	}
```
yacovm, line 246: Is the intent to have this method like this long term - fetch the (ideally) cached version of all subnets and then just return the queried subnet?

StephenButtolph: After granite, this method will be removed entirely.

> // TODO: After Granite, this method should be removed and users should
	// directly call GetWarpValidatorSets.

vms/platformvm/warp/validator.go
StephenButtolph: yes

> This PR deprecated warp.GetCanonicalValidatorSetFromSubnetID. There should be a followup PR to remove it entirely (prior to Granite being activated).

snow/validators/state_test.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		got, err := cached.GetWarpValidatorSet(context.Background(), test.height, test.subnetID)
+		require.ErrorIs(err, test.wantErr)
+		require.Equal(test.want, got)
+		require.Equal(expectCached, !cacheMiss)
+	}
+}
+
+func BenchmarkCachedState_GetWarpValidatorSet_Active(b *testing.B) {
```
StephenButtolph, line 249: As expected, cached values are returned in constant time (~270 ms)
```
BenchmarkCachedState_GetWarpValidatorSet_Active/1-12  	 4232695	       273.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/2-12  	 3868628	       265.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/4-12  	 4001122	       262.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/8-12  	 3786343	       270.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/16-12 	 3903648	       267.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/32-12 	 3749355	       275.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/64-12 	 3863325	       268.3 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/128-12   3791690	       270.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/256-12   3941542	       263.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/512-12   3781414	       269.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkCachedState_GetWarpValidatorSet_Active/1024-12  3869460	       268.3 ns/op	       0 B/op	       0 allocs/op
```

snow/validators/state.go
```diff
@@ -165,3 +175,74 @@ func (n *noValidators) GetCurrentValidatorSet(ctx context.Context, _ ids.ID) (ma
...
+	// generating the full validator set for each P-chain height.
+	//
+	// After Granite, it is expected that the P-chain height will be frozen for
+	// the duration of an epoch, so generating the full set is not expensive as
+	// long as it is cached.
+	activation time.Time
+
+	// Caches validators for all subnets at various heights.
+	// Key: height
+	// Value: mapping height -> subnet ID -> validator set
```
michaelkaplan13:
```suggestion
	// Value: mapping subnet ID -> validator set
```

...or am I misunderstanding something?

chains/manager.go
```diff
@@ -1115,6 +1115,10 @@ func (m *manager) createSnowmanChain(
 			return nil, fmt.Errorf("expected validators.State but got %T", vm)
 		}
 
+		// Wrap the validator state with a cached state so that P-chain lookups
```
michaelkaplan13, line 1118: Is there an extra wrapping occurring here for future (non P-chain) chains?

AFAICT, it will be `cached(locked(cached(valState)))`, making this inner-caching unnecessary.

I _think_ we'd just want to wrap it in the assignment to `ctx.ValidatorState` below, so that the P-Chain version is properly just `cached(valState)` and non P-chain versions are `cached(locked(valState))`

StephenButtolph: I had actually originally done it that way. I decided that this way is likely better.

Specifically, If the P-chain as already requested the validator sets at height `x`, and then later another chain requests the validator sets at height `x`, we probably shouldn't need to recalculate the full validator set for `x`.

(Same logic if another chain had already requested `x` and then the P-chain needed height `x`.)

So essentially the dependency graph is probably better as:
```mermaid
graph TD
    A[X-Chain] --> D[Lock Free Cache]
    B[Subnet] --> H[GRPC Cache]
    H --> I[GRPC]
    I --> D
    C[C-Chain] --> D
    D --> E[Lock]
    E --> F[Global Cache]
    F --> G[P-Chain]
    G --> F
```

Rather than:
```mermaid
graph TD
    A[X-Chain] --> D[Lock Free Cache]
    B[Subnet] --> H[GRPC Cache]
    H --> I[GRPC]
    I --> D
    C[C-Chain] --> D
    D --> E[Lock]
    E --> G[P-Chain]
    G --> F[P-Chain Cache]
    F --> G
```

snow/validators/state.go
StephenButtolph: This number was not selected intelligently

chains/manager.go
michaelkaplan13: I don't think I would have followed without the diagrams, but yeah this makes sense. Thanks for explaining.

snow/validators/state.go
michaelkaplan13: Trying to think if there's any semi-reasonable way we can go about determining a reasonable cache size other than a gut feeling (though it will probably be a gut feeling to a certain extent). Assuming all chains are using epoching, it should be based on the number of chains we expect a single node to need to support, right?

michaelkaplan13 (approved):
LGTM, only open question is if we want to include more of an explanation regarding the validators sets cache size. 🤷

snow/validators/state.go
StephenButtolph: Kind of... Because we have multiple of these caches, even if we added more chains, I don't think we would necessarily cause a bunch of thrashing... Each (subnet) chain would have its own grpc cache. Then we would have X-chain and C-chain hitting the lock free cache (assuming grpc cache is getting hit). With the P-chain hitting the locked cache.

The only time this doesn't hold is during epoch changes (a miss in a higher cache can cause a miss in a lower cache as well). But assuming our epochs are long enough... I think we should be fine as long as the X-chain and C-chain caches aren't stepping on each-others toes... But the X-chain never even uses this (no warp usage there)...

I think the only way to actually understand this is by doing some e2e perf testing here...

Should I add a TODO for us to do during the testing next week?

StephenButtolph (approved):
Approving as I'm a code owner, but someone else should review this (I wrote most of this code)

snow/validators/state.go
```diff
@@ -34,7 +42,9 @@ type State interface {
 	GetWarpValidatorSets(ctx context.Context, height uint64) (map[ids.ID]WarpSet, error)
 
 	// GetWarpValidatorSet returns the canonical warp validator set for the
-	// requested subnet at the requested P-chain height.
+	// requested subnet at the requested P-chain height. If the subnet doesn't
+	// have a canonical warp validator set, either the returned set will be
+	// empty or an error will be returned.
```
geoff-vball, line 47: Can we clarify the behaviour here a bit? Do we know in which cases we return empty and which cases we return an error? 

Or maybe I'm misunderstanding "doesn't have a canonical warp validator set". Is that just for unconverted subnets, or does it also include L1s with now empty validator sets.

geoff-vball (commented):
One question on documentation, but otherwise LGTM 👍

snow/validators/state.go
StephenButtolph: So, the behavior here is:

If caching has been enabled, subnets without a canonical warp validator set (which can only happen if the total stake weight of the validator set overflows uint64) will be returned as if they have no validators (empty).

If caching hasn't been enabled, `FlattenValidatorSet` will return an error (and therefore this function will return an error).

During warp verification, an empty validator set is guaranteed to report an error (because warp messages must be signed by at least one validator). And previously generating the warp validator set would have errored. So this shouldn't change the validity of warp messages.

It is pretty subtle though.

StephenButtolph: I added a test to coreth here: https://github.com/ava-labs/coreth/pull/1311

PR #4172, opened 2025-08-13
Add optional step to archive post-reexecution state to S3

aaronbuchwald:
This PR adds an optional step to the re-execution custom action to archive the S3 post execution state.

This also updates the ARC action to provide an added optional input to the manual workflow trigger that will use it.

TODO: add a R/W IAM role to AWS and GitHub secrets to perform the archive step (only IAM role in secrets currently only has read access).

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -43,6 +43,9 @@ inputs:
     description: 'Whether to push the benchmark result to GitHub.'
     required: true
     default: false
+  push-post-state:
+    description: 'S3 destination to copy the current-state directory after completing re-execution. If empty, this will be skipped.'
+    default: ''
 
 runs:
   using: composite
@@ -100,3 +103,10 @@ runs:
         output-file-path: ${{ env.BENCHMARK_OUTPUT_FILE }}
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
+
+    - uses: ./.github/actions/install-nix
+      if: ${{ inputs.push-post-state != '' }}
+    - name: Copy Post-State to S3
+      if: ${{ inputs.push-post-state != '' }}
+      shell: nix develop --command bash -x {0}
+      run: ./scripts/run_task.sh export-dir-to-s3 LOCAL_SRC=$CURRENT_STATE_DIR S3_DEST=${{ inputs.push-post-state }}
--- a/.github/workflows/c-chain-reexecution-benchmark-arc.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-arc.yml
@@ -24,6 +24,9 @@ on:
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
         required: false
         default: ubuntu-latest
+      push-post-state-dir:
+        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
+        default: ''
 
   schedule:
     - cron: '0 9 * * *' # Runs every day at 09:00 UTC (04:00 EST)
@@ -85,3 +88,4 @@ jobs:
           aws-role: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          push-post-state-dir: ${{ github.event.inputs.push-post-state-dir }}


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -67,6 +70,8 @@ runs:
     - name: Run C-Chain Re-Execution
       uses: ./.github/actions/run-monitored-tmpnet-cmd
       with:
+        # The environment varaibles set in $GITHUB_ENV should be accessible when the task is run, but do not seem to be when executed through
```
aaronbuchwald: Remove

.github/workflows/c-chain-reexecution-benchmark-arc.yml
```diff
@@ -52,7 +56,7 @@ jobs:
           elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
             {
               echo "start-block=101"
-              echo "end-block=250000"
+              echo "end-block=1000"
```
aaronbuchwald: Revert

.github/workflows/c-chain-reexecution-benchmark-arc.yml
```diff
@@ -82,6 +86,7 @@ jobs:
           prometheus-username: ${{ secrets.PROMETHEUS_ID || '' }}
           prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
           push-github-action-benchmark: ${{ github.event_name == 'schedule' }}
-          aws-role: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-role: ${{ secrets.AWS_S3_RW_ROLE }}
```
aaronbuchwald: Give RW only if needed, otherwise provide read only credential.

.github/workflows/c-chain-reexecution-benchmark-arc.yml
```diff
@@ -41,6 +44,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set task parameters
         id: set-params
+        shell: bash
```
aaronbuchwald, line 47: This fixes params not being passed correctly.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -100,3 +105,26 @@ runs:
...
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
+
+    - uses: ./.github/actions/install-nix
+      if: ${{ inputs.push-post-state != '' }}
+    - name: Push Post-State to S3 (if not exists)
+      if: ${{ inputs.push-post-state != '' }}
+      shell: nix develop --command bash -x {0}
+      run: |
+        set -euo pipefail
```
aaronbuchwald: Prefer script over longer scripting here

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -67,7 +70,7 @@ runs:
     - name: Run C-Chain Re-Execution
       uses: ./.github/actions/run-monitored-tmpnet-cmd
       with:
-        run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data
+        run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=${{ env.EXECUTION_DATA_DIR }} SOURCE_BLOCK_DIR=${{ env.SOURCE_BLOCK_DIR }} CURRENT_STATE_DIR=${{ env.CURRENT_STATE_DIR }} START_BLOCK=${{ env.START_BLOCK }} END_BLOCK=${{ env.END_BLOCK }} LABELS=${{ env.LABELS }} BENCHMARK_OUTPUT_FILE=${{ env.BENCHMARK_OUTPUT_FILE }}
```
aaronbuchwald: Seeing behavior that I do not fully understand when passing these via environment variables, so switching to pass them explicitly.

scripts/copy_dir.sh
```diff
@@ -51,4 +51,34 @@ copy_source() {
...
 
+# Function to check the destination directory does not exist to avoid
+# overwrites
+function check_dst_not_exists() {
+    local dst="$1"
+
+    if [[ "$dst" == s3://* ]]; then
+        # Validate the S3 path format using a single regular expression.
+        echo "Checking S3 path format: $dst"
+        if ! [[ "$dst" =~ ^s3://[^/]+/([^/]+/)$ ]]; then
```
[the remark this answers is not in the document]
aaronbuchwald: The error message specifies the expected format, which uses one level of nesting. I think this is pretty clear, but am open to change it if others also find this unclear:

```
echo "Expected format: s3://<bucket-name>/<directory-name>/"
```

maru-ava, line 56: Does an invalid path format not result in s5cmd reporting failure?

scripts/copy_dir.sh
```diff
@@ -51,4 +51,34 @@ copy_source() {
...
+          echo "Error: Invalid S3 path format."
+          echo "Expected format: s3://<bucket-name>/<directory-name>/"
+          exit 1
+        fi
+
+        echo "Checking if S3 path exists: $dst"
+        if s5cmd ls "$dst" | grep -q .; then
```
maru-ava: What happens if s5cmd returns an error for a reason other than the path format (e.g. network connectivity, auth failure)?

scripts/copy_dir.sh
```diff
@@ -51,4 +51,34 @@ copy_source() {
     fi
 }
 
+# Function to check the destination directory does not exist to avoid
+# overwrites
+function check_dst_not_exists() {
+    local dst="$1"
+
+    if [[ "$dst" == s3://* ]]; then
+        # Validate the S3 path format using a single regular expression.
```
aaronbuchwald:
```suggestion
        # Validate the S3 path format as s3://<bucket-name>/<directory-name>/
```

scripts/copy_dir.sh
aaronbuchwald: It's intended to enforce the intended directory structure in the comment below. Updating the comment as well: https://github.com/ava-labs/avalanchego/pull/4172/files#r2279010366

scripts/copy_dir.sh
aaronbuchwald: Good callout. Updated the script to better handle errors. Unfortunately, there doesn't seem to be an easy way to check for an empty s3 directory as I'd like, so I have to check an expected error message. On the bright side, this favors failing to copy as opposed to allowing an accidental overwrite if the behavior changes, which is preferable imo.

Added a comment explaining the behavior.

aaronbuchwald: Here's the results covering each of the error cases:

```bash
# Invalid S3 Path format
avalanchego % ./scripts/copy_dir.sh tdir/ s3://avalanchego-bootstrap-testing/tdir 
Checking S3 path format: s3://avalanchego-bootstrap-testing/tdir
Error: Invalid S3 path format.
Expected format: s3://<bucket-name>/<directory-name>/
# Invalid S3 Path format due to unexpected level of nesting
avalanchego % ./scripts/copy_dir.sh tdir/ s3://avalanchego-bootstrap-testing/tdir/ete/
Checking S3 path format: s3://avalanchego-bootstrap-testing/tdir/ete/
Error: Invalid S3 path format.
Expected format: s3://<bucket-name>/<directory-name>/
# Unexpected error ie. wrong AWS region
avalanchego % ./scripts/copy_dir.sh tdir/ s3://avalanchego-bootstrap-testing/tdir/    
Checking S3 path format: s3://avalanchego-bootstrap-testing/tdir/
Checking if S3 path exists: s3://avalanchego-bootstrap-testing/tdir/
Error: failed to check for contents of s3://avalanchego-bootstrap-testing/tdir/
ERROR "ls s3://avalanchego-bootstrap-testing/tdir/": AuthorizationHeaderMalformed: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-east-2' status code: 400, request id: 357CAEQY6KWJG0N3, host id: Rlnh6x8gQts2WYc7yKl4cbUnvSg6Co9vFn6GoIHDNA06m21zgK3zMV9Cf+NEdrmPH7k068SNX9g=
avalanchego % export AWS_REGION=us-east-2
# Success case
avalanchego % ./scripts/copy_dir.sh tdir/ s3://avalanchego-bootstrap-testing/tdir/
Checking S3 path format: s3://avalanchego-bootstrap-testing/tdir/
Checking if S3 path exists: s3://avalanchego-bootstrap-testing/tdir/
Verified S3 destination: 's3://avalanchego-bootstrap-testing/tdir/' is empty
cp tdir/test s3://avalanchego-bootstrap-testing/tdir/test

real    0m0.245s
user    0m0.019s
sys     0m0.015s
# Refuse accidental overwrite due to non-empty directory
avalanchego % ./scripts/copy_dir.sh tdir/ s3://avalanchego-bootstrap-testing/tdir/
Checking S3 path format: s3://avalanchego-bootstrap-testing/tdir/
Checking if S3 path exists: s3://avalanchego-bootstrap-testing/tdir/
Cannot copy to non-empty destination: 's3://avalanchego-bootstrap-testing/tdir/':
2025/08/18 15:19:51                 6  test
```

Taskfile.yml
```diff
@@ -101,7 +101,7 @@ tasks:
       LOCAL_SRC: '{{.LOCAL_SRC}}'
       S3_DST: '{{.S3_DST}}'
     cmds:
-      - cmd: s5cmd cp {{.LOCAL_SRC}} {{.S3_DST}}
+      - cmd: bash -x ./scripts/copy_dir.sh {{.LOCAL_SRC}} {{.S3_DST}}
```
aaronbuchwald, line 104: Switch to use copy dir script to benefit from check for non-empty s3 directory and avoid accidental overwrites

scripts/copy_dir.sh
```diff
@@ -9,36 +9,36 @@ set -euo pipefail
...
-mkdir -p "$DST"
-
 # Function to copy from a single source to destination
-copy_source() {
+function copy_source() {
     local source="$1"
     local dest="$2"
     
+    # Ensure destination directory exists (after validation)
+    mkdir -p "$dest"
```
joshua-kim: I think the existing code already does this... but won't this create dirs on our local fs with the path `s3://...`?

scripts/copy_dir.sh
```diff
@@ -9,36 +9,36 @@ set -euo pipefail
...
+    
     # Check if source starts with s3://
-    if [[ "$source" == s3://* ]]; then
-        echo "Copying from S3: $source -> $dest"
```
joshua-kim, line 30: Unrelated log removal?

scripts/copy_dir.sh
```diff
@@ -51,4 +51,53 @@ copy_source() {
...
+        # by misreporting non-existent directories as existing, which means
+        # a change in beahvior would cause the script to fail to copy rather
+        # than allow accidental overwrites.
+        echo "Checking if S3 path exists: $dst"
+        if ! OUTPUT=$(s5cmd ls "$dst" 2>&1); then
+            # If the command fails, check for the expected error message.
+            if [[ "$OUTPUT" == *"no object found"* ]]; then
+                echo "Verified S3 destination: '$dst' is empty"
+            else
+                echo "Error: failed to check for contents of $dst"
```
joshua-kim, line 76: If the bucket does not exist - should we be creating it and then copying? Currently I think that if we're trying to copy from a source and the destination bucket doesn't exist, we'll have to create it manually - which is inconsistent w/ the behavior of this script when copying between local dirs where we'll create the directory before proceeding.

.github/workflows/c-chain-reexecution-benchmark-arc.yml
```diff
@@ -24,6 +24,9 @@ on:
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
         required: false
         default: ubuntu-latest
+      push-post-state:
+        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
```
joshua-kim, line 28: We're a bit inconsistent w/ the naming of this (we're calling it post state/post execution state/re-executed in a few places). Since we're already referring to this as re-execute in the current code, could we use that naming in lieu of the new naming?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -100,3 +111,10 @@ runs:
         output-file-path: ${{ env.BENCHMARK_OUTPUT_FILE }}
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
+
+    - uses: ./.github/actions/install-nix
+      if: ${{ inputs.push-post-state != '' }}
+    - name: Push Post-State to S3 (if not exists)
+      if: ${{ inputs.push-post-state != '' }}
+      shell: nix develop --command bash -x {0}
+      run: ./scripts/run_task.sh export-dir-to-s3 LOCAL_SRC=${{ env.EXECUTION_DATA_DIR }}/current-state/ S3_DST=${{ inputs.push-post-state }}
```
joshua-kim, line 120: I found this a bit confusing/I'm not sure what was intended here so just dumping my thoughts:

* It looks like we log `EXECUTION_DATA_DIR` as `"EXECUTION_DATA_DIR=${{ inputs.workspace }}/reexecution-data"` earlier (appended a dir for a log)
* We pass in the env var as `EXECUTION_DATA_DIR=${{ env.EXECUTION_DATA_DIR }} \` (unmodified)
* But the actual directory we're copying over is `LOCAL_SRC=${{ env.EXECUTION_DATA_DIR }}/current-state/` (different appended path, this is also a bit confusing w/ naming since we have an env var for `CURRENT_STATE_DIR`)

Would it make sense for us to pass in `env.EXECUTION_DATA_DIR }}/current-state/` when calling `run_task` and update the log? Based on the README it also looks like it might be possible that we meant to use `CURRENT_STATE_DIR`, since that's documented as having the `current-state` dir format.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -53,7 +56,7 @@ runs:
       run: |
         {
           echo "EXECUTION_DATA_DIR=${{ inputs.workspace }}/reexecution-data"
-          echo "BENCHMARK_OUTPUT_FILE=${{ inputs.workspace }}/reexecute-cchain-range-benchmark-res.txt"
+          echo "BENCHMARK_OUTPUT_FILE=output.txt"
```
joshua-kim, line 59: Why is this re-named? Is this a related change?

.github/workflows/c-chain-reexecution-benchmark-arc.yml
```diff
@@ -82,6 +86,7 @@ jobs:
           prometheus-username: ${{ secrets.PROMETHEUS_ID || '' }}
           prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
           push-github-action-benchmark: ${{ github.event_name == 'schedule' }}
-          aws-role: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
```
joshua-kim, line 89: nit: Is this over-engineering? Feels like we could just use r/w perms since this job now is able to write to s3 when configured to do so

aaronbuchwald: Ehh there's definitely an argument there. I thought this was a reasonable extra protection / defense in depth against accidentally corrupting it.

Writes would only occur via manual trigger, so are run infrequently and if something breaks that would overwrite or push incorrect data by accident, this minimizes the access level of > 90% of runs.

It's also pretty straight forward here imo, so the real cost imo is maintaining two roles instead of one, but that is already set up and they are named explicitly.

scripts/copy_dir.sh
aaronbuchwald: Good catch, moving into the conditional block below where we are using `cp`

scripts/copy_dir.sh
aaronbuchwald: Will add back

scripts/copy_dir.sh
aaronbuchwald: This is true if the bucket does not exist, but it will create the directory for you. This seems like the right default to me as I would not intend to copy directories to a brand new S3 bucket, just copy a directory into an existing bucket.

If you feel strongly that the behavior should match the local file system, I'd prefer to remove `mkdir -p <dst>` from the local file system handling than to create the bucket if it doesn't exist yet here.

.github/workflows/c-chain-reexecution-benchmark-arc.yml
aaronbuchwald: I think this is fine. re-execution refers to the fact we are re-executing blocks from an actual network. Post execution refers to the fact that we are pushing the result after completing execution. post-execution is an adjective here, not a name and post-reexecution doesn't sound right to me.

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: There's some weirdness in how GitHub Action Benchmark handles paths, where it was not handling absolute paths correctly. I found while working on this PR that env vars from $GITHUB_ENV are not picked up correctly by the task run from within `run-monitored-tmpnet-cmd`, so I switched to passing each var in explicitly and that triggered the mishandling of absolute paths.

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: `reexecute-cchain-range` - takes in `CURRENT_STATE_DIR` and `SOURCE_BLOCK_DIR` as separate params and reads from the block dir and executes on and overwrites `CURRENT_STATE_DIR`.

`reexecute-cchain-range-with-copied-data` - this copies data in from S3 buckets or a local directory (supports local directory for testing on snoopy/linus instances with local copy of each)

`EXECUTION_DATA_DIR` is where both of these will be copied into and that's why there's a suffix added to both for the current state and blocks.

That can definitely be confusing. I'd propose using different var names for these two tasks, so that the copy job is clear that `CURRENT_STATE_DIR` and `SOURCE_BLOCK_DIR` are the source to copy from as opposed to what's actually used. I think the root of the confusion is re-using the same var names across both jobs.

If we used `CURRENT_STATE_DIR` here, that would be incorrect because that's the original S3 bucket where we copied the data from.

Since this is using the existing naming, I'd prefer to make the rename a separate PR as a follow up to this. Will make one on top of this PR for now and if you think it should be part of this PR, we can block on it to merge this in.

aaronbuchwald:
Addressed all comments except https://github.com/ava-labs/avalanchego/pull/4172#discussion_r2283250356. As mentioned, will make a separate PR on top of this to improve the names, so that they are not shared between the tasks `reexecute-cchain-range-with-copied-data` and `reexecute-cchain-range`.

aaronbuchwald:
Very strange: GitHub Action Benchmark is failing to checkout the `gh-pages` branch because two very specific files were changed somehow during the workflow: https://github.com/ava-labs/avalanchego/actions/runs/17052404447/job/48342978430?pr=4172#step:6:1578
Screenshot 2025-08-18 at 17 18 03

RodrigoVillar:
> Very strange: GitHub Action Benchmark is failing to checkout the `gh-pages` branch because two very specific files were changed somehow during the workflow: https://github.com/ava-labs/avalanchego/actions/runs/17052404447/job/48342978430?pr=4172#step:6:1578 Screenshot 2025-08-18 at 17 18 03

@aaronbuchwald I'm also facing the same issue on a separate PR: https://github.com/ava-labs/avalanchego/actions/runs/17069126455/job/48393320563?pr=4181

aaronbuchwald:
> > Very strange: GitHub Action Benchmark is failing to checkout the `gh-pages` branch because two very specific files were changed somehow during the workflow: https://github.com/ava-labs/avalanchego/actions/runs/17052404447/job/48342978430?pr=4172#step:6:1578 Screenshot 2025-08-18 at 17 18 03
> 
> @aaronbuchwald I'm also facing the same issue on a separate PR: https://github.com/ava-labs/avalanchego/actions/runs/17069126455/job/48393320563?pr=4181

Looks like this is caused by the `actions/cache` action using the relative path of `./cache`. When https://github.com/ava-labs/avalanchego/pull/4088/files was merged with two changes to files in that path, any future PR that pulls the old cache entry in (it's still getting pulled in because jobs after that PR was merged are hitting the same failure case and therefore not updating the cache) will now see a diff between the up to date source code and the outdated source code that is pulled in and overwriting the cache directory with the code from prior to 4088.

The documentation for Caching in GitHub Actions explicitly uses `./tmp` instead of `./cache` as used in GitHub Actions and I assume this assumes that `./tmp` is included in `.gitignore` so that it doesn't run into issues like this.

It's possible that using a path included in `.gitignore` will interfere with pushing to `gh-pages` though. If that's the case, then it would imply we need to use a path that never interferes with the source code and is not ignored by `.gitignore`.

Switching to use `./tmp` seems to fix the issue: https://github.com/ava-labs/avalanchego/actions/runs/17074658133/job/48412681919?pr=4183

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -10,10 +10,10 @@ inputs:
     default: '250000'
   source-block-dir:
     description: 'The source block directory. Supports S3 directory/zip and local directories.'
-    default: 's3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb.zip'
+    default: 's3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**'
```
yacovm, line 13: can you explain what this ** means?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -67,7 +70,15 @@ runs:
     - name: Run C-Chain Re-Execution
       uses: ./.github/actions/run-monitored-tmpnet-cmd
       with:
-        run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data
+        run: |
+          ./scripts/run_task.sh reexecute-cchain-range-with-copied-data \
```
yacovm, line 74: these capital cased variables are only defined in this file, so what is consuming them via run_task?

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: `*` matches any key at a given level of an S3 bucket and `**` matches any key at any level of recursive nesting within the bucket.

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: Run task invokes the Taskfile (ref: https://taskfile.dev/), which uses this syntax.

These variables are defined here: https://github.com/ava-labs/avalanchego/blob/e5593be724b114c420f27d0888ea485d4769a9ea/Taskfile.yml#L184

yacovm: Yeah I know what that is, it's just that my IDE searched for the strings and couldn't find them. Now it does, odd... must have been still indexing

scripts/copy_dir.sh
```diff
@@ -51,4 +45,53 @@ copy_source() {
...
+        # Note: S3 tooling does not provide a native way to check for an empty
+        # directory. To avoid accidental overwrites, we use a best-effort, brittle
+        # workaround that relies on the expected status code and error message
+        # from s5cmd ls.
+        # If the error message changes, this script would be expected to fail
+        # by misreporting non-existent directories as existing, which means
+        # a change in beahvior would cause the script to fail to copy rather
+        # than allow accidental overwrites.
+        echo "Checking if S3 path exists: $dst"
+        if ! OUTPUT=$(s5cmd ls "$dst" 2>&1); then
```
yacovm, line 71: might be a stupid question but - how do we have the credentials to invoke this command? Do we have them set somewhere as env vars?

scripts/copy_dir.sh
```diff
@@ -51,4 +45,53 @@ copy_source() {
...
 
+# Function to check the destination directory does not exist to avoid
+# overwrites
+function check_dst_not_exists() {
+    local dst="$1"
+
+    if [[ "$dst" == s3://* ]]; then
+        # Validate the S3 path format as s3://<bucket-name>/<directory-name>/
+        echo "Checking S3 path format: $dst"
+        if ! [[ "$dst" =~ ^s3://[^/]+/([^/]+/)$ ]]; then
```
yacovm, line 56: seems like a rather permissive regexp `[^/]+/([^/]+/)`. It basically lets everything that is separated by a / and ends with a /. 

Is there really a concern here that we'll get an incorrect input?

scripts/copy_dir.sh
aaronbuchwald: Yup we assume AWS credentials are set. In CI, we use aws-actions/configure-aws-credentials@v4: https://github.com/ava-labs/avalanchego/blob/7aa6a17b8499513033a483f0840c88d08876aab0/.github/actions/c-chain-reexecution-benchmark/action.yml#L66

scripts/copy_dir.sh
aaronbuchwald: Yes, this PR provides an optional feature to archive the state to S3 and takes in a destination path from the user (this is gated to developers on our team). This enforces that we don't nest two different directories together like putting a current state directory nested inside of another one from a different chain height. It is not perfect, but seemed a good safeguard to add.

Open to better ideas to reduce the likelihood that we corrupt the S3 bucket we're using here.

PR #4173, opened 2025-08-13
use pr to check archive state of state at 250k

aaronbuchwald:
Test PR to check that archiving the state at the end of execution on ARC works as expected.


aaronbuchwald:
Shell param was missing causing environment not to get passed correctly and making me think that Taskfile was handling env variables in an unexpected way: https://github.com/ava-labs/avalanchego/pull/4173/commits/8f306f4e076e9929cb91b0eae6a588dc3718a452.

It seems that GitHub Action Benchmark is not handling the output file path as I'd expect (https://github.com/ava-labs/avalanchego/actions/runs/16969752285/job/48103220490?pr=4173#step:6:1033):

Screenshot 2025-08-14 at 11 42 35

The output file path input is clearly different from the filepath that it is complaining does not exist.

aaronbuchwald:
Not very detailed, but this issue with relative paths may be related: https://github.com/benchmark-action/github-action-benchmark/issues/307

PR #4174, opened 2025-08-14
Use EmptyVoteMetadata in Simplex Proto Messages

samliok:
## Why this should be merged

Metadata for `EmptyVotes` and `EmptyNotarizations` should only include `Round` and `Epoch` values, but where previously the same as typical `Blocks` and `Notarizations`. 

This was updated on the simplex side, but not AvalancheGo
https://github.com/ava-labs/Simplex/pull/215

no


yacovm (approved):

proto/p2p/p2p.proto
```diff
@@ -459,6 +458,13 @@ message ProtocolMetadata {
   bytes prev = 5;
 }
 
+message EmptyVoteMetadata {
```
StephenButtolph, line 461: `ProtocolMetadata` includes a `version` field. It seems to me like we should either remove the `version` field from `ProtocolMetadata` or we should include it here.

fwiw I've always felt like the `version` field was weird...

yacovm: I tend to disagree. It makes sense to have a version in the protocol metadata since we might want to extend the protocol metadata later on. However, for empty votes - we should not have a version as the empty notarization isn't there to have information piggybacked on top of it.

StephenButtolph: > since we might want to extend the protocol metadata later on

I don't disagree that we may want to change the protocol metadata later on, but I think that any changes to the metadata should probably be based on the conventional ways to update proto messages. Rather than with the explicit version field.

That being said, (since in my preferred world the protocol metadata wouldn't have the version) I don't have a problem merging the empty vote metadata as is (without the version).

StephenButtolph (approved):

PR #4175, opened 2025-08-15
Simplify predicate handling

StephenButtolph:
## Why this should be merged

This PR slims down the public API exposed by the `predicate` package for handling predicates. Now only `[]common.Hash` and `[]byte` are exposed publicly.

## How this works

Mainly just deleted code that wasn't needed and simplified a couple implementations.

This PR doesn't touch any result handling logic.

## How this was tested

Unit tests.


vms/evm/predicate/predicatetest/predicatetest.go
```diff
@@ -11,29 +11,30 @@ import (
...
-	packed := pack(predicateBytes)
-	numHashes := len(packed) / common.HashLength
-	storageKeys := make([]common.Hash, numHashes)
-	for i := range storageKeys {
-		start := i * common.HashLength
-		copy(storageKeys[i][:], packed[start:])
-	}
-
+// given precompile address.
+func NewAccessList(address common.Address, b []byte) types.AccessList {
```
StephenButtolph: tbh, this function is so minimal that we might want to just get rid of it entirely.

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
-	copy(bytes, b)
-	bytes[len(b)] = delimiter
-
-	return common.RightPadBytes(bytes, RoundUpTo32(len(bytes)))
-}
-
-// Unpack unpacks a predicate by stripping right padded zeroes, checking for the delimiter,
-// ensuring there is not excess padding, and returning the original message.
-// Returns an error if it finds an incorrect encoding.
-func Unpack(p []byte) ([]byte, error) {
```
StephenButtolph, line 70: We don't want to expose the padded byte slice in the public API.

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
+const Delimiter = 0xff
 
 var (
-	errEmptyPredicate    = errors.New("predicate specified empty predicate")
-	errAllZeroBytes      = errors.New("predicate specified all zero bytes")
-	errExcessPadding     = errors.New("predicate specified excess padding")
-	errWrongEndDelimiter = errors.New("wrong end delimiter")
+	ErrMissingDelimiter  = errors.New("no delimiter found")
+	ErrExcessPadding     = errors.New("predicate included excess padding")
+	ErrWrongEndDelimiter = errors.New("wrong delimiter")
```
StephenButtolph: These are exported because `predicate_test` is compiled as a separate package (and needs them to be exported to test the cases). Technically we could separate `predicate_test` into multiple files to avoid this... But I don't really mind exporting the errors (in case we ever want to write an integration test that hits these errors)

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
 
+// Predicate is a message padded with the delimiter and zeros and chunked into
+// 32-byte chunks.
+type Predicate []common.Hash
+
+// Bytes converts the chunked predicate into the original message.
+//
+// Returns an error if it finds an incorrect encoding.
+func (p Predicate) Bytes() ([]byte, error) {
+	padded := make([]byte, common.HashLength*len(p))
```
StephenButtolph, line 56: Technically this could be made to allocate less bytes, but I don't think we need to be stingy here on the allocation (and it makes the code more clear imo)

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
-		return nil, fmt.Errorf("%w: length (%d)", errAllZeroBytes, len(p))
-	}
-
-	if paddedLength := RoundUpTo32(len(trim)); paddedLength != len(p) {
-		return nil, fmt.Errorf("%w: got length (%d), expected length (%d)", errExcessPadding, len(p), paddedLength)
-	}
-
-	if trim[len(trim)-1] != delimiter {
-		return nil, errWrongEndDelimiter
+		predicates[el.Address] = append(predicates[el.Address], el.StorageKeys)
```
StephenButtolph, line 92: I removed the copy here... If we want to keep the copy to avoid mutating the predicates if `StorageKeys` is later modified, then we can do that... But I didn't see a need.

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
-		// Append a copy of the storage keys to avoid aliasing
-		keys := make([]common.Hash, len(el.StorageKeys))
-		copy(keys, el.StorageKeys)
-		predicateStorageSlots[el.Address] = append(predicateStorageSlots[el.Address], keys)
-	}
-
-	return predicateStorageSlots
-}
-
-func RoundUpTo32(x int) int {
```
StephenButtolph, line 52: We probably don't want to expose this in the public API either.

vms/evm/predicate/predicate.go
```diff
@@ -11,78 +11,68 @@ import (
...
-const delimiter = 0xff
+// Delimiter separates the actual predicate bytes from the padded zero bytes.
+//
+// Predicates are encoded in the Access List of transactions by using in the
+// access tuples. This means that the length must be a multiple of
+// [common.HashLength].
+//
+// Even if the original predicate bytes is a multiple of [common.HashLength],
+// the delimiter must be appended to support decoding.
+const Delimiter = 0xff
```
StephenButtolph: Exported so that the test package could use it rather than copying `0xff`

vms/evm/predicate/predicatetest/predicatetest.go
```diff
@@ -11,29 +11,30 @@ import (
...
-func pack(b []byte) []byte {
-	bytes := make([]byte, len(b)+1)
-	copy(bytes, b)
-	bytes[len(b)] = 0xff
-	return common.RightPadBytes(bytes, predicate.RoundUpTo32(len(bytes)))
+// New constructs a predicate from raw predicate bytes.
+//
+// It chunks the predicate by appending [predicate.Delimiter] and zero-padding
+// to a multiple of 32 bytes.
+func New(b []byte) predicate.Predicate {
```
joshua-kim: I agree we should remove `NewAccessList`... but should we also move this to `predicate` as well? It seems weird to me that we have a way to get the original message out of a predicate, but no way to actually create a predicate without the usage of a test helper.

vms/evm/predicate/predicate_test.go
```diff
@@ -83,213 +87,111 @@ func TestFromAccessListVectors(t *testing.T) {
...
+			name:    "wrong_delimiter",
+			input:   Predicate{{0x42}},
 			want:    nil,
-			wantErr: errExcessPadding,
+			wantErr: ErrWrongEndDelimiter,
 		},
 		{
-			name:    "non-zero padding",
-			input:   append(append([]byte{0x42, 0xff}, bytes.Repeat([]byte{0}, 29)...), 0x01, 0x00),
+			name:    "wrong_delimiter",
```
joshua-kim, line 163: Why do we add underscoring between words? My guess was that tooling doesn't like the white space.... on my IDE it replaces whitespace with `_` so I wasn't sure what the rationale was here.

vms/evm/predicate/predicate_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package predicate
+package predicate_test
```
joshua-kim: Why do we make a separate `_test` package here?

vms/evm/predicate/predicate.go
joshua-kim: nit: *End seems to not be needed to disambigurate since we only have one type of delimiter... could we just call it one thing (always refer to it either as an `EndDelimiter` or a `Delimiter` in code?)

vms/evm/predicate/predicate_test.go
StephenButtolph: The rationale behind replacing `" "` with `"_"` is that if a test fails it is printed with the spaces replaced with underscores, so searching for the test is nicer if the name uses underscores rather than spaces.

vms/evm/predicate/predicate_test.go
StephenButtolph: Because `predicatetest` imports `predicate`

vms/evm/predicate/predicate_test.go
JonathanOppenheimer: if that's the case, results_test also needs to be modified

vms/evm/predicate/predicate_test.go
joshua-kim: ([ref](https://github.com/ava-labs/avalanchego/pull/4175#discussion_r2279530501)) if we could move `New` to `predicate` I think we can avoid the test package + delimiter export. Thoughts?

JonathanOppenheimer (approved):

PR #4177, opened 2025-08-15
Improve results handling

StephenButtolph:
## Why this should be merged

1. This adds explicit tests of the serialization format. This is important because any changes to the serialization format (even if they are symmetric between `Parse` and `Bytes`) could break existing logic.
2. This removes the usage of structs to instead prefer using the maps directly (which the reflect codec fully supports).
3. This aligns the behavior of `Set` and `Get` where after `Set`ing a zero value, `Get` behaves the same as if the value was never set.
4. This aligns the behavior that `Set` had previously (which is required to remain compatible with the C-chain's block building logic)

## How this works

Updated logic to better align with coreth and improved the tests to fully verify the behaviors we expect.

## How this was tested

Unit tests

No.

### DIFF
--- a/vms/evm/predicate/results.go
+++ b/vms/evm/predicate/results.go
@@ -30,14 +30,18 @@ func init() {
 	}
 }
 
-// PrecompileResults is a map of results for each precompile address to the resulting bitset.
-type PrecompileResults map[common.Address]set.Bits
+type (
+	// PrecompileResults is a map of results for each precompile address to the
+	// resulting bitset.
+	PrecompileResults map[common.Address]set.Bits
 
-// BlockResults encodes the precompile predicate results included in a block on a per transaction basis.
-// BlockResults is not thread-safe.
-type BlockResults struct {
-	TxResults map[common.Hash]PrecompileResults `serialize:"true"`
-}
+	// BlockResults maps the transactions in a block to their precompile
+	// results.
+	BlockResults map[common.Hash]PrecompileResults
+
+	// encodedBlockResults is used to serialize and deserialize BlockResults.
+	encodedBlockResults map[common.Hash]map[common.Address][]byte
+)
 
 // ParseBlockResults parses bytes into predicate results.
 func ParseBlockResults(b []byte) (BlockResults, error) {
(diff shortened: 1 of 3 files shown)


vms/evm/predicate/results.go
```diff
@@ -48,49 +52,52 @@ func ParseBlockResults(b []byte) (BlockResults, error) {
...
-// Get returns the predicate results for txHash from precompile address if available.
-func (b *BlockResults) Get(txHash common.Hash, address common.Address) (set.Bits, bool) {
-	result, ok := b.TxResults[txHash][address]
-	return result, ok
+// Get returns the predicate results for txHash from precompile address.
+func (b *BlockResults) Get(txHash common.Hash, address common.Address) set.Bits {
+	if result, ok := (*b)[txHash][address]; ok {
+		return result
+	}
+	return set.NewBits()
```
StephenButtolph, line 72: The zero value of `set.Bits` is not safe to use afaict

vms/evm/predicate/results.go
```diff
@@ -48,49 +52,52 @@ func ParseBlockResults(b []byte) (BlockResults, error) {
...
 
-// Set sets the predicate results for the given txHash. Overrides results if present.
+// Set sets the predicate results for the given txHash. Results are overwritten,
+// not merged.
 func (b *BlockResults) Set(txHash common.Hash, txResults PrecompileResults) {
-	if b.TxResults == nil {
-		b.TxResults = make(map[common.Hash]PrecompileResults)
+	if len(txResults) == 0 {
+		delete(*b, txHash)
+		return
```
StephenButtolph, line 80: This needs to be done to work with the coreth/subnet-evm block building logic.

vms/evm/predicate/results.go
```diff
@@ -48,49 +52,52 @@ func ParseBlockResults(b []byte) (BlockResults, error) {
...
 	}
 
-	return resultsCodec.Marshal(version, &encoded)
-}
-
-// encodedBlockResults is the serialized representation of BlockResults.
-type encodedBlockResults struct {
-	TxResults map[common.Hash]map[common.Address][]byte `serialize:"true"`
+	return resultsCodec.Marshal(version, encoded)
 }
```
StephenButtolph, line 103: We don't need to wrap this into a type (and removing the type wrapper doesn't change the serialization format)

vms/evm/predicate/results_test.go
```diff
@@ -4,227 +4,413 @@
...
-		{
-			name: "multiple txs and gets",
-			sets: []setArgs{
-				{
-					tx: common.Hash{1},
-					results: PrecompileResults{
-						common.Address{2}: set.NewBits(1, 2, 3),
-					},
+			},
+			want: []byte{
```
StephenButtolph, line 224: Explicitly locking in the serialization format makes me a lot more comfortable to make changes here.

vms/evm/predicate/results_test.go
```diff
@@ -4,227 +4,413 @@
...
 		{
-			name: "multiple txs mixed results",
-			results: map[common.Hash]PrecompileResults{
-				{1}: {
-					{2}: set.NewBits(1, 2, 3),
-				},
-				{2}: {
-					{2}: set.NewBits(1, 2, 3),
-					{3}: set.NewBits(3, 2, 1),
+			name: "delete_tx",
```
StephenButtolph, line 173: We need this case to support `Delete` like it had before.

vms/evm/predicate/results_test.go
```diff
@@ -4,227 +4,413 @@
...
+			input := test.input
+			if input == nil {
+				// nil and empty should be considered equivalent
+				input = make(BlockResults)
 			}
-			req.Len(br.TxResults, tc.wantLen)
+
+			parsed, err := ParseBlockResults(got)
+			require.NoError(err)
+			require.Equal(input, parsed)
```
StephenButtolph, line 413: I thought about moving this out to a separate test... But I felt like it was overly repetitive.

vms/evm/predicate/results.go
```diff
@@ -48,49 +52,52 @@ func ParseBlockResults(b []byte) (BlockResults, error) {
...
 
-// Get returns the predicate results for txHash from precompile address if available.
-func (b *BlockResults) Get(txHash common.Hash, address common.Address) (set.Bits, bool) {
-	result, ok := b.TxResults[txHash][address]
-	return result, ok
+// Get returns the predicate results for txHash from precompile address.
+func (b *BlockResults) Get(txHash common.Hash, address common.Address) set.Bits {
+	if result, ok := (*b)[txHash][address]; ok {
+		return result
```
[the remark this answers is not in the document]
StephenButtolph: This is not true.

Issue #4108, opened 2025-07-22
Add generic VM re-execution benchmark support (include P-Chain/X-Chain)


aaronbuchwald:
Update the C-Chain Re-Execution Benchmark to support taking the following inputs to setup the VM

- VM Factory
- Genesis Bytes
- Config Bytes (use to add archive + full benchmark support)
- Upgrade Bytes
- networkID/subnetID/chainID
- XChainID/CChainID (included in `*snow.Context`)
- top-level metrics to report
- Matrix of targets to run on (GitHub runner, ARC w/ different disk configurations ie. ebs, ebs-fast, etc.)

JuanLeon2:
Is the matrix of targets worthwhile independently of the generic VM?  E.g, run EVM with and without Firewood, or Firewood on EBS and SSD (I know these are not 100% correct examples)

aaronbuchwald:
> Is the matrix of targets worthwhile independently of the generic VM? E.g, run EVM with and without Firewood, or Firewood on EBS and SSD (I know these are not 100% correct examples)

It depends, there is a lot we can turn into a matrix and I think it will take a little playing around to decide what we want to run.

Right now, I'd recommend the following:

- run ARC with different storage volume types to evaluate performance on each (can use manual workflow with runner as input as opposed to setting up a matrix for all of them to run on a consistent trigger)
- potentially include a matrix based on the manual triggers we DO want to run on a schedule
- run each DB config that we want to regularly measure performance of (commit adding Firewood as an option is here: https://github.com/ava-labs/avalanchego/commit/99e9d48c61f3db5d419d65e43576be3ae4456d86)

RodrigoVillar:
Was able to get the re-execution test to work for the P-Chain here: 

```
AVL-0W5L7Y:avalanchego rodrigo.villar$ go test -timeout=0 -v -benchtime=1x -bench=BenchmarkReexecuteRange -run=^$ github.com/ava-labs/avalanchego/tests/reexecute/p --block-dir=$P_BLOCK_DIR --current-state-dir=$P_STATE_DIR --start-block=1 --end-block=9
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/tests/reexecute/p
cpu: Apple M1 Max
BenchmarkReexecuteRange
BenchmarkReexecuteRange/[1,9]-Config-default-Runner-dev
[...]

PR #4180, opened 2025-08-18
Add support for passing config and predefined configs to VM re-execution tests

aaronbuchwald:
## Why this should be merged

This PR supports passing config bytes in to the VM for re-execution tests. This is one of the features outlined in https://github.com/ava-labs/avalanchego/issues/4108.

Users specify a predefined config to use when running the benchmark. The currently supported options are "defaultt"/"", "archive", and "firewood". This does not support passing arbitrary JSON atm because the name of the config is used as part of the Benchmark run name and supplied as a Prometheus label, which means we need a simple string and don't want to convert from JSON. If desired, we could add flags to provide both a full arbitrary JSON config and a name to use as the corresponding parameter.

You can see an example of how to add a new pre-defined config here: https://github.com/ava-labs/avalanchego/commit/99e9d48c61f3db5d419d65e43576be3ae4456d86.

Note: this PR also migrates from using ARC to Blacksmith because running on limited infrastructure on ARC with a larger volume of jobs results in a much higher queueing delay with our current configuration.

This PR also reduces the execution range of the scheduled job because it is currently failing at the hour mark seemingly hitting some timeout at this point.

The README has been updated to include instructions on using the new `CONFIG` parameter, a section on adding a new entry to the schedule/PR matrix of runs, and using the GH CLI to trigger a run via manual dispatch:

https://github.com/ava-labs/avalanchego/blob/aaronbuchwald/reexecute-benchmark-configs/tests/reexecute/c/README.md

## How this works

This adds a new flag to the benchmark execution tests that will be parsed as either the name of a pre-defined config or a string to interpret as the JSON config.

## How this was tested

This was tested by adding each of the config options to run the range [101, 250k] on every PR and the schedule job here: https://github.com/ava-labs/avalanchego/actions/runs/17104240451/job/48508879094?pr=4180

Manual workflows as documented in the README to run a single job using GH CLI:

- https://github.com/ava-labs/avalanchego/actions/runs/17214249832

Tested PR triggers via CI, which are a mirror of the schedule run:

- https://github.com/ava-labs/avalanchego/actions/runs/17211815828?pr=4180

No

### DIFF
--- a/.github/workflows/c-chain-reexecution-benchmark-container.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-container.yml
@@ -84,11 +84,11 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install ARC Dependencies
         shell: bash
-        run: |	
-          # xz-utils might be present on some containers. Install if not present.	
-          if ! command -v xz &> /dev/null; then	
-            sudo apt-get update	
-            sudo apt-get install -y xz-utils	
+        run: |
+          # xz-utils might be present on some containers. Install if not present.
+          if ! command -v xz &> /dev/null; then
+            sudo apt-get update
+            sudo apt-get install -y xz-utils
           fi
       - name: Run C-Chain Re-Execution Benchmark
         uses: ./.github/actions/c-chain-reexecution-benchmark


tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -65,6 +65,25 @@ var (
 		"is_ephemeral_node": "false",
 		"chain":             "C",
 	}
+
+	configKey                           = "config"
+	defaultConfigKey                    = "default"
+	predefinedConfigs map[string]string = map[string]string{
```
RodrigoVillar:
```suggestion
	predefinedConfigs  = map[string]string{
```

We declare the type of `predefinedConfigs` twice - maybe just declare it once?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -2,6 +2,9 @@ name: 'C-Chain Re-Execution Benchmark'
 description: 'Run C-Chain re-execution benchmark'
 
 inputs:
+  config:
+    description: 'The config to pass to the VM for the benchmark. Specify "archive" (pre-defined config) or a full JSON blob as a string.'
```
RodrigoVillar: q/no-need-for-change: any preference for specifying archive rather than default ?

.github/workflows/c-chain-reexecution-benchmark-blacksmith.yml
```diff
@@ -1,9 +1,13 @@
-name: C-Chain Re-Execution Benchmark ARC
+name: C-Chain Re-Execution Benchmark Blacksmith
 
 on:
   pull_request:
   workflow_dispatch:
     inputs:
+      config:
+        description: 'The config to pass to the VM for the benchmark. Specify "archive" (pre-defined config) or a full JSON blob as a string.'
```
RodrigoVillar: q/no-need-for-change: same question regarding `archive` vs `default`.

tests/reexecute/c/README.md
```diff
@@ -98,10 +98,47 @@ Provide the parameters explicitly that we have just used locally:
...
 
+## Predefined Configs
+
+To support testing the VM in multiple configurations, the benchmark supports a set of pre-defined configs passed via the Task variable ex. `CONFIG=archive`.
+
+The currently supported options are: "default", "archive", and "firewood".
+
+Note: to execute a benchmark with any of these options, double check to ensure you are using a compatible database via `CURRENT_STATE_DIR`. For example, attempting to execute the VM with Firewood on a database using the default configuration will refuse to startup.
+
+This currently only supports pre-defined configs and not passing a full JSON blob in, so that we have a clear config name to use in the name of the sub-benchmark (used by GitHub Action Benchmark to separate historical results) and added as a label to exported Prometheus metrics.
```
RodrigoVillar, line 109: In the `.yml` files, we have:

```yml
'The config to pass to the VM for the benchmark. Specify "archive" (pre-defined config) or a full JSON blob as a string.'
```

However, this `README` says that it does not support passing in a full JSON blob?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -85,6 +104,13 @@ func TestMain(m *testing.M) {
 	flag.BoolVar(&metricsEnabledArg, "metrics-enabled", true, "Enable metrics collection.")
 	flag.StringVar(&labelsArg, "labels", "", "Comma separated KV list of metric labels to attach to all exported metrics. Ex. \"owner=tim,runner=snoopy\"")
 
+	predefinedConfigKeys := make([]string, 0, len(predefinedConfigs))
+	for k := range predefinedConfigs {
+		predefinedConfigKeys = append(predefinedConfigKeys, k)
+	}
+	predefinedConfigOptionsStr := fmt.Sprintf("[%s]", strings.Join(predefinedConfigKeys, ", "))
+	flag.StringVar(&configNameArg, configKey, "", fmt.Sprintf("Specifies the predefined config to use for the VM. Options include %s.", predefinedConfigOptionsStr))
```
RodrigoVillar: q: why not set the default value for `configNameArg` to just `defaultConfigKey`? This way, we can remove the `""` key from `predefinedConfigs` and remove the if-statement in line 124.

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: Consolidating and updating this comment to just specify the name of the benchmark it's running, so we don't document the set of options in as many places.

.github/workflows/c-chain-reexecution-benchmark-blacksmith.yml
aaronbuchwald: Same comment

tests/reexecute/c/README.md
aaronbuchwald: Nice catch, I removed JSON blob support, this README is correct.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Removing

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Good call

RodrigoVillar (approved):

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -32,57 +36,68 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
       - name: Install ARC Dependencies
-        shell: bash
-        run: |
-          # xz-utils might be present on some containers. Install if not present.
-          if ! command -v xz &> /dev/null; then
-            sudo apt-get update
-            sudo apt-get install -y xz-utils
+        shell: bash
```
maru-ava: Trailing whitespace?

.github/workflows/c-chain-reexecution-benchmark-container.json
```diff
@@ -0,0 +1,25 @@
+{
+    "pull_request": {
+        "include": [
+            {
+                "runner": "ubuntu-latest",
+                "config": "default",
+                "start-block": 101,
```
maru-ava, line 7: (No action required) Maybe define configuration like this outside of github actions in future PRs to simplify local reproducibility?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -85,6 +103,13 @@ func TestMain(m *testing.M) {
 	flag.BoolVar(&metricsEnabledArg, "metrics-enabled", true, "Enable metrics collection.")
 	flag.StringVar(&labelsArg, "labels", "", "Comma separated KV list of metric labels to attach to all exported metrics. Ex. \"owner=tim,runner=snoopy\"")
 
+	predefinedConfigKeys := make([]string, 0, len(predefinedConfigs))
```
maru-ava: (No action required) go 1.23 niceties:

```
predefinedConfigKeys := slices.Collect(maps.Keys(predefinedConfigs))
```

.github/workflows/c-chain-reexecution-benchmark-container.json
aaronbuchwald: By local reproducibility, do you mean to reproduce the exact set of triggers that are going to run on CI? ie. currently you can use the Taskfile to reproduce any individual job, but there's not a local trigger to reproduce the full matrix?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -32,57 +36,68 @@ on:
...
-        run: |
-          # xz-utils might be present on some containers. Install if not present.
-          if ! command -v xz &> /dev/null; then
-            sudo apt-get update
-            sudo apt-get install -y xz-utils
+        run: |	
+          # xz-utils might be present on some containers. Install if not present.	
+          if ! command -v xz &> /dev/null; then	
+            sudo apt-get update	
+            sudo apt-get install -y xz-utils
```
maru-ava: Please strip trailing whitespace on these lines too.

.github/workflows/c-chain-reexecution-benchmark-container.json
maru-ava: Correct, I think a matrix would ideally enable triggering tasks rather than defining configuration to use with tasks.

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: Addressed, sorry didn't realize the previous comment was intended multi-line

PR #4181, opened 2025-08-18
refactor(load): simulator contract

RodrigoVillar:
## Why this should be merged

While researching past transactions on the C-Chain (for the month of July 2025), the following trends became clear:

- Most C-Chain transactions were utilizing more gas than that of the load tests
- Most C-Chain transactions were larger in size (via calldata length) than that of the load tests
- The majority of block processing time was spent on trie operations rather than block execution

While updating the load simulator contract to account for the trends above, I identified the following issues:

1. Several of the methods in the load simulator contract are not relevant in the context of load testing. For example, `simulateMemory()`, while not computationally free, does not produce the type of load relevant for projects like Firewood (i.e. no database r/w operations).
2. The methods of the simulator contract do not have documentation on their gas-usage/calldata lengths. This makes it difficult to create load tests of different profiles (e.g. load tests that target Firewood vs load tests that target SAE).

This PR replaces the `EVMSimulatorContract` contract with `LoadSimulator`, a contract with computationally intensive methods targeted towards load testing. These methods are also documented in terms of their gas-usage and calldata lengths, making it possible to create load tests of different profiles in future PRs.

## How this works

- Replaces `EVMLoadSimulator` with `LoadSimulator` contract
- Adjusted test weights in `NewRandomTest()` to reflect the trends mentioned above 
- Updates `load/README.md` to account for changes

## How this was tested

CI + ran load test locally

No

### DIFF
--- a/tests/load/tests.go
+++ b/tests/load/tests.go
@@ -69,8 +69,11 @@ func NewRandomTest(
 		// value specifies the amount to send in a transfer test
 		value = big.NewInt(1)
 
-		// random values are written to slots to ensure that the same value isn't
-		// being written to a slot, which could reduce the expected gas used of a tx
+		// Random values are written to slots to ensure that the same value isn't
+		// being written to a slot, as the odds of a pRNG choosing the same value
+		// twice is practically zero. Using random values simplifies gas calculations
+		// as it removes the need to use an SLOAD operation to verify a different
+		// value is being written.
 		writeRand  = rand.New(rand.NewSource(0)) //#nosec G404
 		modifyRand = rand.New(rand.NewSource(1)) //#nosec G404
 	)


tests/load/tests.go
```diff
@@ -66,103 +66,85 @@ func NewRandomTest(
...
+			Weight: 10,
 		},
 		{
+			// minimum gas used: 52_300
 			Test: ERC20Test{
 				Contract: tokenContract,
 				Value:    value,
 			},
-			Weight: weight,
+			Weight: 45,
```
RodrigoVillar, line 150: I modified the weights to give higher preference to transactions that occur frequently on the C-Chain (i.e. ERC-20 transfers) or that are computationally-intensive; these weights also push the gas usage/calldata lengths of the load tests to closer match that of the C-Chain.

tests/load/contracts/LoadSimulator.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: BSD 3-Clause
```
maru-ava, line 1: Why did the license change from MIT to BSD? And is there an original copyright holder or is this new code?

tests/load/contracts/LoadSimulator.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @dev Write value to numSlots whose slot value is empty.
+     *
+     * Calldata Length: 4 + 32 + 32 = 68 bytes
+     * Minimum Gas Used: 21_000 + (22_100 * numSlots)
+     */
+    function write(uint256 numSlots, uint256 value) external returns (bool) {
+        assembly {
+            // load value stored at latestEmptySlot
```
maru-ava: (No action required) What is this comment telling me I don't already know?

tests/load/contracts/LoadSimulator.sol
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * Calldata Length: 4 + 32 + 32 = 68 bytes
+     * Minimum Gas Used: 21_000 + (5_000 * numSlots)
+     */
+    function modify(
+        uint256 numSlots,
+        uint256 newValue
+    ) external returns (bool) {
+        assembly {
+            // load value stored at latestEmptySlot
```
maru-ava: (No action required) What is this comment telling me I don't already know?

tests/load/contracts/LoadSimulator.sol
```diff
@@ -0,0 +1,134 @@
...
+     * Minimum Gas Used: 21_000 + (2100 * numSlots)
+     */
+    function read(
+        uint256 offset,
+        uint256 numSlots
+    ) external returns (uint256 sum) {
+        assembly {
+            for {
+                let i := offset
+            } lt(i, add(offset, numSlots)) {
```
maru-ava: (No action required) Would it be worth computing `add(offset, numSlots)` once and reusing the value? Same question for the instance in `write`.

tests/load/tests.go
maru-ava: Do you have a way of verifying that the transaction composition of a load test is representative - to an acceptable hand-waving degree - of current and/or historical transaction composition on mainnet (and maybe testnet)?

tests/load/tests.go
```diff
@@ -66,103 +66,85 @@ func NewRandomTest(
...
+
+		// random values are written to slots to ensure that the same value isn't
+		// being written to a slot, which could reduce the expected gas used of a tx
+		writeRand  = rand.New(rand.NewSource(0)) //#nosec G404
+		modifyRand = rand.New(rand.NewSource(1)) //#nosec G404
 	)
 
 	weightedTests := []WeightedTest{
 		{
+			// minimum gas used: 21_000
```
maru-ava, line 83: (No action required) Any value in exporting the expected minimum gas used as a test metric for comparative purposes (i.e. being able to eyeball that the actual is always greater than equal to the expected minimum)?

tests/load/contracts/LoadSimulator.sol
```diff
@@ -0,0 +1,134 @@
...
+                for {
+                    let i := 0
+                } lt(i, numSlots) {
+                    i := add(i, 1)
+                } {
+                    sstore(i, newValue)
+                }
+            }
+        }
+        return true;
```
maru-ava: (No action required) Is it necessary to return a value if it will always be true? Or should there be a failure condition (e.g. not enough non-empty slots)?

tests/load/tests.go
```diff
@@ -66,103 +66,85 @@ func NewRandomTest(
...
-		// representative of C-Chain usage patterns.
-		weight = uint64(100)
-		// count specifies how many times to repeat an operation (e.g. reads,
-		// writes, or computes) for a test that supports repeated operations.
-		// This value is arbitrary but kept constant to ensure test reproducibility.
-		count = big.NewInt(5)
 		// value specifies the amount to send in a transfer test
 		value = big.NewInt(1)
+
+		// random values are written to slots to ensure that the same value isn't
```
maru-ava: (No action required) How does this prevent the same value from being written to a slot? vs e.g.  read-then-write-or-modify.

tests/load/contracts/LoadSimulator.sol
RodrigoVillar: The original simulator code was not written by me so I don't have context as to why MIT was the license. However, since I wrote this contract which is new code, I decided to follow the AvalancheGo license convention.

tests/load/tests.go
RodrigoVillar: My thought process was that since we're using a pRNG here, the odds of it selecting the same number twice is practically zero. This also comes with the added benefit of making gas calculations simpler (write instead of having to read then write)

tests/load/contracts/LoadSimulator.sol
RodrigoVillar: Removed comment here: 32100c5

tests/load/contracts/LoadSimulator.sol
RodrigoVillar: Removed comment here: 32100c5

tests/load/contracts/LoadSimulator.sol
RodrigoVillar: Done: 73d7757

tests/load/contracts/LoadSimulator.sol
RodrigoVillar: Removed for `write()`: 7d1eb02

In the case of `modify()`, I opted for keeping the boolean return arg but setting it to true if there was enough non-empty slots to modify. Utilizing a `require()` statement instead would have result in the first instances of `ModifyTest` failing since there wouldn't be enough non-empty slots to modify.

tests/load/tests.go
RodrigoVillar: Yup; here's the histogram of gas used for July 2025:

Transaction Gas Used (%)

Here's the histogram for the load test:

Screenshot 2025-08-26 at 09 52 47

My objective was for most transactions to fit into the 50-75k gas used range, which using the current weights achieve.

tests/load/tests.go
maru-ava: Maybe detail your thinking in the comment? I think it would ideally be clear why simpler gas calculations are a worthwhile tradeoff.

RodrigoVillar: Done: d68c063

tests/load/tests.go
maru-ava: When I say no action is required, but I'm asking a question, I would like to see an answer before resolution of the comment.

PR #4184, opened 2025-08-19
Rename height field to numBlocks

samliok:
## Why this should be merged

In avalanchego, `Height` refers to the last accepted height. Renaming it could help avoid confusion between avalanchego's `height` semantics and simplex's.

### DIFF
--- a/simplex/storage.go
+++ b/simplex/storage.go
@@ -36,8 +36,8 @@ var (
 )
 
 type Storage struct {
-	// height represents the number of blocks indexed in storage.
-	height atomic.Uint64
+	// numBlocks represents the number of blocks indexed in storage.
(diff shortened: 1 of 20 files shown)


simplex/storage.go
```diff
@@ -36,8 +36,8 @@ var (
 )
 
 type Storage struct {
-	// height represents the number of blocks indexed in storage.
-	height atomic.Uint64
+	// numBlocks represents the number of blocks indexed in storage.
```
yacovm: // numBlocks represents the number of blocks indexed in storage, also known as the height of the chain

yacovm (approved):

StephenButtolph (approved):

PR #4185, opened 2025-08-19
Remove `app` package

### LINKED ISSUE #2308: Remove `app` package
The `app` package defines a lot of stuff that should exist either in the `node` package or in `main`. Ideally, `node` should define all application-level things related to running an avalanchego node whereas `main` should just handle os signals and how they interact w/ node.

mpignatelli12:
## Why this should be merged
Issue #2308 - The app package defines a lot of stuff that should exist either in the node package or in main. main is already the abstraction for the application the user interacts with and node is the layer of our business logic - so app doesn't have a clear boundary.

## How this works
Moves app/app.go logic to node and main packages, separated into files for node wiring (in node/) and process lifecycle (in main/).

## How this was tested
CI, manual build test

No


node/bootstrap/bootstrap.go
joshua-kim, line 1: What was the reason for us creating a new `bootstrap` package here?

node/bootstrap/bootstrap.go
```diff
@@ -1,33 +1,22 @@
...
-const Header = `     _____               .__                       .__
-    /  _  \___  _______  |  | _____    ____   ____ |  |__   ____    ,_ o
-   /  /_\  \  \/ /\__  \ |  | \__  \  /    \_/ ___\|  |  \_/ __ \   / //\,
-  /    |    \   /  / __ \|  |__/ __ \|   |  \  \___|   Y  \  ___/    \>> |
-  \____|__  /\_/  (____  /____(____  /___|  /\___  >___|  /\___  >    \\
-          \/           \/          \/     \/     \/     \/     \/`
-
 var _ App = (*app)(nil)
 
 type App interface {
```
joshua-kim: What are we using this interface for?

main/run.go
joshua-kim, line 1: Unlike other languages (like Java), there isn't a limit on how many classes/types can exist in a file. Typically in Go you'll see many types in a file as long as they reasonably fit together - so we don't need to put this in a separate file `run.go` - which could be interpreted as confusing since `main.go` is already the entry-point for the binary.

node/bootstrap/bootstrap.go
mpignatelli12: I had kept this to mirror the original app.App, but I was able to refactor without it

main/run.go
mpignatelli12: That makes sense -- I was able to shift that logic into main.go

node/bootstrap/bootstrap.go
mpignatelli12: This was to avoid naming conflicts and separate the logic for wiring up the node, but with a rename it can be added to the node package which would be cleaner.

main/main.go
```diff
@@ -58,15 +60,63 @@ func main() {
...
 		fmt.Printf("couldn't start node: %s\n", err)
 		os.Exit(1)
 	}
 
-	exitCode := app.Run(nodeApp)
-	os.Exit(exitCode)
+	os.Exit(Run(runner))
+}
+
+const Header = `     _____               .__                       .__
```
joshua-kim, line 75: nit: typically `consts` live at the top of the file

main/main.go
```diff
@@ -58,15 +60,63 @@ func main() {
...
+    /  _  \___  _______  |  | _____    ____   ____ |  |__   ____    ,_ o
+   /  /_\  \  \/ /\__  \ |  | \__  \  /    \_/ ___\|  |  \_/ __ \   / //\,
+  /    |    \   /  / __ \|  |__/ __ \|   |  \  \___|   Y  \  ___/    \>> |
+  \____|__  /\_/  (____  /____(____  /___|  /\___  >___|  /\___  >    \\
+          \/           \/          \/     \/     \/     \/     \/`
+
+func Run(app *node.Runner) int {
```
joshua-kim, line 82: I wonder if we even need this function - we only use it once anyways. Should we inline this into main?

node/runner.go
```diff
@@ -1,51 +1,30 @@
...
+// Runner is a wrapper around a *Node that runs in this process
+type Runner struct {
+	node       *Node
+	log        logging.Logger
+	logFactory logging.Factory
+	exitWG     sync.WaitGroup
 }
 
-func New(config nodeconfig.Config) (App, error) {
+func NewRunner(config nodeconfig.Config) (*Runner, error) {
```
joshua-kim, line 27: The code in `main` owns how to actually start a node/when to shutdown... so I think this code belongs in `main` more than it belongs in `node`.

Taking a step back... I'm not sure why we needed the old `App` type to begin with (now called `Runner`). There seems to be a clear abstraction boundary between `node` (our business logic/an avax node), and `main` (the runner/how we spin up a node), but `App`/`Runner` doesn't seem to have a clear boundary of what it owns. I think it might be most reasonable to just in-line all of this code into `main` and remove `Runner` altogether, or keep the abstraction but unexport it and leave it under `main/`.

PR #4186, opened 2025-08-20
fix: blockdb file eviction race issue

### LINKED ISSUE #4194: Flaky Test: BlockDB TestFileCache_Eviction
https://github.com/ava-labs/avalanchego/actions/runs/17131191843/job/48595661223#step:4:601

Image

DracoLi:
## Why this should be merged

Fix race for when a data file handler is evicted but another thread is still using that handler. The eviction tests caught this issue. See [example](https://github.com/ava-labs/avalanchego/actions/runs/17076901774/job/48420613984?pr=4184), [example 2](https://github.com/ava-labs/avalanchego/actions/runs/17080605904/job/48433319015?pr=4159)

Resolves #4194

## How this works

The eviction tests caught this since it simulated high evictions during concurrent calls. The simple solution here is to just retry opening the data file if we get an `os.ErrClosed` error during data file read/write, which should be rare.

## How this was tested

Tests

### DIFF
--- a/x/blockdb/database.go
+++ b/x/blockdb/database.go
@@ -1178,8 +1178,9 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
 				return fmt.Errorf("failed to sync data file after writing block %d: %w", bh.Height, err)
 			}
 		}
-		return nil
+		break
 	}
+	return nil
 }
 
 func (s *Database) updateBlockHeights(writtenBlockHeight BlockHeight) error {
--- a/x/blockdb/database.go
+++ b/x/blockdb/database.go
@@ -221,7 +221,7 @@ func New(config DatabaseConfig, log logging.Logger) (*Database, error) {
 	s := &Database{
 		config: config,
 		log:    databaseLog,
-		fileCache: lru.NewCacheWithOnEvict[int, *os.File](config.MaxDataFiles, func(_ int, f *os.File) {
+		fileCache: lru.NewCacheWithOnEvict(config.MaxDataFiles, func(_ int, f *os.File) {
 			if f != nil {
 				f.Close()
 			}
@@ -480,26 +480,32 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
 
 	// Read the complete block data
 	blockData := make(BlockData, indexEntry.Size)
-	dataFile, localOffset, err := s.getDataFileAndOffset(indexEntry.Offset)
-	if err != nil {
-		s.log.Error("Failed to read block: failed to get data file",
-			zap.Uint64("height", height),
-			zap.Uint64("dataOffset", indexEntry.Offset),
-			zap.Error(err),
-		)
-		return nil, fmt.Errorf("failed to get data file for block at height %d: %w", height, err)
+	for {
+		dataFile, localOffset, fileIndex, err := s.getDataFileAndOffset(indexEntry.Offset)
+		if err != nil {
+			s.log.Error("Failed to read block: failed to get data file",
+				zap.Uint64("height", height),
+				zap.Uint64("dataOffset", indexEntry.Offset),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to get data file for block at height %d: %w", height, err)
+		}
+		_, err = dataFile.ReadAt(blockData, int64(localOffset+uint64(sizeOfBlockEntryHeader)))
+		if err != nil {
+			if errors.Is(err, os.ErrClosed) {
+				s.fileCache.Evict(fileIndex)
+				continue
+			}
+			s.log.Error("Failed to read block: failed to read block data from file",
+				zap.Uint64("height", height),
+				zap.Uint64("localOffset", localOffset),
+				zap.Uint32("blockSize", indexEntry.Size),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to read block data from data file: %w", err)
+		}
+		break
 	}
-	_, err = dataFile.ReadAt(blockData, int64(localOffset+uint64(sizeOfBlockEntryHeader)))
-	if err != nil {
-		s.log.Error("Failed to read block: failed to read block data from file",
-			zap.Uint64("height", height),
-			zap.Uint64("localOffset", localOffset),
-			zap.Uint32("blockSize", indexEntry.Size),
-			zap.Error(err),
-		)
-		return nil, fmt.Errorf("failed to read block data from data file: %w", err)
-	}
-
 	return blockData, nil
 }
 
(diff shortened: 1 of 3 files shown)


x/blockdb/database.go
```diff
@@ -1134,16 +1149,37 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
...
+		if !dataWritten {
+			if _, err := dataFile.WriteAt(combinedBuf, int64(localOffset)); err != nil {
+				if errors.Is(err, os.ErrClosed) {
+					// ensure the file is evicted, otherwise we'll retry forever
+					s.fileCache.Evict(fileIndex)
+					continue
+				}
+				return fmt.Errorf("failed to write block to data file at offset %d: %w", offset, err)
+			}
+			dataWritten = true
```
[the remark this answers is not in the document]
DracoLi: this is for the sync call later, to avoid the write

x/blockdb/database.go
```diff
@@ -1134,16 +1149,37 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
 	combinedBuf := make([]byte, combinedBufSize)
 	copy(combinedBuf, headerBytes)
 	copy(combinedBuf[sizeOfBlockEntryHeader:], block)
-	if _, err := dataFile.WriteAt(combinedBuf, int64(localOffset)); err != nil {
-		return fmt.Errorf("failed to write block to data file at offset %d: %w", offset, err)
-	}
+	dataWritten := false
```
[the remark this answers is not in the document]
DracoLi: no need to rewrite the data if its already written. I think better to keep.

x/blockdb/database.go
```diff
@@ -567,44 +580,51 @@ func (s *Database) ReadBody(height BlockHeight) (BlockData, error) {
...
-	if err != nil {
-		s.log.Error("Failed to read body: body offset calculation overflow",
-			zap.Uint64("height", height),
-			zap.Uint64("headerOffset", headerOffset),
-			zap.Uint32("headerSize", indexEntry.HeaderSize),
-			zap.Error(err),
-		)
-		return nil, fmt.Errorf("calculating body offset would overflow for block at height %d: %w", height, err)
-	}
+	for {
```
DracoLi, line 588: the read body and header methods will be moved soon.

x/blockdb/database.go
```diff
@@ -480,26 +480,32 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
...
-	_, err = dataFile.ReadAt(blockData, int64(localOffset+uint64(sizeOfBlockEntryHeader)))
-	if err != nil {
-		s.log.Error("Failed to read block: failed to read block data from file",
-			zap.Uint64("height", height),
-			zap.Uint64("localOffset", localOffset),
-			zap.Uint32("blockSize", indexEntry.Size),
-			zap.Error(err),
-		)
-		return nil, fmt.Errorf("failed to read block data from data file: %w", err)
+	for {
```
StephenButtolph, line 486: We should probably add a comment as to why this loop is needed.

x/blockdb/database_test.go
```diff
@@ -235,7 +235,7 @@ func TestFileCache_Eviction(t *testing.T) {
 	// Override the file cache with a smaller size to force evictions
 	evictionCount := atomic.Int32{}
 	evictionMu := sync.Mutex{}
-	smallCache := lru.NewCacheWithOnEvict[int, *os.File](3, func(_ int, file *os.File) {
+	smallCache := lru.NewCacheWithOnEvict(3, func(_ int, file *os.File) {
```
yacovm: Is it possible to make a test that fails without the fix with a high probability?

x/blockdb/database.go
```diff
@@ -531,24 +537,31 @@ func (s *Database) ReadHeader(height BlockHeight) (BlockData, error) {
...
+			}
+			s.log.Error("Failed to read header: failed to read header data from file",
+				zap.Uint64("height", height),
+				zap.Uint64("localOffset", localOffset),
+				zap.Uint32("headerSize", indexEntry.HeaderSize),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to read block header data from data file: %w", err)
+		}
+		break
```
yacovm: can we explicitly use `return` here instead?

x/blockdb/database.go
```diff
@@ -567,44 +580,51 @@ func (s *Database) ReadBody(height BlockHeight) (BlockData, error) {
...
+			}
+			s.log.Error("Failed to read body: failed to read body data from file",
+				zap.Uint64("height", height),
+				zap.Uint64("bodyOffset", bodyOffset),
+				zap.Uint32("bodySize", bodySize),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to read block body data from data file: %w", err)
+		}
+		break
```
yacovm: can we explicitly use `return` here?

x/blockdb/database.go
```diff
@@ -1134,14 +1149,36 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
...
+		if s.config.SyncToDisk {
+			if err := dataFile.Sync(); err != nil {
+				if errors.Is(err, os.ErrClosed) {
+					s.fileCache.Evict(fileIndex)
+					continue
+				}
+				return fmt.Errorf("failed to sync data file after writing block %d: %w", bh.Height, err)
+			}
 		}
+		break
```
yacovm: same, can we use `return` here?

x/blockdb/database.go
```diff
@@ -1134,14 +1149,36 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
...
-	if s.config.SyncToDisk {
-		if err := dataFile.Sync(); err != nil {
-			return fmt.Errorf("failed to sync data file after writing block %d: %w", bh.Height, err)
+	for {
+		dataFile, localOffset, fileIndex, err := s.getDataFileAndOffset(offset)
+		if err != nil {
+			return fmt.Errorf("failed to get data file for writing block %d: %w", bh.Height, err)
+		}
+
+		if !dataWritten {
```
yacovm: why do we skip overwriting? 

If we have pending bytes to be written to the file but we closed the file, are we guaranteed that if we re-open the file and call sync, the changes from before will be written? 

Isn't that dependent on the implementation of the file handle? The file handle in the next iteration isn't the same as the one in the previous one.

Can't we just overwrite the changes again? we're not appending to the file but writing to a specific offset, right?  Then the code will also be simpler.

DracoLi: yeah good point, I'm convinced. I think its fine to just rewrite the data in this case if sync fails.

yacovm (approved):

x/blockdb/database_test.go
DracoLi: yep, i added a test that trigger this 100%

x/blockdb/database.go
```diff
@@ -221,7 +221,7 @@ func New(config DatabaseConfig, log logging.Logger) (*Database, error) {
 	s := &Database{
 		config: config,
 		log:    databaseLog,
-		fileCache: lru.NewCacheWithOnEvict[int, *os.File](config.MaxDataFiles, func(_ int, f *os.File) {
```
joshua-kim: nit: remove unrelated diff

x/blockdb/database.go
```diff
@@ -480,26 +480,35 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
...
+			}
+			s.log.Error("Failed to read block: failed to read block data from file",
+				zap.Uint64("height", height),
+				zap.Uint64("localOffset", localOffset),
+				zap.Uint32("blockSize", indexEntry.Size),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to read block data from data file: %w", err)
+		}
+		break
```
joshua-kim: Should we just move the `return` to here?

x/blockdb/database.go
```diff
@@ -480,26 +480,35 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
...
+				zap.Uint64("height", height),
+				zap.Uint64("dataOffset", indexEntry.Offset),
+				zap.Error(err),
+			)
+			return nil, fmt.Errorf("failed to get data file for block at height %d: %w", height, err)
+		}
+		_, err = dataFile.ReadAt(blockData, int64(localOffset+uint64(sizeOfBlockEntryHeader)))
+		if err != nil {
+			if errors.Is(err, os.ErrClosed) {
+				s.fileCache.Evict(fileIndex)
```
joshua-kim, line 499: I don't understand why we need to call evict here - evicting will close the file handle but doesn't this error mean it's already closed?

x/blockdb/database.go
```diff
@@ -1134,16 +1154,35 @@ func (s *Database) writeBlockAt(offset uint64, bh blockEntryHeader, block BlockD
...
+	for {
+		dataFile, localOffset, fileIndex, err := s.getDataFileAndOffset(offset)
+		if err != nil {
+			return fmt.Errorf("failed to get data file for writing block %d: %w", bh.Height, err)
+		}
+
+		if _, err := dataFile.WriteAt(combinedBuf, int64(localOffset)); err != nil {
+			if errors.Is(err, os.ErrClosed) {
+				// ensure the file is evicted, otherwise we'll retry forever
+				s.fileCache.Evict(fileIndex)
```
joshua-kim, line 1168: Ah nvm the reason is documented here.

x/blockdb/database.go
DracoLi: yeah, i missed updating this one

PR #4193, opened 2025-08-21
Rename Engine Types

samliok:
## Why this should be merged

Better rename of the engine types. The previous `ENGINE_TYPE` names were slightly misleading especially since the `SNOWMAN` engine type could soon represent a simplex engine.

### DIFF
--- a/network/p2p/throttler.go
+++ b/network/p2p/throttler.go
@@ -82,6 +82,8 @@ func (s *SlidingWindowThrottler) Handle(nodeID ids.NodeID) bool {
 
 	currentHits := s.windows[s.current].hits
 	current := currentHits[nodeID]
+	currentHits[nodeID]++
+
 	previousFraction := float64(s.period-sinceUpdate) / float64(s.period)
 	previous := s.windows[1-s.current].hits[nodeID]
 	estimatedHits := current + previousFraction*previous
@@ -90,7 +92,6 @@ func (s *SlidingWindowThrottler) Handle(nodeID ids.NodeID) bool {
 		return false
 	}
 
-	currentHits[nodeID]++
 	return true
 }
 
--- a/network/p2p/throttler_test.go
+++ b/network/p2p/throttler_test.go
@@ -121,6 +121,15 @@ func TestSlidingWindowThrottlerHandle(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "throttled requests are counted as hits",
+			limit: 1,
+			calls: []call{
+				{time: currentWindowStartTime},
+				{time: currentWindowStartTime.Add(period).Add(time.Second), throttled: true},
+				{time: currentWindowStartTime.Add(2 * period).Add(time.Second), throttled: true},
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/x/blockdb/database.go
+++ b/x/blockdb/database.go
@@ -480,27 +480,35 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
 
 	// Read the complete block data
 	blockData := make(BlockData, indexEntry.Size)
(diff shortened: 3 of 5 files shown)


proto/p2p/p2p.proto
```diff
@@ -215,8 +215,8 @@ message AcceptedStateSummary {
 enum EngineType {
   ENGINE_TYPE_UNSPECIFIED = 0;
   // Only the X-Chain uses avalanche consensus
```
joshua-kim: nit: update comment

snow/networking/handler/engine_test.go
```diff
@@ -49,7 +49,7 @@ func TestEngineManager_Get(t *testing.T) {
 		{
 			name: "request snowman engine",
```
joshua-kim: nit: update test name

chains/manager.go
```diff
@@ -615,7 +615,7 @@ func (m *manager) createAvalancheChain(
 	defer ctx.Lock.Unlock()
 
 	ctx.State.Set(snow.EngineState{
-		Type:  p2ppb.EngineType_ENGINE_TYPE_AVALANCHE,
+		Type:  p2ppb.EngineType_ENGINE_TYPE_DAG,
```
joshua-kim, line 618: Unrelated to this PR - but I hope we someday stop calling things consensus "engines" hahaha. This naming pre-dates me, but I think it comes from the fact that we have consensus "gears" (bootstrapping, normal op, etc) ([ref](https://github.com/ava-labs/avalanchego/blob/45df9e6926b93aacd11589a7b90a76078ae956d9/snow/engine/common/no_ops_handlers.go#L42)). I find this analogy to a motor vehicle's operation to be unnecessarily confusing... my mental model is that it's just a state transition.

joshua-kim:
CI flake looks related to https://github.com/ava-labs/avalanchego/pull/4186/files

snow/networking/handler/engine_test.go
```diff
@@ -59,8 +59,8 @@ func TestEngineManager_Get(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			e := EngineManager{
-				Avalanche: avalanche,
-				Snowman:   snowman,
+				DAG:   avalanche,
+				Chain: snowman,
```
joshua-kim: nit: we need to update these names too

samliok: i think it's fine having the variable names be `avalanche` and `snowman`

joshua-kim: Ugh.. I just realized there's so many usages of the term `avalanche*` and `snowman*` and I don't want to verify every usage (there's non-code examples too, like in error messages). I think keeping the scope to the proto is reasonable.

snow/networking/handler/handler.go
```diff
@@ -508,7 +508,7 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
 
 	var engineType p2ppb.EngineType
 	switch msg.EngineType {
-	case p2ppb.EngineType_ENGINE_TYPE_AVALANCHE, p2ppb.EngineType_ENGINE_TYPE_SNOWMAN:
+	case p2ppb.EngineType_ENGINE_TYPE_DAG, p2ppb.EngineType_ENGINE_TYPE_CHAIN:
```
StephenButtolph, line 511: The comment below here should be updated.

yacovm (approved):

snow/networking/handler/engine_test.go
StephenButtolph: I feel like there really aren't that many places that reference avalanche vs snowman engines... It feels like we should update these var names (and update the one comment).

I didn't find any other unexpected usage

PR #4195, opened 2025-08-21
Improve createSnowChain readability

samliok:
## Why this should be merged

Only stylistic changes to `createSnowmanChain` of the chain manager. `createSnowmanChain` initializes the `VM`, `Handler` and sets other components required to create a snowman chain. It was overly dense so I just separated it out into many helper functions to make it easier to digest. Very open to addition style suggestions, to make it even more understandable!

### DIFF
--- a/chains/manager.go
+++ b/chains/manager.go
@@ -1544,7 +1544,8 @@ func (m *manager) createSnowmanEngine(
 	vdrs validators.Manager,
 	connectedValidators tracker.Peers,
 	peerTracker *p2p.PeerTracker,
-	halter common.Halter) (*handler.Engine, error) {
+	halter common.Halter,
+) (*handler.Engine, error) {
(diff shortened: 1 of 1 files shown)


chains/manager.go
```diff
@@ -1069,25 +1067,251 @@ func (m *manager) createSnowmanChain(
...
+	chain, exists := m.chains[id]
+	m.chainsLock.Unlock()
+	if !exists {
+		return false
+	}
+
+	return chain.Context().State.Get().State == snow.NormalOp
+}
+
+func (m *manager) registerBootstrappedHealthChecks() error {
```
joshua-kim, line 1142: This diff contains untouched but moved code hunks - which makes this diff harder to review. I'm open to us having helper functions but these code moves look unrelated to me - could we unstage these?

PR #4196, opened 2025-08-21
fix: time remaining should use better estimates

rkuris:
## Why this should be merged

This code does a much better job of estimating completion time, which improves my experience using avalanchego for bootstrapping.

## How this works

Correctly computing an estimate involves:

 - Estimating high early on, but gradually removing the fuzz as the load progresses. This code uses a 20% margin for this.
 - Using a rolling window of samples to see what the rate is. This prevents a few large or slow blocks from making a big impact on the estimate right way. This code uses a window size of 10.
  - Don't report estimates until we have some reasonable number of samples. Estimates will not be reported for about a minute with this code (10 samples at a minimum of 5s apart)
 - Include a percentage complete. I found myself running this calculation all the time and it's nice to see how far along it is as a percentage.

## How this was tested

Too many bootstraps. The 20% margin is a wild guess, but I wrote some code to show how much fuzz makes sense, and settled on 10 samples. It was not tested on a very slow machine.

Not a bad idea.

### DIFF
--- a/snow/engine/snowman/bootstrap/bootstrapper.go
+++ b/snow/engine/snowman/bootstrap/bootstrapper.go
@@ -41,6 +41,9 @@ const (
 	// outstanding when broadcasting.
 	maxOutstandingBroadcastRequests = 50
 
+	// minimumLogInterval is the minimum time between log entries to avoid noise
+	minimumLogInterval = 5 * time.Second
+
 	epsilon = 1e-6 // small amount to add to time to avoid division by 0
 )
 
@@ -95,6 +98,11 @@ type Bootstrapper struct {
 	initiallyFetched uint64
 	// Time that startSyncing was last called
 	startTime time.Time
+	// Time of the last progress update for accurate ETA calculation
+	lastProgressUpdateTime time.Time
+
+	// ETA tracker for more accurate time estimates
+	etaTracker *timer.EtaTracker
 
 	// tracks which validators were asked for which containers in which requests
 	outstandingRequests     *bimap.BiMap[common.Request, ids.ID]
@@ -138,6 +146,8 @@ func New(config Config, onFinished func(ctx context.Context, lastReqID uint32) e
 
 		executedStateTransitions: math.MaxInt,
 		onFinished:               onFinished,
+		lastProgressUpdateTime:   time.Now(),
+		etaTracker:               timer.NewEtaTracker(10, 1.2),
 	}
 
 	timeout := func() {
@@ -407,6 +417,10 @@ func (b *Bootstrapper) startSyncing(ctx context.Context, acceptedBlockIDs []ids.
 	b.initiallyFetched = b.tree.Len()
 	b.startTime = time.Now()
 
+	// Add the first sample to the EtaTracker to establish an accurate baseline
+	// This ensures t.lowestSample is set correctly from the start
+	b.etaTracker.AddSample(b.initiallyFetched, b.tipHeight-b.startingHeight, b.startTime)
+
 	// Process received blocks
 	for _, blk := range toProcess {
 		if err := b.process(ctx, blk, nil); err != nil {
@@ -603,26 +617,37 @@ func (b *Bootstrapper) process(
 		height := blk.Height()
 		b.tipHeight = max(b.tipHeight, height)
 
-		if numPreviouslyFetched/statusUpdateFrequency != numFetched/statusUpdateFrequency {
+		// Check if it's time to log progress (both progress-based and time-based frequency)
+		now := time.Now()
+		shouldLog := numPreviouslyFetched/statusUpdateFrequency != numFetched/statusUpdateFrequency &&
+			now.Sub(b.lastProgressUpdateTime) >= minimumLogInterval
+
+		if shouldLog {
 			totalBlocksToFetch := b.tipHeight - b.startingHeight
-			eta := timer.EstimateETA(
-				b.startTime,
-				numFetched-b.initiallyFetched,         // Number of blocks we have fetched during this run
-				totalBlocksToFetch-b.initiallyFetched, // Number of blocks we expect to fetch during this run
+
+			eta, progressPercentage := b.etaTracker.AddSample(
+				numFetched,
+				totalBlocksToFetch,
+				now,
 			)
 
-			if !b.restarted {
-				b.Ctx.Log.Info("fetching blocks",
-					zap.Uint64("numFetchedBlocks", numFetched),
-					zap.Uint64("numTotalBlocks", totalBlocksToFetch),
-					zap.Duration("eta", eta),
-				)
-			} else {
-				b.Ctx.Log.Debug("fetching blocks",
-					zap.Uint64("numFetchedBlocks", numFetched),
-					zap.Uint64("numTotalBlocks", totalBlocksToFetch),
-					zap.Duration("eta", eta),
-				)
+			// Update the last progress update time and previous progress for next iteration
+			b.lastProgressUpdateTime = now
+
+			// Only log if we have a valid ETA estimate
+			if eta != nil {
+				logger := b.Ctx.Log.Info
+				if b.restarted {
+					// lower log level for restarted bootstrapping
+					// Kept for compatibility; is this desired?
+					logger = b.Ctx.Log.Debug
+				}
+				logger("fetching blocks",
+						zap.Uint64("numFetchedBlocks", numFetched),
+						zap.Uint64("numTotalBlocks", totalBlocksToFetch),
+						zap.Duration("eta", *eta),
+						zap.Float64("pctComplete", progressPercentage),
+					)
 			}
 		}
 	}
--- a/snow/engine/snowman/bootstrap/storage.go
+++ b/snow/engine/snowman/bootstrap/storage.go
@@ -165,6 +165,7 @@ func execute(
 
 		startTime     = time.Now()
 		timeOfNextLog = startTime.Add(logPeriod)
+		etaTracker    = timer.NewEtaTracker(10, 1.2)
 	)
 	defer func() {
 		iterator.Release()
@@ -194,6 +195,9 @@ func execute(
 	log("executing blocks",
 		zap.Uint64("numToExecute", totalNumberToProcess),
 	)
+
+	// Add the first sample to the EtaTracker to establish an accurate baseline
+	etaTracker.AddSample(0, totalNumberToProcess, startTime)
 
 	for !shouldHalt() && iterator.Next() {
 		blkBytes := iterator.Value()
@@ -239,13 +243,22 @@ func execute(
 		if now := time.Now(); now.After(timeOfNextLog) {
 			var (
 				numProcessed = totalNumberToProcess - tree.Len()
-				eta          = timer.EstimateETA(startTime, numProcessed, totalNumberToProcess)
 			)
-			log("executing blocks",
-				zap.Uint64("numExecuted", numProcessed),
-				zap.Uint64("numToExecute", totalNumberToProcess),
-				zap.Duration("eta", eta),
-			)
+
+			// Use the tracked previous progress for accurate ETA calculation
+			currentProgress := numProcessed
+
+			etaPtr, progressPercentage := etaTracker.AddSample(currentProgress, totalNumberToProcess, now)
+			// Only log if we have a valid ETA estimate
+			if etaPtr != nil {
+				log("executing blocks",
+					zap.Uint64("numExecuted", numProcessed),
+					zap.Uint64("numToExecute", totalNumberToProcess),
+					zap.Duration("eta", *etaPtr),
+					zap.Float64("pctComplete", progressPercentage),
+				)
+			}
+
 			timeOfNextLog = now.Add(logPeriod)
 		}
 
--- a/utils/timer/eta.go
+++ b/utils/timer/eta.go
@@ -5,6 +5,7 @@ package timer
 
 import (
 	"encoding/binary"
+	"math"
 	"time"
 )
 
@@ -18,8 +19,108 @@ func ProgressFromHash(b []byte) uint64 {
 	return binary.BigEndian.Uint64(progress[:])
 }
 
-// EstimateETA attempts to estimate the remaining time for a job to finish given
-// the [startTime] and it's current progress.
+// A sample represents a completed amount and the timestamp of the sample
+type sample struct {
+	completed uint64
+	timestamp time.Time
+}
+
+// A EtaTracker tracks the ETA of a job
+type EtaTracker struct {
+	samples        []sample
+	samplePosition uint8
+	maxSamples     uint8
+	lowestSample   uint64
+	totalSamples   uint64
+	slowdownFactor float64
+}
+
+// NewEtaTracker creates a new EtaTracker with the given maximum number of samples
+// and a slowdown factor. The slowdown factor is a multiplier that is added to the ETA
+// based on the percentage completed.
+//
+// The adjustment works as follows:
+//   - At 0% progress: ETA is multiplied by complexityFactor
+//   - At 100% progress: ETA is the raw estimate (no adjustment)
+//   - Between 0% and 100%: Adjustment decreases linearly with progress
+//
+// Example: With slowdownFactor = 2.0:
+//   - At 0% progress: ETA = raw_estimate * 2.0
+//   - At 50% progress: ETA = raw_estimate * 1.5
+//   - At 100% progress: ETA = raw_estimate * 1.0
+//
+// If maxSamples is less than 1, it will default to 5
+func NewEtaTracker(maxSamples uint8, slowdownFactor float64) *EtaTracker {
+	if maxSamples < 1 {
+		maxSamples = 5
+	}
+	return &EtaTracker{
+		samples:        make([]sample, maxSamples),
+		samplePosition: 0,
+		maxSamples:     maxSamples,
+		lowestSample:   math.MaxUint64, // Initialize to maximum value
+		totalSamples:   0,
+		slowdownFactor: slowdownFactor,
+	}
+}
+
+// AddSample adds a sample to the EtaTracker
+// It returns the remaining time to complete the target and the percent complete
+// The returned values are rounded to the nearest second and 2 decimal places respectively
+// This function can return a nil time.Duration indicating that there are not yet enough
+// samples to calculate an accurate ETA.
+//
+// The first sample should be at 0% progress to establish a baseline
+func (t *EtaTracker) AddSample(completed uint64, target uint64, timestamp time.Time) (remaining *time.Duration, percentComplete float64) {
+	if completed < t.lowestSample {
+		t.lowestSample = completed
+	}
+
+	sample := sample{
+		completed: completed,
+		timestamp: timestamp,
+	}
+	// save the oldest sample; this will not be used if we don't have enough samples
+	t.samples[t.samplePosition] = sample
+	t.samplePosition = (t.samplePosition + 1) % t.maxSamples
+	t.totalSamples++
+
+	// If we don't have enough samples, return nil
+	if t.totalSamples < uint64(t.maxSamples) {
+		return nil, 0.0
+	}
+
+	oldestSample := t.samples[t.samplePosition]
+
+	// Calculate the time and progress since the oldest sample
+	timeSinceOldest := sample.timestamp.Sub(oldestSample.timestamp)
+	progressSinceOldest := sample.completed - oldestSample.completed
+
+	// Check if target is already completed or exceeded
+	if sample.completed >= target {
+		zeroDuration := time.Duration(0)
+		percentComplete := float64(sample.completed) / float64(target)
+		roundedPercentComplete := math.Round(percentComplete*10000) / 100 // Return percentage (0.0 to 100.0)
+		return &zeroDuration, roundedPercentComplete
+	}
+
+	rate := float64(progressSinceOldest) / float64(timeSinceOldest)
+	remainingProgress := target - sample.completed
+
+	actualPercentComplete := float64(sample.completed) / float64(target)
+	roundedScaledPercentComplete := math.Round(actualPercentComplete*10000) / 100
+
+	duration := float64(remainingProgress) / rate
+	adjustment := t.slowdownFactor - (t.slowdownFactor-1.0)*actualPercentComplete
+
+	adjustedDuration := duration * adjustment
+	eta := time.Duration(adjustedDuration)
+	roundedEta := eta.Round(time.Second)
+	return &roundedEta, roundedScaledPercentComplete
+}
+
+// This function calculates ETA from start time
+// Deprecated: use EtaTracker instead
 func EstimateETA(startTime time.Time, progress, end uint64) time.Duration {
 	timeSpent := time.Since(startTime)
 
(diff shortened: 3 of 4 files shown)


utils/timer/eta.go
```diff
@@ -18,8 +19,108 @@ func ProgressFromHash(b []byte) uint64 {
...
+// If maxSamples is less than 1, it will default to 5
+func NewEtaTracker(maxSamples uint8, slowdownFactor float64) *EtaTracker {
+	if maxSamples < 1 {
+		maxSamples = 5
+	}
+	return &EtaTracker{
+		samples:        make([]sample, maxSamples),
+		samplePosition: 0,
+		maxSamples:     maxSamples,
+		lowestSample:   math.MaxUint64, // Initialize to maximum value
```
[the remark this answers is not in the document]
rkuris: hmm, I think it's the lowestSample. Perhaps it's also minCompleted but lowestSample better represents how it is used in the code.

utils/timer/eta_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEtaTracker(t *testing.T) {
+	tracker := NewEtaTracker(3, 1.0)
+	now := time.Now()
+	target := uint64(1000)
+
+	thirdSampleEta := 80 * time.Second
```
rkuris: stupid go question: is there a better way to code this? I am only using this later to point to it.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -138,6 +146,8 @@ func New(config Config, onFinished func(ctx context.Context, lastReqID uint32) e
 
 		executedStateTransitions: math.MaxInt,
 		onFinished:               onFinished,
+		lastProgressUpdateTime:   time.Now(),
+		etaTracker:               timer.NewEtaTracker(10, 1.2),
```
alarso16, line 150: Maybe label these constants at the top of the file. The numbers feel really arbitrary, and are reused for the other timer. Also, how did you select these?

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -603,25 +617,36 @@ func (b *Bootstrapper) process(
...
+			// Update the last progress update time and previous progress for next iteration
+			b.lastProgressUpdateTime = now
+
+			// Only log if we have a valid ETA estimate
+			if eta != nil {
+				logger := b.Ctx.Log.Info
+				if b.restarted {
+					// Lower log level for restarted bootstrapping.
+					// Kept for compatibility; is this desired?
+					logger = b.Ctx.Log.Debug
```
alarso16, line 642: I agree, this is weird

snow/engine/snowman/bootstrap/storage.go
```diff
@@ -237,15 +241,22 @@ func execute(
...
-			)
-			log("executing blocks",
-				zap.Uint64("numExecuted", numProcessed),
-				zap.Uint64("numToExecute", totalNumberToProcess),
-				zap.Duration("eta", eta),
-			)
+			numProcessed := totalNumberToProcess - tree.Len()
+
+			// Use the tracked previous progress for accurate ETA calculation
+			currentProgress := numProcessed
```
alarso16, line 247: What's the point of this? Won't these variables always be the same?

utils/timer/eta.go
```diff
@@ -18,8 +19,108 @@ func ProgressFromHash(b []byte) uint64 {
...
 
-// EstimateETA attempts to estimate the remaining time for a job to finish given
-// the [startTime] and it's current progress.
+// A sample represents a completed amount and the timestamp of the sample
+type sample struct {
+	completed uint64
+	timestamp time.Time
+}
+
+// An EtaTracker tracks the ETA of a job
```
DracoLi:
```suggestion
// EtaTracker tracks the ETA of a job
```
nit: start with the struct or func name for convention

utils/timer/eta_test.go
```diff
@@ -0,0 +1,73 @@
...
+		expectedPercent float64
+		description     string
+	}{
+		{
+			name:            "first sample - insufficient data",
+			completed:       0,
+			timestamp:       now,
+			expectedEta:     nil,
+			expectedPercent: 0.0,
+			description:     "should return nil ETA and 0% complete",
```
DracoLi: the `description` property doesn't seem like its used. maybe just add it as a comment instead?

utils/timer/eta_test.go
DracoLi: i think the other option is writing a helper like timePtr

utils/timer/eta.go
```diff
@@ -18,8 +19,108 @@ func ProgressFromHash(b []byte) uint64 {
...
+// AddSample adds a sample to the EtaTracker
+// It returns the remaining time to complete the target and the percent complete
+// The returned values are rounded to the nearest second and 2 decimal places respectively
+// This function can return a nil time.Duration indicating that there are not yet enough
+// samples to calculate an accurate ETA.
+//
+// The first sample should be at 0% progress to establish a baseline
+func (t *EtaTracker) AddSample(completed uint64, target uint64, timestamp time.Time) (remaining *time.Duration, percentComplete float64) {
+	if completed < t.lowestSample {
+		t.lowestSample = completed
```
DracoLi: what's `lowestSample` used for? Looks like its assigned but not used.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -407,6 +417,10 @@ func (b *Bootstrapper) startSyncing(ctx context.Context, acceptedBlockIDs []ids.
 	b.initiallyFetched = b.tree.Len()
 	b.startTime = time.Now()
 
+	// Add the first sample to the EtaTracker to establish an accurate baseline
+	// This ensures t.lowestSample is set correctly from the start
```
DracoLi: nit: should remove ref to `lowestSample`

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: From the commit comment:
> The 20% margin is a wild guess, but I wrote some code to show how much fuzz makes sense, and settled on 10 samples. It was not tested on a very slow machine.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+		completed: completed,
+		timestamp: timestamp,
+	}
+	// save the oldest sample; this will not be used if we don't have enough samples
+	t.samples[t.samplePosition] = sample
+	t.samplePosition = (t.samplePosition + 1) % t.maxSamples
+	t.totalSamples++
+
+	// If we don't have enough samples, return nil
+	if t.totalSamples < uint64(t.maxSamples) {
```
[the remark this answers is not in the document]
rkuris: I think this comment is incorrect. Say there are 2 slots, t.samples[0] and t.samples[1], so maxSamples is 2. On the first call, t.totalSamples will be 1 when you get to this line, but the only sample is t.samples[0], so we correctly return nil. On the second sample, t.samples[0] and t.samples[1] are both populated, so we can proceed.

If I make the change recommended by copilot, we'll not provide an estimate until we've overwritten one sample.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+
+	// Calculate the time and progress since the oldest sample
+	timeSinceOldest := sample.timestamp.Sub(oldestSample.timestamp)
+	progressSinceOldest := sample.completed - oldestSample.completed
+
+	// Check if target is already completed or exceeded
+	if sample.completed >= target {
+		zeroDuration := time.Duration(0)
+		percentComplete := float64(sample.completed) / float64(target)
+		roundedPercentComplete := math.Round(percentComplete*10000) / 100 // Return percentage (0.0 to 100.0)
```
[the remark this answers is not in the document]
rkuris: I think the comment explains what this is doing. I suppose this could be moved into a function for clarity, but that seems a bit overkill for this simple calculation.

utils/timer/eta.go
```diff
@@ -18,8 +19,102 @@ func ProgressFromHash(b []byte) uint64 {
...
+		return &zeroDuration, roundedPercentComplete
```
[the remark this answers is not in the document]
rkuris: If target is 0, we'll be finished before we ever get here.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	}
+
+	if timeSinceOldest == 0 {
+		return nil, 0.0
+	}
+	rate := float64(progressSinceOldest) / float64(timeSinceOldest)
+	if rate == 0 {
```
alarso16, line 102: It seems like really this should be `progressSinceOldest == 0`. Is it possible for floating point operations to mess up the distinction?

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+
+	adjustedDuration := duration * adjustment
+	eta := time.Duration(adjustedDuration)
+	roundedEta := eta.Round(time.Second)
+	return &roundedEta, roundedScaledPercentComplete
+}
+
+// EstimateETA calculates ETA from start time and current progress.
+// Deprecated: use EtaTracker instead
 func EstimateETA(startTime time.Time, progress, end uint64) time.Duration {
```
alarso16, line 124: Can you just delete this? I mean I guess if there's other users, but yours is better!

rkuris: There are still some references to the old method. I didn't change them because I'm not sure how to test them, and this code is focused on the bootstrapping which I can test.

utils/timer/eta.go
rkuris: Nope. Zero divided by anything on any platform is still zero (except when the denominator is zero, that makes NaN still).

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: I'm going to leave this. If someone wants to tackle it, though, it's not a bad idea. Should I create an issue?

snow/engine/snowman/bootstrap/storage.go
rkuris: Just documentation. the parameter for etaTracker is currentProgress, and so I wanted to show why it was being used. If you think that just adds confusion, I can easily remove it.

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -603,25 +617,36 @@ func (b *Bootstrapper) process(
...
-				b.Ctx.Log.Debug("fetching blocks",
+			// Update the last progress update time and previous progress for next iteration
+			b.lastProgressUpdateTime = now
+
+			// Only log if we have a valid ETA estimate
+			if eta != nil {
+				logger := b.Ctx.Log.Info
+				if b.restarted {
+					// Lower log level for restarted bootstrapping.
+					// Kept for compatibility; is this desired?
```
StephenButtolph: Yes it is desired.
```suggestion
```

utils/timer/eta_test.go
```diff
@@ -0,0 +1,74 @@
...
+			completed:       600,
+			timestamp:       now.Add(40 * time.Second),
+			expectedEta:     timePtr(24 * time.Second),
+			expectedPercent: 60.0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			eta, percentComplete := tracker.AddSample(tt.completed, target, tt.timestamp)
```
StephenButtolph: It is discouraged from including mutable state within `t.Run` calls. Specifically, it can prevent running specific test cases.

If I run the "third sample" test, it fails:
```
go test -run ^TestEtaTracker/third_sample_-_sufficient_data_for_ETA$ github.com/ava-labs/avalanchego/utils/timer        
--- FAIL: TestEtaTracker (0.00s)
    --- FAIL: TestEtaTracker/third_sample_-_sufficient_data_for_ETA (0.00s)
        eta_test.go:67: 
                Error Trace:    /Users/stephen/go/src/github.com/ava-labs/avalanchego/utils/timer/eta_test.go:67
                Error:          Not equal: 
                                expected: false
                                actual  : true
                Test:           TestEtaTracker/third_sample_-_sufficient_data_for_ETA
FAIL
FAIL    github.com/ava-labs/avalanchego/utils/timer     0.214s
FAIL
```
The easiest way to fix this is probably just the removal of the call to `t.Run` and include the "`name`" in the `require` calls.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	adjustment := t.slowdownFactor - (t.slowdownFactor-1.0)*actualPercentComplete
+
+	adjustedDuration := duration * adjustment
+	eta := time.Duration(adjustedDuration)
+	roundedEta := eta.Round(time.Second)
+	return &roundedEta, roundedScaledPercentComplete
+}
+
+// EstimateETA calculates ETA from start time and current progress.
+// Deprecated: use EtaTracker instead
```
StephenButtolph:
```suggestion
// EstimateETA calculates ETA from start time and current progress.
//
// Deprecated: use EtaTracker instead
```

It seems that (at least VSCode) needs an additional newline to actually mark it as deprecated.

Without newline: 
image

With newline: 
image

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+}
+
+// AddSample adds a sample to the EtaTracker
+// It returns the remaining time to complete the target and the percent complete
+// The returned values are rounded to the nearest second and 2 decimal places respectively
+// This function can return a nil time.Duration indicating that there are not yet enough
+// samples to calculate an accurate ETA.
+//
+// The first sample should be at 0% progress to establish a baseline
+func (t *EtaTracker) AddSample(completed uint64, target uint64, timestamp time.Time) (remaining *time.Duration, percentComplete float64) {
```
StephenButtolph: named return vars are a bit cursed in golang imo (we aggressively avoid them in avalanchego). While they can be nice for documentation... They can cause some really weird behavior...
```suggestion
func (t *EtaTracker) AddSample(completed uint64, target uint64, timestamp time.Time) (*time.Duration, float64) {
```

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+type sample struct {
+	completed uint64
+	timestamp time.Time
+}
+
+// EtaTracker tracks the ETA of a job
+type EtaTracker struct {
+	samples        []sample
+	samplePosition uint8
+	maxSamples     uint8
```
StephenButtolph: `maxSamples` is just `len(samples)`. We don't need the duplicate information.
```suggestion
```

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	// Calculate the time and progress since the oldest sample
+	timeSinceOldest := sample.timestamp.Sub(oldestSample.timestamp)
+	progressSinceOldest := sample.completed - oldestSample.completed
+
+	// Check if target is already completed or exceeded
+	if sample.completed >= target {
+		zeroDuration := time.Duration(0)
+		percentComplete := float64(sample.completed) / float64(target)
+		roundedPercentComplete := math.Round(percentComplete*10000) / 100 // Return percentage (0.0 to 100.0)
+		return &zeroDuration, roundedPercentComplete
```
StephenButtolph: I feel like it is unexpected for this to return a number greater than 100... And the only reason we do this math is to report that case. Feels like we can just:
```suggestion
		return &zeroDuration, 100
```

If this _is_ needed, we should add test coverage for it, as currently removing this `if` doesn't cause any tests to fail.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	}
+
+	if timeSinceOldest == 0 {
+		return nil, 0.0
```
StephenButtolph, line 99: Should we add some tests for this case?

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	}
+	rate := float64(progressSinceOldest) / float64(timeSinceOldest)
+	if rate == 0 {
+		return nil, 0.0
```
StephenButtolph, line 103: Should we add some tests for this case?

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+	}
+
+	remainingProgress := target - sample.completed
```
StephenButtolph: `remainingProgress` reads weirdly to me. I suppose it isn't wrong... But I feel like progress is normally used to describe work that has been done, rather than work that is still remaining (we have made a lot of progress vs we have a lot of progress left to go)
```suggestion
	remainingWork := target - sample.completed
```

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+//   - At 100% progress: ETA is the raw estimate (no adjustment)
+//   - Between 0% and 100%: Adjustment decreases linearly with progress
+//
+// Example: With slowdownFactor = 2.0:
+//   - At 0% progress: ETA = raw_estimate * 2.0
+//   - At 50% progress: ETA = raw_estimate * 1.5
+//   - At 100% progress: ETA = raw_estimate * 1.0
+//
+// If maxSamples is less than 1, it will default to 5
+func NewEtaTracker(maxSamples uint8, slowdownFactor float64) *EtaTracker {
```
StephenButtolph, line 51: Are slowdown factors normal for ETAs? It feels really odd to me that we aren't just calculating the current rate of progress and extrapolating.

utils/timer/eta.go
```diff
@@ -18,8 +19,109 @@ func ProgressFromHash(b []byte) uint64 {
...
+
+	// Check if target is already completed or exceeded
+	if sample.completed >= target {
+		zeroDuration := time.Duration(0)
+		percentComplete := float64(sample.completed) / float64(target)
+		roundedPercentComplete := math.Round(percentComplete*10000) / 100 // Return percentage (0.0 to 100.0)
+		return &zeroDuration, roundedPercentComplete
+	}
+
+	if timeSinceOldest == 0 {
```
StephenButtolph: This would definitely be a misuse.. But should we bail here if `timeSinceOldest < 0`?

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -603,25 +617,36 @@ func (b *Bootstrapper) process(
...
+			now.Sub(b.lastProgressUpdateTime) >= minimumLogInterval
+
+		if shouldLog {
 			totalBlocksToFetch := b.tipHeight - b.startingHeight
-			eta := timer.EstimateETA(
-				b.startTime,
-				numFetched-b.initiallyFetched,         // Number of blocks we have fetched during this run
-				totalBlocksToFetch-b.initiallyFetched, // Number of blocks we expect to fetch during this run
+
+			eta, progressPercentage := b.etaTracker.AddSample(
```
StephenButtolph: to align with the other callsite
```suggestion
			etaPtr, progressPercentage := b.etaTracker.AddSample(
```

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -603,25 +617,36 @@ func (b *Bootstrapper) process(
 		height := blk.Height()
 		b.tipHeight = max(b.tipHeight, height)
 
-		if numPreviouslyFetched/statusUpdateFrequency != numFetched/statusUpdateFrequency {
+		// Check if it's time to log progress (both progress-based and time-based frequency)
+		now := time.Now()
+		shouldLog := numPreviouslyFetched/statusUpdateFrequency != numFetched/statusUpdateFrequency &&
```
StephenButtolph, line 622: Should we just remove the `statusUpdateFrequency` from here? It was definitely added just because we were lazy and didn't want to add another tracker (which we are doing now with `lastProgressUpdateTime`)

snow/engine/snowman/bootstrap/bootstrapper.go
```diff
@@ -407,6 +417,10 @@ func (b *Bootstrapper) startSyncing(ctx context.Context, acceptedBlockIDs []ids.
 	b.initiallyFetched = b.tree.Len()
 	b.startTime = time.Now()
 
+	// Add the first sample to the EtaTracker to establish an accurate baseline
+	// This ensures t.lowestSample is set correctly from the start
+	b.etaTracker.AddSample(b.initiallyFetched, b.tipHeight-b.startingHeight, b.startTime)
```
StephenButtolph, line 422: Just wanted to note that `startSyncing` is called multiple times, so this sample can be added multiple times.

It probably isn't a problem... but wanted to write that down.

snow/engine/snowman/bootstrap/bootstrapper.go
StephenButtolph: I wish the log rate didn't modify how the eta updates... I suppose it is just up to anyone modifying this to know that doubling the logging rate should also double the number of internally tracked samples in the eta tracker for the same eta volatility.

It might be possible to do some ewma instead to avoid this (like done [here](https://github.com/ava-labs/avalanchego/blob/master/utils/math/continuous_averager.go#L11))... but it isn't immediately obvious to me how best to do that.

snow/engine/snowman/bootstrap/bootstrapper.go
StephenButtolph: I think this comment still needs to be addressed.

utils/timer/eta.go
StephenButtolph: Should we update the last instance of this method from the bootstrapping logic?

rkuris: Can we do that in a followup issue/PR?

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: Removed, but would have been better to explain why this is here.

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: That's possible, can we do this in a followup?

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: Fair point, but that assumes that someone cares about the eta volatility more than this bug fix provides. Even with too few samples (at the minimum of 5), it should still do much better than the old code as well.

snow/engine/snowman/bootstrap/bootstrapper.go
rkuris: I think that's okay. I updated the comment to be clearer about that.

utils/timer/eta.go
rkuris: Slowdown factors are normal for merkle tries, as things are going to slow down as the trie gets deeper.

Of course, if this function is used for something that isn't expected to slow down, slowdownFactor can be set to 1.

snow/engine/snowman/bootstrap/bootstrapper.go
StephenButtolph: If bootstrapping has restarted, then this chain has finished bootstrapping, and it is just waiting for another chain to finish bootstrapping.

So the status is "done" and any additional notifications to the user is more confusing than anything else.

PR #4200, opened 2025-08-22
Aaronbuchwald/add dir src suffix

aaronbuchwald:
This is a followup to improve the naming convention of the re-execution tasks based on this comment: https://github.com/ava-labs/avalanchego/pull/4172#discussion_r2283250356

There are two related tasks `reexecute-cchain-range` and `reexecute-cchain-range-with-copied-data`.

`reexecute-cchain-range` takes in a block and current state DB and executes directly on those two databases. This overwrites the current state DB, so that it cannot be re-used for the same test and is read-only on the block DB.

For convenience, we also have `reexecute-cchain-range-with-copied-data`, which combines a task to copy the required data onto the local machine from S3 and then execute the desired range. This re-used the same variable names of `BLOCK_DIR_SRC` and `CURRENT_STATE_DIR`, which was ambiguous.

This change uses `BLOCK_DIR_SRC` and `CURRENT_STATE_DIR_SRC` to refer to `BLOCK_DIR` and `CURRENT_STATE_DIR` when they are treated as a source to copy from and uses `BLOCK_DIR` and `CURRENT_STATE_DIR` when used in re-execution.


Taskfile.yml
```diff
@@ -88,12 +88,12 @@ tasks:
   export-cchain-block-range:
     desc: Export range of C-Chain blocks from source to target directory.
     vars:
-      SOURCE_BLOCK_DIR: '{{.SOURCE_BLOCK_DIR}}'
-      TARGET_BLOCK_DIR: '{{.TARGET_BLOCK_DIR}}'
+      BLOCK_DIR: '{{.BLOCK_DIR}}'
```
aaronbuchwald: This could be `BLOCK_DIR_SRC` as well since it's also copying, but in this case it's an export as opposed to import. Probably better to update to include the suffix.

scripts/benchmark_cchain_range.sh
```diff
@@ -4,19 +4,19 @@ set -euo pipefail
 
 # This script runs the C-Chain re-execution benchmark with a single iteration.
 # It expects the following environment variables to be set:
-#   SOURCE_BLOCK_DIR: Path or S3 URL to the source block directory or zip.
+#   BLOCK_DIR: Path or S3 URL to the source block directory or zip.
```
maru-ava: (No action required) Maybe update the comment if 'source' is no longer relevant to this variable?

tests/reexecute/c/README.md
```diff
@@ -87,15 +87,15 @@ Now that we've pushed the current-state back to S3, we can run the target range
...
+task reexecute-cchain-range CURRENT_STATE_DIR=$HOME/exec-data/current-state BLOCK_DIR=$HOME/exec-data/blocks START_BLOCK=101 END_BLOCK=200
 ```
 
 Note: if you attempt to re-execute a second time on the same data set, it will fail because the current state has been updated to block 200.
 
 Provide the parameters explicitly that we have just used locally:
 
 ```bash
-task reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=$HOME/reexec-data-params SOURCE_BLOCK_DIR=s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-10k-ldb/** CURRENT_STATE_DIR=s3://avalanchego-bootstrap-testing/cchain-current-state-test/** START_BLOCK=101 END_BLOCK=10000
+task reexecute-cchain-range-with-copied-data EXECUTION_DATA_DIR=$HOME/reexec-data-params BLOCK_DIR_SRC=s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-10k-ldb/** CURRENT_STATE_DIR_SRC=s3://avalanchego-bootstrap-testing/cchain-current-state-test/** START_BLOCK=101 END_BLOCK=10000
```
maru-ava, line 98: (No action required) Why is it desirable to use `*_DIR` for `reexecute-cchain-range` but `*_DIR_SRC` for `reexecute-cchain-range-with-copied-data`?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -88,14 +88,15 @@ var (
 
 func TestMain(m *testing.M) {
 	// Source directory must be a leveldb dir with the required blocks accessible via rawdb.ReadBlock.
-	flag.StringVar(&sourceBlockDirArg, "source-block-dir", sourceBlockDirArg, "DB directory storing executable block range.")
-	// Target block directory to write blocks into when executing TestExportBlockRange.
-	flag.StringVar(&targetBlockDirArg, "target-block-dir", targetBlockDirArg, "DB directory to write blocks into when executing TestExportBlockRange.")
+	flag.StringVar(&blockDirArg, "block-dir", blockDirArg, "DB directory storing executable block range.")
```
maru-ava: (No action required) Why does it no longer make sense to explicitly name the purpose of this dir (e.g. source/src) given that there are multiple block dirs? I think having one undifferentiated 'block dir' and one differentiated 'block dir dest' begs the question of what differs between them and I'm not sure it's ideal to have to read the flag descriptions to determine why.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -88,14 +88,15 @@ var (
...
+	// Destination block directory to write blocks into when executing TestExportBlockRange.
```
maru-ava: (No action required) Could this comment be included in the flag description so that invoking this command wouldn't have to look at the source code to learn these details? Same question for the comment about current state directory.

tests/reexecute/c/README.md
aaronbuchwald: The intent here is that the former is the database used for execution and the `-src` suffix indicates that we are copying from `-src` to `-dst` rather than operating on the data itself.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: They both have their own comments saying something similar, so probably best to remove the comments in favor of using the flag description as its own documentation.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: We're re-using the `block-dir` here for two different use cases. Changing it to use a separate flag.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -109,6 +102,10 @@ func TestMain(m *testing.M) {
 	predefinedConfigOptionsStr := fmt.Sprintf("[%s]", strings.Join(predefinedConfigKeys, ", "))
 	flag.StringVar(&configNameArg, configKey, defaultConfigKey, fmt.Sprintf("Specifies the predefined config to use for the VM. Options include %s.", predefinedConfigOptionsStr))
 
+	// Flags specific to TestExportBlockRange.
+	flag.StringVar(&blockDirSrcArg, "block-dir-src", blockDirSrcArg, "Source block directory to copy from when running TestExportBlockRange.")
+	flag.StringVar(&blockDirDstArg, "block-dir-dst", blockDirDstArg, "Destination blockdirectory to write blocks into when executing TestExportBlockRange.")
```
maru-ava: nit: blockdirectory -> block directory

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -108,6 +102,10 @@ func TestMain(m *testing.M) {
...
```
aaronbuchwald:
```suggestion
	flag.StringVar(&blockDirDstArg, "block-dir-dst", blockDirDstArg, "Destination block directory to write blocks into when executing TestExportBlockRange.")
```

RodrigoVillar (approved):
LGTM, variable names are much more consistent

Issue #4078, opened 2025-07-10
Compress block data in x/blockdb to reduce storage size


DracoLi:
**Context and scope**

Block data is currently not compressed when saved on disk. This means that using `blockdb` instead of `leveldb` results in significantly larger storage requirements.

@rkuris noted that 50 million mainnet blocks take up 122 GB on leveldb (with receipts) and 221 GB in the blockstore  as plain text (without receipts).

We should consider using Snappy to compress block data in blockdb to reduce file sizes to levels comparable to those of leveldb.

StephenButtolph:
> We should consider using Snappy

We should also look into Zstd. We use Zstd for the p2p layer (and based on the metrics I've seen Zstd outperforms Snappy)

rkuris:
Probably worth testing. I pushed the code changes to the rust implementation to enable compile-time switching between implementations but won't have time to test the block migrator size today. Next week perhaps.
https://github.com/ava-labs/blockstore/pull/3/commits/72e7bcbe79f2365682c0378b4218128694d830d9

PR #4201, opened 2025-08-22
BlockDB Compression

DracoLi:
## Why this should be merged

Two main changes to blockdb
- Add config to enable blocks compression
- Remove support for `ReadHeader` and `ReadBody`

Resolves https://github.com/ava-labs/avalanchego/issues/4078

## How this works

- Add an optional config to blockdb config to compress blocks. False by default.
- Added `Compressed` to the Block Entry Header. This allows the blockdb to reindex if the index file is deleted/corrupted.
- Removed `ReadHeader` and `ReadBody` since we will not be using these in the vms.

## How this was tested

Unit tests

No

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -2,6 +2,9 @@ name: 'C-Chain Re-Execution Benchmark'
 description: 'Run C-Chain re-execution benchmark'
 
 inputs:
+  runner_name:
+    description: 'The name of the runner to use and include in the Golang Benchmark name.'
+    required: true
   config:
     description: 'The config to pass to the VM for the benchmark. See BenchmarkReexecuteRange for details.'
     default: ''
@@ -78,8 +81,7 @@ runs:
     - name: Run C-Chain Re-Execution
       uses: ./.github/actions/run-monitored-tmpnet-cmd
       with:
-        run: |
-          ./scripts/run_task.sh reexecute-cchain-range-with-copied-data \
+        run_env: |
             CONFIG=${{ inputs.config }} \
             EXECUTION_DATA_DIR=${{ env.EXECUTION_DATA_DIR }} \
             BLOCK_DIR_SRC=${{ env.BLOCK_DIR_SRC }} \
@@ -87,7 +89,10 @@ runs:
             START_BLOCK=${{ env.START_BLOCK }} \
             END_BLOCK=${{ env.END_BLOCK }} \
             LABELS=${{ env.LABELS }} \
-            BENCHMARK_OUTPUT_FILE=${{ env.BENCHMARK_OUTPUT_FILE }}
+            BENCHMARK_OUTPUT_FILE=${{ env.BENCHMARK_OUTPUT_FILE }} \
+            RUNNER_NAME=${{ inputs.runner_name }}
+        run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data
+
         prometheus_push_url: ${{ inputs.prometheus-push-url }}
         prometheus_username: ${{ inputs.prometheus-username }}
         prometheus_password: ${{ inputs.prometheus-password }}
--- a/.github/workflows/c-chain-reexecution-benchmark-container.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-container.yml
@@ -112,3 +112,4 @@ jobs:
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
           push-post-state: ${{ github.event.inputs.push-post-state }}
+          runner_name: ${{ matrix.runner }}
--- a/.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
@@ -97,3 +97,4 @@ jobs:
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
           push-post-state: ${{ github.event.inputs.push-post-state }}
+          runner_name: ${{ matrix.runner }}
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -186,6 +186,7 @@ tasks:
     vars:
       CURRENT_STATE_DIR: '{{.CURRENT_STATE_DIR}}'
       BLOCK_DIR: '{{.BLOCK_DIR}}'
+      RUNNER_NAME: '{{.RUNNER_NAME | default "dev"}}'
       CONFIG: '{{.CONFIG | default ""}}'
       START_BLOCK: '{{.START_BLOCK}}'
       END_BLOCK: '{{.END_BLOCK}}'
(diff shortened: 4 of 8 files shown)


x/blockdb/database.go
```diff
@@ -218,14 +221,23 @@ func New(config DatabaseConfig, log logging.Logger) (*Database, error) {
 		databaseLog = logging.NoLog{}
 	}
 
+	// from benchmarks, zstd.BestSpeed is about 100% faster than the default
+	// compression level while giving us ~5% better compression ratio than Snappy.
+	var err error
+	compressor, err := compression.NewZstdCompressorWithLevel(math.MaxUint32, zstd.BestSpeed)
```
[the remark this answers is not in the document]
DracoLi: i think this is fine. we have other guards against size before compressor is used.

x/blockdb/config.go
```diff
@@ -97,6 +101,12 @@ func (c DatabaseConfig) WithCheckpointInterval(interval uint64) DatabaseConfig {
 	return c
 }
 
+// WithCompressBlocks returns a copy of the config with CompressBlocks set to the given value.
```
yacovm: super nit but I think `WithCompression` is more succinct. It's safe to assume that you're compressing blocks and not the index.

x/blockdb/compression_test.go
```diff
@@ -0,0 +1,137 @@
...
+	require.NoError(t, err)
+	var blockHeader blockEntryHeader
+	require.NoError(t, blockHeader.UnmarshalBinary(blockHeaderBuf))
+
+	return blockHeader.Compressed
+}
+
+func TestCompression_Enabled(t *testing.T) {
+	config := DefaultConfig().WithCompressBlocks(true)
+	db, cleanup := newTestDatabase(t, config)
```
yacovm: the cleanup makes it think you're deleting the DB although you're just closing it. Maybe rename it to close?

DracoLi: yeah i can rename here.

x/blockdb/database.go
```diff
@@ -377,11 +395,11 @@ func (s *Database) WriteBlock(height BlockHeight, block BlockData, headerSize Bl
 	bh := blockEntryHeader{
 		Height:     height,
 		Size:       blockDataLen,
-		HeaderSize: headerSize,
-		Checksum:   calculateChecksum(block),
+		Checksum:   calculateChecksum(blockToWrite),
```
rrazvan1: is any usecase of the checksum being compared between multiple db instances?

If so, we need to take into account that the checksum for the same block could be different, if the instances use different values for **CompressBlocks** flag.

DracoLi: that won't be the case. 
The checksum is just for verifying the data of the block is correct when reindexing the data files for the same db instance.
Different databases can store different block data so we should never need to compare between them.

x/blockdb/compression_test.go
```diff
@@ -0,0 +1,137 @@
...
+		require.Equal(t, allBlocks[i], readBlock, "Block %d should be readable", height)
+	}
+	closeDB()
+
+	// Step 5: Delete index file
+	indexPath := filepath.Join(dir, indexFileName)
+	require.NoError(t, os.Remove(indexPath), "Failed to delete index file")
+
+	// Step 6: Open database (this should trigger recovery)
+	db, closeDB = newTestDatabase(t, config)
```
rrazvan1: should we also cover the situation of recovering, but also with compression being enabled?

DracoLi: sure, sounds good - i can add a test for this

x/blockdb/README.md
```diff
@@ -75,15 +76,15 @@ Index Entry (16 bytes):
...
+Block Entry Header (23 bytes):
 ┌────────────────────────────────┬─────────┐
 │ Field                          │ Size    │
 ├────────────────────────────────┼─────────┤
 │ Height                         │ 8 bytes │
 │ Size                           │ 4 bytes │
 │ Checksum                       │ 8 bytes │
-│ Header Size                    │ 4 bytes │
 │ Version                        │ 2 bytes │
+│ Compressed                     │ 1 byte  │
```
rkuris: Using a prime number as a block header size guarantees that blocks will never be aligned to a page boundary, resulting in more reads. It might actually be more efficient to add a second byte and call it options, or steal a byte from the version or byte fields.

x/blockdb/database.go
```diff
@@ -478,159 +495,48 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
...
-				zap.Uint64("height", height),
-				zap.Uint64("bodyOffset", bodyOffset),
-				zap.Uint32("bodySize", bodySize),
-				zap.Error(err),
-			)
-			return nil, fmt.Errorf("failed to read block body data from data file: %w", err)
+			return nil, fmt.Errorf("failed to decompress block data: %w", err)
 		}
-		return bodyData, nil
+		return decompressed, nil
```
rkuris: I think we need to validate that the block is the size it should be after decompression, otherwise a partial read might result in a short block being returned.

x/blockdb/README.md
DracoLi: great know. I'll change the version to 1 byte. I think it will be a long time until we need to store more than 255.

DracoLi: actually if we need to verify decompressed block size matches the original block size, we need to store the compressed size in addition to the original block size. So instead of `Compressed`, i plan on changing this to `UncompressedSize`, 4 bytes.

x/blockdb/compression_test.go
DracoLi: Added, see https://github.com/ava-labs/avalanchego/pull/4201/files#diff-5ca79f745e4b589bcb925d89f75c85ff8f883cc580801b82a3f238771a2b5fc7R75-R79

x/blockdb/README.md
DracoLi: https://github.com/ava-labs/avalanchego/pull/4201/commits/1cc6560c0329073d1e6cf11aa56520e0b747b2bd

x/blockdb/database.go
```diff
@@ -505,132 +515,30 @@ func (s *Database) ReadBlock(height BlockHeight) (BlockData, error) {
...
 	}
-
-	headerData := make([]byte, indexEntry.HeaderSize)
-
-	// loop to retry fetching the data file if it got closed between get and read.
-	// If not closed, we read the header and return.
-	for {
-		dataFile, localOffset, fileIndex, err := s.getDataFileAndOffset(indexEntry.Offset)
+	blockData := buf[int(sizeOfBlockEntryHeader):]
+	if bh.UncompressedSize != bh.Size {
```
rrazvan1: This check is not enough to be sure that the data is compressed or not.

There are specific **blockData** bytes that will end up with the same size for compressed data as the uncompressed data.

You can run this test to check:
```go
func TestCompressionCheckBug(t *testing.T) {
	tempDir := t.TempDir()

	config := DefaultConfig().WithDir(tempDir).WithCompression(true)
	db, err := New(config, nil)
	require.NoError(t, err)

	blockData := []byte{0, 1, 2, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
	err = db.WriteBlock(1, blockData)
	require.NoError(t, err)

	dbBlockData, err := db.ReadBlock(1)
	require.NoError(t, err)
	require.Equal(t, blockData, dbBlockData)
}
```

DracoLi: great point, i didn't consider this bug

DracoLi: if the compressed size is the same as the uncompress size. Does it make sense to just store the original uncompressed block data? This way, we still don't need a flag to store if a block is compressed and this check would be correct.

DracoLi: updated to always compress blocks. closing this

rrazvan1 (approved):
LGTM

PR #4206, opened 2025-08-23
Fix validator management deadlock

StephenButtolph:
## Why this should be merged

With the current code, it is possible that `Connected` is called with the context lock held:
- This means that `ctx.Lock -> validators.Lock`.

Additionally, calls such as `Len` and `Has` grab the `validators.Lock` and then `refresh` the validator set, which grabs the `ctx.Lock`.
- This means that `validators.Lock -> ctx.Lock`.

This implies a potential deadlock.

## How this works

This change removes the latter, which ensures that `ctx.Lock -> validators.Lock` is the only order the locks are grabbed.

## How this was tested

Added a unit test.

No.

### DIFF
--- a/network/p2p/validators.go
+++ b/network/p2p/validators.go
@@ -6,6 +6,7 @@ package p2p
 import (
 	"cmp"
 	"context"
+	"fmt"
 	"math"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


network/p2p/validators.go
```diff
@@ -118,11 +138,11 @@ func (v *Validators) refresh(ctx context.Context) {
 
 // Sample returns a random sample of connected validators
 func (v *Validators) Sample(ctx context.Context, limit int) []ids.NodeID {
-	v.lock.Lock()
-	defer v.lock.Unlock()
-
 	v.refresh(ctx)
 
+	v.lock.RLock()
+	defer v.lock.RUnlock()
```
StephenButtolph, line 144: This PR also uses the `RLock` more, which should improve parallelism for most calls.

PR #4208, opened 2025-08-25
feat: Add meterdb for heightindex database

DracoLi:
## Why this should be merged

Similar to `meterdb` for `database.Database`, this `meterdb` is for collecting basic metrics (size, duration, calls) for `database.HeightIndex`

## How this works

Wrapper around a `database.HeightIndex` and collects metrics on size, duration, and calls for Put, Get, Has, and Close.

## How this was tested

Unit tests

### DIFF
--- a/database/heightindexdb/meterdb/db.go
+++ b/database/heightindexdb/meterdb/db.go
@@ -38,7 +38,7 @@ type Database struct {
 	heightDB database.HeightIndex
 
 	calls    *prometheus.CounterVec
-	duration *prometheus.HistogramVec
+	duration *prometheus.GaugeVec
 	size     *prometheus.CounterVec
 }
 
@@ -57,13 +57,11 @@ func New(
 			},
 			methodLabels,
 		),
-		duration: prometheus.NewHistogramVec(
-			prometheus.HistogramOpts{
+		duration: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
 				Namespace: namespace,
 				Name:      "duration",
-				Help:      "time spent in database calls (seconds)",
-				// Buckets to cover the expected performance range of db operations (1μs-1s)
-				Buckets: []float64{.000001, .00001, .0001, .001, .005, .01, .05, .1, .5, 1},
+				Help:      "time spent in database calls (ns)",
 			},
 			methodLabels,
 		),
@@ -89,7 +87,7 @@ func (db *Database) Put(height uint64, block []byte) error {
 	duration := time.Since(start)
 
 	db.calls.With(putLabel).Inc()
-	db.duration.With(putLabel).Observe(duration.Seconds())
+	db.duration.With(putLabel).Add(float64(duration.Nanoseconds()))
 	db.size.With(putLabel).Add(float64(len(block)))
 	return err
 }
@@ -100,7 +98,7 @@ func (db *Database) Get(height uint64) ([]byte, error) {
 	duration := time.Since(start)
 
 	db.calls.With(getLabel).Inc()
-	db.duration.With(getLabel).Observe(duration.Seconds())
+	db.duration.With(getLabel).Add(float64(duration.Nanoseconds()))
 	db.size.With(getLabel).Add(float64(len(block)))
 	return block, err
 }
@@ -111,7 +109,7 @@ func (db *Database) Has(height uint64) (bool, error) {
 	duration := time.Since(start)
 
 	db.calls.With(hasLabel).Inc()
-	db.duration.With(hasLabel).Observe(duration.Seconds())
+	db.duration.With(hasLabel).Add(float64(duration.Nanoseconds()))
 	return has, err
 }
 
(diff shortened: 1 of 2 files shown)


database/heightindexdb/meterdb/db.go
```diff
@@ -0,0 +1,124 @@
...
+			methodLabels,
+		),
+		duration: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: namespace,
+				Name:      "duration",
+				Help:      "time spent in database calls (ns)",
+			},
+			methodLabels,
+		),
```
rrazvan1, line 67: Should we actually use `Histogram`?

Or if we just want a single value, we could stick to `Counter`, since I only see the `Gauge` being used with `.Add(..)`.

database/heightindexdb/meterdb/db.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (db *Database) Put(height uint64, block []byte) error {
+	start := time.Now()
+	err := db.heightDB.Put(height, block)
+	duration := time.Since(start)
+
+	db.calls.With(putLabel).Inc()
+	db.duration.With(putLabel).Add(float64(duration.Nanoseconds()))
+	db.size.With(putLabel).Add(float64(len(block)))
```
rrazvan1, line 91: do we want to store the metrics in case we get an error from the db call? (this question is applicable to all the functions)

DracoLi: yeah I think all metrics should be collected for errors as well (calls, and duration). For size, the length will be 0 on error and I think that's fine.

database/heightindexdb/meterdb/db.go
DracoLi: yeah good point. I used gauge to keep it the same type as the duration in the key-value `meterdb` but I agree that Histogram here is better.

Updated

database/heightindexdb/meterdb/db.go
```diff
@@ -0,0 +1,125 @@
...
+				Help:      "number of calls to the database",
+			},
+			methodLabels,
+		),
+		duration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: namespace,
+				Name:      "duration",
+				Help:      "time spent in database calls (seconds)",
+				Buckets:   prometheus.DefBuckets,
```
rrazvan1: I dont think `prometheus.DefBuckets` is suitable for our use-case since it is not granular enough for db calls. Most of them will fall in `.005` bucket.

```
// DefBuckets are the default Histogram buckets. The default buckets are
// tailored to broadly measure the response time (in seconds) of a network
// service. Most likely, however, you will be required to define buckets
// customized to your use case.
var DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

DracoLi: yeah this definitely needs to be updated. 
I did some local benchmarking and updated the bucket ranges to 1μs-1s. That should cover the expected performance of the db operations here.

database/heightindexdb/meterdb/db.go
```diff
@@ -0,0 +1,126 @@
...
+		duration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: namespace,
+				Name:      "duration",
+				Help:      "time spent in database calls (seconds)",
+				// Buckets to cover the expected performance range of db operations (1μs-1s)
+				Buckets: []float64{.000001, .00001, .0001, .001, .005, .01, .05, .1, .5, 1},
+			},
+			methodLabels,
+		),
```
StephenButtolph: We typically try to avoid using histograms because they significantly increase the cardinality of our metrics (aka cost a lot of money).

Is there a reason we can't use the `Gauge` pattern like we have in the other meterdb implementations?

database/heightindexdb/meterdb/db.go
```diff
@@ -0,0 +1,126 @@
...
+				Name:      "calls",
+				Help:      "number of calls to the database",
+			},
+			methodLabels,
+		),
+		duration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: namespace,
+				Name:      "duration",
+				Help:      "time spent in database calls (seconds)",
```
StephenButtolph: I think using seconds is the recommended opentelemetry/prometheus pattern... But we actually typically use nanoseconds throughout the codebase. I feel like we should probably stay uniform with the rest of our metrics and expose these in ns.

database/heightindexdb/meterdb/db.go
DracoLi: gauge works here too. i initially had gauge but thought histogram might give us some more insights. I hadn't considered the cost aspect, I think gauge is enough for what we need here so I'll revert back.

database/heightindexdb/meterdb/db.go
DracoLi: yeah, if we are reverting to gauge i'll change this back to nanoseconds to be consistent.

PR #4210, opened 2025-08-25
refactor: use generics

alarso16:

### DIFF
--- a/x/sync/manager.go
+++ b/x/sync/manager.go
@@ -125,9 +125,10 @@ func newWorkItem(localRootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[
 }
 
 type Manager[T, U any] struct {
+	proofHandler ProofHandler[T, U]
 	// Must be held when accessing [config.TargetRoot].
 	syncTargetLock sync.RWMutex
-	config         ManagerConfig[T, U]
+	config         ManagerConfig
 
 	workLock sync.Mutex
 	// The number of work items currently being processed.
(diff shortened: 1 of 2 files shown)


x/sync/manager.go
```diff
@@ -234,7 +233,7 @@ func (m *Manager) Start(ctx context.Context) error {
 // sync awaits signal on [m.unprocessedWorkCond], which indicates that there
 // is work to do or syncing completes.  If there is work, sync will dispatch a goroutine to do
 // the work.
-func (m *Manager) sync(ctx context.Context) {
+func (m *Manager[T, UnixNano]) sync(ctx context.Context) {
```
joshua-kim: UnixNano looks odd here? In either case, when the function attached to the receiver has no use for the generic type they can be omitted like `Type[_]` instead of `Type[T]`

alarso16: Stupid autocomplete... but that's great to know

x/sync/manager.go
```diff
@@ -169,8 +168,8 @@ type Manager struct {
 }
 
 // TODO remove non-config values out of this struct
-type ManagerConfig struct {
-	Parser                ProofParser
+type ManagerConfig[T, U any] struct {
+	ProofHandler          ProofHandler[T, U]
```
joshua-kim: We could partially satisfy the above TODO by making this a field/parameter into `NewManager` instead of a part of this config - which will also remove the generic signature on the config type.

PR #4211, opened 2025-08-26
refactor: use reflect.TypeFor

wmypku:
## Why this should be merged

Try to use better api reflect.TypeFor instead of reflect.TypeOf when we have known the type.

More info https://github.com/golang/go/issues/60088

### DIFF
--- a/vms/rpcchainvm/vm_test.go
+++ b/vms/rpcchainvm/vm_test.go
@@ -112,7 +112,7 @@ func TestHelperProcess(t *testing.T) {
 // interface are implemented.
 func TestVMServerInterface(t *testing.T) {
 	var wantMethods, gotMethods []string
-	pb := reflect.TypeOf((*vmpb.VMServer)(nil)).Elem()
+	pb := reflect.TypeFor[vm.VMServer]()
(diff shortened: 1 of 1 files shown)


vms/rpcchainvm/vm_test.go
```diff
@@ -112,7 +112,7 @@ func TestHelperProcess(t *testing.T) {
 // interface are implemented.
 func TestVMServerInterface(t *testing.T) {
 	var wantMethods, gotMethods []string
-	pb := reflect.TypeOf((*vmpb.VMServer)(nil)).Elem()
+	pb := reflect.TypeFor[vm.VMServer]()
```
StephenButtolph, line 115: It seems that we should also be using this below here and in the `codec/reflectcodec` package.

StephenButtolph (changes requested):

vms/rpcchainvm/vm_test.go
wmypku: I don't think there are any places that can be replaced anymore;

You can perform a global search for the keyword `(nil)).Elem`. 😄

StephenButtolph:
closing in favor of #4311.

PR #4212, opened 2025-08-26
feat: add in memory implementation of HeightIndex Database

DracoLi:
## Why this should be merged

Adds an in memory implementation of `database.HeightIndex`. The main use of this will be in tests where we don't need a on-disk height-indexed database.

## How this works

The package is in `database/heightindexdb/memdb`. It uses a map to store data by height.

## How this was tested

Unit tests

No

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -99,14 +99,14 @@ type Database interface {
 type HeightIndex interface {
 	// Put inserts the value into the database at the given height.
 	//
-	// If [value] is nil or an empty slice, then when it's retrieved
-	// it may be nil or an empty slice.
+	// If value is nil or an empty slice, then when it's retrieved it may be nil
+	// or an empty slice.
 	//
-	// [value] is safe to read and modify after calling Put.
+	// value is safe to read and modify after calling Put.
 	Put(height uint64, value []byte) error
 
 	// Get retrieves a value by its height.
-	// Returns ErrNotFound if the key is not present in the database.
+	// Returns [ErrNotFound] if the key is not present in the database.
 	//
 	// Returned []byte is safe to read and modify after calling Get.
 	Get(height uint64) ([]byte, error)
@@ -118,6 +118,6 @@ type HeightIndex interface {
 
 	// Close closes the database.
 	//
-	// Calling Close after Close returns ErrClosed.
+	// Calling Close after Close returns [ErrClosed].
 	io.Closer
 }


x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NewMemoryDatabase creates a new in-memory BlockDB
+func NewMemoryDatabase() *MemoryDatabase {
+	return &MemoryDatabase{
+		blocks: make(map[BlockHeight]BlockData),
+	}
+}
+
+// WriteBlock stores a block in memory
+func (m *MemoryDatabase) WriteBlock(height BlockHeight, block BlockData) error {
```
yacovm: blockData is a slice so why do we copy it and not just save the reference? This is for tests. Do we expect the original slice to be changed?

yacovm (approved):

x/blockdb/memory_db.go
DracoLi: i don't expect it to be changed in tests but it definitely could, which might cause confusion when debugging. 
I opted to copy it since that mirrors the block database behaviour.

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	_, exists := m.blocks[height]
+	return exists, nil
+}
+
+// Inspect returns details about the database
+func (*MemoryDatabase) Inspect() (string, error) {
+	return "", nil
+}
```
joshua-kim: 1. What is this for? Is this to satisfy an interface somewhere?
2. If we kept this - would we need to check `m.closed`?

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+	blocks map[BlockHeight]BlockData
+	closed bool
+}
+
+// NewMemoryDatabase creates a new in-memory BlockDB
+func NewMemoryDatabase() *MemoryDatabase {
+	return &MemoryDatabase{
+		blocks: make(map[BlockHeight]BlockData),
+	}
+}
```
joshua-kim: I don't think we actually need this constructor - this is wrapper around a lock/map/bool which all of which have safe zero values. The gotcha is that we'll have to lazy-initialize the map in `WriteBlock`.

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+// ReadBlock retrieves the full block data for the given height
+func (m *MemoryDatabase) ReadBlock(height BlockHeight) (BlockData, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	if m.closed {
+		return nil, ErrDatabaseClosed
+	}
+
+	block, exists := m.blocks[height]
```
joshua-kim: nit: go typically uses `ok` instead of `exists` for these checks

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+func (m *MemoryDatabase) HasBlock(height BlockHeight) (bool, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	if m.closed {
+		return false, ErrDatabaseClosed
+	}
+
+	_, exists := m.blocks[height]
+	return exists, nil
```
joshua-kim: same nit here

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Close closes the in-memory database
+func (m *MemoryDatabase) Close() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.closed {
+		return nil
+	}
```
joshua-kim: It seems like we don't need this if-check - are we adding this preemptively because we're concerned we might add more stuff to `Close` in the future?

x/blockdb/memory_db.go
DracoLi: 1. interface is here https://github.com/ava-labs/avalanchego/pull/4133
2. yeah good point. i'll add this so it aligns with the behaviour of the actual blockdb implementation

x/blockdb/memory_db.go
DracoLi: i think it should be safe to remove. I don't expect closed to interfere with anything.

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,79 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package blockdb
```
joshua-kim: Should we move this to a testing package to avoid polluting the `blockdb` namespace w/ testing types? Some suggestions are something like `blockdb/blockdbtest`, or `blockdb/memdb`

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,79 @@
...
+
+import "sync"
+
+// MemoryDatabase is an in-memory implementation of BlockDB
+type MemoryDatabase struct {
```
joshua-kim: If we moved this to its own package this could be called `Database` since there'd be no collision against the existing `Database` type

x/blockdb/memory_db.go
```diff
@@ -0,0 +1,79 @@
...
+	return ok, nil
+}
+
+// Close closes the in-memory database
+func (m *MemoryDatabase) Close() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.closed = true
+	m.blocks = nil
```
joshua-kim: random comment: I guess we don't technically have to clear out this db since all future operations are stopped and this operation is idempotent - but I don't think we care either way... so feel free to ignore this...

DracoLi: setting to nil does allow the blocks to be garbage collected in case the db reference is still around.

x/blockdb/memdb/database.go
```diff
@@ -0,0 +1,83 @@
...
+	if m.closed {
+		return false, blockdb.ErrDatabaseClosed
+	}
+
+	_, ok := m.blocks[height]
+	return ok, nil
+}
+
+// Close closes the in-memory database
+func (m *Database) Close() error {
```
joshua-kim: nit: typically receivers are the first letter of the type (default naming pattern), or a well-understood abbreviation (so `d` in this instance or `db` is reasonable).

x/blockdb/memdb/database_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// All operations should fail after close
+	err := db.WriteBlock(height, blockData)
+	require.ErrorIs(t, err, blockdb.ErrDatabaseClosed)
+
+	_, err = db.ReadBlock(height)
+	require.ErrorIs(t, err, blockdb.ErrDatabaseClosed)
+
+	_, err = db.HasBlock(height)
+	require.ErrorIs(t, err, blockdb.ErrDatabaseClosed)
```
joshua-kim: These look like 3 separate test cases to me, should we break these apart into separate tests? Keeping these coupled will mean that we get a loss of visibility if the implementation breaks, since the rest of the test won't run.

x/blockdb/memdb/database_test.go
```diff
@@ -0,0 +1,77 @@
...
+	// Read block back
+	retrievedBlock, err := db.ReadBlock(height)
+	require.NoError(t, err)
+	require.Equal(t, blockData, retrievedBlock)
+
+	// Verify non-existent block
+	nonExistentHeight := blockdb.BlockHeight(999)
+	exists, err = db.HasBlock(nonExistentHeight)
+	require.NoError(t, err)
+	require.False(t, exists)
```
joshua-kim: This similarly looks like a distinct test case

x/blockdb/memdb/database.go
DracoLi: yeah i forgot to update this after the rename. tx

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,85 @@
...
+		return database.ErrClosed
+	}
+
+	if d.data == nil {
+		d.data = make(map[uint64][]byte)
+	}
+
+	if len(data) == 0 {
+		return database.ErrNotFound
+	}
```
joshua-kim: This sentinel error is typically used to indicate that a key is not present in the db - why do we use it in this instance? It seems like we want to prevent zero-length blocks being put into us. I feel like we should either define a new error in this package for that, or we should remove this check and do it in the caller. It's not clear to me how this should behave since the `HeightIndex` interface doesn't document it.

database/heightindexdb/memdb/database_test.go
joshua-kim, line 1: I forgot this existed (and should have left this comment on your previous blockdb PR) - but for interface-testing we could just run this through this https://github.com/ava-labs/avalanchego/blob/45df9e6926b93aacd11589a7b90a76078ae956d9/database/dbtest/dbtest.go#L38

I think all the tests implemented in this file have a test case defined there so we could reuse some code here.

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,85 @@
...
+
+	dataCopy := make([]byte, len(data))
+	copy(dataCopy, data)
+	d.data[height] = dataCopy
+
+	return nil
+}
+
+// Get retrieves data at the given height
+func (d *Database) Get(height uint64) ([]byte, error) {
```
joshua-kim: Sort of random - the go convention is the first letter (`d`) by default, but in some cases you can use abbreviations so `db` is also acceptable here since it's a well known abbreviation for `Database`. (similar for `vm` for a "Virtual Machine"). Both are Go-ey though, so you can ignore this comment.

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	data, ok := d.data[height]
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+
+	dataCopy := make([]byte, len(data))
+	copy(dataCopy, data)
+	return dataCopy, nil
```
joshua-kim: If we're going to implement defensive copies, it might be a good idea to leave a godoc on the interface that these are safe to modify.

database/heightindexdb/memdb/database_test.go
joshua-kim: Nvm I'm being dumb - this is implementing `HeightIndex`, not `Database` so that's why I didn't leave the comment previously.

database/heightindexdb/memdb/database.go
DracoLi: going to remove this. the behaviour should be similar to database.Database, where we can put empty bytes.

database/heightindexdb/memdb/database_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+func TestOperationsAfterClose(t *testing.T) {
```
joshua-kim: Few comments on this test:
1. Why do we share a db instance across tests in the table? We could create a new db instance on each test to avoid issues w/ shared state (not that there are any in this test, but the principle still applies)
2. I generally feel like setup functions in test tables border on being a smell - we're basically just writing a test in the function body. I think if we want to test this we should just write a new test per function (Put/Get/Has...) on a closed db and not try to shoe-horn this into a test-table.

database/heightindexdb/memdb/database_test.go
DracoLi: `HeightIndex` has different interface than `database.Database` so we cannot reuse that package. 
We could create a new package for heightindex db tests though. maybe it make sense for me to just move these tests into `heightindexdb/dbtests` so it can be reused?

database/heightindexdb/memdb/database_test.go
```diff
@@ -0,0 +1,113 @@
...
+func TestPut(t *testing.T) {
+	db := &Database{}
+
+	height := uint64(1)
+	blockData := []byte("test block data")
+	require.NoError(t, db.Put(height, blockData))
+}
+
+func TestGet(t *testing.T) {
+	db := &Database{}
```
joshua-kim: It feels like we could have a test table similar to `Has` on a block that is/isn't in the db with an assertion on `Get`'s expected behavior.

database/heightindexdb/memdb/database_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.fn()
+			require.ErrorIs(t, err, database.ErrClosed)
+		})
+	}
+}
+
+func TestPut(t *testing.T) {
```
joshua-kim: Feels like this better tested as part of `Get`, since you can't verify `Put`'s behavior without a corresponding `Get` - should we remove this and just use the coverage from `TestGet`?

database/heightindexdb/memdb/database_test.go
```diff
@@ -0,0 +1,113 @@
...
+	require.NoError(t, db.Put(height, blockData))
+
+	// Read block back
+	retrievedBlock, err := db.Get(height)
+	require.NoError(t, err)
+	require.Equal(t, blockData, retrievedBlock)
+}
+
+func TestHas(t *testing.T) {
+	t.Run("non-existent block", func(t *testing.T) {
```
joshua-kim: We could use a test-table instead of sub-testing here

database/heightindexdb/memdb/database_test.go
```diff
@@ -0,0 +1,113 @@
...
+		db := &Database{}
+		blockData := []byte("test block data")
+		require.NoError(t, db.Put(uint64(1), blockData))
+		exists, err := db.Has(uint64(1))
+		require.NoError(t, err)
+		require.True(t, exists)
+	})
+}
+
+func TestPut_Overwrite(t *testing.T) {
```
joshua-kim: This could be refactored into the `TestPutGet` test table if we accepted puts to perform and a get to perform after.

I was thinking something along the lines of:
```
type putArgs struct {
    height uint64
    blk []byte
}

test := struct{}{
    name string
    puts []putArgs
    height uint64
    wantBlk []byte
    gotBlk []byte
}

for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) error {
        // setup ...
       for _, p := range puts {
           require.NoError(db.Put(p.height, p.blk))
       }
       // rest of test
    })
}
```

database/heightindexdb/memdb/database.go
DracoLi: i'll change it to db to align with other dbs

database/heightindexdb/memdb/database_test.go
joshua-kim: ~If we are going to have a set of interface-invariant-level tests defined, we might want to eventually refactor similarly to dbtest (not necessarily in this PR, but maybe in the following one) so we don't have to rewrite our coverage across both implementations.~

Posted this comment a little late, but looks like we're on the same page here.

database/heightindexdb/memdb/database.go
DracoLi: good point - i noticed we do mention these kind of behaviour for the other interfaces in database. I'll add some additional comments to the interface on the behaviour, including the nil value bytes.

database/heightindexdb/memdb/database_test.go
DracoLi: yeah i agree, i'll clean up these tests a bit. The tests for this is added as an after thought so a lot of items can be cleaned up.

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Close closes the in-memory database
+func (d *Database) Close() error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	d.closed = true
+	d.data = nil
+	return nil
+}
```
StephenButtolph: I think our other databases return `database.ErrClosed` if we try to close them again after closing.

database/heightindexdb/memdb/database.go
StephenButtolph: I think we should be using [`slices.Clone`](https://pkg.go.dev/slices#Clone)
```suggestion
	return slices.Clone(data), nil
```
Same in `Put`

database/heightindexdb/memdb/database.go
DracoLi: good to know, will update this and also add a note to the interface.

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
 }
 
-// HeightIndex defines the interface for storing and retrieving entries by height.
+// HeightIndex defines the interface for storing and retrieving values by height.
 type HeightIndex interface {
-	// Put inserts the entry into the store at the given height.
-	Put(height uint64, bytes []byte) error
+	// Put inserts the value into the database at the given height.
+	//
+	// A nil or empty slice for [value] is allowed and will be returned as nil
```
DracoLi: clarified the behavior of nil and empty bytes. This should align with the behavior of leveldb now

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+var Tests = map[string]func(t *testing.T, newDB func() database.HeightIndex){
+	"PutGet":              TestPutGet,
+	"Has":                 TestHas,
+	"CloseAndPut":         TestCloseAndPut,
+	"CloseAndGet":         TestCloseAndGet,
+	"CloseAndHas":         TestCloseAndHas,
+	"Close":               TestClose,
+	"ModifyValueAfterPut": TestModifyValueAfterPut,
+	"ModifyValueAfterGet": TestModifyValueAfterGet,
+}
```
joshua-kim, line 26: Could we define this as an array instead of a map? Map iteration isn't defined so our unit tests could become non-deterministic if state is shared.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+	data   []byte
+}
+
+func TestPutGet(t *testing.T, newDB func() database.HeightIndex) {
+	tests := []struct {
+		name        string
+		puts        []putArgs
+		queryHeight uint64
+		expected    []byte
+		expectedErr error
```
joshua-kim: nit: Go typically uses `want*` instead of `expected` for assertions

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+			queryHeight: 1,
+			expected:    nil,
+		},
+		{
+			name: "put and get empty data",
+			puts: []putArgs{
+				{1, []byte{}},
+			},
+			queryHeight: 1,
+			expected:    nil,
```
joshua-kim: nit: up to you if you want to leave this, but this doesn't have to be explicitly defined since nil is the zero value for a pointer. I'm just commenting this since I noticed a lot of this in this PR and i wasn't sure if it was intentional.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+			},
+			queryHeight: 1,
+			expected:    make([]byte, 1000),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			db := newDB()
+			defer db.Close()
```
joshua-kim: Looks like we're ignoring a return value here.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			// Query the specific height
+			retrievedData, err := db.Get(tt.queryHeight)
+			if tt.expectedErr != nil {
+				require.ErrorIs(t, err, tt.expectedErr)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, tt.expected, retrievedData)
+			}
```
joshua-kim: Do we need this if-else branch? Seems like `ErrorIs` should behave fine since you can assert into a `nil` error, and the expected value should be the zero valuef or `retrievedData`.
```suggestion
			require.ErrorIs(t, err, tt.expectedErr)
			require.Equal(t, tt.expected, retrievedData)
```

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+		name        string
+		puts        []putArgs
+		queryHeight uint64
+		expected    bool
+	}{
+		{
+			name:        "non-existent item",
+			puts:        []putArgs{},
+			queryHeight: 1,
+			expected:    false,
```
joshua-kim: Same nit here w.r.t zero values

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			db := newDB()
+			defer db.Close()
+
+			// Perform all puts
+			for _, write := range tt.puts {
+				require.NoError(t, db.Put(write.height, write.data))
+			}
+
+			exists, err := db.Has(tt.queryHeight)
```
joshua-kim: nit: `ok` is typically used for these existence checks

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+func TestHas(t *testing.T, newDB func() database.HeightIndex) {
+	tests := []struct {
+		name        string
+		puts        []putArgs
+		queryHeight uint64
+		expected    bool
+	}{
+		{
+			name:        "non-existent item",
+			puts:        []putArgs{},
```
joshua-kim: Same nit here w.r.t zero values

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+// Tests is a list of all database tests
+var Tests = map[string]func(t *testing.T, newDB func() database.HeightIndex){
```
joshua-kim: One call out (this is true for the existing `dbtest`, but still) is that this doesn't play with the IDE nicely. IDE doesn't seem to be able to pick this up as a test vector so you can't run a single test at a time which is annoying. I'm not sure how we can fix this while also supporting table-tests in the subtests though - so just leaving this as something I noticed.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+	modifiedData := []byte("modified data")
+
+	// Put original data
+	require.NoError(t, db.Put(1, originalData))
+
+	// Modify the original data slice
+	copy(originalData, modifiedData)
+
+	// Get the data back - should still be the original data, not the modified slice
+	retrievedData, err := db.Get(1)
```
joshua-kim: I think we could reasonably refactor this into `TestPutGet` after each put and we wouldn't need this test since we already have a overwritten key case in the test table

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,250 @@
...
+	require.NoError(t, err)
+	require.Equal(t, []byte("original data"), retrievedData)
+	require.NotEqual(t, modifiedData, retrievedData)
+}
+
+func TestModifyValueAfterGet(t *testing.T, newDB func() database.HeightIndex) {
```
joshua-kim: Similar comment w.r.t this test

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,98 @@
...
+	mu     sync.RWMutex
+	data   map[uint64][]byte
+	closed bool
+}
+
+func New() *Database {
+	return &Database{
+		data: make(map[uint64][]byte),
+	}
+}
```
joshua-kim: Do we need this constructor now?

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,98 @@
...
+
+	data, ok := db.data[height]
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+
+	// don't return empty slice if data is nil or empty
+	if data == nil {
+		return nil, nil
+	}
```
joshua-kim: I think this check is redundant - you can read from a zero-value map safely so this code is unreachable since the above if will always return first

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
+	// when retrieved via Get.
+	//
+	// Note: [value] is safe to read and modify after calling Put.
+	Put(height uint64, value []byte) error
 
-	// Get retrieves an entry by its height.
+	// Get retrieves a value by its height.
+	// Returns ErrNotFound if the key is not present in the database.
+	//
+	// Note: Get always returns a new copy of the data.
```
joshua-kim: nit: can we use a comment similar to `Put`? I don't care which we use, but both are basically saying the caller can manipulate the data after calling the api which implies a copy being created so I just want to be consistent here.

database/database.go
joshua-kim: Why do we override the value the caller provided? It seems like a confusing DX if you put `byte{}`, and get back `nil`.

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
 	Get(height uint64) ([]byte, error)
 
-	// Has checks if an entry exists at the given height.
+	// Has checks if a value exists at the given height.
+	//
+	// Return true even if the stored value is nil, or empty.
```
joshua-kim: I'm also confused here, I thought `Get` returned `nil` if you inserted empty - why does `Has` return nil or empty?

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
 	Has(height uint64) (bool, error)
 
 	// Close closes the database.
+	//
+	// Note: Calling Close after Close returns ErrClosed.
```
joshua-kim: `Note:` prefix feels redundant since this is a godoc

database/heightindexdb/memdb/database.go
```diff
@@ -0,0 +1,98 @@
...
+	closed bool
+}
+
+func New() *Database {
+	return &Database{
+		data: make(map[uint64][]byte),
+	}
+}
+
+// Put stores data in memory at the given height
```
joshua-kim: I think godocs on the implementation aren't needed now since we have them on the interface. I would only leave comments if they provide information specific to the implementation and not the public contract of the interface

database/database.go
DracoLi: yeah i added this in to align the behaviour of our leveldb package.
If you are to run `leveldb.Put(key, []byte{})` then `value, err = db.Get(key)`
The value would be nil.
```
Error:          Not equal:
                expected: []byte(nil)
                actual  : []byte{}

                Diff:
                --- Expected
                +++ Actual
                @@ -1,2 +1,3 @@
                -([]uint8) <nil>
                +([]uint8) {
                +}
```

database/heightindexdb/dbtest/dbtest.go
DracoLi: yeah, i left some of these to be explicit what the test is expecting. There are a couple that should be removed though. I'll update.

database/heightindexdb/dbtest/dbtest.go
DracoLi: yep, great point

database/heightindexdb/dbtest/dbtest.go
DracoLi: 👍 nice, much cleaner

database/heightindexdb/memdb/database.go
joshua-kim: `slices.Clone` will not return because the above if will return first. 

```
	data, ok := db.data[height]
	if !ok {
		return nil, database.ErrNotFound
	}
```

This will return since when the map is `nil`, the key will not be present (as a `nil` map contains no keys).

database/database.go
joshua-kim: Yeah I understand we're trying to be consistent w/ leveldb. My question is more if we should care about leveldb's behavior or not (and even if we did, if leveldb is making a bad api decision, should we lower our bar to meet it?).

I think this conversation is outside of the scope of this PR, but we're actively trying to migrate away from leveldb - so maintaining consistency w/ it is a requirement that I'm questioning. I wonder if pebbledb does the same thing here. I don't think we should block this PR on whatever the outcome of this conversation is though - I'm just commenting here since I learned this behavior from this PR.

database/heightindexdb/memdb/database.go
DracoLi: nope, good to remove!

database/heightindexdb/memdb/database.go
DracoLi: yep, no longer needed and we still pass the insert empty bytes and get nil test.

DracoLi: ah yeah sorry, didn't see your comment. I removed it. For context, I added originally because it thought `Clone` was returning an empty byte instead of nil when cloning nil. But this is not true.

database/database.go
DracoLi: I'm definitely okay with changing it. Personally I think aligning the response with the original value makes sense and can avoid issues during tests so I'm open to changing it here.

DracoLi: Going to update this - we can always revert it back if the team has additional thoughts on the counter argument.

database/heightindexdb/memdb/database.go
DracoLi: agreed, will remove

database/database.go
DracoLi: this is to clarify `Has` returns `true` if the value at the key is nil or empty.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+type testCase struct {
+	Name string
+	Test func(t *testing.T, newDB func() database.HeightIndex)
+}
```
joshua-kim: If we're going to not export this type out of the package, could we define these anonymously like we are in the other table tests in this file?
```
var Tests = []struct{
...
} {
...
}
```

DracoLi: yep, make sense

joshua-kim (approved):

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
 
-// HeightIndex defines the interface for storing and retrieving entries by height.
+// HeightIndex defines the interface for storing and retrieving values by height.
 type HeightIndex interface {
-	// Put inserts the entry into the store at the given height.
-	Put(height uint64, bytes []byte) error
+	// Put inserts the value into the database at the given height.
+	//
+	// A nil or empty slice for [value] is accepted and will return their respective
+	// value if retrieved via Get.
```
StephenButtolph: Is this true with the on-disk implementation? I would have expected the on-disk format of `nil` vs `[]byte{}` to be the same.

DracoLi: yep this will be the case for that implementation as well.
In https://github.com/ava-labs/avalanchego/pull/4318, I'm adding the shared tests to it so we will ensure it behaves the same.

database/database.go
StephenButtolph: Differentiating between `nil` and `[]byte{}` is an antipattern imo. I don't think this is a leveldb API issue. I think this is how almost all serializations handle `nil` and `[]byte{}`.

StephenButtolph: See: https://google.github.io/styleguide/go/decisions#nil-slices

While the user of this interface doesn't need to consider `nil`-ness, by defining the behavior on the interface, we require the implementers of the interface to abide by it... Thus introducing the complexity that we realistically wanted to avoid.

DracoLi: > Do not create APIs that force their clients to make distinctions between nil and the empty slice.

Yeah good point. I'll update this to align it with the behaviour of the existing `KeyValueWriter` interface.
```
// If [value] is nil or an empty slice, then when it's retrieved
// it may be nil or an empty slice.
```

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
+// HeightIndex defines the interface for storing and retrieving values by height.
 type HeightIndex interface {
-	// Put inserts the entry into the store at the given height.
-	Put(height uint64, bytes []byte) error
+	// Put inserts the value into the database at the given height.
+	//
+	// If [value] is nil or an empty slice, then when it's retrieved
+	// it may be nil or an empty slice.
+	//
+	// [value] is safe to read and modify after calling Put.
```
StephenButtolph: The bracket notation doesn't work for local variables (and arguments) so we shouldn't use it here.
```suggestion
	// If value is nil or an empty slice, then when it's retrieved it may be nil
	// or an empty slice.
	//
	// value is safe to read and modify after calling Put.
```

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
 
-	// Has checks if an entry exists at the given height.
+	// Has checks if a value exists at the given height.
+	//
+	// Returns true even if the stored value is nil or empty.
 	Has(height uint64) (bool, error)
 
 	// Close closes the database.
+	//
+	// Calling Close after Close returns ErrClosed.
```
StephenButtolph: Bracket notation does work for exported consts/vars
```suggestion
	// Calling Close after Close returns [ErrClosed].
```

database/database.go
```diff
@@ -95,17 +95,29 @@ type Database interface {
...
+	//
+	// If [value] is nil or an empty slice, then when it's retrieved
+	// it may be nil or an empty slice.
+	//
+	// [value] is safe to read and modify after calling Put.
+	Put(height uint64, value []byte) error
 
-	// Get retrieves an entry by its height.
+	// Get retrieves a value by its height.
+	// Returns ErrNotFound if the key is not present in the database.
```
StephenButtolph:
```suggestion
	// Returns [ErrNotFound] if the key is not present in the database.
```

PR #4213, opened 2025-08-26
Simplex ReplicationResponse.LatestRound Might Be Nil

samliok:
## Why this should be merged

A responding node can chose not to send us a latest round if we are already aware of it, therefore this field may be nil


simplex/messages.go
```diff
@@ -162,9 +162,13 @@ func newReplicationResponse(
...
-	latestQR, err := quorumRoundToP2P(latestRound)
-	if err != nil {
-		return nil, err
+	var latestQR *p2p.QuorumRound
+	if latestRound != nil {
+		qr, err := quorumRoundToP2P(latestRound)
+		if err != nil {
+			return nil, err
+		}
+		latestQR = qr
```
StephenButtolph, line 178: This seems like a bad bug, should we be adding some regression tests here? `simplexMessageToOutboundMessage` seems pretty lightly tested rn

samliok: added a test for the replication response

StephenButtolph (approved):

PR #4215, opened 2025-08-26
introduce granite stateless block type for epoch serialization

iansuvak:
## Why this should be merged
introduces new `statelessUnsignedGraniteBlock` type to support serialization of the PChainEpoch struct and switches the timestamp to miliseconds from second


vms/proposervm/block/block.go
```diff
@@ -44,10 +45,18 @@ type SignedBlock interface {
...
 type statelessUnsignedBlock struct {
+	ParentID     ids.ID `serialize:"true" json:"parentID"`
+	Timestamp    int64  `serialize:"true" json:"timestamp"`
+	PChainHeight uint64 `serialize:"true" json:"pChainHeight"`
+	Certificate  []byte `serialize:"true" json:"certificate"`
+	Block        []byte `serialize:"true" json:"block"`
+}
+
+type statelessUnsignedGraniteBlock struct {
+	PChainEpoch  PChainEpoch `serialize:"true" json:"pChainEpoch"`
```
iansuvak: Added all of the fields here directly instead of embedding `statelessUnsignedBlock` since the embedded field wouldn't serialize if it's unexported like that type is.

vms/proposervm/block/block.go
```diff
@@ -86,63 +106,169 @@ func (b *statelessBlock) Bytes() []byte {
...
+
+func (b *statelessGraniteBlock) verify(chainID ids.ID) error {
+	return verifyBlock(b, chainID)
+}
+
+// initializeBlock handles the common initialization logic for both block types
+func initializeBlock(block SignedBlock, bytes []byte) error {
+	switch b := block.(type) {
+	case *statelessBlock:
+		b.bytes = bytes
```
geoff-vball: Are we able to pull out the common logic out from the switch statement, or do we need it inside to access the private fields? If we can't extract the common functionality, it feels weird to call this function from a receiver on the specific block type just to have to switch on the type in this function.

vms/proposervm/block/build.go
```diff
@@ -21,16 +21,29 @@ func BuildUnsigned(
...
+	} else {
+		block = &statelessGraniteBlock{
+			StatelessBlock: statelessUnsignedGraniteBlock{
+				PChainEpoch:  pChainEpoch,
+				ParentID:     parentID,
+				Timestamp:    timestamp.UnixMilli(),
+				PChainHeight: pChainHeight,
+				Certificate:  nil,
+				Block:        blockBytes,
+			},
```
geoff-vball, line 47: Do we need the outer timestamp field here as well?

vms/proposervm/block/build.go
```diff
@@ -77,21 +111,49 @@ func Build(
...
+		return nil, err
+	}
+
+	// Set the signature
+	switch b := block.(type) {
+	case *statelessBlock:
+		b.Signature = signature
+	case *statelessGraniteBlock:
+		b.Signature = signature
+	}
```
geoff-vball: For these, I think it would probably make sense to add private methods to `SignedBlock`, and implement them for each block implementation, so we can just call `block.setSignature(signature)`

vms/proposervm/block/build.go
iansuvak: I didn't add it originally but presumably yes. I think it makes sense since the outer timestamp is the actual `time.Time` field while the inner is the uint64 for serialization

vms/proposervm/block/block.go
iansuvak: It would need more getters and setters to actually dedupe logic, right now I just pulled them apart and documented but it's a shame since the only difference is switch to millisecond timestamps for serialized blocks

iansuvak:
Closing since the commits have been merged into #4238

Issue #4056, opened 2025-07-07
Allow creating platformvm.Client with custom EndpointRequester


nirav-patel-anchor:
**Context and scope**
Currently, [NewClient](https://github.com/ava-labs/avalanchego/blob/507e6bbb5e7d166749fa0b5633e26fd36b753064/vms/platformvm/client.go#L155) appends `/ext/P` to the provided uri. This doesn't work when we need to use a node provider that expects API_KEY as the last part like https://some-node-provider/ext/P/API_KEY

**Discussion and alternatives**
Allow initializing Client with custom RPC EndpointRequester

```go
func NewClientWithEndpointRequester(requester rpc.EndpointRequester) Client {
	return &client{requester: requester}
}
```

**Open questions**
Questions that are still being discussed.

nirav-patel-anchor:
happy to make the PR if it makes sense

PR #4222, opened 2025-08-27
Implement GetProposedHeight API in Proposervm

geoff-vball:
## Why this should be merged
Adds an API to the Proposer VM. This will include more endpoints later in the epoching implementation.

## How this was tested
Unit tests

Is documented in `service.md`


vms/proposervm/client.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/utils/rpc"
+)
+
+var _ Client = (*client)(nil)
+
+type Client interface {
```
yacovm: why is this implemented here if it's not used here? In Go an interface should be in proximity to where it's used, not where it's implemented.

vms/proposervm/service.go
```diff
@@ -0,0 +1,30 @@
...
+type ProposerAPI struct {
+	vm *VM
+}
+
+func (p *ProposerAPI) GetProposedHeight(_ *http.Request, _ *struct{}, reply *api.GetHeightResponse) error {
+	p.vm.ctx.Log.Debug("API called",
+		zap.String("service", "proposervm"),
+		zap.String("method", "getProposedHeight"),
+	)
+	p.vm.ctx.Lock.Lock()
```
yacovm: while this works, I don't think we should do it like this. 

We have the `GetLastAcceptedHeight()` in `vm.go` which isn't used. I think we should lock there and instead call that function. That keeps the lock local. 

I think we should also ideally not have the `VM` struct here as a field but instead an interface with `GetLastAcceptedHeight() uint64`. This lets us test the service better, without needing to initialize a real proposerVM instance but just initialize a mock.

vms/proposervm/vm.go
```diff
@@ -236,6 +249,31 @@ func (vm *VM) Shutdown(ctx context.Context) error {
 	return vm.ChainVM.Shutdown(ctx)
 }
 
+func (vm *VM) GetLastAcceptedHeight() uint64 {
```
yacovm: Looks like this isn't used anywhere.

vms/proposervm/service.go
geoff-vball: Changed the locking flow as suggested. Do we get anything from mocking the service? There's already a test implementation of the proposerVM instance to be used for testing that is trivial to create.

yacovm: I assume we'll expand this struct with further methods, won't we? Such as the epoch information. 
I don't know how easy it is to properly test the additional future methods.

Consider this as a suggestion, up to you. You can leave it as is if you prefer.

yacovm (approved):

vms/proposervm/service.go
```diff
@@ -0,0 +1,28 @@
...
+	vm *VM
+}
+
+func (p *ProposerAPI) GetProposedHeight(_ *http.Request, _ *struct{}, reply *api.GetHeightResponse) error {
+	p.vm.ctx.Log.Debug("API called",
+		zap.String("service", "proposervm"),
+		zap.String("method", "getProposedHeight"),
+	)
+
+	reply.Height = avajson.Uint64(p.vm.GetLastAcceptedHeight())
```
michaelkaplan13: `GetLastAcceptedHeight` will return the last accepted height of this proposerVM chain. What we want is the P-Chain height pointer that this node would use if it were to build a block for this proposerVM chain. 

So we should use `GetMinimumHeight`, similar to how the current PlatformVM API does for this endpoint.

vms/proposervm/vm.go
```diff
@@ -236,6 +249,34 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
```
michaelkaplan13: This can be removed now.

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,38 @@
...
+	defer func() {
+		require.NoError(proVM.Shutdown(context.Background()))
+	}()
+
+	service := &ProposerAPI{vm: proVM}
+
+	reply := api.GetHeightResponse{}
+	require.NoError(service.GetProposedHeight(&http.Request{}, nil, &reply))
+
+	minHeight, err := service.vm.ctx.ValidatorState.GetMinimumHeight(context.Background())
```
samliok: can we mock `minHeight` to return something other than the zero value? I think this would be a slightly better test since `reply.Height` is by default 0.

samliok (approved):
looks good to me, give or take my one comment

vms/proposervm/client.go
```diff
@@ -0,0 +1,30 @@
...
+// NewClient returns a Client for interacting with the ProposerVM API.
+// The provided blockchainName should be the blockchainID or an alias (e.g. "P" for the P-Chain).
+func NewClient(uri string, blockchainName string) *Client {
+	return &Client{
+		Requester: rpc.NewEndpointRequester(uri + fmt.Sprintf("/ext/bc/%s/proposervm", blockchainName)),
+	}
+}
+
+func (c *Client) GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error) {
+	res := &api.GetHeightResponse{}
```
joshua-kim, line 40: 1. Should we be using jsonrpc at all? We now support connectrpc which gives us compatibility w/ h1/h2/etc. clients + proto messages + compatibility with json. I think we wanted to keep jsonrpc for legacy apis (C-Chain), but for other ones that we define I don't think we're locked into using jsonrpc. Some other benefits are that we can re-use the proto definition file as our documentation instead of maintaining markdown docs, and that we can autogenerate clients + not have to maintain corresponding tests for them.
2. The `api` package reads like borderline tech debt to me... I kind of want to reduce the usage of it where possible. We're implementing custom types like `avajson.Uint64`, adding encoding parameters in their messages... and all of the message types are shared across consumers of the package which makes it difficult for apis to evolve independently of each other since they share message definitions. I also think that this message is being used the tip height in other consuming packages - whereas this rpc is returning a likely accepted height, so it feels unusual to re-use this message even if their fields are both communicating block heights.

vms/proposervm/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/ava-labs/avalanchego/utils/metric"
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+)
+
+type Metrics interface {
+	metric.APIInterceptor
+}
```
joshua-kim: Why do we define this interface?

vms/proposervm/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
...
+
+type metrics struct {
+	metric.APIInterceptor
```
joshua-kim: Why do we embed this interface?

vms/proposervm/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package metrics
```
joshua-kim: Why do we define our own package for this? These are service-level metrics, should these live in the same package as the service?

vms/proposervm/service.go
```diff
@@ -0,0 +1,34 @@
...
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/api"
+
+	avajson "github.com/ava-labs/avalanchego/utils/json"
+)
+
+type ProposerAPI struct {
+	vm *VM
+}
```
joshua-kim: nit: naming is odd here - this is `proposervm.ProposerAPI`, but the file-name is `service` (which is idiomatic w/ rpc naming). Should this be `proposervm.service` instead? Additionally it looks like we don't need to export this type the way it's currently being used.

vms/proposervm/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type metrics struct {
+	metric.APIInterceptor
+}
+
+func New(registerer prometheus.Registerer) (Metrics, error) {
+	m := &metrics{}
+	apiRequestMetrics, err := metric.NewAPIInterceptor(registerer)
+	errs := wrappers.Errs{Err: err}
```
joshua-kim: Why do we use `wrappers.Errs` to wrap a single err?

vms/proposervm/metrics/metrics.go
geoff-vball: I didn't write this file, but I imagine it was written out as a stub for any metrics we wanted to include in the proposervm. It's copied over from `https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/metrics/metrics.go`

@michaelkaplan13 not sure if we had any particular metrics in mind we wanted to track. Otherwise I can remove this file.

vms/proposervm/client.go
geoff-vball: @joshua-kim I reimplemented the service using `connectrpc`. I'm not sure how to wire it up properly though - I copied the pattern of `NewHTTPHandler` from `xsvm`, but not sure if that's all I need to do here.

vms/proposervm/vm.go
```diff
@@ -236,6 +239,21 @@ func (vm *VM) Shutdown(ctx context.Context) error {
 	return vm.ChainVM.Shutdown(ctx)
 }
 
+func (vm *VM) NewHTTPHandler(context.Context) (http.Handler, error) {
```
joshua-kim: I think that this implementation will initialize the proposervm api, but not the inner VM's api - which is causing xsvm's CI to fail.

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,34 @@
...
+	)
+
+	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
+	defer func() {
+		require.NoError(proVM.Shutdown(context.Background()))
+	}()
+
+	service := &service{vm: proVM}
+
+	proposedHeightResponse, err := service.GetProposedHeight(context.TODO(), nil)
```
joshua-kim: This is testing package implementation details (this code is non-exported) - if we wanted to write a unit test we would probably want to do so through the exported api (`NewHTTPHandler` returns an `http.Handler`, which is testable).

vms/proposervm/service.go
```diff
@@ -0,0 +1,34 @@
...
+	pb "github.com/ava-labs/avalanchego/connectproto/pb/proposervm"
+)
+
+var _ proposervmconnect.ProposerVMHandler = (*service)(nil)
+
+type service struct {
+	vm *VM
+}
+
+func (p *service) GetProposedHeight(ctx context.Context, _ *connect.Request[pb.GetProposedHeightRequest]) (*connect.Response[pb.GetProposedHeightReply], error) {
```
joshua-kim: nit: receivers typically are the first letter of the type (so `s` in this instance)

vms/proposervm/service.go
```diff
@@ -0,0 +1,34 @@
...
+	p.vm.ctx.Log.Debug("proposervm: GetProposedHeight called")
```
joshua-kim: Our logger already emits files/line numbers, so I think this prefix of `proposervm:` is redundant
```
[08-21|16:24:02.385] INFO network/network.go:1139 shutting down the p2p networking
[08-21|16:24:02.386] INFO nat/nat.go:193 Unmapped all ports
[08-21|16:24:02.386] INFO node/node.go:1738 cleaning up plugin runtimes
[08-21|16:24:02.387] INFO node/node.go:1766 finished node shutdown
```

vms/proposervm/service.go
```diff
@@ -0,0 +1,34 @@
...
+
+	height, err := p.vm.ctx.ValidatorState.GetMinimumHeight(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get minimum height %w", err)
```
joshua-kim: nit: recommend using `could not` in-place of `couldn't` to avoid the special characters - which makes log-diving/parsing annoying since it needs to be escaped.

tests/e2e/c/api.go
```diff
@@ -0,0 +1,111 @@
...
+			ethClient    = e2e.NewEthClient(tc, nodeURI)
+			infoClient   = info.NewClient(nodeURI.URI)
+		)
+
+		upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+		require.NoError(err)
+
+		if !upgrades.IsGraniteActivated(time.Now()) {
+			ginkgo.Skip("skipping test because granite isn't active")
+		}
```
joshua-kim: Why does granite need to active for this test? Don't all VMs run w/ proposervm?

tests/e2e/c/api.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Issue a transaction to the C-Chain to advance past genesis block
+		issueTransaction(tc, ethClient, senderKey, recipientKey.EthAddress(), txAmount)
```
joshua-kim: Do we need to issue a tx for this test?

connectproto/proposervm/service.proto
```diff
@@ -0,0 +1,15 @@
...
+
+service ProposerVM {
+  rpc GetProposedHeight(GetProposedHeightRequest) returns (GetProposedHeightReply);
+}
+
+message GetProposedHeightRequest {}
+
+message GetProposedHeightReply {
+  uint64 height = 1;
+}
```
joshua-kim, line 22: We should probably add docs to this - since I think w/ these apis we'll use the .proto files as the doc source for request/response/rpcs.

tests/e2e/c/api.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/ava-labs/avalanchego/connectproto/pb/proposervm"
+	"github.com/ava-labs/avalanchego/tests"
+	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
+	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
+	"github.com/ava-labs/avalanchego/utils/units"
+
+	pb "github.com/ava-labs/avalanchego/connectproto/pb/proposervm/proposervmconnect"
+)
+
+var _ = e2e.DescribeCChain("[ProposerVM Epoch]", func() {
```
joshua-kim: Epoch seems odd here - this change looks orthogonal to epoch-ing to me

vms/proposervm/client.go
joshua-kim: I think the proposervm api is correct, but we just need to make sure that the inner api is initialized correctly - https://github.com/ava-labs/avalanchego/pull/4222#discussion_r2343049107

tests/e2e/c/api.go
```diff
@@ -0,0 +1,63 @@
...
+
+		httpClient := &http.Client{
+			Transport: &http2.Transport{
+				AllowHTTP: true,
+				DialTLSContext: func(_ context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+					// Skip TLS to use h2c
+					return net.Dial(network, addr)
+				},
+			},
+		}
```
StephenButtolph: I know that this was copied from the `xsvm` test, but I think this should be:
```suggestion
		httpClient := &http.Client{
			Transport: &http2.Transport{
				AllowHTTP: true,
				DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
					var d net.Dialer
					return d.DialContext(ctx, network, addr) // Skip TLS to use h2c
				},
			},
		}
```

StephenButtolph: Can we also make this a helper in the `connectclient` package?
```golang
func New() *http.Client {
	return &http.Client{
		Transport: &http2.Transport{
			AllowHTTP: true,
			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
				var d net.Dialer
				return d.DialContext(ctx, network, addr) // Skip TLS to use h2c
			},
		},
	}
}
```

we should then be able to just do:
```golang
		proposerClient := pb.NewProposerVMClient(
			connectclient.New(),
			nodeURI.URI,
			connect.WithInterceptors(
				connectclient.SetRouteHeaderInterceptor{Route: avalancheCChainID.String()},
			),
		)
```

Similarly in the `xsvm` test.

vms/proposervm/service.go
```diff
@@ -0,0 +1,37 @@
...
+var _ proposervmconnect.ProposerVMHandler = (*service)(nil)
+
+type service struct {
+	vm *VM
+}
+
+func (s *service) GetProposedHeight(ctx context.Context, _ *connect.Request[pb.GetProposedHeightRequest]) (*connect.Response[pb.GetProposedHeightReply], error) {
+	s.vm.ctx.Log.Debug("GetProposedHeight called")
+
+	height, err := s.vm.ctx.ValidatorState.GetMinimumHeight(ctx)
```
StephenButtolph: Shouldn't we be returning the output of `selectChildPChainHeight` based on the preferred block?

vms/proposervm/vm.go
```diff
@@ -236,6 +239,45 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+		grpcreflect.NewStaticReflector(proposervmconnect.ProposerVMName),
+	)
+	vm.ctx.Log.Info("Registering gRPC reflection handler", zap.String("pattern", reflectionPattern))
+	mux.Handle(reflectionPattern, reflectionHandler)
+
+	// Add inner VM handler as fallback if it exists
+	if innerHandler != nil {
+		vm.ctx.Log.Info("Adding inner VM handler as fallback")
+		mux.Handle("/", innerHandler)
+	}
```
StephenButtolph: This shadows reflection handling done by the inner VM's grpc server.

Prior to this change xsvm is able to expose its reflection logic:
```sh
~: grpcurl -plaintext -H "Avalanche-Api-Route:$CHAIN_ID" localhost:9650 xsvm.Ping.Ping
{}
```

With this change:
```sh
~: grpcurl -plaintext -H "Avalanche-Api-Route:$CHAIN_ID" localhost:9650 xsvm.Ping.Ping
Error invoking method "xsvm.Ping.Ping": target server does not expose service "xsvm.Ping"
```

I think we (josh and myself) had talked about adding another header to the `Avalanche-Api-Route` for the proposervm so that we could fully separate the two implementations.

connectproto/proposervm/service.proto
```diff
@@ -0,0 +1,27 @@
...
+// Request to get the current proposed block height.
+message GetProposedHeightRequest {}
+
+// Response containing the current proposed block height.
+message GetProposedHeightReply {
+  // The current proposed block height.
+  // This represents the height of the block that the ProposerVM is
+  // currently proposing to the network for acceptance.
+  uint64 height = 1;
+}
```
StephenButtolph: This reads a bit weirdly. The node isn't always proposing a block.
```suggestion
// Response containing the current proposed P-chain block height.
message GetProposedHeightReply {
  // The P-chain height that would be included in a block if it were proposed
  // right now.
  uint64 height = 1;
}
```

StephenButtolph (commented):
Are we planning on trying to expose this on api.avax.network in the short term?

If so, we should include an implementation using the jsonrpc handlers so that we can easily expose it.

connectproto/proposervm/service.proto
```diff
@@ -0,0 +1,26 @@
...
+
+option go_package = "github.com/ava-labs/avalanchego/connectproto/pb/proposervm";
+
+// ProposerVM service provides RPCs for querying ProposerVM state.
+// This service allows clients to retrieve information about the current
+// state of block proposals in the proposer consensus mechanism.
+service ProposerVM {
+  // GetProposedHeight returns the height of the currently proposed block.
+  // This method is used to query the height of the block that is currently
+  // being proposed by the ProposerVM for network consensus.
```
StephenButtolph: Sorry, should have commented on this in the prior pass as well. But I think the description of this method should be updated similarly to how the comment on the reply message was updated.

tests/e2e/c/api.go
```diff
@@ -0,0 +1,47 @@
...
+		cWallet := baseWallet.C()
+		cBuilder := cWallet.Builder()
+		cContext := cBuilder.Context()
+		avalancheCChainID := cContext.BlockchainID
+
+		proposerClient := pb.NewProposerVMClient(
+			connectclient.New(),
+			nodeURI.URI,
+			connect.WithInterceptors(
+				connectclient.SetRouteHeaderInterceptor{Route: []string{avalancheCChainID.String(), "proposervm"}},
```
StephenButtolph: nit: `proposervm` should probably be a constant like `proposervm.HTTPHeaderRoute` or something.

vms/proposervm/metrics/metrics.go
StephenButtolph, line 1: I don't think we should be adding this file. It doesn't really fit with how we register other metrics in the proposervm.

vms/proposervm/client_test.go
StephenButtolph, line 1: I feel like we should just e2e test this, I don't think the unit test really provides meaningful coverage.

vms/proposervm/vm.go
```diff
@@ -236,6 +249,82 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+			zap.String("path", r.URL.Path),
+			zap.String("header", r.Header.Get(server.HTTPHeaderRoute)),
+		)
+
+		// Check for routing header
+		route, ok := r.Header[server.HTTPHeaderRoute]
+		if !ok {
+			// No routing header - fall back to path-based routing for ProposerVM
+			// This maintains backward compatibility
+			proposerMux.ServeHTTP(w, r)
```
StephenButtolph: This seems to explicitly document some backwards compatibility concern... I'm not really sure what that could be referring to...
```suggestion
			// The top level API router should have already guaranteed that this
			// is present.
			w.WriteHeader(http.StatusBadRequest)
```
But feel free to explain a bit more if I'm missing something.

vms/proposervm/vm.go
```diff
@@ -236,6 +249,82 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+		}
+
+		for _, routeValue := range route {
+			if routeValue == "proposervm" {
+				proposerMux.ServeHTTP(w, r)
+				return
+			}
+		}
+
+		innerHandler.ServeHTTP(w, r)
```
StephenButtolph: I'm not sure what the documented backwards compatibility concern is, so maybe this refactor isn't accurate? But I'm pretty sure this is equivalent and cleaner. (It also fixes the logging of the route)
```suggestion
		route := r.Header[server.HTTPHeaderRoute]
		vm.ctx.Log.Info("ProposerVM routing request",
			zap.String("method", r.Method),
			zap.String("path", r.URL.Path),
			zap.Strings("header", route),
		)

		if len(route) < 2 || route[1] != "proposervm" {
			innerHandler.ServeHTTP(w, r)
			return
		}

		proposerMux.ServeHTTP(w, r)
```

vms/proposervm/client_test.go
ylg-avalabs: Added [e2e tests](https://github.com/ava-labs/avalanchego/pull/4222/files#diff-adde2e3e16337990ad790e06803f6420ef060fa16bbf5cc18a436fb36a69c10eR95-R104).

tests/e2e/c/api.go
```diff
@@ -0,0 +1,103 @@
...
+	ginkgo.It("should provide JSON-RPC API for ProposerVM", func() {
+		// Set up environment and trigger block production
+		avalancheCChainID, nodeURI := setupProposerVMTest()
+
+		// Test the ProposerVM JSON-RPC API
+		client := proposervm.NewClient(nodeURI.URI, avalancheCChainID.String())
+		height, err := client.GetProposedHeight(tc.DefaultContext())
+		require.NoError(err)
+		require.Positive(height, "proposervm height should be greater than 0")
+	})
```
StephenButtolph: By splitting these into two different `It` calls, are we expecting that these are run independently?

I would have expected a single `ginkgo.It` with calls to `tc.By` (one describing the setup where we issue some txs) and then two `tc.By` calls, one for the json-rpc and one for the grpc.

I think this is correct fwiw, just not exactly structured how I would have expected it.

vms/proposervm/client.go
```diff
@@ -0,0 +1,37 @@
...
+	return &client{
+		requester: rpc.NewEndpointRequester(uri + fmt.Sprintf("/ext/bc/%s/proposervm", blockchainName)),
+	}
+}
+
+func (c *client) GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error) {
+	res := &api.GetHeightResponse{}
+	err := c.requester.SendRequest(ctx, "proposervm.getProposedHeight", struct{}{}, res, options...)
+	return uint64(res.Height), err
+}
```
StephenButtolph:
```suggestion
type Client struct {
	Requester rpc.EndpointRequester
}

// NewClient returns a Client for interacting with the ProposerVM API.
// The provided blockchainName should be the blockchainID or an alias (e.g. "P" for the P-Chain).
func NewClient(uri string, blockchainName string) *Client {
	return &Client{
		Requester: rpc.NewEndpointRequester(uri + fmt.Sprintf("/ext/bc/%s/proposervm", blockchainName)),
	}
}

// GetProposedHeight returns the current height of this node's proposer VM.
func (c *Client) GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error) {
	res := &api.GetHeightResponse{}
	err := c.Requester.SendRequest(ctx, "proposervm.getProposedHeight", struct{}{}, res, options...)
	return uint64(res.Height), err
}
```

vms/proposervm/vm.go
```diff
@@ -73,6 +82,8 @@ type VM struct {
 	ctx *snow.Context
 	db  *versiondb.Database
 
+	metrics metric.APIInterceptor
```
StephenButtolph: This doesn't need to be a VM field, we can just create the interceptor in `CreateHandlers`.

api/connectclient/client.go
```diff
@@ -16,24 +20,40 @@ var _ connect.Interceptor = (*SetRouteHeaderInterceptor)(nil)
...
+		}
 		return conn
 	}
 }
 
 func (SetRouteHeaderInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return next
 }
+
+func New() *http.Client {
```
joshua-kim, line 49: This is a nice util

tests/e2e/c/api.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/ava-labs/avalanchego/api/connectclient"
+	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
+	"github.com/ava-labs/avalanchego/vms/proposervm"
+
+	pbproposervm "github.com/ava-labs/avalanchego/connectproto/pb/proposervm"
+	pb "github.com/ava-labs/avalanchego/connectproto/pb/proposervm/proposervmconnect"
+)
+
+var _ = e2e.DescribeCChain("[ProposerVM API]", ginkgo.Label("ProposerVMAPI"), func() {
```
joshua-kim: nit: what's going on with the spacing here w/ `ProposerVMAPI`? Is this supposed to be `ProposerVM API`?

tests/e2e/c/api.go
```diff
@@ -0,0 +1,92 @@
...
+			resp, err := proposerClient.GetProposedHeight(tc.DefaultContext(), &connect.Request[pbproposervm.GetProposedHeightRequest]{})
+			require.NoError(err)
+			require.Positive(resp.Msg.Height, "proposervm height should be greater than 0")
+		})
+
+		tc.By("verifying the jsonrpc service handles API calls", func() {
+			client := proposervm.NewClient(nodeURI.URI, cChainID.String())
+			height, err := client.GetProposedHeight(tc.DefaultContext())
+			require.NoError(err)
+			require.Positive(height, "proposervm height should be greater than 0")
```
joshua-kim: nit: To my knowledge we don't really use these format-string debug messages in the rest of the codebase since it's generally self-describing

vms/proposervm/client.go
```diff
@@ -0,0 +1,43 @@
...
+// Client for interacting with the json API.
+type Client struct {
+	Requester rpc.EndpointRequester
+}
+
+// NewClient returns a Client for interacting with the json API.
+//
+// The provided chain should be the chainID or an alias. Such as "P" for the
+// P-Chain.
+func NewClient(uri string, chain string) *Client {
```
joshua-kim: Could we name this `JSONRPCClient` or something to disambiguate against the connect api?

vms/proposervm/client_jsonrpc.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/rpc"
+)
+
+// Client for interacting with the json API.
+type Client struct {
+	Requester rpc.EndpointRequester
```
joshua-kim, line 17: Does this need to be exported?

vms/proposervm/client.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/rpc"
+)
+
+// Client for interacting with the json API.
```
joshua-kim: nit: `jsonrpc` instead of `json`

vms/proposervm/client.go
```diff
@@ -0,0 +1,43 @@
...
+type Client struct {
+	Requester rpc.EndpointRequester
+}
+
+// NewClient returns a Client for interacting with the json API.
```
joshua-kim: same nit here

vms/proposervm/service.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/ava-labs/avalanchego/api/server"
+	"github.com/ava-labs/avalanchego/connectproto/pb/proposervm/proposervmconnect"
+
+	pb "github.com/ava-labs/avalanchego/connectproto/pb/proposervm"
+	avajson "github.com/ava-labs/avalanchego/utils/json"
+)
+
+var _ proposervmconnect.ProposerVMHandler = (*grpcService)(nil)
+
+type grpcService struct {
```
joshua-kim: `grpcService` is a bit of misnomer since connect is a superset of grpc, I think this should be something ilke `connectService` instead?

vms/proposervm/service.go
```diff
@@ -0,0 +1,86 @@
...
+	vm *VM
+}
+
+func (s *grpcService) GetProposedHeight(ctx context.Context, r *connect.Request[pb.GetProposedHeightRequest]) (*connect.Response[pb.GetProposedHeightReply], error) {
```
joshua-kim: Whatever the name for this ends up being, we should rename the receiver to the first letter of the type

vms/proposervm/service.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Error("failed to get child p-chain height", zap.Error(err))
+		return nil, fmt.Errorf("failed to get child p-chain height: %w", err)
+	}
+
+	return connect.NewResponse(&pb.GetProposedHeightReply{
+		Height: height,
+	}), nil
+}
+
+type jsonService struct {
```
joshua-kim: nit: `jsonRPC` instead of `json`

vms/proposervm/service.go
```diff
@@ -0,0 +1,86 @@
...
+	vm *VM
+}
+
+func (s *jsonService) GetProposedHeight(r *http.Request, _ *struct{}, reply *api.GetHeightResponse) error {
```
joshua-kim: nit: same receiver nit here

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"connectrpc.com/connect"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/api"
+	"github.com/ava-labs/avalanchego/api/connectclient"
+	"github.com/ava-labs/avalanchego/connectproto/pb/proposervm"
+	"github.com/ava-labs/avalanchego/connectproto/pb/proposervm/proposervmconnect"
+)
+
+func TestGRPCService_GetProposedHeight(t *testing.T) {
```
joshua-kim: same naming nit here

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,93 @@
...
+	resp, err := client.GetProposedHeight(context.Background(), req)
+	require.NoError(err)
+	require.NotNil(resp)
+	require.NotNil(resp.Msg)
+
+	// Verify the response matches our expected P-Chain height
+	require.Equal(uint64(pChainHeight), resp.Msg.GetHeight())
+}
+
+func TestJSONService_GetProposedHeight(t *testing.T) {
```
joshua-kim: same nit here

vms/proposervm/vm.go
```diff
@@ -35,6 +42,9 @@ import (
 )
 
 const (
+	HTTPPathEndpoint = "/proposervm"
+	HTTPHeaderRoute  = "proposervm"
```
joshua-kim: Do these need to be exported?

vms/proposervm/vm.go
```diff
@@ -236,6 +246,77 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+		return nil, err
+	}
+
+	// Create ProposerVM multiplexer for ProposerVM-specific routes
+	proposerMux := http.NewServeMux()
+
+	// Add ProposerVM specific handlers to proposerMux
+	service := &grpcService{vm: vm}
+	proposerVMPath, proposerVMHandler := proposervmconnect.NewProposerVMHandler(service)
+	vm.ctx.Log.Info("Registering ProposerVM Connect handler", zap.String("path", proposerVMPath))
```
joshua-kim: nit: typically our logs use lowercasing + I feel like this is somewhat verbose for info. Either way, I feel like we could even do without these logs + it's inconsistent that we would log here but not in `CreateHandlers`.

vms/proposervm/vm.go
```diff
@@ -236,6 +246,77 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+	proposerMux.Handle(proposerVMPath, proposerVMHandler)
+
+	// Add gRPC reflection for ProposerVM
+	reflectionPattern, reflectionHandler := grpcreflect.NewHandlerV1(
+		grpcreflect.NewStaticReflector(proposervmconnect.ProposerVMName),
+	)
+	vm.ctx.Log.Info("Registering ProposerVM gRPC reflection handler", zap.String("pattern", reflectionPattern))
```
joshua-kim: Feel like we could also just get rid of this log as well

vms/proposervm/vm.go
```diff
@@ -236,6 +246,77 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+	proposerMux.Handle(reflectionPattern, reflectionHandler)
+
+	// Create header-based router
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		route := r.Header[server.HTTPHeaderRoute]
+		vm.ctx.Log.Info("ProposerVM routing request",
+			zap.String("method", r.Method),
+			zap.String("path", r.URL.Path),
+			zap.Strings("header", route),
+		)
```
joshua-kim: This will be a really noisy log - we're already logging in the api call as well so I would also remove this log, or at the very least move it to a lower log level.

vms/proposervm/vm.go
```diff
@@ -236,6 +246,77 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+		if len(route) < 2 || route[1] != HTTPHeaderRoute {
+			if innerHandler != nil {
+				innerHandler.ServeHTTP(w, r)
+			} else {
+				w.WriteHeader(http.StatusNotFound)
+			}
+			return
```
joshua-kim: nit: can early return here for readability
```suggestion
			if innerHandler == nil {
			    w.WriteHeader(http.StatusNotFound)
			    return
			}
			
			innerHandler.ServeHTTP(w, r)
			return
```

tests/e2e/c/api.go
```diff
@@ -0,0 +1,92 @@
...
+		cBuilder := cWallet.Builder()
+		cContext := cBuilder.Context()
+		cChainID := cContext.BlockchainID
+
+		tc.By("verifying the grpc service handles API calls", func() {
+			proposerClient := pb.NewProposerVMClient(
+				connectclient.New(),
+				nodeURI.URI,
+				connect.WithInterceptors(
+					connectclient.SetRouteHeaderInterceptor{Route: []string{cChainID.String(), proposervm.HTTPHeaderRoute}},
```
joshua-kim, line 77: Is there a reason we order the routing this way? The deeper in the VM wrapping the api is, the earlier it appears in the route header (e.g in this example it's `[C-Chain Route, ProposerVM Route]`. Would it make sense for us to use the reverse order so it's `[ProposerVM Route, C-ChainRoute]`)?

vms/proposervm/service_test.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+		&struct{}{},
+		&reply,
+	))
+	require.Equal(
+		api.GetHeightResponse{
+			Height: pChainHeight,
+		},
+		reply,
+	)
```
joshua-kim: nit: whitespacing is a bit hard to read here

vms/proposervm/vm.go
```diff
@@ -236,6 +246,77 @@ func (vm *VM) Shutdown(ctx context.Context) error {
...
+
+	// Create header-based router
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		route := r.Header[server.HTTPHeaderRoute]
+		vm.ctx.Log.Info("ProposerVM routing request",
+			zap.String("method", r.Method),
+			zap.String("path", r.URL.Path),
+			zap.Strings("header", route),
+		)
+		if len(route) < 2 || route[1] != HTTPHeaderRoute {
```
joshua-kim: Ah - I think I answered my earlier question in this PR (https://github.com/ava-labs/avalanchego/pull/4222#discussion_r2380747322)  and it seems that the reason that we chose the ordering of appending the proposervm header was so that the inner vm could assume it's route was at the zero'th index to avoid a behavior change? 

One idea would be if we kept the ordering I suggested earlier would be to pop the 0th index each time we forward... but maybe a better idea than popping the route (which destroys information from the requester) would to mutate an internal header that we create when it enters our stack which would be an index in the route list that we're going through and each time we go through the route we would increment it for the next layer.

vms/proposervm/client_jsonrpc.go
geoff-vball: I think @StephenButtolph had commented on wanting to export this but I forget why

StephenButtolph: It is exported because people have requested in the past to be able to provide a custom requester (provide API tokens and the like)

vms/proposervm/vm.go
StephenButtolph: `HTTPHeaderRoute` does. I unexported `HTTPPathEndpoint`.

tests/e2e/c/api.go
StephenButtolph: The route is in the order the server inspects them.

1. First the avalanchego server routes to the requested chain.
2. Then the chain-specific proposervm checks to see if it was routed to

tests/e2e/c/api.go
StephenButtolph: I changed it to proposervm to align more with the `c` and `x` label

vms/proposervm/service.go
StephenButtolph: replaced `grpc` with `connectrpc`

vms/proposervm/service.go
StephenButtolph: I named it `jsonrpc` because I think thats the full name.

vms/proposervm/vm.go
StephenButtolph: made it debug and removed the redundant `ProposerVM` prefix

vms/proposervm/service_test.go
StephenButtolph: I kinda liked it, but 🤷 I removed the whitespace

vms/proposervm/vm.go
StephenButtolph: responded: https://github.com/ava-labs/avalanchego/pull/4222#discussion_r2382370801

StephenButtolph: To clarify, the innervm does not use the routing header at all.

vms/proposervm/client_jsonrpc.go
StephenButtolph: ref: #4056

PR #4223, opened 2025-08-27
Add timeout parameter to C-Chain re-execution jobs

aaronbuchwald:
This PR adds a `timeout-minutes` parameter to the C-Chain re-execution jobs. The default value is 6hrs and some jobs may need to run for longer than this, so this PR opts to make it a parameter rather than hardcode it to a value that is often much too large.


tests/reexecute/c/README.md
```diff
@@ -169,7 +170,8 @@ Copy your desired parameters as JSON into a file or write it out on the command
     "start-block": "101",
     "end-block": "200",
     "source-block-dir": "s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-10k-ldb/**",
-    "current-state-dir": "s3://avalanchego-bootstrap-testing/cchain-current-state-firewood-100/**"
+    "current-state-dir": "s3://avalanchego-bootstrap-testing/cchain-current-state-firewood-100/**",
+    "timeout-minutes": "5"
```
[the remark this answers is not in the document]
aaronbuchwald: This is only partially correct. These parameters are passed in via `gh workflow run`, which only accepts string arguments and afaict handles conversion to the number types correctly as for `start-block` and `end-block`. In the example above for updating the JSON defining the matrix we use for PRs/scheduled runs, we use a number as intended. There is also a comment in this section explaining why we use strings instead of numbers for number fields.

alarso16 (approved):

.github/workflows/c-chain-reexecution-benchmark-gh-native.json
```diff
@@ -7,23 +7,35 @@
...
+                "timeout-minutes": 30
+            },
+            {
+                "runner": "blacksmith-4vcpu-ubuntu-2404",
+                "config": "firewood",
+                "start-block": 101,
+                "end-block": 250000,
+                "block-dir-src": "s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**",
+                "current-state-dir-src": "s3://avalanchego-bootstrap-testing/cchain-current-state-firewood-100/**",
+                "timeout-minutes": 30
```
RodrigoVillar: q: what's the rationale for this additional Firewood workflow?

aaronbuchwald: Good catch, bad merge.

aaronbuchwald: Removed here https://github.com/ava-labs/avalanchego/pull/4223/commits/9672109ae7d6b3c388188b05659b24222add2bb0

PR #4224, opened 2025-08-27
Push benchmark re-execute results on master workflow dispatch

aaronbuchwald:
This PR updates the params passed in to C-Chain re-execution benchmarks to push benchmark results to `gh-pages` via GitHub Action Benchmark when triggered by both scheduled jobs on the master branch and workflow dispatch events on the master branch.

Note: this does give developers the ability to trigger arbitrary runs on the master branch that we will store the results of. This is alright because they will be properly named and separated based on the parameters ie. config and execution range.

What a developer could do is push benchmark results from an arbitrary range that we may not want to persist like [37, 63].

Adding the workflow dispatch has the following benefits that justify this increased chance of pushing un-interesting results:
- easier to debug archiving benchmark resutls
- enables triggering archived test runs before deciding what triggers to add on a persistent basis

### DIFF
--- a/.github/workflows/c-chain-reexecution-benchmark-container.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-container.yml
@@ -100,7 +100,7 @@ jobs:
           current-state-dir: ${{ matrix.current-state-dir }}
           prometheus-username: ${{ secrets.PROMETHEUS_ID || '' }}
           prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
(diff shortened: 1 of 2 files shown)


aaronbuchwald:
CI failure is due to https://github.com/ava-labs/avalanchego/issues/4598. Will re-run and disable this job if it continues to be flaky as it seems to be non-deterministic ie. there's no related changes in this PR and it passed on master and has passed on most runs with no relevant changes.

qusuyan (approved):

Issue #4232, opened 2025-09-02
Load Testing Firewood


RodrigoVillar:
From discussing how Firewood can be integrated into the load tests with the Firewood team, the following should be implemented:

- [x] Add a flag to specify the type of database used in the load test (i.e. LevelDB or LevelDB + Firewood)
- [x] Add the ability to specify different types of load profiles (standard load vs 2x load)
- [ ] Add the ability to specify a pre-populated initial state

PR #4235, opened 2025-09-02
feat(load): add firewood flag

RodrigoVillar:
## Why this should be merged

As mentioned in #4232, we should have the ability to run the load test with Firewood enabled.

## How this works

Adds the `firewood` flag to the load test, which can be called as follows: `task test-load -- --firewood`. When specified, the `firewood` flag overrides the default C-Chain config to use Firewood in addition to LevelDB. 

## How this was tested

CI + ran load test locally with Firewood enabled

No

### DIFF
--- a/tests/load/main/main.go
+++ b/tests/load/main/main.go
@@ -37,7 +37,15 @@ const (
 var (
 	flagVars *e2e.FlagVars
 
-	loadTimeout time.Duration
+	loadTimeoutArg     time.Duration
+	firewoodEnabledArg bool
+
+	firewoodConfig = tmpnet.ConfigMap{
+		"state-scheme":       "firewood",
+		"snapshot-cache":     0,
+		"pruning-enabled":    true,
+		"state-sync-enabled": false,
+	}
 )
 
 func init() {
@@ -46,10 +54,16 @@ func init() {
 	)
 
 	flag.DurationVar(
-		&loadTimeout,
+		&loadTimeoutArg,
 		"load-timeout",
 		0,
 		"the duration that the load test should run for",
+	)
+	flag.BoolVar(
+		&firewoodEnabledArg,
+		"firewood",
+		false,
+		"whether to use Firewood in Coreth",
 	)
 
 	flag.Parse()
@@ -69,9 +83,15 @@ func main() {
 
 	keys, err := tmpnet.NewPrivateKeys(numNodes)
 	require.NoError(err)
+
+	primaryChainConfigs := tmpnet.DefaultChainConfigs()
+	if firewoodEnabledArg {
+		primaryChainConfigs[blockchainID] = firewoodConfig
+	}
 	network := &tmpnet.Network{
-		Nodes:         nodes,
-		PreFundedKeys: keys,
+		Nodes:               nodes,
+		PreFundedKeys:       keys,
+		PrimaryChainConfigs: primaryChainConfigs,
 	}
 
 	e2e.NewTestEnvironment(tc, flagVars, network)
@@ -136,7 +156,7 @@ func main() {
 	)
 	require.NoError(err)
 
-	generator.Run(ctx, log, loadTimeout, testTimeout)
+	generator.Run(ctx, log, loadTimeoutArg, testTimeout)
 }
 
 // newTokenContract deploys an instance of an ERC20 token and distributes the


tests/load/main/main.go
```diff
@@ -37,7 +37,15 @@ const (
 var (
 	flagVars *e2e.FlagVars
 
-	loadTimeout time.Duration
+	loadTimeoutArg     time.Duration
```
RodrigoVillar, line 41: Renamed for consistency

tests/load/main/main.go
```diff
@@ -37,7 +37,15 @@ const (
...
+	firewoodEnabledArg bool
+
+	firewoodConfig = tmpnet.ConfigMap{
+		"state-scheme":       "firewood",
+		"snapshot-cache":     0,
+		"pruning-enabled":    true,
+		"state-sync-enabled": false,
+	}
```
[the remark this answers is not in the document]
RodrigoVillar: Done: 9d8e2c7

alarso16 (approved):
Looks correct to me

tests/load/main/main.go
```diff
@@ -69,9 +87,15 @@ func main() {
 
 	keys, err := tmpnet.NewPrivateKeys(numNodes)
 	require.NoError(err)
+
+	primaryChainConfigs := tmpnet.DefaultChainConfigs()
+	if firewoodEnabledArg {
+		primaryChainConfigs[blockchainID] = firewoodConfig
```
joshua-kim: NB: this is going to override the default flags - is this intended?

tests/load/main/README.md
```diff
@@ -114,6 +114,7 @@ updated by the generator/wallets. The server is configured to be targeted by
 ### Load Test Flags
 
 - `--load-timeout`: Maximum duration to run the load test (default: unlimited)
+- `--firewood`: Whether to use Firewood in Coreth (default: false)
```
joshua-kim, line 117: Would it make sense if we had a flag like `--state-scheme=firewood/hash/path` instead?

tests/load/main/main.go
RodrigoVillar: From previous experiences using Firewood in the load tests, I've never encountered an issue as a result of not using the default flags. However, if a user is solely enabling Firewood, we should aim to keep all other test variables constant - including the default values.

Modified the Firewood config to preserve the default values here: 53a9e54

tests/load/main/README.md
RodrigoVillar: In Coreth, only Firewood and HashDB are supported [[ref](https://github.com/ava-labs/coreth/blob/651f189c826549562b9abe0f55cc99ba87c5e22d/plugin/evm/vm.go#L412)]. Since Firewood is the only non-default state scheme, I thought it made sense to give it a boolean flag since there's only two state schemes.

joshua-kim: Ah - didn't know we didn't support path.

tests/load/main/main.go
```diff
@@ -191,3 +205,26 @@ func newTokenContract(
...
+		primaryChainConfigs[blockchainID] = make(tmpnet.ConfigMap)
+	}
+
+	// firewoodConfig represents the minimum configuration required to enable
+	// Firewood in Coreth.
+	//
+	// Ref: https://github.com/ava-labs/coreth/issues/1180
+	firewoodConfig := tmpnet.ConfigMap{
+		"state-scheme":       "firewood",
+		"snapshot-cache":     0,
```
joshua-kim, line 223: I think this + state-sync-enabled don't need to specified explicitly since these are the default/zero-values that the config uses

RodrigoVillar: During VM initialization, `firewoodConfig` is unmarshaled into the following default config: https://github.com/ava-labs/coreth/blob/651f189c826549562b9abe0f55cc99ba87c5e22d/plugin/evm/config/default_config.go#L16-L99. As a result, not explicitly setting `snapshot-cache` and `state-sync-enabled` to `0` and `false` will result in a failure ([[ref](https://github.com/ava-labs/coreth/blob/651f189c826549562b9abe0f55cc99ba87c5e22d/plugin/evm/vm.go#L402)], [[ref](https://github.com/ava-labs/coreth/blob/651f189c826549562b9abe0f55cc99ba87c5e22d/plugin/evm/vm.go#L408)])

maru-ava:
I traced the e2e-kube failure to https://github.com/ava-labs/avalanchego/issues/4237. The job isn't currently blocking so I see no reason to block merge.

PR #4238, opened 2025-09-03
Epoching ACP181 implementation

geoff-vball:
## Why this should be merged
Implements ProposerVM epochs as described in https://github.com/avalanche-foundation/ACPs/pull/181

## How this works
Adds a new serializable block type 

## How this was tested
E2E and unit tests

Currently all `proposervm` unit tests have their `graniteTime` set to equal their `durangoTime`. We should go through and find any tests we would like to explicitly test for granite. 

Yes


tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+		nodeURI := env.GetRandomNodeURI()
+		ethClient := e2e.NewEthClient(tc, nodeURI)
+
+		infoClient := info.NewClient(nodeURI.URI)
+		upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+		require.NoError(err)
+
+		// Don't run test if granite isn't active.
+		if !upgrades.IsGraniteActivated(time.Now()) {
+			return
```
maru-ava: Please skip with an appropriate message (maybe obviating the need for a comment) rather than silently passing.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		// Genesis does not contain a proposervm block
```
maru-ava: (No action required) Maybe also document how this could happen in an e2e test?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+			require.Greater(
+				advancedEpoch.Number,
+				initialEpoch.Number,
+				"expected epoch number to advance, but it did not",
+			)
+		})
+	})
+})
+
+func blockNumber(tc *e2e.GinkgoTestContext, ethClient *ethclient.Client) uint64 {
```
maru-ava: Maybe inline this?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+	blockNumber, err := ethClient.BlockNumber(tc.DefaultContext())
+	require.NoError(tc, err)
+	return blockNumber
+}
+
+func issueTransaction(
+	tc *e2e.GinkgoTestContext,
```
maru-ava: Please prefer the tests.TestContext interface.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+var _ = e2e.DescribeCChain("[ProposerVM Epoch]", func() {
+	tc := e2e.NewTestContext()
+	require := require.New(tc)
+
+	const txAmount = 10 * units.Avax // Arbitrary amount to send and transfer
+
+	ginkgo.It("should advance the proposervm epoch according to the upgrade config epoch duration", func() {
+		env := e2e.GetEnv(tc)
```
maru-ava: (No action required) Maybe use a var block for these initializations?

tests/e2e/c/proposervm_epoch.go
geoff-vball: `ethClient.BlockNumber()` returns two values. I don't know of a non-icky way to loop over that

tests/e2e/c/proposervm_epoch.go
geoff-vball: Oh, yeah that formatting is way more readable :)

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+		upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+		require.NoError(err)
+
+		// Don't run test if granite isn't active.
+		if !upgrades.IsGraniteActivated(time.Now()) {
+			return
+		}
+
+		// Genesis does not contain a proposervm block
+		for blockNumber(tc, ethClient) == 0 {
```
yacovm: This has the potential to loop endlessly, so perhaps we should use an upper limit to how much we can loop?

api/common_args_responses.go
```diff
@@ -58,6 +58,12 @@ type GetHeightResponse struct {
 	Height avajson.Uint64 `json:"height"`
 }
 
+type GetEpochResponse struct {
+	Number       avajson.Uint64 `json:"number"`
+	StartTime    avajson.Uint64 `json:"startTime"`
+	PChainHeight avajson.Uint64 `json:"pChainHeight"`
+}
```
michaelkaplan13: Should we include the epoch duration here as well, so that callers can determine how much time is left in the given epoch?

vms/proposervm/service.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return fmt.Errorf("couldn't get latest block %s: %w", lastAccepted.String(), err)
+	}
+
+	blk, ok := latestBlock.(block.SignedBlock)
+	if !ok {
+		return errors.New("latest block is not a signed block")
+	}
+
+	epoch := blk.PChainEpoch()
```
michaelkaplan13: If the latest block was a sealing block, what would make the most sense to return in this RPC response?

On one hand, a sealing block is well defined to be part of the epoch that it seals, and the next epoch doesn't begin until the next block is produced, so we don't yet know the next epoch's start time. 

On the other hand, callers of this RPC method (i.e. relayers) will really just want to know what P-Chain height they should use to construct aggregate signatures, and the information currently provided to them does not provide it.

Given the above, I think my suggestion would be to include the response:
- the epoch duration
- the "current" P-Chain height pointer of the latest block
- an `isSealed` boolean denoting whether or not the current epoch is sealed

This provides the caller with all of the information they need to determine:
1. How much time is guaranteed to be remaining in the current epoch
2. The information they need to successfully aggregate a signature for the next epoch (even if it hasn't been started yet) if the current epoch has been sealed.

@yacovm / @StephenButtolph also curious your thoughts on this...including all of the above potentially feels like overkill to me, but it would be useful information for the relayer IMO so including it all in a single RPC would seem to make sense.

tests/e2e/c/proposervm_epoch.go
maru-ava: Logging is an improvement, but I would prefer `ginkgo.Skip` so that it's reflected not just in the test log but also in the test and test suite status.

tests/e2e/c/proposervm_epoch.go
maru-ava: A potential alternative that avoids a separate function and addresses @yacovm's looping concern as a side-effect by ensuring a timeout:

```golang
tc.Eventually(func () bool {
    bn, err := ethClient.BlockNumber(tc.DefaultContext())
    require.NoError(err)
    return bn != 0
},
    e2e.DefaultTimeout,
    e2e.DefaultPollingInterval,
    "failed to see a non-zero block before timeout",
}
```

vms/proposervm/service.go
yacovm: We can just act as if it's just any other block of the epoch. Why should the sealing block have any special treatment? It's an internal mechanism that helps us set the boundary between epochs, it shouldn't have any user facing role. 

Suppose we have 100 blocks in the epoch - $b_i, b_{i+1}, .... b_{i+100}$, and $b_{100}$ is a sealing block. If you have special treatment for the sealing block, can't you argue that $b_{99}$ should also have special treatment, as it's only one block before the sealing block?

tests/e2e/c/proposervm_epoch.go
yacovm: Yeah either that or [gomega's eventually](https://pkg.go.dev/github.com/onsi/gomega#Eventually) is the way to go.

vms/proposervm/service.go
iansuvak: I think that @michaelkaplan13 's point here is very valid since for aggregating a signature without returning extra information we would expect a set of failures around epoch boundaries if there is sustained traffic. 

@yacovm 

> We can just act as if it's just any other block of the epoch. Why should the sealing block have any special treatment? It's an internal mechanism that helps us set the boundary between epochs, it shouldn't have any user facing role.

I don't think that the block is getting a special treatment but extra information about the epoch itself is being returned from `GetEpoch` endpoint which seems appropriate, no?

yacovm: > I think that @michaelkaplan13 's point here is very valid since for aggregating a signature without returning extra information we would expect a set of failures around epoch boundaries if there is sustained traffic.
> 
> @yacovm
> 
> > We can just act as if it's just any other block of the epoch. Why should the sealing block have any special treatment? It's an internal mechanism that helps us set the boundary between epochs, it shouldn't have any user facing role.
> 
> I don't think that the block is getting a special treatment but extra information about the epoch itself is being returned from `GetEpoch` endpoint which seems appropriate, no?


Perhaps special treatment isn't the best terminology to use, but it seemed to me like the isSealed boolean is special treatment because it can only manifest when a sealing block is the last block.

I guess to properly answer the concerns raised here, it is required to understand how the API of ACP-181 is to be used.

I don't know enough about ICM to understand what is the typical way of using ACP-181, but I would assume that if you query the API and assemble the aggregated signature and send it on its way, you can always just query the API once more to see if an epoch change had occurred and your aggregated signature was computed on stale data or not and based on that, retry.

geoff-vball: I think the important thing here is that when a block seals an epoch, we know the next block will use the pchain height of that block, and not the pchain height of that epoch. 

In the relayer's usecase, what we really want to know is "What pchain height will be used to verify the next block", and I think we can only get that with additional information that is not currently included. It is "special treatment" in a way, because this is a special case. A block sealing an epoch is the only time that the next block isn't verified with the current epoch's pchain height.

michaelkaplan13: This API is exposing information about the current epoch, not any block specifically. It's really just a question of what information should be exposed.

Whether or not the current epoch is sealed is relevant to callers generally (and the relayer specifically), because if it is not sealed, the next block is guaranteed to also be in this current epoch, and if it is sealed, the next block is guaranteed to be in the next epoch.

yacovm: In that case, shouldn't we just return the new epoch when the last block is the sealing block? Since any ICM message will get into a block that is in the next epoch as this epoch is sealed.

api/common_args_responses.go
iansuvak: Given that epoch is set on the `upgrade` level which should be the same on the network level, the `StartTime` should be sufficient.

vms/proposervm/service.go
michaelkaplan13: Yes, but the issue there is that the new epoch may not have started yet, since it only starts when the next block is produced/accepted. Specifically, the `startTime` of the next epoch still isn't well defined.

yacovm: Sure, but any transaction validated on-chain will use the new epoch, not the old one. Perhaps that means that we shouldn't return the start time of the epoch in the API in the first place?

michaelkaplan13: There's of course any number of ways we could design/structure this to have the relayer get the information it needs, but IMO it's cleanest and most general to have this API only return information about the current epoch.

It would be more of a special case to have it return values for a future epoch that hasn't begun yet.

michaelkaplan13: Whether or not the current epoch has been sealed indicates if any more blocks can be included in it or not, without making any more assumptions about what the caller may want to do with that information.

vms/proposervm/client.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func (c *Client) GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error) {
+	res := &api.GetHeightResponse{}
+	err := c.Requester.SendRequest(ctx, "proposervm.getProposedHeight", struct{}{}, res, options...)
+	return uint64(res.Height), err
+}
+
+func (c *Client) GetEpoch(ctx context.Context, options ...rpc.Option) (block.PChainEpoch, error) {
```
iansuvak: Should we name this `GetCurrentEpoch` instead? Not planned now but I can imagine having a more general API accepting a epochNumber or something to get information about a historical epoch.

vms/proposervm/service.go
michaelkaplan13: Sorry I was wrong here.

The `startTime` of a given epoch is the timestamp of the sealing block of the previous epoch. Thus, the next epoch begins as soon as the prior epoch is sealed.

So I agree with Yacov, the information about the "current epoch" should be returned, even if no block has been produced in that epoch yet.

tests/e2e/c/proposervm_epoch.go
maru-ava: We don't use gomega, only testify.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+
+			time.Sleep(upgrades.GraniteEpochDuration)
+			issueTransaction(tc, ethClient, senderKey, recipientKey.EthAddress(), txAmount)
+
+			advancedEpoch, err := proposerClient.GetCurrentEpoch(tc.DefaultContext())
+			require.NoError(err)
+
+			tc.Log().Info("advanced epoch", zap.Any("epoch", advancedEpoch))
+
+			require.Equal(
```
maru-ava: (No action required) Given that this test may be executed in parallel with other tests against a shared network, and in CI the tests are run on resource-constrained runners, is there a possibility of the epoch advancing more than once before this check occurs? If so, the previous approach of looking for an increase in the epoch may be more robust than this check for an exact value.

tests/e2e/p/l1.go
```diff
@@ -341,12 +344,74 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
 
-		advanceProposerVMPChainHeight := func() {
-			// We must wait at least [RecentlyAcceptedWindowTTL] to ensure the
-			// next block will reference the last accepted P-chain height.
-			time.Sleep((5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4)
-		}
-		tc.By("advancing the proposervm P-chain height", advanceProposerVMPChainHeight)
+		tc.By("advancing the P-Chain epoch", func() {
+			getCurrentEpochHeight := func() uint64 {
+				proposerClient := proposervm.NewClient(nodeURI.URI, "P")
```
maru-ava: (No action required) Unless the function requires a fresh client, maybe construct the client outside of the function?

tests/e2e/p/l1.go
```diff
@@ -702,7 +767,11 @@ var _ = e2e.DescribePChain("[L1]", func() {
 			})
 		})
 
-		tc.By("advancing the proposervm P-chain height", advanceProposerVMPChainHeight)
+		tc.By("advancing the proposervm P-chain height", func() {
+			// We must wait at least [RecentlyAcceptedWindowTTL] to ensure the
+			// next block will reference the last accepted P-chain height.
+			time.Sleep((5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4)
```
maru-ava: (No action required) Maybe define this interval as a constant given that it is used multiple times in this file?

Also, why it desirable to wait for this duration here but previously for either the duration or the epoch duration depending on which was greater?

tests/e2e/p/l1.go
```diff
@@ -341,12 +344,74 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+			getCurrentEpochHeight := func() uint64 {
+				proposerClient := proposervm.NewClient(nodeURI.URI, "P")
+				epoch, err := proposerClient.GetCurrentEpoch(tc.DefaultContext())
+				require.NoError(err)
+				return epoch.Height
+			}
+
+			epochBefore := getCurrentEpochHeight()
+
+			tc.By("advancing the proposervm P-chain height epoched height", func() {
```
maru-ava: (No action required) Maybe `height epoched height -> epoched height`? If not, I'm confused.

tests/e2e/p/l1.go
```diff
@@ -341,12 +344,74 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+									},
+								},
+							},
+						},
+					},
+					tc.WithDefaultContext(),
+				)
+				require.NoError(err)
+
+				tc.By("ensuring the genesis peer has accepted the tx at "+nodeURI.URI, func() {
```
maru-ava: Given that `nodeURI` is for the node targeted by the test up until now, this block would appear to be confirming the transaction on the same node that it was submitted on. I'm guessing that was not the intent?

tests/e2e/c/proposervm_epoch.go
geoff-vball: Yeah, the current approach definitely risks a flaky test. @michaelkaplan13 are you okay if I revert this change?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+		require.NoError(err)
+
+		if !upgrades.IsGraniteActivated(time.Now()) {
+			ginkgo.Skip("skipping test because granite isn't active")
+		}
+
+		// Issue a transaction to the C-Chain to advance past genesis block
+		issueTransaction(tc, ethClient, senderKey, recipientKey.EthAddress(), txAmount)
+
+		tc.Eventually(func() bool {
```
yacovm: We're doing it to see if the previous transaction issuance succeeded right? So I think we should just move it inside `issueTransaction`.

tests/e2e/c/proposervm_epoch.go
yacovm: You don't have to revert it though, you can just have an isolated network for the test.

yacovm:
```
network := tmpnet.NewDefaultNetwork("ACP-181")

e2e.GetEnv(tc).StartPrivateNetwork(network)
```

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,67 @@ var _ = e2e.DescribePChain("[L1]", func() {
 			})
 		})
 
-		advanceProposerVMPChainHeight := func() {
-			// We must wait at least [RecentlyAcceptedWindowTTL] to ensure the
-			// next block will reference the last accepted P-chain height.
-			time.Sleep((5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4)
+		timeToAdvancePChainContext := (5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4
```
yacovm: Can you explain this 5 * 1/4 computation?

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,67 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+		upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+		require.NoError(err)
+		epochDuration := upgrades.GraniteEpochDuration
+		if timeToAdvancePChainContext < epochDuration {
```
yacovm: I am confused - why are we doing this `if` and comparison? It seems like we're unsure if it's higher or not? 

And do you know where does this 125% modifier comes from?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,112 @@
...
+			bn, err := ethClient.BlockNumber(tc.DefaultContext())
+			require.NoError(err)
+			return bn != 0
+		},
+			e2e.DefaultTimeout,
+			e2e.DefaultPollingInterval,
+			"failed to see a non-zero block before timeout",
+		)
+
+		tc.By("issuing C-Chain transactions to advance the epoch", func() {
```
yacovm: I am wondering why we need an additional indentation / inner function of `tc.By`. We have a single `tc.By` here so why not fold it into the parent scope of `ginkgo.It` ?

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,67 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+			timeToAdvancePChainContext = epochDuration
 		}
-		tc.By("advancing the proposervm P-chain height", advanceProposerVMPChainHeight)
+
+		tc.By("advancing the P-Chain epoch", func() {
+			getCurrentEpochHeight := func() uint64 {
```
yacovm: A bit of a nit, but - i find it a bit redundant to have a function with 4 lines of code to save two occurrences of two lines of code. The code will be simple with just the API call and the assertion. But feel free to keep this if you prefer.

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,67 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+									},
+								},
+							},
+						},
+					},
+					tc.WithDefaultContext(),
+				)
+				require.NoError(err)
+
+				tc.By("ensuring the genesis peer has accepted the tx at "+subnetGenesisNodeURI, func() {
```
yacovm, line 390: what's a genesis peer? Peer appears as a validator in the genesis block?

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,67 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+						},
+						tests.DefaultTimeout,
+						e2e.DefaultPollingInterval,
+						"transaction not accepted",
+					)
+				})
+			})
+
+			epochAfter := getCurrentEpochHeight()
+			require.Greater(epochAfter, epochBefore)
```
yacovm: Why is this a `Greater` comparison and not a comparison where we know the exact epoch?

tests/e2e/p/l1.go
```diff
@@ -16,6 +16,7 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 
+	"github.com/ava-labs/avalanchego/api/info"
```
yacovm, line 19: General comment about the API calls - why do we not check the other fields like number or start time? Is it not possible?

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
 	Block
 
 	PChainHeight() uint64
+	PChainEpoch() PChainEpoch
```
yacovm: how do we distinguish between a granite and an ethna block? do we need to distinguish between the two in this interface? Would it make sense to have a method like `HasEpoch() bool` or perhaps just set a pointer return value to `PChainEpoch` ? 

Asking because a nil `PChainEpoch` seems to me like a valid zero value. Maybe we can have a `IsZero()` method on `PChainEpoch` that would check if `StartTime` is zero or something?

vms/proposervm/block/build.go
```diff
@@ -18,17 +18,34 @@ func BuildUnsigned(
...
-		StatelessBlock: statelessUnsignedBlock{
-			ParentID:     parentID,
-			Timestamp:    timestamp.Unix(),
-			PChainHeight: pChainHeight,
-			Certificate:  nil,
-			Block:        blockBytes,
-		},
-		timestamp: timestamp,
+	var block SignedBlock
+	if pChainEpoch.Number == 0 {
```
yacovm: So this is basically an indirect way of asking `p.vm.Upgrades.IsGraniteActivated` and "is this the first block post granite, right?

vms/proposervm/block.go
```diff
@@ -79,6 +80,34 @@ func (p *postForkCommonComponents) Height() uint64 {
...
+		return block.PChainEpoch{
+			Height:    parentPChainHeight,
+			Number:    parentEpoch.Number + 1,
+			StartTime: parentTimestamp,
+		}
+	}
+	// Otherwise, the parent did not seal the previous epoch, so the child should use the parent's
+	// epoch information. This is true even if the child crosses the epoch boundary, since sealing
+	// blocks are considered to be part of the epoch they seal.
+	return parentEpoch
```
yacovm: I think we should have a unit test just to test this function. Not trying to nitpick but this line isn't covered by tests....  Of course we don't need to have every line to be covered by tests but I think this is a critical function so we probably should have a test just for it that covers all its cases.

vms/proposervm/block.go
```diff
@@ -163,9 +194,18 @@ func (p *postForkCommonComponents) Verify(
 	}
 
 	var contextPChainHeight uint64
-	if p.vm.Upgrades.IsEtnaActivated(childTimestamp) {
+	switch {
+	case p.vm.Upgrades.IsGraniteActivated(childTimestamp):
```
yacovm, line 179: Unrelated to this PR, but, food for thought for whoever is reading this review - shouldn't we unify the code here and the code when we verify a block? It' generally bad practice to have a version of the code that verifies input and a different version that builds the same output. Ideally we'll have a single function that computes the output properly and then the verification function will just compute it exactly as the block build process and then just compare. 

This switch of IsGrantite?IsEtna?IsSomethingElse? should ideally be extracted to a separate function. 

Thoughts?

tests/e2e/c/proposervm_epoch.go
maru-ava: Only tests that are incompatible with a shared network should use a private one, and I don't think that's true of this test. Each test using a private network has to run 1 or more additional nodes whose startup extends test runtime.

vms/proposervm/service.go
```diff
@@ -0,0 +1,75 @@
...
+	height, err := p.vm.ctx.ValidatorState.GetMinimumHeight(r.Context())
+	if err != nil {
+		return fmt.Errorf("couldn't get minimum height %w", err)
+	}
+
+	reply.Height = avajson.Uint64(height)
+	return nil
+}
+
+func (p *ProposerAPI) GetCurrentEpoch(r *http.Request, _ *struct{}, reply *api.GetEpochResponse) error {
```
yacovm: I think this function isn't tested in unit tests

vms/proposervm/client.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func (c *Client) GetProposedHeight(ctx context.Context, options ...rpc.Option) (uint64, error) {
+	res := &api.GetHeightResponse{}
+	err := c.Requester.SendRequest(ctx, "proposervm.getProposedHeight", struct{}{}, res, options...)
+	return uint64(res.Height), err
+}
+
+func (c *Client) GetCurrentEpoch(ctx context.Context, options ...rpc.Option) (block.PChainEpoch, error) {
```
yacovm: I think this function isn't tested in unit tests.

vms/proposervm/vm.go
```diff
@@ -635,6 +669,25 @@ func (vm *VM) parsePreForkBlock(ctx context.Context, b []byte) (*preForkBlock, e
 	}, err
 }
 
+func (vm *VM) GetStatelessSignedBlock(blkID ids.ID) (statelessblock.SignedBlock, error) {
```
yacovm: I am not sure but it seems like this function isn't in use.

tests/e2e/p/l1.go
maru-ava: See: using a shared network.

yacovm: Well I think using the `Greater` vs exactly matching the desired value is worth it.

maru-ava: Why would we be needing to validate something precisely in an e2e test? vs doing that at a lower level?

yacovm: It's the only place we're invoking the API at from an end to end perspective so we might as well do as detailed comparison as possible

tests/e2e/c/proposervm_epoch.go
geoff-vball: I don't think this is required anymore, can remove

tests/e2e/c/proposervm_epoch.go
geoff-vball: We'll just do a greater than comparison here

tests/e2e/p/l1.go
iansuvak: I do not know about the 125% modifier. My guess is that it's a guarantee that the TTL is exceeded to increase reliability of the test. 

The `if` comparison is to make the test work both prior to and after granite activation. We need to wait for the maximum of the two durations and this is an optimization since alternative would be to sleep for the two durations additively.

vms/proposervm/block/block.go
geoff-vball: Now the first block after Granite is activated will have epoch `1`, so we can distinguish from the zero value

tests/e2e/p/l1.go
yacovm: @StephenButtolph do you know where this 5/4 computation is from?

tests/e2e/p/l1.go
geoff-vball: Yes exactly

vms/proposervm/block/build.go
geoff-vball: It's just asking the first part - `IsGraniteActivated`. The value passed is the epoch for the new block, so if the number is 0, granite is not activated and this is not a granite block. Otherwise this is a granite block

tests/e2e/p/l1.go
maru-ava: All the testing we do involves tradeoffs. Unless there is no other way to verify the behavior in question - and I'm not sure that has been established - I don't think verifying this exact value justifies the use of a private network for the entire test.

tests/e2e/p/l1.go
maru-ava: fwiw I don't think e2e tests are where validating every detail of an API call will always make sense. Such validation will rarely benefit from a network of nodes - a single node (or even a minimal API surface) in an integration test could provide the same fidelity for less cost.

vms/proposervm/service.go
```diff
@@ -33,6 +33,7 @@ func (p *ProposerAPI) GetProposedHeight(r *http.Request, _ *struct{}, reply *api
 	return nil
 }
 
+// Returns the epoch information that will be used for the next block to be proposed.
```
yacovm:
```
func (p *ProposerAPI) GetCurrentEpoch(r *http.Request, _ *struct{}, reply *api.GetEpochResponse) error {
	p.vm.ctx.Log.Debug("API called",
		zap.String("service", "proposervm"),
		zap.String("method", "getCurrentEpoch"),
	)

	// This will error if we haven't advanced past the genesis block.
	lastAccepted, err := p.vm.GetLastAccepted()
	if err != nil {
		return fmt.Errorf("couldn't get last accepted block ID %w", err)
	}
```

Why do we return an error?

upgrade/upgrade.go
```diff
@@ -77,28 +77,30 @@ var (
...
+	ApricotPhase6Time            time.Time     `json:"apricotPhase6Time"`
+	ApricotPhasePost6Time        time.Time     `json:"apricotPhasePost6Time"`
+	BanffTime                    time.Time     `json:"banffTime"`
+	CortinaTime                  time.Time     `json:"cortinaTime"`
+	CortinaXChainStopVertexID    ids.ID        `json:"cortinaXChainStopVertexID"`
+	DurangoTime                  time.Time     `json:"durangoTime"`
+	EtnaTime                     time.Time     `json:"etnaTime"`
+	FortunaTime                  time.Time     `json:"fortunaTime"`
+	GraniteTime                  time.Time     `json:"graniteTime"`
+	GraniteEpochDuration         time.Duration `json:"graniteEpochDuration"`
```
StephenButtolph, line 105: Does this epoch duration need to be included in this config? Typically we limit this config for things related to the activation of an upgrade (specifically when they change between fuji + mainnet)

vms/proposervm/block/codec.go
```diff
@@ -24,6 +24,8 @@ func init() {
 	err := errors.Join(
 		lc.RegisterType(&statelessBlock{}),
 		lc.RegisterType(&option{}),
+		lc.RegisterType(&statelessGraniteBlock{}),
+		lc.RegisterType(&statelessUnsignedGraniteBlock{}),
```
StephenButtolph: This is never marshalled as an interface, so we should not be registering its type.
```suggestion
```

api/common_args_responses.go
```diff
@@ -58,6 +58,12 @@ type GetHeightResponse struct {
 	Height avajson.Uint64 `json:"height"`
 }
 
+type GetEpochResponse struct {
```
StephenButtolph: I think we only us `api/common_args_responses.go` for structs that would otherwise be defined a bunch of times (for example, APIs that are implemented by all of platformvm, avm, xsvm, ect).

This seems pretty proposervm specific, so I think we should probably just define it at the same location as it is used.

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
...
+	Certificate  []byte `serialize:"true" json:"certificate"`
+	Block        []byte `serialize:"true" json:"block"`
+}
+
+type statelessUnsignedGraniteBlock struct {
+	ParentID            ids.ID `serialize:"true" json:"parentID"`
+	Timestamp           int64  `serialize:"true" json:"timestamp"`
+	PChainHeight        uint64 `serialize:"true" json:"pChainHeight"`
+	Certificate         []byte `serialize:"true" json:"certificate"`
+	Block               []byte `serialize:"true" json:"block"`
```
StephenButtolph: These fields are the same as with `statelessUnsignedBlock`. I think we should just reuse `statelessUnsignedBlock` rather than re-defining they values.

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
...
+	EpochNumber         uint64 `serialize:"true" json:"epochNumber"`
+	EpochStartTimestamp int64  `serialize:"true" json:"epochStartTimestamp"`
+	EpochHeight         uint64 `serialize:"true" json:"epochHeight"`
+}
+
+type PChainEpoch struct {
+	Height    uint64    `serialize:"true" json:"height"`
+	Number    uint64    `serialize:"true" json:"number"`
+	StartTime time.Time `serialize:"true" json:"startTime"`
 }
```
StephenButtolph: This currently isn't serialized, so my initial comment was going to be to remove the `serialize` tag, but I think I'd prefer to actually use this struct in `statelessUnsignedGraniteBlock`.

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
...
+	Block               []byte `serialize:"true" json:"block"`
+	EpochNumber         uint64 `serialize:"true" json:"epochNumber"`
+	EpochStartTimestamp int64  `serialize:"true" json:"epochStartTimestamp"`
+	EpochHeight         uint64 `serialize:"true" json:"epochHeight"`
+}
+
+type PChainEpoch struct {
+	Height    uint64    `serialize:"true" json:"height"`
+	Number    uint64    `serialize:"true" json:"number"`
+	StartTime time.Time `serialize:"true" json:"startTime"`
```
StephenButtolph: The codec doesn't support serializing `time.Time`. This should probably be changed to `uint64`.

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
...
-	StatelessBlock statelessUnsignedBlock `serialize:"true"`
-	Signature      []byte                 `serialize:"true"`
+	StatelessBlock statelessUnsignedBlock `serialize:"true" json:"statelessBlock"`
+	Signature      []byte                 `serialize:"true" json:"signature"`
+
+	id        ids.ID
+	timestamp time.Time
+	cert      *staking.Certificate
+	proposer  ids.NodeID
+	bytes     []byte
```
StephenButtolph, line 72: These metadata fields are duplicated onto `statelessGraniteBlock`. Can we group them together as a new struct and define `initialize`, `verify` (and the accessors) on that struct? We should be able to avoid the current code duplication.

vms/proposervm/block/block.go
```diff
@@ -35,24 +36,57 @@ type SignedBlock interface {
...
+	PChainHeight        uint64 `serialize:"true" json:"pChainHeight"`
+	Certificate         []byte `serialize:"true" json:"certificate"`
+	Block               []byte `serialize:"true" json:"block"`
+	EpochNumber         uint64 `serialize:"true" json:"epochNumber"`
+	EpochStartTimestamp int64  `serialize:"true" json:"epochStartTimestamp"`
+	EpochHeight         uint64 `serialize:"true" json:"epochHeight"`
+}
+
+type PChainEpoch struct {
+	Height    uint64    `serialize:"true" json:"height"`
```
StephenButtolph: Nit: I feel like calling the struct `Epoch` is a bit more readable than `PChainEpoch`.
```suggestion
type Epoch struct {
	PChainHeight    uint64    `serialize:"true" json:"pChainHeight"`
```

vms/proposervm/block_test.go
```diff
@@ -411,7 +417,76 @@ func TestPreEtnaContextPChainHeight(t *testing.T) {
...
+		parentEpoch,
 	)
 	require.NoError(err)
 	require.Equal(innerChildBlock, gotChild.(*postForkBlock).innerBlk)
 }
+
+func TestNextPChainEpoch(t *testing.T) {
+	var (
+		epochDuration = 5 * time.Minute
+		now           = time.Now()
```
StephenButtolph: It might make sense for us to go through and remove our usage of `time.Time` at some point, but until then, we should be consistent about the granularity).
```suggestion
		now           = time.Now().Truncate(time.Second)
```
I was playing around with `nextPChainEpoch`, and if we were to change the `block.PChainEpoch#StartTime` to be a uint64 then this test would fail weirdly because of partial rounding.

vms/proposervm/client_test.go
StephenButtolph, line 1: While I appreciate trying to test the `client`, I think that we know have a good enough of an e2e testing framework that we should avoid implementing these kinds of unit tests.

The actual method under test here is 3 lines, but we need to:
1. implement a mock struct
2. implement the `SendRequest` method
3. Intercept the type of the reply to mock the call
4. provide an error override
5. Implement the actual test with that mock

I feel like we should just use an e2e test for this and forgo the unit test.

vms/proposervm/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func New(registerer prometheus.Registerer) (Metrics, error) {
+	m := &metrics{}
+	apiRequestMetrics, err := metric.NewAPIInterceptor(registerer)
+	errs := wrappers.Errs{Err: err}
+	m.APIInterceptor = apiRequestMetrics
+
+	return m, errs.Err
+}
```
StephenButtolph: Unless we add additional metrics, it feels like we should just use the API interceptor directly

vms/proposervm/post_fork_block_test.go
```diff
@@ -47,8 +47,9 @@ func TestOracle_PostForkBlock_ImplementsInterface(t *testing.T) {
 	var (
 		activationTime = time.Unix(0, 0)
 		durangoTime    = activationTime
+		graniteTime    = activationTime
 	)
-	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
+	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, graniteTime, 0)
```
StephenButtolph: Not for this PR, but we should update the proposervm to use the `upgradetest` pacakge for specifying upgrades. We should be able to significantly reduce the need to change tests like these if we were to just specify which fork the test was targeting.

vms/proposervm/post_fork_option.go
```diff
@@ -105,11 +111,17 @@ func (b *postForkOption) buildChild(ctx context.Context) (Block, error) {
 		)
 		return nil, err
 	}
+	parentEpoch, err := b.pChainEpoch(ctx)
+	if err != nil {
+		return nil, err
```
StephenButtolph, line 121: Should we add an error log like above?

vms/proposervm/pre_fork_block.go
```diff
@@ -211,6 +211,11 @@ func (b *preForkBlock) buildChild(ctx context.Context) (Block, error) {
 		parentID,
 		newTimestamp,
 		pChainHeight,
+		block.PChainEpoch{
+			Height:    pChainHeight,
+			Number:    0,
+			StartTime: time.Time{},
+		},
```
StephenButtolph: Have we considered whether or not we want to activate the epoch on the local network in this block?

Essentially, If the chain goes from AP4 not being activated to Granite being activated, what do we want to have happen?

vms/proposervm/state/block_state_test.go
```diff
@@ -22,6 +22,11 @@ func testBlockState(require *require.Assertions, bs BlockState) {
 	parentID := ids.ID{1}
 	timestamp := time.Unix(123, 0)
 	pChainHeight := uint64(2)
+	pChainEpoch := block.PChainEpoch{
+		Height:    uint64(2),
+		Number:    uint64(0),
```
StephenButtolph: unnecessary casts:
```suggestion
		Height:    2,
		Number:    0,
```

upgrade/upgrade.go
geoff-vball: We're not going to change it between Fuji and Mainnet, but we do need to be able to change it for local networks and e2e tests. Where would be a better place for it to live?

vms/proposervm/pre_fork_block.go
iansuvak: Resolved out of band but documenting here as well. 

We've decided to keep the current behavior which is that if we go from AP4 not being activated to Granite being activated such as when starting a new chain and building a first child of the genesis block the first block produced will not be a `statelessGraniteBlock` and will not have epoch information in it's serialization, and the first block following that will start the first epoch.

vms/proposervm/block/parse_test.go
```diff
@@ -21,7 +21,11 @@ func TestParseBlocks(t *testing.T) {
 	parentID := ids.ID{1}
 	timestamp := time.Unix(123, 0)
 	pChainHeight := uint64(2)
-	pChainEpoch := Epoch{}
+	pChainEpoch := Epoch{
+		PChainHeight: 2,
+		Number:       1,
+		StartTime:    timestamp.Unix(),
+	}
```
StephenButtolph: This diff doesn't seem needed.
```suggestion
	pChainEpoch := Epoch{}
```

vms/proposervm/client.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
geoff-vball: I think this file is redundant. It's the same client as in `vms/proposervm/client_jsonrpc.go`

vms/proposervm/state/block_state_test.go
```diff
@@ -22,6 +22,11 @@ func testBlockState(require *require.Assertions, bs BlockState) {
 	parentID := ids.ID{1}
 	timestamp := time.Unix(123, 0)
 	pChainHeight := uint64(2)
+	pChainEpoch := block.Epoch{
+		PChainHeight: 2,
+		Number:       1,
+		StartTime:    timestamp.Unix(),
+	}
```
StephenButtolph: This diff also doesn't seem needed.

vms/proposervm/service.go
geoff-vball: Why wouldn't we return an error?

upgrade/upgrade.go
```diff
@@ -77,28 +77,30 @@ var (
 		EtnaTime:                     InitiallyActiveTime,
 		FortunaTime:                  InitiallyActiveTime,
 		GraniteTime:                  UnscheduledActivationTime,
+		GraniteEpochDuration:         30 * time.Second,
```
StephenButtolph, line 82: Should we be populating the mainnet and fuji configs?

upgrade/upgrade.go
StephenButtolph, line 105: I think we should populate this in the `upgradetest` helpers as well.

StephenButtolph, line 105: We never seem to set this in any of our proposervm unit tests. Should we be testing with some more interesting values of this?

tests/e2e/p/l1.go
```diff
@@ -341,12 +346,63 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
-			// We must wait at least [RecentlyAcceptedWindowTTL] to ensure the
-			// next block will reference the last accepted P-chain height.
-			time.Sleep((5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4)
+		upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+		require.NoError(err)
+
+		timeToAdvancePChainContext := (5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4
+		if timeToAdvancePChainContext < upgrades.GraniteEpochDuration {
+			timeToAdvancePChainContext = upgrades.GraniteEpochDuration
 		}
```
StephenButtolph:
```suggestion
		timeToAdvancePChainContext := (5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4
		timeToAdvancePChainContext = max(timeToAdvancePChainContext, upgrades.GraniteEpochDuration)
```

tests/e2e/p/l1.go
```diff
@@ -71,11 +74,13 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
-			keychain   = env.NewKeychain()
-			baseWallet = e2e.NewWallet(tc, keychain, nodeURI)
-			pWallet    = baseWallet.P()
-			pClient    = platformvm.NewClient(nodeURI.URI)
-			owner      = &secp256k1fx.OutputOwners{
+			keychain       = env.NewKeychain()
+			baseWallet     = e2e.NewWallet(tc, keychain, nodeURI)
+			pWallet        = baseWallet.P()
+			pClient        = platformvm.NewClient(nodeURI.URI)
+			proposerClient = proposervm.NewJSONRPCClient(nodeURI.URI, "P")
```
StephenButtolph, line 83: I would probably prefer usage of the grpc service where possible, but I'm fine with this usage.

tests/fixture/tmpnet/network.go
```diff
@@ -184,13 +184,13 @@ func BootstrapNewNetwork(
...
+		return stacktrace.Wrap(fmt.Errorf("failed to check VM binaries: %w", err))
 	}
 	if err := network.EnsureDefaultConfig(ctx, log); err != nil {
-		return stacktrace.Wrap(err)
+		return stacktrace.Wrap(fmt.Errorf("failed to ensure default config: %w", err))
 	}
 	if err := network.Create(rootNetworkDir); err != nil {
-		return stacktrace.Wrap(err)
+		return stacktrace.Wrap(fmt.Errorf("failed to create network: %w", err))
 	}
```
StephenButtolph, line 194: Now that we have a stacktrace, do these additional wrappings actually increase observability?

I feel like the wrapping just shows where the call is being performed, but we already have that information.

vms/proposervm/block.go
```diff
@@ -81,6 +82,34 @@ func (p *postForkCommonComponents) Height() uint64 {
 	return p.innerBlk.Height()
 }
 
+// Calculates a block's P-Chain epoch height based on its ancestor's epoch membership
+func nextPChainEpoch(parentPChainHeight uint64, parentEpoch block.Epoch, parentTimestamp time.Time, epochDuration time.Duration) block.Epoch {
```
StephenButtolph: Currently we separate the logic of "Is the next block post granite" and "what should the epoch of the next block be".

I feel like this function would be best suited if it took in the parent state and returned the child epoch, returning `block.Epoch{}` if epoching hasn't been enabled yet.

That way in the verification code we can simply call this function and verify the provided epoch is equivalent (no fork check) and similarly in building we can simply call this function and then pass the epoch into `Build`.

vms/proposervm/block_test.go
```diff
@@ -101,7 +102,8 @@ func TestPostForkCommonComponents_buildChild(t *testing.T) {
 		context.Background(),
 		parentID,
 		parentTimestamp,
-		pChainHeight-1,
+		pChainHeight,
```
StephenButtolph, line 106: We changed `pChainHeight-1` to `pChainHeight` here... was that expected? I don't know why this would be impacted by this PR... But maybe this is a bug fix?

StephenButtolph: Talked offline. This is needed because the above mock call sets the expected block context, and with epoching enabled we will pass in the parent pchain height.

vms/proposervm/service.go
```diff
@@ -53,6 +54,25 @@ func (c *connectrpcService) GetProposedHeight(ctx context.Context, r *connect.Re
 	}), nil
 }
 
+// GetCurrentEpoch implements proposervmconnect.ProposerVMHandler.
+func (c *connectrpcService) GetCurrentEpoch(ctx context.Context, _ *connect.Request[pb.GetCurrentEpochRequest]) (*connect.Response[pb.GetCurrentEpochReply], error) {
```
StephenButtolph: This function needs locking similar to the locking done in `GetProposedHeight`. Or the locking could be handled in `getCurrentEpoch`.

vms/proposervm/service.go
```diff
@@ -84,3 +104,59 @@ func (j *jsonrpcService) GetProposedHeight(r *http.Request, _ *struct{}, reply *
...
 }
+
+type GetEpochResponse struct {
+	Number       avajson.Uint64 `json:"number"`
+	StartTime    avajson.Uint64 `json:"startTime"`
+	PChainHeight avajson.Uint64 `json:"pChainHeight"`
+}
+
+// Returns the epoch information that will be used for the next block to be proposed.
+func (j *jsonrpcService) GetCurrentEpoch(r *http.Request, _ *struct{}, reply *GetEpochResponse) error {
```
StephenButtolph, line 116: This function needs locking similar to the locking done in `GetProposedHeight`. Or the locking could be handled in `getCurrentEpoch`.

vms/proposervm/service.go
```diff
@@ -84,3 +104,59 @@ func (j *jsonrpcService) GetProposedHeight(r *http.Request, _ *struct{}, reply *
...
+	// This will error if we haven't advanced past the genesis block.
+	lastAccepted, err := vm.GetLastAccepted()
+	if err != nil {
+		return block.Epoch{}, fmt.Errorf("couldn't get last accepted block ID: %w", err)
+	}
+
+	latestBlock, err := vm.getPostForkBlock(ctx, lastAccepted)
+	if err != nil {
+		return block.Epoch{}, fmt.Errorf("couldn't get latest block: %w", err)
+	}
```
StephenButtolph:
```suggestion
	latestBlock, err := vm.getBlock(ctx, vm.preferred)
	if err != nil {
		return block.Epoch{}, fmt.Errorf("couldn't get latest block: %w", err)
	}
```

vms/proposervm/vm.go
StephenButtolph, line 1: diff in this file seems unrelated

vms/proposervm/vm_test.go
StephenButtolph, line 1: Diff in this file doesn't seem relevant.

vms/proposervm/state/block_state_test.go
StephenButtolph, line 1: Diff in this file doesn't seem relevant.

StephenButtolph (approved):

vms/proposervm/block.go
michaelkaplan13: Completely agree with the premise of this comment, and that it's not for this PR either way.

However, for block building specifically this seems difficult/impossible because there are many different valid ways to build the next block, so we can't just simply build locally and compare to what another build proposed.

upgrade/upgrade.go
```diff
@@ -37,6 +37,7 @@ var (
 		EtnaTime:                  time.Date(2024, time.December, 16, 17, 0, 0, 0, time.UTC),
 		FortunaTime:               time.Date(2025, time.April, 8, 15, 0, 0, 0, time.UTC),
 		GraniteTime:               UnscheduledActivationTime,
+		GraniteEpochDuration:      5 * time.Minute,
```
iansuvak, line 40: Did we decide that this is the actual duration? Should this be included in the ACP?

iansuvak (approved):
Focused on the most recent set of changes but took a read of everything. I do like the extraction of the logic into a separate acp181 package and change of interface that it takes in upgrades instead of the epoch duration.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,110 @@
...
+			ethClient  = e2e.NewEthClient(tc, nodeURI)
+			infoClient = info.NewClient(nodeURI.URI)
+		)
+
+		ctx := tc.DefaultContext()
+		upgrades, err := infoClient.Upgrades(ctx)
+		require.NoError(err)
+
+		if !upgrades.IsGraniteActivated(time.Now()) {
+			ginkgo.Skip("skipping test because granite isn't active")
```
yacovm, line 42: stupid question, but - why can't we force granite to activate from within the test?

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		// Issue a transaction to the C-Chain to advance past genesis block
```
yacovm, line 45: Is this essential? What is the purpose of making this first transaction?

upgrade/upgrade.go
```diff
@@ -59,6 +60,7 @@ var (
 		EtnaTime:                  time.Date(2024, time.November, 25, 16, 0, 0, 0, time.UTC),
 		FortunaTime:               time.Date(2025, time.March, 13, 15, 0, 0, 0, time.UTC),
 		GraniteTime:               UnscheduledActivationTime,
+		GraniteEpochDuration:      5 * time.Minute,
 	}
 	Default = Config{
```
yacovm, line 65: Is the default only used in tests? Wondering why it's called that instead of test?

connectproto/proposervm/service.proto
```diff
@@ -20,3 +23,13 @@ message GetProposedHeightReply {
...
   uint64 height = 1;
 }
+
+// Request to get the current epoch.
+message GetCurrentEpochRequest {}
+
+// Response containing the current epoch.
+message GetCurrentEpochReply {
+  uint64 p_chain_height = 1;
+  uint64 number = 2;
```
michaelkaplan13: Nit: I don't think it matters either way for the proto definition here, but it seems a bit odd to have the `p_chain_height` index as the first field and `number` as the second field when `number` is the unique identifier and `p_chain_height` is not necessarily unique.

upgrade/upgrade.go
michaelkaplan13: Yes and yes. It was included in the ACP [here](https://github.com/avalanche-foundation/ACPs/pull/239).

michaelkaplan13 (approved):
I joined the review party late here, but think this got to a very clean implementation that is easy to follow from the specification.

Just had one nit/question about the proto field ordering, but I don't think it's consequential either way.

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,110 @@
...
+		require.NoError(err)
+
+		tc.Log().Info("advanced epoch",
+			zap.Reflect("epoch", advancedEpoch),
+		)
+
+		require.Greater(
+			advancedEpoch.Number,
+			initialEpoch.Number,
+			"expected epoch number to advance, but it did not",
```
joshua-kim: This PR uses this in multiple places but I think that we typically don't use these string require assertions since the require is self-documenting

yacovm (approved):
This PR seems much slimmer than it was a few weeks ago

tests/e2e/c/proposervm_epoch.go
```diff
@@ -0,0 +1,110 @@
...
+		)
+		require.GreaterOrEqual(
+			advancedEpoch.StartTime,
+			initialEpoch.StartTime+int64(upgrades.GraniteEpochDuration.Seconds()),
+			"expected epoch start time to advance, but it did not",
+		)
+		// P-chain height may not increase if no new blocks were created on the P-chain
```
joshua-kim: Should this be above the previous line for locality?

upgrade/upgrade.go
StephenButtolph: The default is used for any networks that aren't Fuji or Mainnet (unless explicitly provided)

tests/e2e/c/proposervm_epoch.go
StephenButtolph: We run the entire E2E suite twice, once with granite enabled once with it disabled.

tests/e2e/p/l1.go
```diff
@@ -342,9 +347,66 @@ var _ = e2e.DescribePChain("[L1]", func() {
 		})
 
 		advanceProposerVMPChainHeight := func() {
-			// We must wait at least [RecentlyAcceptedWindowTTL] to ensure the
-			// next block will reference the last accepted P-chain height.
-			time.Sleep((5 * platformvmvalidators.RecentlyAcceptedWindowTTL) / 4)
+			upgrades, err := infoClient.Upgrades(tc.DefaultContext())
+			require.NoError(err)
+
+			const timeToAdvancePChainWindow = 5 * platformvmvalidators.RecentlyAcceptedWindowTTL / 4
```
joshua-kim: nit: I noticed a few local const declarations in this PR. Our usual convention is to keep const for package-level values, since local constants don’t provide much benefit over regular variables and can make the code look unnecessarily verbose.

tests/e2e/p/l1.go
```diff
@@ -342,9 +347,66 @@ var _ = e2e.DescribePChain("[L1]", func() {
...
+			if !upgrades.IsGraniteActivated(time.Now()) {
+				// We must wait at least [RecentlyAcceptedWindowTTL] to ensure
```
joshua-kim: 1. I think this `[bracket notation]` syntax is generally reserved for godoc to make clickable links to identifiers to either things in an imported package or to something within the same package. For in-line comments like this, it doesn't do anything.
2. It looks like we renamed this var, so do we need to update this as well?

vms/proposervm/acp181/epoch.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"time"
+
+	"github.com/ava-labs/avalanchego/upgrade"
+	"github.com/ava-labs/avalanchego/vms/proposervm/block"
+)
+
+// Epoch returns a child block's epoch based on its parent.
+func Epoch(
```
joshua-kim: Constructors typically default to being named `New*`, so `NewEpoch` might make sense for a name here

tests/e2e/c/proposervm_epoch.go
iansuvak: In cases when Granite is initially active we are going from genesis which is pre-AP4 to granite immediately. We made the choice for the first block in this case to not start the epoch but the immediate block following i.e. the first block with post-fork block as it's parent to start the first epoch

vms/proposervm/acp181/epoch.go
```diff
@@ -0,0 +1,51 @@
...
+
+	epochEndTime := time.Unix(parentEpoch.StartTime, 0).Add(upgrades.GraniteEpochDuration)
+	if parentTimestamp.Before(epochEndTime) {
+		// If the parent was issued before the end of its epoch, then it did not
+		// seal the epoch.
+		return parentEpoch
+	}
+
+	// The parent sealed the epoch. So, the child is the first block of the new
+	// epoch.
```
joshua-kim:
```suggestion
	// The parent sealed the epoch, so the child is the first block of the new
	// epoch.
```

vms/proposervm/acp181/epoch_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/ava-labs/avalanchego/upgrade"
+	"github.com/ava-labs/avalanchego/upgrade/upgradetest"
+
+	statelessblock "github.com/ava-labs/avalanchego/vms/proposervm/block"
+)
+
+func TestEpoch(t *testing.T) {
+	var (
+		now            = time.Now().Truncate(time.Second)
```
joshua-kim: nit: use hard-coded timestamps to keep tests deterministic
```suggestion
		start            = time.Time{}
```

vms/proposervm/block_test.go
```diff
@@ -444,3 +448,115 @@ func TestPreGraniteBlock_NonZeroEpoch(t *testing.T) {
...
+			return nil, errUnknownBlock
+		}
+	}
+	coreVM.BuildBlockF = func(context.Context) (snowman.Block, error) {
+		return coreParentBlk, nil
+	}
+
+	// Build the first proposervm block so that verification is on top of a
+	// post-fork block.
+	parentTime := time.Now().Truncate(time.Second)
```
joshua-kim: Similar comment here

connectproto/proposervm/service.proto
```diff
@@ -20,3 +23,13 @@ message GetProposedHeightReply {
   // right now.
   uint64 height = 1;
 }
+
+// Request to get the current epoch.
+message GetCurrentEpochRequest {}
+
+// Response containing the current epoch.
+message GetCurrentEpochReply {
+  uint64 p_chain_height = 1;
```
joshua-kim: Could we document these fields as well since these proto files are basically the docs for the connect apis?

connectproto/proposervm/service.proto
joshua-kim: +1

PR #3715, opened 2025-02-11
P chain mempool

joshua-kim:

PR #4241, opened 2025-09-04
Adds docs change to #3715

meaghanfitzgerald:
Adds `mempool-gas-capacity` to the config docs. Auto formatted on save.

## Why this should be merged

Documentation

yes, with #3715

### DIFF
--- a/vms/platformvm/config/config.md
+++ b/vms/platformvm/config/config.md
@@ -4,24 +4,23 @@ This document provides details about the configuration options available for the
 
 In order to specify a configuration for the PlatformVM, you need to define a `Config` struct and its parameters. The default values for these parameters are:
 
-| Option                               | Type            | Default            |
-| ------------------------------------ | --------------- | ------------------ |
-| `network`                            | `Network`       | `DefaultNetwork`   |
-| `block-cache-size`                   | `int`           | `64 * units.MiB`   |
-| `tx-cache-size`                      | `int`           | `128 * units.MiB`  |
-| `transformed-subnet-tx-cache-size`   | `int`           | `4 * units.MiB`    |
-| `reward-utxos-cache-size`            | `int`           | `2048`             |
-| `chain-cache-size`                   | `int`           | `2048`             |
-| `chain-db-cache-size`                | `int`           | `2048`             |
-| `block-id-cache-size`                | `int`           | `8192`             |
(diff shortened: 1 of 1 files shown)


vms/platformvm/config/config.md
```diff
@@ -30,25 +31,25 @@ The Network configuration defines parameters that control the network's gossip a
...
-| `pull-gossip-frequency` | `time.Duration` | `1500 * time.Millisecond` | Frequency of pull gossip rounds |
-| `pull-gossip-throttling-period` | `time.Duration` | `10 * time.Second` | Time window for throttling pull requests |
-| `pull-gossip-throttling-limit` | `int` | `2` | Maximum number of pull queries allowed per validator within the throttling window |
-| `expected-bloom-filter-elements` | `int` | `8 * 1024` | Expected number of elements when creating a new bloom filter. Larger values increase filter size |
-| `expected-bloom-filter-false-positive-probability` | `float64` | `0.01` | Target probability of false positives after inserting the expected number of elements. Lower values increase filter size |
-| `max-bloom-filter-false-positive-probability` | `float64` | `0.05` | Threshold for bloom filter regeneration. Filter is refreshed when false positive probability exceeds this value |
+| Field                                              | Type            | Default                   | Description                                                                                                                      |
```
joshua-kim: This diff hunk looks unrelated - can we unstage this? If we want to merge in this hunk we could move it to its own branch w/ `git checkout -p`

meaghanfitzgerald: it's just spacing changes because of my autoformatter for .md files, would that cause any issues? otherwise this would be an annoying nit everytime we update the config markdown file

joshua-kim: It doesn't cause any breaking issues, but it messes with the git history. As an example, if we revert this commit we also inadvertently revert the spacing changes. It seems pedantic and although I am in favor of scope creeping at times to fix tech debt (refactors), cosmetic changes are generally trivial to break out into a separate commit/PR.

If your IDE is auto-formatting on-save, there are sometimes ways to get it to only auto-format on just lines in the diff (but I don't use VS Code, so I don't know how to do it there). If I'm battling my IDE's formatting I'll usually just use the CLI to checkout w/ `git checkout -p`.

meaghanfitzgerald: reverted and pushed the relevant change

joshua-kim (approved):

PR #4246, opened 2025-09-04
Pass x/sync proofs as bytes

alarso16:
## Why this should be merged

Other implementations of x/sync shouldn't be required to implement the proof part in the same format as merkledb, even though they should use the same message format for the x/sync manager

## How this works

Uses `bytes` instead of the proof type in proto, requiring the proofs to be able to marshal themselves

## How this was tested

Existing UT

No

### DIFF
--- a/x/blockdb/README.md
+++ b/x/blockdb/README.md
@@ -9,6 +9,7 @@ BlockDB is a specialized database optimized for blockchain blocks.
 - **Flexible Write Ordering**: Supports out-of-order block writes for bootstrapping
 - **Configurable Durability**: Optional `syncToDisk` mode guarantees immediate recoverability
 - **Automatic Recovery**: Detects and recovers unindexed blocks after unclean shutdowns
+- **Block Compression**: zstd compression for block data
 
 ## Design
 
@@ -66,7 +67,7 @@ Index Entry (16 bytes):
 ├────────────────────────────────┼─────────┤
 │ Data File Offset               │ 8 bytes │
 │ Block Data Size                │ 4 bytes │
-│ Header Size                    │ 4 bytes │
+│ Reserved                       │ 4 bytes │
 └────────────────────────────────┴─────────┘
 ```
 
@@ -75,14 +76,13 @@ Index Entry (16 bytes):
 Each block in the data file is stored with a block entry header followed by the raw block data:
 
 ```
-Block Entry Header (26 bytes):
+Block Entry Header (22 bytes):
 ┌────────────────────────────────┬─────────┐
 │ Field                          │ Size    │
 ├────────────────────────────────┼─────────┤
 │ Height                         │ 8 bytes │
 │ Size                           │ 4 bytes │
 │ Checksum                       │ 8 bytes │
-│ Header Size                    │ 4 bytes │
 │ Version                        │ 2 bytes │
 └────────────────────────────────┴─────────┘
 ```
@@ -93,7 +93,7 @@ BlockDB allows overwriting blocks at existing heights. When a block is overwritt
 
 ### Fixed-Size Index Entries
 
-Each index entry is exactly 16 bytes on disk, containing the offset, size, and header size. This fixed size enables direct calculation of where each block's index entry is located, providing O(1) lookups. For blockchains with high block heights, the index remains efficient, even at height 1 billion, the index file would only be ~16GB.
+Each index entry is exactly 16 bytes on disk, containing the offset, size, and reserved bytes for future use. This fixed size enables direct calculation of where each block's index entry is located, providing O(1) lookups. For blockchains with high block heights, the index remains efficient, even at height 1 billion, the index file would only be ~16GB.
 
 ### Durability and Fsync Behavior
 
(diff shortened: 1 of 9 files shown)


x/merkledb/proof.go
```diff
@@ -22,6 +25,11 @@ import (
 const verificationCacheSize = math.MaxUint16
 
 var (
+	_ encoding.BinaryMarshaler   = (*ChangeProof)(nil)
+	_ encoding.BinaryUnmarshaler = (*ChangeProof)(nil)
+	_ encoding.BinaryMarshaler   = (*RangeProof)(nil)
+	_ encoding.BinaryUnmarshaler = (*RangeProof)(nil)
```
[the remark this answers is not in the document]
alarso16: ???

x/merkledb/proof.go
```diff
@@ -238,6 +246,18 @@ func (proof *Proof) UnmarshalProto(pbProof *pb.Proof) error {
 
 type RangeProof ChangeProof
 
+func (proof *RangeProof) MarshalBinary() (data []byte, err error) {
```
joshua-kim: 1. nit: first letter of receiver should be `r` (convention is typically first letter of type)
2. Avoid named return parameters - can lead to subtle shadowing that's difficult to notice.
3. Could probably just name this `Marshal` - there's no other `Marshal*` function on this type to disambiguate against

x/merkledb/proof.go
```diff
@@ -238,6 +246,18 @@ func (proof *Proof) UnmarshalProto(pbProof *pb.Proof) error {
...
+	return proto.Marshal(proof.ToProto())
+}
+
+func (proof *RangeProof) UnmarshalBinary(data []byte) error {
+	var pbRangeProof pb.RangeProof
+	if err := proto.Unmarshal(data, &pbRangeProof); err != nil {
+		return err
+	}
+	return proof.UnmarshalProto(&pbRangeProof)
```
joshua-kim: nit: whitespace between `}` and `return`

x/merkledb/proof.go
```diff
@@ -445,6 +465,18 @@ type ChangeProof struct {
...
+}
+
+func (proof *ChangeProof) UnmarshalBinary(data []byte) error {
+	var pbChangeProof pb.ChangeProof
+	if err := proto.Unmarshal(data, &pbChangeProof); err != nil {
+		return err
+	}
+	return proof.UnmarshalProto(&pbChangeProof)
+}
+
```
joshua-kim: Similar comments in this hunk

x/sync/network_server.go
```diff
@@ -145,9 +145,13 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 					BytesLimit: req.BytesLimit,
 				},
 				func(rangeProof *merkledb.RangeProof) ([]byte, error) {
+					proofBytes, err := rangeProof.MarshalBinary()
+					if err != nil {
+						return nil, err
+					}
 					return proto.Marshal(&pb.SyncGetChangeProofResponse{
```
joshua-kim: same nit here

x/sync/network_server.go
```diff
@@ -163,9 +167,16 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 		}
 
 		// We generated a change proof. See if it's small enough.
-		proofBytes, err := proto.Marshal(&pb.SyncGetChangeProofResponse{
+		changeProofBytes, err := changeProof.MarshalBinary()
+		if err != nil {
+			return nil, &common.AppError{
+				Code:    p2p.ErrUnexpected.Code,
```
joshua-kim, line 174: Specifying error codes probably isn't needed - the zero value is usable + we don't reason against them except for timeouts anyways

x/merkledb/proof.go
alarso16: I'm just using the `encoding.BinaryMarshaller` interface for easy type constraints - is there a different common interface it should use? The only reason is that it seems like using a common interface would be better for the later type constraint.

I'll fix the other things though

joshua-kim: Ah - if you want to use that interface then it's reasonable to use these names. You could also just define your own interface as well if you don't want to be forced to use that interface as well. Proto is a binary format so satisfying `encoding.BinaryMarshaler` sounds reasonable.

x/sync/network_server.go
```diff
@@ -163,9 +168,15 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 		}
 
 		// We generated a change proof. See if it's small enough.
-		proofBytes, err := proto.Marshal(&pb.SyncGetChangeProofResponse{
+		changeProofBytes, err := changeProof.MarshalBinary()
+		if err != nil {
+			return nil, &common.AppError{
+				Message: fmt.Sprintf("failed to marshal change proof: %s", err),
+			}
```
StephenButtolph: I don't think we ever use `Code: 0`. This should probably specify `Code: p2p.ErrUnexpected.Code`

x/merkledb/proof.go
```diff
@@ -238,19 +246,32 @@ func (proof *Proof) UnmarshalProto(pbProof *pb.Proof) error {
...
+func (r *RangeProof) UnmarshalBinary(data []byte) error {
+	var pbRangeProof pb.RangeProof
+	if err := proto.Unmarshal(data, &pbRangeProof); err != nil {
+		return err
+	}
+
+	return r.UnmarshalProto(&pbRangeProof)
+}
+
+func (r *RangeProof) ToProto() *pb.RangeProof {
```
StephenButtolph, line 262: Can `ToProto` and `UnmarshalProto` (on both `RangeProof` and `ChangeProof`) be unexported?

alarso16: Not yet unfortunately, need to wait for the next PR...
Still used in `x/sync/g_db`

PR #4247, opened 2025-09-04
refactor: verify range proof on DB

alarso16:
## Why this should be merged

Some of the manager attributes (e.g. hasher, tokenSize) really belong to the database. Additionally, the manager explicitly uses merkledb proof verification.

## How this works

Moves range proof verification to a database method. All the proto changes are related to the gRPC stuff in `x/sync/g_db`

## How this was tested

Existing UT

No


x/merkledb/db.go
```diff
@@ -1390,3 +1456,224 @@ func cacheEntrySize(key Key, n *node) int {
...
+	// The largest handled key isn't equal to the end of the range.
+	// Find the start of the next key range to fetch.
+	// Note that [largestHandledKey] can't be Nothing.
+	// Proof: Suppose it is. That means that we got a range/change proof that proved up to the
+	// greatest key-value pair in the database. That means we requested a proof with no upper
+	// bound. That is, [workItem.end] is Nothing. Since we're here, [bothNothing] is false,
+	// which means [workItem.end] isn't Nothing. Contradiction.
+	lastReceivedKey := largestHandledKey.Value()
+
+	if len(endProof) == 0 {
```
alarso16, line 1497: All changes this line and below are cut-and-paste from `x/sync/manager.go`

x/sync/sync_test.go
```diff
@@ -237,92 +234,69 @@ func Test_Sync_FindNextKey_InSync(t *testing.T) {
...
+	require.NoError(dbToSync.Put([]byte{0x11, 0x11}, []byte{2}))
 
-	syncRoot, err := db.GetMerkleRoot(context.Background())
+	// Create empty DB to commit to one key
+	db, err := merkledb.New(ctx, memdb.New(), newDefaultDBConfig())
 	require.NoError(err)
+	require.NoError(db.Put([]byte{0x13}, []byte{3}))
 
-	ctx := context.Background()
-	syncer, err := NewManager(ManagerConfig{
```
alarso16, line 254: All these tests that don't require the manager will be moved to a different package eventually

x/sync/manager.go
```diff
@@ -327,7 +312,7 @@ func (m *Manager) requestChangeProof(ctx context.Context, work *workItem) {
 
 	if work.localRootID == targetRootID {
 		// Start root is the same as the end root, so we're done.
-		m.completeWorkItem(ctx, work, work.end, targetRootID, nil)
+		m.completeWorkItem(work, work.end, targetRootID)
```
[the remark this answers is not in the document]
alarso16: I don't think this is true - weird things will happen to the processed work list

x/merkledb/db.go
```diff
@@ -1390,3 +1456,224 @@ func cacheEntrySize(key Key, n *node) int {
...
+		// We try to find the next key to fetch by looking at the end proof.
+		// If the end proof is empty, we have no information to use.
+		// Start fetching from the next key after [lastReceivedKey].
+		nextKey := lastReceivedKey
+		nextKey = append(nextKey, 0)
```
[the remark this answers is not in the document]
alarso16: I am absolutely not changing this logic if I don't have to. Out of scope for the PR

x/merkledb/db.go
```diff
@@ -87,23 +89,25 @@ type ChangeProofer interface {
...
 	VerifyChangeProof(
 		ctx context.Context,
 		proof *ChangeProof,
 		start maybe.Maybe[[]byte],
 		end maybe.Maybe[[]byte],
 		expectedEndRootID ids.ID,
+		maxLength int,
 	) error
 
 	// CommitChangeProof commits the key/value pairs within the [proof] to the db.
```
demosdemon, line 108: nit: comment needs an update explaining `end` and the return value.

x/sync/manager.go
```diff
@@ -504,32 +489,30 @@ func (m *Manager) handleRangeProofResponse(
...
 		protoutils.ProtoToMaybe(request.StartKey),
 		protoutils.ProtoToMaybe(request.EndKey),
-		request.RootHash,
-		m.tokenSize,
-		m.config.Hasher,
+		root,
+		int(request.KeyLimit),
 	); err != nil {
-		return err
+		return fmt.Errorf("%w due to %w", errInvalidRangeProof, err)
```
joshua-kim: Colons are frequently used to delimit errors
```suggestion
		return fmt.Errorf("%w: %w", errInvalidRangeProof, err)
```

x/merkledb/db.go
```diff
@@ -1198,6 +1254,26 @@ func (db *merkleDB) VerifyChangeProof(
...
 
+func (db *merkleDB) VerifyRangeProof(
+	ctx context.Context,
+	proof *RangeProof,
+	start maybe.Maybe[[]byte],
+	end maybe.Maybe[[]byte],
+	expectedEndRootID ids.ID,
+	maxLength int,
+) error {
+	// use db parameters
```
joshua-kim: What does this comment mean?

alarso16: It was intended to mean "technically, the proof is separate from the DB, but we use the same parameters to verify it", but that doesn't really seem necessary, so I'll just delete the comment

x/merkledb/db.go
```diff
@@ -122,10 +131,35 @@ type RangeProofer interface {
 		maxLength int,
 	) (*RangeProof, error)
 
+	// Returns nil iff all the following hold:
+	//   - [start] <= [end].
+	//   - [proof] is non-empty.
+	//   - [proof.KeyValues] is not longer than [maxLength].
+	//   - All keys in [proof.KeyValues] and [proof.DeletedKeys] are in [start, end].
```
rrazvan1, line 138: DeletedKeys for RangeProof?

x/merkledb/db.go
```diff
@@ -122,10 +131,35 @@ type RangeProofer interface {
...
+	//     If [start] is nothing, all keys are considered > [start].
+	//     If [end] is nothing, all keys are considered < [end].
+	//   - [proof.KeyValues] and [proof.DeletedKeys] are sorted in order of increasing key.
```
rrazvan1: DeletedKeys for RangeProof?

x/merkledb/db.go
```diff
@@ -429,17 +469,26 @@ func (db *merkleDB) CommitChangeProof(ctx context.Context, proof *ChangeProof) e
...
 	view, err := newView(db, db, ViewChanges{BatchOps: ops})
 	if err != nil {
-		return err
+		return maybe.Nothing[[]byte](), err
 	}
-	return view.commitToDB(ctx)
+
+	// Commit the changes to the DB.
+	largestKey := end
+	if len(proof.KeyChanges) > 0 {
```
rrazvan1: This if is always true. We have the early return at line 457 if **KeyChanges** is empty.

x/merkledb/db.go
rrazvan1: Ok, but why do we still keep comments about external stuff (like workItem)?

alarso16: updated!

PR #4248, opened 2025-09-05
feat(load): parameterize worker count

RodrigoVillar:
## Why this should be merged

As mentioned in #4232, we should have a load test where the amount of load is parametrizable. Using more workers increases the load put on the C-Chain but currently, the number of workers is tied to the number of nodes.

## How this works

Adds a `num-workers` flag to allow users to specify the number of workers to use in the load test.

## How this was tested

CI + inspected `genesis.json` file to make sure the number of accounts being funded was equal to the number of workers.

No

### DIFF
--- a/tests/load/main/main.go
+++ b/tests/load/main/main.go
@@ -82,7 +82,6 @@ func main() {
 
 	nodes := tmpnet.NewNodesOrPanic(numNodes)
 
-	require.Positive(numWorkersArg, "failed to create positive number of keys")
 	keys, err := tmpnet.NewPrivateKeys(numWorkersArg)
 	require.NoError(err)
 


tests/load/main/main.go
```diff
@@ -59,6 +60,12 @@ func init() {
 		false,
 		"whether to use Firewood in Coreth",
 	)
+	flag.Uint64Var(
+		&numWorkersArg,
+		"num-workers",
+		defaultNodeCount,
```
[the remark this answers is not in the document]
RodrigoVillar: `defaultNodeCount` is defined in the file.

tests/load/main/main.go
```diff
@@ -75,7 +82,7 @@ func main() {
 
 	nodes := tmpnet.NewNodesOrPanic(numNodes)
 
-	keys, err := tmpnet.NewPrivateKeys(numNodes)
+	keys, err := tmpnet.NewPrivateKeys(int(numWorkersArg))
```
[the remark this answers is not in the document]
RodrigoVillar: I prefer to use a type conversion here to maintain the invariant that `numWorkersArg` is nonnegative.

tests/load/main/README.md
```diff
@@ -115,6 +115,7 @@ updated by the generator/wallets. The server is configured to be targeted by
 
 - `--load-timeout`: Maximum duration to run the load test (default: unlimited)
 - `--firewood`: Whether to use Firewood in Coreth (default: false)
+- `--num-workers`: The number of workers to use for the load test (default: 5)
```
joshua-kim, line 118: Unrelated to this PR - but this documentation is overkill for this tooling/is redundant. Users could (and are more likely to) use the `--help` flag since we document parameter usage there.

tests/load/main/main.go
```diff
@@ -59,6 +60,12 @@ func init() {
 		false,
 		"whether to use Firewood in Coreth",
 	)
+	flag.Uint64Var(
```
joshua-kim: Should we use `flag.IntVar` here? in our code we use it as an `int` anyways - so it would avoid the extra cast. My guess is that we used `uint64` to avoid negatives - but I would argue that:
1. Using `uint` to guarantee non-negatives isn't necessarily great in this case - since we use it as an `int` anyways, which can lead to subtle overflow bugs. 
2. If we _really_ needed the validation - we could just validate the parameter before passing it into our business logic
3. We could just panic if it's negative

RodrigoVillar: Switched to using a `int` type and enforcing that the value passed in is positive: 1949b5e

tests/load/main/README.md
RodrigoVillar: I updated the docs to be consistent with previous PRs - open to streamlining the docs in a future issue/PR

tests/load/main/main.go
```diff
@@ -75,7 +82,8 @@ func main() {
 
 	nodes := tmpnet.NewNodesOrPanic(numNodes)
 
-	keys, err := tmpnet.NewPrivateKeys(numNodes)
+	require.Positive(numWorkersArg, "failed to create positive number of keys")
```
joshua-kim: nit: I think this check is overkill. If it's negative/zero IIRC tmpnet will spit out an error if there aren't any prefunded keys

tests/load/main/README.md
joshua-kim: Yeah it's unrelated - I don't mean to bring it up to block this PR but I'm just making an observation on the nearby code.

tests/load/main/main.go
maru-ava: Correct: https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/tmpnet/genesis.go#L62

RodrigoVillar: Removed: 287a0b2

PR #4249, opened 2025-09-05
Use generics in x/sync

alarso16:

### DIFF
--- a/x/sync/db.go
+++ b/x/sync/db.go
@@ -6,17 +6,15 @@ package sync
 import (
 	"context"
 	"encoding"
+	"errors"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/maybe"
-	"github.com/ava-labs/avalanchego/x/merkledb"
 )
 
 var (
-	_ DB[*merkledb.RangeProof, *merkledb.ChangeProof] = (merkledb.MerkleDB)(nil)
-
-	_ Proof = (*merkledb.RangeProof)(nil)
-	_ Proof = (*merkledb.ChangeProof)(nil)
+	ErrStartRootNotFound = errors.New("start root not found")
+	ErrEndRootNotFound   = errors.New("end root not found")
 )
 
 type DB[TRange, TChange Proof] interface {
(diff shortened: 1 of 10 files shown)


x/sync/db.go
```diff
@@ -3,11 +3,118 @@
...
+	GetMerkleRoot(ctx context.Context) (ids.ID, error)
+	Clear() error
+	ChangeProofer[TChange]
+	RangeProofer[TRange]
+}
+
+type PProof[T any] interface {
+	*T
+	Proof
+}
```
demosdemon: This looks a lot better than I expected once you give this interface a name.

x/sync/network_server.go
```diff
@@ -258,11 +258,11 @@ func (g *GetRangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _ t
...
 // TODO improve range proof generation so we don't need to iteratively
 // reduce the key limit.
-func getRangeProof(
+func getRangeProof[TRange, TChange Proof](
 	ctx context.Context,
-	db DB,
+	db DB[TRange, TChange],
 	req *pb.SyncGetRangeProofRequest,
-	marshalFunc func(*merkledb.RangeProof) ([]byte, error),
+	marshalFunc func(TRange) ([]byte, error),
```
demosdemon, line 246: nit: this func param can go away if, in the change proof case, wrapping with SyncGetChangeProofResponse is done after this returns. that's actually an after-effect of changing the proto wrapper but I just noticed.

alarso16: I also thought this. The only thing is what if the byte limit is 1024, and the marshaled range proof has length 1023? Once it's wrapped in a `ChangeProofResponse`, it will be too long

x/sync/db.go
alarso16: Yeah I was surprised too. I think there might still be some readability stuff that could be improved though around this type

x/sync/network_server.go
demosdemon: > I also thought this. The only thing is what if the byte limit is 1024, and the marshaled range proof has length 1023? Once it's wrapped in a `ChangeProofResponse`, it will be too long

imo, the static well known wrapping that we add shouldn't count towards the byte limit since the peer is aware of what it's responding with but not with what you're wrapping the response with.

PR #4251, opened 2025-09-08
Delete gRPC client for MerkleDB

alarso16:
## Why this should be merged

This is completely unused and slows progress in x/sync refactoring for Firewood.

## How this works

Deletes everything required to host `g_db` in proto

## How this was tested

UT

Not sure, maybe?


x/merkledb/proof_test.go
```diff
@@ -1650,12 +1650,12 @@ func FuzzRangeProofProtoMarshalUnmarshal(f *testing.F) {
 		// Marshal and unmarshal it.
 		// Assert the unmarshaled one is the same as the original.
 		var unmarshaledProof RangeProof
-		protoProof := proof.ToProto()
-		require.NoError(unmarshaledProof.UnmarshalProto(protoProof))
+		protoProof := proof.toProto()
```
alarso16: These fuzz tests fail whenever provided empty keys or values, due to encoding/decoding differences comparing `[]byte(nil)` and `[]byte{}`. Is there a standard way to fix this ?

alarso16: Update: there are deeper issues with trying to use `MarshalBinary` and `UnmarshalBinary` in these tests, namely that proto is not deterministic.

alarso16: Test updated to only check deterministic output

proto/sync/sync.proto
```diff
@@ -61,40 +22,7 @@ message SyncGetChangeProofResponse {
...
-}
-
-message CommitChangeProofRequest {
-  ChangeProof proof = 1;
-}
-
-// For use in sync client, which has a restriction on the size of
-// the response. GetRangeProof in the DB service doesn't.
+// Range proof responses don't require any extra data, so no message is needed.
 message SyncGetRangeProofRequest {
```
joshua-kim: Should we rename these to remove the `Sync*` prefix? I think previously we had to disambiguate to prevent a collision with `GetRangeProofRequest`, but since we're deleting it now we don't need to keep the weird `Sync` prefix anymore.

proto/sync/sync.proto
```diff
@@ -2,47 +2,8 @@ syntax = "proto3";
...
-}
-
-message Proof {
-  bytes key = 1;
-  MaybeBytes value = 2;
-  repeated ProofNode proof = 3;
-}
-
 // For use in sync client, which has a restriction on the size of
 // the response. GetChangeProof in the DB service doesn't.
```
joshua-kim: nit: We need to update/delete this comment

x/merkledb/proof.go
```diff
@@ -53,13 +53,13 @@ var (
...
 	ErrInvalidKeyLength              = errors.New("key length doesn't match bytes length, check specified branchFactor")
-	ErrNilRangeProof                 = errors.New("range proof is nil")
-	ErrNilChangeProof                = errors.New("change proof is nil")
 	ErrNilMaybeBytes                 = errors.New("maybe bytes is nil")
 	ErrNilProof                      = errors.New("proof is nil")
 	ErrNilValue                      = errors.New("value is nil")
 	ErrUnexpectedEndProof            = errors.New("end proof should be empty")
 	ErrUnexpectedStartProof          = errors.New("start proof should be empty")
+	errNilRangeProof                 = errors.New("range proof is nil")
+	errNilChangeProof                = errors.New("change proof is nil")
```
joshua-kim: Are these errors possible to hit now?

x/merkledb/proof.go
```diff
@@ -53,13 +53,13 @@ var (
 	ErrNilValueOrHash                = errors.New("proof node's valueOrHash field is nil")
 	ErrNilKey                        = errors.New("key is nil")
 	ErrInvalidKeyLength              = errors.New("key length doesn't match bytes length, check specified branchFactor")
-	ErrNilRangeProof                 = errors.New("range proof is nil")
-	ErrNilChangeProof                = errors.New("change proof is nil")
 	ErrNilMaybeBytes                 = errors.New("maybe bytes is nil")
 	ErrNilProof                      = errors.New("proof is nil")
 	ErrNilValue                      = errors.New("value is nil")
```
joshua-kim: Dead code after this PR

x/merkledb/proof_test.go
```diff
@@ -1746,144 +1702,75 @@ func TestChangeProofUnmarshalProtoNilValue(t *testing.T) {
...
-	) {
-		require := require.New(t)
-		rand := rand.New(rand.NewSource(randSeed)) // #nosec G404
-
-		// Make a random proof.
-		proofLen := rand.Intn(32)
-		proofPath := make([]ProofNode, proofLen)
-		for i := 0; i < proofLen; i++ {
-			proofPath[i] = newRandomProofNode(rand)
+func generateKeyChanges(rand *rand.Rand, numKeyChanges int, includeNone bool) []KeyChange {
```
joshua-kim, line 1724: Do we need this refactor in this PR? If we keep this, it would be nice if we can move this hunk since this is causing git to get confused and detect `TestChangeProofUnmarshalProtoInvalidMaybe` as a delete/add.

x/merkledb/proof_test.go
```diff
@@ -1746,144 +1702,75 @@ func TestChangeProofUnmarshalProtoNilValue(t *testing.T) {
...
-			},
-		},
-	}
-
-	var proof ChangeProof
-	err := proof.UnmarshalProto(protoProof)
-	require.ErrorIs(t, err, ErrInvalidMaybe)
-}
-
-func FuzzProofProtoMarshalUnmarshal(f *testing.F) {
```
joshua-kim, line 1803: Why is this removed?

x/merkledb/proof_test.go
```diff
@@ -1746,144 +1702,75 @@ func TestChangeProofUnmarshalProtoNilValue(t *testing.T) {
...
-		require.NoError(unmarshaledProof.UnmarshalProto(protoProof))
-		require.Equal(proof, unmarshaledProof)
-
-		// Marshaling again should yield same result.
-		protoUnmarshaledProof := unmarshaledProof.ToProto()
-		require.Equal(protoProof, protoUnmarshaledProof)
-	})
-}
-
-func TestProofProtoUnmarshal(t *testing.T) {
```
joshua-kim, line 1850: Why is this removed?

proto/sync/sync.proto
```diff
@@ -2,47 +2,8 @@ syntax = "proto3";
...
 message SyncGetChangeProofRequest {
```
StephenButtolph: All of the messages that start with `Sync` should probably lose the `Sync` prefix. It seems redundant given the name of the proto file.

proto/sync/sync.proto
StephenButtolph, line 1: Can we update the comments on the `MaybeBytes` fields?

tbh, I think we probably don't even need the field at all. The existence of the field should be sufficient to indicate if the bytes are some or nothing.

x/merkledb/proof_test.go
alarso16: Not technically, but the same code was used in like 5 places. I can restore it if you like

x/merkledb/proof.go
alarso16: Not really (it would be a bug in unexported code) - I deleted the check

x/merkledb/proof_test.go
alarso16: I deleted the code it's testing

x/merkledb/proof_test.go
alarso16: I deleted the code it's testing

proto/sync/sync.proto
alarso16: No we do need the field I think. We need to distinguish between the empty key and no key (or we need a non-trivial functional change)

x/merkledb/proof_test.go
```diff
@@ -1485,7 +1486,7 @@ func TestProofNodeUnmarshalProtoInvalidMaybe(t *testing.T) {
 	rand := rand.New(rand.NewSource(now)) // #nosec G404
 
 	node := newRandomProofNode(rand)
-	protoNode := node.ToProto()
+	protoNode := node.toProto()
```
alarso16, line 1489: These are now testing unexported code... I could delete the tests and the proper operation would be tested in range/change proof tests, but this does check a lot of internal edge cases

x/merkledb/proof_test.go
alarso16: There were changes needed in each place, specifically regarding `[]byte{}` vs. `nil` unmarshaling

x/merkledb/proof_test.go
joshua-kim: I don't think we care too much here - we're not going to maintain merkledb since we're using firewood in the long-term anyways.

joshua-kim (approved):

StephenButtolph (approved):

proto/sync/sync.proto
StephenButtolph: We don't need the field. We can use the presence of the `MaybeBytes` field instead (this is because message types are implicitly optional in proto)

PR #4259, opened 2025-09-09
Remove unnecessary field from MaybeBytes

### LINKED ISSUE #4256: Remove exists field from `MaybeBytes`
Can we update the comments on the `MaybeBytes` fields?

tbh, I think we probably don't even need the field at all. The existence of the field should be sufficient to indicate if the bytes are some or nothing.

_Originally posted by @StephenButtolph in https://github.com/ava-labs/avalanchego/pull/4251#discussion_r2334054023_

alarso16:
## Why this should be merged

Redundant field, closes #4256

## How this works

Use nil as Nothing.

## How this was tested

Existing UT

No

### DIFF
--- a/proto/sync/sync.proto
+++ b/proto/sync/sync.proto
@@ -58,7 +58,9 @@ message Key {
 }
 
 // If the message is present, the value is considered to be "something".
-message MaybeBytes { bytes value = 1; }
+message MaybeBytes {
+  bytes value = 1;
+}
 
 message KeyValue {
   bytes key = 1;
--- a/proto/pb/sync/sync.pb.go
+++ b/proto/pb/sync/sync.pb.go
@@ -548,7 +548,7 @@ func (x *Key) GetValue() []byte {
 	return nil
 }
 
-// If the message is empty, the value is considered to be "nothing".
+// If the message is present, the value is considered to be "something".
 type MaybeBytes struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	Value         []byte                 `protobuf:"bytes,1,opt,name=value,proto3" json:"value,omitempty"`


x/sync/network_server.go
```diff
@@ -53,10 +51,19 @@ var (
...
 func maybeBytesToMaybe(mb *pb.MaybeBytes) maybe.Maybe[[]byte] {
-	if mb != nil && !mb.IsNothing {
-		return maybe.Some(mb.Value)
+	if mb == nil {
+		return maybe.Nothing[[]byte]()
+	}
+	return maybe.Some(mb.Value)
+}
+
+func maybeToMaybeBytes(m maybe.Maybe[[]byte]) *pb.MaybeBytes {
```
alarso16: This is used in `manager.go`, but the name sucks. let me know if you have a less confusing name

StephenButtolph: `maybeToProto` and `protoToMaybe`?

proto/sync/sync.proto
```diff
@@ -57,9 +57,9 @@ message Key {
   bytes value = 2;
 }
 
+// If the message is empty, the value is considered to be "nothing".
```
StephenButtolph: I think the term normally used is [presence](https://protobuf.dev/programming-guides/field_presence/)
```suggestion
// If the message is present, the value is considered to be "something".
```

x/merkledb/proof.go
```diff
@@ -472,16 +473,8 @@ func (c *ChangeProof) unmarshalProto(pbProof *pb.ChangeProof) error {
...
-
-		if kv.Value.IsNothing && len(kv.Value.Value) != 0 {
-			return errInvalidMaybe
-		}
-
 		value := maybe.Nothing[[]byte]()
-		if !kv.Value.IsNothing {
+		if kv.Value != nil {
 			value = maybe.Some(kv.Value.Value)
 		}
```
StephenButtolph: Can we use the `maybeBytesToMaybe` helper?

x/merkledb/proof.go
```diff
@@ -439,12 +436,16 @@ func (c *ChangeProof) toProto() *pb.ChangeProof {
 
 	keyChanges := make([]*pb.KeyChange, len(c.KeyChanges))
 	for i, kv := range c.KeyChanges {
+		var value *pb.MaybeBytes
+		if kv.Value.HasValue() {
+			value = &pb.MaybeBytes{
+				Value: kv.Value.Value(),
+			}
+		}
```
StephenButtolph: Similarly, I think we should use the helper `maybeToMaybeBytes`

x/merkledb/proof.go
```diff
@@ -115,7 +112,7 @@ func (node *ProofNode) unmarshalProto(pbNode *pb.ProofNode) error {
 		node.Children[byte(childIndex)] = childID
 	}
 
-	if !pbNode.ValueOrHash.IsNothing {
+	if pbNode.ValueOrHash != nil {
 		node.ValueOrHash = maybe.Some(pbNode.ValueOrHash.Value)
 	}
```
StephenButtolph: `maybeBytesToMaybe`?

x/merkledb/proof.go
```diff
@@ -70,16 +67,20 @@ type ProofNode struct {
 // toProto converts the ProofNode into the protobuf version of a proof node
 // Assumes [node.Key.Key.length] <= math.MaxUint64.
 func (node *ProofNode) toProto() *pb.ProofNode {
+	var valueOrHash *pb.MaybeBytes
+	if node.ValueOrHash.HasValue() {
+		valueOrHash = &pb.MaybeBytes{
+			Value: node.ValueOrHash.Value(),
+		}
+	}
```
StephenButtolph: `maybeToMaybeBytes `?

x/merkledb/proof.go
alarso16: It's in the `sync` package, seems like we shouldn't import this. I could recreate them in `merkledb` though

StephenButtolph: Is there some place we could put them and export them?

x/sync/protoutils/utils.go
```diff
@@ -0,0 +1,26 @@
...
+		Value: m.Value(),
+	}
+}
+
+func ProtoToMaybe(mb *pb.MaybeBytes) maybe.Maybe[[]byte] {
+	if mb == nil {
+		return maybe.Nothing[[]byte]()
+	}
+	return maybe.Some(mb.Value)
+}
```
joshua-kim, line 26: The cost of copy-paste here isn't bad and avoids the awkward utils package (which typically ends up becoming a trashcan package without a clear abstraction boundary). Even if the cost of copy and paste was worse, we're not maintaining merkledb long-term anyways.

StephenButtolph: I feel like this aligns reasonably with `grpcutils`

joshua-kim (approved):

PR #4262, opened 2025-09-11
[tmpnet] Add optional stack traces to errors originating from tmpnet

maru-ava:
## Why this should be merged

Setting `STACK_TRACE_ERRORS=1` will configure tmpnet to include stack traces with errors it originates. This is intended to aid in debugging by indicating not just the location of the failure but also the chain of callers.

Inspired by: https://github.com/golang/go/issues/63358

## How this works

- Add stacktrace package and ensure all tmpnet errors pass through its functions
- Said functions ensure that the stack trace is collected at the deepest point in the call chain for maximum traceability

## How this was tested

- [x] CI
- [x] tmpnet default behavior
```
$ tmpnetctl start-network --node-count=0
Error: --node-count must be greater than 0 but got 0
```
- [x] tmpnet with stacktraces
```
$ STACK_TRACE_ERRORS=1 tmpnetctl start-network --node-count=0
Error: --node-count must be greater than 0 but got 0
Stack trace:
/home/user/work/avalanchego/a_tmpnet-stacktrace/tests/fixture/tmpnet/tmpnetctl/main.go:76: main.main.func2
/home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985: github.com/spf13/cobra.(*Command).execute
/home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117: github.com/spf13/cobra.(*Command).ExecuteC
/home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041: github.com/spf13/cobra.(*Command).Execute
/home/user/work/avalanchego/a_tmpnet-reuse-state/tests/fixture/tmpnet/tmpnetctl/main.go:322: main.main
/home/user/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.9.linux-arm64/src/runtime/proc.go:272: runtime.main
/home/user/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.9.linux-arm64/src/runtime/asm_arm64.s:1223: runtime.goexit
```

N/A

### DIFF
--- a/tests/fixture/stacktrace/stacktrace.go
+++ b/tests/fixture/stacktrace/stacktrace.go
@@ -60,7 +60,7 @@ func New(msg string) error {
 
 // Errorf adds a stack trace to the last argument provided if it is an
 // error and stack traces are enabled.
-func Errorf(format string, args ...interface{}) error {
+func Errorf(format string, args ...any) error {
 	if !stackTraceErrors {
 		return fmt.Errorf(format, args...)
 	}


tests/fixture/tmpnet/node.go
```diff
@@ -420,7 +421,7 @@ func (n *Node) WaitForHealthy(ctx context.Context) error {
 		healthy, err := n.IsHealthy(ctx)
 		switch {
 		case errors.Is(err, ErrUnrecoverableNodeHealthCheck):
-			return fmt.Errorf("%w for node %q", err, n.NodeID)
+			return stacktrace.Errorf("node %q saw unrecoverable health check: %w", n.NodeID, err)
```
maru-ava, line 424: Needed to reorder the args to ensure the error was last so that the stacktrace could be added to it.

tests/fixture/tmpnet/tmpnetctl/main.go
```diff
@@ -289,12 +290,12 @@ func main() {
...
+				return stacktrace.Wrap(err)
 			}
 			// A valid kubeconfig is required for local kind usage but this is not validated by KubeconfigVars
 			// since unlike kind, tmpnet usage may involve an implicit in-cluster config.
 			if len(kubeconfigVars.Path) == 0 {
-				return errKubeconfigRequired
+				return stacktrace.Wrap(errKubeconfigRequired)
 			}
 			// TODO(marun) Consider supporting other contexts. Will require modifying the kind cluster start script.
 			if len(kubeconfigVars.Context) > 0 && kubeconfigVars.Context != tmpnet.KindKubeconfigContext {
```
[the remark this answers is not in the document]
maru-ava: Talk about missing the forest for the trees...

tests/fixture/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,128 @@
...
+func New(msg string) error {
+	if !stackTraceErrors {
+		return errors.New(msg)
+	}
+	return wrap(errors.New(msg))
+}
+
+// Errorf adds a stack trace to the last argument provided if it is an
+// error and stack traces are enabled.
+func Errorf(format string, args ...interface{}) error {
```
RodrigoVillar:
```suggestion
func Errorf(format string, args ...any) error {
```
nit

joshua-kim:
> but don't you sort of get the same result if you fmt the error before returning it? ex.

It's is similar, but you just get a long error string and have to reconstruct the call stack by searching for error messages (e.g `failed to foo: failed to bar: failed to baz`). It also doesn't handle deduping, so if you have similar error messages in multiple locations you might have some ambiguity. If you want a true stacktrace dump (like the ones that other languages support), to my understanding you just have to implement them yourself and this is considered idiomatic although it may look odd at first glance.

tests/fixture/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,128 @@
...
+// - `fmt.Errorf` -> `stacktrace.Errorf`
+// - `errors.New` -> `stacktrace.New`
+//
+// Additionally, a stack trace can be added to an existing error with
+// `stacktrace.Wrap(err)`.
+
+var stackTraceErrors bool
+
+func init() {
+	if os.Getenv("STACK_TRACE_ERRORS") == "1" {
```
joshua-kim, line 27: I don't know if tmpnet has other env vars it supports, but should we prefix them with something like `TMPNET_` or `TMPNET_DEBUG_`?

tests/fixture/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,128 @@
...
+
+	for {
+		frame, more := frames.Next()
+		frameSlice = append(frameSlice, frame)
+		if !more {
+			break
+		}
+	}
+
+	return StackTraceError{
```
joshua-kim, line 124: Do we have to export this type? It seems like we only use this type as the `error` interface downstream anyways.

maru-ava: I prefer to err on the side of exporting when it comes to library functionality in support of testing to minimize friction with downstream repos like coreth and subnet-evm.

tests/fixture/stacktrace/stacktrace.go
maru-ava: I intentionally avoided prefixing with `TMPNET` because it's not a tmpnet-specific thing. tmpnet is only the first adopter, but there's no reason for this library to be restricted to it (hence not putting it under `tests/fixture/tmpnet`).

maru-ava: Not that I think this env var is ideal, just that I think it's good enough for now.

PR #4267, opened 2025-09-11
Add public network support to tmpnet

felipemadero:
## Why this should be merged
This change enables tmpnet to work with public networks (Fuji and Mainnet) by properly handling their configuration differences. Previously, tmpnet would attempt to configure bootstrap nodes for all networks, but public networks should use avalanchego's built-in bootstrappers instead of custom ones. Additionally, the NetworkID field was missing from persistence, causing configuration loss across sessions.
  
## How this works
- Adds NetworkID field to the serialized network configuration  for proper persistence across tmpnet sessions
- Modifies bootstrap configuration logic to skip custom bootstrap setup for public networks, allowing them to use avalanchego's built-in bootstrappers
  
## How this was tested
CLI L1 fuji validators on local machine

### DIFF
--- a/tests/fixture/tmpnet/network_config.go
+++ b/tests/fixture/tmpnet/network_config.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 
 	"github.com/ava-labs/avalanchego/genesis"
+	"github.com/ava-labs/avalanchego/tests/fixture/stacktrace"
 	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
 	"github.com/ava-labs/avalanchego/utils/perms"
 )
@@ -24,10 +25,10 @@ var errMissingNetworkDir = errors.New("failed to write network: missing network 
 // Read network and node configuration from disk.
 func (n *Network) Read(ctx context.Context) error {
 	if err := n.readNetwork(); err != nil {
-		return err
+		return stacktrace.Wrap(err)
 	}
 	if err := n.readNodes(ctx); err != nil {
-		return err
+		return stacktrace.Wrap(err)
 	}
 	return n.readSubnets()
 }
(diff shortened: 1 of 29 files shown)


tests/fixture/tmpnet/node.go
```diff
@@ -21,13 +21,19 @@ import (
...
 	"github.com/ava-labs/avalanchego/staking"
+	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
 	"github.com/ava-labs/avalanchego/vms/platformvm/signer"
 )
 
 // The Node type is defined in this file (node.go - orchestration) and
 // node_config.go (reading/writing configuration).
 
+// isPublicNetwork returns true if the given network ID corresponds to a public network
```
maru-ava: (No action required) Maybe inline this given that it's only used in one place?

tests/fixture/tmpnet/network_config.go
```diff
@@ -153,6 +153,7 @@ func (n *Network) readConfig() error {
 type serializedNetworkConfig struct {
 	UUID                 string                  `json:"uuid,omitempty"`
 	Owner                string                  `json:"owner,omitempty"`
+	NetworkID            uint32                  `json:"networkID,omitempty"`
```
maru-ava, line 157: Good catch!

tests/fixture/tmpnet/node.go
joshua-kim: I think in-lining looks better as well here

felipemadero: done

joshua-kim:
Looks like we need to re-sign, https://github.com/ava-labs/avalanchego/pull/4267/commits/8684f8c8909651696e0b53dbe95eb3f19e72e90e is missing a signature

felipemadero:
Replaced by https://github.com/ava-labs/avalanchego/pull/4280 due to commit verification issues

PR #4272, opened 2025-09-12
test(reexecute/c): disable metrics collection by default

RodrigoVillar:
## Why this should be merged

As mentioned in offline conversations and in #4214, the metrics collection feature of the re-execution test should be disabled by default as it requires setting up credentials. 

## How this works

Sets metrics collection in the re-execution test to disabled by default, with CI jobs enabling this feature.

## How this was tested

Successful run of re-execution test workflow with metrics enabled: https://github.com/ava-labs/avalanchego/actions/runs/17679308849/job/50249347146?pr=4272#step:3:791

No

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -78,7 +78,8 @@ runs:
         role-to-assume: ${{ inputs.aws-role }}
         aws-region: ${{ inputs.aws-region }}
         role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
-    - name: Run C-Chain Re-Execution
+    - name: Run C-Chain Re-Execution (with monitoring)
+      if: ${{ inputs.metrics-enabled == 'true' }}
       uses: ./.github/actions/run-monitored-tmpnet-cmd
       with:
         run_env: |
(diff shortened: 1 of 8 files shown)


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -87,6 +91,7 @@ runs:
             LABELS=${{ env.LABELS }} \
             BENCHMARK_OUTPUT_FILE=${{ env.BENCHMARK_OUTPUT_FILE }} \
             RUNNER_NAME=${{ inputs.runner_name }}
+            METRICS_ENABLED=${{ env.METRICS_ENABLED }}
         run: ./scripts/run_task.sh reexecute-cchain-range-with-copied-data
 
         prometheus_push_url: ${{ inputs.prometheus-push-url }}
```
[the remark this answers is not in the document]
RodrigoVillar: Addressed: cb8f0e1

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
```diff
@@ -35,6 +35,9 @@ on:
         description: 'Timeout in minutes for the job.'
         required: false
         default: 30
+      metrics-enabled:
+        description: "Whether to enable metrics collection for the benchmark."
+        default: false
```
maru-ava: (No action required) Why would this default to false? I get the intention of not monitoring by default to simplify local usage, but in CI I think we want to always configure monitoring unless the repo secrets are absent (i.e. due to a job executing against a fork branch).

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -95,6 +101,23 @@ runs:
         grafana_dashboard_id: 'Gl1I20mnk/c-chain'
         runtime: "" # Set runtime input to empty string to disable log collection
 
+    - name: Run C-Chain Re-Execution (without monitoring)
```
maru-ava: (No action required) What's the use case for testing in CI without monitoring?

.github/workflows/c-chain-reexecution-benchmark-container.json
```diff
@@ -8,7 +8,8 @@
                 "end-block": 250000,
                 "block-dir-src": "s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**",
                 "current-state-dir-src": "s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100/**",
-                "timeout-minutes": 30
+                "timeout-minutes": 30,
+                "metrics-enabled": true
```
maru-ava: Maybe avoid specifying this field outside of the custom action given that every instance in configuration is `true`?

RodrigoVillar: Done: c832b80

.github/actions/c-chain-reexecution-benchmark/action.yml
RodrigoVillar: Removed

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
RodrigoVillar: Reverted to setting monitoring to always true.

PR #4274, opened 2025-09-12
Add Mempool to XSVM

samliok:
## Why this should be merged

Adding a gossipable mempool to the XSVM whcih allows for mempool transactions to be gossiped to other nodes. This is useful for consensus protocols like simplex, where block proposers can only propose during their proposal window. Gossiping allows block proposers to listen and propose blocks with transactions that might not have been sent to them.


yacovm:
1. Why not just send the transaction to the right node in the test? I think it might make things simpler. We always know which node is the leader in the test since we can get the last block and compute that. 
2. What is the mechanism that evicts the transaction from the mempool once a block is notarized? Without such a mechanism we would send a transaction once and we would have `n` blocks proposed.

PR #2509, opened 2023-12-19
Add gossip Marshaller interface

joshua-kim:
## Why this should be merged

Required to unblock the X-Chain's integration for the sdk gossip

## How this works

Refactors marshal and unmarshal to be on its own interface instead of attached to the `Gossipable` interface

## How this was tested

UTs pass

PR #4275, opened 2025-09-12
Add generics to x/sync

alarso16:
## Why this should be merged

The `x/sync` package shouldn't have to import `merkledb` at all. This moves everything that instantiates a `MerkleDB` to a subpackage, avoiding circular dependencies allowing a sentinel error in x/sync with generics.

## How this works

Moves everything testing for merkledb to a different package. Some tests were semantically changed

## How this was tested

UT

No.


x/sync/merkledb_test/sync_test.go
```diff
@@ -899,36 +789,28 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
...
 
-			return syncer.processedWork.Len() > 0
-		},
-		5*time.Second,
-		5*time.Millisecond,
-	)
+	// Start syncing from the server, will be cancelled by the range proof handler
+	require.NoError(syncer.Start(cancelCtx))
+	err = syncer.Wait(cancelCtx)
+	require.ErrorIs(err, context.Canceled)
```
rrazvan1: replacing the old `require.Eventually` usage with `newRangeProofHandlerCancel`, isnt the same. With your approach, the test isnt 100% that there is at least one job processed.

I think this could work:
```suggestion
	require.NoError(syncer.Start(ctx))

	// Wait until we've processed some work before updating the sync target.
	require.Eventually(func() bool {
		return db.NewIterator().Next()
	}, 5*time.Second, 5*time.Millisecond)
```

x/sync/merkledb_test/sync_test.go
```diff
@@ -945,155 +827,52 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
 	require.Equal(syncRoot, newRoot)
 }
 
-func Test_Sync_Result_Correct_Root_Update_Root_During(t *testing.T) {
-	t.Skip("FLAKY")
-
+func Test_Sync_UpdateSyncTarget(t *testing.T) {
```
rrazvan1: I think this is out of scope. Keep this test skipped or commented if it has any dependency on private fields, and lets fix it in other PR in the proper way.

x/sync/merkledb_test/client_test.go
```diff
@@ -113,6 +114,23 @@ func newFlakyChangeProofHandler(
 	}
 }
 
+type p2pHandlerCancel struct {
+	p2p.Handler
+	action func()
+}
+
+func newRangeProofHandlerCancel(db merkledb.MerkleDB, action func()) *p2pHandlerCancel {
```
rrazvan1: p2pHandler**Cancel**?
Why does it have cancel in name? It accepts any function as action.

x/sync/merkledb_test/sync_test.go
```diff
@@ -37,167 +40,54 @@ func Test_Creation(t *testing.T) {
...
-			start, end = end, start
-		}
-
-		mid = midPoint(maybe.Some(start), maybe.Some(end))
-		require.Equal(-1, bytes.Compare(start, mid.Value()))
-		require.Equal(-1, bytes.Compare(mid.Value(), end))
-	}
 }
 
 func Test_Sync_FindNextKey_InSync(t *testing.T) {
```
rrazvan1: After **findKey** has been removed from manager, this test doesn't make sense anymore.

rrazvan1: This is applicable for every **findKey** test from this file.

rrazvan1: and by deleting/moving them, I think we can make **midpoint** function private again.

x/sync/merkledb_test/client_test.go
joshua-kim, line 1: Would it be fine for `merkledb` to just depend on `x/sync`, instead of making this new testing package? If `x/sync` depends on `merkledb` something is wrong/we should refactor, I think that ideally `x/sync` should define an interface for the database that can be sync'd, and `merkledb` should implement that interface

x/sync/manager.go
```diff
@@ -790,7 +790,7 @@ func (m *Manager) enqueueWork(work *workItem) {
 // start is expected to be less than end
 // Nothing/nil [start] is treated as all 0's
 // Nothing/nil [end] is treated as all 255's
-func midPoint(startMaybe, endMaybe maybe.Maybe[[]byte]) maybe.Maybe[[]byte] {
```
joshua-kim, line 793: Why do we export this?

x/sync/merkledb_test/sync_test.go
alarso16: You're right, these tests really belong in `merkledb`, but I can't move them there yet due to cyclic dependency issues around generating random tries.

x/sync/merkledb_test/client_test.go
alarso16: The next PR #4249 removes the cyclic dependency (since `x/sync` still imports a sentinel error and all the proof structs), so I can move this either in that PR or a separate PR after that

x/sync/manager.go
alarso16: For testing, the midpoint construction is pretty complicated for repetitive code. @rrazvan1 already suggested I revise the test to remove the export

x/sync/merkledb_test/sync_test.go
alarso16: This seems to work, I'll go with it

x/sync/merkledb_test/client_test.go
alarso16: Fixed!

x/sync/merkledb_test/sync_test.go
alarso16: Ok moved - would you prefer in a different file?

x/sync/merkledb_test/client_test.go
alarso16: Moved into merkledb

x/sync/manager.go
alarso16: Unexported!

x/sync/db.go
```diff
@@ -3,12 +3,112 @@
...
+type Proof interface {
+	encoding.BinaryMarshaler
+	encoding.BinaryUnmarshaler
+}
+
+type ChangeProofer[T Proof] interface {
+	// GetChangeProof returns a proof for a subset of the key/value changes in key range
+	// [start, end] that occurred between [startRootID] and [endRootID].
+	// Returns at most [maxLength] key/value pairs.
+	// Returns [ErrInsufficientHistory] if this node has insufficient history
```
rrazvan1: `ErrInsufficientHistory` doesnt exist anymore

x/sync/db.go
```diff
@@ -3,12 +3,112 @@
...
+	// to generate the proof.
+	// Returns ErrEmptyProof if [endRootID] is ids.Empty.
+	// Note that [endRootID] == ids.Empty means the trie is empty
+	// (i.e. we don't need a change proof.)
+	// Returns [ErrNoEndRoot], which wraps [ErrInsufficientHistory], if the
```
rrazvan1: same about `ErrInsufficientHistory`

x/merkledb/history.go
```diff
@@ -15,11 +14,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/buffer"
 	"github.com/ava-labs/avalanchego/utils/heap"
 	"github.com/ava-labs/avalanchego/utils/maybe"
-)
-
-var (
-	ErrInsufficientHistory = errors.New("insufficient history to generate proof")
-	ErrNoEndRoot           = fmt.Errorf("%w: end root not found", ErrInsufficientHistory)
```
rrazvan1, line 22: replace all doc comments: `ErrNoEndRoot` -> `ErrEndRootNotFound`

x/merkledb/history.go
```diff
@@ -144,15 +139,15 @@ func (th *trieHistory) getValueChanges(
...
+		return nil, fmt.Errorf("%w: %s", sync.ErrEndRootNotFound, endRoot)
 	}
 
 	// Confirm there's a change resulting in [startRoot] before
 	// a change resulting in [endRoot] in the history.
 	// [startRootChanges] is the last appearance of [startRoot].
 	startRootChanges, ok := th.getRootChanges(startRoot)
 	if !ok {
-		return nil, fmt.Errorf("%w: start root %s not found", ErrInsufficientHistory, startRoot)
+		return nil, fmt.Errorf("%w: start root %s not found", sync.ErrStartRootNotFound, startRoot)
```
rrazvan1:
```suggestion
		return nil, fmt.Errorf("%w: %s", sync.ErrStartRootNotFound, startRoot)
```

x/merkledb/history.go
```diff
@@ -15,11 +14,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/buffer"
 	"github.com/ava-labs/avalanchego/utils/heap"
 	"github.com/ava-labs/avalanchego/utils/maybe"
-)
-
-var (
-	ErrInsufficientHistory = errors.New("insufficient history to generate proof")
```
rrazvan1, line 21: remove `ErrInsufficientHistory` from doc comments

x/merkledb/proof.go
```diff
@@ -30,6 +31,8 @@ var (
 	_ encoding.BinaryUnmarshaler = (*ChangeProof)(nil)
```
rrazvan1: Can we remove all `encoding.BinaryUnmarshaler` and `encoding.BinaryMarshaler` lines since `sync.Proof` includes them?

x/merkledb/sync_test.go
```diff
@@ -235,7 +107,7 @@ func Test_Sync_FindNextKey_InSync(t *testing.T) {
 func Test_Sync_FindNextKey_Deleted(t *testing.T) {
```
rrazvan1: what do you think about moving all `findnextkey` tests to `db_test.go`?
Those tests don't have any reference to **xsync** package, so they can be extracted.

alarso16: Sure - I was avoiding too big of a diff (since there were some small changes), but I can do that

x/sync/db.go
```diff
@@ -3,12 +3,111 @@
...
+	VerifyChangeProof(
+		ctx context.Context,
+		proof T,
+		start maybe.Maybe[[]byte],
+		end maybe.Maybe[[]byte],
+		expectedEndRootID ids.ID,
+		maxLength int,
+	) error
+
+	// CommitChangeProof commits the key/value pairs within the [proof] to the db.
```
rrazvan1, line 70: what do you think about explaining return value (nextKey)?
Also applicable  for `CommitRangeProof`

x/merkledb/client_test.go
```diff
@@ -113,6 +117,16 @@ func newFlakyChangeProofHandler(
...
+type p2pHandlerAction struct {
+	p2p.Handler
+	action func()
+}
+
+func (h *p2pHandlerAction) AppRequest(ctx context.Context, id ids.NodeID, time time.Time, requestBytes []byte) ([]byte, *common.AppError) {
+	h.action()
+	return h.Handler.AppRequest(ctx, id, time, requestBytes)
+}
+
```
joshua-kim: 1. I don't think we need this primitive (a handler w/ a custom function), since we already have `p2p.TestHandler` which implements the same thing
2. Why are we updating the tests to use this instead of `waitingHandler`?

x/merkledb/proof_test.go
```diff
@@ -820,8 +821,7 @@ func Test_ChangeProof_Missing_History_For_EndRoot(t *testing.T) {
 		maybe.Nothing[[]byte](),
 		50,
 	)
-	require.ErrorIs(err, ErrNoEndRoot)
-	require.ErrorIs(err, ErrInsufficientHistory)
```
joshua-kim, line 824: Why are we updating the behavior of the tests in this PR? Should we undo the rename of the errors and keep the tests checking for both errors being wrapped?

x/sync/client_test.go
```diff
@@ -142,14 +156,3 @@ func (c *counter) Inc() int {
 	c.i++
 	return result
 }
-
-type waitingHandler struct {
```
joshua-kim, line 146: Why are we removing this type + updating tests to not use this?

x/sync/db.go
```diff
@@ -3,12 +3,111 @@
...
+	ErrStartRootNotFound = errors.New("start root not found")
+	ErrEndRootNotFound   = errors.New("end root not found")
+)
+
+type DB[TRange, TChange Proof] interface {
+	GetMerkleRoot(context.Context) (ids.ID, error)
+	Clear() error
+	ChangeProofer[TChange]
+	RangeProofer[TRange]
+}
```
joshua-kim: 1. Our repo typically avoids the pattern of `f1, f2 Foo` and prefers `f1 Foo, f2 Foo` (I don't personally have anything against this style you're using, but it is the style the team uses).
2. Typically, interfaces when composed live at the top of the interface
```
type Foo interface {
    Bar // some interface
    Baz()
}
```
3. Why do we decompose the interface into `ChangeProofer` and `RangeProofer`? It doesn't seem like we use these smaller interfaces anywhere so it feels premature - could we get rid of them and couple them under `DB` for now?
4. We added nice godocs for `ChangeProofer` and `RangeProofer` - it'd be great if we could do the same for `GetMerkleRoot` and `Clear`.

x/sync/db.go
```diff
@@ -3,12 +3,111 @@
...
+type ChangeProofer[T Proof] interface {
+	// GetChangeProof returns a proof for a subset of the key/value changes in key range
+	// [start, end] that occurred between [startRootID] and [endRootID].
+	// Returns at most [maxLength] key/value pairs.
+	// Returns [xsync.ErrStartRootNotFound] if this node has insufficient
+	// history to generate the proof.
+	// Returns ErrEmptyProof if [endRootID] is ids.Empty.
+	// Note that [endRootID] == ids.Empty means the trie is empty
+	// (i.e. we don't need a change proof.)
+	// Returns [xsync.ErrEndRootNotFound], if the history doesn't contain the [endRootID].
```
joshua-kim: I think this won't make a proper godoc link (atleast in my ide) since this file does not alias `sync` to `xsync` (so this should just use `sync`).

x/sync/manager.go
```diff
@@ -94,7 +94,10 @@ func newWorkItem(localRootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[
 	}
 }
 
-type Manager struct {
+type Manager[R, C Proof] struct {
```
joshua-kim: I don't have a strong preference on the generic name we use (I'm seeing `R` and `TRange`), but we should be consistent with whatever we end up choosing.

x/sync/manager.go
```diff
@@ -790,7 +797,7 @@ func (m *Manager) enqueueWork(work *workItem) {
 // start is expected to be less than end
 // Nothing/nil [start] is treated as all 0's
 // Nothing/nil [end] is treated as all 255's
-func midPoint(startMaybe, endMaybe maybe.Maybe[[]byte]) maybe.Maybe[[]byte] {
+func midpoint(startMaybe, endMaybe maybe.Maybe[[]byte]) maybe.Maybe[[]byte] {
```
joshua-kim: nit: undo this rename

x/sync/manager.go
```diff
@@ -867,3 +874,9 @@ func calculateBackoff(attempt int) time.Duration {
 		maxRetryWait,
 	)
 }
+
+// Assumes T is a pointer to a struct.
+func allocateProof[T any]() T {
+	et := reflect.TypeOf((*T)(nil)).Elem()
+	return reflect.New(et.Elem()).Interface().(T)
+}
```
joshua-kim: It looks like you're doing this reflection hack because you can't use `*new(T)` since your `T` you're trying to define is a pointer to a type (so you'll try to dereference a `nil` pointer if you try to use `*new(T)`), but you need to do this with the current `Proof` interface since `Marshal` and `Unmarshal` live on `T` (so you have a chicken-egg problem). There is a hack you can do with generics (see the `GossipableAny` type at https://github.com/ava-labs/avalanchego/pull/2509/files) that I implemented previously, but just making a separate `Marshaler` interface is probably the simplest way to accomplish this without reflection/complex usage of generics.

x/sync/manager_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"bytes"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/utils/maybe"
+)
+
+func Test_Midpoint(t *testing.T) {
```
joshua-kim, line 16: This move seems unnecessary for this diff now

x/sync/network_server.go
```diff
@@ -114,14 +111,6 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 					Message: fmt.Sprintf("failed to get change proof: %s", err),
 				}
 			}
-			if errors.Is(err, merkledb.ErrNoEndRoot) {
```
joshua-kim, line 117: Why is this removed?

x/sync/network_server.go
```diff
@@ -183,7 +172,7 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
 		}
 
 		// The proof was too large. Try to shrink it.
-		keyLimit = uint32(len(changeProof.KeyChanges)) / 2
```
joshua-kim, line 186: This looks like a behavior change?

x/sync/network_server.go
```diff
@@ -250,11 +239,11 @@ func (g *GetRangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _ t
...
 // TODO improve range proof generation so we don't need to iteratively
 // reduce the key limit.
-func getRangeProof(
+func getRangeProof[TRange, TChange Proof](
 	ctx context.Context,
-	db DB,
+	db DB[TRange, TChange],
 	req *pb.GetRangeProofRequest,
-	marshalFunc func(*merkledb.RangeProof) ([]byte, error),
+	marshalFunc func(TRange) ([]byte, error),
```
joshua-kim: I'm not sure when this became possible (or if it always was), but it looks like we don't need to use this `marshalFunc` anymore since we have marshal defined on the interface of our generic type.

x/merkledb/proof_test.go
alarso16: I could undo the rename and move the errors to `xsync`, but they can't remain in `merkledb`, since `network_server.go` uses them to determine whether to drop a range proof request or provide a range proof instead of a change proof

x/sync/manager.go
alarso16: Yeah I originally had it like your `GossipableAny`, but that meant 4 generics, which felt unnecessary. You suggest the Marshaler to be separate? So there will be a `RangeProofMarshaler` and a `ChangeProofMarshaler`?

x/sync/network_server.go
alarso16: The reason it exists is due to the change proof wrapping, but I find it unintuitive anyway - I can refactor it if that diff isn't out of scope

x/sync/network_server.go
alarso16: `merkledb.ErrNoEndRoot = fmt.Errorf("%w: no end root", merkledb.InsufficientHistory)`

So you had to check that you returned an insufficient history error, but also that you have the end root. it's super unintuitive, which is why I made the change

x/sync/network_server.go
alarso16: Yeah slightly. I didn't want to expose the size of the proof (since the server doesn't really need to know anyway)

x/sync/manager.go
joshua-kim: I _think_ you only need to specify two generics (the range/change proof types) and the helper "any" type can be inferred by the compiler - but I'm not positive if that's true or not. Either way, yeah we can have a separate marshaler interface decoupled from the data. We could either have a single `Marshaler` interface or have a separate one per type... I don't have a strong preference for or against either one.

x/sync/network_server.go
joshua-kim: Does it have to be done as part of this PR? It seems unrelated to the scope of the change we're making.

x/sync/network_server.go
joshua-kim: I'm okay with scoping in refactors when we're already re-writing/touching this part of the file anyways - so I'm fine scoping it in here.

x/sync/network_server.go
alarso16: I have to at least move these errors to this package (cyclic dependencies), but we can keep the errors

x/sync/manager.go
alarso16: I went somewhere kind of in the middle - let me know what you think. I think I still prefer the `***Any` strategy, but it seems like you prefer the Marshaler

x/sync/network_server.go
alarso16: I refactored it locally, and it didn't end up being any simpler due to the byte size handling. Say the range proof is of size `MaxBytesLimit` but we're returning the change proof, the packet size will be `MaxBytesLimit + responseOverhead`, from wrapping it in another proto object. It's certainly not more than a couple bytes, but I don't think refactoring this makes the code any simpler

x/merkledb/client_test.go
alarso16: 1. Didn't know about the TestHandler, I'll use that
2. The use of the waitingHandler is unintuitive and relied on access to internal fields in the Manager

alarso16: 1. Didn't know that existed! Updated to use this
2. It's not really "instead of", the tests were just not well made to start with... To use the waitingHandler intuitively required peeking into the implementation (which is no longer possible since the tests moved packages).

x/sync/client_test.go
alarso16: Addressed above

x/merkledb/sync_test.go
```diff
@@ -0,0 +1,582 @@
...
+	require.NotNil(syncer)
+
+	// Start syncing from the server, will be cancelled by the range proof handler
+	require.NoError(syncer.Start(ctx))
+
+	// Wait until we've processed some work before closing
+	require.Eventually(func() bool {
+		return db.NewIterator().Next()
+	}, 5*time.Second, 5*time.Millisecond)
+
```
joshua-kim, line 362: Looks like we forgot to translate the Close here when updating this test

x/sync/db.go
```diff
@@ -3,12 +3,119 @@
...
+
+var (
+	ErrInsufficientHistory = errors.New("insufficient history to generate proof")
+	ErrNoEndRoot           = fmt.Errorf("%w: end root not found", ErrInsufficientHistory)
+)
+
+type ProofFactory[TRange, TChange Proof] interface {
+	NewRangeProof() TRange
+	NewChangeProof() TChange
+}
```
joshua-kim: I'm definitely guilty of even making this naming a trend in our codebase but `*Factory` is really 2008 Spring Java dev and not super Go-ey to me. 

Generally, `NewFoo` naming is commonly used for package-level constructors. I think putting it on a receiver _can_ be okay (and I do this a lot), but it's generally preferred to find some verb that describes the action being performed that creates a type (so in networking packages, `Dial` is the verb to create an instance of a peer as an example, instead of `NewPeer`). In this case though, I think that `foo.NewRangeProof()` reads better than `foo.RangeProof()`, which feels like you're reading data instead of creating something. You could do something like call your interface `ProofBuilder` or `ProofCreator`, but I would argue that `BuildProof` isn't any different from `NewProof` and that although `Build` is a verb it's not really making anything more obvious + these also read like we're forcing Java/C# patterns.

One way we could fix the naming is to just get rid of the interface entirely... if we had an interface like:

```
type Marshaler[T any] interface {
    Marshal(t T) []byte
    Unmarshal(b []byte) T
}
```

We wouldn't even need the `Proof` interface since we're using `any` as our type constraint and we could just define a marshaler for each proof type and pass them both into the syncer.

x/sync/db.go
```diff
@@ -3,12 +3,119 @@
...
+
+type Proof interface {
+	encoding.BinaryMarshaler
+	encoding.BinaryUnmarshaler
+}
+type DB[TRange, TChange Proof] interface {
```
joshua-kim: nit: we typically don't use this syntax in our codebase `TRange, TChange Proof` - we'd probably want something like `TRange Proof, TChange Proof`

x/sync/manager.go
```diff
@@ -148,14 +150,19 @@ type ManagerConfig struct {
 	StateSyncNodes        []ids.NodeID
 }
 
-func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
+func NewManager[R, C Proof](
```
joshua-kim: nit: Can we be consistent w/ the naming for generic types? We're using `TRange` on the struct definition but `R` on this constructor. I don't have a preference for us using `TRange` or `R` (I'd probably use `T` and `U`, or `R` and `C`, but all are fine to use here).

x/sync/manager.go
```diff
@@ -516,7 +525,7 @@ func (m *Manager) handleRangeProofResponse(
 	return nil
 }
 
-func (m *Manager) handleChangeProofResponse(
+func (m *Manager[R, C]) handleChangeProofResponse(
```
joshua-kim, line 533: Seems like we're also using `R, C` here.

x/sync/manager.go
```diff
@@ -786,7 +795,7 @@ func (m *Manager) enqueueWork(work *workItem) {
 	m.unprocessedWork.Insert(second)
 }
 
-// find the midpoint between two keys
+// find the midPoint between two keys
```
joshua-kim: nit: undo diff

x/sync/network_server.go
joshua-kim: Nvm I'm dumb - I should read the calling code.

x/sync/manager.go
joshua-kim: Either the generic black-magic or the marshaler interface are reasonable, but usage of the `reflect` package is entering code-smell territory since we're trying to find out type information... even though we already have it (we know the proof types from the generic).

alarso16: We can use the marshaler interface then!

x/merkledb/client_test.go
```diff
@@ -17,32 +17,36 @@ import (
...
 
 func newFlakyRangeProofHandler(
 	t *testing.T,
-	db merkledb.MerkleDB,
-	modifyResponse func(response *merkledb.RangeProof),
+	db MerkleDB,
+	modifyResponse func(response *RangeProof),
 ) p2p.Handler {
-	handler := NewGetRangeProofHandler(db)
+	handler := xsync.NewGetRangeProofHandler(db, RangeProofMarshaler{})
```
rrazvan1: I would use `marshaler := RangeProofMarshaler{}` and pass it to the `NewGetRangeProofHandler(..)`, and also use it inside `TestHandler.AppRequestF` for better code readability.

x/merkledb/client_test.go
```diff
@@ -72,15 +76,14 @@ func newFlakyRangeProofHandler(
 
 func newFlakyChangeProofHandler(
 	t *testing.T,
-	db merkledb.MerkleDB,
-	modifyResponse func(response *merkledb.ChangeProof),
+	db MerkleDB,
+	modifyResponse func(response *ChangeProof),
 ) p2p.Handler {
-	handler := NewGetChangeProofHandler(db)
+	handler := xsync.NewGetChangeProofHandler(db, RangeProofMarshaler{}, ChangeProofMarshaler{})
```
rrazvan1: same as above, extract to `marshaler := ChangeProofMarshaler{}`

x/sync/network_server.go
```diff
@@ -192,19 +193,21 @@ func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _
...
+type GetRangeProofHandler[TRange any, TChange any] struct {
+	db                  DB[TRange, TChange]
+	rangeProofMarshaler Marshaler[TRange]
 }
 
-func (*GetRangeProofHandler) AppGossip(context.Context, ids.NodeID, []byte) {}
+func (*GetRangeProofHandler[_, _]) AppGossip(context.Context, ids.NodeID, []byte) {}
 
-func (g *GetRangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _ time.Time, requestBytes []byte) ([]byte, *common.AppError) {
+func (g *GetRangeProofHandler[TRange, TChange]) AppRequest(ctx context.Context, _ ids.NodeID, _ time.Time, requestBytes []byte) ([]byte, *common.AppError) {
```
rrazvan1:
```suggestion
func (g *GetRangeProofHandler[TRange, _]) AppRequest(ctx context.Context, _ ids.NodeID, _ time.Time, requestBytes []byte) ([]byte, *common.AppError) {
```

x/merkledb/network_server_test.go
```diff
@@ -117,7 +117,7 @@ func Test_Server_GetRangeProof(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			require := require.New(t)
 
-			handler := NewGetRangeProofHandler(smallTrieDB)
+			handler := xsync.NewGetRangeProofHandler(smallTrieDB, RangeProofMarshaler{})
```
rrazvan1: extract `RangeProofMarshaler{}` to a var? Same reason, it increases the readability.

Maybe declare a global variable for tests that you use everywhere you need.

x/sync/db.go
```diff
@@ -3,12 +3,114 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/maybe"
+)
+
+var (
+	ErrInsufficientHistory = errors.New("insufficient history to generate proof")
+	ErrNoEndRoot           = fmt.Errorf("%w: end root not found", ErrInsufficientHistory)
```
rrazvan1, line 17: wrapping a sentinel error in another one can lead to unexpected behaviour if not used cautiously, more specifically because `errors.Is(ErrNoEndRoot, ErrInsufficientHistory)` is true.

x/sync/db.go
```diff
@@ -3,12 +3,114 @@
...
+
+	// GetChangeProof returns a proof for a subset of the key/value changes in key range
+	// [start, end] that occurred between [startRootID] and [endRootID].
+	// Returns at most [maxLength] key/value pairs.
+	// Returns [sync.ErrStartRootNotFound] if this node has insufficient
+	// history to generate the proof.
+	// Returns ErrEmptyProof if [endRootID] is ids.Empty.
+	// Note that [endRootID] == ids.Empty means the trie is empty
+	// (i.e. we don't need a change proof.)
+	// Returns [sync.ErrEndRootNotFound], if the history doesn't contain the [endRootID].
```
rrazvan1: `ErrEndRootNotFound` doesnt exist anymore, rewrite and also include the error for insufficient history.

x/sync/db.go
```diff
@@ -3,12 +3,114 @@
...
+	// have been are correct prior to [end].
+	CommitChangeProof(ctx context.Context, end maybe.Maybe[[]byte], proof C) (maybe.Maybe[[]byte], error)
+
+	// GetRangeProofAtRoot returns a proof for the key/value pairs in this trie within the range
+	// [start, end] when the root of the trie was [rootID].
+	// If [start] is Nothing, there's no lower bound on the range.
+	// If [end] is Nothing, there's no upper bound on the range.
+	// Returns an error if [rootID] is ids.Empty.
+	// Note that [rootID] == ids.Empty means the trie is empty
+	// (i.e. we don't need a range proof.)
```
rrazvan1, line 83: `ErrInsufficientHistory` is an important error specifically handled in x/sync.
Include it in the comment doc

x/sync/manager.go
```diff
@@ -148,14 +151,18 @@ type ManagerConfig struct {
...
 }
 
-func NewManager(config ManagerConfig, registerer prometheus.Registerer) (*Manager, error) {
+func NewManager[R any, C any](
+	db DB[R, C],
+	config ManagerConfig[R, C],
+	registerer prometheus.Registerer,
+) (*Manager[R, C], error) {
 	switch {
+	case db == nil:
```
rrazvan1, line 162: should we check that `RangeProofMarshaler` and `ChangeProofMarshaler` are `!= nil`?

x/sync/network_server.go
```diff
@@ -45,24 +44,26 @@ var (
...
+	db                   DB[TRange, TChange]
+	rangeProofMarshaler  Marshaler[TRange]
+	changeProofMarshaler Marshaler[TChange]
 }
 
-func (*GetChangeProofHandler) AppGossip(context.Context, ids.NodeID, []byte) {}
+func (*GetChangeProofHandler[TRange, TChange]) AppGossip(context.Context, ids.NodeID, []byte) {}
 
-func (g *GetChangeProofHandler) AppRequest(ctx context.Context, _ ids.NodeID, _ time.Time, requestBytes []byte) ([]byte, *common.AppError) {
+func (g *GetChangeProofHandler[TRange, TChange]) AppRequest(ctx context.Context, _ ids.NodeID, _ time.Time, requestBytes []byte) ([]byte, *common.AppError) {
```
rrazvan1: same, `TChange` can be replaced with `_`

x/sync/db.go
alarso16: I didn't do this - this just moves the errors from merkledb to here. I will make a follow-up since I completely agree, but it's a behavior change

x/merkledb/proof.go
```diff
@@ -183,19 +181,25 @@ func (proof *Proof) Verify(
 	return nil
 }
 
+type RangeProofMarshaler struct{}
```
joshua-kim: nit: usually struct definitions are adjacent to receiver definitions. Can we move this below the below `type RangeProof ChangeProof` definition?

x/sync/db.go
```diff
@@ -3,12 +3,114 @@
...
+	ErrInsufficientHistory = errors.New("insufficient history to generate proof")
+	ErrNoEndRoot           = fmt.Errorf("%w: end root not found", ErrInsufficientHistory)
+)
+
+type Marshaler[T any] interface {
+	Marshal(T) ([]byte, error)
+	Unmarshal([]byte) (T, error)
+}
+
+type DB[R any, C any] interface {
```
joshua-kim, line 25: nit: be consistent with this generic type name. Seems like `TRange` is used most in this PR.

x/sync/db.go
alarso16: Opened an issue

rrazvan1 (approved):

joshua-kim (approved):

Issue #4284, opened 2025-09-16
Upstream ACP-176 math package


michaelkaplan13:
Pull `coreth`'s `acp176` [package](https://github.com/ava-labs/coreth/tree/af368c5237ef982aea3db1483e038f9cc45daaf4/plugin/evm/upgrade/acp176) upstream into AvalancheGo under `vms/evm/upgrade` to allow it to be used by both `coreth` and `subnet-evm` going forward.

This should be done prior to integrating ACP-176 into `subnet-evm` in order to avoid code duplication.

PR #4291, opened 2025-09-17
uplift: Migrate ACP-176 Implementation

ceyonur:
## Why this should be merged

Uplifting acp-176 math package from https://github.com/ava-labs/coreth/tree/master/plugin/evm/upgrade/acp176

## How this works

Copies package from coreth

## How this was tested

copied tests

no

### DIFF
--- a/vms/evm/acp176/acp176.go
+++ b/vms/evm/acp176/acp176.go
@@ -13,9 +13,10 @@ import (
 	"math/big"
 	"sort"
 
+	"github.com/holiman/uint256"
+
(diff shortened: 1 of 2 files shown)


michaelkaplan13:
Closes #4284

michaelkaplan13 (commented):
Confirmed that it is a direct clone from `coreth` using:
```
git diff --no-index avalanchego/vms/evm/acp176/ coreth/plugin/evm/upgrade/acp176/
```
which confirms they are identical.

Looks like we just need to adhere to the AvalancheGo linting.

Issue #4285, opened 2025-09-16
Make ACP-176 math configurable


michaelkaplan13:
Blocked by #4284. 

The existing `acp176` [package](https://github.com/ava-labs/coreth/tree/af368c5237ef982aea3db1483e038f9cc45daaf4/plugin/evm/upgrade/acp176) only supports the existing C-Chain configuration. ACP-224 involves incorporating ACP-176 into Subnet-EVM chains, but allowing them to set their own fee configurations. 

Add a ACP-176 config/params that allows for the following variables (currently hard coded as constants) to be configured by various chains:
```
type Config struct {
	MinGasPrice        gas.Price // M
	TimeToDouble       uint64    // in seconds
	TimeToFillCapacity gas.Gas   // in seconds
}
```

The `TargetToMaxCapacity`, `MinMaxCapacity`, and `TargetToPriceUpdateConversion` will then need to be made functions based on a given configuration. 

For `TargetToPriceUpdateConversion`, a rational approximation of `ln(2)` can be used to ensure precision and avoiding using floating point math.

JonathanOppenheimer:
This is currently blocked by https://github.com/ava-labs/avalanchego/pull/4292. I will begin working on it after that PR is merged.

ceyonur:
I believe this issue can be closed now as we're not going to use acp176 package from avalanchego in SAE (at least not in the current form)

ceyonur:
Moved to https://github.com/ava-labs/strevm/issues/77

PR #4292, opened 2025-09-17
refactor: acp-176/226 to common package

ceyonur:
## Why this should be merged

Refactors common parts from acp-176 and acp-226 into `common` package. 

## How this works

Uses `TargetExcessParam` structure to define common functions, so that different implementations/parameters can use the same functions

## How this was tested

existing tests

no


vms/evm/acp176/acp176.go
```diff
@@ -154,7 +157,7 @@ func (s *State) UpdateTargetExcess(desiredTargetExcess gas.Gas) {
 	)
 
 	// Ensure the gas capacity does not exceed the maximum capacity.
-	newMaxCapacity := mulWithUpperBound(newTargetPerSecond, TargetToMaxCapacity) // C
+	newMaxCapacity := gas.Gas(common.MulWithUpperBound(uint64(newTargetPerSecond), TargetToMaxCapacity)) // C
 	s.Gas.Capacity = min(s.Gas.Capacity, newMaxCapacity)
```
ceyonur, line 181: this now uses golang builtin min, would it be a problem? cc @StephenButtolph

vms/evm/upgrades/acp226/acp226.go
```diff
@@ -24,24 +18,28 @@ const (
 	maxDelayExcess = 46_516_320 // ConversionRate * ln(MaxUint64 / MinDelayMilliseconds) + 1
 )
 
+// acp226Params is the params used for the acp226 upgrade.
+var acp226Params = common.TargetExcessParams{
```
ceyonur: I wonder if a better refactor would be to defer these to coreth/subnet-evm so that they can decide on the parameters and then just call the functions for the parameter structure. But this probably out of the scope for this PR and should be considered with the acp-224 changes. cc @michaelkaplan13 @JonathanOppenheimer

JonathanOppenheimer: Given that in the future, we won't have a `coreth`/`subnet-evm` to defer to, I'd rather not set stuff up in those repositories for the parameters that we'll only have to get rid of and upstream to AvalancheGo eventually. I'll defer to Michael though.

vms/evm/upgrades/acp226/acp226_test.go
michaelkaplan13, line 1: Nit: In `coreth`/`subnet-evm` the subdirectory name is `upgrade` rather than `upgrades`, if that's what we are trying to match here.

ceyonur: I felt like the package is actually a collection of `upgrades`. I don't have strength preference though.

vms/evm/upgrades/acp226/acp226.go
ceyonur: The same can be said for `subnet-evm` or `coreth` packages here

JonathanOppenheimer: Hm wdym?

ceyonur: We eventually will need to have differentiators for subnet-evm and coreth. This can be either two different repositories, or two different packages (or mix of them both). I.e we can pass a config to acp176 package, but the actual config should be defined in either some sub-package in avalanchego (`subnet-evm/coreth`) or directly in the repositories.

michaelkaplan13: Totally agree with @ceyonur. This is the issue for making this math configurable/paramater-izable: https://github.com/ava-labs/avalanchego/issues/4285

Separate from the repo structure/breakdown, we will always need a way for different chains to provide their own settings for these values, which is what Cey is referring to I believe.

JonathanOppenheimer: That makes a lot of sense - I'd personally prefer two different packages rather than two different repositories -- unless there's a strong reason to go against the proposed monorepo, I think PRs etc. across multiple repos for similar, albeit configurable behavior is annoying.

utils/math/exponential.go
```diff
@@ -0,0 +1,77 @@
...
+
+package math
+
+import (
+	"math"
+
+	"github.com/holiman/uint256"
+)
+
+var max256Uint64 = new(uint256.Int).SetUint64(math.MaxUint64)
```
michaelkaplan13, line 12:
```suggestion
var maxUint64 = new(uint256.Int).SetUint64(math.MaxUint64)
```

Nit: to match the existing name

vms/evm/upgrades/common/target_excess.go
```diff
@@ -0,0 +1,63 @@
...
+package common
+
+import (
+	"math"
+	"sort"
+
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+// TargetExcessParams contains the parameters needed for target excess calculations.
```
michaelkaplan13: Now that these are in `common`, I don't think the `Target` prefix is accurate in all cases (i.e. ACP-224). Should we make it just `ExcessParams` and remove the "target" terminology from all of the function names?

i.e:
`Target()`
`DesiredExcess()`
`CalculateValue()`

utils/math/exponential.go
ceyonur: it's redeclared  [here](https://github.com/ava-labs/avalanchego/blob/ceyonur/acp-176-226-refactor/utils/math/safe_math_test.go#L13)
`const maxUint64 uint64 = math.MaxUint64`

vms/evm/upgrades/acp226/acp226.go
```diff
@@ -50,19 +48,5 @@ func DesiredDelayExcess(desiredDelay uint64) uint64 {
...
-// calculateDelayExcess calculates the optimal new DelayExcess for a block proposer to
-// include given the current and desired excess values.
-func calculateDelayExcess(excess, desired uint64) uint64 {
-	change := safemath.AbsDiff(excess, desired)
-	change = min(change, MaxDelayExcessDiff)
-	if excess < desired {
-		return excess + change
-	}
-	return excess - change
+	return acp226Params.DesiredTargetExcess(desiredDelay)
```
michaelkaplan13: Think the above comment can be removed here and just kept with the implementation of the binary search approach.

michaelkaplan13 (commented):
One nit, and think we should update/remove the "target" terminology from `target_excess.go` since it's not just applying to the gas target anymore, but otherwise LGTM.

vms/evm/upgrades/common/target_excess.go
ceyonur: removed prefix

vms/evm/excess/params.go
michaelkaplan13, line 1: Should we rename the file to just `excess.go` as well?

michaelkaplan13 (approved):

utils/math/exponential.go
```diff
@@ -0,0 +1,77 @@
...
+		num   uint256.Int
+		denom uint256.Int
+
+		i              uint256.Int
+		output         uint256.Int
+		numeratorAccum uint256.Int
+
+		maxOutput uint256.Int
+	)
+	num.SetUint64(numerator)     // range is [0, MaxUint64]
```
joshua-kim: nit: All of these range comments seem overly verbose since it's already known that the range of a unit64 is the range of a uint64. Should we get rid of them?

utils/math/exponential.go
```diff
@@ -0,0 +1,77 @@
...
+//			numerator_accum = (numerator_accum * numerator) // (denominator * i)
+//			i += 1
+//		return output // denominator
+//
+// This implementation is optimized with the knowledge that any value greater
+// than MaxUint64 gets returned as MaxUint64. This means that every intermediate
+// value is guaranteed to be at most MaxUint193. So, we can safely use
+// uint256.Int.
+//
+// This function does not perform any memory allocations.
```
joshua-kim: It seems overly verbose to me that we document this - isn't this just a translation of the below source code? Wonder if we should just get rid of this too.

utils/math/exponential.go
```diff
@@ -0,0 +1,77 @@
...
+//
+//nolint:dupword // The python is copied from the EIP-4844 specification
+func CalculateExponential(
```
joshua-kim: Should we call this `ApproximateExponential` so the caller doesn't have to read the godoc to know this is not an exact result?

vms/evm/upgrades/common/excess_params.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Package common provides shared utility functions for ACP implementations.
+package common
```
joshua-kim: I really dislike common packages... I feel like it's unclear what the boundary is for what they own and they tend to turn into dumpsters for things that don't fit elsewhere. Would it be reasonable for us to define an `excess` package instead?

vms/evm/upgrades/common/excess_params.go
```diff
@@ -0,0 +1,63 @@
...
+type ExcessParams struct {
+	MinValue       uint64 // Minimum value (P or M)
+	ConversionRate uint64 // Conversion factor for exponential calculations (D)
+	MaxExcessDiff  uint64 // Maximum change in excess per update (Q)
+	MaxExcess      uint64 // Maximum possible excess value
+}
+
+// AdjustExcess calculates the optimal new excess given the current and desired excess values,
+// ensuring the change does not exceed the maximum excess difference.
+func (p ExcessParams) AdjustExcess(excess, desired uint64) uint64 {
```
joshua-kim: nit: typically receivers start with the first letter of the type (`e`) unless there's a commonly known abbreviation for the type (like `db` or `vm`).

vms/evm/upgrades/common/excess_params.go
```diff
@@ -0,0 +1,63 @@
...
+	"sort"
+
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+// ExcessParams contains the parameters needed for excess calculations.
+type ExcessParams struct {
+	MinValue       uint64 // Minimum value (P or M)
+	ConversionRate uint64 // Conversion factor for exponential calculations (D)
+	MaxExcessDiff  uint64 // Maximum change in excess per update (Q)
```
joshua-kim: Where are these letters coming from? Is there some formula we should doc at the top of this type?

vms/evm/upgrades/common/excess_params.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"math"
+	"sort"
+
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+// ExcessParams contains the parameters needed for excess calculations.
+type ExcessParams struct {
```
joshua-kim: I don't know much about this type so I'm not the person to ask for naming, but should we call this `Excess`? It seems like a math-y type so I assumed everything inside it was a parameter for its calculations.

vms/evm/excess/params.go
```diff
@@ -0,0 +1,63 @@
...
+	return safemath.CalculateExponential(
+		p.MinValue,
+		excess,
+		p.ConversionRate,
+	)
+}
+
+// MulWithUpperBound multiplies two numbers and returns the result. If the
+// result overflows, it returns [math.MaxUint64].
+func MulWithUpperBound(a, b uint64) uint64 {
```
joshua-kim: Does this need to be in this package? Seems like this is only used for acp176 implementation - wondering if we could move and unexport there.

vms/evm/upgrades/common/excess_params.go
ceyonur: Since we have changed the package name, I think we should rename `ExcessParams` > `Params`

vms/evm/upgrades/common/excess_params.go
ceyonur: I moved those letters to relevant acps

vms/evm/upgrades/common/excess_params.go
ceyonur: I think it's still contains parameters for calculations. It's not the real excess, but it's a set of parameters to be used in excess calculations.

vms/evm/excess/params.go
ceyonur: we can move this to acp176 package as an unexported function

utils/math/exponential.go
StephenButtolph: I feel like `max256Uint64` is not really that descriptive. I feel like we should either remove the redeclaration, or name this to something else `maxUint64AsUint256`? `bigMaxUint64`?

utils/math/exponential.go
joshua-kim, line 1: Can we add tests for this new code?

vms/evm/excess/params.go
joshua-kim, line 1: Can we add tests for new code?

utils/math/exponential.go
JonathanOppenheimer: big test table!

vms/evm/excess/params.go
JonathanOppenheimer: big test table!

JonathanOppenheimer (approved):

PR #4297, opened 2025-09-18
C chain reexecution firewood ab test

Elvis339:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,237 +20,424 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  Unit:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [macos-14, ubuntu-22.04, ubuntu-24.04, custom-arm64-jammy, custom-arm64-noble]
+#  Unit:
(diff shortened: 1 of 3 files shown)


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -66,7 +66,7 @@ runs:
   steps:
     - uses: ./.github/actions/setup-go-for-project
     - name: Set task env
-      shell: nix develop --command bash -x {0}
```
maru-ava: I'm not sure why this was ever suggested

.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -82,7 +82,9 @@ runs:
     - name: Run command
       shell: bash
       # --impure ensures the env vars are accessible to the command
-      run: ${{ inputs.run_env }} $GITHUB_ACTION_PATH/nix-develop.sh --impure --command bash -x ${{ inputs.run }}
+      run: |
+        export ${{ inputs.run_env }}
```
maru-ava: As per an offline conversation with Aaron, he's going to stop using run_env (and then I'm going to remove it)) and the job will switch to passing task vars instead.

PR #4300, opened 2025-09-19
ACP-226: add initial delay excess

ceyonur:
## Why this should be merged

Adds default delay excess as 7_970_124 (~2000ms delay)

## How this works

adds the default constant

## How this was tested

changed the test with constant

## Need to be documented in RELEASES.md?'

no

### DIFF
--- a/vms/evm/acp226/acp226.go
+++ b/vms/evm/acp226/acp226.go
@@ -21,9 +21,9 @@ const (
 	// MaxDelayExcessDiff (Q) is the maximum change in excess per update
 	MaxDelayExcessDiff = 200
 
-	// DefaultDelayExcess represents the initial (≈2000ms) delay excess.
+	// InitialDelayExcess represents the initial (≈2000ms) delay excess.
 	// Formula: ConversionRate (2^20) * ln(2000) + 1
-	DefaultDelayExcess = 7_970_124
+	InitialDelayExcess = 7_970_124
 
 	maxDelayExcess = 46_516_320 // ConversionRate * ln(MaxUint64 / MinDelayMilliseconds) + 1
 )
--- a/vms/evm/acp226/acp226.go
+++ b/vms/evm/acp226/acp226.go
@@ -21,6 +21,8 @@ const (
 	// MaxDelayExcessDiff (Q) is the maximum change in excess per update
 	MaxDelayExcessDiff = 200
 
+	DefaultDelayExcess = 7_970_124
+
 	maxDelayExcess = 46_516_320 // ConversionRate * ln(MaxUint64 / MinDelayMilliseconds) + 1
 )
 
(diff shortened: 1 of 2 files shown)


vms/evm/acp226/acp226.go
```diff
@@ -21,6 +21,10 @@ const (
 	// MaxDelayExcessDiff (Q) is the maximum change in excess per update
 	MaxDelayExcessDiff = 200
 
+	// DefaultDelayExcess represents the initial (≈2000ms) delay excess.
+	// Formula: ConversionRate (2^20) * ln(2000) + 1
+	DefaultDelayExcess = 7_970_124
```
StephenButtolph: I'm wondering if this would be more descriptive as the `InitialDelayExcess`. If this is limited to the C-chain, I think it makes more sense as `Initial`, but if there is a situation with subnet-evm that this makes more sense as `Default` then this lgtm.

StephenButtolph (approved):

vms/evm/acp226/acp226.go
ceyonur: I think we also want this to be same default in subnet-evm. Otherwise we should really move them to each repo (see this comment https://github.com/ava-labs/avalanchego/pull/4292#discussion_r2362642275)

ceyonur: (unless we want to have coreth, subnet-evm packages here instead)

StephenButtolph: Yeah I was more wondering on if we are planning on using it as a _default_ (aka none was provided so we use this) or if we are just planning on using this specifically at the time of activation (in which case I think _initial_ is more accurate + aligns with the ACP repo)

RodrigoVillar (approved):

vms/evm/acp226/acp226.go
```diff
@@ -21,6 +21,10 @@ const (
 	// MaxDelayExcessDiff (Q) is the maximum change in excess per update
 	MaxDelayExcessDiff = 200
 
+	// InitialDelayExcess represents the initial (≈2000ms) delay excess.
+	// Formula: ConversionRate (2^20) * ln(2000) + 1
+	InitialDelayExcess = 7_970_124
```
joshua-kim, line 26: These feel like magic numbers even though we use a godoc to document them... could we define these through calculation in code? i.e
```
const conversionRate = 1 << 20 // 2^20 fixed-point scale

var InitialDelayExcess = int(float64(conversionRate)*math.Log(2000)) + 1
```

StephenButtolph: These values are consensus critical, so I don't think it is reasonable to perform floating point operations to generate them. (The actual number is more important than the mechanism used to derive them)

ceyonur: (same reason as what Stephen posted above while I writing this)
I think that would actually end up what we try to avoid with binary search. I think `math.Log` precision can differ from machine to machine and end-up having different initial values.

michaelkaplan13: If we need to make this configurable by minimum delay setting in the future, an approach @StephenButtolph recommended to me for ACP-224 is to use some rational approximation for `ln(x)`, but that would basically be a magic number itself here anyways.

michaelkaplan13 (approved):

PR #4305, opened 2025-09-19
Add simple CLI to fetch and archive C-Chain blocks

aaronbuchwald:


aaronbuchwald:
Failing CI due to `TestMustNotImport` when it tries to import `customethclient` to fetch blocks correctly from the public API:
Screenshot 2025-09-23 at 14 39 24

aaronbuchwald:
Thinking that we should add the following triggers to CI:

PR - create a local network with N (100) blocks, fetch the blocks, and test we can iterate over them successfully
manual workflow - fetch blocks from provided RPC URL, optionally start with existing DB pulled from S3, configure GH timeout, and post to new location in S3
cronjob - maintain an S3 object that contains all of the blocks up to tip and update it once a month

tests/reexecute/c/cli/cmd/fetchBlocks.go
```diff
@@ -0,0 +1,120 @@
...
+		zap.String("dbDir", dbDir),
+	)
+
+	blocksCh := make(chan uint64, concurrency)
+	go func() {
+		defer close(blocksCh)
+		for i := startBlock; i <= endBlock; i++ {
+			blocksCh <- i
+		}
+	}()
```
aaronbuchwald: TODO: add cancellation to avoid leaking a goroutine if the err group below fails.

go.mod
```diff
@@ -88,6 +88,14 @@ require (
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 )
 
+require (
```
joshua-kim, line 91: Looks like we meant to merge this into the below require block?

tests/e2e/c/api.go
```diff
@@ -19,6 +20,46 @@ import (
 	pb "github.com/ava-labs/avalanchego/connectproto/pb/proposervm/proposervmconnect"
 )
 
+func byAdvancingCChainHeight(tc *e2e.GinkgoTestContext, minBlocks int) {
```
joshua-kim, line 23: Why is this code move in this PR?

tests/reexecute/c/cli/cmd/fetch_blocks.go
```diff
@@ -0,0 +1,62 @@
...
+	RunE:  runFetchBlocks,
+}
+
+func init() {
+	rootCmd.AddCommand(fetchBlocksCmd)
+
+	fetchBlocksCmd.Flags().String(dbDirKey, "", "Database to store the fetched blocks")
+	fetchBlocksCmd.Flags().String(rpcURLKey, "http://localhost:9650/ext/bc/C/rpc", "Ethereum RPC URL to fetch blocks from")
+	fetchBlocksCmd.Flags().Uint64(startBlockKey, 0, "Block number to start fetching from (inclusive)")
+	fetchBlocksCmd.Flags().Uint64(endBlockKey, 0, "Block number to stop fetching at (inclusive)")
```
joshua-kim, line 36: Why is stop inclusive? Typically range semantics are inclusive start and exclusive end (like loops), deviating might lead to confusion when using this tooling.

tests/reexecute/c/cli/cmd/fetch_blocks.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	rpcURL, err := cmd.Flags().GetString(rpcURLKey)
+	if err != nil {
+		return fmt.Errorf("failed to get rpc-url flag: %w", err)
+	}
+	concurrency, err := cmd.Flags().GetInt(concurrencyKey)
+	if err != nil {
+		return fmt.Errorf("failed to get concurrency flag: %w", err)
+	}
+	return utils.FetchBlocksToBlockDB(context.Background(), cliLog, dbDir, startBlock, endBlock, rpcURL, concurrency)
```
joshua-kim, line 61: I think we probably want to use `cmd.Context` here - this might cause unexpected behavior if the root context is canceled since this is not a child of the parent.

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
```
joshua-kim, line 37: 1. Why do we define two `init`'s? This introduces complexity since we now have to consciously be aware of how the initialization is ordered - which is not obvious since it is ordered according to build order. It might make sense for us to have a single `init` and merge these two.
2. Go packages are generally fine with big files since there are no limits on numbers of structs/types per file like in other languages. Since the abstraction here is a cli, it would probably be fine to just stuff this all under main, or have one file for the cli and one for main.

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+
+var (
+	cfgFile     string
+	logLevelKey = "log-level"
+	cliLog      logging.Logger
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "argo",
```
joshua-kim, line 24: 1. What is argo? 
2. The naming here isn't self-describing so it feels like it could be improved. We also already use argocd which is what I first though of when reading this.

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+				logging.Colors.ConsoleEncoder(),
+			),
+		)
+		return nil
+	}
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
+	if cfgFile != "" {
```
joshua-kim, line 66: 1. nit: early return for readability
2. Is this config stuff overkill for what we're trying to do here? It seems like for a cli tool that just needs to download blocks in a range using the start/end range parameters would be sufficient - why do we care about supporting config files + env vars?

tests/reexecute/utils/fetch_blocks.go
```diff
@@ -0,0 +1,77 @@
...
+		for i := startBlock; i <= endBlock; i++ {
+			select {
+			case blocksCh <- i:
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	eg, ctx := errgroup.WithContext(ctx)
```
joshua-kim, line 51: Is there a reason we implemented this over using `SetLimit` for concurrency control?

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "argo",
+	Short: "Simple CLI tool to interact with re-execution",
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
```
joshua-kim, line 29: Do we need this helper? It's a helper around a function that's only used once + it's exported even though it's only usage is for main.

tests/reexecute/c/cli/cmd/fetch_blocks.go
```diff
@@ -0,0 +1,62 @@
...
+	dbDirKey       = "db-dir"
+	startBlockKey  = "start-block"
+	endBlockKey    = "end-block"
+	rpcURLKey      = "rpc-url"
+	concurrencyKey = "concurrency"
+)
+
+// fetchBlocksCmd represents the fetchBlocks command
+var fetchBlocksCmd = &cobra.Command{
+	Use:   "fetchBlocks",
```
joshua-kim, line 25: cli tooling typically uses kebab casing for multi-word expressions - additionally this camelCasing is inconsistent with the kebab-cased parameters

tests/reexecute/c/cli/cmd/fetch_blocks.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	dbDirKey       = "db-dir"
+	startBlockKey  = "start-block"
+	endBlockKey    = "end-block"
+	rpcURLKey      = "rpc-url"
+	concurrencyKey = "concurrency"
+)
+
+// fetchBlocksCmd represents the fetchBlocks command
```
joshua-kim, line 23: nit: avoid self-documenting comments

tests/reexecute/utils/blockdb.go
joshua-kim, line 1: We should generally try to avoid utils packages where it's possible - because they never have a clear abstraction boundary it's really easy to turn them into dumpster packages. As an example, the abstraction `utils` here is responsible for defining the `utils.BlockDB` and `utils.FetchBlocksToBlockDB`. It sounds like the package abstraction we wanted was something related to a database instead here.

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var (
+	cfgFile     string
+	logLevelKey = "log-level"
+	cliLog      logging.Logger
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
```
joshua-kim, line 23: Why do we need this root cmd? We only have one sub command anyways, would it make sense to just have a `fetch-blocks` command instead of having to use `argo fetch-blocks`?

tests/reexecute/c/cli/cmd/root.go
```diff
@@ -0,0 +1,86 @@
...
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cli.yaml)")
+	rootCmd.PersistentFlags().String(logLevelKey, logging.Info.String(), "Log level")
+
+	rootCmd.PersistentPreRunE = func(_ *cobra.Command, _ []string) error {
+		logLevelStr, err := rootCmd.PersistentFlags().GetString(logLevelKey)
+		if err != nil {
+			return fmt.Errorf("failed to get %q flag: %w", logLevelKey, err)
+		}
+		logLevel, err := logging.ToLevel(logLevelStr)
```
joshua-kim, line 48: nit: This is in a lot of spots in this PR but typically a line break is done after these braces

tests/e2e/c/fetch_blocks.go
```diff
@@ -0,0 +1,79 @@
...
+			require.NoError(utils.FetchBlocksToBlockDB(
+				tc.DefaultContext(),
+				tc.Log(),
+				fetchBlocksDBDir,
+				startBlock,
+				endBlock,
+				cChainNodeURI,
+				numBlocks,
+			))
+		})
```
joshua-kim, line 49: It feels weird to me that:
1. `utils.FetchBlocksToBlockDB` is exported - this is an implementation detail of the cli 
2. We have an e2e test, but aren't actually testing the cli

I feel like we either need to run a e2e test on the cli binary as a subprocess or do a unit test the root cmd type

Issue #4270, opened 2025-09-12
Add validator set diff by height index


michaelkaplan13:
For both the [weight diffs](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/state.go#L293) and [pub key diffs](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/state.go#L295) tracking in the P-Chain state, add (height+subnet+nodeID) indexes.

These indexes will be required in order to efficiently construct the full validator set states at a given height by starting from the current validator sets (which are always kept in memory) and apply reverse diffs back to the requested height. The current indexes do not allow for this because they would require looping over all subnets that have ever been created.

Similar to the two existing indexes, the new indexes can be added on a "go forwards" basis. That is, nodes can start constructing them when they are updated, but they do not need to recreate them for historic block heights.

This is necessary for ACP-181 to all for efficient look up of all validator sets.

https://github.com/ava-labs/avalanchego/pull/4170

maishivamhoo123:
@team can you please assign this issue to me?

PR #4307, opened 2025-09-22
Validator diffs by height

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/avalanchego/issues/4270?reload=1

### DIFF
--- a/vms/platformvm/state/disk_staker_diff_iterator.go
+++ b/vms/platformvm/state/disk_staker_diff_iterator.go
@@ -41,7 +41,7 @@ func marshalStartDiffKey(subnetID ids.ID, height uint64) []byte {
 
 // marshalStartDiffKeyByHeight is used to determine the starting key when iterating.
 //
-// Invariant: the result is a prefix of [marshalDiffKey] when called with the
+// Invariant: the result is a prefix of [marshalDiffKeyByHeight] when called with the
 // same arguments.
 func marshalStartDiffKeyByHeight(height uint64) []byte {
 	key := make([]byte, database.Uint64Size)


StephenButtolph:
If we made this PR based on `master` and focused just on adding the new state diffs + functions, then I think we should be able to get this merged pretty quickly.

vms/platformvm/state/disk_staker_diff_iterator.go
```diff
@@ -39,6 +39,16 @@ func marshalStartDiffKey(subnetID ids.ID, height uint64) []byte {
 	return key
 }
 
+// marshalStartDiffKey2 is used to determine the starting key when iterating.
+//
+// Invariant: the result is a prefix of [marshalDiffKey] when called with the
+// same arguments.
+func marshalStartDiffKey2(height uint64) []byte {
```
StephenButtolph: I don't think `2` is really descriptive enough. I'm fine changing the name of the prior functions as well. Perhaps `DiffKeyBySubnet` and `DiffKeyByHeight`rather than `DiffKey` and `DiffKey2`?

vms/platformvm/state/state.go
```diff
@@ -209,6 +211,22 @@ type State interface {
...
+	// TODO
+	ApplyValidatorWeightDiffsByHeight(
+		ctx context.Context,
+		validators map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput,
+		startHeight uint64,
+		endHeight uint64,
+	) error
+
+	// TODO
+	ApplyValidatorPublicKeyDiffsByHeight(
```
StephenButtolph:
```suggestion
	ApplyAllValidatorWeightDiffs(
		ctx context.Context,
		validators map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput,
		startHeight uint64,
		endHeight uint64,
	) error

	// TODO
	ApplyAllValidatorPublicKeyDiffs(
```

vms/platformvm/state/state.go
```diff
@@ -389,8 +407,10 @@ type state struct {
 	pendingSubnetDelegatorBaseDB database.Database
 	pendingSubnetDelegatorList   linkeddb.LinkedDB
 
-	validatorWeightDiffsDB    database.Database
-	validatorPublicKeyDiffsDB database.Database
+	validatorWeightDiffsDB            database.Database
+	validatorWeightDiffsByHeightDB    database.Database
+	validatorPublicKeyDiffsDB         database.Database
```
StephenButtolph:
```suggestion
	validatorWeightDiffsBySubnetIDDB            database.Database
	validatorWeightDiffsByHeightDB    database.Database
	validatorPublicKeyDiffsBySubnetIDDB         database.Database
```

vms/platformvm/state/state.go
```diff
@@ -1385,6 +1409,71 @@ func (s *state) SetCurrentSupply(subnetID ids.ID, cs uint64) {
...
+
+		weightDiff, err := unmarshalWeightDiff(diffIter.Value())
+		if err != nil {
+			return err
+		}
+
+		vdrs, ok := allValidators[subnetID]
+		if !ok {
+			// If this subnet previously had no validators, add the map back
+			vdrs = make(map[ids.NodeID]*validators.GetValidatorOutput, 1)
```
StephenButtolph: We don't really know the size here. Probably not worth putting the suggested size at `1`
```suggestion
			vdrs = make(map[ids.NodeID]*validators.GetValidatorOutput)
```

vms/platformvm/state/state.go
```diff
@@ -1477,6 +1566,60 @@ func applyWeightDiff(
...
+			break
+		}
+
+		vdrs, ok := allValidators[subnetID]
+		if !ok {
+			// A subnet that is eventually removed from the map for having no validators may have a key diff before it was removed
+			continue
+		}
+
+		vdr, ok := vdrs[nodeID]
```
StephenButtolph: The prior code is correct, this is just more succinct. 
```suggestion
		vdr, ok := allValidators[subnetID][nodeID]
```

vms/platformvm/state/state.go
```diff
@@ -1477,6 +1566,60 @@ func applyWeightDiff(
...
+		} else {
+			vdr.PublicKey = bls.PublicKeyFromValidUncompressedBytes(pkBytes)
+		}
+	}
+
+	// Note: this does not fallback to the linkeddb index because the linkeddb
+	// index does not contain entries for when to remove the public key.
+	//
+	// Nodes may see inconsistent public keys for heights before the new public
+	// key index was populated.
```
StephenButtolph: This comment doesn't really make any sense here. We will never have stored lookups by height in the linkeddbs.

vms/platformvm/state/state.go
```diff
@@ -2605,6 +2748,25 @@ func (s *state) writeValidatorDiffs(height uint64) error {
...
+		}
+		if !bytes.Equal(diff.prevPublicKey, diff.newPublicKey) {
+			err := s.validatorPublicKeyDiffsByHeightDB.Put(
+				diffKey2,
+				diff.prevPublicKey,
+			)
+			if err != nil {
+				return err
+			}
+		}
```
StephenButtolph: These `if`s are duplicates of the prior `if`s. We should just write to both indices in the prior `if`s and remove the duplicate checks.

vms/platformvm/state/state.go
```diff
@@ -2605,6 +2748,25 @@ func (s *state) writeValidatorDiffs(height uint64) error {
 				return err
 			}
 		}
+		diffKey2 := marshalDiffKey2(height, subnetIDNodeID.subnetID, subnetIDNodeID.nodeID)
+		if diff.weightDiff.Amount != 0 {
+			err := s.validatorWeightDiffsByHeightDB.Put(
+				diffKey2,
+				marshalWeightDiff(&diff.weightDiff),
```
StephenButtolph: Once we unify the `if`s, we should only marshal the weight diff once.

snow/validators/manager.go
```diff
@@ -274,6 +277,17 @@ func (m *manager) Sample(subnetID ids.ID, size int) ([]ids.NodeID, error) {
...
+func (m *manager) GetAllMaps() map[ids.ID]map[ids.NodeID]*GetValidatorOutput {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	set := make(map[ids.ID]map[ids.NodeID]*GetValidatorOutput, len(m.subnetToVdrs))
+	for subnetID, vdrs := range m.subnetToVdrs {
+		set[subnetID] = vdrs.Map()
+	}
+	return set
+}
```
StephenButtolph, line 289: Hmmm this seems really subtle. The manager's lock only protects `m.subnetToVdrs`, it does _not_ protect against modifications of the _values_ in the map.

This is fine for all of the other methods because they are performed on a single validator set. However, here we expect consistency over _all_ validator sets.

I am specifically concerned about the case that:
1. we are calling `GetAllMaps()`
2. There is another goroutine calling `AddWeight(subnetID0, ...)` then `AddWeight(subnetID1, ...)`

`GetAllMaps()` (if properly atomic) would guaranteed that we have either:
- Not seen these weight changes
- Seen only the weight change for `subnetID0`
- Seen the weight change for both `subnetID0` and `subnetID1`

As it is currently implemented, it is possible for:
- Only the weight change for `subnetID1` to be reflected.

I'm not sure that this is really a problem... This _is_ effectively a race anyways (which isn't something we need to be worried about in the platformvm because we already guarantee that at most one thread is executing on the P-chain side). But I feel like we should handle this case properly (like we have with the case documented in `AddWeight`).

vms/platformvm/validators/manager.go
```diff
@@ -319,6 +371,15 @@ func (m *manager) makeValidatorSet(
 	return validatorSet, currentHeight, err
 }
 
+func (m *manager) getAllCurrentValidatorSets(
+	ctx context.Context,
+) (map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput, uint64, error) {
+	subnetsMap := m.cfg.Validators.GetAllMaps()
+	// TODO is there a race-condition here?
```
StephenButtolph: There is not
```suggestion
```

vms/platformvm/state/state.go
geoff-vball: The linkeddb part doesn't make sense. The second half does make sense though, right? Nodes may see inconsistent public keys for heights before they started using this index

PR #4310, opened 2025-09-22
refactor: uplift `gossip` package from evm repositories 

JonathanOppenheimer:
## Why this should be merged
There are packages in subnet-evm and coreth that share a lot of functionality / are mirrored. These packages should be uplifted to AvalancheGo to prevent the duplicative maintenance of two sets of the same code. I recently was syncing changes between coreth and subnet-evm for these gossip changes, and it would be better if this code was just in AvalancheGo 

Here are the packages being unified and uplifted: 
**coreth**: https://github.com/ava-labs/coreth/tree/master/plugin/evm/gossip
**subnet-evm**: https://github.com/ava-labs/subnet-evm/tree/master/plugin/evm/gossip

## How this was tested

No


vms/evm/gossip/handler.go
```diff
@@ -0,0 +1,81 @@
...
+	return &txGossipHandler{
+		appGossipHandler:  handler,
+		appRequestHandler: validatorHandler,
+	}, nil
+}
+
+type txGossipHandler struct {
+	appGossipHandler  p2p.Handler
+	appRequestHandler p2p.Handler
+}
```
RodrigoVillar: nit: this should live above `NewTxGossiphandler()` [[ref](https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering)]

JonathanOppenheimer: thanks!

vms/evm/gossip/handler.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/network/p2p/gossip"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+var _ p2p.Handler = (*txGossipHandler)(nil)
+
+type txGossipHandler struct {
```
RodrigoVillar, line 22: Since the consumers of this type will live outside the `gossip` package, could we export this type?

JonathanOppenheimer: The only place we use this type is internal to the package at the moment - what would the rationale be to export it?

RodrigoVillar: My rationale was that _eventually_, `coreth` and `subnet-evm` would be the consumers of the package.

After looking at how `coreth` uses `txGossipHandler` and [this](https://go.dev/doc/effective_go#generality), keeping `txGossipHandler` unexported is fine with me; I think we could return `p2p.Handler` instead of `*txGossipHandler` in `NewTxGossipHandler()`.

JonathanOppenheimer: `coreth` and `subnet-evm` will be consumers of this package immediately after it's merged into AvalancheGo. 

The second part aligns more with the docs I agree, although we need to return two `p2p.Handler`s, which is why the type exists if I had to guess. Would you prefer a tuple or something?

JonathanOppenheimer: Nevermind - this is fixed!

ARR4N:
Is this the "temporary" directory, or is it the final destination for this code? If the latter, I'll provide more thorough comments on style.

JonathanOppenheimer:
> Is this the "temporary" directory, or is it the final destination for this code? If the latter, I'll provide more thorough comments on style.

This is the final destination for the code. 'uplifting' is the is the process of cleaning up, aligning with AvalancheGo standards, etc. so these typically get more thorough reviews.

JonathanOppenheimer:
This PR was superseded by https://github.com/ava-labs/avalanchego/pull/4819

PR #4316, opened 2025-09-23
[tooling] Switch to go 1.24's `go tool` to manage CLI dependencies

maru-ava:
## Why this should be merged

Previously, golangci-lint was invoked with `go run` and on every minor version golang upgrade (e.g. 1.23 -> 1.24), a now-incompatible cached version of golangci-lint compiled for the previous golang version would report an error when attempting to run `task lint`.

Thankfully, as of golang 1.24, `go tool` can manage user tooling in a way that is version-sensitive. That means a version change will automatically build a compatible version of golangci-lint - no more incompatible cached versions.

Rather than polluting the main go.mod with our CLI tooling dependencies, a separate tooling module is used at `./tools/go.mod`. Instructions in that file document usage and maintenance.

## How this was tested

- CI
- Locally upgraded to golang 1.25 and checked that `task lint` executed without the previously-seen version error.

N/A


scripts/actionlint.sh
```diff
@@ -2,11 +2,12 @@
...
 set -euo pipefail
 
-go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.1 "${@}"
+AVALANCHE_PATH="$(cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )"
+"${AVALANCHE_PATH}"/scripts/run_tool.sh actionlint "${@}"
 
 echo "Checking use of scripts/* in GitHub Actions workflows..."
 SCRIPT_USAGE=
-for file in .github/workflows/*.{yml,yaml}; do
+for file in "${AVALANCHE_PATH}"/.github/workflows/*.{yml,yaml}; do
```
maru-ava, line 10: This change ensures the script is compatible with being invoked from a path other than the repo root.

PR #4318, opened 2025-09-23
chore: update blockdb to use HeightIndex interface and errors in database pkg

DracoLi:
## Why this should be merged

Follow up to https://github.com/ava-labs/avalanchego/pull/4133, update `x/blockdb` Database to align with the `database.HeightIndex` interface and use the errors from `database/errors.go` for database closed and not found.

## How this works

- Renamed public methods WriteBlock, ReadBlock, and HasBlock to Put, Get, and Has
- Replaced `ErrDatabaseClosed` and `ErrBlockNotFound` with the common errors in `database/errors.go`
- Add HeightIndex dbtests to tests and aligned behaviour

## How this was tested

Unit tests

No

### DIFF
--- a/x/blockdb/writeblock_test.go
+++ b/x/blockdb/writeblock_test.go
@@ -19,7 +19,7 @@ import (
 	safemath "github.com/ava-labs/avalanchego/utils/math"
 )
 
-func TestPut_GetIntegrity(t *testing.T) {
+func TestPutGet(t *testing.T) {
 	tests := []struct {
 		name  string
 		block []byte
@@ -54,7 +54,7 @@ func TestPut_GetIntegrity(t *testing.T) {
 	}
 }
 
-func TestPut_MaxHeightVerification(t *testing.T) {
+func TestPut_MaxHeight(t *testing.T) {
 	customConfig := DefaultConfig().WithMinimumHeight(10)
 
 	tests := []struct {


x/blockdb/database.go
```diff
@@ -20,15 +20,16 @@ import (
...
 	"github.com/ava-labs/avalanchego/utils/logging"
 
 	safemath "github.com/ava-labs/avalanchego/utils/math"
 )
 
 const (
-	indexFileName          = "blockdb.idx"
-	dataFileNameFormat     = "blockdb_%d.dat"
+	indexFileName          = "db.idx"
+	dataFileNameFormat     = "db_%d.dat"
```
joshua-kim: Diff looks unrelated - can we unstage?

x/blockdb/database.go
```diff
@@ -38,8 +39,8 @@ const (
 	// IndexFileVersion is the version of the index file format.
 	IndexFileVersion uint64 = 1
 
-	// BlockEntryVersion is the version of the block entry.
-	BlockEntryVersion uint16 = 1
+	// DataEntryVersion is the version of a data entry.
+	DataEntryVersion uint16 = 1
```
joshua-kim: Diff looks unrelated

x/blockdb/database.go
```diff
@@ -336,12 +339,6 @@ func (s *Database) WriteBlock(height BlockHeight, block BlockData) error {
 		return fmt.Errorf("%w: block size cannot exceed %d bytes", ErrBlockTooLarge, math.MaxUint32)
 	}
 
-	blockDataLen := uint32(blockSize)
-	if blockDataLen == 0 {
-		s.log.Error("Failed to write block: empty block", zap.Uint64("height", height))
-		return ErrBlockEmpty
```
joshua-kim, line 342: Seems like we want to figure out what the expected behavior is for a zero-length block size. In the memdb PR it was an error but here our behavior is inconsistent.

x/blockdb/database_test.go
joshua-kim, line 1: similar comment here - we might want to use our db test package if we're now satisfying the `database.Database` interface

x/blockdb/writeblock_test.go
```diff
@@ -320,3 +309,22 @@ func TestWriteBlock_Errors(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, []byte{}, block)
+	})
+
+	t.Run("zero length block", func(t *testing.T) {
+		require.NoError(t, store.Put(1, []byte{}))
+		block, err := store.Get(1)
+		require.NoError(t, err)
+		require.Equal(t, []byte{}, block)
+	})
```
joshua-kim: We should be able to refactor and use a testing table here.

x/blockdb/database.go
DracoLi: change is covered in the pr description. I can create a separate pr for this if needed.

x/blockdb/database.go
DracoLi: change is documented in the pr description. I didn't opt to create a separate pr since its a small change in `/x/` and partly related to making blockdb not tied to blocks. can create a separate pr if needed.

x/blockdb/database.go
DracoLi: memdb will be updated. we should support saving empty length data to align with kv database behaviour.

x/blockdb/database_test.go
DracoLi: the tests in this package tests the underlying implementation logic (ie eviction, index files). I think we can create separate tests for the HeghtIndex interface but I don't plan on including that in this PR.

joshua-kim: Oh I'm dumb... we're satisfying the `HeightIndex` interface, not the `Database` interface.

DracoLi:
Once https://github.com/ava-labs/avalanchego/pull/4212 is merged in. I'll also add the common interface tests to blockdb tests to ensure consistent behaviour.

x/blockdb/database.go
DracoLi: removed this - will create separate PR to group these clean up changes

x/blockdb/writeblock_test.go
```diff
@@ -214,18 +215,6 @@ func TestWriteBlock_Errors(t *testing.T) {
...
-			height:  0,
-			block:   nil,
-			wantErr: ErrBlockEmpty,
-		},
-		{
-			name:    "empty block zero length",
-			height:  0,
-			block:   []byte{},
-			wantErr: ErrBlockEmpty,
-		},
```
joshua-kim, line 228: Shouldn't these test cases still be covered? It seems like we removed them because we now allow us to write nil/empty bytes - I think we still want to add coverage on these cases since they now work but should expect a non-nil error. Would it make sense to add them to `TestWriteBlock_Basic`?

DracoLi: yep i agree. we should ensure consistent behaviour within blockdb. I'll add them back.

DracoLi: https://github.com/ava-labs/avalanchego/pull/4318/commits/2ef8932e0792f508bd4c852a47d850496f6f52d9

x/blockdb/writeblock_test.go
```diff
@@ -136,16 +172,16 @@ func TestWriteBlock_Basic(t *testing.T) {
...
 			}
 
 			// Verify all written blocks are readable and data is correct
 			for h, expectedBlock := range blocksWritten {
-				readBlock, err := store.ReadBlock(h)
-				require.NoError(t, err, "ReadBlock failed at height %d", h)
+				readBlock, err := store.Get(h)
+				require.NoError(t, err, "Get failed at height %d", h)
 				require.Equal(t, expectedBlock, readBlock)
 			}
```
joshua-kim: We test this behavior in TestWriteBlock_Basic that a block inserted via Put is returned through Get. It feels like we're duplicating coverage here, since it seems that this test is intended to verify heights.

Some ideas might be to either:

1. Refactor the tests to share the same test table, since it seems that testing the correctness of Put requires us to verify behavior on Get and the height apis
2. Remove this assertion entirely from this test, so we'd have a single test for Put/Get and one for Put/heights?

DracoLi: I think 2 makes sense here. The basic Put/Get tests should be covered and we can focus this tests on just the height verification. I'll remove this dup assertion to keep this tests more focused

joshua-kim: Sweet. We might also want to rename the tests here to reflect that the former is testing the contract between Put/Get and this one is testing the contract between Put/Height.

PR #4323, opened 2025-09-23
Add test-unit-fast task

StephenButtolph:
## Why this should be merged

```sh
time task test-unit
```

reports:
```
task test-unit  322.95s user 113.78s system 401% cpu 1:48.80 total
```

```sh
time task test-unit-fast
```

reports:
```
task test-unit-fast  30.95s user 63.20s system 732% cpu 12.848 total
```

aka almost 2m to run the unit tests replaced with about 13s.

## How this works

Adds a task to run the unit tests with:
- Race detection disabled (much faster)
- Shuffling disabled (allows the test runs to be cached)

## How this was tested

Locally

no

### DIFF
--- a/scripts/build_test.sh
+++ b/scripts/build_test.sh
@@ -7,6 +7,18 @@ AVALANCHE_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )
 # Load the constants
 source "$AVALANCHE_PATH"/scripts/constants.sh
 
+# Run the tests with shuffling unless NO_SHUFFLE is set.
+shuffle="-shuffle=on"
+if [[ -n "${NO_SHUFFLE:-}" ]]; then
+    shuffle=""
+fi
+
+# Run the tests with race detection unless NO_RACE is set.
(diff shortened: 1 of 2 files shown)


scripts/build_test.sh
```diff
@@ -18,4 +30,4 @@ fi
 TEST_TARGETS="$(eval "go list ./... ${EXCLUDED_TARGETS}")"
 
 # shellcheck disable=SC2086
-go test -tags test -shuffle=on -race -timeout="${TIMEOUT:-120s}" -coverprofile="coverage.out" -covermode="atomic" ${TEST_TARGETS}
+go test -tags test ${shuffle:-} ${race:-} -timeout="${TIMEOUT:-120s}" -coverprofile="coverage.out" -covermode="atomic" ${TEST_TARGETS}
```
aaronbuchwald, line 33: Seems like we are handling treating this as an empty string, twice? This statement is saying if set and not empty string, use the value, but if we just evaluated it, it would also be the empty string. Still probably cleaner to handle this way.

PR #4328, opened 2025-09-24
Move ginkgo tool usage to the main go.mod from tools/go.mod

maru-ava:
## Why this should be merged

CI logs were reporting that the version of the ginkgo CLI was divergent from the version used by our test suites. Since most of ginkgo's dependencies are already represented in the main module, the risk of unwanted dependency pollution is low and managing ginkgo only in the main module avoids having to synchronize the version across the 2 modules.

## How this was tested

CI

N/A

### DIFF
--- a/.github/actions/setup-go-for-project/action.yml
+++ b/.github/actions/setup-go-for-project/action.yml
@@ -20,3 +20,6 @@ runs:
       uses: actions/setup-go@v5
       with:
         go-version-file: go.mod
+        cache-dependency-path: |
+          go.sum
+          tools/go.sum


.github/CODEOWNERS
```diff
@@ -28,6 +28,7 @@
 /tests/ @joshua-kim @maru-ava
 /tests/*.md @joshua-kim @maru-ava @meaghanfitzgerald
 /tests/reexecute/ @aaronbuchwald
+/tools/* @joshua-kim @maru-ava
```
maru-ava, line 31: Forgot to add this in the PR moving CLI tooling to `go tool`.

go.mod
```diff
@@ -206,3 +212,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+tool github.com/onsi/ginkgo/v2/ginkgo
```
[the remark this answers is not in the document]
maru-ava: This is the output of golang tooling, I'm not going to hand-edit it to satisfy you.

.github/actions/setup-go-for-project/action.yml
```diff
@@ -20,3 +20,6 @@ runs:
       uses: actions/setup-go@v5
       with:
         go-version-file: go.mod
+        cache-dependency-path: |
+          go.sum
+          tools/go.sum
```
maru-ava, line 25: Realized that the new tools module wasn't being targeted for caching.

Elvis339: Not sure if it's related, but I notice weird [output](https://github.com/ava-labs/avalanchego/actions/runs/17921591860/job/50957709134) when comparing caching enabled vs disabled. Didn't affect results, just strange logs. Not a blocker - more of a 'what's happening here?' situation.

maru-ava: The job you linked to appears to be using both setup-go-for-project and install-nix. With this PR, it should either be one or the other. Do you have an example of strange behavior where only setup-go-for-project is exhibiting strange behavior?

Elvis339: I don't but that makes sense, thanks for clearing it up.

PR #4331, opened 2025-09-24
uplift `database` package from evm repositories

JonathanOppenheimer:
## Why this should be merged
There are packages in subnet-evm and coreth that share a lot of functionality / are mirrored. As has been discussed as an evm team wide effort, we are working to uplift these packages to AvalancheGo to prevent the duplicative maintenance of two sets of the same code. This particular package, `database` is a relatively easy lift. 

Here are the packages being unified and uplifted: 
**coreth**: https://github.com/ava-labs/coreth/tree/master/plugin/evm/database
**subnet-evm**: https://github.com/ava-labs/subnet-evm/tree/master/plugin/evm/database

## How this was tested

No

### DIFF
--- a/vms/evm/database/database_test.go
+++ b/vms/evm/database/database_test.go
@@ -121,9 +121,7 @@ func (it *testSnapshotIterator) Value() []byte {
 	return it.pairs[it.index].value
 }
 
-func (*testSnapshotIterator) Release() {
-	// No cleanup needed for snapshot iterator
-}
+func (*testSnapshotIterator) Release() {}
 
 func (*testSnapshotIterator) Error() error {
 	return nil


vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return db.Database.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+// NewIteratorWithStart implements ethdb.Database
+func (db ethDBWrapper) NewIteratorWithStart(start []byte) ethdb.Iterator {
+	return db.Database.NewIteratorWithStart(start)
+}
+
+// wrappedBatch implements ethdb.wrappedBatch
```
RodrigoVillar:
```suggestion
// wrappedBatch implements ethdb.Batch
```

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// NewIterator implements ethdb.Database
+//
+// Note: This method assumes that the prefix is NOT part of the start, so there's
+// no need for the caller to prepend the prefix to the start.
+func (db ethDBWrapper) NewIterator(prefix []byte, start []byte) ethdb.Iterator {
+	// avalanchego's database implementation assumes that the prefix is part of the
+	// start, so it is added here (if it is provided).
+	if len(prefix) > 0 {
```
RodrigoVillar:
```suggestion
	if len(prefix) != 0 {
```

The built-in `len()` function will always return a non-negative number [[ref](https://go.dev/ref/spec#Array_types)], so checking if it's equal to `0` is sufficient here.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"errors"
+
+	"github.com/ava-labs/libevm/ethdb"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+var (
+	_ ethdb.KeyValueStore = (*ethDBWrapper)(nil)
```
RodrigoVillar: Could we also add a type assertion for `wrappedBatch`? i.e.

```
var _ ethdb.Batch = (*wrappedBatch)(nil)
```

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,69 @@
...
+	ErrSnapshotNotSupported = errors.New("snapshot is not supported")
+)
+
+func WrapDatabase(db database.Database) ethdb.KeyValueStore { return ethDBWrapper{db} }
+
+// ethDbWrapper implements ethdb.Database
+type ethDBWrapper struct{ database.Database }
+
+// Stat implements ethdb.Database
+func (ethDBWrapper) Stat(string) (string, error) { return "", database.ErrNotFound }
```
RodrigoVillar: Q: Normally, we return `database.ErrNotFound` if a key doesn't exist. Since it seems that we're erroring here due to this method not being supported, do you think it would make sense to instead return `ErrStatNotSupported`?

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,69 @@
...
+	return db.Database.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+// NewIteratorWithStart implements ethdb.Database
+func (db ethDBWrapper) NewIteratorWithStart(start []byte) ethdb.Iterator {
+	return db.Database.NewIteratorWithStart(start)
+}
+
+// wrappedBatch implements ethdb.wrappedBatch
+type wrappedBatch struct{ database.Batch }
```
RodrigoVillar: Q: would it make sense to rename this type to `ethBatchWrapper`? This way, we have:
- `ethDBWrapper`
- `ethBatchWrapper`

and so we're consistent with our type naming.

JonathanOppenheimer:
Thanks for the suggestions rodrigo :)

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,71 @@
...
+	_ ethdb.Batch         = (*ethBatchWrapper)(nil)
+	_ ethdb.KeyValueStore = (*ethDBWrapper)(nil)
+
+	ErrSnapshotNotSupported = errors.New("snapshot is not supported")
+	ErrStatNotSupported     = errors.New("stat is not supported")
+)
+
+func WrapDatabase(db database.Database) ethdb.KeyValueStore { return ethDBWrapper{db} }
+
+// ethDbWrapper implements ethdb.Database
```
RodrigoVillar: Can we remove all instances of `X implements Y`?

JonathanOppenheimer: done

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+
+var (
+	_ ethdb.Batch         = (*ethBatchWrapper)(nil)
+	_ ethdb.KeyValueStore = (*ethDBWrapper)(nil)
+
+	ErrSnapshotNotSupported = errors.New("snapshot is not supported")
+	ErrStatNotSupported     = errors.New("stat is not supported")
+)
+
+func WrapDatabase(db database.Database) ethdb.KeyValueStore { return ethDBWrapper{db} }
```
joshua-kim: This is an adapter/wrapper pattern - I would just name this `New`. You can come up with more creative names like `FromDatabase` or `WithDatabase` to make it clear you're adapting something or deriving a type from an existing one, but since there's no other constructor defined in this package I would just go super simple and boring and go for `New` because there's nothing to disambiguate against.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+var (
+	_ ethdb.Batch         = (*ethBatchWrapper)(nil)
+	_ ethdb.KeyValueStore = (*ethDBWrapper)(nil)
+
+	ErrSnapshotNotSupported = errors.New("snapshot is not supported")
+	ErrStatNotSupported     = errors.New("stat is not supported")
```
joshua-kim: It seems like we don't care about these error types - I would just unexport here to remove some package api bloat.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func WrapDatabase(db database.Database) ethdb.KeyValueStore { return ethDBWrapper{db} }
+
+type ethDBWrapper struct{ database.Database }
```
joshua-kim: 1. nit: I would just call this `database` or `db` or something. It's implied that this is an eth-compatible db because we're in the `evm` package. It's also obvious from looking at the struct definition that this is a wrapper pattern.
2. avoid embed if we can, see later comment

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+		start = newStart
+	}
+	return db.Database.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+func (db ethDBWrapper) NewIteratorWithStart(start []byte) ethdb.Iterator {
+	return db.Database.NewIteratorWithStart(start)
+}
+
+type ethBatchWrapper struct{ database.Batch }
```
joshua-kim: nit: avoid embeds where possible unless we need it - it autopromotes functions/fields into our type and that introduces subtle behavior changes that can be difficult to notice. That being said - we only use this type and the earlier type as their interfaces, so the autopromoted methods never actually make a difference in practice.

We avoid some boilerplate code of implementing passthroughs by embedding, but the cost of maintaing passthrough code is trivial and creates more obvious + readable behavior than using embeds.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (batch ethBatchWrapper) ValueSize() int { return batch.Batch.Size() }
```
joshua-kim: Receiver is called `batch` - not a single letter or an abbreviation. Could we use the first letter of the type?

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+
+func WrapDatabase(db database.Database) ethdb.KeyValueStore { return ethDBWrapper{db} }
+
+type ethDBWrapper struct{ database.Database }
+
+func (ethDBWrapper) Stat(string) (string, error) { return "", ErrStatNotSupported }
+
+func (db ethDBWrapper) NewBatch() ethdb.Batch { return ethBatchWrapper{db.Database.NewBatch()} }
+
+// TODO: propagate size through avalanchego Database interface
```
joshua-kim: I can't imagine we're ever going to do this TODO because it seems like such a premature optimization - I'd just delete it.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+	"errors"
+
+	"github.com/ava-labs/libevm/ethdb"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+var (
+	_ ethdb.Batch         = (*ethBatchWrapper)(nil)
+	_ ethdb.KeyValueStore = (*ethDBWrapper)(nil)
```
joshua-kim: I don't know if this is a real convention or not but I almost always see `_` vars defined after other "real" named vars.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+func (db ethDBWrapper) NewBatchWithSize(int) ethdb.Batch {
+	return ethBatchWrapper{db.Database.NewBatch()}
+}
+
+func (ethDBWrapper) NewSnapshot() (ethdb.Snapshot, error) {
+	return nil, ErrSnapshotNotSupported
+}
+
+// This method assumes that the prefix is NOT part of the start, so there's
+// no need for the caller to prepend the prefix to the start.
```
joshua-kim: This comment just seems weird to me - the function signature is literally:
```
NewIterator(prefix []byte, start []byte)
```

The prefix and start are two distinct parameters, so I'm not sure why anyone would ever assume that the start should have the prefix prepended to it. I feel like we could just get rid of this godoc.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+func (db ethDBWrapper) NewIterator(prefix []byte, start []byte) ethdb.Iterator {
+	// avalanchego's database implementation assumes that the prefix is part of the
+	// start, so it is added here (if it is provided).
```
joshua-kim: Similarly this comment is describing the guarantees of the avalanchego `Database` interface, so if there's something unclear I feel like the doc should live on that interface instead of here. I would remove this comment too.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(prefix) != 0 {
```
joshua-kim: Do we even need this length check here? It seems to me like this code works with zero-length prefixes to me unless I'm missing something. Maybe this is a performance optimization but I don't think iterators are expensive to create, the cost is typically on the iteration.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+		newStart := make([]byte, len(prefix)+len(start))
+		copy(newStart, prefix)
+		copy(newStart[len(prefix):], start)
+		start = newStart
+	}
+	return db.Database.NewIteratorWithStartAndPrefix(start, prefix)
```
joshua-kim: nit: typically add a new line before a return

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,61 @@
...
+
+type ethDBWrapper struct{ database.Database }
+
+func (ethDBWrapper) Stat(string) (string, error) { return "", ErrStatNotSupported }
+
+func (db ethDBWrapper) NewBatch() ethdb.Batch { return ethBatchWrapper{db.Database.NewBatch()} }
+
+// TODO: propagate size through avalanchego Database interface
+func (db ethDBWrapper) NewBatchWithSize(int) ethdb.Batch {
+	return ethBatchWrapper{db.Database.NewBatch()}
```
joshua-kim: We could just return `db.NewBatch()` here

vms/evm/database/wrapped_database.go
JonathanOppenheimer: 1. We can't do `database` (it's the import) or `db` bc of the intervening var definition. I'd say we do `ethdb`, but that's also an import. Any other suggestions? 
2. done

vms/evm/database/wrapped_database.go
JonathanOppenheimer: If we do that, then we does not implement `ethdb.KeyValueStore`:
```
		have NewBatch() "github.com/ava-labs/avalanchego/database".Batch
		want NewBatch() ethdb.Batch
```

vms/evm/database/wrapped_database.go
joshua-kim: You can alias imports if there's a namespace collision (you could call `database` `avalanchegodatabase` or `avalanchegodb` or something)

JonathanOppenheimer: Done!

vms/evm/database/wrapped_database.go
joshua-kim: This comment seems to have been marked as resolved but it's still in the diff

JonathanOppenheimer: My apologies - now it's actually resolved!

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,79 @@
...
+	copy(newStart, prefix)
+	copy(newStart[len(prefix):], start)
+	start = newStart
+
+	return db.db.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+func (db database) NewIteratorWithStart(start []byte) ethdb.Iterator {
+	return db.db.NewIteratorWithStart(start)
+}
```
joshua-kim: This is dead code - this isn't on the `KeyValueStore` interface and we return the interface in `New`. Can we delete this? It looks like the old comment that we deleted `// NewIteratorWithStart implements ethdb.Database` was incorrect.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (e ethBatchWrapper) ValueSize() int { return e.batch.Size() }
+
+func (e ethBatchWrapper) Write() error { return e.batch.Write() }
+
+func (e ethBatchWrapper) Reset() { e.batch.Reset() }
+
+func (e ethBatchWrapper) Replay(w ethdb.KeyValueWriter) error { return e.batch.Replay(w) }
+
+func (e ethBatchWrapper) Inner() avalanchegodb.Batch { return e.batch.Inner() }
```
joshua-kim: Do we need this? we're using this type as the `ethdb.Batch` interface type, so we don't need to implement the avalanchego interface.

vms/evm/database/wrapped_database.go
```diff
@@ -0,0 +1,79 @@
...
+	start = newStart
+
+	return db.db.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+func (db database) NewIteratorWithStart(start []byte) ethdb.Iterator {
+	return db.db.NewIteratorWithStart(start)
+}
+
+type ethBatchWrapper struct {
```
joshua-kim: Maybe we should just call this batch (also we want to be consistent w/ database, since we're not calling it `databaseWrapper`) - It seems self-documenting that this is a wrapper pattern.

vms/evm/database/database.go
```diff
@@ -0,0 +1,79 @@
...
+
+var (
+	errSnapshotNotSupported = errors.New("snapshot is not supported")
+	errStatNotSupported     = errors.New("stat is not supported")
+
+	_ ethdb.Batch         = (*ethBatchWrapper)(nil)
+	_ ethdb.KeyValueStore = (*database)(nil)
+)
+
+type database struct {
```
joshua-kim, line 22: This is production code, so we should have corresponding test coverage for it. eth exposes test suites to check interface implementations through `dbtest.TestDatabaseSuite`.

The problem is that we're implementing the interface but not satisfying its public contract - we error on `Stat` and `Snapshot` (so tests will fail). We could just return a `nil` error for `Stat` and we'd technically satisfy the interface since we could just say that our stat is nothing, but for `Snapshot` we need to somehow freeze the k/v pairs.

In prod code we never actually use `Snapshot`, so we could just iterate over the entire db and copy the k/v pairs into a map and write a wrapper that implements the `Snapshot` interface... and then we'd be able to get test coverage without having to write tests. The implementation would not be suitable for prod but it'd be enough to pass the test suite.

Some other ideas would be to just use the tests and filter out failed tests... but that's brittle because of string matching + you'd have to update any tooling that runs tests so i think this workaround is unusable.

vms/evm/database/database.go
joshua-kim, line 1: nit: I feel like this file name could be better - I don't have a strong opinion on what to name it but `wrapped_*` no longer fits here... maybe something like `database.go` or something.

vms/evm/database/wrapped_database.go
JonathanOppenheimer: It's not dead as it is used in coreth, but yeah, it's a needless layer of abstraction. Removed.

vms/evm/database/database.go
JonathanOppenheimer: This is a very nice approach - I will separate out the test implementation from the actual implementation though and still return the errNotSupported errors on the actual database. I'm not sure if you proposing this, but we really should very explicitly not support this implementation of snapshot for production code and I can imagine someone accidentally using it.

vms/evm/database/database.go
JonathanOppenheimer: done!

vms/evm/database/database.go
```diff
@@ -0,0 +1,73 @@
...
+	start = newStart
+
+	return db.db.NewIteratorWithStartAndPrefix(start, prefix)
+}
+
+type batch struct {
+	batch avalanchegodb.Batch
+}
+
+func (e batch) Put(key, value []byte) error { return e.batch.Put(key, value) }
```
joshua-kim: nit: update receiver

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	database
+}
+
+// NewSnapshot creates a test-only snapshot that captures the current database state.
+// This implementation is designed for test suite compliance only and is NOT suitable
+// for production use as it loads the entire database into memory.
+//
+// Creates a snapshot by iterating over the entire database and copying key-value pairs.
+func (db testDatabase) NewSnapshot() (ethdb.Snapshot, error) {
+
```
joshua-kim: nit: extra whitespace

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+
+	return &testSnapshot{data: snapshotData}, nil
+}
+
+// testSnapshot implements ethdb.Snapshot by storing a copy of the database state.
+// This is a test-only implementation that loads the entire database into memory
+// for the purpose of satisfying the dbtest.TestDatabaseSuite interface requirements.
+// It is NOT suitable for production use due to memory and performance implications.
```
joshua-kim: Same comment here

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+type testSnapshot struct {
+	data map[string][]byte
+}
+
+func (s *testSnapshot) Get(key []byte) ([]byte, error) {
```
joshua-kim: nit: update receiver

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	value, exists := s.data[string(key)]
```
joshua-kim: nit: Use `ok` naming for existence checks

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	if !exists {
+		return nil, avalanchegodb.ErrNotFound
```
joshua-kim: nit: It doesn't make a difference here, but we don't have to use avalanchego's database error here. Any error here works - we could just return some anonymous error like `errors.New("not found")`

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return value, nil
+}
+
+func (s *testSnapshot) Has(key []byte) (bool, error) {
+	_, exists := s.data[string(key)]
```
joshua-kim: same nit here

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database/memdb"
+
+	avalanchegodb "github.com/ava-labs/avalanchego/database"
+)
+
+// testDatabase wraps the production database with test-only snapshot functionality
+type testDatabase struct {
+	database
```
joshua-kim: Why do we wrap the `database` type? I think if we wrap the `eth.KeyValueStore` interface we can avoid the cast we do later.

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (*testSnapshotIterator) Error() error {
+	return nil
+}
+
+func TestInterface(t *testing.T) {
+	dbtest.TestDatabaseSuite(t, func() ethdb.KeyValueStore {
+		baseDB := memdb.New()
+		wrappedDB := New(baseDB)
+		return &testDatabase{database: wrappedDB.(database)}
```
joshua-kim: 1. If we use the interface as per the earlier comment we can remove this cast. 
2. nit: I generally try to avoid creating variables when it's avoidable because it creates more things the reader needs to track. 
```suggestion
		return &testDatabase{KeyValueStore: New(memdb.New())}
```

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+// testDatabase wraps the production database with test-only snapshot functionality
+type testDatabase struct {
+	database
+}
+
+// NewSnapshot creates a test-only snapshot that captures the current database state.
+// This implementation is designed for test suite compliance only and is NOT suitable
+// for production use as it loads the entire database into memory.
```
joshua-kim: We don't have to document this because this is a part of the test - the actual package doesn't actually see this so it's not possible to use in production.

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	return value, nil
+}
+
+func (s *testSnapshot) Has(key []byte) (bool, error) {
+	_, exists := s.data[string(key)]
+	return exists, nil
+}
+
+func (*testSnapshot) Release() {
+	// No cleanup needed for snapshot
```
joshua-kim: nit: This comment is self-documenting. I've heard Arran describe this as don't document what/how comments document "why" . "What" and "how" can be deduced from reading the code (and if it's not readable, the code should be made readable), but "why" is not obvious from reading the code.

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (it *testSnapshotIterator) Value() []byte {
+	if it.index < 0 || it.index >= len(it.pairs) {
+		return nil
+	}
+	return it.pairs[it.index].value
+}
+
+func (*testSnapshotIterator) Release() {
+	// No cleanup needed for snapshot iterator
```
joshua-kim: Same nit here

vms/evm/database/database.go
```diff
@@ -0,0 +1,73 @@
...
+	"errors"
+
+	"github.com/ava-labs/libevm/ethdb"
+
+	avalanchegodb "github.com/ava-labs/avalanchego/database"
+)
+
+var (
+	ErrSnapshotNotSupported = errors.New("snapshot is not supported")
+	ErrStatNotSupported     = errors.New("stat is not supported")
```
joshua-kim: Do these need to be exported?

vms/evm/database/database_test.go
joshua-kim: Same nit here

vms/evm/database/database.go
```diff
@@ -0,0 +1,73 @@
...
+	_ ethdb.KeyValueStore = (*database)(nil)
+)
+
+type database struct {
+	db avalanchegodb.Database
+}
+
+func New(db avalanchegodb.Database) ethdb.KeyValueStore { return database{db} }
+
+func (database) Stat(string) (string, error) { return "", ErrStatNotSupported }
```
joshua-kim: This isn't even a nit - but I have no idea what we should do here. Do we return an error or not? The interface says that `Stat` returns a stat corresponding to the property passed to us... so I'm not sure if returning a `"", nil"` is valid or not. This code is also not used anywhere in geth... and it's also not tested... so I guess both returning an error and `nil` are equally valid and not valid...

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	defer iter.Release()
+
+	for iter.Next() {
+		key := iter.Key()
+		value := iter.Value()
+		keyCopy := make([]byte, len(key))
+		valueCopy := make([]byte, len(value))
+		copy(keyCopy, key)
+		copy(valueCopy, value)
+		snapshotData[string(keyCopy)] = valueCopy
```
joshua-kim: You don't have to copy the key bytes again - `string(key)` already performs a copy of the bytes because we allocate a new string + strings are immutable but byte slices are mutable so Go must create a new string to provide immutability guarantee. Currently we're actually creating two copies.

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,166 @@
...
+	dbtest.TestDatabaseSuite(t, func() ethdb.KeyValueStore {
+		baseDB := memdb.New()
+		wrappedDB := New(baseDB)
+		return &testDatabase{database: wrappedDB.(database)}
+	})
+}
+
+func TestProductionErrors(t *testing.T) {
+	baseDB := memdb.New()
+	wrappedDB := New(baseDB)
```
joshua-kim: We're sharing a db instance across tests - we should either create separate tests or create the db per-sub-test. Also - same nit about minimizing vars, we could do something like `db := New(memdb.New())`

vms/evm/database/database_test.go
JonathanOppenheimer: sure - guess we don't need the avalanchegodb

vms/evm/database/database.go
JonathanOppenheimer: Hmm -- I think at least to me returning an error makes intuitively more sense than `nil`

joshua-kim: Yeah this is just a random thing I noticed 😄 , but I generally don't like hand-wavy explanations and I can't come up with a non-hand-wavy explanation for or against either approach (I also think it "intuitively" looks better + it's consistent/ w how geth implemented their memorydb, but that's not really a technical reason) - so I'm just rambling.

vms/evm/database/database_test.go
JonathanOppenheimer: Without the `copy(keyCopy, key)` I get the below error 

```
Running tool: /Users/jonathan.oppenheimer/.go/bin/go test -timeout 30s -tags test -run ^TestInterface$ github.com/ava-labs/avalanchego/vms/evm/database

--- FAIL: TestInterface (0.00s)
    --- FAIL: TestInterface/Snapshot (0.00s)
        /Users/jonathan.oppenheimer/Desktop/GitHub/avalanchego/vms/evm/database/testsuite.go:339: not found
FAIL
FAIL	github.com/ava-labs/avalanchego/vms/evm/database	0.277s
FAIL
```

I follow what you're saying, but don't know enough about Go to know why that doesn't happen here.

vms/evm/database/database_test.go
JonathanOppenheimer: If you're talking about the receiver, we can't have a receiver that isn't `_` or we fail lint.

joshua-kim: I should clarify better... I meant that the comment was self-documenting in reference to https://github.com/ava-labs/avalanchego/pull/4331/files#r2461442411

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if err := iter.Error(); err != nil {
+		return nil, err
+	}
+
+	return &testSnapshot{data: snapshotData}, nil
+}
+
+// testSnapshot implements ethdb.Snapshot by storing a copy of the database state.
```
joshua-kim: imported packages need brackets to link properly
```suggestion
// testSnapshot implements [ethdb.Snapshot] by storing a copy of the database state.
```

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+func TestInterface(t *testing.T) {
+	dbtest.TestDatabaseSuite(t, func() ethdb.KeyValueStore {
+		return &testDatabase{KeyValueStore: New(memdb.New())}
+	})
+}
+
+func TestProductionErrors(t *testing.T) {
```
joshua-kim: I feel a little weird about this naming - "production" fits oddly because we're always testing production code
```suggestion
func TestUnimplemented(t *testing.T) {
```

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,152 @@
...
+	t.Run("NewSnapshot_ReturnsError", func(t *testing.T) {
+		wrappedDB := New(memdb.New())
+		_, err := wrappedDB.NewSnapshot()
```
joshua-kim: nit: var name feels leaky
```suggestion
		db := New(memdb.New())
		_, err := db.NewSnapshot()
```

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,152 @@
...
+		require.ErrorIs(t, err, errSnapshotNotSupported)
+	})
+
+	t.Run("Stat_ReturnsError", func(t *testing.T) {
+		wrappedDB := New(memdb.New())
+		_, err := wrappedDB.Stat("test")
```
joshua-kim:
```suggestion
		db := New(memdb.New())
		_, err := db.Stat("test")
```

vms/evm/database/database_test.go
```diff
@@ -0,0 +1,152 @@
...
+	value, ok := t.data[string(key)]
+	if !ok {
+		return nil, errors.New("not found")
+	}
+	return value, nil
+}
+
+func (t *testSnapshot) Has(key []byte) (bool, error) {
+	_, exists := t.data[string(key)]
+	return exists, nil
```
joshua-kim:
```suggestion
	_, ok := t.data[string(key)]
	return ok, nil
```

PR #4333, opened 2025-09-24
Remove UptimeManager#IsConnected

joshua-kim:
## Why this should be merged

Slim down interface. Having `IsConnected` on the interface makes some edge-cases look confusing where we check for if the inner uptime manager is in-sync with us, which I'm not sure is possible or not ([ref](https://github.com/ava-labs/subnet-evm/blob/53f53056e9d161ada1ea42506b9537755018715d/plugin/evm/validators/uptime/pausable_manager.go#L40))

## How this works

Removes `IsConnected` from `UptimeManager`

## How this was tested

UT added

No

### DIFF
--- a/snow/uptime/manager.go
+++ b/snow/uptime/manager.go
@@ -30,7 +30,6 @@ type Tracker interface {
 	StartedTracking() bool
 
 	Connect(nodeID ids.NodeID) error
-	IsConnected(nodeID ids.NodeID) bool
 	Disconnect(nodeID ids.NodeID) error
 }
 
--- a/vms/platformvm/service.go
+++ b/vms/platformvm/service.go
@@ -868,7 +868,7 @@ func (s *Service) getPrimaryOrSubnetValidators(subnetID ids.ID, nodeIDs set.Set[
 				if err != nil {
 					return nil, err
 				}
-				isConnected := s.vm.uptimeManager.IsConnected(currentStaker.NodeID)
+				isConnected := s.vm.Network.Peers().Has(currentStaker.NodeID)
 				connected = &isConnected
 				uptime = &currentUptime
 			}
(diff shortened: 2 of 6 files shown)


vms/platformvm/service.go
```diff
@@ -868,7 +868,7 @@ func (s *Service) getPrimaryOrSubnetValidators(subnetID ids.ID, nodeIDs set.Set[
 				if err != nil {
 					return nil, err
 				}
-				isConnected := s.vm.uptimeManager.IsConnected(currentStaker.NodeID)
+				isConnected := s.vm.Network.Peers().Has(currentStaker.NodeID)
```
joshua-kim, line 871: I guess this is now race-y with the uptime information from the `uptimeManager`, but I don't think that we care here?

snow/uptime/manager.go
```diff
@@ -30,7 +30,6 @@ type Tracker interface {
 	StartedTracking() bool
 
 	Connect(nodeID ids.NodeID) error
-	IsConnected(nodeID ids.NodeID) bool
```
ceyonur, line 33: Why we removed `IsConnected` but kept `Connect` and `Disconnect`?(BTW I don't think we can remove any of them) 

The connected here is not equal to peer list. I.e a peer can still be connected in p2p layer but can be benched (which is disconnected for uptime manager).

It might not impose a critical issue for the API layer but it's weird as an API to not provide this. Also if we remove this I don't think any inner manager can be reliable.

vms/platformvm/service.go
ceyonur: It's indeed racey and more than being racey ~~it's not reflect the same outcome with `IsConnected` as a peer could be benched ([here](https://github.com/ava-labs/avalanchego/blob/joshua-kim/remove-is-connected/snow/networking/router/chain_router.go#L568))~~

We have discussed this and seems they're guaranteed to be the same.

ceyonur (approved):

PR #4334, opened 2025-09-25
test: add blockdb config to reexecution tests

DracoLi:
## Why this should be merged

Enable running reexecution tests with blockdb enabled for workflow runs.

The config is added in https://github.com/ava-labs/coreth/pull/1189

## How this works

Enables block database for the c-chain. Block database is disabled by default.

### DIFF
--- a/tests/reexecute/c/vm_reexecute_test.go
+++ b/tests/reexecute/c/vm_reexecute_test.go
@@ -81,6 +81,9 @@ var (
 			"pruning-enabled": true,
 			"state-sync-enabled": false
 		}`,
+		"blockdb": `{
+			"block-database-enabled": true
+		}`,
 	}
 
 	configNameArg  string


aaronbuchwald:
What's the intended path for this PR on the C-Chain?

Fine to add, but I would think we can easily run the re-execution test on a PR to evaluate its performance and decide based on that if we want to pursue it rather than merging it upfront. If the lifecycle of integrating / optimizing it will be longer, then merging support for this config makes sense to me though.

Could also add support for new configs by following up on https://github.com/ava-labs/avalanchego/issues/4108 to support passing arbitrary JSON along with a name to use for the config.

DracoLi:
@aaronbuchwald 
yeah i agree - I don't think i'll need to run this very often right now. I'll close this PR and instead run it locally or on a box for now.
I think https://github.com/ava-labs/avalanchego/issues/4108 is a better improvement so we won't need this once that is done.

aaronbuchwald:
Sounds good, if you find it easiest to re-trigger by pushing to a PR where the config is supported, I'd just open to a draft PR. I am still working on migrating from executing the initial phase of the chain to a much later point (33m) and then to near tip.

To test the blockdb changes, you may want to re-generate that state from scratch, so that the database is not incorrectly much larger due to having written blocks to the original format up to that checkpoint.

aaronbuchwald:
If you want to generate your own checkpoints, I vibe coded this script that I decided against checking in to AvalancheGo: https://gist.github.com/aaronbuchwald/3d222f657c926ebfef44098beea77ab2

PR #4335, opened 2025-09-25
Refactor Warp Verification Tests

StephenButtolph:
## Why this should be merged

This PR separates testing of `FlattenValidatorSet` and `BitSetSignature.Verify` into separate tests (and removes usage of `gomock`).

This PR also added explicit benchmarking of those two functions for usage when repricing warp verification after ACP-181.

## How this works

1. Removed mocking
2. Broke 1 big test into 2 small tests
3. Added 2 new benchmarks

## How this was tested

CI

No

### DIFF
--- a/vms/platformvm/warp/signature_test.go
+++ b/vms/platformvm/warp/signature_test.go
@@ -4,17 +4,14 @@
 package warp
 
 import (
-	"context"
 	"errors"
-	"math"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"go.uber.org/mock/gomock"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/validators"
-	"github.com/ava-labs/avalanchego/snow/validators/validatorsmock"
 	"github.com/ava-labs/avalanchego/utils"
 	"github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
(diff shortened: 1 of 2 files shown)


vms/platformvm/warp/signature_test.go
```diff
@@ -137,757 +134,268 @@ func TestNumSigners(t *testing.T) {
...
-		},
+	unsignedMsg, err := NewUnsignedMessage(
+		constants.UnitTestID,
+		sourceChainID,
+		nil,
+	)
+	require.NoError(t, err)
+	unsignedBytes := unsignedMsg.Bytes()
+
+	sign := func(signers ...bls.Signer) [bls.SignatureLen]byte {
```
joshua-kim, line 145: Wondering if this should live as a test util in a package somewhere?

vms/platformvm/warp/validator_test.go
```diff
@@ -346,3 +348,107 @@ func BenchmarkGetCanonicalValidatorSet(b *testing.B) {
...
+				nodeID := ids.GenerateTestNodeID()
+				secretKey, err := localsigner.New()
+				require.NoError(b, err)
+				publicKey := secretKey.PublicKey()
+				vdrs[nodeID] = &validators.GetValidatorOutput{
+					NodeID:    nodeID,
+					PublicKey: publicKey,
+					Weight:    1,
+				}
+			}
```
joshua-kim, line 447: Why do we do this setup inside of `b.Run`? I thought this counted towards timing/allocation stats. Aren't we supposed to either do this setup before `b.Run`, or call `b.ResetTimer()` before the benchmarking loop?

StephenButtolph: `b.Loop()` handles this for us

joshua-kim: I'm being dumb, maybe I should RTFM 🤦 . The first call to `Loop` resets stats...

joshua-kim (approved):

PR #4336, opened 2025-09-25
Granite ACP-176

alarso16:
## Why this should be merged

Enables ava-labs/coreth#1169 - Capacity refill and gas calculations need to be adjusted to be millisecond based in granite, but keeping the same method of calculation.

Additionally, one of the old ACP-176 tests wasn't testing what it intended to after a parameter change, so the test adds additional comments to ease recalculation in the future if necessary.

## How this works

Converts TargetPerSecond to TargetPerMS (rounding down when necessary) for rate calculations based on the millisecond

## How this was tested

New UT

No.

### DIFF
--- a/vms/evm/acp176/acp176.go
+++ b/vms/evm/acp176/acp176.go
@@ -117,7 +117,7 @@ func (s *State) AdvanceSeconds(seconds uint64) {
 func (s *State) AdvanceMilliseconds(milliseconds uint64) {
 	targetPerSecond := s.Target()
 	targetPerMS := targetPerSecond / 1000
-	maxPerMS := mulWithUpperBound(targetPerMS, TargetToMax)            // R - this can't overflow since 1000 > TargetToMax.
+	maxPerMS := targetPerMS * TargetToMax                              // R - this can't overflow since 1000 > TargetToMax.
 	maxPerSecond := mulWithUpperBound(targetPerSecond, TargetToMax)    // rate used for calculating maxCapacity
 	maxCapacity := mulWithUpperBound(maxPerSecond, TimeToFillCapacity) // C
 	s.Gas = s.Gas.AdvanceTime(
--- a/vms/evm/acp176/acp176.go
+++ b/vms/evm/acp176/acp176.go
@@ -96,9 +96,10 @@ func (s *State) GasPrice() gas.Price {
 	return gas.CalculatePrice(MinGasPrice, s.Gas.Excess, priceUpdateConversion)
 }
 
-// AdvanceTime increases the gas capacity and decreases the gas excess based on
+// AdvanceSeconds increases the gas capacity and decreases the gas excess based on
 // the elapsed seconds.
-func (s *State) AdvanceTime(seconds uint64) {
+// This is used in Fortuna.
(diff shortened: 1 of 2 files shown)


vms/evm/acp176/acp176.go
```diff
@@ -110,6 +111,23 @@ func (s *State) AdvanceTime(seconds uint64) {
 	)
 }
 
+// AdvanceMilliseconds increases the gas capacity and decreases the gas excess based on
+// the elapsed milliseconds.
+// This is used in Granite.
+func (s *State) AdvanceMilliseconds(milliseconds uint64) {
+	targetPerSecond := s.Target()
```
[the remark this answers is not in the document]
alarso16: Is this comment necessary?

vms/evm/acp176/acp176.go
```diff
@@ -110,6 +111,23 @@ func (s *State) AdvanceTime(seconds uint64) {
...
+	targetPerMS := targetPerSecond / 1000
+	maxPerMS := mulWithUpperBound(targetPerMS, TargetToMax)            // R - this can't overflow since 1000 > TargetToMax.
```
StephenButtolph: Feels weird to document that it can't overflow but still check for overflow.
```suggestion
	maxPerMS := targetPerMS * TargetToMax                              // R - this can't overflow since 1000 > TargetToMax.
```

We could actually enforce this at compile time with:
```golang
const (
	// Enforce at compile time that TargetToMax <= 1000 to ensure
	// [State.AdvanceMilliseconds] can not overflow.
	_ uint = 1000 - TargetToMax
)
```
But that might be a bit overkill for this.

vms/evm/acp176/acp176_test.go
```diff
@@ -342,6 +343,167 @@ var (
...
+				},
+				// Set MaxCapacity to MaxUint64
+				// Requires Target >= MaxUint64 / TargetToMaxCapacity
+				// Target in seconds is used for calculating maxCapacity
+				TargetExcess: 947_688_692, // ceil(TargetConversion * ln(MaxUint64 / MinMaxCapacity))
+			},
+			milliseconds: 1000,
+			expected: State{
+				Gas: gas.State{
+					Capacity: 3_689_348_878_490_564_000,                        // greater than MaxUint64/10
```
StephenButtolph: Should we replace all the `/10` with `/ TargetToMaxCapacity`?

vms/evm/acp176/acp176_test.go
```diff
@@ -342,6 +343,167 @@ var (
...
 		},
 	}
+	advanceMillisecondsTests = []struct {
+		name         string
+		initial      State
+		milliseconds uint64
+		expected     State
+	}{
+		{
+			name: "0_seconds",
```
StephenButtolph: I probably would have named these based on `ms` (eg `1000_ms` rather than `1_seconds`) but it's pretty nit picky. I'm fine either way.

vms/evm/acp176/acp176.go
alarso16: Since we won't ever make TargetToMax that big, it does seem overkill - but that does seem like a goo dstrategy

PR #4338, opened 2025-09-25
refactor(metervm): remove mockable clock

RodrigoVillar:
## Why this should be merged

In `metervm`, we have a mockable clock which we use to get the current time. We don't use any of the mock features of the clock and so making a call to `vm.clock.Time()` is equivalent to `time.Now()`, which is easier to understand and reduces code indirection.

## How this works

Replaces all calls to the mockable clock with `time.Now()`.

## How this was tested

CI

No

### DIFF
--- a/vms/metervm/batched_vm.go
+++ b/vms/metervm/batched_vm.go
@@ -31,8 +31,7 @@ func (vm *blockVM) GetAncestors(
 		maxBlocksSize,
 		maxBlocksRetrivalTime,
 	)
-	end := time.Now()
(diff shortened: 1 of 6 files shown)


yacovm (approved):

StephenButtolph (commented):
all of the:
```golang
end := time.Now()
duration := float64(end.Sub(start))
```
Can be replaced with:
```golang
duration := float64(time.Since(start))
```

StephenButtolph (approved):

PR #4339, opened 2025-09-25
Move ledger keychain implementation to dedicated package

felipemadero:
## Why this should be merged
This refactoring improves code organization by separating ledger-specific implementation from the generic keychain interfaces. The current structure mixes generic keychain interfaces with ledger-specific implementation details in the same package, making it harder to maintain and extend. 

## How this works
- Package reorganization: All ledger-related functionality (ledgerKeychain, ledgerSigner, ledger interfaces, tests, and mocks) is moved to utils/crypto/keychain/ledger/
- Interface preservation: The core Signer and Keychain interfaces remain in the main keychain package unchanged
- Ledger device constructor now returns a concrete *Ledger struct while the ledger keychain package defines a Ledger interface for its dependency needs, enabling clean testing and dependency injection

## How this was tested
- unit tests


sukantoraymond (approved):

felipemadero:
ledger implementation to be removed by https://github.com/ava-labs/avalanchego/pull/4413

PR #4341, opened 2025-09-25
ci: composable avalanchego action

Elvis339:
## Why this should be merged
Enables composable CI by allowing repositories to setup AvalancheGo with custom dependency versions. This solves the problem where teams need AvalancheGo context for testing but can't easily compose different repo versions in their CI workflows.

[More context](https://docs.google.com/document/d/1ynQmmt7U_77PzojlckvZcMfBG3wMXA4o64zDGckLwxs/edit?usp=sharing)

https://github.com/ava-labs/avalanchego/blob/6de7ef24da194951a670ad8f3c96999471a15e7f/scripts/setup_firewood.sh#L34

## How this works
Prepares build environment with dependency replacements for custom workflows

## How this was tested
**Test Run**: https://github.com/ava-labs/avalanchego/actions/runs/18103163342/job/51511123516

**Workflow**: https://github.com/ava-labs/avalanchego/blob/1674dfaf1135cc21d63c4a8a6bfbaee6c915b36a/.github/workflows/ci.yml#L285

This test validates the setup action by configuring Firewood v0.0.12 and v0.0.13, then running the c-chain-reexecution benchmark. The setup phase works correctly - the benchmark failure is unrelated and outside this PR's scope.

**Note**: The [setup script](https://github.com/ava-labs/avalanchego/blob/6de7ef24da194951a670ad8f3c96999471a15e7f/scripts/setup_firewood.sh#L34) depends on a custom branch in [Firewood](https://github.com/ava-labs/firewood/blob/composable-ci-action/.github/scripts/build.sh).

No. Internal CI change


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -98,6 +98,7 @@ runs:
         prometheus_password: ${{ inputs.prometheus-password }}
         grafana_dashboard_id: 'Gl1I20mnk/c-chain'
         runtime: "" # Set runtime input to empty string to disable log collection
+        prometheus_url: ${{ inputs.prometheus-push-url }}
```
RodrigoVillar: q: why do we need this diff?

Elvis339: It's one of the required parameters in `run-monitored-tmpnet-cmd` and it used to control starting of metrics collection that has been fixed so this is no longer needed I just need to merge master once that patch gets in.

.github/actions/avalanchego-build-action/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      with:
+        repository: 'ava-labs/avalanchego'
+        ref: ${{ inputs.avalanchego }}
+        path: ${{ inputs.checkout-path }}
+    - name: Setup Go for project
+      uses: ./.github/actions/setup-go-for-project  # Note: If Nix-specific functionality is needed, consumer should install Nix as a prerequisite to this action
+    - name: Setup Firewood FFI
+      if: inputs.firewood != ''
+      id: firewood
+      uses: ava-labs/firewood/.github/actions/build-action@composable-ci-action
```
aaronbuchwald: Where is this action defined? I don't see it on Firewood main

.github/actions/avalanchego-build-action/action.yml
```diff
@@ -0,0 +1,120 @@
...
+        fi
+
+        # Replace LibEVM if provided and not default
+        if [ "${{ inputs.libevm }}" != "" ] && [ "${{ inputs.libevm }}" != "main" ]; then
+          echo "Replacing LibEVM with local checkout: ../libevm"
+          go mod edit -replace github.com/ava-labs/libevm=../libevm
+        fi
+
+        go mod tidy
+        go mod download
```
aaronbuchwald: Won't this leave a diff in the resulting checked out code that may cause future actions to find an unexpected diff and fail? Should that be documented or cleaned up?

.github/actions/avalanchego-build-action/action.yml
```diff
@@ -0,0 +1,120 @@
...
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Checkout AvalancheGo
+      uses: actions/checkout@v4
+      with:
+        repository: 'ava-labs/avalanchego'
+        ref: ${{ inputs.avalanchego }}
+        path: ${{ inputs.checkout-path }}
```
aaronbuchwald: Why does AvalancheGo use checkout path and everything else just uses the name of the repo it's checking out?

Would it make sense to take in one parameter to use as a working directory and then check out each required remote as a subdirectory?

.github/actions/avalanchego-build-action/action.yml
```diff
@@ -0,0 +1,120 @@
...
+        # Replace Coreth if provided and not default
+        if [ "${{ inputs.coreth }}" != "" ] && [ "${{ inputs.coreth }}" != "master" ]; then
+          echo "Replacing Coreth with local checkout: ../coreth"
+          go mod edit -replace github.com/ava-labs/coreth=../coreth
+        fi
+
+        # Replace LibEVM if provided and not default
+        if [ "${{ inputs.libevm }}" != "" ] && [ "${{ inputs.libevm }}" != "main" ]; then
+          echo "Replacing LibEVM with local checkout: ../libevm"
+          go mod edit -replace github.com/ava-labs/libevm=../libevm
```
aaronbuchwald: Given we are checking out versions that are available in GitHub, we can just use go get with the updated version rather than checking them each out locally. If we wanted to checkout and compile a specific version of Firewood, we'd need the source code, but currently this is just using the published golang ffi bindings of Firewood as well.

.github/actions/avalanchego-build-action/action.yml
Elvis339: It's defined [here](https://github.com/ava-labs/firewood/blob/composable-ci-action/.github/actions/build-action/action.yml) `composable-ci-action` branch. I used it to test these things can work.

.github/actions/avalanchego-build-action/action.yml
Elvis339: Yes it will, it should probably be cleaned up. Will handle that case.

.github/actions/avalanchego-build-action/README.md
```diff
@@ -0,0 +1,101 @@
...
+  with:
+    checkout-path: "build/avalanchego"
+    coreth: "my-feature-branch"
+    libevm: "experimental-branch"
+
+- name: Run custom build commands
+  run: |
+    cd build/avalanchego
+    ./scripts/run_task.sh reexecute-cchain-range
+    ./scripts/run_task.sh my-custom-task
```
aaronbuchwald: This would not work because the current task definition uses `go run`, so it would not use the the compiled test binary.

.github/actions/avalanchego-build-action/action.yml
Elvis339: That's good input, I'll run `go get` instead of `go mod edit` and repo checkout.

The Firewood workflow handles building form source and using pre-built FFI. 
https://github.com/ava-labs/firewood/blob/composable-ci-action/.github/actions/build-action/action.yml

.github/actions/avalanchego-build-action/action.yml
aaronbuchwald: This is a bit hard to verify as a reviewer since it's not on main and doesn't have a reference to the branch where it's available. How does this work for the workflow runs linked in the PR?

.github/actions/avalanchego-build-action/action.yml
aaronbuchwald: I think we could probably do less here and only support setup rather than adding a script that builds different binaries once setup has completed.

.github/actions/avalanchego-build-action/action.yml
Elvis339: Added link to Firewood action in PR description, thanks for pointing it out.

This [run](https://github.com/ava-labs/avalanchego/actions/runs/18034362817?pr=4341) is using Firewood action along with AvalancheGo to run reexecution test on self-hosted runners. You can find the commit [here](https://github.com/ava-labs/avalanchego/blob/fb8a6cf11fb3c4ae84e2c38f35589db3e4c6258b/.github/workflows/ci.yml#L285). (I'll attach commits in the PR description as well).

The workflow in the [commit](https://github.com/ava-labs/avalanchego/blob/fb8a6cf11fb3c4ae84e2c38f35589db3e4c6258b/.github/workflows/ci.yml#L285) proves that:
1. Firewood build action is:
     - Correctly building from [source (random commit)](https://github.com/ava-labs/avalanchego/blob/fb8a6cf11fb3c4ae84e2c38f35589db3e4c6258b/.github/workflows/ci.yml#L316) and exposing FFI for AvalancheGo build action to link
     - Correctly expose [pre-built binary](https://github.com/ava-labs/avalanchego/blob/fb8a6cf11fb3c4ae84e2c38f35589db3e4c6258b/.github/workflows/ci.yml#L285)
2. AvalancheGo build action is linking dependencies and running reexecution benchmark in composable way

.github/actions/avalanchego-build-action/action.yml
Elvis339: The avalanchego checkout path is the main workspace where we build the final binary. It needs to be configurable because consumers might want to integrate it into existing directory structures.

Firewood has some storage requirements its database operations benefit from NVMe for performance, but compilation can happen anywhere. The Firewood action handles its own optimal storage detection internally.
A single working directory would force all components to use the same storage type. 

The current approach lets each component use optimal storage:
- Firewood runtime: NVMe when available (handled by its action)  
- AvalancheGo: Standard storage (sufficient for compilation)
- Coreth/LibEVM: Handled by Go's dependency system

.github/actions/avalanchego-build-action/README.md
Elvis339: It works here.

- [Workflow](https://github.com/ava-labs/avalanchego/blob/dfa8c8164e7578d2bdec93a5c1595dffcdf1732e/.github/workflows/ci.yml#L285)
- [Run](https://github.com/ava-labs/avalanchego/actions/runs/18019895212/job/51274339039)

maru-ava (commented):
Why was this not implemented as a task that could be executed both locally and in CI? A task has the advantage of being easier to implement and maintain - since iteration doesn't require cycling through CI - and then it would be possible to trivially reproduce a binary without CI.

Elvis339:
> Why was this not implemented as a task that could be executed both locally and in CI?

The Action approach was chosen for cross-repository consumption. But now that you mention Task I guess it could be made to to run locally and in CI and be exposed as GH action for cross-repository consumption.

.github/actions/avalanchego-build-action/action.yml
```diff
@@ -0,0 +1,120 @@
...
+  coreth:
+    description: 'Coreth version (commit SHA, branch name, or tag)'
+    required: false
+    default: ''
+  libevm:
+    description: 'LibEVM version (commit SHA, branch name, or tag)'
+    required: false
+    default: ''
+
+outputs:
```
JuanLeon2: Can we also output the artifact name (rather than solely relying on the convention described in README.md)

Elvis339:
Switched to task that can execute both locally and in CI.

- Extended `build.sh` with `-f` option signaling Firewood version
- Added new [script](https://github.com/ava-labs/avalanchego/blob/6de7ef24da194951a670ad8f3c96999471a15e7f/scripts/setup_firewood.sh) `setup_firewood.sh` 
- Refactored CI to only support setting Avalanchego without building targets

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,53 @@
...
+    default: ''
+  libevm:
+    description: 'LibEVM version (commit SHA, branch name, or tag)'
+    required: false
+    default: ''
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Checkout AvalancheGo
```
maru-ava, line 33: This is the only non-task step that I would expect to see in an action with the remain steps being part of a task.

maru-ava (commented):
Given that coreth is the integration point for firewood, and coreth is only compatible with specific versions of avalanchego, I'm not sure this open-ended approach (supporting independent configuration of the versions of firewood, coreth and avalanchego) makes sense.

I'm also not sure why avalanchego should be the integration point. Coreth already runs avalanchego e2e tests for its pinned version of avalanchego. So maybe that same mechanism can be extended to support a coreth job for each of the targeted versions of firewood?

Elvis339:
>Given that coreth is the integration point for firewood, and coreth is only compatible with specific versions of avalanchego, I'm not sure this open-ended approach (supporting independent configuration of the versions of firewood, coreth and avalanchego) makes sense.

>I'm also not sure why avalanchego should be the integration point. Coreth already runs avalanchego e2e tests for its pinned version of avalanchego. So maybe that same mechanism can be extended to support a coreth job for each of the targeted versions of firewood?

Thanks for the feedback. Let me clarify why I chose avalanchego as the integration point:

The c-reexecution-benchmark tool lives in avalanchego and needs to test firewood
directly. If we configure this through coreth:

1. We'd update firewood version in coreth
2. Then update coreth in avalanchego
3. Then run the benchmark in avalanchego

This adds an extra dependency hop when the benchmark tool is already in avalanchego
and could consume firewood directly.

As you mentioned, e2e tests exist in coreth but they're run from the context of avalanchego. My idea is that each of these libs (firewood, coreth) provides a way to build and setup themselves, and then avalanchego (or others) becomes the "lego builder" - composing these components as needed rather than going through the
dependency chain.

Does that reasoning make sense, or am I missing something about how coreth's e2e
infrastructure could handle this use case?

_edit by maru-ava: sorry, I accidentally edited this when attempting to reply._

.github/actions/avalanchego-setup-action/action.yml
Elvis339: That's a fair point. I can see the appeal of consolidating all go.mod updates in one place.

I'm thinking of keeping coreth/libevm updates explicit in the action makes the dependency changes visible and optional. They're simple `go get` commands, so putting  them in setup_firewood.sh would add context/coupling without clear benefit - the script would need to handle optional arguments for dependencies it doesn't own.

That said, I'm not strongly opposed to consolidating them. Do you see a clear advantage to moving them into the script? Or is this more about having a consistent pattern for where dependency management happens?

Happy to restructure if there's a compelling reason I'm missing.

scripts/setup_firewood.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if [ -d "${FIREWOOD_CLONE_DIR}" ]; then
+  echo "Removing existing Firewood directory..." >&2
+  rm -rf "${FIREWOOD_CLONE_DIR}"
+fi
+
+echo "Setting up Firewood FFI version: ${FIREWOOD_VERSION}" >&2
+echo "Using workspace: ${WORKSPACE_PATH}" >&2
+
+git clone https://github.com/ava-labs/firewood "${FIREWOOD_CLONE_DIR}" \
```
maru-ava, line 43: Why is it desirable to clone just to get the file rather than using a [raw url](https://raw.githubusercontent.com/ava-labs/firewood/refs/heads/main/benchmark/setup-scripts/build-firewood.sh)? And then the `build-firewood.sh` script will go off and do a full clone anyway.

I'm not sure why this would be at all preferable to just [building with nix](https://github.com/ava-labs/firewood/pull/1319).

scripts/setup_firewood.sh
```diff
@@ -0,0 +1,77 @@
...
+SETUP_FIREWOOD_SCRIPT="${FIREWOOD_CLONE_DIR}/benchmark/setup-scripts/build-firewood.sh"
+
+if [ ! -f "${SETUP_FIREWOOD_SCRIPT}" ]; then
+  echo "Error: Setup Firewood script not found at ${SETUP_FIREWOOD_SCRIPT}" >&2
+  exit 1
+fi
+
+# Build or fetch Firewood FFI with custom workspace
+# Capture only the last line which is the FFI path
+FFI_PATH=$("${SETUP_FIREWOOD_SCRIPT}" "${FIREWOOD_VERSION}" --workspace "${WORKSPACE_PATH}" | tail -n 1)
```
maru-ava, line 55: It does not appear that the target script accepts a version. afaik only way to build a specific version would be to clone at said version. Is there an updated script that hasn't yet merged to firewood main?

scripts/setup_firewood.sh
```diff
@@ -0,0 +1,77 @@
...
+  exit 1
+fi
+
+echo "Updating go.mod with FFI path: ${FFI_PATH}" >&2
+go mod edit -replace github.com/ava-labs/firewood-go-ethhash/ffi="${FFI_PATH}"
+
+go mod tidy
+go mod download
+
+# Output FFI path to stdout for consumption by other scripts
```
maru-ava, line 76: Are you intending that the script call itself? I don't see any other case of tailing the output to get the path. Maybe just assume `$PWD/firewood/ffi`?

Taskfile.yml
```diff
@@ -11,6 +11,12 @@ tasks:
     desc: Builds avalanchego
     cmd: ./scripts/build.sh
 
+  build-with-firewood:
+    desc: Builds avalanchego with Firewood FFI (specify version with FIREWOOD_VERSION)
+    vars:
+      FIREWOOD_VERSION: '{{.FIREWOOD_VERSION | default "ffi/v0.0.12"}}'
```
maru-ava, line 20: Maybe make it clear that what is being accepted here is any of `branch|tag|SHA`?

scripts/setup_firewood.sh
```diff
@@ -0,0 +1,77 @@
...
+cd "${AVALANCHE_PATH}"
+
+# Verify go.mod exists
+if [ ! -f "go.mod" ]; then
+  echo "Error: go.mod not found in ${AVALANCHE_PATH}" >&2
+  exit 1
+fi
+
+echo "Updating go.mod with FFI path: ${FFI_PATH}" >&2
+go mod edit -replace github.com/ava-labs/firewood-go-ethhash/ffi="${FFI_PATH}"
```
maru-ava, line 71: It is not ideal that after I execute this script, my working tree is now dirty and I have to make sure I don't accidentally commit the change.

.github/actions/avalanchego-setup-action/action.yml
maru-ava: The pattern for the majority of our CI jobs is that they are trivial to reproduce locally. That means, apart from CI-specific setup like cloning, caching and installing non-nix deps, the meat of a given job is encapsulated in a task that can be invoked locally. CI jobs are then just the glue between Github Actions and our tasks. This ensures both reproducibility and simplifies development and maintenance of our jobs.

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,61 @@
...
+      working-directory: ${{ inputs.checkout-path }}
+      run: |
+        echo "Updating Coreth to version: ${{ inputs.coreth }}"
+        go get github.com/ava-labs/coreth@${{ inputs.coreth }}
+    - name: Update LibEVM dependency
+      if: inputs.libevm != ''
+      shell: bash
+      working-directory: ${{ inputs.checkout-path }}
+      run: |
+        echo "Updating LibEVM to version: ${{ inputs.libevm }}"
```
maru-ava, line 51: What guarantees are there that a given version of libevm is going to be compatible with given versions of avalanchego and coreth?

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        repository: 'ava-labs/avalanchego'
+        ref: ${{ inputs.avalanchego }}
+        path: ${{ inputs.checkout-path }}
+    - name: Update Coreth dependency
+      if: inputs.coreth != ''
+      shell: bash
+      working-directory: ${{ inputs.checkout-path }}
+      run: |
+        echo "Updating Coreth to version: ${{ inputs.coreth }}"
+        go get github.com/ava-labs/coreth@${{ inputs.coreth }}
```
maru-ava, line 45: What guarantees are there that a given version of coreth is going to be compatible with given versions of avalanchego or firewood?

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,61 @@
+name: 'Setup AvalancheGo'
+description: 'Setup AvalancheGo with custom dependencies (Firewood, Coreth, LibEVM)'
+
+inputs:
+  avalanchego:
+    description: 'AvalancheGo version (commit SHA, branch name, or tag)'
+    required: false
+    default: ${{ github.sha }}
```
maru-ava, line 8: This reference only makes sense if the custom action will only be used in the avalanchego repo.

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,61 @@
...
+  checkout-path:
+    description: 'Path where AvalancheGo will be checked out'
+    required: false
+    default: '.'
+  firewood:
+    description: 'Firewood version (commit SHA, branch, tag, or ffi/vX.Y.Z for pre-built)'
+    required: false
+    default: ''
+  firewood-path:
```
maru-ava, line 17: Under what circumstances will this actually be used?

.github/actions/avalanchego-setup-action/action.yml
```diff
@@ -0,0 +1,61 @@
...
+      working-directory: ${{ inputs.checkout-path }}
+      run: |
+        echo "Updating LibEVM to version: ${{ inputs.libevm }}"
+        go get github.com/ava-labs/libevm@${{ inputs.libevm }}
+    - name: Setup Firewood FFI
+      if: inputs.firewood != ''
+      shell: bash
+      run: |
+        if [ -n "${{ inputs.firewood-path }}" ]; then
+          "${{ inputs.checkout-path }}/scripts/setup_firewood.sh" "${{ inputs.firewood }}" "${{ inputs.firewood-path }}"
```
maru-ava: Counter to my insistence on a task being the preferred entrypoint, the only way this custom action is going to be able to be used in other repos isif they have this same script in the same location, or the script is in the action dir and referred to with a relative path. Which doesn't preclude the use of a task, said task would just need to point to the updated script path.

.github/actions/avalanchego-setup-action/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Security Model
+- Repository Restriction: Only allows dependencies from `github.com/ava-labs/*` repositories
+- No Custom Forks: Prevents supply chain attacks from malicious forks
+- Commit Validation: Validates dependency versions reference ava-labs repositories only
+
+## Inputs
+
+| Input | Description | Required | Default               |
+|-------|-------------|----------|-----------------------|
+| `checkout-path` | Directory path where AvalancheGo will be checked out | No | `'.'`                 |
```
maru-ava, line 24: Why is this suggested?

.github/actions/avalanchego-setup-action/README.md
```diff
@@ -0,0 +1,44 @@
...
+| `avalanchego` | AvalancheGo version (commit SHA, branch, tag) | No | `'${{ github.sha }}'` |
+| `firewood` | Firewood version. Consumer should run Firewood shared workflow first | No | `''`                  |
```
maru-ava, line 26: Where's the firewood dir?

maru-ava:
> This adds an extra dependency hop when the benchmark tool is already in avalanchego 
and could consume firewood directly.

It doesn't add a dependency hop that isn't already there. coreth depends directly on both firewood and avalanchego. coreth and avalanchego are always mutually pinned in their master branches. Updating the pinning is a pain (ask any coreth maintainer), and ideally firewood benchmarking would be performed independently of changes to the pinning to minimize the pain.

iirc @aaronbuchwald has suggested moving the re-execution testing to the coreth repo? That could help minimize the work involved in testing firewood, since then the most recent version of the testing would always be available to run against a compatible version of firewood. Otherwise, being able to apply an updated test to a new version of firewood could require updating the pinning between avalanchego and coreth and that can be painful when unrelated breaking changes have been made on the avalanchego side.

The potential for the master branch of avalanchego containing breaking changes for coreth is why I'm not convinced of the wisdom of being able to configure SHAs of all 3 repos. I would expect instead to see avalanchego capable of testing against whatever version of coreth it's pinned to and a compatible version of firewood, coreth being able to test with whatever version of avalanchego it's pinned to and a compatible version of firewood, and firewood being able to test with a compatible version of coreth and its pinned version of avalanchego.

> As you mentioned, e2e tests exist in coreth but they're run from the context of avalanchego. My idea is that each of these libs (firewood, coreth) provides a way to build and setup themselves, and then avalanchego (or others) becomes the "lego builder" - composing these components as needed rather than going through the 
dependency chain.

> Does that reasoning make sense, or am I missing something about how coreth's e2e 
infrastructure could handle this use case?

I think your reasoning only makes sense if one ignores the potential for avalanchego introducing changes that break coreth (between updates to mutual pinning). I've been frustrated enough times trying to make cross-repo CI changes that I think that potential is worth considering. 

The e2e job of which you speak validates that a change to coreth is compatible with the currently pinned version of avalanchego. An update to the version of firewood could be such a change. The machinery involved could even be reused to enable firewood testing e.g. build ffi, clone coreth, go mod replace for firewood ffi, run coreth script to clone the appropriate version of avalanchego and go mod replace for coreth, and then run the desired test from the avalanchego repo.

scripts/setup_firewood.sh
Elvis339: Thanks for the suggestion. We can leverage [building with nix](https://github.com/ava-labs/firewood/pull/1319) once it's merged.

Elvis339:
> > This adds an extra dependency hop when the benchmark tool is already in avalanchego
> > and could consume firewood directly.
> 
> It doesn't add a dependency hop that isn't already there. coreth depends directly on both firewood and avalanchego. coreth and avalanchego are always mutually pinned in their master branches. Updating the pinning is a pain (ask any coreth maintainer), and ideally firewood benchmarking would be performed independently of changes to the pinning to minimize the pain.
> 
> iirc @aaronbuchwald has suggested moving the re-execution testing to the coreth repo? That could help minimize the work involved in testing firewood, since then the most recent version of the testing would always be available to run against a compatible version of firewood. Otherwise, being able to apply an updated test to a new version of firewood could require updating the pinning between avalanchego and coreth and that can be painful when unrelated breaking changes have been made on the avalanchego side.
> 
> The potential for the master branch of avalanchego containing breaking changes for coreth is why I'm not convinced of the wisdom of being able to configure SHAs of all 3 repos. I would expect instead to see avalanchego capable of testing against whatever version of coreth it's pinned to and a compatible version of firewood, coreth being able to test with whatever version of avalanchego it's pinned to and a compatible version of firewood, and firewood being able to test with a compatible version of coreth and its pinned version of avalanchego.
> 
> > As you mentioned, e2e tests exist in coreth but they're run from the context of avalanchego. My idea is that each of these libs (firewood, coreth) provides a way to build and setup themselves, and then avalanchego (or others) becomes the "lego builder" - composing these components as needed rather than going through the
> > dependency chain.
> 
> > Does that reasoning make sense, or am I missing something about how coreth's e2e
> > infrastructure could handle this use case?
> 
> I think your reasoning only makes sense if one ignores the potential for avalanchego introducing changes that break coreth (between updates to mutual pinning). I've been frustrated enough times trying to make cross-repo CI changes that I think that potential is worth considering.
> 
> The e2e job of which you speak validates that a change to coreth is compatible with the currently pinned version of avalanchego. An update to the version of firewood could be such a change. The machinery involved could even be reused to enable firewood testing e.g. build ffi, clone coreth, go mod replace for firewood ffi, run coreth script to clone the appropriate version of avalanchego and go mod replace for coreth, and then run the desired test from the avalanchego repo.

Your argument is that testing firewood variations from coreth's context provides better isolation from the avalanchego ↔ coreth synchronization pain. Since updating the pinning between avalanchego and coreth is frequently painful (especially when unrelated breaking changes exist), testing firewood from coreth means we only need to deal with firewood version changes while relying on coreth's already-validated avalanchego pin. This minimizes friction because we're varying one dimension (firewood) instead of potentially coordinating compatibility across all three repos.

Is that an accurate representation of your concern? Correct me if I'm wrong.

---

I think your approach is valid, and I want to explain why I'm proposing avalanchego as the integration point despite the challenges you've raised.

The current reality is that both `load-test` and `c-chain-reexecution` benchmarks live in avalanchego. Avalanchego has organically become the testing context where realistic integration testing happens - not by design, but because it's where dependency resolution naturally converges.

My goal is to enable teams (particularly the Firewood team) to easily run these existing tests without writing code or navigating the dependency chain. For example, from Firewood's CI (pseudocode):

```yaml
- uses: ava-labs/setup-avalanchego@v1
  with:
    firewood-version: v0.0.13

- uses: ava-labs/load-test@v1
  with:
    duration: 10m
```

The vision is composability: each component (firewood, coreth, avalanchego) provides standardized setup/build scripts, and any repo can compose versions as needed. Since the test tools already exist in avalanchego, this approach allows teams to reuse them without duplicating infrastructure or coordination overhead.

To properly address your concerns, we'd likely need a broader plan to move these tests to more appropriate homes (perhaps coreth for firewood integration testing, or dedicated test harnesses). But that's a larger refactoring effort.

Does this help clarify my reasoning? I'm open to discussing whether the composability benefits outweigh the compatibility risks you've identified, or whether we should prioritize moving the tests first.

joshua-kim:
Closing. If this was closed in error please re-open.

PR #4342, opened 2025-09-25
Rebase validator diffs by height

geoff-vball:
## Why this should be merged
Creates a new index for the validator diffs DB. We now key diffs by `height, subnetID, nodeID` in addition to the previous index of `subnetID, height, nodeID`. This will allow us to apply diffs for all validators of all subnets at a given height.

N/A

### DIFF
--- a/snow/validators/manager.go
+++ b/snow/validators/manager.go
@@ -95,6 +95,9 @@ type Manager interface {
 	Sample(subnetID ids.ID, size int) ([]ids.NodeID, error)
 
 	// Map of the validators in this subnet
+	GetAllMaps() map[ids.ID]map[ids.NodeID]*GetValidatorOutput
+
+	// Map of the validators in this subnet
 	GetMap(subnetID ids.ID) map[ids.NodeID]*GetValidatorOutput
 
 	// When a validator is added, removed, or its weight changes, the listener
@@ -272,6 +275,17 @@ func (m *manager) Sample(subnetID ids.ID, size int) ([]ids.NodeID, error) {
 	}
 
 	return set.Sample(size)
+}
+
+func (m *manager) GetAllMaps() map[ids.ID]map[ids.NodeID]*GetValidatorOutput {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	set := make(map[ids.ID]map[ids.NodeID]*GetValidatorOutput, len(m.subnetToVdrs))
+	for subnetID, vdrs := range m.subnetToVdrs {
+		set[subnetID] = vdrs.Map()
+	}
+	return set
 }
 
 func (m *manager) GetMap(subnetID ids.ID) map[ids.NodeID]*GetValidatorOutput {
--- a/vms/platformvm/state/disk_staker_diff_iterator.go
+++ b/vms/platformvm/state/disk_staker_diff_iterator.go
@@ -39,17 +39,17 @@ func marshalStartDiffKey(subnetID ids.ID, height uint64) []byte {
 	return key
 }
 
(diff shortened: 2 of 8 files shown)


snow/validators/manager.go
```diff
@@ -274,6 +277,17 @@ func (m *manager) Sample(subnetID ids.ID, size int) ([]ids.NodeID, error) {
 	return set.Sample(size)
 }
 
+func (m *manager) GetAllMaps() map[ids.ID]map[ids.NodeID]*GetValidatorOutput {
```
StephenButtolph, line 263: Can we add a test for this function?

vms/platformvm/state/disk_staker_diff_iterator.go
StephenButtolph, line 1: Can we add tests similar to the tests for the prior functions?

vms/platformvm/state/state.go
```diff
@@ -67,33 +67,35 @@ var (
...
+	CurrentPrefix                         = []byte("current")
+	PendingPrefix                         = []byte("pending")
+	ValidatorPrefix                       = []byte("validator")
+	DelegatorPrefix                       = []byte("delegator")
+	SubnetValidatorPrefix                 = []byte("subnetValidator")
+	SubnetDelegatorPrefix                 = []byte("subnetDelegator")
+	ValidatorWeightDiffsPrefix            = []byte("flatValidatorDiffs")
+	ValidatorWeightDiffsByHeightPrefix    = []byte("flatValidatorDiffsByHeight")
+	ValidatorPublicKeyDiffsPrefix         = []byte("flatPublicKeyDiffs")
+	ValidatorPublicKeyDiffsByHeightPrefix = []byte("flatPublicKeyDiffsByHeight")
```
StephenButtolph: Can we add these to the state layout comment above the `type state struct {` definition?

vms/platformvm/state/state.go
```diff
@@ -209,6 +211,42 @@ type State interface {
...
+	//
+	// Note: Because this function iterates towards the genesis, [startHeight]
+	// will typically be greater than or equal to [endHeight]. If [startHeight]
+	// is less than [endHeight], no diffs will be applied.
+	ApplyAllValidatorPublicKeyDiffs(
+		ctx context.Context,
+		validators map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput,
+		startHeight uint64,
+		endHeight uint64,
+	) error
```
StephenButtolph, line 248: Can we add tests for these in this package? (see: `TestState_ApplyValidatorDiffs`)

I believe there is currently a bug, so hopefully this will act as a regression test 🙂

StephenButtolph (approved):

vms/platformvm/validators/manager.go
```diff
@@ -288,6 +372,14 @@ func (m *manager) makeValidatorSet(
 	return validatorSet, currentHeight, err
 }
 
+func (m *manager) getAllCurrentValidatorSets(
+	ctx context.Context,
+) (map[ids.ID]map[ids.NodeID]*validators.GetValidatorOutput, uint64, error) {
+	subnetsMap := m.cfg.Validators.GetAllMaps()
+	currentHeight, err := m.getCurrentHeight(ctx)
+	return subnetsMap, currentHeight, err
```
michaelkaplan13, line 380: I'm assuming it's impossible for the height to advance in between these calls to `GetAllMaps` and `getCurrentHeight`, but could someone confirm/explain why?

StephenButtolph: This function is called with the P-chain context lock held

PR #4344, opened 2025-09-25
Move EthKeychain interface to utils/crypto/keychain package

felipemadero:
## Why this should be merged
This refactoring improves code organization by centralizing keychain interfaces in a single package. Currently, the EthKeychain interface is defined in wallet/chain/c, which creates an awkward dependency where core keychain functionality is scattered across different packages. This separation makes it unclear that keychain implementations should provide both AVAX and Ethereum address support.  

## How this works
- Interface relocation: The EthKeychain interface with GetEth(addr common.Address) (Signer, bool) and EthAddresses() set.Set[common.Address] methods is moved to the main keychain package

## How this was tested
- Unit tests


yacovm (dismissed):

StephenButtolph:
Isn't it supposed to be preferred to define the interface where it is used, rather than next to the struct that defines it?

I know that isn't necessarily what we have done in avalanchego in the past, but this PR seems to be pushing us back into a bad pattern (rather than pushing us towards a cleaner pattern).

felipemadero:
> Isn't it supposed to be preferred to define the interface where it is used, rather than next to the struct that defines it?

You're right about the Go convention. Looking at usage, EthKeychain is primarily used in wallet/chain/c (signer) and wallet/subnet/primary (wallet creation). I moved it to match the existing Keychain pattern and also wanted to provide a single point of reference for keychain implementations (ledger, cubesigner, secp256k1fx). This makes it easier to see what interfaces need to be implemented when adding new keychain types IMO. But given the more limited usage compared to Keychain, would you prefer it back in wallet/chain/c where it's most heavily used?

felipemadero:
Unneeded Feature

PR #4345, opened 2025-09-25
use only relevant package to subnet-evm

ceyonur:
## Why this should be merged

The check was too strict, it should only cover those potentially can be imported from subnet-evm

## How this works

relaxes the test by only checking:

* direct `tests` package
* direct `tests/antithesis` package
* `tests/fixture` and subpackages

## How this was tested

UT

no

### DIFF
--- a/tests/imports_test.go
+++ b/tests/imports_test.go
@@ -15,13 +15,16 @@ import (
 func TestMustNotImport(t *testing.T) {
 	require := require.New(t)
 
+	illegalPaths := []string{
+		"github.com/ava-labs/coreth/params",
+		"github.com/ava-labs/coreth/plugin/evm/customtypes",
+	}
 	mustNotImport := map[string][]string{
 		// Importing these packages configures libevm globally. This must not be
 		// done to support both coreth and subnet-evm.
-		"tests/...": {
-			"github.com/ava-labs/coreth/params",
-			"github.com/ava-labs/coreth/plugin/evm/customtypes",
-		},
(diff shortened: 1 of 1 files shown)


tests/imports_test.go
```diff
@@ -15,13 +15,16 @@ import (
 func TestMustNotImport(t *testing.T) {
```
maru-ava, line 22: Please add a docstring that explains the intent behind this test. Otherwise it's not obvious to me how to review changes and I am guessing I am not the only one. We should only be approving what we understand.

ceyonur: There was a inline comment that explained it, I moved it to above.

aaronbuchwald: Could you include in the comment as you explained on Slack why a test is needed here as opposed to just letting it fail loudly when attempting to double register extras?

maru-ava: Can you elaborate on what 'configure libevm globally' means? And what's stopping a transitory import (i.e. an e2e test imports an avalanchego package that imports libevm) from 'configuring libevm globally'?

ceyonur: * importing `libevm` is fine, the problem is importing certain packages that use libevm globals like `coreth/params`, `coreth/plugin/evm/customtypes`
* Transitory imports can definitely happen, and this test actually checks them through `packages.GetDependencies`.

maru-ava: That's helpful, thank you. Please add those details to the docstring.

PR #4347, opened 2025-09-25
Update ledger dependencies to v1.1.0

felipemadero:
## Why this should be merged
- The previous ledger dependency no longer works with Avalanche Ledger app v1.3.7; adopting v1.1.0 restores support for current Ledger firmware.
- Upstream has moved the Go package into its own repository (github.com/ava-labs/ledger-avalanche-go)

## How this works
- Updates utils/crypto/ledger/ledger.go to import the module from the new standalone repository.
- Refreshes go.mod/go.sum so the project depends on ledger-avalanche-go v1.1.0


tests/fixture/bootstrapmonitor/bootstrap_test_config_test.go
```diff
@@ -299,7 +299,7 @@ func TestStateSyncEnabledFromEnvVars(t *testing.T) {
 			chainConfigContent: marshalAndEncode(t,
 				map[string]chains.ChainConfig{
 					"C": {
-						Config: []byte("{\"state-sync-enabled\":1234}"),
+						Config: []byte("{\"state-sync-enabled\":{}}"),
```
StephenButtolph, line 302: Why is this changed?

go.mod
```diff
@@ -88,6 +88,15 @@ require (
...
+require (
+	github.com/cenkalti/backoff/v5 v5.0.2 // indirect
+	github.com/go-viper/mapstructure/v2 v2.3.0 // indirect
+	github.com/sagikazarmark/locafero v0.9.0 // indirect
+	github.com/sourcegraph/conc v0.3.0 // indirect
+	github.com/zondax/golem v0.27.0 // indirect
+	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+)
+
 require (
```
StephenButtolph: super pedantic, but why do are add a new require block for this? We currently have 1 require block for direct and 1 for indirect dependencies
```suggestion
```

tests/fixture/bootstrapmonitor/bootstrap_test_config_test.go
maru-ava: The updated version of the cast library is more permissive when it comes to boolean input. Previously, a numeric input was rejected with an error, but now a non-zero input is considered `true`. I chose `{}` (empty object in json) to trigger an error condition for a boolean cast after reviewing the cast implementation.

go.mod
felipemadero: not pedantic at all :)

StephenButtolph (approved):

PR #4348, opened 2025-09-25
Add retry logic on top of ledger dependency update

felipemadero:
## Why this should be merged
- Ledger operations against the moved ledger-avalanche-go module still fail sporadically with hidapi: unknown failure and APDU 0x6987 errors.
- Introducing a bounded retry with backoff around device calls restores stability when the hardware wallet needs a moment to respond.
  
## How this works
- Adds retryOnHIDAPIError, which retries selected ledger calls up to five times with a linear backoff.
- Wraps app discovery, key derivation, signing, version reads, and disconnect in the new retry helper

## How this was tested
- A example program made to test all p/x/c txs, signed with ledger, on local network + (someones on fuji) was made to check the stability of the ledger interaction and to check the availability of sign vs sign hash on new ledger app. Nano S+. Macos.
After the fix, the program executes many times with no issues.

### DIFF
--- a/utils/crypto/ledger/ledger.go
+++ b/utils/crypto/ledger/ledger.go
@@ -5,6 +5,8 @@ package ledger
 
 import (
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/crypto/keychain"
@@ -19,9 +21,38 @@ const (
 	rootPath          = "m/44'/9000'/0'" // BIP44: m / purpose' / coin_type' / account'
 	ledgerBufferLimit = 8192
 	ledgerPathSize    = 9
+	maxRetries        = 5
+	initialRetryDelay = 200 * time.Millisecond
 )
 
 var _ keychain.Ledger = (*Ledger)(nil)
+
+// retryOnHIDAPIError executes a function up to maxRetries times if it encounters
+// the specific "hidapi: unknown failure" error or APDU error 0x6987
+func retryOnHIDAPIError(fn func() error) error {
+	var err error
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		err = fn()
+		if err == nil {
+			return nil
+		}
+
+		// Check if the error contains the specific HIDAPI failure message or APDU 0x6987
+		if strings.Contains(err.Error(), "hidapi: unknown failure") ||
+			strings.Contains(err.Error(), "APDU Error Code from Ledger Device: 0x6987") {
+			if attempt < maxRetries {
+				// Calculate backoff delay: 200ms, 400ms, 600ms, 800ms
+				delay := time.Duration(attempt) * initialRetryDelay
+				time.Sleep(delay)
+				continue
+			}
+		}
+
+		// If it's not a retryable error or we've exhausted retries, return the error
+		return err
+	}
+	return err
+}
 
 // Ledger is a wrapper around the low-level Ledger Device interface that
 // provides Avalanche-specific access.
@@ -30,8 +61,13 @@ type Ledger struct {
 	epk    *bip32.Key
 }
 
-func New() (keychain.Ledger, error) {
-	device, err := ledger.FindLedgerAvalancheApp()
+func New() (*Ledger, error) {
+	var device *ledger.LedgerAvalanche
+	err := retryOnHIDAPIError(func() error {
+		var err error
+		device, err = ledger.FindLedgerAvalancheApp()
+		return err
+	})
 	return &Ledger{
 		device: device,
 	}, err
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -88,6 +88,15 @@ require (
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 )
 
+require (
```
maru-ava: Any idea why this gets added as a separate section?

utils/crypto/ledger/ledger.go
```diff
@@ -19,19 +21,53 @@ const (
...
+func retryOnHIDAPIError(fn func() error) error {
+	var err error
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		err = fn()
+		if err == nil {
+			return nil
+		}
+
+		// Check if the error contains the specific HIDAPI failure message or APDU 0x6987
+		if strings.Contains(err.Error(), "hidapi: unknown failure") ||
```
maru-ava: Does the use of `Contains` rather than checking for an exact match indicate that there is potentially more text involved? If so, might those details be useful enough to justify being logged for traceability when within maxRetries?

utils/crypto/ledger/ledger.go
```diff
@@ -19,19 +21,53 @@ const (
...
+			strings.Contains(err.Error(), "APDU Error Code from Ledger Device: 0x6987") {
+			if attempt < maxRetries {
+				// Calculate backoff delay: 200ms, 400ms, 600ms, 800ms
```
maru-ava, line 44: (No action required) Maybe indicate in the comment why this scheme is appropriate?

utils/crypto/ledger/ledger.go
```diff
@@ -19,19 +21,53 @@ const (
...
+				delay := time.Duration(attempt) * initialRetryDelay
+				time.Sleep(delay)
+				continue
+			}
+		}
+
+		// If it's not a retryable error or we've exhausted retries, return the error
```
maru-ava: Maybe remove this return? The final return would appear to suffice.

go.mod
felipemadero: I understand `go get` is doing that in this case.

utils/crypto/ledger/ledger.go
felipemadero: it is just that msg, changed to exact match

utils/crypto/ledger/ledger.go
felipemadero: added a brief comment

utils/crypto/ledger/ledger.go
felipemadero: It is needed because at this point, I don't want to retry. Changed for a break instead

utils/crypto/ledger/ledger.go
```diff
@@ -42,7 +79,12 @@ func addressPath(index uint32) string {
 }
 
 func (l *Ledger) Address(hrp string, addressIndex uint32) (ids.ShortID, error) {
-	resp, err := l.device.GetPubKey(addressPath(addressIndex), true, hrp, "")
+	var resp *ledger.ResponseAddr
+	err := retryOnHIDAPIError(func() error {
+		var err error
+		resp, err = l.device.GetPubKey(addressPath(addressIndex), false, hrp, "")
```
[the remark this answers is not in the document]
felipemadero: it is an intended change, to avoid the user to need to authorize this query operation

utils/crypto/ledger/ledger.go
```diff
@@ -19,19 +20,55 @@ const (
...
 // Ledger is a wrapper around the low-level Ledger Device interface that
 // provides Avalanche-specific access.
 type Ledger struct {
 	device *ledger.LedgerAvalanche
 	epk    *bip32.Key
 }
 
-func New() (keychain.Ledger, error) {
-	device, err := ledger.FindLedgerAvalancheApp()
+func New() (*Ledger, error) {
```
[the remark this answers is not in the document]
felipemadero: it is deliberate change. It is indeed breaking change but will not impact known users. Users can expect to receive an object that implements the interface.

utils/crypto/ledger/ledger.go
```diff
@@ -19,19 +20,55 @@ const (
...
+// the specific "hidapi: unknown failure" error or APDU error 0x6987
+func retryOnHIDAPIError(fn func() error) error {
+	var err error
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		err = fn()
+		if err == nil {
+			return nil
+		}
+
+		// Check if the error is HIDAPI failure message or APDU 0x6987
```
maru-ava: Maybe reword to explain why these errors are appropriate to retry? Otherwise the comment doesn't provide any more detail than already obvious from the code.

felipemadero: improved comment

yacovm:
> * A example program made to test all p/x/c txs, signed with ledger, on local network + (somes on fuji) was made to check the stability of the ledger interaction and to check the availability of sign vs sign hash on new ledger app. Nano S+. Macos.
>   After the fix, the program executes many times with no issues.

What example program?

utils/crypto/ledger/ledger.go
```diff
@@ -5,6 +5,7 @@ package ledger
 
 import (
 	"fmt"
+	"time"
```
yacovm, line 8: 1. Is the code in this file tested in CI? If so, then we should have a test that fails without the changes in this PR.
2. If the code is not tested in CI (likely because we don't have a ledger connected to our github runners) then I don't think we should have this code in avalanchego.

felipemadero:

PR #4350, opened 2025-09-25
uplift: add `vmerrors` package from evm repositories

JonathanOppenheimer:
## Why this should be merged
There are packages in subnet-evm and coreth that share a lot of functionality / are mirrored. As has been discussed as an evm team wide effort, we are working to uplift these packages to AvalancheGo to prevent the duplicative maintenance of two sets of the same code. This particular package, `vmerrors` is a relatively easy lift, and is the union of both EVM repositories `error.go` file (which were disjoint). 

I will note however, that the fact this package exists is kind of stupid, IMO. Merging this PR would also then enabled the following issue to be resolved more easily across both repositories:

https://github.com/ava-labs/subnet-evm/issues/1747

## How this was tested

No


RodrigoVillar (approved):

alarso16:
This does make a lot of sense to uplift - however, I worry that once we resolve ava-labs/coreth#144 (specifically, errcheck and forbidigo) we will need more global errors here. It might even be worth moving some block verification errors into this package prior (e.g. `see coreth/plugin/evm/wrapped_block.go`). Is it worth waiting to uplift these three errors until the errors are more globally managed in coreth? Feel free to disagree

JonathanOppenheimer:
> This does make a lot of sense to uplift - however, I worry that once we resolve [ava-labs/coreth#144](https://github.com/ava-labs/coreth/pull/144) (specifically, errcheck and forbidigo) we will need more global errors here. It might even be worth moving some block verification errors into this package prior (e.g. `see coreth/plugin/evm/wrapped_block.go`). Is it worth waiting to uplift these three errors until the errors are more globally managed in coreth? Feel free to disagree

Fully agree - would love to see vmerrors be the source of all vmerrors, and used across the board - I think our error declaration in the EVM repos could use some work. Will mark as draft for now.

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package vmerrors
```
joshua-kim, line 4: Why do we define a `vmerrors` package, instead of defining these in the components that produce the errors? `evm.ErrFoo` would imply that it is a error the vm knows about. I generally feel that common packages are undesirable because it creates an unclear abstraction boundary that makes it very easy to dump things into.

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "errors"
+
+var (
+	ErrGenerateBlockFailed         = errors.New("failed to generate block")
+	ErrBlockVerificationFailed     = errors.New("failed to verify block")
+	ErrWrapBlockFailed             = errors.New("failed to wrap block")
+	ErrInvalidCoinbase             = errors.New("invalid coinbase")
+	ErrSenderAddressNotAllowListed = errors.New("cannot issue transaction from non-allowlisted address")
```
joshua-kim, line 13: Why is this a vm error? It seems like a error that the txpool defines. It looks like we have this in the `vmerrors` package because we wanted to avoid duplicating the error type, but it's two usages are:
* in the txpool [here](https://github.com/ava-labs/subnet-evm/blob/b02c096945bbd5979b010fbaf0411a3e8a1949ba/core/txpool/validation.go#L283), where we could just unexport this error since we care about its value in tests
* in state transition (I'm not familiar with this code) [here](https://github.com/ava-labs/subnet-evm/blob/0eabe2f692460c368c269f411a1742236316145b/core/state_transition.go#L364), where we're not even testing for it anyways. We don't reason about this error or test it - so we could just define it in-line here instead of forcing them to share the same error for what appears to be no clear benefit, since our prod code doesn't even reason about this error at all anyways.

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
...
+
+package vmerrors
+
+import "errors"
+
+var (
+	ErrGenerateBlockFailed         = errors.New("failed to generate block")
+	ErrBlockVerificationFailed     = errors.New("failed to verify block")
+	ErrWrapBlockFailed             = errors.New("failed to wrap block")
+	ErrInvalidCoinbase             = errors.New("invalid coinbase")
```
joshua-kim, line 12: This is only used for a unit test and it's defined by `dummy`. It seems we have an integration test [here](//github.com/ava-labs/coreth) that does an assertion on it, but I would argue that:
1. The test looks significantly more complicated than it needs to be, I think the minimal case you need is to build a block and overwrite the coinbase and it gives us this error (i'm not sure though)
2. Since this is defined by `dummy` and you don't have an import issue, we could just define this in `dummy` (since it defines this error anyways) and do an assert against `dummy.ErrInvalidCoinbase`

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
...
+// See the file LICENSE for licensing terms.
+
+package vmerrors
+
+import "errors"
+
+var (
+	ErrGenerateBlockFailed         = errors.New("failed to generate block")
+	ErrBlockVerificationFailed     = errors.New("failed to verify block")
+	ErrWrapBlockFailed             = errors.New("failed to wrap block")
```
joshua-kim, line 11: This is defined by code in `evm`, so we could move this there

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package vmerrors
+
+import "errors"
+
+var (
+	ErrGenerateBlockFailed         = errors.New("failed to generate block")
+	ErrBlockVerificationFailed     = errors.New("failed to verify block")
```
joshua-kim, line 10: Same w/ `ErrWrapBlockFailed`

vms/evm/vmerrors/errors.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package vmerrors
+
+import "errors"
+
+var (
+	ErrGenerateBlockFailed         = errors.New("failed to generate block")
```
joshua-kim, line 9: Same w/ `ErrWrapBlockFailed`

JonathanOppenheimer:
cc: @alarso16 - let's make changes to coreth/subnet-evm to get rid of this package entirely.

PR #4351, opened 2025-09-25
Set `activationTime` equal to `upgrade.InitiallyActiveTime` for `propservm`

JonathanOppenheimer:
## Why this should be merged
This PR builds towards using the upgradetest package for forks, and that package uses the upgrade.InitiallyActiveTime values. Next we can replace all the values with the fork. 

## How this was tested
Existing CI

No

### DIFF
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -12,8 +12,10 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 )
 
+const InitiallyActiveTimeUnix = 1599696000
+
 var (
-	InitiallyActiveTime       = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
+	InitiallyActiveTime       = time.Date(2020, time.September, 10, 0, 0, 0, 0, time.UTC)
 	UnscheduledActivationTime = time.Date(9999, time.December, 1, 0, 0, 0, 0, time.UTC)
 
 	Mainnet = Config{
--- a/vms/proposervm/post_fork_block_test.go
+++ b/vms/proposervm/post_fork_block_test.go
@@ -45,8 +45,8 @@ func TestOracle_PostForkBlock_ImplementsInterface(t *testing.T) {
 
 	// setup
 	var (
-		activationTime = time.Unix(0, 0)
-		durangoTime    = activationTime
+		activationTime = upgrade.InitiallyActiveTime
+		durangoTime    = upgrade.InitiallyActiveTime
 	)
 	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
 	defer func() {
@@ -90,7 +90,7 @@ func TestBlockVerify_PostForkBlock_PreDurango_ParentChecks(t *testing.T) {
 	require := require.New(t)
 
 	var (
-		activationTime = time.Unix(0, 0)
+		activationTime = upgrade.InitiallyActiveTime
 		durangoTime    = upgrade.UnscheduledActivationTime
 	)
 	coreVM, valState, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
@@ -181,8 +181,8 @@ func TestBlockVerify_PostForkBlock_PostDurango_ParentChecks(t *testing.T) {
 	require := require.New(t)
 
 	var (
-		activationTime = time.Unix(0, 0)
-		durangoTime    = activationTime // post Durango
+		activationTime = upgrade.InitiallyActiveTime
+		durangoTime    = upgrade.InitiallyActiveTime // post Durango
 	)
 	coreVM, valState, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
 	defer func() {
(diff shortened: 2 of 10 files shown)


upgrade/upgrade.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 )
 
+const InitiallyActiveTimeUnix = 1599696000
+
 var (
-	InitiallyActiveTime       = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
+	InitiallyActiveTime       = time.Date(2020, time.September, 10, 0, 0, 0, 0, time.UTC)
```
StephenButtolph: Changing this should be done in a different PR. It isn't relevant to improving the proposervm tests.

vms/proposervm/post_fork_block_test.go
```diff
@@ -181,8 +181,8 @@ func TestBlockVerify_PostForkBlock_PostDurango_ParentChecks(t *testing.T) {
 	require := require.New(t)
 
 	var (
-		activationTime = time.Unix(0, 0)
-		durangoTime    = activationTime // post Durango
+		activationTime = upgrade.InitiallyActiveTime
+		durangoTime    = upgrade.InitiallyActiveTime // post Durango
```
StephenButtolph:
```suggestion
		durangoTime    = upgrade.InitiallyActiveTime
```

vms/proposervm/pre_fork_block_test.go
StephenButtolph, line 1: The changes in this file don't look accurate.

JonathanOppenheimer: These changes are now in line with the other suggested changes of not setting `durangoTime = ActivationTime`

upgrade/upgrade.go
JonathanOppenheimer: It has been reverted

upgrade/upgrade.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/ava-labs/avalanchego/utils/constants"
 )
 
+const InitiallyActiveTimeUnix = 1607144400
```
joshua-kim: Why do we need this const? Couldn't callers get get this information from `InitiallyActiveTime.Unix()`?

vms/proposervm/vm_test.go
```diff
@@ -2542,8 +2542,8 @@ func TestTimestampMetrics(t *testing.T) {
 
 	innerBlock := snowmantest.BuildChild(snowmantest.Genesis)
 
-	outerTime := time.Unix(314159, 0)
-	innerTime := time.Unix(142857, 0)
+	outerTime := time.Unix(upgrade.InitiallyActiveTimeUnix+1000, 0)
+	innerTime := time.Unix(49463, 0) // the actual numbers do not matter here, we just verify the metrics are populated correctly
```
joshua-kim: Could we pick some other number offset from `upgrade.InitiallyActiveTimeUnix` like we did with the outer time? We might accidentally break this test if we change the timestamp. It seems like previously this timestamp was inbetween the two, so maybe we could do `upgrade.InitiallyActiveTimeUnix+500`?

vms/proposervm/vm_test.go
```diff
@@ -2542,8 +2542,8 @@ func TestTimestampMetrics(t *testing.T) {
 
 	innerBlock := snowmantest.BuildChild(snowmantest.Genesis)
 
-	outerTime := time.Unix(314159, 0)
-	innerTime := time.Unix(142857, 0)
+	outerTime := time.Unix(upgrade.InitiallyActiveTimeUnix+1000, 0)
```
joshua-kim: It's hard to interpret what `+1000` means, so we could write this as `upgrade.InitiallyActiveTime.Add(1000 * time.Second).Unix()` or `upgrade.InitiallyActiveTime.Add(15 * time.Minute).Unix()` (since 1k seconds ~= 15 min)

upgrade/upgrade.go
StephenButtolph: It is so that it can be a const rather than a var

vms/proposervm/vm_test.go
JonathanOppenheimer: Yeah sure - I changed it to +500 - the number literally does not matter though.

upgrade/upgrade.go
joshua-kim: Chatted offline, seems like we don't actually need to have these consts since the source-of-truth is from the var anyways.

JonathanOppenheimer: the const has been removed.

joshua-kim (approved):

vms/proposervm/vm_test.go
```diff
@@ -2542,8 +2542,8 @@ func TestTimestampMetrics(t *testing.T) {
 
 	innerBlock := snowmantest.BuildChild(snowmantest.Genesis)
 
-	outerTime := time.Unix(314159, 0)
-	innerTime := time.Unix(142857, 0)
+	outerTime := time.Unix(upgrade.InitiallyActiveTime.Add(1000*time.Second).Unix(), 0)
+	innerTime := time.Unix(upgrade.InitiallyActiveTime.Add(500*time.Second).Unix(), 0) // the actual numbers do not matter here, we just verify the metrics are populated correctly
```
StephenButtolph:
```suggestion
	// The actual numbers do not matter here, we just verify the metrics are
	// populated correctly.
	outerTime := upgrade.InitiallyActiveTime.Add(1000 * time.Second)
	innerTime := upgrade.InitiallyActiveTime.Add(500 * time.Second)
```

StephenButtolph (approved):

PR #4352, opened 2025-09-26
feat: Use the new ETA tracker for load tests

rkuris:
Replaced the deprecated EstimateETA function with EtaTracker.

## Why this should be merged

Because the new EtaTracker is much better

## How this works

Removed the deprecated call and replaced it with the new one

## How this was tested

Reran the test and see that it now has better estimates

No

### DIFF
--- a/tests/reexecute/c/vm_reexecute_test.go
+++ b/tests/reexecute/c/vm_reexecute_test.go
@@ -335,9 +335,10 @@ type vmExecutorConfig struct {
 }
 
 type vmExecutor struct {
-	config  vmExecutorConfig
-	vm      block.ChainVM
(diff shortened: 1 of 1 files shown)


RodrigoVillar:
Could we rename the PR title to `feat: Use the new ETA tracker for re-execution tests` or something similar? Usually, load tests refer to the code in the `tests/load` package 😄

aaronbuchwald (approved):
Code change LGTM

Would you mind attaching logs from the two comparison runs before/after this PR to show that the ETA is more accurate with this change, so that there's at least one clear data point?

Approving regardless since either way this is replacing deprecated code.

PR #4361, opened 2025-09-29
Implement Firewood static state sync

alarso16:
## Why this should be merged

This is the first of many PRs to enable state-syncing Firewood in the EVM. Although not in the perfect state to be used, this marks being able to create the `xsync.Manager` object with range proofs, and this functionality is tested.

## How this works

Creates the interfaces needed to be a `xsync.DB`, with stubs for everything change-proof related. Additionally, the FFI's `FindNextKey` simply returns the last key in the range proof, so there's some hacky stuff to make that work

## How this was tested

New UT

No

### DIFF
--- a/firewood/syncer/syncer_test.go
+++ b/firewood/syncer/syncer_test.go
@@ -5,7 +5,6 @@ package firewood
 
 import (
 	"bytes"
-	"context"
 	"errors"
 	"math/rand"
 	"testing"
@@ -69,13 +68,16 @@ func Test_Firewood_Sync(t *testing.T) {
 }
 
 func testSync(t *testing.T, seed int64, clientKeys int, serverKeys int) {
-	ffiServer, close := generateDB(t, serverKeys, seed)
-	defer close(t.Context())
-	serverDB := &database{db: ffiServer}
-	clientDB, close := generateDB(t, clientKeys, seed+1) // guarantee different data
-	defer close(t.Context())
-
 	ctx := t.Context()
+
+	ffiServer := generateDB(t, serverKeys, seed)
+	serverDB := &database{db: ffiServer}
+	clientDB := generateDB(t, clientKeys, seed+1) // guarantee different data
+	defer func() {
+		require.NoError(t, serverDB.db.Close(ctx))
+		require.NoError(t, clientDB.Close(ctx))
+	}()
+
 	root, err := serverDB.GetMerkleRoot(ctx)
 	require.NoError(t, err)
 
@@ -101,20 +103,14 @@ func testSync(t *testing.T, seed int64, clientKeys int, serverKeys int) {
 // generateDB creates a new Firewood database with up to [numKeys] random key/value pairs.
 // The function returned closes the database, waiting on the provided context, if there were no test failures.
 // Note that each key/value pair may not be unique, so the resulting database may have fewer than [numKeys] entries.
-func generateDB(t *testing.T, numKeys int, seed int64) (*ffi.Database, func(context.Context)) {
+func generateDB(t *testing.T, numKeys int, seed int64) *ffi.Database {
 	t.Helper()
 	db, err := ffi.New(t.TempDir())
 	require.NoError(t, err)
 	require.NotNil(t, db)
 
-	close := func(ctx context.Context) {
-		if !t.Failed() {
-			require.NoError(t, db.Close(ctx))
-		}
-	}
-
 	if numKeys == 0 {
-		return db, close
+		return db
 	}
 
 	var (
@@ -144,7 +140,7 @@ func generateDB(t *testing.T, numKeys int, seed int64) (*ffi.Database, func(cont
 	_, err = db.Update(keys, vals)
 	require.NoError(t, err)
 
-	return db, close
+	return db
 }
 
 // logDiff logs the differences between two Firewood databases.


x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	_, err = fw.Update(keys, vals)
+	require.NoError(t, err)
+
+	return db
+}
+
+// logDiff logs the differences between two Firewood databases.
+// Useful when debugging state mistmatches after sync.
+func logDiff(t *testing.T, db1, db2 *syncDB) {
```
alarso16: I wrote this function for debugging, but that doesn't mean I have to merge it in. Any thoughts?

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+			EmptyRoot:             ids.ID(types.EmptyRootHash),
+		},
+		prometheus.NewRegistry(),
+	)
+	require.NoError(err)
+	require.NotNil(syncer)
+
+	require.NoError(syncer.Start(ctx))
+	err = syncer.Wait(ctx)
+	if errors.Is(err, xsync.ErrFinishedWithUnexpectedRoot) {
```
alarso16: The most common error I encountered while trying to make this work as a timeout, so I wonder if making a context with a timeout to log differences would be more beneficial than letting it stall for 2 minutes without helpful logs

x/firewood/sync_db.go
```diff
@@ -0,0 +1,121 @@
...
+	_ xsync.DB[*RangeProof, *ChangeProof] = (*syncDB)(nil)
+)
+
+// syncDB wraps a Firewood FFI database to implement the xsync.DB interface.
+type syncDB struct {
+	fw   *ffi.Database
+	lock sync.Mutex // TODO: remove this lock once FFI is thread-safe
+}
+
+func wrapSyncDB(db *ffi.Database) *syncDB {
```
alarso16: I'm thinking that it would be more logical to export a function like this:
```
func NewManager(*ffi.Database, *ManagerConfig) (*xsync.Manager[*RangeProof, *ChangeProof], error)
```
rather than exposing a database and requiring the user to understand the types. Should I save this for a different PR? Do you agree?

x/firewood/sync_db.go
```diff
@@ -0,0 +1,121 @@
...
+	// Extra data must be provided at commit time.
+	// TODO: remove this once the FFI no longer requires it.
+	proof.root = expectedEndRootID
+	proof.maxLength = maxLength
+
+	return proof.ffi.Verify(ffi.Hash(expectedEndRootID), start, end, uint32(maxLength))
+}
+
+// Commit the range proof to the database.
+func (db *syncDB) CommitRangeProof(_ context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
```
alarso16: I wouldn't look to closely at this function. it's all a little janky until Firewood has a proper FindNextKey implementation, in which case we don't have to copy the byte slice, and we don't have to analyze the returned values.

x/firewood/proof.go
```diff
@@ -0,0 +1,58 @@
...
+	_ xsync.Marshaler[*RangeProof]  = RangeProofMarshaler{}
+	_ xsync.Marshaler[*ChangeProof] = ChangeProofMarshaler{}
+)
+
+type RangeProofMarshaler struct{}
+
+func (RangeProofMarshaler) Marshal(r *RangeProof) ([]byte, error) {
+	if r.ffi == nil {
+		return nil, nil
+	}
```
joshua-kim: Do we need to do this `nil` check? It looks like we're guaranteed non-nil pointer here since `GetRangeProofAtRoot` never returns `nil`

x/sync/firewood/proof.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (ChangeProofMarshaler) Marshal(*ChangeProof) ([]byte, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (ChangeProofMarshaler) Unmarshal([]byte) (*ChangeProof, error) {
+	return nil, errors.New("not implemented")
+}
+
+type ChangeProof struct{}
```
joshua-kim: nit: We could remove this struct definition and satisfy the generic w/ `struct{}` in places where this is used to avoid prematurely introducing this type - but the intent of this type is pretty clear so it's also fine to leave this as a `TODO`.

x/firewood/proof.go
```diff
@@ -0,0 +1,58 @@
...
+	if err := proof.UnmarshalBinary(data); err != nil {
+		return nil, err
+	}
+	return &RangeProof{
+		ffi: proof,
+	}, nil
+}
+
+type RangeProof struct {
+	ffi       *ffi.RangeProof
```
joshua-kim: nit: I think we went with `ffi` as a name to make it clear that we're talking to the ffi - but although it makes clear the domain boundary we're communicating across it makes it more unclear what the type we're working with is (`r.ffi.MarshalBinary()` reads like we're serializing the ffi). 

Some recommendations for other names would be:
* `rangeProof` - verbose, defaulting to the type name is never bad
* `ffiRangeProof` - if you want to highlight that we're working with the ffi's version of this domain type
* `inner` - since this is a wrapper pattern
* `r` or `rp` - Can be used for shorter scoped code and since this a wrapper pattern we can argue that the context is given by the wrapping type so `r.r.MarshalBinary()` is clearly us marshaling the type we wrap.

If I was the author I would probably have used `r` - but this comment is entering opinionated territory so don't feel bound to my preferences.

x/firewood/proof.go
```diff
@@ -0,0 +1,58 @@
...
+
+type RangeProofMarshaler struct{}
+
+func (RangeProofMarshaler) Marshal(r *RangeProof) ([]byte, error) {
+	if r.ffi == nil {
+		return nil, nil
+	}
+
+	data, err := r.ffi.MarshalBinary()
+	return data, err
```
joshua-kim: This can be directly returned

x/firewood/proof.go
```diff
@@ -0,0 +1,58 @@
...
+package firewood
+
+import (
+	"errors"
+
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
+
+	"github.com/ava-labs/avalanchego/ids"
+
+	xsync "github.com/ava-labs/avalanchego/x/sync"
```
joshua-kim: nit: Unnecessary import alias? It's possible you left this in to avoid ambiguity with the stdlib `sync` package but I generally avoid these unless there is an actual conflict

firewood/syncer/proof.go
joshua-kim, line 1: Did we intend to put this in `x/sync/firewood`?

x/firewood/sync_db.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/ava-labs/avalanchego/utils/maybe"
+
+	xsync "github.com/ava-labs/avalanchego/x/sync"
+)
+
+var _ xsync.DB[*RangeProof, *ChangeProof] = (*syncDB)(nil)
+
+// syncDB wraps a Firewood FFI database to implement the xsync.DB interface.
+type syncDB struct {
+	fw   *ffi.Database
```
joshua-kim: nit: similar naming comment as in the other file, but naming this type something like `DB` with `db` as a field might make sense

x/firewood/sync_db.go
joshua-kim: I agree + I would do it as part of this PR - currently it's not clear how this package comes together without this.

I think we can define our own config type here though instead of accepting `sync.ManagerConifg` - we already define marshaling so we can accept a subset of the `ManagerConfig` instead. Currently that type is being abused and is used as a DI container (it accepts fields that are required that don't have sensible defaults), so we can explicitly accept things like the p2p clients as fields to make the API a bit cleaner. If we do this - we should be able to unexport the marshaler types as well.

x/firewood/sync_db.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (db *syncDB) GetMerkleRoot(context.Context) (ids.ID, error) {
+	root, err := db.fw.Root()
+	if err != nil {
+		return ids.ID{}, err
+	}
+	return ids.ID(root), nil
+}
+
+// GetRangeProofAtRoot returns a range proof for x/sync between [start, end].
```
joshua-kim: nit: w.r.t the functions on this type - these are documented on the interface so duplicative to document here on the implementation

x/firewood/sync_db.go
```diff
@@ -0,0 +1,119 @@
...
+	// TODO: remove this once the FFI no longer requires it.
+	proof.root = expectedEndRootID
+	proof.maxLength = maxLength
+
+	return proof.ffi.Verify(ffi.Hash(expectedEndRootID), start, end, uint32(maxLength))
+}
+
+// Commit the range proof to the database.
+func (db *syncDB) CommitRangeProof(_ context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
+	// Prevent concurrent commits to the database.
```
joshua-kim: These comments are overly verbose - generally try to document "why" comments instead of "how" or "what". "How" and "what" can be derived from the reader - and if it's not possible to or difficult to do so it's a sign that the code is not readable enough. In this specific example locks are always used for concurrency control so it just adds noise to the code which is already readable

x/firewood/sync_db.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// We can now get the FindNextKey iterator.
+	nextKeyRange, err := proof.ffi.FindNextKey()
+	if err != nil || nextKeyRange == nil {
+		// No error indicates the range is complete.
+		return maybe.Nothing[[]byte](), err
+	}
+
+	// It is borrowed from the ffi, so we need to make our own copy.
```
joshua-kim: This is an example of a good comment - since we're making a defensive copy. Does the ffi actually modify this value though? I'm wondering if this defensive copy is something that we actually need to do.

firewood/syncer/syncer_test.go
joshua-kim, line 1: We define some sync tests in `merkledb` - it might make sense to make test suite that accepts an implementation of `sync.DB` to verify that you satisfy the interface (similar to what we do for our database tests) so we can reuse our test coverage. This might not actually be possible though since merkledb doesn't have the same restrictions w.r.t key length that firewood does w/ eth hash enabled.

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+)
+
+var (
+	rangeProofMarshaler  = RangeProofMarshaler{}
+	changeProofMarshaler = ChangeProofMarshaler{}
+)
+
+// Test_Firewood_Sync tests syncing a Firewood database from empty to a full database.
+func Test_Firewood_Sync(t *testing.T) {
+	t.Parallel()
```
joshua-kim: Do these calls meaningfully improve test execution time, or are these speculative improvements? I generally avoid the usage of `Parallel` unless there is a tangible benefit because it otherwise adds noise + might even add overhead to the test.

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	dbSizes := []int{0, 1, 1_000, 10_000, 100_000}
+	for _, serverSize := range dbSizes {
```
joshua-kim: Can we write this as a testing table?

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+		})
+	}
+}
+
+// Test_Firewood_Sync_Overwrite tests syncing when the client already has some data.
+// This data is expected to be overwritten by the server's data.
+func Test_Firewood_Sync_Overwrite(t *testing.T) {
+	t.Parallel()
+	dbSizes := []int{0, 1, 1_000, 10_000, 100_000}
+	initialSize := 500
```
joshua-kim: nit: I would in-line this to avoid creating extra named variables on the stack which clutters the scope since it's only used in one place anyways

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	for _, serverSize := range dbSizes {
+		t.Run(fmt.Sprintf("numKeys=%d", serverSize), func(t *testing.T) {
+			t.Parallel()
+			testSync(t, initialSize, serverSize)
+		})
+	}
+}
+
+func testSync(t *testing.T, clientKeys int, serverKeys int) {
+	require := require.New(t)
```
joshua-kim: I think we've been trying to avoid usages of the require instance in favor of the functions `require.Equal(t, ...)` since there's a potential footgun w/ subtesting (although it's used correctly here).

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	seed := time.Now().UnixNano()
```
joshua-kim: Can we avoid using time as the seed? This is non-deterministic and will result on tests having different outputs which can lead to confusing unit test debugging experiences - maybe we can accept a hard-coded seed from the call-site, or always use zero as the starting seed?

If we intentionally want this to be randomized making this a fuzz test might be reasonable here.

x/firewood/sync_test.go
joshua-kim: I'm not aware of a great way of doing this - if we use our own timeout logic we don't respect the runtime's testing timeout and although `t.Deadline` exists trying to fire something right before the timeout is race-y.

x/sync/firewood/syncer_test.go
```diff
@@ -0,0 +1,274 @@
...
+	t.Helper()
+	path := filepath.Join(t.TempDir(), "firewood.db")
+
+	fw, err := ffi.New(path, ffi.DefaultConfig())
+	require.NoError(t, err)
+	require.NotNil(t, fw)
+	t.Cleanup(func() {
+		ctx := context.WithoutCancel(t.Context())
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second) // allow some time for garbage collection
+		defer cancel()
```
joshua-kim: Why do we do this?

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/ava-labs/avalanchego/network/p2p/p2ptest"
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	xsync "github.com/ava-labs/avalanchego/x/sync"
+)
+
+var (
+	rangeProofMarshaler  = RangeProofMarshaler{}
+	changeProofMarshaler = ChangeProofMarshaler{}
+)
```
joshua-kim: nit: What is this for? I generally try to avoid globals where possible since it clutters the package scope - this also doesn't create more readable code since `*proofMarshaler` is nearly equivalent to `*ProofMarshaler{}`.

firewood/syncer/syncer_test.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		// Random length between minLength and maxLength inclusive
+		valLen := r.Intn(maxLength-minLength+1) + minLength
+
+		key := make([]byte, keyLen)
+		val := make([]byte, valLen)
+
+		// Fill with random bytes
+		_, err := r.Read(key)
+		require.NoError(t, err, "read never errors")
```
joshua-kim, line 132: nit: self documenting code (although Arran has been using these logs - so maybe this is preference)

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	var (
+		r         = rand.New(rand.NewSource(seed)) // #nosec G404
+		keys      = make([][]byte, numKeys)
+		vals      = make([][]byte, numKeys)
+		minLength = 1
+		maxLength = 64
+	)
+	for i := 0; i < numKeys; i++ {
```
joshua-kim: nit: this can be written as `for range numKeys`

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+		val := make([]byte, valLen)
+
+		// Fill with random bytes
+		_, err := r.Read(key)
+		require.NoError(t, err, "read never errors")
+		_, err = r.Read(val)
+		require.NoError(t, err, "read never errors")
+
+		keys[i] = key
+		vals[i] = val
```
joshua-kim: nit: We can use `append` instead of using indices to be more readable

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+		vals      = make([][]byte, numKeys)
+		minLength = 1
+		maxLength = 64
+	)
+	for i := 0; i < numKeys; i++ {
+		// Key length must match geth structure
+		keyLen := 32
+		if r.Intn(2) == 0 {
+			keyLen = 64
+		}
```
joshua-kim:
```suggestion
		// Key length must match geth structure
		keyLen := 32 * (r.Intn(2) + 1)
```

x/firewood/sync_test.go
joshua-kim: If it was helpful to you it'll likely be helpful to others when debugging

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+	require.NoError(t, err)
+
+	return db
+}
+
+// logDiff logs the differences between two Firewood databases.
+// Useful when debugging state mismatches after sync.
+func logDiff(t *testing.T, db1, db2 *syncDB) {
+	t.Helper()
+	require := require.New(t)
```
joshua-kim: Similar comment w.r.t `require` instances

x/sync/syncer.go
```diff
@@ -516,8 +517,8 @@ func (s *Syncer[R, _]) handleRangeProofResponse(
 	if err := s.db.VerifyRangeProof(
 		ctx,
 		rangeProof,
-		protoutils.ProtoToMaybe(request.StartKey),
-		protoutils.ProtoToMaybe(request.EndKey),
+		work.start,
+		work.end,
```
joshua-kim: is this a separate diff/bugfix?

x/sync/syncer.go
```diff
@@ -150,6 +150,7 @@ type Config[R any, C any] struct {
 	SimultaneousWorkLimit int
 	Log                   logging.Logger
 	TargetRoot            ids.ID
+	EmptyRoot             ids.ID
```
joshua-kim, line 153: What is firewood's value for the empty root? Is it not just 32 zeroes?

x/sync/firewood/sync_db.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return maybe.Some(newSlice), nil
+}
+
+//nolint:revive // TODO: implement this method.
+func (db *syncDB) GetChangeProof(_ context.Context, startRootID ids.ID, endRootID ids.ID, start maybe.Maybe[[]byte], end maybe.Maybe[[]byte], maxLength int) (*ChangeProof, error) {
+	return nil, errors.New("change proofs are not implemented")
+}
+
+//nolint:revive // TODO: implement this method.
```
joshua-kim: What linter are we breaking here?

x/firewood/sync_test.go
```diff
@@ -0,0 +1,274 @@
...
+		t.Run(fmt.Sprintf("numKeys=%d", serverSize), func(t *testing.T) {
+			t.Parallel()
+			testSync(t, 0, serverSize)
+		})
+	}
+}
+
+// Test_Firewood_Sync_Overwrite tests syncing when the client already has some data.
+// This data is expected to be overwritten by the server's data.
+func Test_Firewood_Sync_Overwrite(t *testing.T) {
```
joshua-kim: Seems like this can be refactored with the above test if we write a table

x/firewood/proof.go
alarso16: Probably not

x/sync/firewood/proof.go
alarso16: I did consider that, but it felt more intuitive to provide the infrastructure here, since we need some dummy marshaller anyway

x/firewood/proof.go
alarso16: Ah that is unclear. I'll choose `rp` out of preference, but I don't mind any of those names

x/firewood/proof.go
alarso16: That is the reason, but makes sense to me

firewood/syncer/proof.go
alarso16: That's a great question! And highly related to our conversation in #4661. I think we really just need to decide whether we cluster all our Firewood code in one place, or if we group it close to its functionality. I'm sure you have an opinion about this, but I don't really

x/firewood/sync_db.go
alarso16: I named it `syncDB` because it doesn't need exported after we wrap, and I was curious if you would drop in your firewood implementation for the X-Chain alongside all this (or even the EVM database), in which case this would certainly conflict

x/firewood/sync_db.go
alarso16: I removed the comments that I thought may be unhelpful. Let me know what you think

x/firewood/sync_db.go
alarso16: I can actually remove this! It was originally borrowed from Rust, but the FFI was updated to copy the value before sending it to us. As of v0.0.16, this is no longer required.

x/sync/firewood/syncer_test.go
alarso16: If there's a memory management error, closing the database can hang to prevent undefined behavior. I would want to provide the test context, but it is already canceled. Additionally, I wouldn't want to timeout while closing the database. You should see the documentation of the `Close()` function, since it was hotly debated on the Firewood team

firewood/syncer/syncer_test.go
alarso16: Yeah the comment is useless, but I wasn't sure what to do about the always nil error. nolint it and ignore? I figured still checking the error would be better

x/firewood/sync_test.go
alarso16: I no longer need this, but that is much cleaner

x/sync/syncer.go
alarso16: It's just unnecessary to convert from proto when we have it right there. So yes? I forgot this was in the diff, I can remove it

x/sync/syncer.go
alarso16: Ethereum's empty root is not 32 zeros. An empty trie has the following hash:
https://github.com/ava-labs/libevm/blob/749b6cefda2894bab9bbdbe2d27086d57ca0d393/core/types/hashes.go#L27

x/sync/firewood/sync_db.go
alarso16: There's unused variables. I left them for clarity, so you don't have to go look at the interface definition at change-proof time to know what `ids.ID` is

x/firewood/sync_test.go
alarso16: Interesting. I'll leave it for now, and see if we run into more issues like this in the future.

firewood/syncer/syncer_test.go
alarso16: It turns out that the key length restriction put in place here was only necessary due to a bug in Firewood, which is now fixed. I think we can reduce the duplicate code, but it would touch quite a bit of merkledb code too. What if after we merge this PR, I refactor all the testing to work for both databases? I think most of the benefit will come once we want to start testing change proofs as well. I worry the scope of refactoring all these tests will creep into implementation

x/firewood/sync_test.go
alarso16: Locally the difference is about 1.6 vs. 2.5 seconds. So I would say it does meaningfully improve execution time percentage wise, but it's not super tremendous

x/firewood/sync_db.go
alarso16: Renamed to `db`, and we can adjust later if it conflicts with something

x/firewood/sync_test.go
alarso16: I set the seed to 0, but it does seem like we would want to know if it ever fails. I think that fuzzing makes sense to add separately?

x/firewood/sync_test.go
alarso16: Ah right, we talked about this once. I'll delete it

x/sync/firewood/sync_db.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+type Config struct {
+	RangeProofClient      *p2p.Client
+	ChangeProofClient     *p2p.Client
+	SimultaneousWorkLimit int
+	Log                   logging.Logger
+	TargetRoot            ids.ID
+	StateSyncNodes        []ids.NodeID
+}
```
joshua-kim, line 39: `x/sync`'s Config type is abused as a DI container + a go config pattern  - can we abstract over this to avoid leaking the bad implementation into this package? The proof clients are really dependencies of the syncer and the `TargetRoot` is really a parameter for which to start the syncer - I feel like we could probably do something like have `Config` know about stuff with safe zero values that have reasonable defaults to make tests easier to write + each VM integrating with this won't have to re-define their own defaults (which will all probably be the same). I think `SimultaneousWorkLimit` and `StateSyncNodes` have reasonable default values, and the logger can default to not logging. We could also pass the metrics instance through the config and the default can be a new registry - only prod cares to configure to use its own metrics instance.

`TargetRoot` i actually think is not a true config value because as part of state sync we need to define the point to sync to - so I think the caller should be explicitly providing this through a parameter. Similarly I think there aren't reasonable default values for the network clients so those should be defined explicitly as well.

x/sync/firewood/sync_db.go
```diff
@@ -0,0 +1,135 @@
...
+
+func NewSyncer(fw *ffi.Database, config Config, register prometheus.Registerer) (*xsync.Syncer[*RangeProof, *ChangeProof], error) {
```
joshua-kim: 1. nit: `fw` is deviating from the abstractions of `ffi` and `Database` - would calling this something like `db` make more sense? Similarly `register` might want to be `registerer` or `r`.
2. Locality for `NewSyncer` and `Config`'s definition might want to be after all of the function definitions on `db` or before `db`'s definition - currently this locality reads like a constructor for `db`.

x/sync/firewood/sync_db.go
joshua-kim, line 1: Should we update the file name here? Maybe `syncer`?

x/sync/firewood/sync_db.go
```diff
@@ -0,0 +1,135 @@
...
+	defer db.lock.Unlock()
+	_, err := db.db.VerifyAndCommitRangeProof(proof.rp, start, end, ffi.Hash(proof.root), uint32(proof.maxLength))
+	if err != nil {
+		return maybe.Nothing[[]byte](), err
+	}
+
+	nextKeyRange, err := proof.rp.FindNextKey()
+	if err != nil || nextKeyRange == nil {
+		// No error indicates the range is complete.
+		return maybe.Nothing[[]byte](), err
```
joshua-kim: Although semantically correct - this is not clear/deviates from Go's typical pattern of error handling where `err` typically implies a non-nil error. This can also cause unexpected behavior if the ffi returns a non-nil next key range and a non-nil error (we shouldn't because it breaks the aforementioned idiom of a meaningful error and a meaningful return value, but our repository breaks this idiom without good reason at times). At any rate - I would break this into two separate `if`'s for clarity.

x/sync/firewood/sync_db.go
joshua-kim: We should listen to the linter here - it doesn't matter what the parameters are/mean because they're not used and this code is clear/self-documenting enough to convey this to the reader

x/sync/firewood/sync_db.go
```diff
@@ -0,0 +1,135 @@
...
+	// TODO: remove this once the FFI no longer requires it.
+	proof.root = expectedEndRootID
+	proof.maxLength = maxLength
+
+	return proof.rp.Verify(ffi.Hash(expectedEndRootID), start, end, uint32(maxLength))
+}
+
+func (db *db) CommitRangeProof(_ context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
+	db.lock.Lock()
+	defer db.lock.Unlock()
```
joshua-kim: It's not clear to me what the concurrency guarantees we want to provide are - why do we lock here but not in the other methods? Are only updates to the database required to be locked?

x/firewood/sync_test.go
joshua-kim: 1 second is almost a rounding error - it's not worth adding overhead/complexity/non-determinism that might mess up with globals and debugging for something that won't be noticed when running tests

x/sync/firewood/sync_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p/p2ptest"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/x/sync"
+)
+
+func Test_Firewood_Sync(t *testing.T) {
+	for _, serverSize := range []int{0, 1, 1_000, 10_000, 100_000} {
+		for _, clientSize := range []int{0, 1_000} {
+			t.Run(fmt.Sprintf("numKeys=%d_clientKeys=%d", serverSize, clientSize), func(t *testing.T) {
```
joshua-kim: This works but it's deviating from the idiomatic table pattern - could we use a testing table instead? They tend to be easier to maintain over time and are more easily readable and don't require the string formatting that we're doing.

x/sync/firewood/sync_test.go
```diff
@@ -0,0 +1,239 @@
...
+	root, err := serverDB.GetMerkleRoot(ctx)
+	require.NoError(t, err)
+
+	syncer, err := NewSyncer(
+		clientDB,
+		Config{
+			RangeProofClient:      p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, sync.NewGetRangeProofHandler(serverDB, rangeProofMarshaler{})),
+			ChangeProofClient:     p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, sync.NewGetChangeProofHandler(serverDB, rangeProofMarshaler{}, changeProofMarshaler{})),
+			SimultaneousWorkLimit: 5,
+			Log:                   logging.NoLog{},
```
joshua-kim: Related to my earlier comment about defining some defaults - but it would simplify this test setup

x/sync/firewood/syncer_test.go
joshua-kim: Isn't this racey? If `Close` takes more than 5 seconds we can get a false positive by canceling early. Isn't it safe for us to timeout while closing because these are temporary db instances for testing purposes? If we mess up closure subsequent runs of the tests should not be impacted.

If there is a memory management bug in firewood - it's also not something that the syncer tests should care about anyways - it's an implementation detail of firewood. If we need better ways of debugging it I think we would want to add tests on `Close` closer to the ffi/firewood implementation than adding complexity to the syncer tests.

x/sync/firewood/sync_test.go
```diff
@@ -0,0 +1,239 @@
...
+		next2 = iter2.Next()
+		count2++
+		missingCount++
+		require.NoError(t, iter2.Err(), "iter2 error")
+	}
+	if missingCount > 0 {
+		t.Logf("%d keys missing from DB1 starting with %x", missingCount, prevKey)
+	}
+
+	t.Logf("DB1 had %d keys, DB2 had %d keys", count1, count2)
```
joshua-kim: Should we update some logs/naming of db1/db2 to be something like `got` and `want` to make it clear what the expected value was?

x/sync/syncer.go
joshua-kim: We should avoid unrelated changes where possible - I'm open to merging this in separately but we want to keep PRs scoped to a logical change. As an example reverting a feature by reverting the commit inadvertently also reverting unrelated cleanups means that we now need to manually revert a diff.

x/sync/firewood/proof.go
joshua-kim: We do need the dummy marshaler but this type isn't needed - we don't implement change proofs so `struct{}{}` works as a sentinel value (and actually this type being exposed out of `firewood/syncer` suggests that they are implemented).

joshua-kim (dismissed):

x/sync/firewood/syncer_test.go
alarso16: The proofs have finalizers attached. The fact that these finalizers work is tested in Firewood, but of course, if this code mistakenly holds a references to this afterwards, this could cause a memory leak. For it to timeout, the finalizers (which aren't technically guaranteed to run) would have to now have run for 5 seconds, even though closing the database directly calls `runtime.GC()`. It is technically racy, but I would be really, really surprised if it ever flaked.

What would your alternative be? There's no other decent context to pass to this function, and any subsequent test runs are unaffected

x/sync/firewood/sync_db.go
alarso16: I created an issue for fixing the x/sync API (deemed not very urgent) and I'll fix this config

x/sync/firewood/sync_db.go
alarso16: I can actually delete this now, but I'll explain why. The case here is trying to commit multiple proofs simultaneously. We need to prevent this, since committing a proof requires the most recent state of the database, which would be altered simultaneously. There's now locking on the Firewood-side (as of v0.0.17) to address this more directly

x/sync/firewood/sync_test.go
alarso16: I changed to be table-driven. Not sure it's a strict improvement, but maybe there's some naming issues that could be improved

x/sync/firewood/sync_test.go
alarso16: There's lots of items (an iterator, revision, etc.), so I'm not sure it's cleaner, but I'll mark this as resolved

x/sync/firewood/syncer.go
```diff
@@ -0,0 +1,146 @@
...
+// database wraps a Firewood FFI database to implement the xsync.DB interface.
+type database struct {
+	db *ffi.Database
+}
+
+type Config struct {
+	SimultaneousWorkLimit int                   // optional, defaults to 8
+	Log                   logging.Logger        // optional
+	StateSyncNodes        []ids.NodeID          // optional
+	Registerer            prometheus.Registerer // optional
```
joshua-kim: This is overdocumenting - it being a `Config` means that all the values are optional. The caller also doesn't care about the defaults because they are owned by this package

x/sync/firewood/syncer.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func NewSyncer(config Config, db *ffi.Database, targetRoot ids.ID, rangeProofClient *p2p.Client, changeProofClient *p2p.Client) (*xsync.Syncer[*RangeProof, struct{}], error) {
+	r := config.Registerer
+	if r == nil {
+		r = prometheus.NewRegistry()
+	}
```
joshua-kim: The creation of these intermediate variables makes it less clear that we're updating config defaults. Since the config is a value and not a pointer - it's fine to mutate it directly since we don't mess with the caller's state
```suggestion
    if config.Registerer == nil {
        config.Registerer = prometheus.NewRegistry()
    }
```

x/sync/firewood/syncer.go
```diff
@@ -0,0 +1,146 @@
...
+	logger := config.Log
+	if logger == nil {
+		logger = logging.NoLog{}
+	}
+	workLimit := config.SimultaneousWorkLimit
+	if workLimit <= 0 {
+		workLimit = defaultSimultaneousWorkLimit
+	}
```
joshua-kim: This is deviating a bit from the zero values -> override pattern configs use, which creates a misleading api since if the caller intentionally specifies a negative value we actually override it (we're mixing responsibilities of default setting and sanitization). I recommend throwing an error if this is negative (I think x/sync needs a regression test on this).

```suggestion
	if config.SimultaneousWorkLimit == 0 {
		config.SimultaneousWorkLimit = defaultSimultaneousWorkLimit
	}
```

x/sync/firewood/syncer_test.go
```diff
@@ -0,0 +1,237 @@
...
+		clientSize int
+		serverSize int
+	}{
+		{name: "both empty", clientSize: 0, serverSize: 0},
+		{name: "one request from empty", clientSize: 0, serverSize: 1000},
+		{name: "server empty", clientSize: 1000, serverSize: 0},
+		{name: "one request replace all", clientSize: 1000, serverSize: 1000},
+		{name: "10,000 keys from empty", clientSize: 0, serverSize: 10_000},
+		{name: "100,000 keys from empty", clientSize: 0, serverSize: 100_000},
+		{name: "10,000 keys replace all", clientSize: 10_000, serverSize: 10_000},
```
joshua-kim: nit: indenting here? Typically these test cases take the form of:
```
{
   name: <name>,
   clientSize: <clientSize>,
 ...
}
```

Single-line struct initialization is fine but in the specific example of test tables it's not idiomatic

x/sync/firewood/syncer_test.go
joshua-kim: > It is technically racy, but I would be really, really surprised if it ever flaked.

It being racey is a reason to not include it - if a unit test is flaky is is not a useful signal and especially on stressed CI workers sleeps being used to patch racy logic is unreliable. 

> What would your alternative be? There's no other decent context to pass to this function, and any subsequent test runs are unaffected

`t.Context` is only called if the test calls `FailNow` - maybe we could check `t.Failed` during `Cleanup` and only close the database if we haven't failed yet? If we do that - we can use `t.Context` without having to detach from it, and we don't care about closure if the test failed earlier anyways.

w.r.t timeouts - we could not define a timeout to avoid racey behavior with `Close` and rely on the `testing.T` timeout - it trades off a flakey test for extra time, but it means that the signal we get will be meaningful. Additionally developers after seeing one flake will likely not block until the test execution completes to iterate on the test - they can see the test is visibly hanging to infer that the test is still broken.

firewood/syncer/proof.go
joshua-kim: I think for now we could put this under `firewood/syncer` for better discoverability - `x/sync/firewood` makes it seem like firewood is an implementation of a syncing strategy when this is reflecting that firewood is a syncable database.

I also think with the revised code ownership on the repo getting reviews will not be an issue so we don't have to rely on the usage of `x` to unblock development, and I also don't want us to fall into the mindset that `x` should be held to a different standard than the rest of the codebase (or that it should exist at all).

x/sync/firewood/proof.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type RangeProof struct {
+	rp        *ffi.RangeProof
+	root      ids.ID
+	maxLength int
+}
+
+// TODO: implement an actual ChangeProof marshaler.
+type dummyMarshaler struct{}
```
joshua-kim: nit: changeProofMarshaler? `dummy*` gives the impression that this is a test double

x/sync/firewood/syncer.go
```diff
@@ -0,0 +1,146 @@
...
+	Registerer            prometheus.Registerer // optional
+}
+
+func NewSyncer(config Config, db *ffi.Database, targetRoot ids.ID, rangeProofClient *p2p.Client, changeProofClient *p2p.Client) (*xsync.Syncer[*RangeProof, struct{}], error) {
+	r := config.Registerer
+	if r == nil {
+		r = prometheus.NewRegistry()
+	}
+	logger := config.Log
+	if logger == nil {
```
joshua-kim: Same comment here

x/sync/firewood/syncer.go
alarso16: It does! I'l create an issue and write the test this afternoon in a separate PR

firewood/syncer/syncer_test.go
```diff
@@ -0,0 +1,267 @@
...
+		})
+	}
+}
+
+func testSync(t *testing.T, seed int64, clientKeys int, serverKeys int) {
+	ffiServer, close := generateDB(t, serverKeys, seed)
+	defer close(t.Context())
+	serverDB := &database{db: ffiServer}
+	clientDB, close := generateDB(t, clientKeys, seed+1) // guarantee different data
+	defer close(t.Context())
```
joshua-kim:
```suggestion
	ffiServer := generateDB(t, serverKeys, seed)
	serverDB := &database{db: ffiServer}
	clientDB := generateDB(t, clientKeys, seed+1) // guarantee different data
	
	defer func() {
		require.NoError(t, ffiServer.Close(t.Context()))
		require.NoError(t, clientDB.Close(t.Context()))
	}()
```

joshua-kim (approved):

firewood/syncer/syncer_test.go
```diff
@@ -0,0 +1,263 @@
...
+	require.NotNil(t, db)
+
+	if numKeys == 0 {
+		return db
+	}
+
+	var (
+		r         = rand.New(rand.NewSource(seed)) // #nosec G404
+		keys      = make([][]byte, numKeys)
+		vals      = make([][]byte, numKeys)
```
AminR443, line 119: Doesn't this create a slice of length numKeys, all nil? should be `make([][]byte, 0, numKeys)`.

PR #4363, opened 2025-09-29
Move canonical warp ordering to validators package

StephenButtolph:
## Why this should be merged

We are working to add another method to the `validators.State` interface. The method will return all warp-canonical validator sets that exist on the network.

Therefore, we need to move the warp-canonical validator set definitions from the `warp` package into the `validators` package.

## How this works

Renames + moves:
- `warp.Validator` -> `validators.Warp`
- `warp.CanonicalValidatorSet` -> `validators.WarpSet`
- `warp.FlattenValidatorSet` -> `validators.FlattenValidatorSet`

This PR does not remove the previous types/function, but it does deprecate them and alias them to the new versions.

## How this was tested

CI

No.

### DIFF
--- a/snow/validators/warp_test.go
+++ b/snow/validators/warp_test.go
@@ -1,0 +1,168 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package validators
+
+import (
+	"math"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls/signer/localsigner"
+
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+)
+
+var (
+	_ utils.Sortable[*testWarp] = (*testWarp)(nil)
+
+	testVdrs []*testWarp
+)
+
+type testWarp struct {
+	nodeID ids.NodeID
+	sk     bls.Signer
+	vdr    *Warp
+}
+
+func (v *testWarp) Compare(o *testWarp) int {
+	return v.vdr.Compare(o.vdr)
+}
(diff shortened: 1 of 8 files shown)


snow/validators/warp_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+var (
+	_ utils.Sortable[*testWarp] = (*testWarp)(nil)
+
+	testVdrs []*testWarp
+)
+
+type testWarp struct {
+	nodeID ids.NodeID
+	sk     bls.Signer
```
joshua-kim: If `sk` isn't ever used, it seems like we don't need this type and we could use `Warp` instead?

vms/platformvm/warp/validator.go
```diff
@@ -32,79 +26,33 @@ type ValidatorState interface {
...
 }
 
-type CanonicalValidatorSet struct {
-	// Validators slice in canonical ordering of the validators that has public key
-	Validators []*Validator
-	// The total weight of all the validators, including the ones that doesn't have a public key
-	TotalWeight uint64
-}
+type (
+	// Deprecated: use [validators.WarpSet] instead.
```
joshua-kim, line 30: Is there a reason we're doing this deprecation notice + aliasing instead of deleting these types and updating callsides to use the moved code?

StephenButtolph: I didn't want to update coreth atomically

snow/validators/warp_test.go
```diff
@@ -0,0 +1,152 @@
...
+	t.Helper()
+
+	vdrs := make([]*Warp, n)
+	for i := range vdrs {
+		vdrs[i] = newWarp(t)
+	}
+	utils.Sort(vdrs)
+	return WarpSet{
+		Validators:  vdrs,
+		TotalWeight: 3 * n,
```
joshua-kim: Why do we use a weight of 3 instead of 1?

StephenButtolph: changed to 1 and added a comment

michaelkaplan13 (approved):
This honestly would probably be a good time to start making the `Warp` -> `ICM` terminology update in the code, but I don't want to block on it.

Issue #4282, opened 2025-09-16
Expand Top Level Metrics from Re-Execution Benchmark


aaronbuchwald:
The re-execution benchmarks measure the total amount of time to execute a range of blocks [N, N+K], measures the total gas executed within that range and provides the mgas/s metric based on the total amount of elpased time during the benchmark.

Metrics are also exported to Grafana for ease of use, but they are not integrated into the benchmark results that are archived via GitHub Action Benchmark.

You can compare two runs based solely on the mgas/s, but it wouldn't immediately tell you where the change came from. This makes it desirable to account for every ms of time spent in the benchmark and provide the most accurate possible assessment, report them to the benchmark for easy archiving and comparison (report via [ReportMetric](https://github.com/ava-labs/avalanchego/blob/ed0f66caee4dae03279ab832182b1ce9426dc2a8/tests/reexecute/c/metrics.go#L130), and so that we can look at a run that impacts performance and point to the exact metrics that were changed and resulted in a meaningful diff in mgas/s.

To do this, we can utilize [MeterVM](https://github.com/ava-labs/avalanchego/tree/master/vms/metervm/) to account for the time spent in each of parse/verify/accept and break each of these down further as needed.

The sum of parse/verify/accept should be nearly equal to the elapsed time of the benchmark or else there may be a greater share of time spent reading blocks from disk to keep the execute sequence channel full (this would be unexpected imo and should be possible to avoid).

The next place we have existing metrics to break down further is within `InsertBlock`, which should take up most of the time during block verification. Here, we have detailed metrics that are broken down [here](https://github.com/ava-labs/avalanche-monitoring/blob/69e4c404aa04f3ecf5834d80bc31890c1d543cc3/grafana/dashboards/c_chain.json#L377).

We should similarly be able to sum these metrics, so that the sum is close to equal to the time we spend in block verification. If that is not the case, we may be spending an unexpected amount of time somewhere unexpected and have room for improvement or at least to better understand where execution is spending time.

aaronbuchwald:
Start on this is here: https://github.com/ava-labs/avalanchego/blob/ed0f66caee4dae03279ab832182b1ce9426dc2a8/tests/reexecute/c/metrics.go#L115

[...]

PR #4369, opened 2025-09-30
feat(reexecution/c): add metervm metrics

RodrigoVillar:
## Why this should be merged

As mentioned in #4282, it would be great if we could have a more precise assessment for the re-execution test rather than printing the average `mGas/s` for the entire benchmark. The starting point point for increasing the precision of the re-execution test would be to record the time spent in block parsing/verification/acceptance (whose sum should equal the total re-execution benchmark time). 

## How this works

- Adds metervm metrics to re-execution test
- Adds the metric unit `ms/gGas` to allow for a more straightforward comparison between metrics.

## How this was tested

CI + ran re-execution test locally to verify that the sum of block parsing/verification/acceptance is just about equal to the total re-execution benchmark time.

No


tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -592,27 +605,3 @@ func parseCustomLabels(labelsStr string) (map[string]string, error) {
...
 	return labels, nil
 }
-
-func getTopLevelMetrics(b *testing.B, registry prometheus.Gatherer, elapsed time.Duration) {
-	r := require.New(b)
-
-	gasUsed, err := getCounterMetricValue(registry, "avalanche_evm_eth_chain_block_gas_used_processed")
-	r.NoError(err)
-	mgasPerSecond := gasUsed / 1_000_000 / elapsed.Seconds()
-	b.ReportMetric(mgasPerSecond, "mgas/s")
```
aaronbuchwald, line 610: We should keep mgas/s as even if we are going to provide a better breakdown, it should still be our primary top-level metric.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -269,6 +275,13 @@ func newMainnetCChainVM(
 		ids.Empty:       constants.PrimaryNetworkID,
 	}
 
+	meterVMRegistry, err := metrics.MakeAndRegister(meterVMGatherer, "C")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create meterVM registry: %w", err)
+	}
```
aaronbuchwald: Can we pass in the fully formed registry to pass to `metervm.NewBlockVM` rather than splitting it inside of this function and before passing it in?

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Done: 25dbab1

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Done: 6345add

tests/reexecute/c/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, metric := range meterVMMetrics {
+		metricVal, err := getGaugeMetricValue(registry, metric.query)
+		r.NoError(err)
+
+		metricValMS := (metricVal / nsPerMs) / totalGGas
+		totalMSTrackedPerGGas += metricValMS
+		b.ReportMetric(metricValMS, metric.name+"_ms/gGas")
+	}
+
+	b.ReportMetric(totalMSTrackedPerGGas, "tracked_ms/gGas")
```
aaronbuchwald: What is the purpose of reporting this as a top-level metric? Anything reported this way will be archived and graphed.

tests/reexecute/c/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	return 0, fmt.Errorf("metric %s not found", query)
+}
+
+type topLevelMetric struct {
+	name  string
+	query string
+}
```
aaronbuchwald, line 76: Now that we using both gauge and counters, should the metrics themselves describe which to use rather than embedding this assumption in the function below?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -168,6 +169,9 @@ func benchmarkReexecuteRange(
 	vmMultiGatherer := metrics.NewPrefixGatherer()
 	r.NoError(prefixGatherer.Register("avalanche_evm", vmMultiGatherer))
 
+	meterVMRegistry := prometheus.NewRegistry()
+	r.NoError(prefixGatherer.Register("avalanche_meterchainvm_C", meterVMRegistry))
```
aaronbuchwald: Can you link example metrics from this run and DataDog showing that the metrics are reported the same way in production as they do here?

tests/reexecute/c/metrics.go
RodrigoVillar: 1. This previously existed in your metrics PR [[ref](https://github.com/ava-labs/avalanchego/blob/ed0f66caee4dae03279ab832182b1ce9426dc2a8/tests/reexecute/c/metrics.go#L133)], so I thought this was a metric you wanted to be printed out
2. Printing this metric allows for easy verification that `sum(parse + verify + accept) = total time per gGas`

Happy to remove this metric if you think it makes sense.

aaronbuchwald:
Can we use mgas/s and ggas/s rather than camel case? That's how I've typically seen it and follows the current use of `mgas/s`.

tests/reexecute/c/metrics.go
RodrigoVillar: Added a `metricKind` type for metrics to describe themselves: 63899f1

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Modified metervm prefix to match production naming here: 23a47e3

Datadog: https://app.datadoghq.com/metric/explorer?fromUser=false&start=1759252380002&end=1759255980002&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGABuMMiVGJgimWwwOAB0tfWNcE75Oo0YohWqThiZ2fIIaXiqwH2iyJk8IDwAulSu7niYoeHrqpsYMaXxS6sgGnJoOaDydYgIOUk4ME6ZWxoQmYlool1yisrpb5QL4-Jz0JjKERuDxoJb8CD2TBYJx-B7fERKFY8PhneTfBAAYSkwhgKBEmzQPCAA

Prometheus: https://prometheus-poc.avax-dev.network/query?g0.expr=avalanche_meterchainvm_accept_sum%7Bchain%3D%22C%22%2C+gh_run_id%3D%2218139163525%22%7D&g0.show_tree=0&g0.tab=graph&g0.range_input=1h&g0.res_type=auto&g0.res_density=medium&g0.display_mode=lines&g0.show_exemplars=0

The prometheus logs correspond from this GH job run: https://github.com/ava-labs/avalanchego/actions/runs/18139163525/job/51625555044?pr=4369

tests/reexecute/c/metrics.go
aaronbuchwald: Yup, that was in a mid-way state for debugging. I don't think we should report this because it does not make a useful top-level metric to be archived imo.

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: LGTM

aaronbuchwald:
Note: need to ensure that the newly added metrics show up as desired in the benchmarks page.

tests/reexecute/c/metrics.go
RodrigoVillar: Removed `totalMSTrackedPerGGas` here: cbc6298

Elvis339 (approved):

PR #4370, opened 2025-09-30
StatelessGraniteBlock types

iansuvak:
## Why this should be merged

Splits out the new type and serialization changes out from the epoching implementation in #4238 to shrink that PR. 

## How this works

Contains implementations of the new types as well as `statelessBlockMetadata` and the `Build` method, but all instances of calls to it are entering emptyEpoch and verification just verifies that the epoch is not set.  

## How this was tested

Existing CI


vms/proposervm/block/parse_test.go
```diff
@@ -81,6 +109,11 @@ func TestParse(t *testing.T) {
 	parentID := ids.ID{1}
 	timestamp := time.Unix(123, 0)
 	pChainHeight := uint64(2)
+	pChainEpoch := Epoch{
+		PChainHeight: 2,
+		Number:       0,
+		StartTime:    timestamp.Unix(),
+	}
```
geoff-vball, line 112: This should be invalid, right? We should either have the empty epoch struct, or `Number` should be `>= 1`

vms/proposervm/block/block.go
```diff
@@ -43,94 +46,163 @@ type SignedBlock interface {
...
+}
+
+func (b *statelessGraniteBlock) initialize(bytes []byte) error {
+	return b.statelessBlockMetadata.initialize(&b.StatelessGraniteBlock.StatelessBlock, b.Signature, bytes)
+}
+
+func (b *statelessGraniteBlock) verify(chainID ids.ID) error {
+	if b.StatelessGraniteBlock.Epoch.Number == 0 {
+		return errZeroEpoch
+	}
```
geoff-vball: So I think what @StephenButtolph was saying is that we should check if granite has been activated here. Updated nodes could create a malicious block with `Epoch.Number > 0`, which would pass verification on updated nodes but fail on un-updated nodes (at least in this PR). In the full PR, I think the check for `nextPChainEpoch()` would catch this discrepancy.

vms/proposervm/block/block.go
```diff
@@ -16,9 +16,11 @@ import (
 
 var (
 	_ SignedBlock = (*statelessBlock)(nil)
+	_ SignedBlock = (*statelessGraniteBlock)(nil)
 
 	errUnexpectedSignature = errors.New("signature provided when none was expected")
 	errInvalidCertificate  = errors.New("invalid certificate")
+	errZeroEpoch           = errors.New("epoch number must be greater than 0 prior to granite")
```
StephenButtolph:
```suggestion
	errZeroEpoch           = errors.New("epoch number must be greater than 0 after granite")
```

vms/proposervm/block/block.go
```diff
@@ -43,94 +46,163 @@ type SignedBlock interface {
...
+
+func (b *statelessGraniteBlock) PChainEpoch() Epoch {
+	return b.StatelessGraniteBlock.Epoch
+}
+
+func (b *statelessGraniteBlock) initialize(bytes []byte) error {
+	return b.statelessBlockMetadata.initialize(&b.StatelessGraniteBlock.StatelessBlock, b.Signature, bytes)
+}
+
+func (b *statelessGraniteBlock) verify(chainID ids.ID) error {
```
StephenButtolph, line 203: I think we should probably update the doc on `Parse` and `ParseWithoutVerification` after this change. It currently explicitly says that verification is specific to signature verification... But I think we should just say that the verification guarantees that the block has a valid representation (or something)

vms/proposervm/block/build.go
```diff
@@ -18,17 +18,30 @@ func BuildUnsigned(
...
 			Timestamp:    timestamp.Unix(),
 			PChainHeight: pChainHeight,
 			Certificate:  nil,
 			Block:        blockBytes,
-		},
-		timestamp: timestamp,
+		}
+		block SignedBlock
+	)
+	if epoch.Number == 0 {
```
StephenButtolph: Do we need to use the number as a special field? I feel like it would be a bit more clear (at least within this PR) to do `if epoch == (Epoch{}) {` rather than `if epoch.Number == 0`.

vms/proposervm/block/parse_test.go
iansuvak: Yes sorry this was just from the existing PR but agree should be invalid. Will expand tests to cover this case since right now only the `Number` field is checked.

vms/proposervm/block/block.go
iansuvak: `Epoch.Number` right now is globally checked to be zero in `postforkcomponents.Verify` so the verification should fail on updated nodes as well.

vms/proposervm/pre_fork_block.go
StephenButtolph, line 97: We need to verify the epoch of the child block somewhere here.

vms/proposervm/pre_fork_block_test.go
StephenButtolph, line 1: Does it make sense to add a test for the epoch verification (pre-Granite) in this PR?

vms/proposervm/block.go
```diff
@@ -105,6 +107,11 @@ func (p *postForkCommonComponents) Verify(
 		return errPChainHeightNotMonotonic
 	}
 
+	// TODO: Remove this check once the implementation is in place.
+	if child.PChainEpoch().Number != 0 {
+		return errEpochNotZero
+	}
```
StephenButtolph, line 113: Does it make sense to add a test for this (pre-Granite) in this PR?

vms/proposervm/block/parse_test.go
```diff
@@ -179,6 +213,7 @@ func TestParseBytes(t *testing.T) {
 		expectedErr error
 	}{
 		{
+			// TODO: Include the epoch information in the serialized block
```
StephenButtolph: Should this be addressed in this PR?

vms/proposervm/block/parse_test.go
StephenButtolph, line 104: Should we be verifying that both pre-Granite and post-Granite blocks are able to be parsed here?

vms/proposervm/block/build_test.go
StephenButtolph, line 1: Similar comment as to parse, we should probably have tests for both pre-granite and post-granite building here

vms/proposervm/block/block.go
geoff-vball: So, I think that would only check against honestly produced blocks. If someone maliciously creates a granite block before granite is activated with 
```
Epoch{
		PChainHeight: 1,
		Number:       0,
		StartTime:    1,
	}
```
I think it either won't get deserialized properly on un-updated nodes, or the signature won't match the serialization (because it will match - unless the bytes which include the epoch get deserialized fine, and we only check the signature over the bytes representation

vms/proposervm/pre_fork_block.go
iansuvak: Thanks addressed below

vms/proposervm/block/block.go
iansuvak: On any non-updated nodes the deserialization will fail. On updated nodes with this PR merged, the check in `postforkcomponents.Verify` will ensure that even if the block deserializes correctly it will fail verification if it deserializes to a non-zero epoch and any nodes constructing a post-granite block will require that it has non-zero epoch so this case shouldn't be reachable

vms/proposervm/pre_fork_block_test.go
iansuvak: Added `TestPreForkBlock_NonZeroEpoch`

vms/proposervm/block/parse_test.go
iansuvak: removed the comment since explicit tests were added

vms/proposervm/block/block.go
```diff
@@ -43,94 +46,163 @@ type SignedBlock interface {
...
+	if b.StatelessGraniteBlock.Epoch.Number == 0 {
```
StephenButtolph:
```suggestion
	if b.StatelessGraniteBlock.Epoch == (Epoch{}) {
```

vms/proposervm/block/block.go
```diff
@@ -16,9 +16,11 @@ import (
 
 var (
 	_ SignedBlock = (*statelessBlock)(nil)
+	_ SignedBlock = (*statelessGraniteBlock)(nil)
 
 	errUnexpectedSignature = errors.New("signature provided when none was expected")
 	errInvalidCertificate  = errors.New("invalid certificate")
+	errZeroEpoch           = errors.New("epoch number must be greater than 0 after granite")
```
StephenButtolph:
```suggestion
	errZeroEpoch           = errors.New("epoch must be provided after granite")
```

vms/proposervm/block/build.go
StephenButtolph:
```suggestion
	if epoch == (Epoch{}) {
```

vms/proposervm/block.go
```diff
@@ -38,6 +38,7 @@ var (
 	errProposerMismatch         = errors.New("proposer mismatch")
 	errProposersNotActivated    = errors.New("proposers haven't been activated yet")
 	errPChainHeightTooLow       = errors.New("block P-chain height is too low")
+	errEpochNotZero             = errors.New("epoch number must be 0 prior to granite")
```
StephenButtolph:
```suggestion
	errEpochNotZero             = errors.New("epoch must not be provided prior to granite")
```

vms/proposervm/block/parse_test.go
```diff
@@ -196,7 +268,9 @@ func TestParseBytes(t *testing.T) {
 			bytes, err := hex.DecodeString(test.hex)
 			require.NoError(err)
 
-			_, err = Parse(bytes, chainID)
+			blk, err := Parse(bytes, chainID)
+
+			require.IsType(test.expectedType, blk)
```
StephenButtolph: It feels weird to me to test the returned value even though an error it returned.
```suggestion
			_, err = Parse(bytes, chainID)
```
I think this was added to make sure it was parsed as a `statelessGraniteBlock`, but we already are doing this based on the expected error (I think).

vms/proposervm/block_test.go
```diff
@@ -415,3 +417,103 @@ func TestPreEtnaContextPChainHeight(t *testing.T) {
...
+	postForkNonZeroEpochBlock := &postForkBlock{
+		SignedBlock: nonZeroEpochStatelessBlock,
+		postForkCommonComponents: postForkCommonComponents{
+			vm:       vm,
+			innerBlk: innerChildBlock,
+		},
+	}
+	err = blk.Verify(context.Background(), parentTimestamp, pChainHeight, postForkNonZeroEpochBlock)
+	require.ErrorIs(err, errEpochNotZero)
+}
```
StephenButtolph, line 459: While admirable to try to test this without creating a VM instance... I feel like it would just be easier to use the existing test helpers to setup the boilerplate.
```suggestion
// Confirm that VM rejects blocks with non-zero epoch prior to granite upgrade activation
func TestPreGraniteBlock_NonZeroEpoch(t *testing.T) {
	require := require.New(t)

	var (
		activationTime = upgrade.InitiallyActiveTime
		durangoTime    = upgrade.InitiallyActiveTime
	)
	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
	defer func() {
		require.NoError(proVM.Shutdown(context.Background()))
	}()

	innerBlk := snowmantest.BuildChild(snowmantest.Genesis)
	slb, err := statelessblock.Build(
		proVM.preferred,
		proVM.Time(),
		100, // pChainHeight,
		statelessblock.Epoch{
			PChainHeight: 1,
			Number:       1,
			StartTime:    1,
		},
		proVM.StakingCertLeaf,
		innerBlk.Bytes(),
		proVM.ctx.ChainID,
		proVM.StakingLeafSigner,
	)
	require.NoError(err)
	proBlk := postForkBlock{
		SignedBlock: slb,
		postForkCommonComponents: postForkCommonComponents{
			vm:       proVM,
			innerBlk: innerBlk,
		},
	}
	err = proBlk.Verify(context.Background())
	require.ErrorIs(err, errEpochNotZero)
}
```
This also remove all usage of mocks which we've been trying to make progress towards to make our test coverage more useful.

vms/proposervm/block/parse_test.go
iansuvak: Yeah we are but it's implicit. Happy to remove it but here wanted to feel more of a clue of what the harcoded encoded bytes actually are.

vms/proposervm/block/block.go
iansuvak: I have no idea how I missed this when this is exactly where the initial comment was 😞

StephenButtolph (approved):

geoff-vball (approved):

michaelkaplan13 (approved):

PR #4371, opened 2025-09-30
Refactor `initTestProposerVM` to take in `upgradetest.fork` 

JonathanOppenheimer:
## Why this should be merged
This PR modifies the testing setup function to directly take in a fork, rather than two timestamps. This PR follows up on the cleanup in https://github.com/ava-labs/avalanchego/pull/4351. 

If different forks/values are desired for tests, I am happy to change them. 

## How this was tested
Existing UT 

No

### DIFF
--- a/vms/proposervm/block_test.go
+++ b/vms/proposervm/block_test.go
@@ -413,11 +413,7 @@ func TestPreEtnaContextPChainHeight(t *testing.T) {
 func TestPreGraniteBlock_NonZeroEpoch(t *testing.T) {
 	require := require.New(t)
 
-	var (
-		activationTime = upgrade.InitiallyActiveTime
-		durangoTime    = upgrade.InitiallyActiveTime
-	)
-	_, _, proVM, _ := initTestProposerVM(t, activationTime, durangoTime, 0)
+	_, _, proVM, _ := initTestProposerVM(t, upgradetest.Latest, 0)
 	defer func() {
 		require.NoError(proVM.Shutdown(context.Background()))
 	}()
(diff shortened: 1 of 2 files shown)


vms/proposervm/batched_vm_test.go
```diff
@@ -939,10 +942,17 @@ func initTestRemoteProposerVM(
...
+		upgrades = upgradetest.GetConfigWithUpgradeTime(fork, forkActivationTime[0])
+	} else {
+		upgrades = upgradetest.GetConfig(fork)
+	}
+
 	proVM := New(
 		coreVM,
 		Config{
-			Upgrades:            upgradetest.GetConfigWithUpgradeTime(fork, forkActivationTime),
+			Upgrades:            upgrades,
```
JonathanOppenheimer, line 956: Here is the meaningful difference.

StephenButtolph (approved):

vms/proposervm/vm_test.go
```diff
@@ -130,14 +133,18 @@ func initTestProposerVM(
 		}
 	}
 
+	var upgrades upgrade.Config
+	if len(forkActivationTime) > 0 {
+		upgrades = upgradetest.GetConfigWithUpgradeTime(fork, forkActivationTime[0])
```
DracoLi, line 138: why does forkActivationTime need to be a slice if only the first ~~time~~ item is used?

JonathanOppenheimer: This might be my lack of Go knowledge, but from my understanding, variadic parameters are automatically slices.

> If f is [variadic](https://go.dev/ref/spec#Function_types) with a final parameter p of type ...T, then within f the type of p is equivalent to type []T.

I thought this is how optional parameters were done in Go, and did it this way for cleanliness. If there's a better way, please let me know!

DracoLi: ah yeah sorry my bad - I forgot this is just an optional param.

PR #4373, opened 2025-09-30
Add warp accessors to `validators.State`

StephenButtolph:
## Why this should be merged

[ACP-181](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/181-p-chain-epoched-views/README.md) adds epoched views of the P-chain for all warp verification.

For a given P-chain height and subnetID, warp message verification can be cached through the generation of the canonical validator set. After that, message-specific logic is performed. Such as aggregating a subset of public keys or verifying the aggregate signature.

The intent of change is to introduce caching inside the `validators.State` implementation to allow caching to be handled once by the P-chain.

## How this works

This PR adds two functions: `GetWarpValidatorSets(height)` and `GetWarpValidatorSet(height, subnetID)` to the `validators.State` interface and implements them on all of the `validators.State` implementations.

## How this was tested

Added tests for all new implementations:
- [x] `Client` @ `snow/validators/gvalidators/validator_state_client.go`
- [x] `manager` @ `vms/platformvm/validators/manager.go`

These implementations are "trivial" and therefore remain untested.
- `lockedState` @ `snow/validators/state.go`
- `noValidators` @ `snow/validators/state.go`
- `tracedState` @ `snow/validators/traced_state.go`

no

### DIFF
--- a/proto/validatorstate/validator_state.proto
+++ b/proto/validatorstate/validator_state.proto
@@ -14,6 +14,13 @@ service ValidatorState {
   rpc GetCurrentHeight(google.protobuf.Empty) returns (GetCurrentHeightResponse);
   // GetSubnetID returns the subnetID of the provided chain.
   rpc GetSubnetID(GetSubnetIDRequest) returns (GetSubnetIDResponse);
+  // GetWarpValidatorSets returns the canonical warp validator set for all
+  // subnets at the requested P-chain height. If a subnet doesn't have a valid
+  // canonical warp validator set, it is omitted from the response.
+  rpc GetWarpValidatorSets(GetWarpValidatorSetsRequest) returns (GetWarpValidatorSetsResponse);
+  // GetWarpValidatorSet returns the canonical warp validator set for the
+  // provided subnet at the requested P-chain height.
+  rpc GetWarpValidatorSet(GetWarpValidatorSetRequest) returns (GetWarpValidatorSetResponse);
   // GetValidatorSet returns the weights of the nodeIDs for the provided
   // subnet at the requested P-chain height.
   rpc GetValidatorSet(GetValidatorSetRequest) returns (GetValidatorSetResponse);
@@ -38,6 +45,15 @@ message GetSubnetIDResponse {
   bytes subnet_id = 1;
 }
 
+message GetWarpValidatorSetsRequest {
+  uint64 height = 1;
+}
+
+message GetWarpValidatorSetRequest {
+  uint64 height = 1;
+  bytes subnet_id = 2;
+}
+
 message GetValidatorSetRequest {
   uint64 height = 1;
   bytes subnet_id = 2;
@@ -56,6 +72,27 @@ message Validator {
   bool is_active = 6; // can be empty
   bytes validation_id = 7; // can be empty
   bool is_l1_validator = 8; // can be empty
+}
+
+message GetWarpValidatorSetsResponse {
+  repeated WarpValidatorSet validator_sets = 1;
+}
+
+message GetWarpValidatorSetResponse {
+  uint64 total_weight = 1;
+  repeated WarpValidator validators = 2;
+}
+
+message WarpValidatorSet {
+  bytes subnet_id = 1;
+  uint64 total_weight = 2;
+  repeated WarpValidator validators = 3;
+}
+
+message WarpValidator {
+  bytes public_key = 1; // Uncompressed public key, must not be empty
+  uint64 weight = 2;
+  repeated bytes node_ids = 3; // must not be empty
 }
 
 message GetValidatorSetResponse {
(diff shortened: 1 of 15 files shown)


snow/validators/gvalidators/validator_state_client.go
```diff
@@ -142,3 +210,30 @@ func (c *Client) GetCurrentValidatorSet(
...
+			nodeIDs[j] = nodeID
+		}
+		pkBytes := vdr.GetPublicKey()
+		pk := bls.PublicKeyFromValidUncompressedBytes(pkBytes)
+		if pk == nil {
+			return nil, errFailedPublicKeyDeserialize
+		}
+		vdrs[i] = &validators.Warp{
+			PublicKey:      pk,
+			PublicKeyBytes: pkBytes,
```
yacovm, line 233: why do we also keep the raw form of the PK? Is it only for sorting?

vms/platformvm/validators/manager.go
```diff
@@ -230,12 +230,39 @@ func (m *manager) getCurrentHeight(context.Context) (uint64, error) {
...
+) (map[ids.ID]validators.WarpSet, error) {
+	allValidators, err := m.makeAllValidatorSets(ctx, targetHeight)
+	if err != nil {
+		return nil, err
+	}
+
+	validatorSets := make(map[ids.ID]validators.WarpSet, len(allValidators))
+	for subnetID, vdrSet := range allValidators {
+		ws, err := validators.FlattenValidatorSet(vdrSet)
+		if err != nil {
```
yacovm, line 245: shall we log here why we can't do it? This shouldn't happen by looking at the current code of flatten but maybe in the future it will change.

vms/platformvm/validators/manager_test.go
```diff
@@ -123,3 +130,211 @@ func TestGetValidatorSet_AfterEtna(t *testing.T) {
...
+		pk0Bytes = bls.PublicKeyToUncompressedBytes(pk0)
+		pk1      = newPublicKey(t)
+		pk1Bytes = bls.PublicKeyToUncompressedBytes(pk1)
+	)
+	if bytes.Compare(pk0Bytes, pk1Bytes) > 0 {
+		pk0, pk1 = pk1, pk0
+		pk0Bytes, pk1Bytes = pk1Bytes, pk0Bytes
+	}
+
+	var (
```
yacovm, line 155: a bit of a nit but do we really need to populate all these fields for the test?

yacovm (approved):

snow/validators/gvalidators/validator_state_client.go
StephenButtolph: It avoids repeatedly re-generating this value from the pk.

vms/platformvm/validators/manager.go
StephenButtolph: This actually _can_ happen (there is a test which triggers this case).

We don't have access to a logger here... I suppose I could add one if needed... But I'm not sure it's really needed

proto/validatorstate/validator_state.proto
```diff
@@ -58,6 +74,27 @@ message Validator {
...
+message WarpValidatorSet {
+  bytes subnet_id = 1;
+  uint64 total_weight = 2;
+  repeated WarpValidator validators = 3;
+}
+
+message WarpValidator {
+  bytes public_key = 1; // Uncompressed public key, must not be empty
+  uint64 weight = 2;
+  repeated bytes node_ids = 3; // must not be empty
```
michaelkaplan13, line 95: [No action] Relayers/Aggregators will need the `node_ids` to know which nodes to query for signatures, but I don't think VMs actually need them to verify an ICM signature.

I'm guessing that this needs to support both use cases, and isn't worth over optimizing for the VM case?

vms/platformvm/validators/manager_test.go
StephenButtolph: ya you're right. I removed the non-required fields.

proto/validatorstate/validator_state.proto
StephenButtolph: Yeah the only usage of `NodeIDs` is with the aggregator code. I don't think it will really matter because this evaluation will be cached.

snow/validators/state.go
```diff
@@ -24,6 +24,28 @@ type State interface {
 	// GetSubnetID returns the subnetID of the provided chain.
 	GetSubnetID(ctx context.Context, chainID ids.ID) (ids.ID, error)
 
+	// GetWarpValidatorSets returns the canonical warp validator set for all
+	// subnets at the requested P-chain height.
+	//
+	// If a subnet is not present in the returned map, that indicates that the
+	// subnet is not currently able to send warp messages.
```
michaelkaplan13:
```suggestion
	// subnet is not currently able to produce valid warp message signatures.
```

Nit: "send" is a bit of misnomer IMO

PR #4374, opened 2025-09-30
Uptime Tracker refactor

joshua-kim:

### DIFF
--- a/vms/evm/uptimetracker/uptime_tracker.go
+++ b/vms/evm/uptimetracker/uptime_tracker.go
@@ -61,6 +61,7 @@ func New(
 // GetUptime returns the uptime of the validator corresponding to validationID
 func (u *UptimeTracker) GetUptime(validationID ids.ID) (
 	time.Duration,
+	time.Time,
 	bool,
 	error,
 ) {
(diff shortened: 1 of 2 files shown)


vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -15,296 +15,276 @@ import (
...
-		clock:          clock,
+		manager: uptime.NewManager(s, clock),
+		clock:  clock,
+		height: -1,
+		state:   s,
 	}, nil
 }
 
-// GetValidator returns the validator data for the given validationID
-func (u *UptimeTracker) GetValidator(vID ids.ID) (Validator, bool) {
```
joshua-kim, line 74: This looks confusing - I'm getting rid of all of the Get* apis. I don't think we need them - EVM already knows about the P-Chain, so we could just grab the validator set from there, and get the uptimes from `UptimeTracker`.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -15,296 +15,276 @@ import (
...
-		ValidationID:   data.validationID,
-		NodeID:         data.NodeID,
-		StartTimestamp: data.StartTime,
-		IsActive:       data.IsActive,
-		Weight:         data.Weight,
-		IsL1Validator:  data.IsL1Validator,
-	}, true
-}
+// GetUptime returns the uptime of the validator corresponding to validationID
+func (u *UptimeTracker) GetUptime(validationID ids.ID) (
```
joshua-kim, line 62: Just realized we need the last updated time. Gimme a second to push something.

joshua-kim: Pushed.

PR #4376, opened 2025-10-01
fix: correct ledger buffer size calculation for Sign function

felipemadero:
## Why this should be merged
The previous buffer size calculation incorrectly assumed signing and change paths were stored separately in the ledger buffer. However, the ledger library deduplicates these paths, and the buffer format includes a 1-byte path count prefix. This fixes the calculation to accurately reflect the actual buffer usage.

### DIFF
--- a/x/blockdb/README.md
+++ b/x/blockdb/README.md
@@ -148,16 +148,16 @@ defer db.Close()
 // Write a block
 height := uint64(100)
 blockData := []byte("block data")
-err := db.WriteBlock(height, blockData)
+err := db.Put(height, blockData)
 if err != nil {
     fmt.Println("Error writing block:", err)
     return
 }
 
 // Read a block
-blockData, err := db.ReadBlock(height)
+blockData, err := db.Get(height)
 if err != nil {
-    if errors.Is(err, blockdb.ErrBlockNotFound) {
+    if errors.Is(err, database.ErrNotFound) {
         fmt.Println("Block doesn't exist at this height")
         return
     }
--- a/x/blockdb/database.go
+++ b/x/blockdb/database.go
@@ -20,6 +20,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/avalanchego/cache/lru"
+	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/utils/compression"
 	"github.com/ava-labs/avalanchego/utils/logging"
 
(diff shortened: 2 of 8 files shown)


utils/crypto/ledger/ledger.go
```diff
@@ -103,15 +108,15 @@ func (l *Ledger) SignHash(hash []byte, addressIndices []uint32) ([][]byte, error
...
-		// app. When the tx that is being signed is too large, we sign with hash
-		// instead.
-		//
-		// Ref: https://github.com/ava-labs/avalanche-wallet-sdk/blob/9a71f05e424e06b94eaccf21fd32d7983ed1b040/src/Wallet/Ledger/provider/ZondaxProvider.ts#L68
+	// We pass addressIndices both as signing paths and change paths.
+	// The ledger library deduplicates them, so the buffer contains len(addressIndices) paths.
+	// Buffer format: 1 byte (path count) + paths + transaction bytes
+	//
+	// Ref: https://github.com/ava-labs/ledger-avalanche-go/blob/main/common.go (ConcatMessageAndChangePath)
+	bufferSize := 1 + len(addressIndices)*ledgerPathSize + len(txBytes)
```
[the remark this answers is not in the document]
felipemadero: ledgerPathSize does take into account the correct size of 9 on the ledger library, adding comments for it

utils/crypto/ledger/ledger.go
```diff
@@ -103,15 +113,15 @@ func (l *Ledger) SignHash(hash []byte, addressIndices []uint32) ([][]byte, error
 }
```
yacovm, line 114: This is a general comment, unrelated to this PR, but - shouldn't this code be in https://github.com/ava-labs/ledger-avalanche-go ?

utils/crypto/ledger/ledger.go
```diff
@@ -103,15 +113,15 @@ func (l *Ledger) SignHash(hash []byte, addressIndices []uint32) ([][]byte, error
...
-		// There is a limit on the tx length that can be parsed by the ledger
-		// app. When the tx that is being signed is too large, we sign with hash
-		// instead.
-		//
-		// Ref: https://github.com/ava-labs/avalanche-wallet-sdk/blob/9a71f05e424e06b94eaccf21fd32d7983ed1b040/src/Wallet/Ledger/provider/ZondaxProvider.ts#L68
+	// We pass addressIndices both as signing paths and change paths.
+	// The ledger library deduplicates them, so the buffer contains len(addressIndices) paths.
+	// Buffer format: 1 byte (path count) + paths + transaction bytes
+	//
+	// Ref: https://github.com/ava-labs/ledger-avalanche-go/blob/main/common.go (ConcatMessageAndChangePath)
```
yacovm, line 120: The [typescript file](https://github.com/ava-labs/avalanche-wallet-sdk/blob/9a71f05e424e06b94eaccf21fd32d7983ed1b040/src/Wallet/Ledger/provider/ZondaxProvider.ts#L69) represents the previous implementation so why is the typescript file not changed accordingly? 

Also how does the golang file [represent](https://github.com/ava-labs/ledger-avalanche-go/blob/main/common.go#L195) the new logic?

felipemadero:

PR #4377, opened 2025-10-01
refactor: ledger interface to use public keys instead of addresses

felipemadero:
## Why this should be merged
The current ledger interface returns addresses (`ids.ShortID`) directly, which couples the ledger abstraction to a specific address format. This refactor changes the interface to return public keys instead, enabling:

- Support for multiple address formats (Avalanche and Ethereum) from the same ledger key
- **Satisfies `EthKeychain` interface, enabling ledger to be used to fully sign C-chain transactions on wallet**
  
## How this works
**Ledger Interface Changes:**
- `Ledger.Address()` → `Ledger.PubKey()` - returns `*secp256k1.PublicKey`
- `Ledger.Addresses()` → `Ledger.PubKeys()` - returns `[]*secp256k1.PublicKey`

**Ledger Keychain Changes:**
- Stores public keys internally instead of addresses
- Derives both Avalanche and Ethereum addresses from public keys
- Added `EthAddresses()` and `GetEth()` methods for EthKeychain implementation

**Implementation:**
- The ledger now queries the hardware device for public keys without requiring to provide HRP or display
- Public keys are cached with their ledger indices
- Addresses are derived on-demand using `pubKey.Address()` and `pubKey.EthAddress()`
- Both address types share the same derivation path from the ledger

## How this was tested
- unit tests
- ledger operations on CLI to deploy a L1 locally

### DIFF
--- a/utils/crypto/keychain/keychain.go
+++ b/utils/crypto/keychain/keychain.go
@@ -6,8 +6,13 @@ package keychain
 import (
 	"errors"
 	"fmt"
+	"math"
+
+	"github.com/ava-labs/libevm/common"
+	"golang.org/x/exp/maps"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
 	"github.com/ava-labs/avalanchego/utils/set"
 )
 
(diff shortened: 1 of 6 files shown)


utils/crypto/keychain/keychain.go
```diff
@@ -63,6 +74,9 @@ func NewLedgerKeychain(l Ledger, numToDerive int) (Keychain, error) {
 
 	indices := make([]uint32, numToDerive)
 	for i := range indices {
+		if i > math.MaxUint32 {
+			return nil, fmt.Errorf("index %d exceeds uint32 max", i)
+		}
```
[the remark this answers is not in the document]
felipemadero: moved the check upwards

utils/crypto/keychain/keychain.go
```diff
@@ -39,27 +44,36 @@ type Keychain interface {
...
+}
+
 // ledgerKeychain is an abstraction of the underlying ledger hardware device,
 // to be able to get a signer from a finite set of derived signers
 type ledgerKeychain struct {
-	ledger    Ledger
-	addrs     set.Set[ids.ShortID]
-	addrToIdx map[ids.ShortID]uint32
+	ledger           Ledger
+	avaAddrToKeyInfo map[ids.ShortID]*keyInfo    // Maps Avalanche addresses to key info
```
yacovm, line 57: why not `set.Set[keyInfo]` ?

utils/crypto/keychain/keychain.go
```diff
@@ -39,27 +44,36 @@ type Keychain interface {
...
+	ethAddrToKeyInfo map[common.Address]*keyInfo // Maps Ethereum addresses to key info
 }
 
 // ledgerSigner is an abstraction of the underlying ledger hardware device,
 // to be able sign for a specific address
 type ledgerSigner struct {
 	ledger Ledger
 	idx    uint32
-	addr   ids.ShortID
+	pubKey *secp256k1.PublicKey
```
yacovm, line 66: why a pointer vs a non pointer?

felipemadero:

PR #4378, opened 2025-10-01
More nits

joshua-kim:

### DIFF
--- a/vms/evm/uptimetracker/state.go
+++ b/vms/evm/uptimetracker/state.go
@@ -15,13 +15,10 @@ import (
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/utils/set"
 )
 
-const (
-	codecVersion  uint16         = 0
-	updatedStatus dbUpdateStatus = iota
-	deletedStatus
-)
+const codecVersion uint16 = 0
 
 var (
 	codecManager codec.Manager
@@ -41,8 +38,6 @@ func init() {
 	}
 }
 
-type dbUpdateStatus int
-
 type validator struct {
 	UpDuration    time.Duration `serialize:"true"`
 	LastUpdated   uint64        `serialize:"true"`
@@ -61,7 +56,8 @@ type state struct {
 
 	validators             map[ids.ID]*validator
 	nodeIDsToValidationIDs map[ids.NodeID]ids.ID
-	updates                map[ids.ID]dbUpdateStatus
+	updatedValidators set.Set[ids.ID]
+	deletedValidators set.Set[ids.ID]
 }
 
 func newState(db database.Database) (*state, error) {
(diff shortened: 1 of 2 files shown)


vms/evm/uptimetracker/state.go
```diff
@@ -176,33 +157,31 @@ func (s *state) deleteValidator(validationID ids.ID) bool {
...
-
-			if err := batch.Put(validationID[:], validatorBytes); err != nil {
-				return fmt.Errorf("failed to put validator: %w", err)
-			}
-		case deletedStatus:
-			if err := batch.Delete(validationID[:]); err != nil {
-				return fmt.Errorf("failed to delete validator: %w", err)
-			}
-		default:
-			return fmt.Errorf("unknown update status: %v", updateStatus)
```
joshua-kim, line 205: We can get rid of this impossible case since we know we only have two cases (updates/deletes)

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -175,70 +175,57 @@ func (u *UptimeTracker) update(
 	}
 
 	// Add or update validators
-	for validationID, next := range newValidators {
-		if prev, ok := u.state.getValidatorByValidationID(validationID); ok {
-			// We are updating a validator we know about
-			isActiveUpdated := prev.IsActive == next.IsActive
```
joshua-kim, line 181: This bool was very poorly named - it was named the opposite of what it actually did. This should have been `!=`, but it doesn't matter because we deleted this bool by returning the updated bool in the function below.

vms/evm/uptimetracker/state.go
```diff
@@ -138,33 +132,20 @@ func (s *state) GetStartTime(nodeID ids.NodeID) (time.Time, error) {
...
 
-func (s *state) updateValidator(
-	validationID ids.ID,
-	isActive bool,
-	weight uint64,
-) error {
-	v, ok := s.validators[validationID]
-	if !ok {
-		return database.ErrNotFound
-	}
```
joshua-kim, line 152: This not possible to hit - we always do a check that the validator was present before calling this function

vms/evm/uptimetracker/state.go
```diff
@@ -138,33 +132,20 @@ func (s *state) GetStartTime(nodeID ids.NodeID) (time.Time, error) {
...
+func (s *state) updateValidator(validationID ids.ID, isActive bool) bool {
+	v := s.validators[validationID]
 
-	updated := deletedStatus
```
joshua-kim, line 154: We never handle deletes here - in the calling code we check that we have this validator.

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -175,70 +175,57 @@ func (u *UptimeTracker) update(
...
 			})
 
-			if next.IsActive {
+			if newValidator.IsActive {
 				continue
 			}
 
-			// This validator not
-			if err := u.deactivate(next.NodeID); err != nil {
+			// This validator is not active and is treated is offline
```
joshua-kim, line 224: I probably messed up typing this comment

vms/evm/uptimetracker/state.go
```diff
@@ -138,33 +132,20 @@ func (s *state) GetStartTime(nodeID ids.NodeID) (time.Time, error) {
...
+	updated := false
 	if v.IsActive != isActive {
 		v.IsActive = isActive
-		updated = updatedStatus
-	}
-
-	if v.Weight != weight {
```
joshua-kim, line 160: I don't think we care about weights anymore? We used to since we served the validators through a Get api, but that api doesn't exist now.

JonathanOppenheimer (approved):

Issue #4372, opened 2025-09-30
Load testing against a live network


RodrigoVillar:
For load testing against a live network, we'll need to add the following features to the `load` package:

- [ ] The ability to print the Grafana dashboard link for client-side metrics when load testing an external network
- [ ] The ability to load in a list of private keys to use for workers
- [ ] The ability to specify which API nodes to target (i.e. being able to pass in a list of websocket URIs)

Issue #4290, opened 2025-09-17
Add Grafana Link To Re-Execution Benchmark


RodrigoVillar:
When running the re-execution benchmark locally with metrics scrapping enabled, a link to the corresponding Grafana dashboard is not printed. This is in contrast to running the benchmark in CI, which does print the Grafana dashboard link as it's handled by `run-monitored-tmpnet-cmd` GH action. 

Ideally, we should print the dashboard link, similar to how `tmpnet` already does it here:

https://github.com/ava-labs/avalanchego/blob/115407b7cc8243898791913c431d27451877ffa4/tests/fixture/tmpnet/network.go#L415-L418

PR #4383, opened 2025-10-02
feat(reexecution/c): print grafana link during local run

RodrigoVillar:
## Why this should be merged

As mentioned in #4290, the Grafana dashboard link is not printed out when running the re-execution tests locally. This PR fixes this by printing the link even when running locally.

The changes in the `tmpnet` package are relevant in #4372, as we'll call `NewGrafanaLink()` to generate the dashboard link when running against an external network.

## How this works

Prints the Grafana dashboard link of a re-execution test when running locally.

## How this was tested

CI + ran re-execution test to verify that the dashboard link is printed and is as expected.

No

### DIFF
--- a/tests/fixture/tmpnet/network.go
+++ b/tests/fixture/tmpnet/network.go
@@ -1120,7 +1120,7 @@ func getRPCVersion(log logging.Logger, command string, versionArgs ...string) (u
 // with the given UUID. The start and end times are accepted as strings to support the
 // use of Grafana's time range syntax (e.g. `now`, `now-1h`).
 func MetricsLinkForNetwork(networkUUID string, startTime string, endTime string) string {
-	return NewGrafanaLink(
+	return NewGrafanaURI(
 		networkUUID,
 		startTime,
 		endTime,
@@ -1128,8 +1128,8 @@ func MetricsLinkForNetwork(networkUUID string, startTime string, endTime string)
 	)
 }
 
-// NewGrafanaLink returns a Grafana dashboard link.
-func NewGrafanaLink(
+// NewGrafanaURI returns a Grafana dashboard URI.
+func NewGrafanaURI(
 	networkUUID string,
 	startTime string,
 	endTime string,
--- a/tests/reexecute/c/vm_reexecute_test.go
+++ b/tests/reexecute/c/vm_reexecute_test.go
@@ -601,7 +601,7 @@ func collectRegistry(tb testing.TB, log logging.Logger, name string, gatherer pr
 	log.Info("metrics available via grafana",
 		zap.String(
 			"url",
-			tmpnet.NewGrafanaLink(networkUUID, startTime, "", grafanaURI),
+			tmpnet.NewGrafanaURI(networkUUID, startTime, "", grafanaURI),
 		),
 	)
 }


tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -590,6 +591,19 @@ func collectRegistry(tb testing.TB, name string, gatherer prometheus.Gatherer, l
 		Labels:  labels,
 	}, true /* withGitHubLabels */)
 	r.NoError(err)
+
+	var (
+		dashboardPath = "d/Gl1I20mnk/c-chain"
+		grafanaURI    = tmpnet.DefaultBaseGrafanaURI + dashboardPath
```
maru-ava, line 597: (No action required) The reason the entire URI can be overridden via `GRAFANA_URI` is to ensure clean support for a local grafana (as per the tmpnet README). Optional for whether the re-execute test wants to maintain the same configurability.

tests/fixture/tmpnet/network.go
```diff
@@ -60,8 +60,11 @@ const (
 	// eth address: 0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC
 	HardHatKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
 
+	// Default base grafana URI.
+	DefaultBaseGrafanaURI = "https://grafana-poc.avax-dev.network/"
+
 	// Default grafana URI used to construct metrics links. Can be overridden by setting GRAFANA_URI env var.
-	defaultGrafanaURI = "https://grafana-poc.avax-dev.network/d/kBQpRdWnk/avalanche-main-dashboard"
+	defaultGrafanaURI = DefaultBaseGrafanaURI + "d/kBQpRdWnk/avalanche-main-dashboard"
```
aaronbuchwald, line 67: nit: can we use a more meaningful variable name here than `defaultGrafanaURI` ? This is pretty similar to `DefaultBaseGrafanaURI`, but something like `defaultGrafanaMainDashboardURI` would be more descriptive as the previous one makes clear it's the base, but this doesn't make clear how it extends the base.

tests/fixture/tmpnet/network.go
```diff
@@ -1117,13 +1120,28 @@ func getRPCVersion(log logging.Logger, command string, versionArgs ...string) (u
...
+	return NewGrafanaURI(
+		networkUUID,
+		startTime,
+		endTime,
+		GetEnvWithDefault("GRAFANA_URI", defaultGrafanaURI),
+	)
+}
+
+// NewGrafanaURI returns a Grafana dashboard URI.
+func NewGrafanaURI(
```
aaronbuchwald, line 1132: How do the two different names NewGrafanaURI and MetricsLinkForNetwork make it clear what the difference between the two is?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -590,6 +591,19 @@ func collectRegistry(tb testing.TB, name string, gatherer prometheus.Gatherer, l
 		Labels:  labels,
 	}, true /* withGitHubLabels */)
 	r.NoError(err)
+
+	var (
+		dashboardPath = "d/Gl1I20mnk/c-chain"
```
aaronbuchwald, line 596: nit: can we use a more descriptive name here?

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: +1 - good call to add this

aaronbuchwald:
Could you include in the PR description or a comment the output from running this locally to make it easier to verify as a reviewer?

Found in CI here: https://github.com/ava-labs/avalanchego/actions/runs/18201835265/job/51822459436?pr=4383#step:4:969

Screenshot 2025-10-02 at 15 20 31


LGTM

PR #4385, opened 2025-10-02
refactor: explicit libevm registration

ARR4N:
## Why this should be merged

Required by ava-labs/coreth#1305, which enforces explicit calls of `libevm` registration as a means of allowing both `coreth` and `subnet-evm` to be imported simultaneously.

## How this works

Bumps `coreth` to the merged commit of the aforementioned PR and calls `plugin/evm.RegisterAllLibEVMExtras()` in tests and `package main` as required. This call MUST NOT be performed outside of tests / `main`s as this will pollute other packages and undo the changes.

## How this was tested

Existing e2e tests.

No


x/merkledb/network_server_test.go
```diff
@@ -22,6 +24,11 @@ import (
 	xsync "github.com/ava-labs/avalanchego/x/sync"
 )
 
+func TestMain(m *testing.M) {
+	evm.RegisterAllLibEVMExtras()
+	os.Exit(m.Run())
+}
+
```
StephenButtolph: Why is this needed?

ARR4N: It's not; reverting. The test is flaky, I assumed it was broken because of my change, I added this and it "fixed" it. It times out and panics, so I saw the stack dump just like with unregistered extras. My bed is calling loudly!

PR #4387, opened 2025-10-03
feat(sync/customrawdb): migrate customrawdb package from coreth

### LINKED ISSUE #4386: Migrate `customrawdb` package from `coreth` to `avalanchego`
Migrate https://github.com/ava-labs/coreth/tree/master/plugin/evm/customrawdb to `vms/evm/sync`.

powerslider:
## Why this should be merged

Check #4386 

## How this works

- Move `plugin/evm/customrawdb` in `coreth` to `vms/evm/sync` in `avalanchego` as part of state sync code migration effort.
- Replace `geth`-style zero/nil read returns with explicit errors across customrawdb.
- Add exported sentinel errors:
  - `ErrEntryNotFound`, `ErrInvalidData`, `ErrStateSchemeConflict`, `ErrInvalidArgument`
- Update accessors to return errors consistently.
- Align tests with new errors and function signatures.
- Add more tests to cover all exported functionality. 

## How this was tested

Existing UT that come with the migrated code plus newly added UT to cover the entire package API.

no

resolves #4386

Signed-off-by: Tsvetan Dimitrov (tsvetan.dimitrov23@gmail.com)

### DIFF
--- a/vms/evm/sync/customrawdb/markers.go
+++ b/vms/evm/sync/customrawdb/markers.go
@@ -150,10 +150,7 @@ func WriteChainConfig[T any](db ethdb.KeyValueWriter, hash common.Hash, config *
 	if err != nil {
 		return fmt.Errorf("%w: %w", errFailedToMarshalUpgradeConfig, err)
 	}
-	if err := db.Put(upgradeConfigKey(hash), data); err != nil {
-		return err
-	}
-	return nil
+	return db.Put(upgradeConfigKey(hash), data)
 }
 
 // NewAccountSnapshotsIterator returns an iterator for walking all of the accounts in the snapshot
@@ -219,10 +216,6 @@ func readTimeMarker(db ethdb.KeyValueReader, key []byte) (time.Time, error) {
 	data, err := db.Get(key)
 	if err != nil {
 		return time.Time{}, err
-	}
-
-	if len(data) == 0 {
-		return time.Time{}, fmt.Errorf("%w: empty value", errInvalidData)
 	}
 
 	var unix uint64


vms/evm/sync/customrawdb/database_ext.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+)
+
+// InspectDatabase traverses the entire database and checks the size
+// of all different categories of data.
+func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
```
joshua-kim: This looks like dead code, why do we have this?

vms/evm/sync/customrawdb/db_test.go
joshua-kim, line 1: Why do we have this file? This contains dead code/a function that isn't actually a test.

vms/evm/sync/customrawdb/markers.go
joshua-kim, line 1: What does the `_ext` suffix on these files mean? Should we remove it?

vms/evm/sync/customrawdb/accessors_snapshot_ext.go
```diff
@@ -0,0 +1,46 @@
...
+// be used during updates, so a crash or failure will mark the entire snapshot
+// invalid.
+func DeleteSnapshotBlockHash(db ethdb.KeyValueWriter) {
+	if err := db.Delete(snapshotBlockHashKey); err != nil {
+		log.Crit("Failed to remove snapshot block hash", "err", err)
+	}
+}
+
+// IterateAccountSnapshots returns an iterator for walking all of the accounts in the snapshot
+func IterateAccountSnapshots(db ethdb.Iteratee) ethdb.Iterator {
```
joshua-kim: nit: Maybe name this `NewAccountSnapshotsIterator`? `IterateAccountSnapshots` makes it sound like this function is iterating over the database when it is just a constructor for an iterator.

Referencing `rawdb` - I can't tell what the intended convention is because they also do something similar by mixing the two naming styles ([ref](https://github.com/ethereum/go-ethereum/blob/e42af536c552538a8ca8c353a26669fd2e9eb150/core/rawdb/accessors_snapshot.go#L117-L118)).

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+	return common.BytesToHash(root), nil
+}
+
+// WriteSyncRoot writes root as the root of the main trie of the in-progress sync.
+func WriteSyncRoot(db ethdb.KeyValueWriter, root common.Hash) error {
+	return db.Put(syncRootKey, root[:])
+}
+
+// AddCodeToFetch adds a marker that we need to fetch the code for `hash`.
+func AddCodeToFetch(db ethdb.KeyValueWriter, hash common.Hash) {
```
joshua-kim: Naming in this package is inconsistent - we're calling this `Add` but the other apis are all `Write*` for db writes.

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+	if err := db.Put(codeToFetchKey(hash), nil); err != nil {
+		log.Crit("Failed to put code to fetch", "codeHash", hash, "err", err)
+	}
+}
+
+// DeleteCodeToFetch removes the marker that the code corresponding to `hash` needs to be fetched.
+func DeleteCodeToFetch(db ethdb.KeyValueWriter, hash common.Hash) {
+	if err := db.Delete(codeToFetchKey(hash)); err != nil {
+		log.Crit("Failed to delete code to fetch", "codeHash", hash, "err", err)
```
joshua-kim: nit: we call this `codeHash` but it's called `hash` in code.

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+
+// WriteSyncRoot writes root as the root of the main trie of the in-progress sync.
+func WriteSyncRoot(db ethdb.KeyValueWriter, root common.Hash) error {
+	return db.Put(syncRootKey, root[:])
+}
+
+// AddCodeToFetch adds a marker that we need to fetch the code for `hash`.
+func AddCodeToFetch(db ethdb.KeyValueWriter, hash common.Hash) {
+	if err := db.Put(codeToFetchKey(hash), nil); err != nil {
+		log.Crit("Failed to put code to fetch", "codeHash", hash, "err", err)
```
joshua-kim: nit: we call this `codeHash` but it's called `hash` in code.

vms/evm/sync/customrawdb/accessors_metadata.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return time.Unix(int64(unix), 0), nil
+}
+
+// WriteOfflinePruning writes a time marker of the last attempt to run offline pruning.
+// The marker is written when offline pruning completes and is deleted when the node
+// is started successfully with offline pruning disabled. This ensures users must
+// disable offline pruning and start their node successfully between runs of offline
+// pruning.
+func WriteOfflinePruning(db ethdb.KeyValueStore) error {
```
joshua-kim: I'm noticing some inconsistency that some of our functions return errors, while others don't.

It seems that we're following geth-style naming, so I'm also assuming we intended to follow geth's api style - referencing their state accessors [here](https://github.com/ethereum/go-ethereum/blob/master/core/rawdb/accessors_chain.go#L220) it seems that they don't return errors... they usually return a zero/nil value and log errors instead of letting the caller how to handle the error. It seems that they use `Crit` for failed writes to exit, and callers performing reads are expected to check the value returned.

We could deviate from geth's style but since we have inconsistency in our own package I want to ask if this was intended. If we're going to deviate from geth's style, I think it would just be easiest to return the error + let the caller handle the error and not have the caller infer failure from reasoning about the returned value of a read.

vms/evm/sync/customrawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,153 @@
...
+		log.Error("Invalid upgrade config JSON", "err", err)
+		return nil
+	}
+
+	return config
+}
+
+// WriteChainConfig writes the chain config settings to the database.
+// The provided [upgradeConfig] (any JSON-marshalable type) will be stored alongside the chain config.
+func WriteChainConfig[T any](db ethdb.KeyValueWriter, hash common.Hash, config *params.ChainConfig, upgradeConfig T) {
```
joshua-kim: This function signature looks like it changed from subnetevm and coreth. Why do we pass in `upgradeConfig`? It looks like in subnet evm we would write whatever was in `params.GetExtra` and in coreth we would write the config provided.

vms/evm/sync/customrawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,153 @@
...
+// WriteAcceptorTip writes `hash` as the last accepted block that has been fully processed.
+func WriteAcceptorTip(db ethdb.KeyValueWriter, hash common.Hash) error {
+	return db.Put(acceptorTipKey, hash[:])
+}
+
+// ReadAcceptorTip reads the hash of the last accepted block that was fully processed.
+// If there is no value present (the index is being initialized for the first time), then the
+// empty hash is returned.
+func ReadAcceptorTip(db ethdb.KeyValueReader) (common.Hash, error) {
+	has, err := db.Has(acceptorTipKey)
```
joshua-kim: nit: typically go names these vars `ok` for existence checks

vms/evm/sync/customrawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return common.Hash{}, err
+	}
+	if !has {
+		// If the index is not present on disk, the [acceptorTipKey] index has not been initialized yet.
```
joshua-kim: These `[brackets]` to my understanding are used to render in godoc and only work on exported code in either imported packages or in the same package - using it in-line in a function won't work and either way this var isn't exported.

vms/evm/sync/customrawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,153 @@
...
+		return common.Hash{}, err
+	}
+	if len(h) != common.HashLength {
+		return common.Hash{}, fmt.Errorf("value has incorrect length %d", len(h))
+	}
+	return common.BytesToHash(h), nil
+}
+
+// ReadChainConfig retrieves the consensus settings based on the given genesis hash.
+// The provided [upgradeConfig] (any JSON-unmarshalable type) will be populated if present on disk.
```
joshua-kim: Same comment here - i think for vars for a function \`backticks\` are used

vms/evm/sync/customrawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,153 @@
...
+func ReadChainConfig[T any](db ethdb.KeyValueReader, hash common.Hash, upgradeConfig *T) *params.ChainConfig {
```
joshua-kim: Same comment w.r.t the parameter for `upgradeConfig`.

vms/evm/sync/customrawdb/accessors_snapshot_ext.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/log"
+)
+
+// ReadSnapshotBlockHash retrieves the hash of the block whose state is contained in
+// the persisted snapshot.
+func ReadSnapshotBlockHash(db ethdb.KeyValueReader) common.Hash {
+	data, _ := db.Get(snapshotBlockHashKey)
+	if len(data) != common.HashLength {
+		return common.Hash{}
```
joshua-kim: It seems like when reads fail the rest of the geth rawdb Read* apis would error log

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/log"
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+)
+
+// ReadSyncRoot reads the root corresponding to the main trie of an in-progress
+// sync and returns common.Hash{} if no in-progress sync was found.
+func ReadSyncRoot(db ethdb.KeyValueReader) (common.Hash, error) {
+	has, err := db.Has(syncRootKey)
```
joshua-kim: Same `ok` naming nit here

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+// key returned from NewSyncSegmentsIterator.
+func UnpackSyncSegmentKey(keyBytes []byte) (common.Hash, []byte) {
+	keyBytes = keyBytes[len(syncSegmentsPrefix):] // skip prefix
+	root := common.BytesToHash(keyBytes[:common.HashLength])
+	start := keyBytes[common.HashLength:]
+	return root, start
+}
+
+// packSyncSegmentKey packs root and account into a key for storage in db.
+func packSyncSegmentKey(root common.Hash, start common.Hash) []byte {
```
joshua-kim: Why do we refactor this into a separate function if it's only used in one place, and the call-site is basically a direct pass through into this?

vms/evm/sync/customrawdb/accessors_state_sync.go
```diff
@@ -0,0 +1,223 @@
...
+// key returned from NewSyncStorageTriesIterator.
+func UnpackSyncStorageTrieKey(keyBytes []byte) (common.Hash, common.Hash) {
+	keyBytes = keyBytes[len(syncStorageTriesPrefix):] // skip prefix
+	root := common.BytesToHash(keyBytes[:common.HashLength])
+	account := common.BytesToHash(keyBytes[common.HashLength:])
+	return root, account
+}
+
+// packSyncStorageTrieKey packs root and account into a key for storage in db.
+func packSyncStorageTrieKey(root common.Hash, account common.Hash) []byte {
```
joshua-kim: Similar comment w.r.t refactored function

vms/evm/sync/customrawdb/accessors_state_sync_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"slices"
+	"testing"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/stretchr/testify/require"
+)
+
+func TestClearPrefix(t *testing.T) {
```
joshua-kim: It looks like `ClearAllSyncSegments` is the only function tested (and that this test name is wrong) - could we add coverage on this package api?

vms/evm/sync/customrawdb/schema.go
```diff
@@ -0,0 +1,64 @@
...
+
+package customrawdb
+
+import (
+	"github.com/ava-labs/libevm/common"
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+)
+
+var (
```
joshua-kim: Could we group all package-level global vars under a single block? If we want to group them by intent, we can add in-line comments

vms/evm/sync/customrawdb/schema_ext.go
```diff
@@ -0,0 +1,64 @@
...
+	syncPerformedPrefix = []byte("sync_performed")
+	// syncPerformedKeyLength is the length of the key for the sync performed metadata key,
+	// and is equal to [syncPerformedPrefix] + block number as uint64.
+	syncPerformedKeyLength = len(syncPerformedPrefix) + wrappers.LongLen
+)
+
+var FirewoodScheme = "firewood"
+
+// upgradeConfigKey = upgradeConfigPrefix + hash
+func upgradeConfigKey(hash common.Hash) []byte {
```
joshua-kim: This is a weird place for this function to live - it's defined in a file that's not close to where it's used. I think the file structure of this package more generally speaking could be improved - Go code usually will place code into files where there are clear boundaries between abstractions (big files with multiple types are fine). It's not clear to me what this file actually owns/represents.

vms/evm/sync/customrawdb/db_test.go
powerslider: Yes, I agree with this. I will check why this was historically added in the first place, but we can definitely remove it from here.

vms/evm/sync/customrawdb/database_ext.go
powerslider: It is used only in `database_ext_test.go`, so it is a bit unnecessary to be present here. I will check though historically why it was added in the first place since I am not the original author.

vms/evm/sync/customrawdb/accessors_metadata_ext.go
powerslider: I changed the signature exactly because I wanted to decouple the `params.GetExtra` which is local to `subnet-evm` from the implementation of this function which I want to be here. `coreth` uses the `libevm` implementation and `subnet-evm` uses this one. I will consult with the team if potentially we should keep this code here or maybe do something else.

powerslider:
@joshua-kim I added lots of new tests that enable having the following coverage for the package:
```
coverage: 86.7% of statements
ok  	github.com/ava-labs/avalanchego/vms/evm/sync/customrawdb	0.213s	coverage: 86.7% of statements
```
I fixed most of the minor comments and will get back on the rest of your points once I have more historical context from the team.

vms/evm/sync/customrawdb/markers.go
powerslider: Discussed with the team to remove these suffixes and they are removed now.

vms/evm/sync/customrawdb/db_test.go
powerslider: `InspectDatabase` is deleted from the file only `ParseStateScheme` remains.

vms/evm/sync/customrawdb/accessors_metadata.go
powerslider: Discussed with the team to omit the `geth` api style. So now all exported methods return errors and are well tested with additional tests. We should be consistent now.

powerslider:
@joshua-kim all comments should be fixed. Discussed with the team to omit `geth` api style and to remove the `_ext` suffixes. Please check these changes.

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+			// Not present initially.
+			_, err := tc.read(db)
+			require.ErrorIs(t, err, ErrEntryNotFound)
+
+			// Write marker and read back a reasonable recent time.
+			require.NoError(t, tc.write(db))
+			ts, err := tc.read(db)
+			require.NoError(t, err)
+			require.WithinDuration(t, time.Now(), ts, 2*time.Second)
```
joshua-kim: This test is inherently flaky - if we refactor so that we can pass in a timestamp like `WriteOfflinePruning(db, time.Now())` we can test against a timestamp known by the callsite and do an assertion against it.

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+			db := rawdb.NewMemoryDatabase()
+			for _, h := range tc.writes {
+				require.NoError(t, WriteAcceptorTip(db, h))
+			}
+			tip, err := ReadAcceptorTip(db)
+			if tc.expectedErr != nil {
+				require.ErrorIs(t, err, tc.expectedErr)
+			} else {
+				require.NoError(t, err)
+			}
```
joshua-kim: You can safely use `require.ErrorIs` even if the expected error asserted against is `nil`, so you don't have to do this if/else check.

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+	_, err := ReadAcceptorTip(db)
+	require.ErrorIs(t, err, ErrInvalidData)
+}
+
+func TestWriteAcceptorTip(t *testing.T) {
+	cases := []struct {
+		name        string
+		writes      []common.Hash
+		want        common.Hash
+		expectedErr error
```
joshua-kim: nit: we're mixing usage of `want` and `expected` naming in tests - I think the go idiom is technically `want*`, could we use consistent naming across tests?

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+			name:        "none",
+			writes:      nil,
+			want:        common.Hash{},
+			expectedErr: ErrEntryNotFound,
+		},
+		{
+			name:        "single_write",
+			writes:      []common.Hash{common.HexToHash("0xabc1")},
+			want:        common.HexToHash("0xabc1"),
+			expectedErr: nil,
```
joshua-kim: nit: It's up to your preference if you want to be explicit or not, but we don't have to populate zero values like `nil` and `common.Hash{}` since those are the default zero values.

vms/evm/sync/customrawdb/errors.go
```diff
@@ -0,0 +1,17 @@
...
+
+var (
+	// ErrEntryNotFound indicates the requested key/value was not present in the DB.
+	ErrEntryNotFound = errors.New("customrawdb: entry not found")
+	// ErrInvalidData indicates the stored value exists but is malformed or undecodable.
+	ErrInvalidData = errors.New("customrawdb: invalid data")
+	// ErrStateSchemeConflict indicates the provided state scheme conflicts with what is on disk.
+	ErrStateSchemeConflict = errors.New("customrawdb: state scheme conflict")
+	// ErrInvalidArgument indicates a caller-provided argument was invalid.
+	ErrInvalidArgument = errors.New("customrawdb: invalid argument")
```
joshua-kim: I think we need to export `ErrEntryNotFound`, but do we need to export these? It seems like if we hit these errors we'd never reason about the errors anyways.

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	if len(h) != common.HashLength {
+		return common.Hash{}, fmt.Errorf("%w: length %d", ErrInvalidData, len(h))
+	}
+	return common.BytesToHash(h), nil
+}
+
+// ReadChainConfig retrieves the consensus settings based on the given genesis hash.
+// The provided `upgradeConfig` (any JSON-unmarshalable type) will be populated if present on disk.
+func ReadChainConfig[T any](db ethdb.KeyValueReader, hash common.Hash, upgradeConfig *T) (*params.ChainConfig, error) {
```
joshua-kim, line 119: 1. Do we need `*T` if the caller specifies a pointer-type as `T`?
2. It seems weird that the comment specifies that `T` is known to be a json-unmarshable type, but our type constraint is `any`. Should we use `json.Unmarshaler`?

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,173 @@
...
+	if err := json.Unmarshal(upgrade, upgradeConfig); err != nil {
+		return nil, ErrInvalidData
+	}
+
+	return config, nil
+}
+
+// WriteChainConfig writes the chain config settings to the database.
+// The provided `upgradeConfig` (any JSON-marshalable type) will be stored alongside the chain config.
+func WriteChainConfig[T any](db ethdb.KeyValueWriter, hash common.Hash, config *params.ChainConfig, upgradeConfig T) error {
```
joshua-kim, line 143: Same comment as on `ReadChainConfig`

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+			read:   ReadOfflinePruning,
+			write:  WriteOfflinePruning,
+			delete: DeleteOfflinePruning,
+		},
+		{
+			name:   "populate_missing_tries",
+			read:   ReadPopulateMissingTries,
+			write:  WritePopulateMissingTries,
+			delete: DeletePopulateMissingTries,
+		},
```
joshua-kim: It looks like we're trying to avoid duplication, but I think the saying "duplication is cheaper than the wrong abstraction" is valid here:
1. I think especially in tests, it's preferable to duplicate than have less clear/obvious test code
2. I think the duplication here (set up a db and do read/write/delete calls) is trivial
3. Generally, if a test table involves setup functions as parameters it generally reads like a smell to me that the test is hard to reason about

Should we just split this across tests for each marker type?

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,173 @@
...
+
+// ReadChainConfig retrieves the consensus settings based on the given genesis hash.
+// The provided `upgradeConfig` (any JSON-unmarshalable type) will be populated if present on disk.
+func ReadChainConfig[T any](db ethdb.KeyValueReader, hash common.Hash, upgradeConfig *T) (*params.ChainConfig, error) {
+	config := rawdb.ReadChainConfig(db, hash)
+	if config == nil {
+		return nil, ErrEntryNotFound
+	}
+
+	upgrade, _ := db.Get(upgradeConfigKey(hash))
```
joshua-kim: Why do we ignore the error value here?

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var unix uint64
+	if err := rlp.DecodeBytes(data, &unix); err != nil {
+		return time.Time{}, fmt.Errorf("%w: %w", ErrInvalidData, err)
+	}
+
+	return time.Unix(int64(unix), 0), nil
+}
+
+// upgradeConfigKey = upgradeConfigPrefix + hash
```
joshua-kim: nit: Some of these in-line comments are documenting self-documenting code - could we remove these? Comments have a maintenance cost associated with them since they need to be kept in-sync w/ code, so if code is reasonably self-documenting we should generally avoid adding comments to them.

vms/evm/sync/customrawdb/accessors_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+				require.ErrorIs(t, err, tc.expectedErr)
+			} else {
+				require.NoError(t, err)
+			}
+			require.Equal(t, tc.want, tip)
+		})
+	}
+}
+
+func TestTimeMarkers_BadEncoding(t *testing.T) {
```
joshua-kim: Similar comment on this test w.r.t eating the duplicate cost of code to get a simpler test

vms/evm/sync/customrawdb/accessors_snapshot.go
```diff
@@ -0,0 +1,61 @@
...
+		return common.Hash{}, ErrInvalidData
+	}
+	return common.BytesToHash(data), nil
+}
+
+// WriteSnapshotBlockHash stores the root of the block whose state is contained in
+// the persisted snapshot.
+func WriteSnapshotBlockHash(db ethdb.KeyValueWriter, blockHash common.Hash) error {
+	if err := db.Put(snapshotBlockHashKey, blockHash[:]); err != nil {
+		log.Error("Failed to store snapshot block hash", "err", err)
```
joshua-kim: This behavior looks inconsistent w/ other apis - why do we log the error here? Would it make sense for us to just let the caller decide how to handle this?

vms/evm/sync/customrawdb/accessors_snapshot.go
```diff
@@ -0,0 +1,61 @@
...
+	return nil
+}
+
+// DeleteSnapshotBlockHash deletes the hash of the block whose state is contained in
+// the persisted snapshot. Since snapshots are not immutable, this  method can
+// be used during updates, so a crash or failure will mark the entire snapshot
+// invalid.
+func DeleteSnapshotBlockHash(db ethdb.KeyValueWriter) error {
+	if err := db.Delete(snapshotBlockHashKey); err != nil {
+		log.Error("Failed to remove snapshot block hash", "err", err)
```
joshua-kim: Same comment here

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,212 @@
...
+
+// NewSyncPerformedIterator returns an iterator over all block numbers the VM
+// has state synced to.
+func NewSyncPerformedIterator(db ethdb.Iteratee) ethdb.Iterator {
+	return rawdb.NewKeyLengthIterator(db.NewIterator(syncPerformedPrefix, nil), syncPerformedKeyLength)
+}
+
+// UnpackSyncPerformedKey returns the block number from keys the iterator returned
+// from NewSyncPerformedIterator.
+func UnpackSyncPerformedKey(key []byte) uint64 {
```
joshua-kim: super-nit: I'm not sure when we use `Unpack*` naming or why we started using it... but I feel like now would be a reasonable time to stop using it and use something like `Parse` or `Unmarshal` which we use more commonly (also consistent w/ the `ParseStateScheme` code)

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	require.NoError(t, db.Put(key, []byte("bar")))
+
+	require.NoError(t, ClearAllSyncSegments(db))
+
+	count := 0
+	it := db.NewIterator(syncSegmentsPrefix, nil)
+	defer it.Release()
+	for it.Next() {
+		count++
+	}
```
joshua-kim: 1. We're doing this iteration in a few spots to check the number of k/v pairs - would it make sense for us to use something like this so we can avoid re-implementing the length checks?
```
memDB := memorydb.New()
db := rawdb.NewDatabase(baseDB)
...
memDB.Len()
```

2.  Shouldn't we be checking the actual row instead of just the count?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	require.Equal(t, 1, count)
+}
+
+func TestWriteReadSyncRoot(t *testing.T) {
+	db := rawdb.NewMemoryDatabase()
+
+	// No root written yet
+	root, err := ReadSyncRoot(db)
+	require.ErrorIs(t, err, ErrEntryNotFound)
+	require.Equal(t, common.Hash{}, root)
```
joshua-kim: nit: `require.Zero` is a function to check for zero values

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	h2 := common.HexToHash("0x22")
+
+	require.NoError(t, WriteCodeToFetch(db, h1))
+	require.NoError(t, WriteCodeToFetch(db, h2))
+
+	// Insert a malformed key that should be ignored by the iterator (wrong length)
+	bad := append(append([]byte{}, CodeToFetchPrefix...), append(h1.Bytes(), 0x00)...)
+	require.NoError(t, db.Put(bad, []byte("x")))
+
+	seen := map[common.Hash]bool{}
```
joshua-kim: nit: the `bool` value in this map is actually redundant since the key is sufficient to know if it's in the map or not, so we'd want `map[T]struct{}`, which we have the alias `set.Set` for

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	root := common.HexToHash("0xabc")
+	cases := []struct {
+		name  string
+		clear func(db ethdb.KeyValueStore) error
+		iter  func(db ethdb.Iteratee) ethdb.Iterator
+	}{
+		{
+			name:  "segments_no_keys",
+			clear: func(db ethdb.KeyValueStore) error { return ClearSyncSegments(db, root) },
+			iter:  func(db ethdb.Iteratee) ethdb.Iterator { return NewSyncSegmentsIterator(db, root) },
```
joshua-kim: similar comment w.r.t duplication on this test

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	latest, err := GetLatestSyncPerformed(db)
+	require.NoError(t, err)
+	require.Equal(t, uint64(20), latest)
+}
+
+func TestGetLatestSyncPerformedEmpty(t *testing.T) {
+	db := rawdb.NewMemoryDatabase()
+	latest, err := GetLatestSyncPerformed(db)
+	require.NoError(t, err)
+	require.Equal(t, uint64(0), latest)
```
joshua-kim: I think `require.Zero` also works here as well (I thought our linter would catch this - so maybe I'm wrong?)

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+func TestSyncSegmentsByRootTable(t *testing.T) {
+	type entry struct {
+		root   common.Hash
+		starts []common.Hash
+	}
+	entries := []entry{
+		{root: common.HexToHash("0xaaa"), starts: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2")}},
+		{root: common.HexToHash("0xbbb"), starts: []common.Hash{common.HexToHash("0x3")}},
+	}
+	db := rawdb.NewMemoryDatabase()
```
joshua-kim: Why do we share this db instance across the sub-tests?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+
+func TestSyncSegmentsByRootTable(t *testing.T) {
+	type entry struct {
+		root   common.Hash
+		starts []common.Hash
+	}
+	entries := []entry{
+		{root: common.HexToHash("0xaaa"), starts: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2")}},
+		{root: common.HexToHash("0xbbb"), starts: []common.Hash{common.HexToHash("0x3")}},
+	}
```
joshua-kim: Why do we deviate from the table-test style we did in the other tests here?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+	for it.Next() {
+		count++
+	}
+	require.NoError(t, it.Error())
+	require.Equal(t, 1, count)
+}
+
+func TestClear_NoKeys(t *testing.T) {
+	root := common.HexToHash("0xabc")
+	cases := []struct {
```
joshua-kim: super nitty: we usually call this var for test-tables `tests`

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,507 @@
...
+		{
+			name:   "three_keep",
+			hashes: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2"), common.HexToHash("0x3")},
+			del:    nil,
+			want:   3,
+		},
+		{
+			name:   "three_delete_one",
+			hashes: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2"), common.HexToHash("0x3")},
+			del:    func() *common.Hash { h := common.HexToHash("0x2"); return &h }(),
```
joshua-kim: TIL you can use semicolons in Go to create these multi-line expressions...

vms/evm/sync/customrawdb/accessors_state_sync_test.go
```diff
@@ -0,0 +1,507 @@
...
+func TestChainConfigCases(t *testing.T) {
+	type upgrade struct {
+		X int `json:"x"`
+	}
+	cases := []struct {
+		name        string
+		cfg         *params.ChainConfig
+		up          any
+		mutate      func(db ethdb.KeyValueStore, h common.Hash)
+		expectedErr error
```
joshua-kim: nit: mixed usage of `expected*` and `want*`, could we defer to `want*`?

vms/evm/sync/customrawdb/accessors_state_sync_test.go
```diff
@@ -0,0 +1,507 @@
...
+			up:          struct{}{},
+			mutate:      nil,
+			expectedErr: ErrEntryNotFound,
+			wantUp:      nil,
+		},
+		{
+			name:        "invalid-upgrade-json",
+			cfg:         &params.ChainConfig{ChainID: big.NewInt(2)},
+			up:          upgrade{X: 1},
+			mutate:      func(db ethdb.KeyValueStore, h common.Hash) { _ = db.Put(upgradeConfigKey(h), []byte("{")) },
```
joshua-kim: This setup function makes the test harder to reason about - could we either just not test this or move it to its own test?

vms/evm/sync/customrawdb/errors.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package customrawdb
+
+import "errors"
+
+var (
+	// ErrEntryNotFound indicates the requested key/value was not present in the DB.
+	ErrEntryNotFound = errors.New("customrawdb: entry not found")
```
joshua-kim: 👍 I've seen this package name prefixing in stdlib and although we don't do it in our codestyle I think it seems like a nice suggestion since it adds meaningful context to wrapped errors

vms/evm/sync/customrawdb/schema.go
```diff
@@ -0,0 +1,49 @@
...
+	// FirewoodScheme is the scheme for the Firewood storage scheme.
+	FirewoodScheme = "firewood"
+
+	// === Snapshot metadata keys ===
```
joshua-kim: So on the file structure of this package:

There's a lot of sprawl - we have individual files, some of which holding little code at all (database.go, schema.go). Generally I feel like code should try to live close to where it's used, so these key vars living separate from the files they're used in feels weird to me.

I think Go does typically have dedicated files for stuff, but typically where there is a clear domain boundary (like peer.go, http.go, and database.go might be reasonable boundaries). It seems like all of the keys documented here though fall under the state sync grouping of data according to the comments, so would it make sense to just merge all of the files into one file? I feel like we could just go full mono-file and call it `db.go` or `customrawdb.go` or something. I feel like we could also drop the `accessors_` suffix.

vms/evm/sync/customrawdb/database.go
```diff
@@ -0,0 +1,25 @@
...
+
+package customrawdb
+
+import (
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+)
+
+// ParseStateScheme parses the state scheme from the provided string.
+func ParseStateScheme(provided string, disk ethdb.Database) (string, error) {
```
joshua-kim: nit: we call this `disk` which is inconsistent w/ our previous `db` naming

vms/evm/sync/customrawdb/schema.go
powerslider: I do agree with you, this file structure and naming was mostly inherited via the original rawdb package, but since we are doing custom implementations here it would be worthwhile to move away from it.

powerslider: ok, here is my newly proposed file structure:

- `markers.go`: All node/chain metadata markers in one place (offline pruning, populate tries, pruning disabled, acceptor tip, chain config, snapshot). These are simple KV flags/values with small helpers.
- `sync_progress.go`: State sync bookkeeping only (sync root, segments, storage tries, code-to-fetch, sync-performed) plus iterators and clear helpers. Keeps progress metadata separate from generic node markers.
- `db.go`: Package entry-point items with broad relevance (state scheme parsing, shared errors). Avoids scattering common definitions or tying them to a feature file.

It just did not make sense to me from a domain perspective to slap everything in one file, because we still have separate concerns. Plus I am not also a fan of way too long files so I think this is a well balanced structure that should be easily read by both people with advanced state sync knowledge and people with less context.

vms/evm/sync/customrawdb/markers.go
powerslider: no reason, I copied the code that way, so I'll fix it

vms/evm/sync/customrawdb/markers.go
powerslider: 1. Yes, I can omit the pointer from T.

2. We keep `T` unconstrained (`any`) because `encoding/json` can decode plain structs and built-ins without `json.Unmarshaler`. The comment clarifies that "unmarshalable" is about compatibility with `encoding/json`, not the interface.

vms/evm/sync/customrawdb/accessors_snapshot.go
powerslider: This is leftover from the `geth` style way. I'll remove it.

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,241 @@
...
+	return common.BytesToHash(data), nil
+}
+
+// WriteSnapshotBlockHash stores the root of the block whose state is contained in
+// the persisted snapshot.
+func WriteSnapshotBlockHash(db ethdb.KeyValueWriter, blockHash common.Hash) error {
+	if err := db.Put(snapshotBlockHashKey, blockHash[:]); err != nil {
+		return err
+	}
+	return nil
```
joshua-kim: the if check isn't needed here if we're directly returning the result
```suggestion
	return db.Put(snapshotBlockHashKey, blockHash[:])
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,241 @@
...
+
+// DeleteSnapshotBlockHash deletes the hash of the block whose state is contained in
+// the persisted snapshot. Since snapshots are not immutable, this  method can
+// be used during updates, so a crash or failure will mark the entire snapshot
+// invalid.
+func DeleteSnapshotBlockHash(db ethdb.KeyValueWriter) error {
+	if err := db.Delete(snapshotBlockHashKey); err != nil {
+		return err
+	}
+	return nil
```
joshua-kim: same nit here

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+	require.NoError(t, db.Put(bad, []byte("x")))
+
+	// Collect hashes from iterator and assert presence.
+	vals := mapIterator(t, NewCodeToFetchIterator(db), func(key []byte) common.Hash {
+		return common.BytesToHash(key[len(CodeToFetchPrefix):])
+	})
+	seen := set.Set[common.Hash]{}
+	seen.Add(vals...)
+	require.True(t, seen.Contains(h1))
+	require.True(t, seen.Contains(h2))
```
joshua-kim: Since this is a map alias:
```suggestion
	require.Contains(t, seen, h1)
	require.Contains(t, seen, h2)
```

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+	// Insert a malformed key that should be ignored by the iterator (wrong length)
+	bad := append(append([]byte{}, CodeToFetchPrefix...), append(h1.Bytes(), 0x00)...)
+	require.NoError(t, db.Put(bad, []byte("x")))
+
+	// Collect hashes from iterator and assert presence.
+	vals := mapIterator(t, NewCodeToFetchIterator(db), func(key []byte) common.Hash {
+		return common.BytesToHash(key[len(CodeToFetchPrefix):])
+	})
+	seen := set.Set[common.Hash]{}
+	seen.Add(vals...)
```
joshua-kim:
```suggestion
	seen := set.Of(vals...)
```

vms/evm/sync/customrawdb/sync_progress_test.go
joshua-kim: This looks like it wasn't resolved

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+	db := rawdb.NewMemoryDatabase()
+
+	require.NoError(t, WriteSyncPerformed(db, 10))
+	require.NoError(t, WriteSyncPerformed(db, 20))
+	require.NoError(t, WriteSyncPerformed(db, 15))
+
+	// Iterator yields all
+	vals := mapIterator(t, NewSyncPerformedIterator(db), ParseSyncPerformedKey)
+
+	require.ElementsMatch(t, []uint64{10, 20, 15}, vals)
```
joshua-kim: We know the ordering is guaranteed, shouldn't we assert that the values are 10 -> 15 -> 20?

If we do that - I think we can replace all usages of `ElementsMatch` with `Equals` in this package to get stricter constraints on our assertions.

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+		},
+		{
+			name:     "def",
+			root:     common.HexToHash("0xdef"),
+			accounts: []common.Hash{common.HexToHash("0x3")},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run("storage_"+tc.name, func(t *testing.T) {
```
joshua-kim: Why do we prefix the name here?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+		root   common.Hash
+		starts []common.Hash
+	}{
+		{
+			name:   "aaa",
+			root:   common.HexToHash("0xaaa"),
+			starts: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2")},
+		},
+		{
+			name: "bbb", root: common.HexToHash("0xbbb"),
```
joshua-kim: Can we fix whitespacing here?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func TestSyncStorageTriesByRootTable(t *testing.T) {
+	tests := []struct {
+		name     string
+		root     common.Hash
+		accounts []common.Hash
+	}{
+		{
+			name:     "abc",
```
joshua-kim: can we have better test names? It's hard to reason about these when debugging - maybe something like single hash, multiple hashes?

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func TestSyncSegmentsByRootTable(t *testing.T) {
+	tests := []struct {
+		name   string
+		root   common.Hash
+		starts []common.Hash
+	}{
+		{
+			name:   "aaa",
```
joshua-kim: Similar comment on test naming.

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,402 @@
...
+			starts: []common.Hash{common.HexToHash("0x1"), common.HexToHash("0x2")},
+		},
+		{
+			name: "bbb", root: common.HexToHash("0xbbb"),
+			starts: []common.Hash{common.HexToHash("0x3")},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run("segments_"+tc.name, func(t *testing.T) {
```
joshua-kim: Same comment w.r.t prefixing

vms/evm/sync/customrawdb/sync_progress_test.go
powerslider: I did remove the iteration duplication via the `mapIterator` function, or maybe I misunderstood your remark?

joshua-kim: I was referencing https://github.com/ava-labs/avalanchego/pull/4387#discussion_r2430498403 to make these separate tests to avoid the functions in the tables

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,414 @@
...
+	start3 := common.HexToHash("0x03")
+
+	require.NoError(t, WriteSyncSegment(db, rootA, start1))
+	require.NoError(t, WriteSyncSegment(db, rootA, start2))
+	require.NoError(t, WriteSyncSegment(db, rootB, start3))
+
+	// Iterate only over rootA and assert exact keys.
+	keys := mapIterator(t, NewSyncSegmentsIterator(db, rootA), common.CopyBytes)
+	expectedA := [][]byte{buildSegmentKey(rootA, start1), buildSegmentKey(rootA, start2)}
+	require.ElementsMatch(t, keys, expectedA)
```
joshua-kim: It looks like we're still using `ElementsMatch`, see related https://github.com/ava-labs/avalanchego/pull/4387#discussion_r2434091583

vms/evm/sync/customrawdb/sync_progress_test.go
powerslider: ooo, ok got it now, I understood something totally different

vms/evm/sync/customrawdb/db.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"errors"
+
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+)
+
+var (
+	// ErrEntryNotFound indicates the requested key/value was not present in the DB.
+	ErrEntryNotFound = errors.New("entry not found")
```
DracoLi: we have a general `database.ErrNotFound` for db not found errors. 
I'm wondering if we should just use that instead of creating another custom not found error. wdyt?

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	if !ok {
+		return time.Time{}, ErrEntryNotFound
+	}
+
+	data, err := db.Get(key)
+	if err != nil {
+		return time.Time{}, err
+	}
+	if len(data) == 0 {
```
DracoLi: can this ever happen normally? i'm wondering f `errInvalidData` is a better error for this.

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// WriteSnapshotBlockHash stores the root of the block whose state is contained in
+// the persisted snapshot.
+func WriteSnapshotBlockHash(db ethdb.KeyValueWriter, blockHash common.Hash) error {
+	return db.Put(snapshotBlockHashKey, blockHash[:])
+}
+
+// DeleteSnapshotBlockHash deletes the hash of the block whose state is contained in
+// the persisted snapshot. Since snapshots are not immutable, this  method can
```
DracoLi:
```suggestion
// the persisted snapshot. Since snapshots are not immutable, this method can
```

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,234 @@
...
+	keyBytes = keyBytes[len(syncStorageTriesPrefix):] // skip prefix
+	root := common.BytesToHash(keyBytes[:common.HashLength])
+	account := common.BytesToHash(keyBytes[common.HashLength:])
+	return root, account
+}
+
+// WriteSyncPerformed logs an entry in `db` indicating the VM state synced to `blockNumber`.
+func WriteSyncPerformed(db ethdb.KeyValueWriter, blockNumber uint64) error {
+	syncPerformedPrefixLen := len(syncPerformedPrefix)
+	bytes := make([]byte, syncPerformedPrefixLen+wrappers.LongLen)
```
DracoLi: can `syncPerformedKeyLength` be used here?

vms/evm/sync/customrawdb/markers.go
powerslider: Good catch - with our writers, the value is always RLP-encoded uint64 (non-empty). An empty value implies corruption or external misuse, not a normal state, so I'll change the error.

vms/evm/sync/customrawdb/db.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+)
+
+var (
+	// errStateSchemeConflict indicates the provided state scheme conflicts with what is on disk.
+	errStateSchemeConflict = errors.New("state scheme conflict")
+	// FirewoodScheme is the scheme for the Firewood storage scheme.
+	FirewoodScheme = "firewood"
+)
```
StephenButtolph:
```suggestion
// FirewoodScheme is the scheme for the Firewood storage scheme.
const FirewoodScheme = "firewood"

// errStateSchemeConflict indicates the provided state scheme conflicts with
// what is on disk.
var errStateSchemeConflict = errors.New("state scheme conflict")
```

vms/evm/sync/customrawdb/db_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Provided Firewood on empty disk -> allowed.
+	scheme, err := ParseStateScheme(FirewoodScheme, db)
+	require.NoError(t, err)
+	require.Equal(t, FirewoodScheme, scheme)
+
+	// Simulate disk has non-empty path scheme by writing persistent state id.
+	rawdb.WritePersistentStateID(db, 1)
+	scheme2, _ := ParseStateScheme(FirewoodScheme, db)
+	require.Empty(t, scheme2)
```
StephenButtolph:
```suggestion
	scheme, err = ParseStateScheme(FirewoodScheme, db)
	require.ErrorIs(t, err, errStateSchemeConflict)
	require.Empty(t, scheme)
```

vms/evm/sync/customrawdb/db_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Pass-through to rawdb for non-Firewood using a fresh empty DB.
+	db2 := rawdb.NewMemoryDatabase()
+	scheme, err = ParseStateScheme("hash", db2)
+	require.NoError(t, err)
+	require.Equal(t, "hash", scheme)
```
StephenButtolph:
```suggestion
	scheme, err = ParseStateScheme(rawdb.HashScheme, db2)
	require.NoError(t, err)
	require.Equal(t, rawdb.HashScheme, scheme)
```

vms/evm/sync/customrawdb/db_test.go
```diff
@@ -0,0 +1,31 @@
...
+package customrawdb
+
+import (
+	"testing"
+
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseStateScheme(t *testing.T) {
```
StephenButtolph, line 13: No action required, but I'd probably slightly prefer this as a test vector... Admittedly this might be more succinct as is.

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	// snapshotBlockHashKey tracks the block hash of the last snapshot.
+	snapshotBlockHashKey = []byte("SnapshotBlockHash")
+)
+
+// WriteOfflinePruning writes a time marker of the last attempt to run offline pruning.
+// The marker is written when offline pruning completes and is deleted when the node
+// is started successfully with offline pruning disabled. This ensures users must
+// disable offline pruning and start their node successfully between runs of offline
+// pruning.
+func WriteOfflinePruning(db ethdb.KeyValueStore, ts time.Time) error {
```
StephenButtolph: Everywhere in this file that we have `ethdb.KeyValueStore`, we should either be using `ethdb. KeyValueReader` or `ethdb.KeyValueWriter`
```suggestion
func WriteOfflinePruning(db ethdb.KeyValueWriter, ts time.Time) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return writeTimeMarker(db, offlinePruningKey, ts)
+}
+
+// ReadOfflinePruning reads the most recent timestamp of an attempt to run offline
+// pruning if present.
+func ReadOfflinePruning(db ethdb.KeyValueStore) (time.Time, error) {
```
StephenButtolph:
```suggestion
func ReadOfflinePruning(db ethdb.KeyValueReader) (time.Time, error) {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return readTimeMarker(db, offlinePruningKey)
+}
+
+// DeleteOfflinePruning deletes any marker of the last attempt to run offline pruning.
+func DeleteOfflinePruning(db ethdb.KeyValueStore) error {
```
StephenButtolph:
```suggestion
func DeleteOfflinePruning(db ethdb.KeyValueWriter) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return db.Delete(offlinePruningKey)
+}
+
+// WritePopulateMissingTries writes a marker for the current attempt to populate
+// missing tries.
+func WritePopulateMissingTries(db ethdb.KeyValueStore, ts time.Time) error {
```
StephenButtolph:
```suggestion
func WritePopulateMissingTries(db ethdb.KeyValueWriter, ts time.Time) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return writeTimeMarker(db, populateMissingTriesKey, ts)
+}
+
+// ReadPopulateMissingTries reads the most recent timestamp of an attempt to
+// re-populate missing trie nodes.
+func ReadPopulateMissingTries(db ethdb.KeyValueStore) (time.Time, error) {
```
StephenButtolph:
```suggestion
func ReadPopulateMissingTries(db ethdb.KeyValueReader) (time.Time, error) {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return readTimeMarker(db, populateMissingTriesKey)
+}
+
+// DeletePopulateMissingTries deletes any marker of the last attempt to
+// re-populate missing trie nodes.
+func DeletePopulateMissingTries(db ethdb.KeyValueStore) error {
```
StephenButtolph:
```suggestion
func DeletePopulateMissingTries(db ethdb.KeyValueWriter) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return db.Delete(populateMissingTriesKey)
+}
+
+// WritePruningDisabled writes a marker to track whether the node has ever run
+// with pruning disabled.
+func WritePruningDisabled(db ethdb.KeyValueStore) error {
```
StephenButtolph:
```suggestion
func WritePruningDisabled(db ethdb.KeyValueWriter) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	return db.Put(pruningDisabledKey, nil)
+}
+
+// HasPruningDisabled returns true if there is a marker present indicating that
+// the node has run with pruning disabled at some point.
+func HasPruningDisabled(db ethdb.KeyValueStore) (bool, error) {
```
StephenButtolph:
```suggestion
func HasPruningDisabled(db ethdb.KeyValueReader) (bool, error) {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+
+// DeleteSnapshotBlockHash deletes the hash of the block whose state is contained in
+// the persisted snapshot. Since snapshots are not immutable, this method can
+// be used during updates, so a crash or failure will mark the entire snapshot
+// invalid.
+func DeleteSnapshotBlockHash(db ethdb.KeyValueWriter) error {
+	return db.Delete(snapshotBlockHashKey)
+}
+
+func writeTimeMarker(db ethdb.KeyValueStore, key []byte, ts time.Time) error {
```
StephenButtolph:
```suggestion
func writeTimeMarker(db ethdb.KeyValueWriter, key []byte, ts time.Time) error {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	data, err := rlp.EncodeToBytes(uint64(ts.Unix()))
+	if err != nil {
+		return err
+	}
+	return db.Put(key, data)
+}
+
+func readTimeMarker(db ethdb.KeyValueStore, key []byte) (time.Time, error) {
```
StephenButtolph:
```suggestion
func readTimeMarker(db ethdb.KeyValueReader, key []byte) (time.Time, error) {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	upgrade, err := db.Get(upgradeConfigKey(hash))
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errFailedToGetUpgradeConfig, err)
+	}
+
+	if len(upgrade) == 0 {
+		return config, nil
+	}
+
+	if err := json.Unmarshal(upgrade, &upgradeConfig); err != nil {
```
StephenButtolph: Shouldn't we be enforcing that `upgradeConfig` is a pointer? As is, this function allows providing a value, but in that case the caller's upgradeConfig is not read properly.
```suggestion
func ReadChainConfig[T any](db ethdb.KeyValueReader, hash common.Hash, upgradeConfig *T) (*params.ChainConfig, error) {
	config := rawdb.ReadChainConfig(db, hash)
	if config == nil {
		return nil, database.ErrNotFound
	}

	upgrade, err := db.Get(upgradeConfigKey(hash))
	if err != nil {
		return nil, fmt.Errorf("%w: %w", errFailedToGetUpgradeConfig, err)
	}

	if len(upgrade) == 0 {
		return config, nil
	}

	if err := json.Unmarshal(upgrade, upgradeConfig); err != nil {
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return common.Hash{}, err
+	}
+	if !ok {
+		return common.Hash{}, database.ErrNotFound
+	}
+	h, err := db.Get(acceptorTipKey)
+	if err != nil {
+		return common.Hash{}, err
+	}
```
StephenButtolph, line 110: Are any of the methods where we call `Has` followed by `Get` on a performance critical path?

I had originally written a comment about checking the error on `Get` rather than calling `Has`... But is appears that libevm doesn't return sentinel errors to inspect if the fetching failed due to a disk error or just because the entry isn't present...

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+// has state synced to.
+func NewSyncPerformedIterator(db ethdb.Iteratee) ethdb.Iterator {
+	return rawdb.NewKeyLengthIterator(db.NewIterator(syncPerformedPrefix, nil), syncPerformedKeyLength)
+}
+
+// ParseSyncPerformedKey returns the block number from keys the iterator returned
+// from NewSyncPerformedIterator.
+func ParseSyncPerformedKey(key []byte) uint64 {
+	return binary.BigEndian.Uint64(key[len(syncPerformedPrefix):])
+}
```
StephenButtolph: Do we need to export this?

StephenButtolph: Should we document that this function panics if `key` doesn't have sufficient length?

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+
+// ParseSyncStorageTrieKey returns the root and account for a storage trie
+// key returned from NewSyncStorageTriesIterator.
+func ParseSyncStorageTrieKey(keyBytes []byte) (common.Hash, common.Hash) {
+	keyBytes = keyBytes[len(syncStorageTriesPrefix):] // skip prefix
+	root := common.BytesToHash(keyBytes[:common.HashLength])
+	account := common.BytesToHash(keyBytes[common.HashLength:])
+	return root, account
+}
```
StephenButtolph, line 173: Can we document that this panics if the length is not sufficient?

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+func clearPrefix(db ethdb.KeyValueStore, prefix []byte, keyLen int) error {
+	it := db.NewIterator(prefix, nil)
+	defer it.Release()
+
+	batch := db.NewBatch()
+	for it.Next() {
+		key := common.CopyBytes(it.Key())
+		if len(key) != keyLen {
+			continue
+		}
```
StephenButtolph: No point copying `key`s if we aren't going to hold a reference to them during the next iteration.
```suggestion
		key := it.Key()
		if len(key) != keyLen {
			continue
		}
		key = common.CopyBytes(key)
```

vms/evm/sync/customrawdb/markers_test.go
```diff
@@ -0,0 +1,244 @@
...
+func TestNewAccountSnapshotsIterator(t *testing.T) {
+	db := rawdb.NewMemoryDatabase()
+
+	// Keys that match and don't match the iterator length filter.
+	a1 := common.HexToHash("0x01")
+	a2 := common.HexToHash("0x02")
+	key1 := append(append([]byte{}, rawdb.SnapshotAccountPrefix...), a1.Bytes()...)
+	key2 := append(append([]byte{}, rawdb.SnapshotAccountPrefix...), a2.Bytes()...)
+	// Non-matching: extra byte appended.
+	bad := append(append([]byte{}, rawdb.SnapshotAccountPrefix...), append(a1.Bytes(), 0x00)...)
```
StephenButtolph:
```suggestion
	key1 := slices.Concat(rawdb.SnapshotAccountPrefix, a1.Bytes())
	key2 := slices.Concat(rawdb.SnapshotAccountPrefix, a2.Bytes())
	// Non-matching: extra byte appended.
	bad := slices.Concat(key1, []byte{0x00})
```

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,403 @@
...
+	db := rawdb.NewMemoryDatabase()
+
+	h1 := common.HexToHash("0x11")
+	h2 := common.HexToHash("0x22")
+
+	require.NoError(t, WriteCodeToFetch(db, h1))
+	require.NoError(t, WriteCodeToFetch(db, h2))
+
+	// Insert a malformed key that should be ignored by the iterator (wrong length)
+	bad := append(append([]byte{}, CodeToFetchPrefix...), append(h1.Bytes(), 0x00)...)
```
StephenButtolph:
```suggestion
	bad := slices.Concat(CodeToFetchPrefix, h1.Bytes(), []byte{0x00})
```

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,403 @@
...
+	root := common.HexToHash("0xabc")
+	// Keys that should be cleared
+	require.NoError(t, WriteSyncStorageTrie(db, root, common.HexToHash("0x01")))
+	require.NoError(t, WriteSyncStorageTrie(db, root, common.HexToHash("0x02")))
+	// Key that should not be cleared due to wrong length.
+	bad := make([]byte, 0, len(syncStorageTriesPrefix)+2*common.HashLength+1)
+	bad = append(bad, syncStorageTriesPrefix...)
+	bad = append(bad, root.Bytes()...)
+	bad = append(bad, common.HexToHash("0xff").Bytes()...)
+	bad = append(bad, byte(0x00))
```
StephenButtolph:
```suggestion
	bad := slices.Concat(
		syncStorageTriesPrefix,
		root.Bytes(),
		common.HexToHash("0xff").Bytes(),
		[]byte{0x00},
	)
```

vms/evm/sync/customrawdb/sync_progress_test.go
```diff
@@ -0,0 +1,403 @@
...
+	return buildKey(syncStorageTriesPrefix, root, account, syncStorageTriesKeyLength)
+}
+
+func buildKey(prefix []byte, h1, h2 common.Hash, keyLen int) []byte {
+	b := make([]byte, keyLen)
+	copy(b, prefix)
+	copy(b[len(prefix):], h1[:])
+	copy(b[len(prefix)+common.HashLength:], h2[:])
+	return b
+}
```
StephenButtolph, line 396:
```suggestion
func buildSegmentKey(root, start common.Hash) []byte {
	return slices.Concat(syncSegmentsPrefix, root[:], start[:])
}

func buildStorageTrieKey(root, account common.Hash) []byte {
	return slices.Concat(syncStorageTriesPrefix, root[:], account[:])
}
```

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+func NewSyncStorageTriesIterator(db ethdb.Iteratee, seek []byte) ethdb.Iterator {
+	return rawdb.NewKeyLengthIterator(db.NewIterator(syncStorageTriesPrefix, seek), syncStorageTriesKeyLength)
+}
+
+// WriteSyncStorageTrie adds a storage trie for account (with the given root) to be synced.
+func WriteSyncStorageTrie(db ethdb.KeyValueWriter, root common.Hash, account common.Hash) error {
+	bytes := make([]byte, 0, syncStorageTriesKeyLength)
+	bytes = append(bytes, syncStorageTriesPrefix...)
+	bytes = append(bytes, root[:]...)
+	bytes = append(bytes, account[:]...)
```
StephenButtolph: Everywhere else uses copy, feels a bit weird to me that we use append here.

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+	root := common.BytesToHash(keyBytes[:common.HashLength])
+	account := common.BytesToHash(keyBytes[common.HashLength:])
+	return root, account
+}
+
+// WriteSyncPerformed logs an entry in `db` indicating the VM state synced to `blockNumber`.
+func WriteSyncPerformed(db ethdb.KeyValueWriter, blockNumber uint64) error {
+	syncPerformedPrefixLen := len(syncPerformedPrefix)
+	bytes := make([]byte, syncPerformedKeyLength)
+	copy(bytes[:syncPerformedPrefixLen], syncPerformedPrefix)
```
StephenButtolph: optional:
```suggestion
	copy(bytes, syncPerformedPrefix)
```
we don't need to slice this.

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+	binary.BigEndian.PutUint64(bytes[syncPerformedPrefixLen:], blockNumber)
+	return db.Put(bytes, []byte{0x01})
```
StephenButtolph: Why do we write a `[]byte{0x01}` here?
```suggestion
	return db.Put(bytes, nil)
```

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// WriteSyncSegment adds a trie segment for root at the given start position.
+func WriteSyncSegment(db ethdb.KeyValueWriter, root common.Hash, start common.Hash) error {
+	// packs root and account into a key for storage in db.
+	bytes := make([]byte, syncSegmentsKeyLength)
+	copy(bytes, syncSegmentsPrefix)
+	copy(bytes[len(syncSegmentsPrefix):], root[:])
+	copy(bytes[len(syncSegmentsPrefix)+common.HashLength:], start.Bytes())
+	return db.Put(bytes, []byte{0x01})
```
StephenButtolph: Why do we write a `[]byte{0x01}` here?
```suggestion
	return db.Put(bytes, nil)
```

vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -0,0 +1,235 @@
...
+	return rawdb.NewKeyLengthIterator(db.NewIterator(syncStorageTriesPrefix, seek), syncStorageTriesKeyLength)
+}
+
+// WriteSyncStorageTrie adds a storage trie for account (with the given root) to be synced.
+func WriteSyncStorageTrie(db ethdb.KeyValueWriter, root common.Hash, account common.Hash) error {
+	bytes := make([]byte, 0, syncStorageTriesKeyLength)
+	bytes = append(bytes, syncStorageTriesPrefix...)
+	bytes = append(bytes, root[:]...)
+	bytes = append(bytes, account[:]...)
+	return db.Put(bytes, []byte{0x01})
```
StephenButtolph: Why do we write a `[]byte{0x01}` here?
```suggestion
	return db.Put(bytes, nil)
```

vms/evm/sync/customrawdb/markers.go
powerslider: Good catch, I actually meant for this to be a pointer, but I was doing some experiments and forgot to return it back as a pointer.

vms/evm/sync/customrawdb/markers.go
powerslider: `Has -> Get` patterns appear in `ReadSyncRoot`, `ReadAcceptorTip`, `ReadSnapshotBlockHash`, and `readTimeMarker`. None of these are on hot paths - they’re used at startup, during maintenance/snapshot workflows, or as sync metadata checks - not in per-block or tight loops. So the extra lookup isn’t performance-sensitive.

vms/evm/sync/customrawdb/sync_progress.go
powerslider: Based on the current internal usage no, so I'll unexport it for now.

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+
+	data, err := json.Marshal(upgradeConfig)
+	if err != nil {
+		return fmt.Errorf("%w: %w", errFailedToMarshalUpgradeConfig, err)
+	}
+	if err := db.Put(upgradeConfigKey(hash), data); err != nil {
+		return err
+	}
+	return nil
```
StephenButtolph:
```suggestion
	return db.Put(upgradeConfigKey(hash), data)
```

vms/evm/sync/customrawdb/markers.go
```diff
@@ -0,0 +1,238 @@
...
+
+	data, err := db.Get(key)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	if len(data) == 0 {
+		return time.Time{}, fmt.Errorf("%w: empty value", errInvalidData)
+	}
+
```
StephenButtolph: I think we will already error if the data is empty
```suggestion
```

StephenButtolph (approved):

PR #4389, opened 2025-10-03
refactor: move most ledger transactions from hash signing to full-bytes signingmove hash signing decision out of wallet into ledger keychain

felipemadero:
## Why this should be merged

Currently, the wallet layer needs to know when to use `SignHash` vs `Sign` for different transaction types, creating tight coupling between wallet logic and signing implementation details. This PR moves that decision-making into the ledger keychain itself, simplifying the wallet interface and centralizing signing logic where it belongs.

Additionally, this PR incorporates the refactoring changes from #4339 (ledger-keychain-move) to resolve circular dependency issues between the keychain and secp256k1fx packages.

### Ledger Signing Improvements

The ledger keychain's `Sign` method now internally determines when to use `SignHash` based on transaction type:

- **For TransferSubnetOwnershipTx only**: Uses `SignHash` (required by Ledger hardware)
- **For all other P/X/C-Chain transactions**: Uses the regular `Sign` method

### Transactions that moved from SignHash → Sign

After comprehensive testing with Ledger hardware, these transactions were verified to work with full-bytes signing and **no longer require hash signing**:

**P-Chain:**
- RemoveSubnetValidatorTx
- TransformSubnetTx
- AddPermissionlessValidatorTx
- AddPermissionlessDelegatorTx
- ConvertSubnetToL1Tx
- RegisterL1ValidatorTx
- SetL1ValidatorWeightTx
- IncreaseL1ValidatorBalanceTx
- DisableL1ValidatorTx

**C-Chain:**
- ImportTx
- ExportTx

### Architecture Changes

- Wallet layer no longer needs conditional logic to choose between signing methods - it always calls `Sign`
- Ledger-specific code moved to `utils/crypto/keychain/ledger` package to fix circular dependencies

## How this was tested

- Unit tests for all transaction types
- CLI flows to deploy an L1 with Ledger hardware
- Comprehensive Ledger hardware testing for all P/X/C-Chain transaction types to verify which require SignHash vs Sign
- Example programs for subnet/chain creation and P<->C transfers


felipemadero:

PR #4393, opened 2025-10-03
Simplify waitForProposerWindow helper

StephenButtolph:
## Why this should be merged

There are some places in our unit tests where we specify a p-chain height for block building that may not be accurate.

This PR removes the requirement for manually specifying the value in the helper, and ensures the value is correctly generated in the helper.

## How this works

fetches the preferred block from the VM and waits until the local node can build a block on the preferred block rather than taking in a block that is assumed to be preferred.

## How this was tested

CI

no.

### DIFF
--- a/vms/proposervm/batched_vm_test.go
+++ b/vms/proposervm/batched_vm_test.go
@@ -198,7 +198,7 @@ func TestGetAncestorsPostForkOnly(t *testing.T) {
 	// prepare build of next block
 	require.NoError(builtBlk1.Verify(context.Background()))
 	require.NoError(proRemoteVM.SetPreference(context.Background(), builtBlk1.ID()))
-	require.NoError(waitForProposerWindow(proRemoteVM, builtBlk1, 0))
+	require.NoError(proRemoteVM.waitForProposerWindow())
 
 	coreBlk2 := snowmantest.BuildChild(coreBlk1)
 	coreVM.BuildBlockF = func(context.Context) (snowman.Block, error) {
@@ -210,7 +210,7 @@ func TestGetAncestorsPostForkOnly(t *testing.T) {
 	// prepare build of next block
 	require.NoError(builtBlk2.Verify(context.Background()))
 	require.NoError(proRemoteVM.SetPreference(context.Background(), builtBlk2.ID()))
-	require.NoError(waitForProposerWindow(proRemoteVM, builtBlk2, 0))
+	require.NoError(proRemoteVM.waitForProposerWindow())
 
 	coreBlk3 := snowmantest.BuildChild(coreBlk2)
 	coreVM.BuildBlockF = func(context.Context) (snowman.Block, error) {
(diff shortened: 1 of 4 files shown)


vms/proposervm/batched_vm_test.go
```diff
@@ -198,7 +198,7 @@ func TestGetAncestorsPostForkOnly(t *testing.T) {
 	// prepare build of next block
 	require.NoError(builtBlk1.Verify(context.Background()))
 	require.NoError(proRemoteVM.SetPreference(context.Background(), builtBlk1.ID()))
-	require.NoError(waitForProposerWindow(proRemoteVM, builtBlk1, 0))
```
StephenButtolph, line 201: It is not guaranteed for the pchain height to be 0 here.

vms/proposervm/batched_vm_test.go
```diff
@@ -210,7 +210,7 @@ func TestGetAncestorsPostForkOnly(t *testing.T) {
 	// prepare build of next block
 	require.NoError(builtBlk2.Verify(context.Background()))
 	require.NoError(proRemoteVM.SetPreference(context.Background(), builtBlk2.ID()))
-	require.NoError(waitForProposerWindow(proRemoteVM, builtBlk2, 0))
```
StephenButtolph, line 213: It is not guaranteed for the pchain height to be 0 here.

vms/proposervm/batched_vm_test.go
```diff
@@ -689,7 +689,7 @@ func TestBatchedParseBlockPostForkOnly(t *testing.T) {
 	// prepare build of next block
 	require.NoError(builtBlk1.Verify(context.Background()))
 	require.NoError(proRemoteVM.SetPreference(context.Background(), builtBlk1.ID()))
-	require.NoError(waitForProposerWindow(proRemoteVM, builtBlk1, 0))
```
StephenButtolph, line 692: It is not guaranteed for the pchain height to be 0 here.

vms/proposervm/vm_test.go
```diff
@@ -217,13 +217,22 @@ func initTestProposerVM(
...
+	if err != nil {
+		return fmt.Errorf("couldn't get P-Chain height from tip: %w", err)
+	}
+
 	var (
-		ctx              = context.Background()
-		childBlockHeight = chainTip.Height() + 1
-		parentTimestamp  = chainTip.Timestamp()
+		childBlockHeight = preferred.Height() + 1
+		parentTimestamp  = preferred.Timestamp()
```
yacovm, line 234: not related to this PR but since we're simplifying the function, shouldn't we just do:

```
		if now.After(delayUntil) {
			return nil
		}
```

instead of:

```
		if !now.Before(delayUntil) {
			return nil
		}
```

?

StephenButtolph: if `now == delayUntil` we should return

michaelkaplan13 (approved):

Issue #4388, opened 2025-10-03
Add API endpoint GetAllValidatorsAt()


geoff-vball:
**Context and scope**
In order to support relayer/aggregator use cases post-granite, we want to be able to fetch all non-empty warp validator sets from the API, in a similar way we already use `GetValidatorsAt()`.

**Discussion and alternatives**

**Open questions**

PR #4394, opened 2025-10-03
Add GetAllValidatorsAt endpoint

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/avalanchego/issues/4388

## How this works
Adds an API endpoint to the platform vm API to return the canonical validator sets of all chains at a given height.

## How this was tested
Unit tests and E2E tests

### DIFF
--- a/vms/platformvm/service.go
+++ b/vms/platformvm/service.go
@@ -1782,6 +1782,133 @@ func (s *Service) GetTimestamp(_ *http.Request, _ *struct{}, reply *GetTimestamp
 	return nil
 }
 
+// GetAllValidatorsAtArgs are the arguments for GetAllValidatorsAt
+type GetAllValidatorsAtArgs struct {
+	Height platformapi.Height `json:"height"`
+}
+
+// GetAllValidatorsAtReply is the response from GetAllValidatorsAt
+type GetAllValidatorsAtReply struct {
+	Validators map[ids.ID]validators.WarpSet `json:"validators"`
+}
+
+type jsonWarpSet struct {
+	Validators  []*jsonWarpValidatorOutput `json:"validators"`
+	TotalWeight avajson.Uint64             `json:"totalWeight"`
+}
+
+type jsonWarpValidatorOutput struct {
+	PublicKey *string        `json:"publicKey"`
+	Weight    avajson.Uint64 `json:"weight"`
+	NodeIDs   []ids.NodeID   `json:"nodeIDs"`
+}
+
+// GetValidatorsAt returns the weights of the validator set of a provided subnet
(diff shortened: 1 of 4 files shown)


vms/platformvm/service.md
```diff
@@ -1846,6 +1846,79 @@ curl -X POST --data '{
...
+Get the validators and their weights of all Subnets and the Primary Network at a given P-Chain height.
+
+**Signature:**
+
+```
+platform.getAllValidatorsAt(
+    {
+        height: [int|string],
+    }
+)
```
michaelkaplan13, line 1860: Can/should we add the response signature?

geoff-vball: Added !

StephenButtolph: I'm not really sure what the response signature would look like... tbh I feel like the example request + response is sufficiently descriptive.

geoff-vball: Oh damn, I thought I pushed the change I had made, but github was having a moment yesterday morning. I had the same difficulty in knowing how to describe it, but here's what Claude gave me if you want to add it or something similar

```) -> {
platform.getAllValidatorsAt(
    {
        height: [int|string]
    }
) -> {
    validatorSets: {
        [subnetID]: {
            validators: [
                {
                    publicKey: string,
                    weight: string,
                    nodeIDs: []string
                }
            ],
            totalWeight: string
        }
    }
}

File: vms/platformvm/service.go [hunk 2]
```diff
@@ -1867,13 +1914,9 @@ func (s *Service) GetValidatorsAt(r *http.Request, args *GetValidatorsAtArgs, re
...
-	var err error
-	height := uint64(args.Height)
-	if args.Height.IsProposed() {
-		height, err = s.vm.GetMinimumHeight(ctx)
-		if err != nil {
-			return fmt.Errorf("failed to get proposed height: %w", err)
-		}
+	height, err := s.getQueryHeight(ctx, args.Height)
+	if err != nil {
+		return fmt.Errorf("failed to get query height: %w", err)
```
iansuvak: with the current implementation that is unlikely to change without another sentinel value like "proposed" this will only ever be an error when fetching the proposed height still but I agree that when abstracting out into a `getQueryHeight` function it makes sense to be more general with the error message.

StephenButtolph: After re-reading it, I think I misread your comment.

But I moved the error message into the helper, so that we would have fine-grained control of the error message based on the query

iansuvak: Sorry for the not very readable comment. I do think that this is an improvement =)

PR #4402, opened 2025-10-07
Cache subnetIDs in cachedState

michaelkaplan13:
## Why this should be merged
Optimizes look ups of the subnet ID for a given blockchain ID, particularly for chains with `validatorState` instances that would otherwise require going over grpc for the look up each time (i.e. `subnet-evm` instances). 

See discussion [here](https://github.com/ava-labs/coreth/pull/1324#discussion_r2410937838).

## How this works
Uses an LRU cache for subnet IDs for given blockchain IDs.

## How this was tested
Unit tests 

No

### DIFF
--- a/snow/validators/state_test.go
+++ b/snow/validators/state_test.go
@@ -96,7 +96,7 @@ func TestCachedState_GetSubnetID(t *testing.T) {
 			uncached.GetSubnetIDF = nil
 		}
 
-		got, err := cached.GetSubnetID(context.Background(), test.blockchainID)
+		got, err := cached.GetSubnetID(t.Context(), test.blockchainID)
 		require.ErrorIs(err, test.wantErr)
 		require.Equal(test.want, got)
 		require.Equal(test.expectCached, !cacheMiss)


michaelkaplan13:
@StephenButtolph called out that this does not actually improve the worst case verification times/cost for flows dependent on `GetSubnetID`, since bogus blockchain ID parameters could be provided, which will never be cached.

It's strictly a best case performance optimization, but should not be relied upon for gas price calculations.

snow/validators/state.go
```diff
@@ -189,10 +192,15 @@ type cachedState struct {
...
+	// Caches the subnet ID for given blockchain IDs.
+	// Key: blockchain ID
+	// Value: subnet ID
+	subnetIDsCache cache.Cacher[ids.ID, ids.ID]
+
 	// Caches validators for all subnets at various heights.
 	// Key: height
 	// Value: mapping subnet ID -> validator set
-	cache cache.Cacher[uint64, map[ids.ID]WarpSet]
+	validatorsSetsCache cache.Cacher[uint64, map[ids.ID]WarpSet]
```
StephenButtolph: The double plural reads a bit weirdly. It is multiple validator sets
```suggestion
	validatorSetsCache cache.Cacher[uint64, map[ids.ID]WarpSet]
```

michaelkaplan13: 👍 updated

PR #4405, opened 2025-10-08
chore(sync/types): add sync/types package that contains the main Syncer interface

### LINKED ISSUE #4404: Migrate syncer interface types to `avalanchego` from `coreth`
## Summary 

Migrate the `Syncer` interface to `vms/evm/sync` which is the core contract used by all syncer implementation that will follow up in later PRs.

powerslider:
## Why this should be merged

Check #4404 

## How this works

- Migrate `Syncer` interface type that will be the main contract for all syncer implementations migrated in follow up PRs.

## How this was tested

existing UT

no

resolves #4404

### DIFF
--- a/vms/evm/sync/types/types.go
+++ b/vms/evm/sync/types/types.go
@@ -1,0 +1,22 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package types
+
+import "context"
+
+// Syncer is the common interface for all sync operations.
+// This provides a unified interface for atomic state sync and state trie sync.
+type Syncer interface {
+	// Sync completes the full sync operation, returning any errors encountered.
+	// The sync will respect context cancellation.
+	Sync(ctx context.Context) error
+
+	// Name returns a human-readable name for this syncer implementation.
+	Name() string
+
+	// ID returns a stable, machine-oriented identifier (e.g., "state_block_sync", "state_code_sync",
+	// "state_evm_state_sync", "state_atomic_sync"). Implementations should ensure this is unique and
+	// stable across renames for logging/metrics/deduplication.
+	ID() string
+}


vms/evm/sync/types/types.go
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
DracoLi, line 1: `vms/evm/sync/types/types.go`

This is in the correct folder? what other types will be in `sync/types`?

powerslider: Here is the original [code](https://github.com/ava-labs/coreth/blob/master/sync/types.go). Probably I'll add only `SummaryProvider` in a later PR when I move the code that will use it. Only the `Extender` interface will not be migrated for now, because we will not migrate the `atomic` syncer (which uses it).

vms/evm/sync/types/types.go
```diff
@@ -0,0 +1,22 @@
...
+// See the file LICENSE for licensing terms.
+
+package types
```
joshua-kim, line 4: It's not clear to me why this package exists - it doesn't have the same name, but it feels like an "interfaces" package which is an anti-pattern. This is only used in coreth's `vmsync` package for the common syncer types defined in `sync/statesync`. Would it make sense to move to one of those two packages? Go typically defines interfaces in the consumer but if there's a well-agreed upon abstraction and multiple packages are using it as an interface can be reasonable to define it in the producer.

On an unrelated note... why is `vmsync` called `vmsync`? It seems like we named it that way to disambiguate against `sync`, but you can use import aliasing to get around this when there is a conflict.

vms/evm/sync/types/types.go
```diff
@@ -0,0 +1,22 @@
...
+	// The sync will respect context cancellation.
+	Sync(ctx context.Context) error
+
+	// Name returns a human-readable name for this syncer implementation.
+	Name() string
+
+	// ID returns a stable, machine-oriented identifier (e.g., "state_block_sync", "state_code_sync",
+	// "state_evm_state_sync", "state_atomic_sync"). Implementations should ensure this is unique and
+	// stable across renames for logging/metrics/deduplication.
+	ID() string
```
joshua-kim, line 21: The caller [here](https://github.com/ava-labs/coreth/blob/f0db9b584e197a3a3f3459985d12fac2f0dbee9a/plugin/evm/vmsync/client.go#L194) already knows about the concrete syncer types being registered, so we don't need to make this part of the interface public contract. Even if we did, it feels like overkill that we're adding this to the interface for duplicate handling because we own the syncers that we're registering - so it shouldn't happen anyways. It also feels leaky that something that just knows about how to perform a sync operation also needs to define its own metrics namespace and it's own "human-readable name".

vms/evm/sync/types/types.go
powerslider: Ok, I'll start addressing all your points one by one:
- The state sync code in coreth is mostly split in two places roughly: `plugin/evm` and the root `sync` package:
   -  [sync/types.go](https://github.com/ava-labs/coreth/blob/master/sync/types.go) contains the interfaces that are implemented across both places. The reason why they are separated is because we don't want to have cyclic imports. The `Syncer` interface is implemented by:  
      - [sync/statesync/state_syncer.go](https://github.com/ava-labs/coreth/blob/master/sync/statesync/state_syncer.go) (for the HashDB state syncer) 
      - [sync/statesync/blocksync/syncer.go](https://github.com/ava-labs/coreth/blob/master/sync/blocksync/syncer.go) (for the block syncer)
      - [sync/statesync/code_syncer.go](https://github.com/ava-labs/coreth/blob/master/sync/statesync/code_syncer.go) (for the code syncer)
      - [plugin/evm/atomic/sync/syncer.go](https://github.com/ava-labs/coreth/blob/master/plugin/evm/atomic/sync/syncer.go) (for the atomic syncer)
 - [plugin/evm/vmsync](https://github.com/ava-labs/coreth/tree/master/plugin/evm/vmsync) is named this way because I needed to migrate this code initially from the root `sync` package in `coreth to `plugin/evm`. I am not saying that this is the best name ever, but it made sense given the structure in `coreth` at least temporarily. 
 - How I envision the new structure to look like is the following:
```
 # avalanchego project
 
 vms
    evm
       sync
          block
             syncer.go -> implements the Syncer interface
          code
             syncer.go -> implements the Syncer interface
          state
             syncer.go -> implements the Syncer interface
          types
             types.go - > contains the Syncer interface
          registry (up for debate)
             ... all files from the coreth vmsync package
```

```
#coreth project

plugin
   evm
     atomic
       sync
          syncer.go -> imports the Syncer interface from avalanchego and implements it
          # This will import "github.com/ava-labs/avalanchego/vms/evm/sync/types"
```

PS: I know that the coreth structure is really bad and the idea is to finally unify it under vms/evm/sync in avalanchego. For now the atomic syncer will remain in coreth based on the decision we made on the meetings beforehand. So we also need the flexibility of importing the Syncer interface that will be in avalanchego now in coreth as well.

powerslider:
Closed due to change in priorities. This code will be first moved to `graft/evm` and at some point afterwards will it get moved to `vms/evm/sync`.

PR #4406, opened 2025-10-09
[nix] Extract Go derivation to nested flake for cross-repo reuse

maru-ava:
## Why this should be merged

Exposing the go derivation via a flake allows the firewood flake to use it to provide its go version without preventing the avalanchego flake from depending on the firewood flake.

## How this was tested

- CI
- Manually
  - Locally: `nix ./nix/go`
  - Remotely: `nix run 'github:ava-labs/avalanchego?dir=nix/go&ref=9cd1c3aee7c2ef1b46621799d7ef0698c2326046'`

N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/DataDog/zstd v1.5.2
 	github.com/StephenButtolph/canoto v0.17.2
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/coreth v0.15.4-rc.3.0.20251002221438-a857a64c28ea
+	github.com/ava-labs/coreth v0.15.4-rc.4
 	github.com/ava-labs/ledger-avalanche-go v1.1.0
 	github.com/ava-labs/libevm v1.13.15-0.20251002164226-35926db4d661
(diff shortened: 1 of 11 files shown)


Elvis339:
Just to confirm I understand the issue correctly; before this change, did the firewood flake need to depend on the entire avalanchego flake just to get the Go version, which would create a circular dependency since avalanchego also depends on firewood? And this PR solves it by letting firewood import only the Go derivation from nix/go instead of the full avalanchego flake - is that right?

maru-ava:
> Just to confirm I understand the issue correctly; before this change, did the firewood flake need to depend on the entire avalanchego flake just to get the Go version, which would create a circular dependency since avalanchego also depends on firewood? And this PR solves it by letting firewood import only the Go derivation from nix/go instead of the full avalanchego flake - is that right?

Correct. And I'd like to reuse avalancheg's go derivation in the firewood flake instead of: 
 1. Using the nix version which is not guaranteed to match the version defined in firewood's go.mod
    - This will still work because golang is smart enough to download and switch to the correct version, but cleaner to just use the correct version to start with
 2. Duplicating the go derivation in the firewood repo

PR #4410, opened 2025-10-09
Add SetPreferenceWithContext

### LINKED ISSUE #4409: PlatformVM mempool verification should use child epoch height post-Granite activation
Currently, [the PlatformVM verifies ICM messages included in txs to be added to the mempool using the height returned by `GetMinimumHeight`](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/block/executor/manager.go#L135). With introduction of ProposerVM epochs in ACP-181, this will no longer be viable because the P-Chain height to verify ICM messages in blocks will be the epoched height, which can vary significantly from the minimum height of a given node (which stays close to the tip of the chain).

To provide best effort mempool verification, ICM messages in transactions should be verified using the epoched height of a child block should one be created on top of the current preferred block. 

This will require exposing the epoch height information to the `manager` from the ProposerVM.

michaelkaplan13:
## Why this should be merged

Fixes #4409

## How this works

- Passes the P-Chain height to be used for verification of a child of the preferred block in `SetPreferenceWithContext`.
- Uses the preferred context when verifying P-Chain txs before the enter the mempool.

## How this was tested

- [x] Manual local tests
- [x] Unit tests

No

### DIFF
--- a/vms/proposervm/vm.go
+++ b/vms/proposervm/vm.go
@@ -385,7 +385,7 @@ func (vm *VM) SetPreference(ctx context.Context, preferred ids.ID) error {
 	// preference with the P-Chain height to be used to verify a child of the
 	// preferred block.
 	if vm.setPreferenceVM != nil && preferredEpoch != (statelessblock.Epoch{}) {
-		// The P-Chain height used to verify a child of the preferred will
+		// The P-Chain height used to verify a child of the preferred block will
 		// potentially be different than the P-Chain height used to verify the
 		// preferred block if the preferred block seals the current epoch.
 		preferredPChainHeight, err := blk.pChainHeight(ctx)


StephenButtolph:
Testing locally this seems to correctly be verifying warp messages in the mempool against the p-chain height of the next block's epoch.

vms/proposervm/vm.go
```diff
@@ -370,7 +374,51 @@ func (vm *VM) SetPreference(ctx context.Context, preferred ids.ID) error {
...
+		return err
+	}
+
 	innerBlkID := blk.getInnerBlk().ID()
+
+	// If the inner VM implements SetPreferenceWithContext, use it to set the
+	// preference with the P-Chain height to be used to verify a child of the
+	// preferred block.
+	if vm.setPreferenceVM != nil && preferredEpoch != (statelessblock.Epoch{}) {
+		// The P-Chain height used to verify a child of the preferred will
```
geoff-vball:
```suggestion
		// The P-Chain height used to verify a child of the preferred block will
```

snow/engine/snowman/block/blocktest/set_preference_vm.go
```diff
@@ -0,0 +1,42 @@
...
+var (
+	errSetPreferenceWithContext = errors.New("unexpectedly called SetPreferenceWithContext")
+
+	_ block.SetPreferenceWithContextChainVM = (*SetPreferenceVM)(nil)
+)
+
+type SetPreferenceVM struct {
+	T *testing.T
+
+	CantSetPreferenceWithContext bool
```
geoff-vball, line 26: I would try to avoid `bool`s named `cant`, since at quick glance I think most people will read it as `can`

StephenButtolph: Fair point... This is currently how all the mocked VM helpers are though, so if we want to change how this works we should do a large pass.

`Cant` is used so that the default value (false) is more permissive

PR #4412, opened 2025-10-10
wallet: add signing options infrastructure for remote signers

felipemadero:
## Why this should be merged
This PR lays the groundwork for integrating remote signing services (like CubeSigner) into AvalancheGo by introducing a signing options infrastructure. This enables wallet signers to receive essential transaction context while maintaining backward compatibility with existing implementations.

## How this works
The implementation introduces a functional options pattern for wallet signing:
- **SigningOptions Infrastructure**: New `SigningOption` type with `WithChainAlias()` and `WithNetworkID()` helpers to provide signing context
- **Extended Signer Interface**: The `keychain.Signer` interface now accepts variadic `SigningOption` parameters, allowing signers to receive network ID and chain alias context
- **Wallet Integration**: Updated P-chain, X-chain, and C-chain wallet signers to pass chain alias and network ID to `Sign()` calls
- **Backward Compatibility**: Existing signers (Ledger, secp256k1) accept but ignore options, maintaining the `signHash` parameter for Ledger compatibility

This provides the foundation needed for remote signing services to properly format addresses and validate transactions based on chain context, which enables future CubeSigner integration.

## How this was tested
- Unit tests

### DIFF
--- a/wallet/chain/c/signer.go
+++ b/wallet/chain/c/signer.go
@@ -60,16 +60,18 @@ type SignerBackend interface {
 }
 
 type txSigner struct {
-	avaxKC  keychain.Keychain
-	ethKC   EthKeychain
-	backend SignerBackend
+	avaxKC    keychain.Keychain
+	ethKC     EthKeychain
+	backend   SignerBackend
+	networkID uint32
 }
 
-func NewSigner(avaxKC keychain.Keychain, ethKC EthKeychain, backend SignerBackend) Signer {
+func NewSigner(avaxKC keychain.Keychain, ethKC EthKeychain, backend SignerBackend, networkID uint32) Signer {
 	return &txSigner{
-		avaxKC:  avaxKC,
-		ethKC:   ethKC,
-		backend: backend,
+		avaxKC:    avaxKC,
+		ethKC:     ethKC,
+		backend:   backend,
+		networkID: networkID,
 	}
 }
 
@@ -80,10 +82,10 @@ func (s *txSigner) SignAtomic(ctx context.Context, tx *atomic.Tx) error {
 		if err != nil {
 			return err
 		}
-		return sign(tx, true, signers)
+		return sign(tx, true, signers, s.networkID)
 	case *atomic.UnsignedExportTx:
 		signers := s.getExportSigners(utx.Ins)
-		return sign(tx, true, signers)
+		return sign(tx, true, signers, s.networkID)
 	default:
 		return fmt.Errorf("%w: %T", errUnknownTxType, tx)
 	}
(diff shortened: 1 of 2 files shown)


utils/crypto/keychain/keychain.go
```diff
@@ -22,10 +22,10 @@ var (
 )
 
 // Signer implements functions for a keychain to return its main address and
-// to sign a hash
+// to sign a hash or transaction
 type Signer interface {
 	SignHash([]byte) ([]byte, error)
-	Sign([]byte) ([]byte, error)
+	Sign([]byte, ...SigningOption) ([]byte, error)
```
joshua-kim, line 28: I'm wondering why this has to change the interface - do the values for chain id/network id ever change for our usage of cube signer? If not, I'm wondering if this can live in the cube signer implementation when it's initialized instead of us making every signer know about the chain and network id even though only one implementation cares about it.

felipemadero: I see the concern about interface complexity. Let me explain the constraints I'm working with:

X-chain support is the main blocker: With the current design, all three chains work with a single keychain instance. If we create separate keychains for P and C at wallet construction (passing chain context there), we lose X-chain operations—and I expect CubeSigner/Ledger users will want X-chain support.

Why signers need this context:
- CubeSigner (and any future remote signers) need chain/network info to decide whether to block/review transactions. CubeSigner's API specifically requires these parameters, and may ask for tx type in the future.
- Chain information is necessary to deserialize transactions and extract details.
- Network ID is convenience info that's easy for the wallet to provide from existing context, avoiding the need for keychain implementations to depend on tx internals (reducing coupling).

Ledger will also need this: There's currently tx-specific logic in the wallet (deciding which txs need SignHash vs Sign). A subsequent PR will move this decision to the Ledger implementation, which also needs chain context to make the right choice.

Why the current approach works: The variadic options pattern:
- Doesn't break existing implementations (they trivially ignore the parameter)
- Provides a general wallet→signer communication channel for future extensibility
- Avoids tight coupling (implementations don't need to parse/inspect tx bytes)
- Keeps keychains chain-agnostic at construction (chain/network aren't conceptually part of a keychain's identity)

Do you see an implementation approach I'm missing that would avoid these tradeoffs?

joshua-kim: I think it can be okay for the a keychain implementation that does need to know the network/chain-id to be know about it at initialization time instead of it living on the interface signature. In practice the keychains are only used for our wallets to sign txs, and our wallets already need to know about the chain id (which is similar to network id and chain alias) because they track utxo sets ([ref](https://github.com/ava-labs/avalanchego/blob/45df9e6926b93aacd11589a7b90a76078ae956d9/wallet/subnet/primary/wallet.go#L118)).   

If we do this, I think we can remove the diff in avalanchego entirely since we'll push the network/chain alias down into the sdk.

felipemadero: Ok will close this PR then. Will create a custom MakeWallet inside SDK ref https://github.com/ava-labs/avalanchego/blob/45df9e6926b93aacd11589a7b90a76078ae956d9/wallet/subnet/primary/wallet.go#L81 that takes in three keychains, P/X/C, so there is no context ambiguity between P and X.
I will also fix https://github.com/ava-labs/avalanchego/pull/4413 so as to not depend on signing options.

felipemadero:
Closing as per https://github.com/ava-labs/avalanchego/pull/4412#discussion_r2434216885

PR #4413, opened 2025-10-10
wallet: remove ledger support in favor of golang SDK implementation of ledger keychain

felipemadero:
## Why this should be merged
This PR removes built-in Ledger hardware wallet support from AvalancheGo's wallet implementation, also delegating hash signing decisions to keychain implementations instead.
This change:
- Simplifies the wallet signing interface by removing the dual-path signing logic (signHash vs Sign)
- Enables keychain implementations (like CubeSigner SDK or future Ledger Go SDK) to make their own signing decisions based on transaction context
- Reduces maintenance burden by removing direct hardware wallet integration code

## How this works
The implementation removes:
- **Ledger Integration**: All Ledger-specific code (`utils/crypto/ledger`, ledger keychain, tests, and mocks)
- **signHash Parameter**: Removed from P-chain and C-chain wallet signers, eliminating the conditional signing logic
- **Dual Signing Paths**: Wallets now only call `Sign(unsignedBytes, ...SigningOption)` with full transaction bytes and context

Keychain implementations receive the full transaction bytes along with signing context (chain alias, network ID) and can decide internally how to sign. For example:
- A remote signing service SDK can send firm transaction information for contextual validation
- Ledger Go SDK can hash the bytes internally if needed for hardware compatibility

## How this was tested
- Unit tests


utils/crypto/keychain/keychain.go
```diff
@@ -4,25 +4,12 @@
...
-	ErrInvalidNumAddrsToDerive = errors.New("number of addresses to derive should be greater than 0")
-	ErrInvalidNumAddrsDerived  = errors.New("incorrect number of ledger derived addresses")
-	ErrInvalidNumSignatures    = errors.New("incorrect number of signatures")
-)
-
 // Signer implements functions for a keychain to return its main address and
-// to sign a hash
+// to sign a hash or transaction
 type Signer interface {
 	SignHash([]byte) ([]byte, error)
```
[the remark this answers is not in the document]
felipemadero: The SignHash method is kept in the interface for backward compatibility

wallet/chain/p/signer/visitor.go
```diff
@@ -327,13 +326,11 @@ func (s *visitor) getAuthSigners(ownerID ids.ID, auth verify.Verifiable) ([]keyc
...
 }
 
-// TODO: remove [signHash] after the ledger supports signing all transactions.
-func sign(tx *txs.Tx, signHash bool, txSigners [][]keychain.Signer) error {
+func sign(tx *txs.Tx, txSigners [][]keychain.Signer) error {
 	unsignedBytes, err := txs.Codec.Marshal(txs.CodecVersion, &tx.Unsigned)
 	if err != nil {
 		return fmt.Errorf("couldn't marshal unsigned tx: %w", err)
 	}
-	unsignedHash := hashing.ComputeHash256(unsignedBytes)
```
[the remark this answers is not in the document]
felipemadero: We've been working with keychain implementations to ensure they are updated to handle hashing internally based on their specific requirements.

utils/crypto/keychain/keychain.go
joshua-kim: Do we still need this?

yacovm: i think not

felipemadero: Removed

StephenButtolph (approved):

Issue #4362, opened 2025-09-29
Expose metrics server in tests without starting Prometheus agent / remote write


AminR443:
### Context and scope
In test runs (e.g., re-execution), `avalanchego` only exposes metrics when `METRICS_ENABLED` is set. That path currently (a) starts a Prometheus agent configured for remote write and (b) opens a metrics HTTP server on a random port. For firewood re-execution tests, we want to pull metrics locally (scrape `/ext/metrics`) without starting Prometheus agent/remote write and need a deterministic way to discover or set the listen port. We need these changes so we can configure our own Prometheus to scrape from it and monitor the metrics using Grafana during the re-execution test.

**References:** 
* random port bind in tests: [https://github.com/ava-labs/avalanchego/blob/47550e9d21c36acca6862c8749bc1de6c3392fc1/tests/prometheus_server.go#L17](https://github.com/ava-labs/avalanchego/blob/47550e9d21c36acca6862c8749bc1de6c3392fc1/tests/prometheus_server.go#L17)
*  metrics-enabled path that starts both agent and server: [https://github.com/ava-labs/avalanchego/blob/47550e9d21c36acca6862c8749bc1de6c3392fc1/tests/reexecute/c/vm_reexecute_test.go#L544-L555](https://github.com/ava-labs/avalanchego/blob/47550e9d21c36acca6862c8749bc1de6c3392fc1/tests/reexecute/c/vm_reexecute_test.go#L544-L555)

### Discussion and alternatives
Proposed code changes:

* Refactor test metrics setup so the HTTP server and Prometheus agent can be started independently.
* Support explicit port for metrics.

We can do this in two ways:

1. **Split env toggles**: `METRICS_SERVER_ENABLED` / `METRICS_PUSH_ENABLED`.
2. **Mode selector**: `METRICS_MODE=server|push|both`.

We could also add `METRICS_PORT` or agree on a constant to use instead of `0`.

aaronbuchwald:
Setting up prometheus agent and server independently (and configuring them independently) sounds like a good change to me. Same for adding an option to specify the port the server will listen on. iirc this was originally an explicit choice to listen on a random port because the metrics were going to be collected anyways, so a hardcoded port was unnecessary and could lead to misconfiguration. I think it makes perfect sense to expose an option to configure it for ease of development though.

I'd opt to split into two separately configured services for the prom agent and prom server, but open to alternatives.

[...]

PR #4415, opened 2025-10-12
feat(reexecute/c): decouple metrics server and collector

RodrigoVillar:
## Why this should be merged

As mentioned in #4362, splitting up the Prometheus server and collector is ideal for clients of Firewood who want access to VM metrics but would prefer to use their own monitoring stack. 

Although #4362 also discusses the option of hardcoding the metrics server port, I've opted to split this out into a separate PR.

## How this works

Changes the `METRICS_ENABLED` parameter to `METRICS_MODE`, which has the following three options:
- `disabled`: does not start both the Prometheus server and collector
- `server-only`: starts only the Prometheus server
- `full`: starts both the Prometheus server and collector

I've opted for keeping a single environment variable rather than splitting `METRICS_ENABLED` since there should never be a case for starting the Prometheus collector but not the Prometheus server. In either the `server-only` or `full` cases, a metrics endpoint or Grafana URL is printed out, respectively.

## How this was tested

CI

No

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -30,6 +30,7 @@
 /tests/*.md @joshua-kim @maru-ava @meaghanfitzgerald
 /tests/reexecute/ @aaronbuchwald
 /tools/* @joshua-kim @maru-ava
+/vms/evm/ @ARR4N @joshua-kim @StephenButtolph
 /x/blockdb @DracoLi
 /x/merkledb @joshua-kim @rrazvan1
 /x/sync @joshua-kim @rrazvan1
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## Pending
+
+### Configs
+
+- Changed default block delay for L1s (other than Primary Network) to 0.
+
 ## [v1.13.5](https://github.com/ava-labs/avalanchego/releases/tag/v1.13.5)
 
 This version is backwards compatible to [v1.13.0](https://github.com/ava-labs/avalanchego/releases/tag/v1.13.0). It is optional, but encouraged.
--- a/config/config.go
+++ b/config/config.go
@@ -1103,6 +1103,13 @@ func getSubnetConfigsFromDir(v *viper.Viper, subnetIDs []ids.ID) (map[ids.ID]sub
 }
 
 func getDefaultSubnetConfig(v *viper.Viper) subnets.Config {
+	subnetDefaults := getPrimaryNetworkConfig(v)
+	// Allow L1s (other than Primary Network) to use their own throttling mechanisms.
+	subnetDefaults.ProposerMinBlockDelay = 0
+	return subnetDefaults
+}
+
+func getPrimaryNetworkConfig(v *viper.Viper) subnets.Config {
 	return subnets.Config{
 		ConsensusParameters:         getConsensusConfig(v),
 		ValidatorOnly:               false,
@@ -1326,7 +1333,7 @@ func GetNodeConfig(v *viper.Viper) (node.Config, error) {
 		return node.Config{}, fmt.Errorf("couldn't read subnet configs: %w", err)
 	}
 
-	primaryNetworkConfig := getDefaultSubnetConfig(v)
+	primaryNetworkConfig := getPrimaryNetworkConfig(v)
 	if err := primaryNetworkConfig.Valid(); err != nil {
 		return node.Config{}, fmt.Errorf("invalid consensus parameters: %w", err)
 	}
--- a/network/config.go
+++ b/network/config.go
@@ -181,4 +181,8 @@ type Config struct {
 	// Specifies how much disk usage each peer can cause before
 	// we rate-limit them.
 	DiskTargeter tracker.Targeter `json:"-"`
+
+	// If true, connects to all validators regardless of primary network validator
+	// status or of configured tracked subnets.
+	ConnectToAllValidators bool `json:"connectToAllValidators"`
 }
--- a/network/ip_tracker.go
+++ b/network/ip_tracker.go
@@ -44,6 +44,7 @@ func newIPTracker(
 	trackedSubnets set.Set[ids.ID],
 	log logging.Logger,
 	registerer prometheus.Registerer,
+	connectToAllValidators bool,
 ) (*ipTracker, error) {
 	bloomMetrics, err := bloom.NewMetrics("ip_bloom", registerer)
 	if err != nil {
@@ -64,11 +65,12 @@ func newIPTracker(
 			Name: "tracked_subnets",
 			Help: "number of subnets this node is monitoring",
 		}),
-		bloomMetrics:   bloomMetrics,
-		tracked:        make(map[ids.NodeID]*trackedNode),
-		bloomAdditions: make(map[ids.NodeID]int),
-		connected:      make(map[ids.NodeID]*connectedNode),
-		subnet:         make(map[ids.ID]*gossipableSubnet),
+		connectToAllValidators: connectToAllValidators,
+		bloomMetrics:           bloomMetrics,
+		tracked:                make(map[ids.NodeID]*trackedNode),
+		bloomAdditions:         make(map[ids.NodeID]int),
+		connected:              make(map[ids.NodeID]*connectedNode),
+		subnet:                 make(map[ids.ID]*gossipableSubnet),
 	}
 	err = errors.Join(
 		registerer.Register(tracker.numTrackedPeers),
@@ -230,6 +232,8 @@ type ipTracker struct {
 	connected map[ids.NodeID]*connectedNode
 	// subnet tracks all the subnets that have at least one gossipable ID.
 	subnet map[ids.ID]*gossipableSubnet
+
+	connectToAllValidators bool
 }
 
 // ManuallyTrack marks the provided nodeID as being desirable to connect to.
@@ -267,6 +271,7 @@ func (i *ipTracker) ManuallyGossip(subnetID ids.ID, nodeID ids.NodeID) {
 //  1. The node has been manually tracked.
 //  2. The node has been manually gossiped on a tracked subnet.
 //  3. The node is currently a validator on a tracked subnet.
+//  4. The node is currently a validator on any subnet and connectToAllValidators is true.
 func (i *ipTracker) WantsConnection(nodeID ids.NodeID) bool {
 	i.lock.RLock()
 	defer i.lock.RUnlock()
@@ -305,6 +310,7 @@ func (i *ipTracker) ShouldVerifyIP(
 //  1. The provided IP is from a node whose connection is desired on a tracked
 //     subnet.
 //  2. This IP is newer than the most recent IP we know of for the node.
+//  3. The node is a validator and connectToAllValidators is true.
 //
 // If this IP is replacing a gossipable IP, this IP will also be marked as
 // gossipable.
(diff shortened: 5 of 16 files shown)


RodrigoVillar:
The `full` option is enabled in CI as seen here: https://github.com/ava-labs/avalanchego/actions/runs/18465753103/job/52607232591?pr=4415#step:4:968

When running locally with `METRICS_MODE=server-only`, you get the following:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/tests/reexecute/c
cpu: Apple M1 Max
BenchmarkReexecuteRange
BenchmarkReexecuteRange/[1,1000]-Config-default-Runner-dev
[10-13|08:45:03.588] INFO c-chain-reexecution c/vm_reexecute_test.go:616 metrics endpoint available {"url": "http://127.0.0.1:62042/ext/metrics"}
```

When running locally with `METRICS_MODE=full`, you get the following:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/tests/reexecute/c
cpu: Apple M1 Max
BenchmarkReexecuteRange
BenchmarkReexecuteRange/[1,1000]-Config-default-Runner-dev
[10-13|08:46:22.299] INFO prometheus tmpnet/monitor_processes.go:370 collector already running {"cmd": "prometheus"}
[10-13|08:46:22.299] INFO prometheus tmpnet/monitor_processes.go:610 waiting for collector readiness {"cmd": "prometheus", "url": "http://127.0.0.1:9090/-/ready", "logPath": "/Users/rodrigo.villar/.tmpnet/prometheus/prometheus.log"}
[10-13|08:46:22.300] INFO prometheus tmpnet/monitor_processes.go:634 collector ready {"cmd": "prometheus"}
[10-13|08:46:22.300] INFO prometheus tmpnet/monitor_processes.go:60 To stop: tmpnetctl stop-metrics-collector
[10-13|08:46:22.300] INFO c-chain-reexecution c/vm_reexecute_test.go:665 metrics available via grafana {"url": "https://grafana-poc.avax-dev.network/d/Gl1I20mnk/c-chain?&var-filter=network_uuid%7C%3D%7C8ce59182-3fcf-4a8d-8272-eeadbcea1537&var-filter=is_ephemeral_node%7C%3D%7Cfalse&from=1760359582300&to=now"}
```

Taskfile.yml
```diff
@@ -203,7 +203,7 @@ tasks:
       END_BLOCK: '{{.END_BLOCK}}'
       LABELS: '{{.LABELS | default ""}}'
       BENCHMARK_OUTPUT_FILE: '{{.BENCHMARK_OUTPUT_FILE | default ""}}'
-      METRICS_ENABLED: '{{.METRICS_ENABLED | default "false"}}'
+      METRICS_MODE: '{{.METRICS_MODE | default "disabled"}}'
```
aaronbuchwald: Will this work if we set the default to the empty string instead of disabled since it will take on the default behavior rather than providing a separate default here?

Taskfile.yml
```diff
@@ -228,7 +228,7 @@ tasks:
       END_BLOCK: '{{.END_BLOCK | default "250000"}}'
       LABELS: '{{.LABELS | default ""}}'
       BENCHMARK_OUTPUT_FILE: '{{.BENCHMARK_OUTPUT_FILE | default ""}}'
-      METRICS_ENABLED: '{{.METRICS_ENABLED | default "false"}}'
+      METRICS_MODE: '{{.METRICS_MODE | default "disabled"}}'
```
aaronbuchwald: same question on setting the default here

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -94,6 +101,41 @@ var (
...
+		*m = MetricsDisabled
+	case "server-only":
+		*m = MetricsServerOnly
+	case "full":
+		*m = MetricsFull
+	default:
+		return fmt.Errorf("invalid metrics mode: %s (valid options: disabled, server-only, full)", s)
+	}
+	return nil
+}
```
aaronbuchwald: Could we just use a string here and perhaps an alias type rather than implementing `Set`?

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -185,9 +227,8 @@ func benchmarkReexecuteRange(
 	r.NoError(prefixGatherer.Register("avalanche_snowman", consensusRegistry))
 
 	log := tests.NewDefaultLogger("c-chain-reexecution")
-
-	if metricsEnabled {
-		collectRegistry(b, log, "c-chain-reexecution", prefixGatherer, labels)
+	if metricsMode.shouldStartServer() {
+		collectRegistry(b, log, "c-chain-reexecution", prefixGatherer, labels, metricsMode.shouldStartCollector())
```
aaronbuchwald: Can we clean this up a little bit? It seems odd that we decompose `metricsMode` into two separate booleans and use one to handle the if condition here and the other half of it as an argument that gets passed in.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -554,20 +595,40 @@ func newConsensusMetrics(registry prometheus.Registerer) (*consensusMetrics, err
...
+	tb testing.TB,
+	log logging.Logger,
+	name string,
+	gatherer prometheus.Gatherer,
+	labels map[string]string,
+	startCollector bool,
+) {
 	r := require.New(tb)
 
+	server, err := tests.NewPrometheusServer(gatherer)
```
aaronbuchwald, line 571: This will still listen on a random port, so doesn't seem to provide the full configurability requested by https://github.com/ava-labs/avalanchego/issues/4362

Taskfile.yml
RodrigoVillar: Great point; updated so that we set the default to the empty string here (the reexecution test will provide us with a default): be761ac

Taskfile.yml
RodrigoVillar: Done: be761ac

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: This is referenced in the PR description (i.e. hardcoding a port is a feature separate from separating the metrics server and collector).

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Simplified the `metricsMode` type here: a7cb056696109f2f60ac980ef4341b2faba9188c

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Cleaned up the logic for starting the metrics server vs starting the metrics server and collector here: 9e319d0

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -546,9 +569,31 @@ func newConsensusMetrics(registry prometheus.Registerer) (*consensusMetrics, err
 	return m, nil
 }
 
-// collectRegistry starts prometheus and collects metrics from the provided gatherer.
-// Attaches the provided labels + GitHub labels if available to the collected metrics.
-func collectRegistry(tb testing.TB, log logging.Logger, name string, gatherer prometheus.Gatherer, labels map[string]string) {
+// startServer starts a Prometheus server for the provided gatherer.
+func startServer(
```
aaronbuchwald, line 564: Can we avoid duplicating the code between `startServer` and `startServerAndColletor` ?

tests/reexecute/c/README.md
```diff
@@ -42,7 +42,18 @@ export AWS_REGION=us-east-2
 
 ### Metrics Collection
 
-If running with metrics collection, enabled in CI and configured locally with `METRICS_ENABLED=true`, follow the instructions in the e2e [README](../../e2e/README.md#monitoring) to set the required Prometheus environment variables.
+If running locally, there are three options for metrics collection:
+
+- `METRICS_MODE=disabled`: no metrics are available.
+- `METRICS_MODE=server-only`: starts a Prometheus server exporting VM metrics. A
+  link to the metrics endpoint is logged during execution.
+- `METRICS_MODE=full`: starts both a Prometheus server exporting VM metrics and
```
aaronbuchwald: Given these names are not very self-explanatory (it's not clear what server-only and full refer to without the descriptions), I think it would be better to simply configure the two separately and require that if grafana is enabled, then the prometheus server must be enabled as well.

It's fine imo for the metrics server to be enabled and grafana disabled by default since that won't require setting any extra credentials.

RodrigoVillar: Done: 5160738

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: Done: 915ae5d

tests/reexecute/c/README.md
```diff
@@ -42,7 +42,14 @@ export AWS_REGION=us-east-2
 
 ### Metrics Collection
 
-If running with metrics collection, enabled in CI and configured locally with `METRICS_ENABLED=true`, follow the instructions in the e2e [README](../../e2e/README.md#monitoring) to set the required Prometheus environment variables.
+If running locally, metrics collection can be customized via the following parameters:
+
+- `METRICS_SERVER_ENABLED`: starts a Prometheus server exporting VM metrics.
+- `METRICS_COLLECTOR_ENABLED`: starts a Prometheus collector (if enabled, then `METRICS_SERVER_ENABLED` must be enabled as well).
```
maru-ava, line 48: Why isn't this just implicit?

RodrigoVillar: On it's own, I'm not opposed to `METRICS_COLLECTOR_ENABLED=true` implicitly setting `METRICS_SERVER_ENABLED=true` as well. However, considering that this PR will be followed up by #4418 (which adds the ability to configure a port for the metrics server), I think this becomes confusing (i.e. it isn't clear what happens if `METRICS_COLLECTOR_ENABLED=true` and `METRICS_PORT=X` without reading the description of `METRICS_COLLECTOR_ENABLED` ).

This could be fixed by renaming `METRICS_COLLECTOR_ENABLED` to `METRICS_SERVER_AND_COLLECTOR_ENABLED`, but this looks similar to a previous iteration of this PR which received this review comment: https://github.com/ava-labs/avalanchego/pull/4415#discussion_r2429225890

maru-ava: I'm not going to block on this, but a rethink is definitely suggested. I don't think the comment you linked to suggesting seperate flags precludes improving what appears in this PR.

RodrigoVillar: Can you elaborate?  The only options regarding the flag design are as follows:
- Enabling the collector implicitly enables the server.
- Enabling the collector does not implicitly enable the server.

With this PR choosing the latter, I'm not sure what else needs to be rethought (if you think enabling the collector _**needs**_ to implicitly enable the server, I'm happy to follow-up on that).

maru-ava: I think the issue is the chosen terminology. `ENABLE_METRICS_SERVER` starts a `PrometheusServer` instance, and that name could be confused with an actual Prometheus server (one that collects and aggregate metrics). I recommend changing `ENABLE_METRICS_SERVER` to `ENABLE_METRICS_EXPORT` and `PrometheusServer` to `MetricsExporter`. That way there is a unambiguous relationship between exporting metrics and collecting those exported metrics. Once that change is made, I think it would be a no-brainer to implicitly enable metrics export when collection is enabled rather than requiring that the user explicitly configure that.

RodrigoVillar: Thank you for the feedback. Repeating what I said in #4418, enabling the metrics collector in #4418 will implicitly enable the metrics server with a port of `0` (as of commit [830e8c1](https://github.com/ava-labs/avalanchego/pull/4418/commits/830e8c1303d4cd93bd61f3192b3a9ddc0dc63c2d))

PR #4418, opened 2025-10-14
feat(reexecute/c): explicit metrics port

RodrigoVillar:
## Why this should be merged

As referenced in #4362, this PR allows clients of the reexecution test to specify the port which the metrics server will listen on.

## How this works

Adds a `metrics-server-port` flag to the reexecution test.

## How this was tested

CI + queried metrics server locally.

No

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -10,6 +10,7 @@ set -euo pipefail
 #   END_BLOCK: The ending block height (inclusive).
 #   LABELS (optional): Comma-separated key=value pairs for metric labels.
 #   BENCHMARK_OUTPUT_FILE (optional): If set, benchmark output is also written to this file.
+#   METRICS_SERVER_ENABLED (optional): If set, enables the metrics server.
 #   METRICS_SERVER_PORT (optional): If set, determines the port the metrics server will listen to.
 #   METRICS_COLLECTOR_ENABLED (optional): If set, enables the metrics collector.
 
@@ -27,6 +28,7 @@ cmd="go test -timeout=0 -v -benchtime=1x -bench=BenchmarkReexecuteRange -run=^$ 
   --start-block=\"${START_BLOCK}\" \
   --end-block=\"${END_BLOCK}\" \
   ${LABELS:+--labels=\"${LABELS}\"} \
+  ${METRICS_SERVER_ENABLED:+--metrics-server-enabled=\"${METRICS_SERVER_ENABLED}\"} \
   ${METRICS_SERVER_PORT:+--metrics-server-port=\"${METRICS_SERVER_PORT}\"} \
   ${METRICS_COLLECTOR_ENABLED:+--metrics-collector-enabled=\"${METRICS_COLLECTOR_ENABLED}\"}"
 
--- a/tests/reexecute/c/README.md
+++ b/tests/reexecute/c/README.md
@@ -44,12 +44,13 @@ export AWS_REGION=us-east-2
 
 If running locally, metrics collection can be customized via the following parameters:
 
-- `METRICS_SERVER_PORT`: if set, starts a Prometheus server exporting VM metrics and sets the port the server will listen to.
-- `METRICS_COLLECTOR_ENABLED`: starts a Prometheus collector. If `METRICS_SERVER_PORT` is not set, enabling the collector implicitly sets `METRICS_SERVER_PORT` to `0`.
+- `METRICS_SERVER_ENABLED`: starts a Prometheus server exporting VM metrics.
+- `METRICS_SERVER_PORT`: if set, determines the port the Prometheus server will listen to (set to `0` by default).
+- `METRICS_COLLECTOR_ENABLED`: starts a Prometheus collector. If `METRICS_SERVER_ENABLED` is not set, enabling the collector implicitly sets `METRICS_SERVER_ENABLED` to `true`.
 
 When utilizing the metrics collector feature, follow the instructions in the e2e [README](../../e2e/README.md#monitoring) to set the required Prometheus environment variables.
 
-Running the re-execution test in CI will always set `METRICS_COLLECTOR_ENABLED=true`.
+Running the re-execution test in CI will always set `METRICS_SERVER_ENABLED=true` and `METRICS_COLLECTOR_ENABLED=true`.
 
 ## Quick Start
 
@@ -237,7 +238,7 @@ The `CONFIG` parameter currently only supports pre-defined configs and not passi
 
 The C-Chain benchmarks export VM metrics to the same Grafana instance as AvalancheGo CI: https://grafana-poc.avax-dev.network/.
 
-To export metrics for a local run, simply set the Taskfile variable `METRICS_COLLECTOR_ENABLED=true` either via environment variable or passing it at the command line.
+To export metrics for a local run, simply set the Taskfile variables `METRICS_SERVER_ENABLED=true` and `METRICS_COLLECTOR_ENABLED=true` either via environment variable or passing it at the command line.
 
 You can view granular C-Chain processing metrics with the label attached to this job (job="c-chain-reexecution") [here](https://grafana-poc.avax-dev.network/d/Gl1I20mnk/c-chain?orgId=1&from=now-5m&to=now&timezone=browser&var-datasource=P1809F7CD0C75ACF3&var-filter=job%7C%3D%7Cc-chain-reexecution&var-chain=C&refresh=10s).
 
--- a/tests/reexecute/c/vm_reexecute_test.go
+++ b/tests/reexecute/c/vm_reexecute_test.go
@@ -63,7 +63,8 @@ var (
 	executionTimeout   time.Duration
 	labelsArg          string
 
-	metricsServerPort          *uint64
+	metricsServerEnabledArg    bool
+	metricsServerPortArg       uint64
 	metricsCollectorEnabledArg bool
 
 	networkUUID string = uuid.NewString()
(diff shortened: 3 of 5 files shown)


RodrigoVillar:
Here's an instance of setting the metrics port locally with the command: `task reexecute-cchain-range START_BLOCK=1 END_BLOCK=50_000 METRICS_SERVER_PORT=5000`

Screenshot 2025-10-20 at 15 42 55

tests/prometheus_server.go
```diff
@@ -28,25 +32,32 @@ type PrometheusServer struct {
 // NewPrometheusServer creates and starts a Prometheus server with the provided gatherer
 // listening on 127.0.0.1:0 and serving /ext/metrics.
 func NewPrometheusServer(gatherer prometheus.Gatherer) (*PrometheusServer, error) {
+	return NewPrometheusServerWithPort(gatherer, defaultMetricsPort)
+}
+
+// NewPrometheusServer creates and starts a Prometheus server with the provided gatherer
```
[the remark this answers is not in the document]
RodrigoVillar: Thanks Copilot: ac02cbf

scripts/benchmark_cchain_range.sh
```diff
@@ -28,7 +29,8 @@ cmd="go test -timeout=0 -v -benchtime=1x -bench=BenchmarkReexecuteRange -run=^$
   --end-block=\"${END_BLOCK}\" \
   ${LABELS:+--labels=\"${LABELS}\"} \
   ${METRICS_SERVER_ENABLED:+--metrics-server-enabled=\"${METRICS_SERVER_ENABLED}\"} \
-  ${METRICS_COLLECTOR_ENABLED:+--metrics-collector-enabled=\"${METRICS_COLLECTOR_ENABLED}\"}"
+  ${METRICS_COLLECTOR_ENABLED:+--metrics-collector-enabled=\"${METRICS_COLLECTOR_ENABLED}\"} \
+  ${METRICS_SERVER_PORT:+--metrics-server-port=\"${METRICS_SERVER_PORT}\"}"
```
aaronbuchwald: Could we change the ordering here and do a pass throughout to ensure that the metric server port param is next to metrics server enabled so that these are properly clustered together by the resource that they change?

Is it possible to use a single parameter rather than a metrics server port and metrics server enabled param? For example, we could use only the port param, treat an unset param as disabled, and then treat 0 or another port param as enabling it with the specified port.

RodrigoVillar: > Is it possible to use a single parameter rather than a metrics server port and metrics server enabled param? For example, we could use only the port param, treat an unset param as disabled, and then treat 0 or another port param as enabling it with the specified port.

Yes, I've updated the PR to use a single parameter `metrics-server-port` as described above.

> Could we change the ordering here and do a pass throughout to ensure that the metric server port param is next to metrics server enabled so that these are properly clustered together by the resource that they change?

Now that we have just one metrics server parameter, this comment is no longer applicable.

maru-ava:
I would like to see the following comment addressed: https://github.com/ava-labs/avalanchego/pull/4415#discussion_r2448881857

In keeping with that previous comment, I remain unconvinced of the wisdom of requiring that someone set a port to enable collection. Implicitly setting port 0 when collection is enabled would be consistent with how tmpnet collectors are configured by default, and I don't think there's a good reason to deviate here.

RodrigoVillar:
@maru-ava I've updated this PR so that if `METRICS_COLLECTOR_ENABLED=true`, then the metrics server port is implicitly set to `0` here: 527b652f0ec6d1ec2e23809c013c576a1c6d62f9.

tests/prometheus_server.go
```diff
@@ -28,25 +32,32 @@ type PrometheusServer struct {
 // NewPrometheusServer creates and starts a Prometheus server with the provided gatherer
 // listening on 127.0.0.1:0 and serving /ext/metrics.
 func NewPrometheusServer(gatherer prometheus.Gatherer) (*PrometheusServer, error) {
+	return NewPrometheusServerWithPort(gatherer, defaultMetricsPort)
+}
+
+// NewPrometheusServerWithPort creates and starts a Prometheus server with the provided gatherer
+// listening on 127.0.0.1:port and serving /ext/metrics.
+func NewPrometheusServerWithPort(gatherer prometheus.Gatherer, port uint64) (*PrometheusServer, error) {
```
aaronbuchwald, line 40: Consider using functional options pattern here, but think this is fine since it's just adding one extra constructor.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -119,9 +128,8 @@ func TestMain(m *testing.M) {
 
 	flag.Parse()
 
-	if metricsCollectorEnabledArg && !metricsServerEnabledArg {
-		fmt.Fprint(os.Stderr, "metrics collector is enabled but metrics server is disabled.\n")
-		os.Exit(1)
+	if metricsCollectorEnabledArg && metricsServerPort == nil {
+		metricsServerPort = new(uint64)
```
aaronbuchwald: (no changes needed) this override feels a bit odd, but I think this works out pretty nicely in terms of the collector and port flags supported.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -104,7 +104,16 @@ func TestMain(m *testing.M) {
...
+	flag.Func("metrics-server-port", "Starts a metrics server and sets the port it will listen to", func(s string) error {
+		port, err := strconv.ParseUint(s, 10, 64)
+		if err != nil {
+			return err
+		}
+
+		metricsServerPort = new(uint64)
+		*metricsServerPort = port
+		return nil
+	})
```
aaronbuchwald: Isn't there a way to simply check if a flag is set without using a Func like this?

tests/reexecute/c/README.md
```diff
@@ -44,12 +44,12 @@ export AWS_REGION=us-east-2
 
 If running locally, metrics collection can be customized via the following parameters:
 
-- `METRICS_SERVER_ENABLED`: starts a Prometheus server exporting VM metrics.
-- `METRICS_COLLECTOR_ENABLED`: starts a Prometheus collector (if enabled, then `METRICS_SERVER_ENABLED` must be enabled as well).
+- `METRICS_SERVER_PORT`: if set, starts a Prometheus server exporting VM metrics and sets the port the server will listen to.
+- `METRICS_COLLECTOR_ENABLED`: starts a Prometheus collector. If `METRICS_SERVER_PORT` is not set, enabling the collector implicitly sets `METRICS_SERVER_PORT` to `0`.
```
aaronbuchwald: This works out pretty nicely

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: If we keep `metricsServerPort` as a `uint64` value, then there's no way to differentiate a client setting `METRICS_SERVER_PORT=0` and a client declining to start the metrics server (in both cases, `metricsServerPort` will be `0`).

Edit: as to the usage of `flag.Func()`, this is necessary as we need to initialize `metricsServerPort`. Using the following code:

```go
flag.Uint64Var(metricsServerPort, "metrics-server-port", 0, "Starts a metrics server and sets the port it will listen to")
```

results in a panic as `flag.Uint64Var()` attempts to write to a `nil` pointer.

tests/reexecute/c/README.md
RodrigoVillar: W

tests/prometheus_server.go
RodrigoVillar: I would prefer the additional constructor for the following reasons:
1. `NewXWithY` constructors are idiomatic in Go [[ref](https://cs.opensource.google/go/go/+/master:src/net/http/request.go;l=889)]
2. The cost of having to create a new `Options` struct type along with getters and setters outweighs the benefits of having a single constructor IMO (similarly, passing in a `Config` struct type seems overkill for a single parameter (`port`))

tests/reexecute/c/vm_reexecute_test.go
RodrigoVillar: I wouldn't consider this an override, but rather a good default to address the following comment: https://github.com/ava-labs/avalanchego/pull/4415#discussion_r2433064173. Agree that this works out well though.

scripts/benchmark_cchain_range.sh
```diff
@@ -10,7 +10,7 @@ set -euo pipefail
 #   END_BLOCK: The ending block height (inclusive).
 #   LABELS (optional): Comma-separated key=value pairs for metric labels.
 #   BENCHMARK_OUTPUT_FILE (optional): If set, benchmark output is also written to this file.
-#   METRICS_SERVER_ENABLED (optional): If set, enables the metrics server.
+#   METRICS_SERVER_PORT (optional): If set, determines the port the metrics server will listen to.
```
maru-ava, line 14: I would like to see `METRICS_SERVER_PORT` be in addition to `METRICS_SERVER_ENABLED` rather than its replacement. The default port should remain zero - dynamic - and that detail shouldn't be required knowledge for those who don't want to set a specific port.

RodrigoVillar: Done: 3651f47

scripts/benchmark_cchain_range.sh
RodrigoVillar: Update: as requested in https://github.com/ava-labs/avalanchego/pull/4418#discussion_r2471540087, I've reverted the unification of the metrics server flags: 3651f47

tests/reexecute/c/vm_reexecute_test.go
aaronbuchwald: Ah I was thinking of the viper style syntax which supports differentiating if a flag was explicitly set or not.

scripts/benchmark_cchain_range.sh
aaronbuchwald: Why the preference to make this in addition instead of a single flag @maru-ava ?

Made an opposite comment here https://github.com/ava-labs/avalanchego/pull/4418#discussion_r2470145919

maru-ava: The rule of thumb here is 'Separate the Decision from the Details/Decouple Orthogonal Concerns': If two aspects of configuration answer different questions, they should be separate controls. Whether metrics export is enabled is a separable concern from configuring the port used for export. While conflating them seems like a reasonable simplification, best practice is to keep them separate for clarity and composability.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -105,6 +106,7 @@ func TestMain(m *testing.M) {
 	flag.DurationVar(&executionTimeout, "execution-timeout", 0, "Benchmark execution timeout. After this timeout has elapsed, terminate the benchmark without error. If 0, no timeout is applied.")
 
 	flag.BoolVar(&metricsServerEnabledArg, "metrics-server-enabled", false, "Whether to enable the metrics server.")
+	flag.Uint64Var(&metricsServerPortArg, "metrics-server-port", 0, "The port the metrics server will listen to.")
```
maru-ava: Given that that port numbers are 1-bit, is there a reason to prefer `Uint64Var` to  `Uint16Var`?

maru-ava, line 109: Given that that port numbers are 16 bit, is there a reason to prefer `Uint64Var` to  `Uint16Var`?

RodrigoVillar: Yes, `Uint16Var` does not exist [[ref](https://pkg.go.dev/flag#pkg-functions)]

maru-ava: Aha, I was looking at cobra. I guess that begs the question as to why we're using stdlib flag instead of cobra. The only reason we use stdlib flag in e2e testing is for compatiblity with ginkgo flag parsing, but that isn't a constraint here. In any case, not an issue for this PR.

PR #4425, opened 2025-10-17
feat(blockdb): add lru cache for block entries

DracoLi:
## Why this should be merged

I noticed that during C-Chain block execution testing, each block accepted will trigger 3 block data fetches (header, body, and receipts). This happens when collecting unflattened logs in the acceptor queue ([code](https://github.com/ava-labs/coreth/blob/master/core/blockchain.go#L617)).

Fetching block data should be fast but decompressing the data each time adds a lot of overhead. This overhead can be completely avoided with a cache because the data should have been just written to the database.

A cache was not originally added to blockdb because the VMs all have caches for blocks already. But they can be bypassed due to oversight and having a cache at the db level ensures these kind of issues are avoided.

## How this works

- Added an lru cache with configurable size (defaults to 256).
- Cache is keyed by block height and is updated on `Put` and `Get`. 
- Cache is used on `Get` and `Has` to avoid disk access.

## How this was tested

Unit tests and reexecution benchmark


x/blockdb/database.go
```diff
@@ -371,6 +376,7 @@ func (s *Database) Put(height BlockHeight, block BlockData) error {
 		)
 		return err
 	}
+	s.entryCache.Put(height, slices.Clone(block))
```
yacovm: when does this happen now? upon verify or accept? the latter, right?

x/blockdb/database.go
```diff
@@ -385,12 +391,6 @@ func (s *Database) Put(height BlockHeight, block BlockData) error {
 // It returns database.ErrNotFound if the block does not exist.
 func (s *Database) readBlockIndex(height BlockHeight) (indexEntry, error) {
 	var entry indexEntry
```
yacovm, line 389: is this moved to fix unsafe access to s.closed?

x/blockdb/database.go
```diff
@@ -176,6 +177,7 @@ type Database struct {
 	log        logging.Logger
 	closed     bool
 	fileCache  *lru.Cache[int, *os.File]
+	entryCache *lru.Cache[BlockHeight, BlockData]
```
yacovm: entry cache is a really poor name to pick for a cache that caches blocks.

yacovm: how about block cache?

x/blockdb/database.go
```diff
@@ -223,6 +225,7 @@ func New(config DatabaseConfig, log logging.Logger) (*Database, error) {
 				f.Close()
 			}
 		}),
+		entryCache: lru.NewCache[BlockHeight, BlockData](config.EntryCacheSize),
```
yacovm: this is one approach of doing it.

I think another perhaps more elegant approach would be to have a wrapper that intercepts Put and Has and Get and has the cache. The latter is much more testable as you can mock the inner backend. 

I'm fine with this approach though, just wanted to suggest another idea.

x/blockdb/entry_cache_test.go
```diff
@@ -0,0 +1,107 @@
...
+	height := uint64(30)
+	block := randomBlock(t)
+	require.NoError(t, db.Put(height, block))
+
+	has, err := db.Has(height)
+	require.NoError(t, err)
+	require.True(t, has)
+
+	// Verify block is in cache
+	cached, ok := db.entryCache.Get(height)
```
yacovm: this test doesn't really test that we don't do I/O and use the cache. 

what you can do is populate the cache artificially with a block that is not in the ledger and then Get/Has that block.

x/blockdb/cache_db_test.go
```diff
@@ -0,0 +1,107 @@
...
+	// Modify the original block after Put
+	clone[0] = 99
+
+	// Cache should have the original unmodified data
+	cached, ok := db.entryCache.Get(height)
+	require.True(t, ok)
+	require.Equal(t, block, cached)
+}
+
+func TestCacheGetReturnsClone(t *testing.T) {
```
yacovm, line 76: why are we cloning the data, can you explain? Blocks are immutable.

x/blockdb/database.go
DracoLi: yeah on accept

x/blockdb/database.go
DracoLi: I think that's a good idea. It avoids adding more logic to the `Database` and can support database with no cache easily. I'll test this out.

x/blockdb/database.go
DracoLi: I have plans to remove references to block when this package will be moved to `database.heightindexdb`, thus I didn't include block in the name here. 
The db doesn't have to just store blocks so ideally we don't reference blocks in the package. I've also considered `entryDataCache` instead of `entryCache`, thoughts?

x/blockdb/database.go
DracoLi: Yeah this is needed for the `Has` changes so we can do the cache check after closed check instead of it being part of `readBlockIndex`.
```
	if s.closed {
		s.log.Error("Failed Get: database closed", zap.Uint64("height", height))
		return nil, database.ErrClosed
	}

	if c, ok := s.entryCache.Get(height); ok {
		return slices.Clone(c), nil
	}
```
Also probably better to do the closed check on the public methods instead of a helper method.

x/blockdb/entry_cache_test.go
DracoLi: :+1: good idea

x/blockdb/cache_db_test.go
DracoLi: This is to prevent client modifications to the returned data affecting the cached data. Same for Put

This is from the interface.
```
	// Get retrieves a value by its height.
	// Returns [ErrNotFound] if the key is not present in the database.
	//
	// Returned []byte is safe to read and modify after calling Get.
```

x/blockdb/database.go
DracoLi: I tested this out but unfortunately encountered some drawbacks.
- `Put` and `Has` will have some dup logic to handle `db.closed` since it should only use cache if store is not closed. This means we will need to have dup `closeMu` and `db.closed` logic in both functions and ensure we don't have nested RLocks (we can do this by creating helpers like `getWithoutLock`).
- `New` can no longer just return `Database` if cache is used, it needs to return the interface. This makes tests more complicated as we now need to type assert if accessing private vars of Database. Or alternatively we create a separate `NewWithCache` method that returns a `cachedDatabase`. But now feels wrong since all our config is in `DatabaseConfig` but now need to have a separate initializer to enable caching. 

Given that, and since the caching logic is not too complex in Database, I think might be better to keep it as is.

x/blockdb/database.go
yacovm: > The db doesn't have to just store blocks so ideally we don't reference blocks in the package.

But you have blocks in your method signature, so like it or not, that's what you do in the code. 

Entry doesn't mean anything. You have a file cache which is clear what it is, and another entry cache which makes it oblivious what it stores.

x/blockdb/cache_db_test.go
yacovm: I get that, I am asking _why_ we're doing this.

Why would anyone using this would try to change something that is immutable?

x/blockdb/database.go
yacovm: I get the duplication and I still think it's worth it, but I'll defer to your decision 🤷

x/blockdb/cache_db_test.go
DracoLi: I see your point, at least for the current usage, block data is immutable so caller should not change it.
I'll create a issue to look into if just removing that Get behaviour from the interface. I agree that if we don't expect this usage right now, then we should just remove it to keep things simple.

DracoLi: https://github.com/ava-labs/avalanchego/issues/4453

x/blockdb/database.go
DracoLi: I'll name it to `blockCache` now. We can rename in a single PR if needed when we move the package to `database.heightindexdb`

x/blockdb/database.go
DracoLi: @yacovm tried refactoring this again - I think its a good state now. https://github.com/ava-labs/avalanchego/pull/4425/commits/20b70f1562db90ee4533d115c2dbb2373897c71b

x/blockdb/cache_db_test.go
```diff
@@ -0,0 +1,119 @@
...
+	require.True(t, ok)
+	require.Equal(t, b1, cached)
+
+	// Put second block at same height and verify it overrides the first one
+	b2 := randomBlock(t)
+	require.NoError(t, db.Put(height, b2))
+	cached, ok = db.cache.Get(height)
+	require.True(t, ok)
+	require.Equal(t, b2, cached)
+	require.NotEqual(t, b1, cached)
```
rrazvan1: Is this only for the situation when the random blocks **b1** and **b2** are equal? 😅

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	data, err := c.db.getWithoutLock(height)
+	if err != nil {
+		return nil, err
+	}
+	c.cache.Put(height, slices.Clone(data))
+	return data, nil
+}
+
+func (c *cacheDB) Put(height BlockHeight, data BlockData) error {
```
rrazvan1, line 59: Concurrent **_Puts_** can lead to data mismatch between cache and actual db.

I confirmed using the following test (I used high concurrent writes to get consistent results)

```go
func TestCacheConcurrentPuts(t *testing.T) {
	db := newCacheDatabase(t, DefaultConfig())

	height := uint64(1)
	concurrentWrites := 1000

	wg := &sync.WaitGroup{}
	for i := 0; i < concurrentWrites; i++ {
		wg.Add(1)
		go func() {
			err := db.Put(height, randomBlock(t))
			require.NoError(t, err)

			wg.Done()
		}()
	}

	wg.Wait()

	dbBlock, err := db.db.Get(height)
	require.NoError(t, err)

	cachedBlock, ok := db.cache.Get(height)
	require.True(t, ok)

	require.Equal(t, dbBlock, cachedBlock)
}
```

x/blockdb/datasplit_test.go
```diff
@@ -54,22 +53,19 @@ func TestDataSplitting(t *testing.T) {
 
 	// reopen and verify all blocks are readable
 	require.NoError(t, store.Close())
-	config = config.WithDataDir(store.config.DataDir).WithIndexDir(store.config.IndexDir)
-	store, err = New(config, store.log)
-	require.NoError(t, err)
-	store.compressor = compression.NewNoCompressor()
-	defer store.Close()
```
rrazvan1, line 61: don't we need to keep the db close deferring?

x/blockdb/helpers_test.go
```diff
@@ -11,26 +11,39 @@ import (
...
 	if config.DataDir == "" {
 		config = config.WithDataDir(dir)
 	}
 	db, err := New(config, logging.NoLog{})
-	require.NoError(t, err, "failed to create database")
+	require.NoError(t, err)
+	return db
+}
 
-	cleanup := func() {
```
rrazvan1, line 30: by removing this a lot of tests are not closing the DB anymore.

x/blockdb/cache_db_test.go
DracoLi: yeah, just to make it clear we expect cached to be different than original data due to the override. It's not testing anything different and mainly to show intention. 
Actually, I'll just remove this require, if b1 do happen to be same as b2, this would be a flake

x/blockdb/datasplit_test.go
DracoLi: no we are using tmp dir, don't really need to close it unless we need it for testing purposes

x/blockdb/helpers_test.go
DracoLi: see above, we don't really need to close it since each test is using tmp dir.

x/blockdb/cache_db.go
DracoLi: Yeah great point, this can definitely happen.
While I think blockdb should never be used for high concurrent writes at the same height, I think we can fix this easily without much downside if it makes sense.
I've tested adding sharded locking to the Put operation. Concurrent writes to the same height now properly serialize, while writes to different heights still run in parallel with minimal contention.

DracoLi: Alternatively, we can also just document that this is expected since the db is currently never expected to have concurrent writes at the same height for different data.

rrazvan1: I think we need to use the sharded locking inside **Get** aswell.

If there are 2 concurrent method calls, **Put** and **Get** (which I guess isn't something unexpected) it may happen the following scenario:

1. **Get**: get old value from DB
2. **Put**: update DB with new value
3. **Put**: update cache with new value
4. **Get**: update cache with old value

=> **DB** will have the new value, while **cache** will have the old value

DracoLi: Getting old data is fine IMO on Has and Get, we do not need to guarantee every Get is the most recent data.
I also think we are overcomplicating this for an use-case that will likely never be needed. The database should not be used in a case where we need to write different data for a height concurrently.
I think the solution is to document this behaviour where cached data can be inconsistent with underlying db data.

https://github.com/ava-labs/avalanchego/pull/4425/commits/d666deb7cd37615571567edb9c8a99f35f788086 thoughts?

rrazvan1: I was probably too picky, but you have more context about the use-cases.

Sounds reasonable, but I would be more specific: "concurrent access to the same height", bcs not only concurrent writes can lead to db-cache inconsistencies.

DracoLi: updated https://github.com/ava-labs/avalanchego/pull/4425/commits/9264d55b3b35446d9181eee7d2c37840b862c85d

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+var _ database.HeightIndex = (*cacheDB)(nil)
+
+// cacheDB caches data from the underlying [Database].
+//
+// Operations (Get, Has, Put) are not atomic with the underlying database.
+// Concurrent access to the same height can result in cache inconsistencies where
+// the cache and database contain different values. This limitation is acceptable
+// because concurrent writes to the same height are not an intended use case.
```
rrazvan1, line 23:
```suggestion
// because concurrent access to the same height is not an intended use case.
```

DracoLi: concurrent Get/Has to the same height is fine and expected. its the writes to the same height that is not intended usage and will result in access inconsistencies from cached Has and Get

rrazvan1: Then it sounds wrong to me:

```
Concurrent access to the same height can result in cache inconsistencies [...] is acceptable because concurrent writes to the same height are not an intended use case.
```

DracoLi: reverted to the original comment. https://github.com/ava-labs/avalanchego/pull/4425/commits/9792f389b9e04ed02f13290712670510e352dbc4
I think both are right depending on how you interpret it but I think we can just call out the writes.

rrazvan1 (approved):

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+
+func newCacheDB(db *Database, size uint16) *cacheDB {
+	return &cacheDB{
+		db:    db,
+		cache: lru.NewCache[BlockHeight, BlockData](int(size)),
+	}
+}
+
+func (c *cacheDB) Get(height BlockHeight) (BlockData, error) {
+	c.db.closeMu.RLock()
```
yacovm: Why are we doing this? Accessing a mutex of an inner field is usually an anti-pattern.

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+	defer c.db.closeMu.RUnlock()
+
+	if c.db.closed {
+		c.db.log.Error("Failed Get: database closed", zap.Uint64("height", height))
```
yacovm, line 44: Why not just have the ability to close this DB like we can close the inner DB?

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, database.ErrClosed
+	}
+
+	if cached, ok := c.cache.Get(height); ok {
```
yacovm, line 48: Is there a current use case in which we mutate the block data? If not, then why not just return the reference and let the consumer worry about not mutating this?

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+		return slices.Clone(cached), nil
+	}
+	data, err := c.db.getWithoutLock(height)
```
yacovm: I think it's generally not a good idea to share the lock of an inner field.

x/blockdb/cache_db.go
DracoLi: no current use case right now but I'd like to not make that change in this PR. There is already an issue https://github.com/ava-labs/avalanchego/issues/4453, we should update the interface first to remove that req then update this.

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+	c.cache.Put(height, slices.Clone(data))
```
yacovm, line 55: Is it mandatory to clone the data yet again when we insert into the cache?

yacovm: I know it's a shallow copy but still - why are we doing it?

DracoLi: Its required by the the interface spec https://github.com/ava-labs/avalanchego/blob/master/database/database.go#L105.
This is discussed and there's an issue already created. https://github.com/ava-labs/avalanchego/pull/4425#discussion_r2482602376 (i updated the issue to include Put as well)

x/blockdb/cache_db.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// Operations (Get, Has, Put) are not atomic with the underlying database.
+// Concurrent writes to the same height can result in cache inconsistencies where
+// the cache and database contain different values. This limitation is acceptable
+// because concurrent writes to the same height are not an intended use case.
+type cacheDB struct {
+	db    *Database
+	cache *lru.Cache[BlockHeight, BlockData]
+
+	closeMu sync.RWMutex
```
yacovm, line 28: do we really need a mutex here? Can't we use an atomic.Bool ?

x/blockdb/readblock_test.go
```diff
@@ -92,8 +94,7 @@ func TestReadOperations(t *testing.T) {
 				config = &defaultConfig
 			}
 
-			store, cleanup := newTestDatabase(t, *config)
```
yacovm, line 95: where did the cleanup go to?

x/blockdb/database.go
```diff
@@ -494,6 +499,15 @@ func (s *Database) Has(height BlockHeight) (bool, error) {
...
 
+	if s.closed {
+		s.log.Error("Failed Has: database closed", zap.Uint64("height", height))
+		return false, database.ErrClosed
+	}
+
+	return s.hasWithoutLock(height)
+}
+
+func (s *Database) hasWithoutLock(height BlockHeight) (bool, error) {
```
yacovm, line 506: can we rollback this change now that we don't need it?

x/blockdb/writeblock_test.go
```diff
@@ -150,9 +150,7 @@ func TestPut_MaxHeight(t *testing.T) {
 			if config.CheckpointInterval == 0 {
 				config = DefaultConfig()
 			}
-
-			store, cleanup := newTestDatabase(t, config)
-			defer cleanup()
```
yacovm, line 155: where did the cleanup go to?

x/blockdb/readblock_test.go
DracoLi: removed the cleanup function since I realized we don't need to cleanup the db in the tests whenever its created. A test can always close it via `db.Close()` if needed.

x/blockdb/cache_db.go
DracoLi: Yeah, we need the mutex. An atomic bool would only protect the close flag itself, but the RWMutex ensures that Close() waits for in-flight operations to finish before flushing the cache and closing the db, and prevents new operations from starting during close.

yacovm: I get that, I was only wondering - what's the harm in not waiting for in-flight operations since we're anyway shutting down. 

But I'll defer that to you.

x/blockdb/cache_db.go
DracoLi: agreed, updated cacheDb to manage its own mutex

PR #4430, opened 2025-10-23
feat: `emulate` package for temporary `libevm` registration

ARR4N:
## Why this should be merged

Provides a single point of entry for emulation of `coreth` and `subnet-evm` behaviour with respect to temporary `libevm` registration.

## How this works

Syntactic sugar. The changes to `go.mod` are simply due to running `go get` at the latest commits on `master.

> [!NOTE]
>  Open question: this adds `subnet-evm` to `go.mod`. Do we want to actively block its import in any other package?

## How this was tested

Basic integration test to demonstrate usage of the functions with respective repo "extra" types.

No


vms/evm/emulate/emulate_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/stretchr/testify/require"
+
+	cchain "github.com/ava-labs/coreth/plugin/evm/customtypes"
+	subnet "github.com/ava-labs/subnet-evm/plugin/evm/customtypes"
+)
+
+// setAndGetMillis is an arbitrary function that can be run if and only if
+// emulating either `coreth` or `subnet-evm`. If the resepective emulation isn't
```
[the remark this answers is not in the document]
ARR4N: Wow, AI is going to replace us!.

vms/evm/emulate/emulate.go
```diff
@@ -0,0 +1,43 @@
...
+func CChainVal[T any](fn func() (T, error)) (T, error) {
+	return val(CChain, fn)
+}
+
+// SubnetEVMVal executes `fn` as if running in a `subnet-evm` node.
+func SubnetEVMVal[T any](fn func() (T, error)) (T, error) {
+	return val(SubnetEVM, fn)
+}
+
+func val[T any](wrap func(func() error) error, fn func() (T, error)) (T, error) {
```
StephenButtolph: I find the current code pretty un-parsable. Can we use line-breaks so that each input argument is clear?
```suggestion
func val[T any](
	wrap func(func() error) error, 
	fn func() (T, error),
) (T, error) {
```

Alternatively we could make these callbacks their own types

vms/evm/emulate/emulate_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// setAndGetMillis is an arbitrary function that can be run if and only if
+// emulating either `coreth` or `subnet-evm`. If the respective emulation isn't
+// active then it will cause `libevm` to panic. In addition to the panicking
+// behaviour, it asserts that it is the only active emulation.
+func setAndGetMillis[T interface {
+	any
+	*cchain.HeaderExtra | *subnet.HeaderExtra
+}](
```
StephenButtolph: I would prefer either of these two signatures over the current code:
```golang
func setAndGetMillis[T any](
```

```golang
func setAndGetMillis[T *cchain.HeaderExtra | *subnet.HeaderExtra](
```

ARR4N: TIL that you can do the second one!

vms/evm/emulate/emulate_test.go
```diff
@@ -0,0 +1,91 @@
...
+		t.Parallel()
+		<-start
+		for range 1000 {
+			got, err := SubnetEVMVal(onSubnetEVM)
+			require.ErrorIs(t, err, sentinel)
+			require.Equal(t, milli, *got)
+		}
+	})
+
+	close(start)
```
StephenButtolph, line 87: Why does `start` exist?
```suggestion
	t.Run("coreth", func(t *testing.T) {
		t.Parallel()
		for range 1000 {
			got, err := CChainVal(onCChain)
			require.ErrorIs(t, err, sentinel)
			require.Equal(t, milli, *got)
		}
	})

	t.Run("subnet-evm", func(t *testing.T) {
		t.Parallel()
		for range 1000 {
			got, err := SubnetEVMVal(onSubnetEVM)
			require.ErrorIs(t, err, sentinel)
			require.Equal(t, milli, *got)
		}
	})
```

I suppose it will catch if someone removes the `t.Parallel()`?

vms/evm/emulate/emulate_test.go
ARR4N: Done.

vms/evm/emulate/emulate_test.go
ARR4N: > Why does start exist?

Something I do to ensure the first one doesn't finish before the second one starts.

> I suppose it will catch if someone removes the `t.Parallel()`?

True.

> "I don't care either way" -- @StephenButtolph IRL 

I'm gonna leave it then.

joshua-kim (approved):

ARR4N:
~Blocked by https://github.com/ava-labs/subnet-evm/pull/1823~

PR #4431, opened 2025-10-24
Remove unexpected workarounds of the prealloc linter

StephenButtolph:
## Why this should be merged

We enforce the `prealloc` linter, which will complain if we lazily populate a slice. However, it seems that there has been some code added that avoids the linter without actually achieving the linter's goal.

This PR removes such instances.

## How this works

Either leave slices as nil or pre-allocates the slices.

## How this was tested

I tried to find a linter / configuration that would catch this in CI... But I couldn't figure out how to do it.

I found all these instances by greping for instances that matched the `make\(.*, 0\)` regex.

No

### DIFF
--- a/cache/lru/cache_test.go
+++ b/cache/lru/cache_test.go
@@ -88,8 +88,10 @@ func TestCacheOnEvict(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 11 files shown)


maru-ava:
Can we please exclude `tests/` from this kind of linter? Code that doesn't run in production does not require this kind of rigor.

DracoLi (approved):

PR #4434, opened 2025-10-27
Remove bool return from `uptimetracker.GetUptime` 

JonathanOppenheimer:
## Why this should be merged
This cleans up the function signature for `uptimetracker.GetUptime.` It really does not make sense to have two different variables to indicate an error state, and we should return an explanatory error if the validator whose uptime is being requested. can't be found. 

## How this was tested
Existing CI 

No

### DIFF
--- a/vms/evm/uptimetracker/uptime_tracker_test.go
+++ b/vms/evm/uptimetracker/uptime_tracker_test.go
@@ -487,7 +487,7 @@ func TestUptimeTracker_GetUptime(t *testing.T) {
 					return validatorsMap, pChainHeight, nil
 				}
 
-				require.NoError(uptimeTracker.Sync(context.Background()))
+				require.NoError(uptimeTracker.Sync(t.Context()))
 			}
 
 			gotUptime, gotLastUpdated, err := uptimeTracker.GetUptime(
@@ -528,7 +528,7 @@ func TestUptimeTracker_Restart(t *testing.T) {
 	uptimeTracker, err := New(validatorState, subnetID, db, clock)
 	require.NoError(err)
 
-	require.NoError(uptimeTracker.Sync(context.Background()))
+	require.NoError(uptimeTracker.Sync(t.Context()))
 	require.NoError(uptimeTracker.Connect(nodeID))
 
 	clock.Set(start.Add(10 * time.Second))
@@ -542,7 +542,7 @@ func TestUptimeTracker_Restart(t *testing.T) {
 	clock.Set(start.Add(20 * time.Second))
 	uptimeTracker, err = New(validatorState, subnetID, db, clock)
 	require.NoError(err)
-	require.NoError(uptimeTracker.Sync(context.Background()))
+	require.NoError(uptimeTracker.Sync(t.Context()))
 
 	uptime, lastUpdated, err = uptimeTracker.GetUptime(validationID)
 	require.NoError(err)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,6 +60,7 @@ linters:
     - tagalign
     - testifylint
     - unconvert
+    - usetesting
     - unparam
     - unused
     - usestdlibvars
@@ -199,6 +200,14 @@ linters:
       disable:
         - go-require
         - float-compare
+    usetesting:
+      os-create-temp:     true # Disallow `os.CreateTemp("", ...)`
+      os-mkdir-temp:      true # Disallow `os.MkdirTemp()`
+      os-setenv:          true # Disallow `os.Setenv()`
+      os-temp-dir:        true # Disallow `os.TempDir()`
+      os-chdir:           true # Disallow `os.Chdir()`
+      context-background: true # Disallow `context.Background()`
+      context-todo:       true # Disallow `context.TODO()`
     unused:
       # Mark all struct fields that have been written to as used.
       # Default: true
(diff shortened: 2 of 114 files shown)


vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -57,27 +57,26 @@ func New(
...
-	bool,
 	error,
 ) {
 	u.lock.Lock()
 	defer u.lock.Unlock()
 
 	nodeID, ok := u.state.getNodeID(validationID)
 	if !ok {
-		return 0, time.Time{}, false, nil
+		return 0, time.Time{}, fmt.Errorf("validator %s not found", validationID)
```
joshua-kim: 1. > It really does not make sense to have two different variables to indicate an error state

    The reason I implemented it this way was because looking up the uptime and not having it I didn't consider to be an error (analagous to a k/v lookup in a map). EVM doesn't distinguish between errors/the validator not being present, so we could argue that current bool is YAGNI and go for the smaller return signature.

2. `validator %s...` is slightly misleading because this is the tx id (we refer to it as a validation id in this package, but we call it a tx id in the p-chain) corresponding to a validator, not the actual node id of the validator - so we might want to call this `validationID` to be consistent w/ this package's context.

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -33,11 +33,9 @@ func TestUptimeTracker_GetUptime(t *testing.T) {
 
 		wantUptime      time.Duration
 		wantLastUpdated time.Time
-		wantOk          bool
 	}{
 		{
-			name:   "no validators",
-			wantOk: false,
```
joshua-kim, line 40: I think to test this we'll need an assertion against the expected error type - otherwise if CI is passing on this diff it means we have a gap in coverage.

vms/evm/uptimetracker/uptime_tracker.go
JonathanOppenheimer: 1. This is a fair point, and I only thought about this more thoroughly when pulling this PR into subnet-evm, that yeah there is no distinction between it being missing and there being an error. I see the argument that this isn't really an error although I find this somewhat interesting as my mind still works in Python and looking up a key in a map (`dict()`) that doesn't exist actually IS an error - a `KeyError,` and you can use another data structure if you don't want this behavior.  
2. 👍

vms/evm/uptimetracker/uptime_tracker_test.go
JonathanOppenheimer: We can use `wantError` instead?

JonathanOppenheimer:
Whether or not you think this PR is up to you! (e.g. if you disagree, just close it)

I understand both the YAGNI argument (having literally just made a mistake because of it), and the argument that not finding the validation ID is not actually an error of `GetUptime.`

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -488,9 +493,13 @@ func TestUptimeTracker_GetUptime(t *testing.T) {
...
-			require.NoError(err)
-			require.Equal(tt.wantLastUpdated, gotLastUpdated)
-			require.Equal(tt.wantUptime, gotUptime)
+			if tt.wantError {
+				require.Error(err)
+			} else {
+				require.NoError(err)
+				require.Equal(tt.wantLastUpdated, gotLastUpdated)
+				require.Equal(tt.wantUptime, gotUptime)
+			}
```
joshua-kim: This if-else branch isn't needed - if your `wantError` is `nil`, `require.ErrorIs` handles it correctly (i.e `require.ErrorIs(err, nil)` works).

vms/evm/uptimetracker/uptime_tracker_test.go
```diff
@@ -33,9 +33,11 @@ func TestUptimeTracker_GetUptime(t *testing.T) {
 
 		wantUptime      time.Duration
 		wantLastUpdated time.Time
+		wantError       bool
```
joshua-kim: nit: assertions typically follow `want*`, and errors in go are typically named `err`, so this var is usually named `wantErr` (or `expectedErr`, but less commonly so).

vms/evm/uptimetracker/uptime_tracker.go
```diff
@@ -57,27 +60,30 @@ func New(
...
 	defer u.lock.Unlock()
 
 	nodeID, ok := u.state.getNodeID(validationID)
 	if !ok {
-		return 0, time.Time{}, false, nil
+		return 0, time.Time{}, fmt.Errorf(
+			"validationID %s: %w",
+			validationID,
+			errValidationIDNotFound,
+		)
```
joshua-kim, line 74: Small nit - this error format will spit out validation id twice (`validationID <id>: validationID not found` vs `validationID not found: <id>` )
```suggestion
		return 0, time.Time{}, fmt.Errorf(
			"%w: %s",
			errValidationIDNotFound,
			validationID,
		)
```

vms/evm/uptimetracker/uptime_tracker_test.go
JonathanOppenheimer: you taught me always want over expected!

JonathanOppenheimer:
I know we discussed, but you can still assert that the first argument takes precedence over the other - I'm just fixing your comments because you left them.

PR #4435, opened 2025-10-27
Es/taskify reexecution

Elvis339:
## Why this should be merged

The polyrepo migration requires benchmark workflows to run in any repository (Firewood, Coreth, etc.) without depending on AvalancheGo infrastructure.

Current workflows use custom actions that reference themselves:
```yaml
- uses: ./.github/actions/run-monitored-tmpnet-cmd
```

This breaks in other repos because they don't have AvalancheGo's custom actions directory. Copying actions to each repo creates duplication and maintenance burden.

This PR inlines all custom action logic directly into the workflow:
- Monitoring configuration moved into workflow steps
- Task configuration defined in JSON matrix
- Three duplicate workflows consolidated into one

The result is a self-contained trivially locally reproducable workflow with zero custom action dependencies. It can be invoked to any repository and work immediately. Execution simplifies to: `./scripts/run_task.sh <task-name>` with config from JSON.

## How this works

The workflow reads task configurations from `c-chain-reexecution-benchmark-config.json` based on the trigger event (pull_request, schedule, workflow_dispatch). It splits configurations into separate matrices for native GitHub runners and self-hosted ARC container runners to avoid compatibility issues.

On pull requests, it runs fast benchmarks (101-250k blocks) for validation. On schedule, it runs long benchmarks (33m-33m500k blocks) on performance-optimized runners. Manual dispatch accepts inputs to override the config for ad-hoc testing.

To reproduce locally just execute a task defined in Taskfile.

## How this was tested
CI

No - internal CI/CD infrastructure.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -270,8 +270,8 @@ tasks:
           TASK_NAME: '{{.TASK}}'
           START_BLOCK: 101
           END_BLOCK: 250000
-          BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-mainnet-blocks-1m-ldb/**'
-          CURRENT_STATE_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-current-state-hashdb-full-100/**'
+          BLOCK_DIR_SRC: 'cchain-mainnet-blocks-1m-ldb'
+          CURRENT_STATE_DIR_SRC: 'cchain-current-state-hashdb-full-100'
           CONFIG: default
 
   c-chain-reexecution-hashdb-archive-101-250k:
@@ -282,8 +282,8 @@ tasks:
           TASK_NAME: '{{.TASK}}'
           START_BLOCK: 101
           END_BLOCK: 250000
-          BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-mainnet-blocks-1m-ldb/**'
-          CURRENT_STATE_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-current-state-hashdb-archive-100/**'
+          BLOCK_DIR_SRC: 'cchain-mainnet-blocks-1m-ldb'
+          CURRENT_STATE_DIR_SRC: 'cchain-current-state-hashdb-archive-100'
           CONFIG: archive
 
   c-chain-reexecution-hashdb-33m-33m500k:
@@ -294,8 +294,8 @@ tasks:
           TASK_NAME: '{{.TASK}}'
           START_BLOCK: 33000001
           END_BLOCK: 33500000
-          BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-mainnet-blocks-30m-40m-ldb/**'
-          CURRENT_STATE_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-current-state-hashdb-full-33m/**'
+          BLOCK_DIR_SRC: 'cchain-mainnet-blocks-30m-40m-ldb'
+          CURRENT_STATE_DIR_SRC: 'cchain-current-state-hashdb-full-33m'
           CONFIG: default
 
   c-chain-reexecution-firewood-101-250k:
@@ -306,8 +306,8 @@ tasks:
           TASK_NAME: '{{.TASK}}'
           START_BLOCK: 101
           END_BLOCK: 250000
-          BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-mainnet-blocks-1m-ldb/**'
-          CURRENT_STATE_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-current-state-firewood-100/**'
+          BLOCK_DIR_SRC: 'cchain-mainnet-blocks-1m-ldb'
+          CURRENT_STATE_DIR_SRC: 'cchain-current-state-firewood-100'
           CONFIG: firewood
 
   c-chain-reexecution-firewood-33m-33m500k:
@@ -318,8 +318,8 @@ tasks:
           TASK_NAME: '{{.TASK}}'
           START_BLOCK: 33000001
           END_BLOCK: 33500000
-          BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-mainnet-blocks-30m-40m-ldb/**'
-          CURRENT_STATE_DIR_SRC: '{{.S3_BOOTSTRAP_TESTING_PREFIX}}/cchain-current-state-firewood-33m/**'
+          BLOCK_DIR_SRC: 'cchain-mainnet-blocks-30m-40m-ldb'
+          CURRENT_STATE_DIR_SRC: 'cchain-current-state-firewood-33m'
           CONFIG: firewood
 
   test-bootstrap-monitor-e2e:
--- a/scripts/copy_dir.sh
+++ b/scripts/copy_dir.sh
@@ -3,13 +3,17 @@
 set -euo pipefail
 
 # Usage: ./scripts/copy_dir.sh source_directory destination_directory
-# Sources can be S3 URLs (s3://bucket/path) or a local file path
+# Sources can be:
+#   - S3 URLs (s3://bucket/path)
+#   - S3 object keys (will be expanded to s3://avalanchego-bootstrap-testing/<key>/**)
+#   - Local file paths
 # Assumes s5cmd has been installed and is available in the PATH.
 # s5cmd is included in the nix dev shell.
 
 if [ $# -ne 2 ]; then
     echo "Usage: $0 <source_directory> <destination_directory>"
-    echo "Import from S3 Example: $0 's3://bucket1/path1' /dest/dir"
+    echo "Import from S3 URL Example: $0 's3://bucket1/path1' /dest/dir"
+    echo "Import from S3 object key Example: $0 'cchain-mainnet-blocks-1m-ldb' /dest/dir"
     echo "Export to S3 Example: $0 '/local/path1' 's3://bucket2/path2'"
     echo "Local Example: $0 '/local/path1' /dest/dir"
     exit 1
@@ -18,11 +22,17 @@ fi
 SRC="$1"
 DST="$2"
 
+# If SRC doesn't start with s3:// or /, assume it's an S3 object key
+if [[ "$SRC" != s3://* ]] && [[ "$SRC" != /* ]]; then
+    SRC="s3://avalanchego-bootstrap-testing/${SRC}/**"
+    echo "Expanded object key to: $SRC"
+fi
+
 # Function to copy from a single source to destination
 function copy_source() {
     local source="$1"
     local dest="$2"
-    
+
     # Check if source starts with s3://
     if [[ "$source" == s3://* || "$dest" == s3://* ]]; then
         # Use s5cmd to copy from S3
@@ -30,7 +40,7 @@ function copy_source() {
         time s5cmd cp --show-progress "$source" "$dest"
     else
         # Use cp for local filesystem with recursive support
-        
+
         # Ensure destination directory exists
         mkdir -p "$dest"
 


Taskfile.yml
```diff
@@ -255,6 +257,99 @@ tasks:
           METRICS_SERVER_ENABLED: '{{.METRICS_SERVER_ENABLED}}'
           METRICS_COLLECTOR_ENABLED: '{{.METRICS_COLLECTOR_ENABLED}}'
 
+  c-chain-reexecution-hashdb-101-250k:
+    desc: C-Chain re-execution from block 101 to 250k with hashdb
+    vars:
+      MONITORING_ENABLED: '{{if or .PROMETHEUS_URL .LOKI_URL}}true{{else}}false{{end}}'
```
maru-ava: The run-monitored-tmpnet-cmd custom action similarly configures collection by checking if a non-empty value is provided for a required variable. But it does this is to ensure that CI jobs will still be able to run when run against fork branches. Fork branches don't have access to the repo secrets that provide collector configuration (i.e. URLs, usernames, passwords).

What is proposed here appears superficially to follow that same pattern, but rather than being related to CI configuration, monitoring is forced whenever collector URLs are set. I think it would be preferable for this to be controlled directly as per the example of `TMPNET_START_METRICS_COLLECTOR`. That approach prefers an explicit flag/env var so that there is no ambiguity as to how to enable and disable collection. I don't think it should be necessary to know which of the 4 flags configuring metrics collection are also responsible for enabling and disabling that collection.

Note also that reexecute testing doesn't collect logs, so `LOKI_URL` is not relevant for configuring its monitoring.

Taskfile.yml
```diff
@@ -255,6 +257,99 @@ tasks:
...
+      TIMESTAMP: '{{now | date "20060102-150405"}}'
```
maru-ava: Maybe move these 2 steps - repeated in each task definition - to either the called task or the underlying script? The part that varies across steps appears to be `{{.TASK}}` (the name of the task).

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -1,119 +0,0 @@
...
-          echo "CURRENT_STATE_DIR_SRC=${{ inputs.current-state-dir-src }}"
-        } >> $GITHUB_ENV
-    - name: Configure AWS Credentials
-      uses: aws-actions/configure-aws-credentials@v4
-      with:
-        role-to-assume: ${{ inputs.aws-role }}
-        aws-region: ${{ inputs.aws-region }}
-        role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
-    - name: Run C-Chain Re-Execution
-      uses: ./.github/actions/run-monitored-tmpnet-cmd
```
maru-ava, line 85: When I talked about the issue with this custom action, I meant that this relative import (`./.github/...`) prevents reuse across repos and that it would be necessary to extract the necessary steps from the `run-monitored-tmpnet-cmd` and inline them in _this_ custom action. I also mentioned that `run-monitored-tmpnet-cmd` itself inlined nix installation rather than use the `install-nix` to ensure that it could be used by other repos.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -1,119 +0,0 @@
-name: 'C-Chain Re-Execution Benchmark'
```
maru-ava, line 1: My request was to make this custom action reusable for firewood CI by replacing the use of the `run-monitored-tmpnet-cmd` custom action with steps copied from it.

Taskfile.yml
```diff
@@ -255,6 +257,99 @@ tasks:
...
+      EXECUTION_DATA_DIR: '{{.EXECUTION_DATA_DIR | default (printf "/tmp/cchain-reexecution-hashdb-101-250k-%s" .TIMESTAMP)}}'
+    cmds:
+      - task: reexecute-cchain-range-with-copied-data
+        vars:
+          AWS_REGION: us-east-2
```
maru-ava: This appears to be set for some tasks and not others? `import-s3-to-dir` also sets it. Does it maybe belong in the global env for the taskfile?

Taskfile.yml
```diff
@@ -255,6 +257,99 @@ tasks:
...
+          EXECUTION_DATA_DIR: '{{.EXECUTION_DATA_DIR}}'
+          START_BLOCK: 101
+          END_BLOCK: 250000
+          BLOCK_DIR_SRC: s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**
+          CURRENT_STATE_DIR_SRC: s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100/**
+          CONFIG: default
+          METRICS_SERVER_ENABLED: '{{.MONITORING_ENABLED}}'
```
maru-ava: Is it desirable to prevent configuring these 2 variables independently?

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -0,0 +1,389 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix-native: ${{ steps.define-matrix.outputs.matrix-native }}
+      matrix-self-hosted: ${{ steps.define-matrix.outputs.matrix-self-hosted }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Define Matrix
+        id: define-matrix
+        shell: bash
+        run: |
```
maru-ava: Non-trivial bash (and this certainly qualifies) needs to be defined in its own file both to give us the benefit of proper syntax highlighting and to ensure that shellcheck is catching obvious footguns.

.github/workflows/c-chain-reexecution-benchmark.yml
```diff
@@ -0,0 +1,389 @@
...
+            nix develop --command go mod download -modfile=tools/go.mod
+          fi
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-region: us-east-2
+      - name: Show Grafana Dashboard Link
+        shell: bash
+        run: |
```
maru-ava: See: non-trivial bash

Taskfile.yml
Elvis339: That makes sense, I couldn't figure out how to infer the called task name (child) so it looks like this.

https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/Taskfile.yml#L228

https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/Taskfile.yml#L263

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Got it. Here's updated custom [c-chain-reexecution-benchmark](https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/.github/actions/c-chain-reexecution-benchmark/action.yml)

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: This should enable it.

https://github.com/ava-labs/avalanchego/tree/33c73c31620e08fe90db8aed7785677aa0b8f124/.github/actions/c-chain-reexecution-benchmark

Taskfile.yml
Elvis339: That makes sense for now.

https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/Taskfile.yml#L8

.github/workflows/c-chain-reexecution-benchmark.yml
Elvis339: Yes that makes total sense for future maintenance. 

https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/.github/workflows/c-chain-reexecution-matrix.sh

Taskfile.yml
Elvis339: I kind of copied it from run-monitored-tmpnet.

https://github.com/ava-labs/avalanchego/blob/33c73c31620e08fe90db8aed7785677aa0b8f124/.github/actions/c-chain-reexecution-benchmark/action.yml#L156

Elvis339:
> While a transition from the previous json-defined workflows might be desirable, is it required to enable running reexecute tests as part of firewood CI?
> 
> If not, maybe focus this PR only on ensuring tasks for each reexecute configuration and a reexecute custom action that is reusable across repos?

It's not required, but I think it composes together nicely and serves as verification that everything works. Here's my reasoning:
For firewood CI, we need:

- Taskfile tasks for each config (done)
- Reusable custom action (main deliverable)

The workflow refactoring adds:
- Immediate verification the custom action works
- Eliminates ~250 lines of duplication

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+    default: ${{ github.run_attempt }}
+  job:
+    description: 'Job ID for monitoring labels'
+    required: false
+    default: ${{ github.job }}
+  grafana_dashboard_id:
+    description: 'The identifier of the Grafana dashboard to use, in the format <UID>/<dashboard-name>.'
+    required: false
+    default: 'Gl1I20mnk/c-chain'
+  filter_by_owner:
```
maru-ava: This can be removed - it is only required when using multiple networks with tmpnet.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
     default: false
   push-post-state:
     description: 'S3 destination to copy the current-state directory after completing re-execution. If empty, this will be skipped.'
+    required: false
```
maru-ava: (No action required) Default values can be omitted in yaml, and a boolean field always defaults to false.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+    default: ''
+  repository_owner:
```
maru-ava: (No action required) Maybe duplicate the comment from run-monitored-tmpnet-cmd documenting that these vars never need to be provided by the caller?

maru-ava: Also, maybe use kebab-case for consistency? `run-monitored-tmpnet-cmd` is the outlier here.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+    description: 'Repository owner for monitoring labels'
```
maru-ava: Not clear why descriptions are necessary or desirable for internal-only vars.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+      with:
+        github_access_token: ${{ inputs.github-token }}
+    - run: $GITHUB_ACTION_PATH/nix-develop.sh --command echo "dependencies installed"
+      shell: bash
+    # Cache Go modules (architecture-independent)
+    - uses: actions/cache@v4
+      id: go-mod-cache
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-mod-${{ hashFiles('go.sum', 'tools/go.sum') }}
```
maru-ava: (No action required) Maybe tools/go.sum can be skipped given that reexecute doesn't depend on tools?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+      shell: bash
+      run: |
+        $GITHUB_ACTION_PATH/nix-develop.sh --impure --command bash -x ./scripts/run_task.sh ${{ inputs.task }} \
+          RUNNER_NAME="${{ inputs.runner_name }}" \
+          BENCHMARK_OUTPUT_FILE="${{ env.BENCHMARK_OUTPUT_FILE }}" \
+          EXECUTION_DATA_DIR="${{ env.EXECUTION_DATA_DIR }}"
+      env:
+        TMPNET_START_METRICS_COLLECTOR: ${{ inputs.prometheus-username != '' }}
+        TMPNET_CHECK_METRICS_COLLECTED: ${{ inputs.prometheus-username != '' }}
+        TMPNET_START_LOGS_COLLECTOR: "" # Set to empty string to disable log collection
```
maru-ava: This line can be omitted, `TMPNET_START_LOGS_COLLECTOR` defaults to false.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -59,51 +47,127 @@ inputs:
...
+      run: |
+        {
+          echo "EXECUTION_DATA_DIR=${{ inputs.workspace }}/reexecution-data"
+          echo "BENCHMARK_OUTPUT_FILE=output.txt"
+        } >> $GITHUB_ENV
+    - name: Run C-Chain Re-execution Benchmark
+      shell: bash
+      run: |
+        $GITHUB_ACTION_PATH/nix-develop.sh --impure --command bash -x ./scripts/run_task.sh ${{ inputs.task }} \
+          RUNNER_NAME="${{ inputs.runner_name }}" \
```
maru-ava: (No action required) I don't really understand why reexecute testing uses task vars rather than having the wrapper script be configured by env vars. The utility of a task definition is in enabling discoverability via a common entrypoint rather than in implementing non-trivial functionality.

.github/actions/c-chain-reexecution-benchmark/nix-develop.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
maru-ava, line 1: What's the rational for this addition?

.github/actions/c-chain-reexecution-benchmark/output-metrics-url.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
maru-ava: Maybe enforce that this file must be an exact copy of the run-monitored-tmpnet-cmd version to minimize the potential for unintended drift?

Taskfile.yml
```diff
@@ -170,10 +173,6 @@ tasks:
     desc: Runs static analysis tests of golang code
     cmd: ./scripts/lint.sh
 
-  lint-fix:
```
maru-ava, line 173: I'm assuming this deletion was unintended?

Taskfile.yml
```diff
@@ -255,6 +255,66 @@ tasks:
           METRICS_SERVER_ENABLED: '{{.METRICS_SERVER_ENABLED}}'
           METRICS_COLLECTOR_ENABLED: '{{.METRICS_COLLECTOR_ENABLED}}'
 
+  c-chain-reexecution-hashdb-101-250k:
+    desc: C-Chain re-execution from block 101 to 250k with hashdb
+    cmds:
+      - task: reexecute-cchain-range-with-copied-data
+        vars:
+          TASK_NAME: c-chain-reexecution-hashdb-101-250k
```
maru-ava: Maybe use `{{.TASK}}` as the value of `TASK_NAME` for this and all other modified tasks?

Taskfile.yml
```diff
@@ -255,6 +255,66 @@ tasks:
...
+          START_BLOCK: 101
+          END_BLOCK: 250000
+          BLOCK_DIR_SRC: s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**
```
maru-ava: (No action required) Given that the protocol, bucket and `/**` file glob are all common, maybe only require that the object path (i.e. `cchain-mainnet-blocks-1m-ldb`) be specified in this and the other tasks?

Taskfile.yml
```diff
@@ -255,6 +255,66 @@ tasks:
...
+          CURRENT_STATE_DIR_SRC: s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100/**
+          CONFIG: default
```
maru-ava, line 275: (No action required) Maybe avoid  specifying defaults like this one?

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Thanks for pointing that out. Removed.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: https://github.com/ava-labs/avalanchego/blob/2b676126837bb7aec844657ae515b5a101230ddd/.github/actions/c-chain-reexecution-benchmark/action.yml#L50

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Good point, removed.

https://github.com/ava-labs/avalanchego/blob/2b676126837bb7aec844657ae515b5a101230ddd/.github/actions/c-chain-reexecution-benchmark/action.yml#L83

https://github.com/ava-labs/avalanchego/blob/2b676126837bb7aec844657ae515b5a101230ddd/.github/actions/c-chain-reexecution-benchmark/action.yml#L89

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Good point. 

I updated `RUNNER_NAME` to be part of env in the [script](https://github.com/ava-labs/avalanchego/blob/2b676126837bb7aec844657ae515b5a101230ddd/scripts/benchmark_cchain_range.sh).

.github/actions/c-chain-reexecution-benchmark/output-metrics-url.sh
Elvis339: I added sym link.

Taskfile.yml
Elvis339: Good catch reverted. 

https://github.com/ava-labs/avalanchego/blob/2b676126837bb7aec844657ae515b5a101230ddd/Taskfile.yml#L177

Taskfile.yml
Elvis339: What's the requirement? Adding `{{ .TASK }}` as `TASK_NAME` results in parent name vs name of the executed task.

`reexecute-cchain-range-with-copied-data-20251029-130410`

Taskfile.yml
Elvis339: I think it should be part of the script and the script is made to be a bit more flexible.

https://github.com/ava-labs/avalanchego/pull/4435/commits/0635d0591a8569acf890ba7ab864fff20c8b782a

Elvis339:
> > * Taskfile tasks for each config (done)
> > * Reusable custom action (main deliverable)
> > 
> > The workflow refactoring adds:
> > 
> > * Immediate verification the custom action works
> > * Eliminates ~250 lines of duplication
> 
> I was not arguing against the potential benefit, but I'm still not convinced that this PR is the appropriate place to be making such invasive changes. The shift to a reusable action should have negligible impact on the workflows, and the shift to using tasks is also unlikely to require complicated changes to the workflows. Limiting the focus of this PR to those changes would make for a much easier PR.
> 
> Taking the initiative to reduce duplication is laudable, but I don't think such changes should have to come at the cost of making separable changes harder to review.

You're right, and I apologize for overloading this PR. I should have been more thoughtful about separating concerns.
My thinking was that since we're already touching all these workflow files to switch to the reusable action, it made sense to include the task refactoring in the same pass especially because the CI provides immediate verification that everything works end-to-end. Testing a reusable action in isolation is tricky, and having real workflows exercise it felt like the safest way to validate the changes.

Would it help if I split this into two PRs one for the task definitions + their immediate usage, and another for the reusable action refactor?

I want to make this easier to review while still getting the validation benefits of CI. What would work best for you?

.github/actions/c-chain-reexecution-benchmark/output-metrics-url.sh
maru-ava: ~A symlink won't work when the action is used from other repos.~ 

Edit: You'll need to confirm that a symlink will work when the action is used by other repos.

maru-ava:
> You're right, and I apologize for overloading this PR. I should have been more thoughtful about separating concerns. My thinking was that since we're already touching all these workflow files to switch to the reusable action, it made sense to include the task refactoring in the same pass especially because the CI provides immediate verification that everything works end-to-end. Testing a reusable action in isolation is tricky, and having real workflows exercise it felt like the safest way to validate the changes.

> 
> Would it help if I split this into two PRs one for the task definitions + their immediate usage, and another for the reusable action refactor?
> 
> I want to make this easier to review while still getting the validation benefits of CI. What would work best for you?

I would like to see this PR focus only on necessary changes - a custom action that doesn't rely on relative imports, workflows that externalize reexecute configuration to tasks. Anything more than that can be deferred to one or more subsequent PRs.

PR #4439, opened 2025-10-30
Allow new nodes to sync delinquent L1s

StephenButtolph:
## Why this should be merged

L1s that did not properly update to support the Granite activation on Fuji may have finalized blocks past the Granite activation time without epoch information.

This allows new nodes to sync off of these non-conforming validators.

## How this works

Moves epoch verification inside the bootstrapped check.

## How this was tested

Added a unit test that fails on master.

No.

### DIFF
--- a/vms/proposervm/vm_test.go
+++ b/vms/proposervm/vm_test.go
@@ -2746,6 +2746,140 @@ func TestSelectChildPChainHeight(t *testing.T) {
 	}
 }
 
+// This tests the case where a chain missed the Granite activation and continued
+// producing blocks without epochs. When a new node joins the network, it needs
+// to allow the delayed activation of Granite.
+func TestBootstrappingWithDelayedGraniteActivation(t *testing.T) {
+	require := require.New(t)
+
+	// innerVMBlks is appended to throughout the test, which modifies the
+	// behavior of coreVM.
+	innerVMBlks := []*snowmantest.Block{
+		snowmantest.Genesis,
+	}
+
+	coreVM := &blocktest.VM{
+		VM: enginetest.VM{
+			T: t,
+			InitializeF: func(context.Context, *snow.Context, database.Database, []byte, []byte, []byte, []*common.Fx, common.AppSender) error {
+				return nil
+			},
+		},
+		ParseBlockF: func(_ context.Context, blkBytes []byte) (snowman.Block, error) {
+			for _, blk := range innerVMBlks {
+				if bytes.Equal(blk.Bytes(), blkBytes) {
+					return blk, nil
+				}
+			}
+			return nil, errUnknownBlock
+		},
+		GetBlockF: func(_ context.Context, blkID ids.ID) (snowman.Block, error) {
+			for _, blk := range innerVMBlks {
+				if blk.Status == snowtest.Accepted && blk.ID() == blkID {
+					return blk, nil
+				}
+			}
+			return nil, database.ErrNotFound
+		},
+		LastAcceptedF: func(context.Context) (ids.ID, error) {
+			var (
+				lastAcceptedID     ids.ID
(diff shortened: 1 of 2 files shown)


ceyonur (approved):

vms/proposervm/vm_test.go
```diff
@@ -2746,6 +2746,140 @@ func TestSelectChildPChainHeight(t *testing.T) {
...
+	require.NoError(err)
+
+	require.NoError(block1.Verify(t.Context()))
+	require.NoError(block1.Accept(t.Context()))
+
+	// This block should include an epoch. But since the node is still
+	// bootstrapping, it should allow the block.
+	innerBlock2 := snowmantest.BuildChild(innerBlock1)
+	innerVMBlks = append(innerVMBlks, innerBlock2)
+	statelessBlock2, err := statelessblock.Build(
```
yacovm, line 2864: To play devil's advocate - I am not sure this construction properly proves it replicates the problem (it does, but it doesn't show it in the test), as we're building the block with the software version of Granite, just with an empty epoch. The only way of being convinced that the test properly replicates the problem is look at the old version of the `Build` function and the existing one and see that passing an empty epoch activates the pre Granite code path which is identical to the current one.

StephenButtolph: I suppose the alternative would be to spin up two different VMs (one running with Granite scheduled and one running with Granite unscheduled) and then showing that the node with granite scheduled can sync off of the node with granite unscheduled.

I felt like that would probably add quite a bit of code to the test, but I can do that if you feel like it would make the test better

yacovm: > I suppose the alternative would be to spin up two different VMs (one running with Granite scheduled and one running with Granite unscheduled) and then showing that the node with granite scheduled can sync off of the node with granite unscheduled.
> 
> I felt like that would probably add quite a bit of code to the test, but I can do that if you feel like it would make the test better

But both VMs would be running a software that is in version v1.14. The old software version is no longer with us. 
I think if we wish to be as pristine as possible, we should have a chain with hardcoded blocks made with Fortuna and then see that a bootstrapping proposerVM can still parse / verify them. 

Thoughts?

StephenButtolph: I feel like hardcoding the blocks is less discoverable than manually generating blocks with the prior format.

Wouldn't we essentially just be taking the current code, dumping out the blocks from the test and then embedding those blocks and removing the code that generated them?

yacovm: Yeah I think if we have code that manually generates the blocks exactly as it works in Fortuna, without calling the `Build()` function, that works best.

StephenButtolph: Just to make sure we're on the same page. That is what the current test does (perhaps confusing because `BuildUnsigned` could have alternatively been called `NewUnsigned`). This test doesn't actually call `proVM.BuildBlock`

yacovm: But we're calling [statelessblock.Build](https://github.com/ava-labs/avalanchego/commit/534f2737ea5763260d1cff444ec7554ab98df978#diff-0a19ad2bab665d2f2f629cb3a692b258012a9416ebe786715959abfd915a575dR2864-R2872)

StephenButtolph: Sure... If that function was called `statelessblock.New` (without modifying any of the behavior), would that not be doing exactly what you are suggesting?

yacovm: The function is doing what I am suggesting even now. My point is that when you read the test, you need to trust that passing an empty epoch makes the function use its pre-granite logic. Had the code of the test constructed the stateless block explicitly, it would have been more clear.

PR #4442, opened 2025-10-30
feat(reexecute/c): add firewood archival config option

RodrigoVillar:
## Why this should be merged

Now that Firewood can be run in archival mode, we can use this option in the reexecution test.

## How this works

Adds a `firewood-archive` config option for running Firewood in archive mode.

## How this was tested

Ran reexecution test with `firewood-archive` set and verified that archival mode was enabled by running `tree $CURRENT_STATE_DIR -L 3`

```
├── chain-data-dir
│   └── firewood
│       ├── firewood.db # Firewood database
│       └── root_store # Archival state directory
└── db
```

No


RodrigoVillar:
Note to reviewers: I've marked this PR as **Do Not Merge** since this will work only after the Coreth migration. However, this PR can still be reviewed since there will be 0 code changes after the Coreth migration.

tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -89,6 +89,12 @@ var (
 			"pruning-enabled": true,
 			"state-sync-enabled": false
 		}`,
+		"firewood-with-archive": `{
```
aaronbuchwald:
```suggestion
		"firewood-archive": `{
```

RodrigoVillar: Done: 6ed1b6c

PR #4443, opened 2025-10-30
chore: task reexecution

Elvis339:
## Why this should be merged
This patch is breakdown of https://github.com/ava-labs/avalanchego/pull/4435

This patch contains a custom action that doesn't rely on relative imports, workflows that externalize reexecute configuration to tasks.

### DIFF
--- a/.github/workflows/c-chain-reexecution-benchmark-container.yml
+++ b/.github/workflows/c-chain-reexecution-benchmark-container.yml
@@ -98,11 +98,11 @@ jobs:
         uses: ./.github/actions/c-chain-reexecution-benchmark
         with:
           task: ${{ matrix.task }}
-          config: ${{ matrix.config || inputs.config }}
-          start-block: ${{ matrix.start-block || inputs.start-block }}
-          end-block: ${{ matrix.end-block || inputs.end-block }}
-          block-dir-src: ${{ matrix.block-dir-src || inputs.block-dir-src }}
-          current-state-dir-src: ${{ matrix.current-state-dir-src || inputs.current-state-dir-src }}
+          config: ${{ matrix.config }}
+          start-block: ${{ matrix.start-block }}
+          end-block: ${{ matrix.end-block }}
+          block-dir-src: ${{ matrix.block-dir-src }}
+          current-state-dir-src: ${{ matrix.current-state-dir-src }}
           prometheus-url: ${{ secrets.PROMETHEUS_URL || '' }}
           prometheus-push-url: ${{ secrets.PROMETHEUS_PUSH_URL || '' }}
           prometheus-username: ${{ secrets.PROMETHEUS_USERNAME || '' }}
@@ -112,4 +112,4 @@ jobs:
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
           push-post-state: ${{ github.event.inputs.push-post-state }}
-          runner_name: ${{ matrix.runner || inputs.runner }}
+          runner_name: ${{ matrix.runner }}


scripts/benchmark_cchain_range.sh
```diff
@@ -8,18 +8,34 @@ set -euo pipefail
...
+#   METRICS_SERVER_PORT (optional): Port for the metrics server (defaults to "0").
+#   METRICS_COLLECTOR_ENABLED (optional): If set to "true", enables the metrics collector (defaults to "false").
 
 : "${BLOCK_DIR:?BLOCK_DIR must be set}"
 : "${CURRENT_STATE_DIR:?CURRENT_STATE_DIR must be set}"
-: "${RUNNER_NAME:?RUNNER_NAME must be set}"
 : "${START_BLOCK:?START_BLOCK must be set}"
 : "${END_BLOCK:?END_BLOCK must be set}"
 
+# Set defaults for optional variables
```
Elvis339: https://github.com/ava-labs/avalanchego/actions/runs/18954133887/job/54126429470#step:5:570

Added for debug capability, clear view of environment variables helps reducing the ambiguity.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +44,115 @@ inputs:
...
+        echo "EXECUTION_DATA_DIR=${EXEC_DIR}" >> "$GITHUB_ENV"
+        echo "BENCHMARK_OUTPUT_FILE=${BENCHMARK_FILE}" >> "$GITHUB_ENV"
+    - name: Run C-Chain Re-execution Benchmark
+      shell: nix develop --impure --command bash -x {0}
+      run: ./scripts/run_task.sh ${{ inputs.task }} \
+        BENCHMARK_OUTPUT_FILE="${{ env.BENCHMARK_OUTPUT_FILE }}" \
+        EXECUTION_DATA_DIR="${{ env.EXECUTION_DATA_DIR }}"
+      env:
+        RUNNER_NAME: ${{ inputs.runner_name }}
+        TMPNET_START_METRICS_COLLECTOR: ${{ inputs.prometheus-username != '' }}
```
maru-ava: It appears this this env var isn't necessary, METRICS_COLLECTOR_ENABLED replaces it.

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -100,18 +79,14 @@ jobs:
...
         with:
-          config: ${{ matrix.config }}
-          start-block: ${{ matrix.start-block }}
-          end-block: ${{ matrix.end-block }}
-          block-dir-src: ${{ matrix.block-dir-src }}
-          current-state-dir-src: ${{ matrix.current-state-dir-src }}
+          task: ${{ matrix.task }}
+          runner_name: ${{ matrix.runner }}
+          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-region: 'us-east-2'
```
maru-ava: nit: Maybe minimize the required diff by not re-ordering vars that are not changing (i.e. aws-region, aws-role, runner_name, push-github-action-benchmark)?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -100,18 +79,14 @@ jobs:
...
           prometheus-username: ${{ secrets.PROMETHEUS_USERNAME || '' }}
           prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
-          push-github-action-benchmark: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master') }}
-          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
-          aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          push-post-state: ${{ github.event.inputs.push-post-state }}
-          runner_name: ${{ matrix.runner }}
+          push-github-action-benchmark: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master') }}
+          push-post-state: ${{ github.event.inputs.push-post-state || '' }}
```
maru-ava: (No action required) Why is the value of this var changing?

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
```diff
@@ -90,18 +69,14 @@ jobs:
...
         uses: ./.github/actions/c-chain-reexecution-benchmark
         with:
-          config: ${{ matrix.config }}
-          start-block: ${{ matrix.start-block }}
-          end-block: ${{ matrix.end-block }}
-          block-dir-src: ${{ matrix.block-dir-src }}
-          current-state-dir-src: ${{ matrix.current-state-dir-src }}
+          task: ${{ matrix.task }}
+          runner_name: ${{ matrix.runner }}
+          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
```
maru-ava: nit: Same comment about avoiding non-functional reordering in this PR.

scripts/copy_dir.sh
```diff
@@ -18,19 +22,25 @@ fi
 SRC="$1"
 DST="$2"
 
+# If SRC doesn't start with s3:// or /, assume it's an S3 object key
```
maru-ava, line 21: Nice!

scripts/benchmark_cchain_range.sh
maru-ava: Is there a way to do this without setting explicit defaults? Ideally there is one source of truth for those defaults, and what is proposed here results in defaults being defined in both the script and the go entrypoint.

scripts/benchmark_cchain_range.sh
```diff
@@ -8,18 +8,34 @@ set -euo pipefail
...
 #   START_BLOCK: The starting block height (exclusive).
 #   END_BLOCK: The ending block height (inclusive).
+#   RUNNER_NAME (optional): Name of the runner (defaults to "dev").
+#   CONFIG (optional): Configuration name (defaults to "default").
 #   LABELS (optional): Comma-separated key=value pairs for metric labels.
 #   BENCHMARK_OUTPUT_FILE (optional): If set, benchmark output is also written to this file.
-#   METRICS_SERVER_ENABLED (optional): If set, enables the metrics server.
-#   METRICS_SERVER_PORT (optional): If set, determines the port the metrics server will listen to.
-#   METRICS_COLLECTOR_ENABLED (optional): If set, enables the metrics collector.
+#   METRICS_SERVER_ENABLED (optional): If set to "true", enables the metrics server (defaults to "false").
```
maru-ava: Is it necessary to require these explicit values (i.e. true, false)? Without this PR the flag parsing treats any non-empty value as true.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +44,115 @@ inputs:
...
+    default: ${{ github.workflow }}
+  run-id:
+    default: ${{ github.run_id }}
+  run-number:
+    default: ${{ github.run_number }}
+  run-attempt:
+    default: ${{ github.run_attempt }}
+  job:
+    default: ${{ github.job }}
+  grafana-dashboard-id:
```
maru-ava: Maybe just inline this? It was only necessary as a parameter to support configuring run-monitored-tmpnet-cmd. That also suggests that the dashboard ID can be similarly hard-coded in run-monitored-tmpnet-cmd since only the reexecute custom action needed to override it.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +44,115 @@ inputs:
...
+        echo "BENCHMARK_OUTPUT_FILE=${BENCHMARK_FILE}" >> "$GITHUB_ENV"
+    - name: Run C-Chain Re-execution Benchmark
+      shell: nix develop --impure --command bash -x {0}
+      run: ./scripts/run_task.sh ${{ inputs.task }} \
+        BENCHMARK_OUTPUT_FILE="${{ env.BENCHMARK_OUTPUT_FILE }}" \
+        EXECUTION_DATA_DIR="${{ env.EXECUTION_DATA_DIR }}"
+      env:
+        RUNNER_NAME: ${{ inputs.runner_name }}
+        TMPNET_START_METRICS_COLLECTOR: ${{ inputs.prometheus-username != '' }}
+        TMPNET_CHECK_METRICS_COLLECTED: ${{ inputs.prometheus-username != '' }}
```
maru-ava: It appears this env var is also not necessary - the check for collected metrics is always performed if metrics collection is enabled.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +44,115 @@ inputs:
...
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
-
-    - name: Push Post-State to S3 (if not exists)
-      if: ${{ inputs.push-post-state != '' }}
-      shell: nix develop --command bash -x {0}
-      run: ./scripts/run_task.sh export-dir-to-s3 SRC=${{ env.EXECUTION_DATA_DIR }}/current-state/ DST=${{ inputs.push-post-state }}
+    - name: Push Post-State to S3
+      if: inputs.push-post-state != ''
+      shell: nix develop --impure --command bash -x {0}
```
maru-ava, line 193: (No action required) Note that `--impure` is only required if the command needs to access env vars present in the parent shell of `nix develop`.

scripts/benchmark_cchain_range.sh
maru-ava: Instead of doing this in the script, maybe update the golang entrypoint to output a summary of its configuration after flag parsing has been performed?

scripts/benchmark_cchain_range.sh
Elvis339: It's not necessary but tough it might be valuable to have them in the script.

It's a just copy / paste of [vm_reexecute.go](https://github.com/ava-labs/avalanchego/blob/776ed1e6d1dfec082baeddb43707be4d093c4d69/tests/reexecute/c/vm_reexecute_test.go#L108).

Now when I think about it it might be messy to keep documentation up to date on both places. 

For `RUNNER_NAME`, I have to add a default of "dev" to match the Go flag default. This removes friction for local development while still allowing CI to override it.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Thanks for the explanation. Access to AWS env vars is required for this job.

scripts/benchmark_cchain_range.sh
Elvis339: Makes sense.

https://github.com/ava-labs/avalanchego/blob/10c5ef7d8c84b840479a3b2470b32ebe2bb3f290/tests/reexecute/c/vm_reexecute_test.go#L211

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Because of `|| ''` directive, I removed it.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +44,111 @@ inputs:
...
-        grafana_dashboard_id: 'Gl1I20mnk/c-chain'
-        runtime: "" # Set runtime input to empty string to disable log collection
-
+    - name: Set task env
+      shell: bash
+      run: |
+        TIMESTAMP=$(date '+%Y%m%d-%H%M%S')
+        EXEC_DIR="/tmp/reexecution-data-${TIMESTAMP}"
+        BENCHMARK_FILE="${GITHUB_WORKSPACE}/benchmark-output.txt"
+        echo "EXECUTION_DATA_DIR=${EXEC_DIR}" >> "$GITHUB_ENV"
```
maru-ava: (No action required) Is it necessary to define EXEC_DIR separately rather than inlining the definition on this line? Same question for the next line.

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: Is TMPNET_START_METRICS_COLLECTOR used by reexec testing after all?

scripts/benchmark_cchain_range.sh
```diff
@@ -16,10 +16,12 @@ set -euo pipefail
 
 : "${BLOCK_DIR:?BLOCK_DIR must be set}"
 : "${CURRENT_STATE_DIR:?CURRENT_STATE_DIR must be set}"
-: "${RUNNER_NAME:?RUNNER_NAME must be set}"
 : "${START_BLOCK:?START_BLOCK must be set}"
 : "${END_BLOCK:?END_BLOCK must be set}"
 
+# Defaults to "dev" set at vm_reexecute_test.go `runnerNameArg`
```
maru-ava: Where an argument is optional with a default as per this case, maybe avoid duplicating the default by conditionally appending the `--runner` argument only when a value is provided?

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: No strong reason to keep them separate they're only used once. Inlined for simplicity.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: It's not. I removed it and set only `METRICS_COLLECTOR_ENABLED=true` and I was able to push metrics.

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -4,26 +4,9 @@ on:
...
-        default: 250000
-      block-dir-src:
-        description: 'The source block directory. Supports S3 directory/zip and local directories.'
-        required: false
-        default: s3://avalanchego-bootstrap-testing/cchain-mainnet-blocks-1m-ldb/**
-      current-state-dir-src:
-        description: 'The current state directory. Supports S3 directory/zip and local directories.'
-        required: false
-        default: s3://avalanchego-bootstrap-testing/cchain-current-state-hashdb-full-100/**
+      task:
```
aaronbuchwald, line 7: Would it make sense to compromise between keeping CI easily reproducible with well-defined nix tasks and enabling workflow dispatch of dynamically defined re-execution jobs?

If we do this, then it would now require a code change to trigger a job over a different range or with a different input state. For example, I am working on creating a state sync'd version of the current state at different checkpoints (as opposed to the much larger current state that gets created when re-executing up to that point in time). Right now, it's very easy to kick that off by changing a single parameter, but with this change I'd need to make a PR to get a new task definition merged in, which would make for slower development.

maru-ava: Yes, having both options would make sense. So long as task-based is the default there's no reason a more granular approach can't be provided.

Taskfile.yml
```diff
@@ -201,44 +201,42 @@ tasks:
...
-      RUNNER_NAME: '{{.RUNNER_NAME | default "dev"}}'
       CONFIG: '{{.CONFIG | default ""}}'
       START_BLOCK: '{{.START_BLOCK}}'
       END_BLOCK: '{{.END_BLOCK}}'
       LABELS: '{{.LABELS | default ""}}'
       BENCHMARK_OUTPUT_FILE: '{{.BENCHMARK_OUTPUT_FILE | default ""}}'
-      METRICS_SERVER_ENABLED: '{{.METRICS_SERVER_ENABLED | default "false"}}'
-      METRICS_SERVER_PORT: '{{.METRICS_SERVER_PORT}}'
-      METRICS_COLLECTOR_ENABLED: '{{.METRICS_COLLECTOR_ENABLED | default "false"}}'
+      TIMESTAMP: '{{.TIMESTAMP | default (now | date "20060102-150405")}}'
```
aaronbuchwald, line 213: What's the significance of the default date here?

scripts/copy_dir.sh
```diff
@@ -18,19 +22,25 @@ fi
...
+if [[ "$SRC" != s3://* ]] && [[ "$SRC" != /* ]]; then
+    SRC="s3://avalanchego-bootstrap-testing/${SRC}/**"
```
aaronbuchwald: Could we define the "default bucket" at the top of the file if we are going to include a default like this here?

aaronbuchwald: What do you think of making the bucket name a separate parameter, so that the taskfile can define it once and re-use it for simpler configuration? This avoids hardcoding it in this script.

aaronbuchwald (changes requested):
Nice work!

Nice improvement to provide well-defined tasks for specific jobs in CI. We should imho continue to support running dynamically defined jobs via workflow dispatch as this has proven to be extremely helpful imo.

Taskfile.yml
maru-ava: This is just a way to define a date format for golang templating. `now` is the default date, '2006...' is just a format string.

Elvis339: And the reason it's added is to support parallel execution which would conflict otherwise because of the same directory name this task is producing.

Elvis339:
Triggered workflow with custom inputs.

```
 gh workflow run "C-Chain Re-Execution Benchmark GH Native" \       
  --ref es/task-reexecution \
  -f task="" \
  -f block-dir-src=cchain-mainnet-blocks-1m-ldb \
  -f current-state-dir-src=cchain-current-state-hashdb-full-100 \
  -f start-block=101 \
  -f end-block=250000 \
  -f config=default \
  -f runner=ubuntu-latest \
  -f timeout-minutes=360
```

Run: https://github.com/ava-labs/avalanchego/actions/runs/19149324486/job/54735171547

scripts/copy_dir.sh
Elvis339: Ah I see, copy_dir.sh could be used for things other than reexecution. That makes sense. https://github.com/ava-labs/avalanchego/pull/4443/commits/e703fe0a5bd411b5b1d01086cd84374abb0449d9 and https://github.com/ava-labs/avalanchego/pull/4443/commits/d359c7aeedbc02394db783eeb845d22c2049931b

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Done.

https://github.com/ava-labs/avalanchego/pull/4443/commits/a800ff1545c0fedb598f686d4369ef21ab788925

Taskfile.yml
```diff
@@ -201,63 +204,127 @@ tasks:
...
       START_BLOCK={{.START_BLOCK}} \
       END_BLOCK={{.END_BLOCK}} \
       LABELS={{.LABELS}} \
       BENCHMARK_OUTPUT_FILE={{.BENCHMARK_OUTPUT_FILE}} \
-      METRICS_SERVER_ENABLED={{.METRICS_SERVER_ENABLED}} \
-      METRICS_SERVER_PORT={{.METRICS_SERVER_PORT}} \
-      METRICS_COLLECTOR_ENABLED={{.METRICS_COLLECTOR_ENABLED}} \
+      EXECUTION_DATA_DIR={{.EXECUTION_DATA_DIR}} \
       bash -x ./scripts/benchmark_cchain_range.sh
+    # Runtime context variables are read from environment by the script:
```
maru-ava, line 225: (No action required) Not entirely sure why these details need to be embedded in the taskfile. Or, to be honest, why this task makes sense at all. It might be cleaner to only add tasks for pre-defined test configurations and leave ad-hoc usage to direct script invocation.

maru-ava (approved):

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +60,139 @@ inputs:
...
+  run-attempt:
+    default: ${{ github.run_attempt }}
+  job:
+    default: ${{ github.job }}
 
 runs:
   using: composite
   steps:
-    - name: Set task env
+    - uses: cachix/install-nix-action@02a151ada4993995686f9ed4f1be7cfbb229e56f #v31
```
aaronbuchwald, line 87: So we are duplicating the code from run monitored tmpnet command here to avoid the relative import? Is there no better way to share code across these two actions than to duplicate the shared code?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +60,139 @@ inputs:
...
+        GH_RUN_ID: ${{ inputs.run_id }}
+        GH_RUN_NUMBER: ${{ inputs.run_number }}
+        GH_RUN_ATTEMPT: ${{ inputs.run_attempt }}
+        GH_JOB_ID: ${{ inputs.job }}
     - name: Compare Benchmark Results
       uses: benchmark-action/github-action-benchmark@v1
       with:
         tool: 'go'
-        output-file-path: ${{ env.BENCHMARK_OUTPUT_FILE }}
+        output-file-path: benchmark-output.txt
```
aaronbuchwald: What's the reason for this change?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +60,139 @@ inputs:
...
-        prometheus_password: ${{ inputs.prometheus-password }}
-        grafana_dashboard_id: 'Gl1I20mnk/c-chain'
-        runtime: "" # Set runtime input to empty string to disable log collection
-
+            BLOCK_DIR_SRC=${{ inputs.block-dir-src }} \
+            CURRENT_STATE_DIR_SRC=${{ inputs.current-state-dir-src }} \
+            START_BLOCK=${{ inputs.start-block }} \
+            END_BLOCK=${{ inputs.end-block }} \
+            BENCHMARK_OUTPUT_FILE="${{ env.BENCHMARK_OUTPUT_FILE }}" \
+            TASK_NAME="custom-reexecution"
```
aaronbuchwald: Why do we need a task name here?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,64 +60,139 @@ inputs:
...
+        fi
+      env:
```
aaronbuchwald, line 170: Why were some env vars removed in this change? It seems like metrics server enabled and labels were dropped here

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -6,34 +6,37 @@ on:
...
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
-        required: false
         default: ubuntu-latest
+        type: choice
+        options:
+          - ubuntu-latest
+          - blacksmith-4vcpu-ubuntu-2404
+          - avalanche-avalanchego-runner-2ti
+          - avago-runner-m6i-4xlarge-ebs-fast
+          - avago-runner-i4i-4xlarge-local-ssd
```
aaronbuchwald: Can we remove this restriction for now?

We are still adding new labels behind ARC on a somewhat regular basis as we test different instance / storage types and this would require merging a code change to test those new labels, whereas right now we can easily trigger manual workflows.

aaronbuchwald:
> Triggered workflow with custom inputs.
> 
> ```
>  gh workflow run "C-Chain Re-Execution Benchmark GH Native" \       
>   --ref es/task-reexecution \
>   -f task="" \
>   -f block-dir-src=cchain-mainnet-blocks-1m-ldb \
>   -f current-state-dir-src=cchain-current-state-hashdb-full-100 \
>   -f start-block=101 \
>   -f end-block=250000 \
>   -f config=default \
>   -f runner=ubuntu-latest \
>   -f timeout-minutes=360
> ```
> 
> Run: https://github.com/ava-labs/avalanchego/actions/runs/19149324486/job/54735171547

Could we update the re-execution README to make sure that instructions to run this via the GH CLI are up to date?

Also, does this support both running a named task and granularly defined task via GH CLI?

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: Same comment for the other file that mirrors this one.

aaronbuchwald (changes requested):

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: The alternative is duplicating the custom actions into each repo's `.github/actions/` directory. Inlining avoids that while giving us:
- Local reproducibility: `./scripts/run_task.sh <task-name>` just works
- Zero cross-repo dependencies 
- Self-documenting workflows

The logic seems simple enough that duplication beats dependency management here, especially during the polyrepo migration where independence is valuable.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Without `TASK_NAME`, parallel local execution of c-chain-reexecution-* tasks would conflict they'd all try to create file with the same name. `TASK_NAME` ensures each task gets unique context rich identifiers.

This isn't required for CI because each matrix job runs on a separate runner with isolated resources, but it's required for local development where you might run multiple tasks simultaneously on the same machine.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: In [current master](https://github.com/ava-labs/avalanchego/blob/29b4e6bc541bcd913da6c51c1c5403fff2e377a4/.github/actions/c-chain-reexecution-benchmark/action.yml#L95) labels were never set, not even in the [parent action](https://github.com/ava-labs/avalanchego/blob/master/.github/actions/run-monitored-tmpnet-cmd/action.yml) for example we can see labels are empty on this [run](https://github.com/ava-labs/avalanchego/actions/runs/19275195679/job/55113180235#step:3:835).

Regarding metrics server enabled we consolidated it making it implicitly controlled by whether the collector is enabled.

The `benchmark_cchain_range.sh` script reads these from the environment using the `${VAR:+--flag}` pattern, so they're still available just passed through environment variables instead of explicit task parameters. This makes the workflow cleaner since monitoring config is grouped in the `env:` section.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Valid, done.

https://github.com/ava-labs/avalanchego/pull/4443/commits/3483737ff9af4ab0287822da1c0e51a5a403d6f8

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Good catch I used it for testing relative paths on self-hosted environments.

Reverted this change: https://github.com/ava-labs/avalanchego/pull/4443/commits/0cb04dc9def691259e08d66b93bff267985f74fe

.github/workflows/c-chain-reexecution-benchmark-container.yml
maru-ava: I think we should have predefined options to minimize the expertise required, and that this shouldn't preclude a manual override.

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: Yeah, unfortunately there is no way to share code between actions intended for reuse across repos. `uses` requires either a relative import - which precludes use in other repos without copying the target action to the same relative path in other repos - or an absolute versioned import (i.e. github.com/my/repo/my/action@version) which only makes sense for custom actions in other repos.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Yes, having override option makes sense for both use-cases.
https://github.com/ava-labs/avalanchego/pull/4443/commits/4615490f680f6a39ded6c7be65faccbb0642f253

maru-ava (approved):

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -100,11 +110,12 @@ jobs:
       - name: Run C-Chain Re-Execution Benchmark
         uses: ./.github/actions/c-chain-reexecution-benchmark
         with:
-          config: ${{ matrix.config }}
-          start-block: ${{ matrix.start-block }}
-          end-block: ${{ matrix.end-block }}
-          block-dir-src: ${{ matrix.block-dir-src }}
-          current-state-dir-src: ${{ matrix.current-state-dir-src }}
+          task: ${{ matrix.task }}
```
aaronbuchwald, line 100: Since we use `matrix.task` here and do not push the dynamic params into the matrix, doesn't this also mean that we can no longer run dynamic task definitions from the JSON files?

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: imho this is not worth the complexity.

Supporting dynamic configuration means that you can mess it up, so we don't actually avoid misconfiguration and we now need to handle the switchover between the two. Could we revert this?

aaronbuchwald (changes requested):

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: If you mean to update [JSON file](https://github.com/ava-labs/avalanchego/blob/c8b4a5ae0450a26f1bb7d516ee945762d8b1941b/.github/workflows/c-chain-reexecution-benchmark-container.json) with dynamic inputs (runner, start, end block...) then no, you'd have to create a task. 

What's the need for this when we support dynamic dispatch assuming I understood the question.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: The misconfiguration can happen either way imo. Even if we just had a single input field, someone can still type a runner name that doesn't exist. And runner is just a label for job scheduling if it's wrong, the job fails to start immediately. It's not affecting any logic or causing partial failures.

The main reason for the presets is discoverability. People can see the 4 approved runners right there in the dropdown instead of having to dig through docs or ask around. The custom field is just for edge cases like testing new ARC labels without having to merge a PR first which is literally the problem you mentioned before.

The bash code is like 6 lines and runs once when the workflow starts. It's not affecting logic therefore it's not on-going complexity.

What's your main concern here is it the bash logic itself or something about controlling which runners get used for benchmarks?

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: When creating a new batch of jobs, it's often easier to trigger them all at once via a PR, so that they are all organized side by side. Would it be possible to continue to support dynamic configurations via the JSON file?

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: My concern here is that it adds complexity at minimal benefit (avoid shooting yourself in the foot, but doesn't actually prevent it). The dropdown is a nice feature within the UI, but I was thinking more from the perspective of running them via the GH CLI or pushing code as I use that flow more frequently and the error case for that flow stays the same with this change. That's why from my perspective, this adds complexity (not just the bash, but combining two inputs to configure the runner type) and doesn't solve a real problem.

I always copy paste, so have not had any issues with using an incorrect label. Discoverability is a good point for new users though. I'd prefer to remove this and document the available runners, but feel free to ignore and resolve this comment if you prefer.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: You should be able to use dynamic configurations, the workflow supports both approaches.

You could have a JSON config like:

```json
{
  "include": [
    {
      "runner": "ubuntu-latest",
      "start-block": "500", 
      "end-block": "1000",
      "block-dir-src": "custom-blocks",
      "current-state-dir-src": "custom-state"
    }
  ]
}
```

The current JSON files just happen to use the task-based approach, but nothing stops you from mixing both.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: You are frequent user and code owner of this workflow so I decided to revert. 

https://github.com/ava-labs/avalanchego/pull/4443/commits/fa2ebb55d8c68309c29ea2d67aaba15faee586dc

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: I don't think this would work for multiple entries though because it uses `matrix.task` (which allows it to pull from multiple entries), but then uses `inputs.XXX` for the remainder of the parameters.

There is only one set of inputs and the intermediate step to write them into multiple matrix entries is how this was supported before. Am I misunderstanding something here about how this would work after this PR?

Elvis339: I see, thanks for explaining. 

When I look at it, both the JSON configs and the Taskfile tasks are doing the same thing defining preset parameter combinations.

The JSON has:
```json
{
  "runner": "ubuntu-latest", 
  "task": "c-chain-reexecution-hashdb-101-250k"
}
```

And the Taskfile has:
```yaml
c-chain-reexecution-hashdb-101-250k:
  START_BLOCK: 101
  END_BLOCK: 250000
  BLOCK_DIR_SRC: 'cchain-mainnet-blocks-1m-ldb'
```

So when would you actually need granular parameters in the JSON when the task already defines everything? 

If you want different parameter combinations, you can just define more tasks (and the workflow is same, update the code -> open PR)

What scenario are you thinking of where you'd need this?

aaronbuchwald: This would be to keep the flow of triggering a new batch of tasks outside of the Taskfile as simple as possible and keep it fully featured from just the JSON.

You can do it from the Taskfile, but I personally find it easier to trigger a batch by updating the JSON file only, so would prefer to keep support for that flow rather than forcing use of the taskfile.

All we'd need to do to support this, is to revert back to parsing the JSON / inputs as we did before into a matrix and then using the matrix entries here.

Elvis339: To get things moving, here's the implementation that supports [dynamic configurations via JSON](https://github.com/ava-labs/avalanchego/pull/4443/commits/171634f6c65da905859e46572e5e7f9531351bb3).

Here's a concrete run showcasing that it works: https://github.com/ava-labs/avalanchego/actions/runs/19716541948/job/56489942376

aaronbuchwald (approved):

aaronbuchwald:
It seems the task described in the README to test Firewood fails: https://github.com/ava-labs/avalanchego/actions/runs/19719001624/job/56497689011#step:4:899

This may be a result of a faulty current state db pushed to the s3 bucket for Firewood. Since Firewood re-execution tests are not currently reliable due to https://github.com/ava-labs/coreth/issues/1158, it may be better to update the README to use a hashdb example that passes, but don't think that needs to block this PR.

PR #4450, opened 2025-10-31
chore(customrawdb): unexport `customrawdb.NewSyncPerformedIterator`

powerslider:
## Why this should be merged

- This function is no longer needed by `coreth` so it can be unexported.

## How this works

- Rename `NewSyncPerformedIterator ` to `newSyncPerformedIterator `

## How this was tested

existing UT

no

### DIFF
--- a/vms/evm/sync/customrawdb/sync_progress.go
+++ b/vms/evm/sync/customrawdb/sync_progress.go
@@ -180,20 +180,14 @@ func WriteSyncPerformed(db ethdb.KeyValueWriter, blockNumber uint64) error {
 	return db.Put(bytes, nil)
 }
 
-// NewSyncPerformedIterator returns an iterator over all block numbers the VM
-// has state synced to.
-func NewSyncPerformedIterator(db ethdb.Iteratee) ethdb.Iterator {
-	return rawdb.NewKeyLengthIterator(db.NewIterator(syncPerformedPrefix, nil), syncPerformedKeyLength)
-}
-
 // GetLatestSyncPerformed returns the latest block number state synced performed to.
 func GetLatestSyncPerformed(db ethdb.Iteratee) (uint64, error) {
-	it := NewSyncPerformedIterator(db)
+	it := newSyncPerformedIterator(db)
 	defer it.Release()
 
 	var latestSyncPerformed uint64
 	for it.Next() {
-		syncPerformed := ParseSyncPerformedKey(it.Key())
+		syncPerformed := parseSyncPerformedKey(it.Key())
 		if syncPerformed > latestSyncPerformed {
 			latestSyncPerformed = syncPerformed
 		}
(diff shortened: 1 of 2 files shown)


vms/evm/sync/customrawdb/sync_progress.go
```diff
@@ -231,10 +231,10 @@ func clearPrefix(db ethdb.KeyValueStore, prefix []byte, keyLen int) error {
 	return batch.Write()
 }
 
-// parseSyncPerformedKey returns the block number from keys the iterator returned
+// ParseSyncPerformedKey returns the block number from keys the iterator returned
```
alarso16: This is not related, but this comment is confusing to me. Any chance you could fix this first sentence to make sense? I've read it 10 times and I have no idea what it means

alarso16 (approved):

StephenButtolph:
Where is this needed in coreth?

powerslider:
> Where is this needed in coreth?

It is needed here https://github.com/ava-labs/coreth/blob/master/plugin/evm/vmtest/test_syncervm.go#L645, but instead of `UnpackSyncPerfomedKey` it is now called `ParseSyncPerformedKey`.

StephenButtolph:
I feel like that test is very weird no? Why would that package test the internals of this package? Shouldn't it just assert the exported behavior of this package?

powerslider:
> I feel like that test is very weird no? Why would that package test the internals of this package? Shouldn't it just assert the exported behavior of this package?

True, it is used more in the sense of an additional assertion step [here](https://github.com/ava-labs/coreth/blob/master/plugin/evm/vmtest/test_syncervm.go#L481) and [here](https://github.com/ava-labs/coreth/blob/master/plugin/evm/vmtest/test_syncervm.go#L492) when testing the VM behavior so it is not in the scope of a unit test for this particular package. I personally don't mind removing it. Wdyt @StephenButtolph ?

StephenButtolph:
> True, it is used more in the sense of an additional assertion step [here](https://github.com/ava-labs/coreth/blob/master/plugin/evm/vmtest/test_syncervm.go#L481) and [here](https://github.com/ava-labs/coreth/blob/master/plugin/evm/vmtest/test_syncervm.go#L492) when testing the VM behavior so it is not in the scope of a unit test for this particular package. I personally don't mind removing it. Wdyt @StephenButtolph ?

Looking more closely, it seems like we should either expose both `NewSyncPerformedIterator` & `parseSyncPerformedKey` or neither.

I generally feel like reducing cross-package dependencies helps us more in the long-run...

Looking more closely at coreth, it seems that we currently use these to assert:
1. That the set is empty
2. The the set has a single entry

Couldn't we just use `GetLatestSyncPerformed` for these tests?
1. Verify that the result is `0`
2. Verify that the result is equal to the provided entry

If we are okay with changing those tests to use `GetLatestSyncPerformed`, can we actually change this PR to unexport `NewSyncPerformedIterator`?

powerslider:
> If we are okay with changing those tests to use GetLatestSyncPerformed, can we actually change this PR to unexport NewSyncPerformedIterator?

Yes, absolutely! Now that you've mentioned it, I never actually realized that `GetLatestSyncPerformed` was doing the same thing as those two functions which is odd why it wasn't used in the first place. 

In that case I'll unexport `NewSyncPerformedIterator` and I'll finish off the fix in coreth. Thanks!

PR #4451, opened 2025-10-31
refactor(reexcute/c): generic VM reexecution

RodrigoVillar:
## Why this should be merged

As part of #4108, this PR allows for generic VM reexecution by refactoring the `reexecute/c` package to only include C-Chain related logic, and migrates all other code to the `reexecute` package.

An example of how other VMs can utilize this refactored code for their reexecution tests is #4458, which adds reexecution tests for the P-Chain.

## How this works

- Uplifts the `VMExecutor` and extends it by integrating the block channel and consensus metrics into it.
- Adds a `NewMainnetVM()` constructor to facilitate creating/initializing mainnet VMs (such as the EVM and PlatformVM).
- Uplifts `StartServer()` and `StartCollector()`
- Uplifts `ExecuteBlockRange()`

## How this was tested

CI

No


tests/reexecute/vm_executor.go
```diff
@@ -0,0 +1,209 @@
...
+	etaTracker       *timer.EtaTracker
+}
+
+func newVMExecutor(
+	log logging.Logger,
+	vm block.ChainVM,
+	registry prometheus.Registerer,
+	executionTimeout time.Duration,
+	startBlock uint64,
+	endBlock uint64,
```
RodrigoVillar: I removed the `vmExecutorConfig` type as I felt that it's fine for `newVMExecutor()` to just expand its function parameters given that `BenchmarkExecutor` is its only caller.

RodrigoVillar:
@copilot Can you do a code review similar to the review that you would do if I marked this PR as ready for review?

Update: this failed miserably.

RodrigoVillar:
Example of metrics working from the latest CI run ([ref](https://github.com/ava-labs/avalanchego/actions/runs/19054254116/job/54421122410?pr=4451)): 
Screenshot 2025-11-04 at 08 20 48

tests/reexecute/vm_executor.go
```diff
@@ -0,0 +1,359 @@
...
+func blockKey(height uint64) []byte {
+	return binary.BigEndian.AppendUint64(nil, height)
+}
+
+type consensusMetrics struct {
+	lastAcceptedHeight prometheus.Gauge
+}
+
+// newConsensusMetrics creates a subset of the metrics from snowman consensus
+// [engine](../../snow/engine/snowman/metrics.go).
```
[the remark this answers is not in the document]
RodrigoVillar: W callout, fixed: ee16107

tests/reexecute/vm_executor.go
```diff
@@ -0,0 +1,359 @@
...
+//
+// The registry passed in is expected to be registered with the prefix
+// "avalanche_snowman" and the chain label (ex. chain="C") that would be handled
+// by the[chain manager](../../../chains/manager.go).
```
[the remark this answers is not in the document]
RodrigoVillar: ee16107

tests/reexecute/vm_executor.go
```diff
@@ -0,0 +1,359 @@
...
+
+	sharedMemoryDB := prefixdb.New([]byte("sharedmemory"), db)
+	atomicMemory := atomic.NewMemory(sharedMemoryDB)
+
+	chainIDToSubnetID := map[ids.ID]ids.ID{
+		mainnetXChainID:  constants.PrimaryNetworkID,
+		MainnetCChainID:  constants.PrimaryNetworkID,
+		vmParams.ChainID: vmParams.SubnetID,
+		ids.Empty:        constants.PrimaryNetworkID,
+	}
```
aaronbuchwald: It feels a little bit odd to me to mix the params that are hardcoded vs. supplied in this way. I would prefer to make this a parameter as well and provide some helper to create the correct value of it based on the networkID, but this is a good first step imo.

tests/reexecute/export.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/database/leveldb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/units"
+)
+
+// ExportBlockRange copies blocks from a source LevelDB directory to a
+// destination LevelDB directory for the specified block range [startBlock, endBlock].
+func ExportBlockRange(tb testing.TB, blockDirSrc string, blockDirDst string, startBlock, endBlock uint64, chanSize int) {
```
aaronbuchwald: Could we decouple this from `testing.TB` and just return an error if it fails?

aaronbuchwald: I think this should also be organized alongside the definition of `blockKey` so that the definition of the blockdb we are using here is in one clear place.

aaronbuchwald (approved):

tests/reexecute/vm_executor.go
RodrigoVillar: Hmm good point. Could I ask why `ids.Empty` was originally mapped to the primary network? In the case of the C-Chain reexecution tests, the `mainnetXChainID` and `MainnetCChainID` mappings make sense (cross chain transfers) but `ids.Empty` seems odd.

tests/reexecute/export.go
RodrigoVillar: Moved `ExportBlockRange()` to `vm_executor.go` so that it's close to `blockKey()` here: 3b45a8d

RodrigoVillar: > Could we decouple this from testing.TB and just return an error if it fails?

We currently use `testing.TB` to attach a cleanup function to close the db in `createBlockChanFromLevelDB()`: https://github.com/ava-labs/avalanchego/blob/3b45a8d4fb4f3f82856c216a2e48fae5aea8fdec/tests/reexecute/vm_executor.go#L310-L312

I could get rid of the cleanup function and add a `defer` statement to close the DB in the goroutine here: https://github.com/ava-labs/avalanchego/blob/3b45a8d4fb4f3f82856c216a2e48fae5aea8fdec/tests/reexecute/vm_executor.go#L314-L316

aaronbuchwald: Probably best to ensure the DB gets closed so that this doesn't produce issues when working with long-running jobs. We don't currently wait for that goroutine, so fine to keep imo and defer any change here to separate work.

tests/reexecute/vm_executor.go
aaronbuchwald: `ids.Empty` is the P-Chain's blockchainID as well as the subnetID for the primary network.

Elvis339:
Quality work on this. 

I've been looking at the C-chain and P-chain test files, and I'm noticing they're mostly identical. Both have the same flag definitions, similar TestMain setup, identical metrics server/collector logic, and the same database initialization pattern. The main differences are really just:
- Which VM factory to instantiate
- How to extract genesis bytes
- Metrics prefix (avalanche_evm vs avalanche_p)
- C-chain's predefined configs vs P-chain not having any
- Different Grafana dashboard paths

I wonder if could we can reduce the boilerplate? Something like having a single `tests/reexecute/benchmark_test.go` that accepts a `--chain` flag, and then dispatch to chain-specific configuration objects?

I'm imagining something like:
```
type ChainConfig interface {
    CreateVMFactory() vms.Factory
    GetVMParams(configBytes []byte) VMParams
    MetricsPrefix() string
    DashboardPath() string
    DefaultLabels(networkUUID string) map[string]string
    // Chain-specific metrics reporting
    ReportMetrics(b *testing.B, gatherer prometheus.Gatherer, elapsed time.Duration)
}
```

A few questions I'm curious about:
- What's the reasoning for keeping separate test files? Is there an anticipated need for chains to diverge significantly in their test structure? Or is this more about package organization?
- How do you see this evolving when we add more chains? If someone wants to add X-chain or subnet VM reexecution, would they copy-paste another test file, or is there a better path?
- Config handling - C-chain has this nice system with predefined configs (default/archive/firewood). Should this be part of the abstraction, or intentionally kept C-chain specific? What happens if P-chain or other chains need similar configurability?

We could eliminate duplication per chain by having one test file, and make adding new chains as simple as implementing the interface and registering it. But I'm also aware there might be reasons to keep things separate.

What do you think?

tests/reexecute/export.go
RodrigoVillar: Ack.

tests/reexecute/vm_executor.go
RodrigoVillar: Ah gotcha, I made `chainIDToSubnetID` a parameter in `VMConfig` here: ced37fe. I think we could add a `NewChainToSubnet()` helper function in the PR that adds the P-Chain.

RodrigoVillar:
Going to revisit this after #4508 is complete.

joshua-kim:
Closing due to staleness. If this was closed in error please re-open.

PR #4454, opened 2025-11-01
Implement RTT measurement

yacovm:
This commit adds measurement of round trip time using the ping pong messages.

## Why this should be merged

This commit lets the administrators of a node monitor the network round trip time of all peers the network is connected to.
In case the network of the node experiences delay, it will be reflected in this metric and can be used to explain phenomenon such as block acceptance taking too long.

## How this works

The Round-Trip-Time is measured by measuring the time between the ping pong messages. When a ping is sent to a peer `p`, the time is sampled. When a pong is received, the elapsed time is measured and the round_trip_count and round_trip_sum counter and gauge are emitted.

## How this was tested

I cherry picked this commit on a Fuji 1.14 node I have deployed a week ago, and re-compiled the node and ran it.

I created in Grafana the dashboard below using the command: `increase(avalanche_network_round_trip_sum[5m]) / increase(avalanche_network_round_trip_count[5m])` 

image

No

### DIFF
--- a/network/peer/peer.go
+++ b/network/peer/peer.go
@@ -616,6 +616,10 @@ func (p *peer) writeMessage(writer io.Writer, msg message.OutboundMessage) {
 		return
 	}
 
+	if msg.Op() == message.PingOp {
+		atomic.StoreInt64(&p.lastPingSent, p.Clock.Time().UnixMilli())
+	}
+
 	// Write the message
 	var buf net.Buffers = [][]byte{msgLenBytes[:], msgBytes}
 	if _, err := io.CopyN(writer, &buf, int64(wrappers.IntLen+msgLen)); err != nil {
@@ -629,10 +633,6 @@ func (p *peer) writeMessage(writer io.Writer, msg message.OutboundMessage) {
 	now := p.Clock.Time()
 	p.storeLastSent(now)
 	p.Metrics.Sent(msg)
-
-	if msg.Op() == message.PingOp {
-		atomic.StoreInt64(&p.lastPingSent, now.UnixMilli())
-	}
 }
 
 func (p *peer) sendNetworkMessages() {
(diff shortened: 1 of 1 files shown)


network/peer/peer.go
```diff
@@ -819,7 +826,17 @@ func (p *peer) getUptime() uint32 {
...
 }
 
-func (*peer) handlePong(*p2p.Pong) {}
+func (p *peer) handlePong(*p2p.Pong) {
+	lastPing := atomic.LoadInt64(&p.lastPingSent)
+	if lastPing == 0 {
+		return
+	}
+
+	elapsed := p.Clock.Time().UnixMilli() - lastPing
```
StephenButtolph: Seems like we might as well enforce that we actually sent a ping.
```suggestion
	pingSent := atomic.SwapInt64(&p.lastPingSent, 0)
	if pingSent == 0 {
		p.Log.Debug(malformedMessageLog,
			zap.Stringer("nodeID", p.id),
			zap.Stringer("messageOp", message.PongOp),
			zap.String("reason", "received unexpected pong"),
		)
		p.StartClose()
		return
	}

	elapsed := p.Clock.Time().UnixMilli() - pingSent
```

network/peer/peer.go
```diff
@@ -626,6 +629,10 @@ func (p *peer) writeMessage(writer io.Writer, msg message.OutboundMessage) {
 	now := p.Clock.Time()
 	p.storeLastSent(now)
 	p.Metrics.Sent(msg)
+
+	if msg.Op() == message.PingOp {
+		atomic.StoreInt64(&p.lastPingSent, now.UnixMilli())
+	}
```
StephenButtolph: Couple questions for this:
1. Is it a problem that we are setting this variable _after_ we have sent the message? Technically seems racy between handling the pong and setting this ping time here.
2. Would it make more sense to set this where we send the ping message (`peer.sendNetworkMessages`) rather than here? That would avoid needing to actually inspect the op type.

yacovm: I thought about (1) but we don't send the ping there, we just enqueue it to be being sent. It would be inaccurate because if we have some number of fat messages enqueued before us, it would skew the statistics. 

I moved the code to right before we write the message down the wire.

PR #4456, opened 2025-11-03
Skip flaky tests

StephenButtolph:
## Why this should be merged

I added a `t.Skip("FLAKY: ${REASON}")` to all the merkle/sync tests that have been flaking recently. Interestingly, all of them are flaking due to timeouts.

A flaky test is worse than no test. It decreases developer velocity (people just re-run failed tests). While it's possible that these tests could still catch persistently failing changes, neither merkledb nor sync are currently in production.

## How this works

Skip the flaky tests.

## How this was tested

N/A

No.


JonathanOppenheimer:
Should an issue be created to track fixing these flaky tests?

alarso16 (approved):

emilypi-avalabs:
Agree with @JonathanOppenheimer here - if they're flaky, eventually we want them resolved when there's time for a deeper dive

PR #4459, opened 2025-11-04
Update versions for v1.14.0

StephenButtolph:
## Why this should be merged

Fuji went well 🥳 

## How this works

Schedules Granite for 11AM ET on November 19th.

## How this was tested

Fuji Activation

Yes.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -13,11 +13,11 @@ The ACPs in this upgrade go into effect at 11 AM ET (4 PM UTC) on Wednesday, Nov
 
 **All Mainnet nodes must upgrade before 11 AM ET, November 19th 2025.**
 
+The plugin version is updated to `44`; all plugins must update to be compatible.
+
 ### LibEVM Hook Registration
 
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -2,9 +2,40 @@
...
+
+The ACPs in this upgrade go into effect at 11 AM ET (4 PM UTC) on Wednesday, November 19th, 2025 on Mainnet.
+
+**All Mainnet nodes must upgrade before 11 AM ET, November 19th 2025.**
+
+### LibEVM Hook Registration
+
+This release includes changes for how EVM modifications are registered with libevm. For consumers of avalanchego as a library, manual registration of libevm callbacks is now required.
+
+The plugin version is updated to `44` all plugins must update to be compatible.
```
joshua-kim: Is this supposed to be in the above header instead of under the `LibEVM...` one?

Also... I just realized this but I think this template for `The plugin version is updated to `x` all plugins must update to be compatible.` has weird grammar... i feel like we need some type of punctuation before `all`.

StephenButtolph: updated both

michaelkaplan13 (approved):

PR #4465, opened 2025-11-05
Cleanup the version package

StephenButtolph:
## Why this should be merged

1. Removes an interface
2. Replaces an `error` with a `bool`
3. Generally simplifies some code
4. Removes `Semantic`
5. Removes duplicate function `Before`

## How this works

A cleanup that I factored out of my wip of the version logging work.

## How this was tested

CI

No

### DIFF
--- a/version/application.go
+++ b/version/application.go
@@ -21,8 +21,6 @@ type Application struct {
 	str         string
 }
 
-// The only difference here between Application and Semantic is that Application
-// prepends the client name rather than "v".
 func (a *Application) String() string {
 	a.makeStrOnce.Do(a.initString)
 	return a.str
@@ -36,10 +34,6 @@ func (a *Application) initString() {
 		a.Minor,
 		a.Patch,
 	)
-}
-
-func (a *Application) Before(o *Application) bool {
-	return a.Compare(o) < 0
 }
 
 // Compare returns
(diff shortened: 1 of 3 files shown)


version/application.go
```diff
@@ -42,27 +38,21 @@ func (a *Application) initString() {
...
-func (a *Application) Compatible(o *Application) error {
-	switch {
-	case a.Major > o.Major:
-		return errDifferentMajor
-	default:
-		return nil
-	}
-}
-
 func (a *Application) Before(o *Application) bool {
```
joshua-kim: I feel like we could even get rid of this function too since we already have `Compare`

version/version.go
```diff
@@ -24,35 +25,37 @@ type Semantic struct {
...
 }
 
 // The only difference here between Semantic and Application is that Semantic
 // prepends "v" rather than the client name.
 func (s *Semantic) String() string {
-	strIntf := s.str.Load()
-	if strIntf != nil {
-		return strIntf.(string)
-	}
+	s.makeStrOnce.Do(s.initString)
```
joshua-kim: Do we actually use this in a concurrent context? Wondering if we can just do a simple if not nil check instead

version/application.go
joshua-kim, line 1: I remember thinking this at some point in the past but I don't remember the answer... is there a reason we re-use the `Semantic` code to implement `Application`? It feels like we could define some prefix for `Semantic`'s string representation to reduce a bit of duplicate code...

alarso16 (approved):

version/application.go
alarso16: Yeah that does seem unnecessary

PR #4469, opened 2025-11-05
Apply latest tag only when the image tag matches a semantic version

sukantoraymond:
Changes scripts/build_image.sh so :latest is only applied when the image tag matches a semantic version (e.g., v1.3.0). Master builds continue to be tagged as :master (and by commit hash), not :latest, unless FORCE_TAG_LATEST is set.

### DIFF
--- a/scripts/build_image.sh
+++ b/scripts/build_image.sh
@@ -116,8 +116,8 @@ if [[ -z "${SKIP_BUILD_RACE}" ]]; then
                  "$AVALANCHE_PATH" -f "$AVALANCHE_PATH/Dockerfile"
 fi
 
-# Only tag the latest image for the master branch when images are pushed to a registry
-if [[ "${DOCKER_IMAGE}" == *"/"* && $image_tag == "master" ]]; then
+# Tag latest when pushing to a registry and the tag is a release (vMAJOR.MINOR.PATCH)
+if [[ "${DOCKER_IMAGE}" == *"/"* && $image_tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
   echo "Tagging current avalanchego images as $DOCKER_IMAGE:latest"
   docker buildx imagetools create -t "$DOCKER_IMAGE:latest" "$DOCKER_IMAGE:$commit_hash"
 fi


StephenButtolph:
I think it makes way more sense to `latest` to point to the most recent release, rather than `master`. Will defer to @maru-ava for reviewing the actual code tho

maru-ava:
> Changes scripts/build_image.sh so :latest is only applied when the image tag matches a semantic version (e.g., v1.3.0). Master builds continue to be tagged as :master (and by commit hash), not :latest, unless FORCE_TAG_LATEST is set.

This describes the _what_ of the PR, but not the _why_. Given that `latest` is a floating image tag that should never be used in production, what is the rational behind ensuring that it only targets release tags?

StephenButtolph:
> This describes the what of the PR, but not the why. Given that latest is a floating image tag that should never be used in production, what is the rational behind ensuring that it only targets release tags?

I agree the description could be improved. This had come up during our 1<->1, so I can probably talk to the rationale.

I felt that "latest" being an alias of "master" is a bit unexpected (as master should probably not be consumed externally). Having it map to the latest release felt less error prone.

While depending on a floating tag is not generally recommended, it can be nice for ease of use. A potential alternative could be including a `latest-stable` release or something.

maru-ava:
> I felt that "latest" being an alias of "master" is a bit unexpected (as master should probably not be consumed externally). Having it map to the latest release felt less error prone.

> While depending on a floating tag is not generally recommended, it can be nice for ease of use. A potential alternative could be including a latest-stable release or something.

Frankly, I'm not sure anyone has a good reason to use `latest` externally. It's a huge footgun for anybody not well-versed in how floating tags work. Maybe `latest-stable` and `latest-unstable` and remove `latest` entirely so folks know what they are getting into?

fwiw the bootstrap testing currently depends on `latest` being the last image published to master, whatever we decide there will need to be a floating tag that is updated every time a new image is build from HEAD of master.

StephenButtolph:
> whatever we decide there will need to be a floating tag that is updated every time a new image is build from HEAD of master.

That is the "master" tag no?

StephenButtolph:
> Maybe latest-stable ... and remove latest entirely so folks know what they are getting into?

I think that makes sense. Do you also think we should rename `master` to `latest-unstable`?

maru-ava:
> > whatever we decide there will need to be a floating tag that is updated every time a new image is build from HEAD of master.
> 
> That is the "master" tag no?

I guess so, yeah, TIL we have a master tag.

maru-ava:
> > Maybe latest-stable ... and remove latest entirely so folks know what they are getting into?
> 
> I think that makes sense. Do you also think we should rename `master` to `latest-unstable`?

I think that makes sense.

sukantoraymond:
> > > Maybe latest-stable ... and remove latest entirely so folks know what they are getting into?
> > 
> > 
> > I think that makes sense. Do you also think we should rename `master` to `latest-unstable`?
> 
> I think that makes sense.

So the only thing left is to change master to latest-unstable? @maru-ava

maru-ava:
> So the only thing left is to change master to latest-unstable? @maru-ava

I think this PR should change `latest` to `latest-unstable` as well. Where are we actually tagging `master` though?

sukantoraymond:
> > So the only thing left is to change master to latest-unstable? @maru-ava
> 
> I think this PR should change `latest` to `latest-unstable` as well. Where are we actually tagging `master` though?

Master is already tagged at 
```
${DOCKER_CMD} -t "$DOCKER_IMAGE:$commit_hash" -t "$DOCKER_IMAGE:$image_tag" \
              "$AVALANCHE_PATH" -f "$AVALANCHE_PATH/Dockerfile"
```

sukantoraymond:
> > So the only thing left is to change master to latest-unstable? @maru-ava
> 
> I think this PR should change `latest` to `latest-unstable` as well. Where are we actually tagging `master` though?

I have now tagged master as "latest-unstable"

maru-ava:
> > > So the only thing left is to change master to latest-unstable? @maru-ava
> > 
> > 
> > I think this PR should change `latest` to `latest-unstable` as well. Where are we actually tagging `master` though?
> 
> I have now tagged master as "latest-unstable"

I wasn't asking for an additional tagging step. Something is already tagging master and that would ideally be changed to tag `latest-unstable` instead and we remove the master tag.

StephenButtolph:
> I wasn't asking for an additional tagging step. Something is already tagging master and that would ideally be changed to tag `latest-unstable` instead and we remove the master tag.

`./scripts/image_tag.sh` defines the the `image_tag` as the human readable name of the commit that is checked out (either using a branch name or with a tag).

The `publish_docker_image` job is triggered upon:
```yaml
on:
  workflow_dispatch:
  push:
    tags:
      - "*"
    branches:
      - master
      - dev
```
So that's where the "master" tag is coming from.

maru-ava:
Tagging `DO NOT MERGE` until I get the bootstrap monitor updated and deployed to target the `master` tag instead of `latest`.

scripts/build_image.sh
```diff
@@ -116,8 +116,14 @@ if [[ -z "${SKIP_BUILD_RACE}" ]]; then
...
 
-# Only tag the latest image for the master branch when images are pushed to a registry
-if [[ "${DOCKER_IMAGE}" == *"/"* && ($image_tag == "master" || -n "${FORCE_TAG_LATEST}") ]]; then
+# Tag latest when pushing to a registry and the tag is a release (vMAJOR.MINOR.PATCH), or when forced
+if [[ "${DOCKER_IMAGE}" == *"/"* && ( "$image_tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ || -n "${FORCE_TAG_LATEST}" ) ]]; then
   echo "Tagging current avalanchego images as $DOCKER_IMAGE:latest"
   docker buildx imagetools create -t "$DOCKER_IMAGE:latest" "$DOCKER_IMAGE:$commit_hash"
 fi
+
+# Also tag master builds as latest-unstable when images are pushed to a registry
```
maru-ava: As per #4490, I'm going to just use the `master` tag so this can be left out for now.

scripts/build_image.sh
```diff
@@ -116,8 +116,14 @@ if [[ -z "${SKIP_BUILD_RACE}" ]]; then
                  "$AVALANCHE_PATH" -f "$AVALANCHE_PATH/Dockerfile"
 fi
 
-# Only tag the latest image for the master branch when images are pushed to a registry
-if [[ "${DOCKER_IMAGE}" == *"/"* && ($image_tag == "master" || -n "${FORCE_TAG_LATEST}") ]]; then
+# Tag latest when pushing to a registry and the tag is a release (vMAJOR.MINOR.PATCH), or when forced
+if [[ "${DOCKER_IMAGE}" == *"/"* && ( "$image_tag" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ || -n "${FORCE_TAG_LATEST}" ) ]]; then
```
maru-ava: @StephenButtolph This change appears to assume that we only ever tag new releases. Do we ever tag for a previous release (e.g. a bugfix release) such that `latest` could end up pointing to a previous release?

scripts/build_image.sh
sukantoraymond: removed

scripts/build_image.sh
maru-ava: Stil looking for an answer to this before approval

StephenButtolph: Just discussed this. We do not ever re-tag or tag old releases. So I think this makes sense 👍.

maru-ava: tl;dr We only plan on tagging the latest release, so this should be fine.

PR #4470, opened 2025-11-06
refactor: replace Split in loops with more efficient SplitSeq

huyejia:
## Why this should be merged

Optimize code using a more modern writing style which can make the code more efficient and cleaner.
More info: https://github.com/golang/go/issues/61901

strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.


maru-ava:
Why would it be desirable to optimize a loop that is used only by a testing library and likely to process at most 10 iterations per test run?

joshua-kim:
Closing, this optimization feels premature + no perf data.

PR #4471, opened 2025-11-06
Remove gosec from test linting

alarso16:
## Why this should be merged

The linter is intended to catch mistakes, but (almost) all of the nolint clauses for gosec in testing were for random number generation. These are intentional, so it's feels unreasonable to lint in the first place.

## How this works

Removed linter, and did a global search for `#nosec` and removed all in testing files.

## How this was tested

CI

No


joshua-kim (approved):

maru-ava (approved):
Nice cleanup!

PR #4472, opened 2025-11-06
[tooling] Update to use `go tool` for mockgen and canoto

maru-ava:
## Why this should be merged

During the transition to `go tool` as of 1.24, mockgen and canoto were somehow missed for migration to our new tools module (tools/go.mod). This was noticed when working on the coreth migration since coreth's mockgen usage had already been transitioned to go tool.

## How this was tested

CI

N/A


maru-ava:
As per @StephenButtolph these modules were intentionally excluded from tools/go.mod to avoid divergent versions vendored in go.mod and tools/go.mod.

PR #4478, opened 2025-11-06
Name and configure the H upgrade: Helicon

JonathanOppenheimer:
## Why this should be merged
This is perhaps the most important PR I have ever opened. We held an internal vote about what to name the next upgrade, and @bernard-avalabs's suggestion won! 

## How this works
This is just structural code for when logic for the next upgrade is added. 

## How this was tested
CI

No

### DIFF
--- a/upgrade/new_upgrade_checklist.md
+++ b/upgrade/new_upgrade_checklist.md
@@ -4,15 +4,13 @@
 
 This checklist ists all the required steps when when adding a new network upgrade/fork to AvalancheGo. Note that this list is not exhaustive nor future-proof.
 
-**Example:** Adding upgrade "Ionosphere" after "Helicon"
-
 ---
 
 ## 1. Core Configuration Files
 
 ### [`upgrade/upgrade.go`](./upgrade.go)
 
-- [ ] Add new time field to `Config` struct (e.g., `IonoTime time.Time`)
+- [ ] Add new time field to `Config` struct
 - [ ] Add `UnscheduledActivationTime` activation time to all configs:
   - `Mainnet` config
   - `Fuji` config


upgrade/upgradetest/fork.go
```diff
@@ -19,6 +19,7 @@ const (
 	Etna
 	Fortuna
 	Granite
+	Helicon // naming credit to @bernard-avalabs!
```
StephenButtolph: Not trying to take away any credit... But it looks weird in the code
```suggestion
	Helicon
```

StephenButtolph (dismissed):
Please model this PR after #3964

Bonus points for adding tests to automatically discover discrepancies between this PR and the above linked PR.

upgrade/upgradetest/fork.go
JonathanOppenheimer: I'll remember @bernard-avalabs... I'll remember...

image

api/info/service.md
```diff
@@ -675,28 +675,29 @@ curl -X POST --data '{
...
+    "apricotPhase6Time": "2022-09-06T20:00:00Z",
+    "apricotPhasePost6Time": "2022-09-07T03:00:00Z",
+    "banffTime": "2022-10-18T16:00:00Z",
+    "cortinaTime": "2023-04-25T15:00:00Z",
+    "cortinaXChainStopVertexID": "jrGWDh5Po9FMj54depyunNixpia5PN4aAYxfmNzU8n752Rjga",
+    "durangoTime": "2024-03-06T16:00:00Z",
+    "etnaTime": "2024-12-16T17:00:00Z",
+    "fortunaTime": "2025-04-08T15:00:00Z",
+    "graniteTime": "2025-11-19T16:00:00Z",
+    "heliconTime": "9999-12-01T00:00:00Z"
```
JonathanOppenheimer, line 700: I know this an example from [your comment](https://github.com/ava-labs/avalanchego/pull/3964/files#r2091504079) on the last PR -- I find it confusing (as Yacov did) that our example would have intentionally incorrect values, regardless it being a semantically correct response. I think they should be the actual values.

tests/e2e/e2e_test.go
```diff
@@ -51,11 +51,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	subnets := vms.XSVMSubnetsOrPanic(nodes...)
 
 	upgrades := upgrade.Default
-	if flagVars.ActivateGranite() {
-		upgrades.GraniteTime = upgrade.InitiallyActiveTime
-		upgrades.GraniteEpochDuration = 4 * time.Second
+	if flagVars.ActivateHelicon() {
+		upgrades.HeliconTime = upgrade.InitiallyActiveTime
+		upgrades.HeliconEpochDuration = 4 * time.Second
```
StephenButtolph: Why does Helicon have an epoch duration?

upgrade/upgrade.go
```diff
@@ -80,6 +84,8 @@ var (
 		FortunaTime:                  InitiallyActiveTime,
 		GraniteTime:                  InitiallyActiveTime,
 		GraniteEpochDuration:         30 * time.Second,
+		HeliconTime:                  InitiallyActiveTime,
```
StephenButtolph:
```suggestion
		HeliconTime:                  UnscheduledActivationTime,
```

tests/e2e/e2e_test.go
JonathanOppenheimer: My understanding is that each upgrade should have these from ACP-181. Does only Granite get an epoch duration?

JonathanOppenheimer: and I would assume when Helicon is activated, we'd use it's epochs in `epoch.go`

StephenButtolph: [ACP-181](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/181-p-chain-epoched-views#setting-the-epoch-duration) specifies:

> Any changes to $D$ in the future would require another network upgrade.

Unless we are changing $D$, we shouldn't be adding a new parameter.

JonathanOppenheimer: Ah I see. I will revert these additions. Perhaps this is a skill issue on my part, but in that case, I find the labeling of the variables is confusing. 

If Epoch durations are independent of major network versions, why would it be called `GraniteEpochDuration`? This would result that during Helicon, the current epoch duration would be `GraniteEpochDuration`, as in the `I` upgrade etc. This seems confusing at first glance to a reader. 

In my opinion, the epoch duration should be a single variable `EpochDuration` which would be used in a production with a different methodology to track historical epoch durations.

tests/e2e/e2e_test.go
```diff
@@ -51,11 +51,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	subnets := vms.XSVMSubnetsOrPanic(nodes...)
 
 	upgrades := upgrade.Default
-	if flagVars.ActivateGranite() {
-		upgrades.GraniteTime = upgrade.InitiallyActiveTime
+	if flagVars.ActivateHelicon() {
+		upgrades.HeliconTime = upgrade.InitiallyActiveTime
 		upgrades.GraniteEpochDuration = 4 * time.Second
```
StephenButtolph: Why do we only override the granite epoch duration when Helicon is activated?

tests/e2e/e2e_test.go
StephenButtolph: There was actually quite a bit of conversation around this variable even existing in this config... I agree that it really doesn't fit in very well.

tests/e2e/e2e_test.go
JonathanOppenheimer: Yeah I see we don't have to - we have the old `GraniteEpochDuration` stored from when Granite was activated -- removed.

tests/e2e/e2e_test.go
JonathanOppenheimer: Yeah alternatively, it could just not be in this config at all haha

tests/e2e/e2e_test.go
```diff
@@ -51,11 +50,10 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	subnets := vms.XSVMSubnetsOrPanic(nodes...)
 
 	upgrades := upgrade.Default
-	if flagVars.ActivateGranite() {
-		upgrades.GraniteTime = upgrade.InitiallyActiveTime
-		upgrades.GraniteEpochDuration = 4 * time.Second
```
StephenButtolph, line 56: By removing this, we are changing the epoch duration from `4s` to `30s` in our tests. This causes some of our E2E tests to sleep longer...

I think this is fine... The CI time seems to be around 15m-16m before and after this change... So doesn't seem like a significant slowdown... But if we wanted to keep it the same, we should override the epoch duration regardless of whether Helicon is activated.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	})
+
+	t.Run("upgradetest Fork constants match config fields", func(*testing.T) {
+		// For each time field in upgrade.Config (except Apricot phases and special cases),
+		// there should be a corresponding Fork constant
+		expectedForks := make(map[string]bool)
+		for _, fieldName := range timeFields {
+			upgradeName := strings.TrimSuffix(fieldName, "Time")
+			// Skip Apricot variants as they have complex naming
```
maru-ava: (No action required) Maybe prefer early exit?

```golang
if strings.HasPrefix(upgradeName, "Apricot") {
    // Skip Apricot variants as they have complex naming
    continue
}
expectedForks[upgradeName] = false
```

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
...
+			if !strings.HasPrefix(upgradeName, "Apricot") {
+				expectedForks[upgradeName] = false
+			}
+		}
+
+		// Check that each fork can be stringified
+		for fork := NoUpgrades; fork <= Latest; fork++ {
+			forkName := fork.String()
```
maru-ava: (No action required) Am I missing something, or does `forkName` need to be checked to be something other than `"Unknown"`?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
...
+				continue
+			}
+
+			if _, exists := expectedForks[forkName]; exists {
+				expectedForks[forkName] = true
+			}
+		}
+
+		// Verify all expected forks were found
+		for forkName, found := range expectedForks {
```
maru-ava: (No action required) Rather than conflating the checks for existence with stringifiable, maybe add something like `Fork.FromString(forkName string) Fork` to handle an existence check so as to avoid having to use the same map for both checks?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
...
+		testTime := time.Date(2030, time.January, 1, 0, 0, 0, 0, time.UTC)
+
+		for fork := ApricotPhase1; fork <= Latest; fork++ {
+			forkName := fork.String()
+
+			t.Run(forkName, func(*testing.T) {
+				config := upgrade.Config{}
+				SetTimesTo(&config, fork, testTime)
+
+				// Verify that at least this fork's time is set
```
maru-ava: Maybe ensure that this test verifies that SetTimesTo fulfills its stated contract to set the upgrade time of the provided fork _and all prior forks_? As written, there would appear to be an opportunity for someone to forget a `fallthrough` statement for any given fork and only set the time for the targeted fork.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	})
+
+	t.Run("Latest fork is properly defined", func(*testing.T) {
+		// Latest should be a valid fork constant
+		require.NotEqual(NoUpgrades, Latest, "Latest should not be NoUpgrades")
+		require.NotEqual("Unknown", Latest.String(), "Latest should have a valid string representation")
+
+		// Latest should be >= all other named forks
+		require.GreaterOrEqual(Latest, Granite, "Latest should be >= Granite")
```
maru-ava: (No action required) Is this intended to be maintained at each upgrade or is `>= Granite` enough in perpetuity?

upgrade/upgradetest/upgrade_consistency_test.go
maru-ava: Or is the absence of a check for `"Unknown"` because `TestForkStringCompleteness` covers that case?

tests/fixture/e2e/flags.go
```diff
@@ -40,7 +40,7 @@ type FlagVars struct {
 	stopNetwork    bool
 	restartNetwork bool
 
-	activateGranite bool
+	activateHelicon bool
```
maru-ava: (No action required) Do we need to change this flag every release, or could we just use `Latest` instead?

tests/e2e/e2e_test.go
```diff
@@ -51,11 +50,10 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	if flagVars.ActivateHelicon() {
```
maru-ava: (No action required) Do we need to change the target every release or could we just use `Latest`?

.github/workflows/ci.yml
```diff
@@ -60,15 +60,15 @@ jobs:
           loki_push_url: ${{ secrets.LOKI_PUSH_URL || '' }}
           loki_username: ${{ secrets.LOKI_USERNAME || '' }}
           loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
-  e2e_post_granite:
+  e2e_post_helicon:
```
maru-ava: (No action required) Do we need to change this job every release, or could we just use Latest instead?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,251 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
maru-ava: (No action required) Not that I don't appreciate your diligence, but I'm not entirely sure how many of these tests are worth the cost of implementation, review and maintenance. Could the set presented here be simplified to only the essential ones?

maru-ava: I realize what I'd really like to see is a checklist (similar to the one for bumping go version in go.mod) for the tasks associated with configuring a new upgrade. Fine to have some of the easy characteristics of configuring a new upgrade validated automatically, and even better if we have a checklist that documents the steps involved for both implementer and reviewers.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: Yup, `TestForkStringCompleteness` covers that case

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: Good suggestion!

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: No this, would have to be maintained in perpetuity, so it's not really a very good test. I will remove it.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: Yes this makes a lot of sense

tests/e2e/e2e_test.go
JonathanOppenheimer: How would you suggest using latest here? My understanding is that it needs to be this way to set the specific upgrade variables when each release is triggered.

tests/fixture/e2e/flags.go
JonathanOppenheimer: Actually we could add a flag for this.

tests/e2e/e2e_test.go
```diff
@@ -51,11 +51,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	subnets := vms.XSVMSubnetsOrPanic(nodes...)
 
 	upgrades := upgrade.Default
-	if flagVars.ActivateGranite() {
-		upgrades.GraniteTime = upgrade.InitiallyActiveTime
-		upgrades.GraniteEpochDuration = 4 * time.Second
+	if flagVars.ActivateLatest() {
+		// Activate all upgrades up to and including the latest
+		upgradetest.SetTimesTo(&upgrades, upgradetest.Latest, upgrade.InitiallyActiveTime)
```
JonathanOppenheimer: Is this acceptable? This is functional change, as it will set all prior upgrade times, rather than just the specific upgrade.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: I removed the tests that would require changes to pass - this is fair criticism, and tests are not very meaningful if you have to update them to check what you're verifying anyway.

JonathanOppenheimer: Here is the checklist: https://github.com/ava-labs/avalanchego/blob/04518ba88f674ec63f2b2708cce49e976ab2ec8a/upgrade/new_upgrade_checklist.md

tests/e2e/e2e_test.go
maru-ava: @StephenButtolph Is this acceptable? Or would it make sense to add a function that only sets the latest time?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+		field := configType.Field(i)
+		if field.Type == reflect.TypeOf(time.Time{}) {
+			timeFields = append(timeFields, field.Name)
+		}
+		if field.Type == reflect.TypeOf(time.Duration(0)) {
+			durationFields = append(durationFields, field.Name)
+		}
+	}
+
+	require.NotEmpty(timeFields, "upgrade.Config should have time.Time fields")
```
maru-ava: Given that this check is never going to fail because we're never going to delete legacy times, why is this check desirable?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	t.Run("all time fields have activation methods", func(*testing.T) {
```
maru-ava: Given that the only reason for these methods to exist is to be used to query activation, and there are already a ton of examples of the expected form for previous releases, why is it desirable to have to maintain this test? If an activation method isn't added when we configure a new release, it will be trivial to add one that matches the expected form when needed.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+		testTime := time.Date(2025, time.January, 1, 0, 0, 0, 0, time.UTC)
+		for _, fieldName := range timeFields {
+			field := configValue.FieldByName(fieldName)
+			require.True(field.IsValid(), "Field %s should exist", fieldName)
+			field.Set(reflect.ValueOf(testTime))
+		}
+
+		// Set duration fields to valid values
+		for _, fieldName := range durationFields {
```
maru-ava: Given that durationFields is orthogonal to timeFields, and Config.Validate doesn't reference duration fields, this block appears entirely superfluous.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+			// Verify method signature: func(*upgrade.Config, time.Time) bool
+			require.Equal(2, method.Type.NumIn(), "Method %s should have 2 inputs (receiver, time.Time)", expectedMethodName)
+			require.Equal(1, method.Type.NumOut(), "Method %s should have 1 output (bool)", expectedMethodName)
+			require.Equal(reflect.TypeOf(time.Time{}), method.Type.In(1), "Method %s should take time.Time as parameter", expectedMethodName)
+			require.Equal(reflect.TypeOf(true), method.Type.Out(0), "Method %s should return bool", expectedMethodName)
+		}
+	})
+
+	t.Run("all time fields are in Validate method", func(*testing.T) {
```
maru-ava: I think this test is validating 2 separable concerns: whether all forks have upgrade times and whether the upgrade time ordering is as expected.  Maybe simplify by separating the testing?
 - Define a mapping of fork -> forkTime
 - Add a test that validates that each fork is present in the mapping 
 - Move the check for upgrade time ordering to upgrade/upgrade_test.go

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+		// Set the last time field to before the first
+		if len(timeFields) >= 2 {
+			lastField := configValue.FieldByName(timeFields[len(timeFields)-1])
+			lastField.Set(reflect.ValueOf(testTime.Add(-time.Hour)))
+			err := testConfig.Validate()
+			require.ErrorIs(err, upgrade.ErrInvalidUpgradeTimes, "Validate should fail when upgrade times are out of order")
+		}
+	})
+
+	t.Run("all time fields are set in network configs", func(t *testing.T) {
```
maru-ava: How about we add a method to upgrade.Config that given a Fork value, it returns the activation time of that fork? That would make it trivial to iterate through all forks and ensure the time is non-zero.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+					field := configValue.FieldByName(fieldName)
+					require.True(field.IsValid(), "%s.%s should exist", tc.name, fieldName)
+					timeValue := field.Interface().(time.Time)
+					require.False(timeValue.IsZero(), "%s.%s should be set to a non-zero time", tc.name, fieldName)
+				}
+			})
+		}
+	})
+
+	t.Run("all Fork constants are stringifiable", func(*testing.T) {
```
maru-ava: Maybe rewrite this to perform bidirectional verification? i.e. for each Fork, stringify, then ensure that string can be converted back to the same Fork value. And maybe an additional check for an invalid string.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+		// Check that each fork can produce a valid string representation
+		for fork := NoUpgrades; fork <= Latest; fork++ {
+			forkName := fork.String()
+			require.NotEqual("Unknown", forkName, "Fork %d should be stringifiable", fork)
+		}
+	})
+
+	t.Run("upgradetest Fork constants match config fields", func(*testing.T) {
```
maru-ava: This test will not be necessary if a test is added (as per [a previous comment](https://github.com/ava-labs/avalanchego/pull/4478/files#r2523776714)) that validates that an upgrade time exists for each Fork. Also, I don't think we need to be validating that time fields have a `Time` suffix - the presence of so many existing examples of how to name things will make it very hard for a reviewer to miss any discrepancy.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+	})
+}
+
+// TestUpgradeFieldNaming ensures consistent naming conventions across the codebase.
+func TestUpgradeFieldNaming(t *testing.T) {
+	require := require.New(t)
+
+	configType := reflect.TypeOf(upgrade.Config{})
+
+	t.Run("time fields end with Time suffix", func(*testing.T) {
```
maru-ava: What prompted this test? Without context, I'm not sure why it needs to exist.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+				require.True(
+					strings.HasSuffix(field.Name, "Time"),
+					"time.Time field %s should have 'Time' suffix",
+					field.Name,
+				)
+			}
+		}
+	})
+
+	t.Run("duration fields end with Duration or Height suffix", func(*testing.T) {
```
maru-ava: What prompted this test? Without context, I'm not sure why it needs to exist.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+				)
+			}
+		}
+	})
+
+	t.Run("json tags use camelCase", func(*testing.T) {
+		for i := 0; i < configType.NumField(); i++ {
+			field := configType.Field(i)
+			jsonTag := field.Tag.Get("json")
+			if jsonTag != "" {
```
maru-ava: Why would it be acceptable not to have a json tag?

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+				// Extract just the name part (before any commas)
+				jsonName := strings.Split(jsonTag, ",")[0]
+				require.NotEmpty(jsonName, "Field %s should have non-empty json tag", field.Name)
```
maru-ava: I don't think the empty or lower-case letter check are what we want to be checking. Instead, check that the json tag is the lower-camel version of the field name.

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+	t.Run("valid fork names", func(*testing.T) {
+		// Test that all valid fork constants can be converted to string and back
+		for fork := NoUpgrades; fork <= Latest; fork++ {
+			forkStr := fork.String()
+			parsedFork := FromString(forkStr)
+			require.Equal(fork, parsedFork, "FromString(%q) should return %d, got %d", forkStr, fork, parsedFork)
+		}
+	})
+
+	t.Run("invalid fork names", func(*testing.T) {
```
maru-ava: This test should be redundant after refactoring the test on 117

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+		if fork <= Latest {
+			require.NotEqual("Unknown", forkStr, "Fork %d should not return 'Unknown'", fork)
+		} else {
+			// Forks beyond Latest should return "Unknown"
+			require.Equal("Unknown", forkStr, "Fork %d (beyond Latest) should return 'Unknown'", fork)
+		}
+	}
+}
+
+// TestFromString ensures FromString correctly maps all fork names to their constants.
```
maru-ava: This test should be redundant after refactoring the test on 117

upgrade/upgradetest/upgrade_consistency_test.go
```diff
@@ -0,0 +1,290 @@
...
+					"Field %s json tag '%s' should start with lowercase letter",
+					field.Name,
+					jsonName,
+				)
+			}
+		}
+	})
+}
+
+// TestForkStringCompleteness ensures all Fork constants have String() cases.
```
maru-ava: This test should be redundant after refactoring the test on 117

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,117 @@
+<!-- markdownlint-disable MD024 -->
```
maru-ava, line 1: This is more complicated than I was hoping. As per the example of the go.mod checklist, please simplify to include a single list for both implementers and reviewers and avoid including details that should be obvious from a consistency standpoint (e.g. suffix with Time) or will be validated automatically (e.g. json tag needs to be X).

tests/e2e/e2e_test.go
StephenButtolph: In practice, this isn't actually a functional change, so I think its fine.

tests/e2e/e2e_test.go
JonathanOppenheimer: Would you like to keep it the same?

tests/e2e/e2e_test.go
JonathanOppenheimer: because all the prior upgrade times are already set, correct?

upgrade/new_upgrade_checklist.md
JonathanOppenheimer: I have vastly simplified the checklist - let me know if you would like it to be even simpler 

(e.g. just a list of file names like the go.mod)

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: I have removed it - I assume if someone removed the legacy times, some other test would scream.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: What additional maintainence would this test require? It doesn't ever need to be udpated unless we change the `isActive()` method template. If adding an `isActive()` method is part of setting up a new upgrade (as it was in Stephen's PR), I think this test makes sense.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: It should not be.

upgrade/upgradetest/upgrade_consistency_test.go
maru-ava: A checklist should be enough. If for some reason we miss it, and nothing in the PR requires it, no biggie. The first person that needs to use it can add it. No code is the best code.

upgrade/new_upgrade_checklist.md
maru-ava: Much cleaner, thank you!

For future, please follow the workflow of:
 - respond to a request for a change (done,explanation,more questions,etc)
 - leave it to your reviewer(s) to resolve conversations when they are satisfied

JonathanOppenheimer: Oh man I have been resolving comments after I have addressed them (in my opinion at least) since I started - I've never been told otherwise lol - will start doing!

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: What are your thoughts on this? If I take in a forkName value, the function would have to determine which configuration, and then we'd have to maintain a map of forks to fork activation times. This would be another thing to update when prepping a new upgrade version.

JonathanOppenheimer: Actually I see in a prior comment you ask for a defined mapping. I think this is an additional thing to maintain, but would mean you would not need reflection.

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: I think adding a map is additional overhead that would need to be maintained, and would only be used for this test -- seems like overkill, but I will do so.

JonathanOppenheimer: Additionally, this wouldn't actually test that someone added the proper fields to the Config structure -- just wehther they updated the map.

maru-ava: You were right to push back initially. I'm looking for a more testable thing that doesn't require `reflect`, but the shape of it isn't fixed in my head so if what I say doesn't make sense asking for clarification is generally preferable than following bad advice.

tests/e2e/e2e_test.go
JonathanOppenheimer: I will keep the name

upgrade/upgradetest/upgrade_consistency_test.go
JonathanOppenheimer: This is tough as I don't think I can change the serialization format for all the upgrades - I think we can try to use maps by default.

JonathanOppenheimer: Take a look at the current approach - we mirror the string strategy by declaring custom getter/switch, so the we test the code as directly as we do in other cases. You can't update the getter without updating the actual production code.

tests/e2e/e2e_test.go
JonathanOppenheimer: See current code changes - we set all

tests/e2e/e2e_test.go
JonathanOppenheimer: using `SetTimesTo()` actually does not work for the E2E, because it breaks if the --set-latest is not sent - it bricks every upgrade. I'll use a new function!

upgrade/upgradetest/config.go
```diff
@@ -76,3 +79,40 @@ func SetTimesTo(c *upgrade.Config, fork Fork, upgradeTime time.Time) {
 		c.ApricotPhase1Time = upgradeTime
 	}
 }
+
+// SetTimeOnly sets the upgrade time of only the specified fork to the provided
+// upgradeTime, without affecting any other fork times.
+func SetTimeOnly(c *upgrade.Config, fork Fork, upgradeTime time.Time) {
```
JonathanOppenheimer: This is in the `upgradetest` package but used in production code -- however doing it otherwise results in an cyclic import loop. This seems bad.

JonathanOppenheimer:
`e2e_post_granite` will never run because it doesn't exist anymore.

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,44 @@
...
+  - [ ] Add configuration field(s) to `Config` struct
+  - [ ] Add to all three configs (`Mainnet`, `Fuji`, `Default`)
+  - [ ] Update `Validate()` if validation is needed
+
+## 2. Test Helper Files
+
+### [`upgrade/upgradetest/fork.go`](./upgradetest/fork.go)
+
+- [ ] Add new fork constant before `Latest`
+- [ ] Update `Latest` constant to point to the new fork
```
StephenButtolph:
```suggestion
```

upgrade/upgradetest/fork.go
```diff
@@ -62,3 +71,80 @@ func (f Fork) String() string {
...
+	case ApricotPhase3:
+		return upgrade.Configs[networkID].ApricotPhase3Time
+	case ApricotPhase2:
+		return upgrade.Configs[networkID].ApricotPhase2Time
+	case ApricotPhase1:
+		return upgrade.Configs[networkID].ApricotPhase1Time
+	default:
+		return time.Time{}
+	}
+}
```
StephenButtolph: `FromString` seems fairly reasonable to have (to align with `String()`). But adding this code as a mechanism to test this (already) testing code seems pretty cumbersome...

upgrade/upgradetest/upgrade_consistency_test.go
StephenButtolph, line 1: I feel like this file is more trouble than it is worth 🪓

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,43 @@
...
+- [ ] Add `UnscheduledActivationTime` activation time to all configs:
+  - `Mainnet` config
+  - `Fuji` config
+  - `Default` config
+- [ ] Add field to `Validate()` method's `upgrades` slice
+- [ ] Add `IsXActivated(time.Time) bool` method
+- [ ] If the upgrade has associated configuration (like `GraniteEpochDuration`):
+  - [ ] Add configuration field(s) to `Config` struct
+  - [ ] Add to all three configs (`Mainnet`, `Fuji`, `Default`)
+  - [ ] Update `Validate()` if validation is needed
```
StephenButtolph: Typically the full scope of an upgrade isn't actually known when we do this boilerplate.. I think we can leave this to the development of the features.
```suggestion
```

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,43 @@
...
+
+## 2. Test Helper Files
+
+### [`upgrade/upgradetest/fork.go`](./upgradetest/fork.go)
+
+- [ ] Add new fork constant before `Latest`
+- [ ] Add case to `String()` and `FromString()` method
```
StephenButtolph:
```suggestion
- [ ] Add case to `String()` method
```

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,43 @@
...
+
+### [`upgrade/upgradetest/config.go`](./upgradetest/config.go)
+
+- [ ] Add case to `SetTimesTo()` function
+
+---
```
StephenButtolph, line 40: Can we also add that the rpcchainvm initialize needs to be updated

upgrade/upgrade_test.go
```diff
@@ -36,13 +37,47 @@ func TestValidDefaultUpgrades(t *testing.T) {
...
+
+	// Now test that Validate catches out-of-order times
+	// Set the last time field to before the first
+	if len(timeFields) >= 2 {
+		lastField := configValue.FieldByName(timeFields[len(timeFields)-1])
+		lastField.Set(reflect.ValueOf(testTime.Add(-time.Hour)))
+		err := testConfig.Validate()
+		require.ErrorIs(t, err, ErrInvalidUpgradeTimes, "Validate should fail when upgrade times are out of order")
+	}
 }
```
StephenButtolph: 🪓 I think we can get rid of the tests that use reflect.

vms/rpcchainvm/vm_server.go
```diff
@@ -925,5 +928,6 @@ func convertNetworkUpgrades(pbUpgrades *vmpb.NetworkUpgrades) (upgrade.Config, e
 		EtnaTime:                     etna,
 		FortunaTime:                  fortuna,
 		GraniteTime:                  granite,
+		HeliconTime:                  helicon,
```
StephenButtolph, line 933: Can we add a test for if this is missing? Should be easy with the upgradetest package

upgrade/upgradetest/fork.go
```diff
@@ -19,15 +19,18 @@ const (
 	Etna
 	Fortuna
 	Granite
+	Helicon
 
-	Latest = Granite
+	Latest = iota - 1
```
StephenButtolph: My bad
```suggestion
	Latest Fork = iota - 1
```

tests/e2e/e2e_test.go
```diff
@@ -50,13 +50,12 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-	if flagVars.ActivateGranite() {
-		upgrades.GraniteTime = upgrade.InitiallyActiveTime
-		upgrades.GraniteEpochDuration = 4 * time.Second
-	} else {
-		upgrades.GraniteTime = upgrade.UnscheduledActivationTime
+	upgradeToActivate := upgradetest.Latest
+	if !flagVars.ActivateLatest() {
+		upgradeToActivate--
 	}
+	upgrades := upgradetest.GetConfig(upgradetest.Fork(upgradeToActivate))
```
StephenButtolph:
```suggestion
	upgrades := upgradetest.GetConfig(upgradeToActivate)
```

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,48 @@
...
+- [ ] Add `IsXActivated(time.Time) bool` method
+
+### [`vms/rpcchainvm/vm_client.go`](../vms/rpcchainvm/vm_client.go)
+
+- [ ] Add new time field to `NetworkUpgrades` struct
+
+### [`vms/rpcchainvm/vm_server.go`](../vms/rpcchainvm/vm_server.go)
+
+- [ ] Add a new `grpcutils.TimestampAsTime` block to `convertNetworkUpgrades()` function
+- [ ] Return the new time in the same function (e.g., `IonoTime time.Time`)
```
StephenButtolph: These really seem like a single task to me.
```suggestion
- [ ] Add a new `grpcutils.TimestampAsTime` block to `convertNetworkUpgrades()` function
```

upgrade/upgrade.go
```diff
@@ -38,7 +38,9 @@ var (
 		FortunaTime:               time.Date(2025, time.April, 8, 15, 0, 0, 0, time.UTC),
 		GraniteTime:               time.Date(2025, time.November, 19, 16, 0, 0, 0, time.UTC),
 		GraniteEpochDuration:      5 * time.Minute,
+		HeliconTime:               UnscheduledActivationTime,
 	}
+
```
StephenButtolph:
```suggestion
```

upgrade/upgrade.go
```diff
@@ -61,7 +63,9 @@ var (
 		FortunaTime:               time.Date(2025, time.March, 13, 15, 0, 0, 0, time.UTC),
 		GraniteTime:               time.Date(2025, time.October, 29, 15, 0, 0, 0, time.UTC),
 		GraniteEpochDuration:      5 * time.Minute,
+		HeliconTime:               UnscheduledActivationTime,
 	}
+
```
StephenButtolph:
```suggestion
```

upgrade/upgrade_test.go
StephenButtolph, line 1: File diff can probably be reverted.

tests/e2e/e2e_test.go
JonathanOppenheimer: This doesn't work:

cannot use upgradeToActivate (variable of type int) as upgradetest.Fork value in argument to upgradetest.GetConfigcompilerIncompatibleAssign

JonathanOppenheimer: Oh nevermind saw your other suggestion

vms/rpcchainvm/vm_server.go
JonathanOppenheimer: Yup added

upgrade/upgrade_test.go
JonathanOppenheimer: Reverted

vms/rpcchainvm/vm_test.go
```diff
@@ -255,3 +256,39 @@ func TestNewHTTPHandler(t *testing.T) {
...
+		HeliconTime:                   grpcutils.TimestampFromTime(wantConfig.HeliconTime),
+	}
+
+	gotConfig, err := convertNetworkUpgrades(pbUpgrades)
+	require.NoError(t, err)
+
+	// not passed through the RPC protocol, but we need to set it to match the
+	// expected value for comparison.
+	// TODO(JonathanOppenheimer): this should definitely be removed from the config
+	gotConfig.GraniteEpochDuration = wantConfig.GraniteEpochDuration
```
JonathanOppenheimer: another reason this should be removed.

vms/rpcchainvm/vm_test.go
```diff
@@ -255,3 +256,39 @@ func TestNewHTTPHandler(t *testing.T) {
...
+		ApricotPhasePost_6Time:        grpcutils.TimestampFromTime(wantConfig.ApricotPhasePost6Time),
+		BanffTime:                     grpcutils.TimestampFromTime(wantConfig.BanffTime),
+		CortinaTime:                   grpcutils.TimestampFromTime(wantConfig.CortinaTime),
+		CortinaXChainStopVertexId:     wantConfig.CortinaXChainStopVertexID[:], // not passed through the RPC protocol
+		DurangoTime:                   grpcutils.TimestampFromTime(wantConfig.DurangoTime),
+		EtnaTime:                      grpcutils.TimestampFromTime(wantConfig.EtnaTime),
+		FortunaTime:                   grpcutils.TimestampFromTime(wantConfig.FortunaTime),
+		GraniteTime:                   grpcutils.TimestampFromTime(wantConfig.GraniteTime),
+		HeliconTime:                   grpcutils.TimestampFromTime(wantConfig.HeliconTime),
+	}
```
StephenButtolph: This is copied from `vm_client.go`. This should be made into a helper (similarly to how `convertNetworkUpgrades` is a helper the other way).

vms/rpcchainvm/vm_test.go
StephenButtolph: Isn't this just a bug? If we are passing through things like `ApricotPhase_4MinPChainHeight` and `CortinaXChainStopVertexId`, I'd expect `GraniteEpochDuration` to be passed through.

vms/rpcchainvm/vm_test.go
```diff
@@ -255,3 +256,39 @@ func TestNewHTTPHandler(t *testing.T) {
 	handler.ServeHTTP(w, httptest.NewRequest(http.MethodGet, "/", nil))
 	require.Equal(http.StatusOK, w.Code)
 }
+
+// TestConvertNetworkUpgrades_AllFieldsHandled ensures that all fields in
+// upgrade.Config are properly handled when converting from the gRPC protobuf message.
+func TestConvertNetworkUpgrades_AllFieldsHandled(t *testing.T) {
+	wantConfig := upgrade.Default
```
StephenButtolph: Can we test this for all of:
```
constants.MainnetID
constants.FujiID
constants.LocalID
constants.UnitTestID
```
?

JonathanOppenheimer: yup!

vms/rpcchainvm/vm_test.go
JonathanOppenheimer: Turned into a helper!

vms/rpcchainvm/vm_test.go
JonathanOppenheimer: As I shared prior, I don't think either the `ApricotPhase_4MinPChainHeight` or the `CortinaXChainStopVertexId` should really be stored in this structure -- it doesn't make sense to me. But that is probably a seperate PR, and I can add `GraniteEpochDuration` to the `NetworkUpgrades`

vms/rpcchainvm/vm_client.go
```diff
@@ -269,6 +253,29 @@ func (vm *VMClient) Initialize(
 	return err
 }
 
+func getNetworkUpgrades(chainCtx *snow.Context) *vmpb.NetworkUpgrades {
```
StephenButtolph:
```suggestion
func getNetworkUpgrades(u upgrade.Config) *vmpb.NetworkUpgrades {
```

vms/rpcchainvm/vm_test.go
```diff
@@ -255,3 +258,44 @@ func TestNewHTTPHandler(t *testing.T) {
...
+		},
+		{
+			name:      "Fuji",
+			networkID: constants.FujiID,
+		},
+		{
+			name:      "Local",
+			networkID: constants.LocalID,
+		},
+	}
```
StephenButtolph, line 286:
```suggestion
	tests := []networks{
		{
			name:      "Mainnet",
			networkID: constants.MainnetID,
		},
		{
			name:      "Fuji",
			networkID: constants.FujiID,
		},
		{
			name:      "Local",
			networkID: constants.LocalID,
		},
		{
			name:      "UnitTest",
			networkID: constants.UnitTestID,
		},
	}
```

vms/rpcchainvm/vm_client.go
```diff
@@ -269,6 +254,29 @@ func (vm *VMClient) Initialize(
...
+		ApricotPhase_3Time:            grpcutils.TimestampFromTime(u.ApricotPhase3Time),
+		ApricotPhase_4Time:            grpcutils.TimestampFromTime(u.ApricotPhase4Time),
+		ApricotPhase_4MinPChainHeight: u.ApricotPhase4MinPChainHeight, // not passed through the RPC protocol
+		ApricotPhase_5Time:            grpcutils.TimestampFromTime(u.ApricotPhase5Time),
+		ApricotPhasePre_6Time:         grpcutils.TimestampFromTime(u.ApricotPhasePre6Time),
+		ApricotPhase_6Time:            grpcutils.TimestampFromTime(u.ApricotPhase6Time),
+		ApricotPhasePost_6Time:        grpcutils.TimestampFromTime(u.ApricotPhasePost6Time),
+		BanffTime:                     grpcutils.TimestampFromTime(u.BanffTime),
+		CortinaTime:                   grpcutils.TimestampFromTime(u.CortinaTime),
+		CortinaXChainStopVertexId:     u.CortinaXChainStopVertexID[:], // not passed through the RPC protocol
```
StephenButtolph:
```suggestion
		ApricotPhase_4MinPChainHeight: u.ApricotPhase4MinPChainHeight,
		ApricotPhase_5Time:            grpcutils.TimestampFromTime(u.ApricotPhase5Time),
		ApricotPhasePre_6Time:         grpcutils.TimestampFromTime(u.ApricotPhasePre6Time),
		ApricotPhase_6Time:            grpcutils.TimestampFromTime(u.ApricotPhase6Time),
		ApricotPhasePost_6Time:        grpcutils.TimestampFromTime(u.ApricotPhasePost6Time),
		BanffTime:                     grpcutils.TimestampFromTime(u.BanffTime),
		CortinaTime:                   grpcutils.TimestampFromTime(u.CortinaTime),
		CortinaXChainStopVertexId:     u.CortinaXChainStopVertexID[:],
```

upgrade/new_upgrade_checklist.md
```diff
@@ -0,0 +1,47 @@
...
+
+**Example:** Adding upgrade "Ionosphere" after "Helicon"
+
+---
+
+## 1. Core Configuration Files
+
+### [`upgrade/upgrade.go`](./upgrade.go)
+
+- [ ] Add new time field to `Config` struct (e.g., `IonoTime time.Time`)
```
alarso16: What is `Iono`?

alarso16 (approved):

upgrade/new_upgrade_checklist.md
JonathanOppenheimer: I removed the example entirely!

PR #4480, opened 2025-11-06
Avoid unnecessary RST_STREAM and PING frames causing GOAWAY messages

michaelkaplan13:
## Why this should be merged
Avoids GOAWAY errors from various API servers by avoiding client implementations that send extra unnecessary RST_STREAM and PING frames. The workaround is adopted from the Cloudflare blog linked below, and in comments.

## How this works
See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive

## How this was tested
Using a minimal reproduction script [here](https://github.com/ava-labs/avalanchego/tree/michaelkaplan13/goaway-error-repro/goawaytest). On that branch, running the script with the changes in `json.go` should yield no errors, but removing the changes in `json.go` and running the script consistently results in ~125 GOAWAY errors.

Open to suggestions for how to better test it in an automated fashion. I could not think of a good way to do so myself since it depends on the external server setup.

No

### DIFF
--- a/api/metrics/client.go
+++ b/api/metrics/client.go
@@ -60,10 +60,5 @@ func (c *Client) GetMetrics(ctx context.Context) (map[string]*dto.MetricFamily, 
 	}
 
 	var parser expfmt.TextParser
-	metrics, err := parser.TextToMetricFamilies(resp.Body)
-	if err != nil {
-		return nil, err
-	}
-
-	return metrics, nil
+	return parser.TextToMetricFamilies(resp.Body)
 }


api/metrics/client.go
```diff
@@ -52,6 +53,10 @@ func (c *Client) GetMetrics(ctx context.Context) (map[string]*dto.MetricFamily,
 
 	// Return an error for any non successful status code
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		// Avoid sending unnecessary RST_STREAM and PING frames by ensuring the whole body is read.
```
maru-ava: (No action required) Rather than duplicating this pattern everywhere, maybe use a function whose name is self-documenting? e.g.

```golang
// CleanlyCloseBody avoids sending unnecessary RST_STREAM and PING frames by ensuring the whole body is read before being closed.
// See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive
func CleanlyCloseBody(body io.ReadCloser) error {
    _, _ = io.Copy(io.Discard, body)
	return body.Close()
}

File: utils/dynamicip/ifconfig_resolver.go [hunk 2]
```diff
@@ -31,7 +31,12 @@ func (r *ifConfigResolver) Resolve(ctx context.Context) (netip.Addr, error) {
...
 	}
-	defer resp.Body.Close()
+	defer func() {
+		// Avoid sending unnecessary RST_STREAM and PING frames by ensuring the whole body is read.
+		// See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
 
 	ipBytes, err := io.ReadAll(resp.Body)
```
joshua-kim, line 38: We already read until `EOF` here - do we still need to drain the response body?

tests/fixture/tmpnet/check_monitoring.go
joshua-kim, line 1: Same comment in this diff in this package w.r.t `io.ReadAll`

api/metrics/client.go
joshua-kim, line 1: Do we need this client behavior update in this file? This is only used in e2e tests so I feel like this is over-kill for an environment that is just a smoke test.

utils/rpc/json.go
```diff
@@ -49,15 +50,27 @@ func SendJSONRequest(
 
 	// Return an error for any non successful status code
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		// Avoid sending unnecessary RST_STREAM and PING frames by ensuring the whole body is read.
+		// See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive
+		_, _ = io.Copy(io.Discard, resp.Body)
```
joshua-kim: If the server is malicious they can dump an infinite length response body into us. We're not in danger of OOM'ing because `io.Copy` allocates a fixed buffer, but we could spin indefinitely. I think a more robust strategy might be to drain the response body if we can, or send a `RST_STREAM` to the server if the caller cancels our context.

api/metrics/client.go
michaelkaplan13: (duh) thanks.

It took a couple `nolint`s because the linter get confused between some combination of the close being in helper function and needing to called from multiple places, but AFAICT it's still the preferred option to keep the exact error handling behavior.

api/metrics/client.go
```diff
@@ -53,16 +56,17 @@ func (c *Client) GetMetrics(ctx context.Context) (map[string]*dto.MetricFamily,
...
 	metrics, err := parser.TextToMetricFamilies(resp.Body)
 	if err != nil {
 		// Drop any error during close to report the original error
-		_ = resp.Body.Close()
+		_ = rpc.CleanlyCloseBody(resp.Body)
 		return nil, err
 	}
-	return metrics, resp.Body.Close()
+
+	return metrics, rpc.CleanlyCloseBody(resp.Body)
```
maru-ava: (No action required) Is there ever a useful return value from `resp.Body.Close()`? We ignore it in defers, ignore it in previous returns in this function, and yet here we return it.

tests/fixture/tmpnet/check_monitoring.go
```diff
@@ -114,7 +115,7 @@ func queryLoki(
 	if err != nil {
 		return 0, stacktrace.Errorf("failed to execute request: %w", err)
 	}
-	defer resp.Body.Close()
+	defer func() { _ = rpc.CleanlyCloseBody(resp.Body) }()
```
maru-ava: (No action required) I'm guessing the errcheck linter has specific exclusions for ignoring the result of`resp.Body.Close()` since it's such a common pattern to ignore the error. Without that exclusion, it appears our options are a function wrapper to enable ignoring the error or `//nolint:errcheck` here and `//nolint:bodyclose` where the response is created. Is this function wrapping pattern common enough for future maintainers to understand it's import, or would the explicit directives provide more clarity?

maru-ava (approved):

api/metrics/client.go
michaelkaplan13: > Is there ever a useful return value from resp.Body.Close()?

Practically speaking, not that I'm aware of.

Theoretically speaking, there could be custom implementations of `io.ReadCloser`, so it's possible. I found it annoying here since it prohibits a single deferred call, but didn't want to change any behavior unrelated to the purpose of the PR.

utils/rpc/json.go
```diff
@@ -7,12 +7,21 @@ import (
...
 	rpc "github.com/gorilla/rpc/v2/json2"
 )
 
+// CleanlyCloseBody avoids sending unnecessary RST_STREAM and PING frames by ensuring
+// the whole body is read before being closed.
+// See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive
+func CleanlyCloseBody(body io.ReadCloser) error {
+	_, _ = io.Copy(io.Discard, body)
+	return body.Close()
+}
```
StephenButtolph: I remember when first adding the error returning logic to the prod clients that it was kind of "we should probably return the error if one happens"... But honestly, if the client was able to get the response... I feel like dropping the error isn't unreasonable.

Should we just ignore this everywhere and make this function not return an error? We could ignore the error once here, and then everywhere just `defer CleanlyCloseBody(resp.Body)`
```suggestion
// CleanlyCloseBody avoids sending unnecessary RST_STREAM and PING frames by
// ensuring the whole body is read before being closed.
// See https://blog.cloudflare.com/go-and-enhance-your-calm/#reading-bodies-in-go-can-be-unintuitive
func CleanlyCloseBody(body io.ReadCloser) {
	_, _ = io.Copy(io.Discard, body)
	_ = body.Close()
}
```

michaelkaplan13: This seems very reasonable and I'm happy to make this change because it cleans up the code, though I do think it's a potential behavior change outside the scope of the intention of this PR.

joshua-kim: Related to my testing, but `io.Copy` can return a cancelation error since it respects the context - so we might want to bubble up that error instead.

api/metrics/client.go
michaelkaplan13: I don't think it makes a difference either way practically speaking, but I think applying best practices everywhere is better than not, particularly now that we're just uniformly deferring the call to the helper everywhere (which I realize is a change from when this comment was left). Thoughts?

tests/fixture/tmpnet/check_monitoring.go
michaelkaplan13: Similarly, now that we have the `CleanlyCloseBody` helper, I think it's most defensive to use it uniformly across the board rather than depending on the fact that the body is always fully read otherwise.

utils/dynamicip/ifconfig_resolver.go
michaelkaplan13: Now that we have the CleanlyCloseBody helper, I think it's most defensive to use it uniformly across the board rather than depending on the fact that the body is always fully read otherwise.

api/metrics/client.go
```diff
@@ -45,24 +47,23 @@ func (c *Client) GetMetrics(ctx context.Context) (map[string]*dto.MetricFamily,
...
 	var parser expfmt.TextParser
 	metrics, err := parser.TextToMetricFamilies(resp.Body)
 	if err != nil {
-		// Drop any error during close to report the original error
-		_ = resp.Body.Close()
 		return nil, err
 	}
-	return metrics, resp.Body.Close()
+
+	return metrics, nil
```
StephenButtolph:
```suggestion
	return parser.TextToMetricFamilies(resp.Body)
```

PR #4482, opened 2025-11-07
Unsync LevelDB Compactions Across Nodes

samliok:
## Why this should be merged

We have noticed synchronous spikes in LevelDB compactions across our nodes. This potentially causes network blips if Disk IO or CPU resources get overconsumed by many nodes in the network at the same time.

Screenshot 2025-11-07 at 11 21 19 AM

## How this works

Our nodes are configured by using the default `levelDB` configuration. This may be the reason for why we are seeing synchrony in compactions. We update this by deterministically randomizing two levelDB config parameters when levelDB is set up. These parameters should provide enough variance to offset the compaction synchrony. 

`CompactionL0Trigger` ->  The default for this is 4. This means compaction for level-0 tables will start when 4 L0 files are created. We set this value to be a random value(seeded by nodeID) from 3-7. By changing this value we get variance in when L0 compactions are triggered.

`CompactionTableSize` -> The default value for this is 2MiB. It defines the target size for files produced by compaction. By increasing this, compaction runs less often but consumes more cpu & I/O. We now randomize this value from `2MiB - 4MiB`

## How this was tested

I spun up a mainnet node on an EC2 instance. Initially it was created with the default compaction config for levelDB, but then I restarted the node with two different sets of configs. One with `CompactionTrigger` to `3` and Size to `4MiB` and then another restart with a `CompationL0Trigger` of 7 and Size of `3MiB`. The node appeared healthy after the restart.

### DIFF
--- a/database/leveldb/db.go
+++ b/database/leveldb/db.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"math/rand"
 	"slices"
 	"sync"
 	"time"
@@ -60,6 +61,18 @@ const (
 	// levelDBByteOverhead is the number of bytes of constant overhead that
 	// should be added to a batch size per operation.
 	levelDBByteOverhead = 8
+
+	// minCompactionL0Trigger is the minimum value for CompactionL0Trigger.
+	minCompactionL0Trigger = 3
+
+	// maxCompactionL0Trigger is the maximum value for CompactionL0Trigger.
+	maxCompactionL0Trigger = 7
+
+	// minCompactionTableSize is the minimum value for CompactionTableSize.
+	minCompactionTableSize = 2 * opt.MiB
+
+	// maxCompactionTableSize is the maximum value for CompactionTableSize.
+	maxCompactionTableSize = 4 * opt.MiB
 )
 
 var (
@@ -98,40 +111,40 @@ type config struct {
 	// BlockSize is the minimum uncompressed size in bytes of each 'sorted table'
 	// block.
 	//
-	// The default value is 4KiB.
+	// The default value is [opt.DefaultBlockSize].
 	BlockSize int `json:"blockSize"`
 	// CompactionExpandLimitFactor limits compaction size after expanded.
 	// This will be multiplied by table size limit at compaction target level.
 	//
-	// The default value is 25.
+	// The default value is [opt.DefaultCompactionExpandLimitFactor].
 	CompactionExpandLimitFactor int `json:"compactionExpandLimitFactor"`
 	// CompactionGPOverlapsFactor limits overlaps in grandparent (Level + 2)
 	// that a single 'sorted table' generates.  This will be multiplied by
 	// table size limit at grandparent level.
 	//
-	// The default value is 10.
+	// The default value is [opt.DefaultCompactionGPOverlapsFactor].
 	CompactionGPOverlapsFactor int `json:"compactionGPOverlapsFactor"`
 	// CompactionL0Trigger defines number of 'sorted table' at level-0 that will
 	// trigger compaction.
 	//
-	// The default value is 4.
+	// The default value is a random value between [minCompactionL0Trigger] and [maxCompactionL0Trigger].
 	CompactionL0Trigger int `json:"compactionL0Trigger"`
 	// CompactionSourceLimitFactor limits compaction source size. This doesn't apply to
 	// level-0.
 	// This will be multiplied by table size limit at compaction target level.
 	//
-	// The default value is 1.
+	// The default value is [opt.DefaultCompactionSourceLimitFactor].
 	CompactionSourceLimitFactor int `json:"compactionSourceLimitFactor"`
 	// CompactionTableSize limits size of 'sorted table' that compaction generates.
 	// The limits for each level will be calculated as:
 	//   CompactionTableSize * (CompactionTableSizeMultiplier ^ Level)
 	// The multiplier for each level can also fine-tuned using CompactionTableSizeMultiplierPerLevel.
 	//
-	// The default value is 2MiB.
+	// The default value is a random value between [minCompactionTableSize] and [maxCompactionTableSize].
 	CompactionTableSize int `json:"compactionTableSize"`
 	// CompactionTableSizeMultiplier defines multiplier for CompactionTableSize.
 	//
-	// The default value is 1.
+	// The default value is [opt.DefaultCompactionTableSizeMultiplier].
 	CompactionTableSizeMultiplier float64 `json:"compactionTableSizeMultiplier"`
 	// CompactionTableSizeMultiplierPerLevel defines per-level multiplier for
 	// CompactionTableSize.
@@ -145,11 +158,11 @@ type config struct {
 	// The multiplier for each level can also fine-tuned using
 	// CompactionTotalSizeMultiplierPerLevel.
 	//
-	// The default value is 10MiB.
+	// The default value is [opt.DefaultCompactionTotalSize].
 	CompactionTotalSize int `json:"compactionTotalSize"`
 	// CompactionTotalSizeMultiplier defines multiplier for CompactionTotalSize.
 	//
-	// The default value is 10.
+	// The default value is [opt.DefaultCompactionTotalSizeMultiplier].
 	CompactionTotalSizeMultiplier float64 `json:"compactionTotalSizeMultiplier"`
 	// DisableSeeksCompaction allows disabling 'seeks triggered compaction'.
 	// The purpose of 'seeks triggered compaction' is to optimize database so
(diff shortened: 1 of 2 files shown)


database/leveldb/db.go
```diff
@@ -185,8 +188,32 @@ type config struct {
 	MetricUpdateFrequency time.Duration `json:"metricUpdateFrequency"`
 }
 
+// randomizeCompaction sets compaction parameters in cfg to values derived
+// from nodeID to spread out compaction behavior across different nodes.
+func randomizeCompaction(nodeID ids.NodeID, cfg *config) {
```
rkuris: nit: It's not truly random, so maybe compactionParamsFromNodeId would be a better name

database/leveldb/db.go
```diff
@@ -196,6 +223,9 @@ func New(file string, configBytes []byte, log logging.Logger, reg prometheus.Reg
 		MaxManifestFileSize:    DefaultMaxManifestFileSize,
 		MetricUpdateFrequency:  DefaultMetricUpdateFrequency,
 	}
+
+	randomizeCompaction(nodeID, &parsedConfig)
```
rkuris: may want to mention here that configuration will override the returns from here

database/leveldb/db.go
```diff
@@ -185,8 +188,32 @@ type config struct {
...
+	bytes := nodeID.Bytes()
+
+	// sanity check
+	if len(bytes) < 8 {
+		paddedBytes := make([]byte, 8)
+		copy(paddedBytes[8-len(bytes):], bytes)
+		bytes = paddedBytes
+	}
+	seed := binary.BigEndian.Uint64(bytes)
+	r := rand.New(rand.NewSource(int64(seed)))
```
rkuris: nit: I would have pulled out the creation of a random source from a NodeId into a separate, testable function

rkuris (approved):
Approved with suggestions / nits

database/leveldb/db.go
samliok: added tests for this function + renamed

database/leveldb/db_test.go
```diff
@@ -74,3 +75,28 @@ func BenchmarkInterface(b *testing.B) {
...
+	t.Run("same nodeID results in same compaction params", func(t *testing.T) {
+		id := ids.GenerateTestNodeID()
+		var cfg1, cfg2 config
+		setCompactionParamsFromNodeID(id, &cfg1)
+		setCompactionParamsFromNodeID(id, &cfg2)
+
+		require.Equal(t, cfg1.CompactionL0Trigger, cfg2.CompactionL0Trigger)
+		require.Equal(t, cfg1.CompactionTableSize, cfg2.CompactionTableSize)
+	})
+}
```
rkuris, line 86: nit: maybe a test that 100 of them don't all generate exactly the same configuration?

database/leveldb/db.go
```diff
@@ -60,6 +63,18 @@ const (
 	// levelDBByteOverhead is the number of bytes of constant overhead that
 	// should be added to a batch size per operation.
 	levelDBByteOverhead = 8
+
+	// minCompactionL0Trigger is the minimum value for CompactionL0Trigger.
```
yacovm, line 65: Have we observed that two of these parameters are needed to be changed at the same time? 

Have we seen what happens when only one of them is changed?

database/leveldb/db.go
```diff
@@ -185,8 +200,24 @@ type config struct {
...
+	// generate random defaults for CompactionTableSize
+	compactionTableSize := (r.Intn(maxCompactionTableSize-minCompactionTableSize+1) + minCompactionTableSize) * opt.MiB
+
+	cfg.CompactionL0Trigger = compactionL0Trigger
+	cfg.CompactionTableSize = compactionTableSize
+}
+
 // New returns a wrapped LevelDB object.
-func New(file string, configBytes []byte, log logging.Logger, reg prometheus.Registerer) (database.Database, error) {
+func New(nodeID ids.NodeID, file string, configBytes []byte, log logging.Logger, reg prometheus.Registerer) (database.Database, error) {
```
DracoLi: Does making this take a `NodeID` make sense? 
`db.leveldb` is a general kv db but this makes db initialization dependent on having a NodeID when this package is used.

database/leveldb/db.go
```diff
@@ -60,6 +63,18 @@ const (
 	// levelDBByteOverhead is the number of bytes of constant overhead that
 	// should be added to a batch size per operation.
 	levelDBByteOverhead = 8
+
+	// minCompactionL0Trigger is the minimum value for CompactionL0Trigger.
+	minCompactionL0Trigger = 3
+
+	// maxCompactionL0Trigger is the maximum value for CompactionL0Trigger.
```
DracoLi, line 68: Have we tested how much compactions will be spread out if we are randomizing these two values? Its not clear to me if it will spread out enough to make a big impact.
If the total size per level is still the same, won't nodes hit a level 4 or 5 compaction at around the same time when previous level is full due to similar db size? It should not be the exact same time due to L0 trigger config but since L0 size is small, it might still be very close together.

database/leveldb/db.go
rkuris: You just get less variance if you do this, but I think we're shooting for the maximum possible variance.

database/leveldb/db.go
rkuris: If we don't have a `NodeID` then we don't care about being out of sync. Or at least we don't know what we're being out of sync with.

This lets us maintain the exact same configuration on a restart with the same `NodeID` so IMO this is the right approach.

database/leveldb/db.go
rkuris: I think we'd have to see it in production to know for sure, but this is the best we can do so it's worth a shot.

database/leveldb/db.go
yacovm: I get that, but if (for example) variable A is responsible for 99% of the variance and variable B is responsible for 1% of it then it makes sense to only tune A.

rkuris: Based on the algorithm, it shouldn't be that skewed, unless something is wrong with the implementation. I'm not sure this is worth testing to verify though, since even if it is as bad as 99/1 there's not any harm in changing both.

database/leveldb/db.go
DracoLi: should this just take something like a seed int? factory can convert the NodeID to a seed when creating a leveldb. I think its kind of weird needing to pass ids.EmptyNodeID when creating a leveldb.

rkuris:
You might just want another constructor that doesn't randomize these
values. What is the use case?

database/leveldb/db.go
samliok: ended up removing the `NodeID` because I did find it weird requiring that to be a part of the constructor. Plus I don't think we _really_ care if the levelDB config is randomized on restart(as long as it different changes don't break the node, which seems like they don't).

database/factory/factory.go
StephenButtolph, line 1: The diff in this file should be reverted. (The argument naming fixes in the comments are arguably good... But the `if` statement doesn't make sense, and the file as a whole is otherwise unmodified).

database/leveldb/db_test.go
```diff
@@ -73,3 +73,18 @@ func BenchmarkInterface(b *testing.B) {
 		}
 	}
 }
+
+func TestSetCompactionParamsFromNodeId(t *testing.T) {
```
StephenButtolph: This name seems stale.

database/leveldb/db_test.go
```diff
@@ -73,3 +73,18 @@ func BenchmarkInterface(b *testing.B) {
...
+	t.Run("randomized compaction params within expected ranges", func(t *testing.T) {
```
StephenButtolph: Why are we `t.Run`ing one time in this test? Why don't we just directly run the test in the outer function?

database/leveldb/db_test.go
```diff
@@ -73,3 +73,18 @@ func BenchmarkInterface(b *testing.B) {
...
+		for range 100 {
```
StephenButtolph: I can see arguments in the other direction, but tbh I don't really think we need to run this 100 times... Unless we're going to try to inspect the averages/variance or something... We're already going to be running this a ton in CI anyways

database/leveldb/db.go
```diff
@@ -185,6 +198,21 @@ type config struct {
...
+
+	// generate random defaults for CompactionL0Trigger
+	compactionL0Trigger := r.Intn(maxCompactionL0Trigger-minCompactionL0Trigger+1) + minCompactionL0Trigger
+
+	// generate random defaults for CompactionTableSize
+	compactionTableSize := (r.Intn(maxCompactionTableSize-minCompactionTableSize+1) + minCompactionTableSize)
+
+	cfg.CompactionL0Trigger = compactionL0Trigger
+	cfg.CompactionTableSize = compactionTableSize
+}
```
StephenButtolph, line 212: Few questions/comments: 
1. Why do we make a new source here? I think we should just use [rand.Intn](https://pkg.go.dev/math/rand#Intn) directly.
2. The in-line comments don't seem too insightful.
3. Having the arguments as constants makes it a bit more clear on exactly what is being passed into the functions imo.

```suggestion
// randomizeCompactionParams sets compaction parameters in cfg to random values
// to spread out compaction behavior across different nodes.
//
// #nosec G404 // non-crypto randomness is fine here
func randomizeCompactionParams(cfg *config) {
	const (
		rangeCompactionL0Trigger = maxCompactionL0Trigger - minCompactionL0Trigger + 1
		rangeCompactionTableSize = maxCompactionTableSize - minCompactionTableSize + 1
	)
	cfg.CompactionL0Trigger = rand.Intn(rangeCompactionL0Trigger) + minCompactionL0Trigger
	cfg.CompactionTableSize = rand.Intn(rangeCompactionTableSize) + minCompactionTableSize
}
```

database/leveldb/db.go
StephenButtolph: > which seems like they don't

This is a requirement for merging anyways, as we are going to be deploying this on nodes that are already running a different config.

StephenButtolph:
This PR did not meaningfully change the compaction cadence for the nodes running this version.

PR #4485, opened 2025-11-10
feat(evm): add database with separate storage for block data

### LINKED ISSUE #4491: Add evm database that supports separate storage for block data
## Summary

Add an EVM database wrapper that routes block-related data into `heightindexdb` while keeping non-block data on the existing KV store. This is intended to reduce KV usage/compaction pressure.

## Scope

- Block routing:
  - Wrap the existing EVM `ethdb.Database`.
  - Route block headers, bodies, and receipts to height-indexed databases.
  - Keep all non-block data on the underlying KV store.
- Integration with `x/blockdb`:
   - Store headers/bodies/receipts in dedicated height-indexed stores managed by `x/blockdb`.
- Migration:
  - Provide a mechanism to migrate existing canonical block data from the KV store into the new height-indexed databases.
  - Migration should be safe to run in the background and resilient across restarts.

DracoLi:
## Why this should be merged

Adds a `ethdb.Database` implementation that moves EVM block data out of the shared KV store into dedicated height-indexed databases optimized for block storage, reducing KV store size and compaction cost.

- Headers, bodies, and receipts are stored in `heightindexdb`; all other data remains in the underlying KV store.
- A decrease in total LevelDB/PebbleDB size and total compaction time is expected. Detailed metrics will be tracked in #4523.

## How this works

The `blockdb.Database` wraps `ethdb.Database` and routes block data by key prefix, while leaving non-block data in the KV store.

- Header/body/receipt keys for blocks with height ≥ the configured minimum are written to height-indexed DBs; below that threshold they use the KV store.
- At most one block per height is stored. Writing at an occupied height overwrites the previous value and is not an intended use case. Deletes of block data are no-ops.
- If a block isn’t yet in the height-indexed DBs and migration isn’t complete, reads fall back to the KV store.
- Block-related keys bypass the KV batch and write directly to the height-indexed DBs; non-block keys are batched on the underlying KV store.
- Deferred init: With `allowDeferredInit`, initialization can wait until the minimum block height is known (e.g., when state sync is enabled). Database min height is persisted and cannot be changed without recreating the block databases.

### Migration

Separated to https://github.com/ava-labs/avalanchego/pull/4771

## How this was tested

- Unit test for routing and migration behavior.
- Two Mainnet nodes (with state sync enabled and disabled) with `blockdb` enabled running for >4 weeks.

No


vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+// New defers initializing the height-indexed block databases until
+// [Database.InitBlockDBs] is called.
+//
+// The bool result is true if the block databases were initialized immediately,
+// and false if initialization was deferred.
+func New(
+	stateDB database.Database,
+	evmDB ethdb.Database,
+	dbPath string,
+	allowDeferredInit bool,
```
ARR4N, line 139: Why might someone want to defer? Why would the _consumer_ know the minimum block height later, but not now? I'm thinking that this could instead be `defaultMinimBlockHeight uint64`, which you use instead of `1` when the current argument is `false`.

The `InitBlockDBs` method could then be unexported and the `heightsDBReady bool` field (which feels like a race-condition waiting to happen) removed.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/heightindexdb/meterdb"
+	"github.com/ava-labs/avalanchego/database/prefixdb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	heightindexdb "github.com/ava-labs/avalanchego/x/blockdb"
+)
+
+var (
```
ARR4N, line 26: Although it's idiomatic in other languages, it's not a requirement in Go to declare all globals at the top like this. The primary drawback is that they lack context relative to where they're used so a reader has to jump back and forth between declaration and usage sites.

Please move the declarations (perhaps still in groups, but also feel free to break them up if warranted) closer to where they're used.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	_ ethdb.Database = (*Database)(nil)
+	_ ethdb.Batch    = (*batch)(nil)
+
+	migratorDBPrefix = []byte("migrator")
```
ARR4N: Do we want to prefix our prefixes to guarantee that they never clash with a geth value? Like "ava-migrator"?

Is there a downside to doing that? FWIW they use `b` for block bodies and `blt-` for something else so we don't need to worry about a substring clash and starting with `a` is (at first glance) safe.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	errUnexpectedKey        = errors.New("unexpected database key")
+	errNotInitialized       = errors.New("database not initialized")
+	errAlreadyInitialized   = errors.New("database already initialized")
+	errInvalidEncodedLength = errors.New("invalid encoded length")
+)
+
+// Key prefixes for block data in [ethdb.Database].
+// This is copied from libevm because they are not exported.
+// Since the prefixes should never be changed, we can avoid libevm changes by
+// duplicating them here.
```
ARR4N: Agreed.

Just noting in case any other reviewer asks me about libevm.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+var (
+	evmHeaderPrefix    = []byte("h")
+	evmBlockBodyPrefix = []byte("b")
+	evmReceiptsPrefix  = []byte("r")
+)
+
+const (
+	hashDataElements = 2
```
ARR4N:
```suggestion
```

I don't think this particular constant adds much value, but it does force the reader to go and check it. The two below are feasibly configurable, but the number of elements is a property of the implementation because `readHashAndData()` (the only place it's used) is just the inverse of `writeHashAndData()`, which has 2 arguments of interest.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	blockNumberSize  = 8
+	blockHashSize    = 32
+
+	headerDBName   = "headerdb"
+	bodyDBName     = "bodydb"
+	receiptsDBName = "receiptsdb"
```
ARR4N:
```suggestion
```

These are each only ever used in a single place so I'd recommend removing them too. If it becomes necessary to have them coupled somewhere in the future (e.g. in a test) then they can be abstracted, but for now they're [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+var (
+	evmHeaderPrefix    = []byte("h")
+	evmBlockBodyPrefix = []byte("b")
+	evmReceiptsPrefix  = []byte("r")
+)
+
+const (
+	hashDataElements = 2
+	blockNumberSize  = 8
+	blockHashSize    = 32
```
ARR4N: (fyi) To better explain my rationale for removing the other constants, I think these are useful because they couple multiple usage sites that all MUST match.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+)
+
+// Database wraps an [ethdb.Database] and routes block headers, bodies, and receipts
+// to separate [database.HeightIndex] databases for blocks at or above the minimum height.
+// All other data uses the underlying [ethdb.Database] directly.
+type Database struct {
+	ethdb.Database
+
+	// Databases
+	stateDB    database.Database
```
ARR4N:
```suggestion
	disk    database.Database
```

The concept of a "state DB" has very strong connotations in EVM code (i.e. [`state.StateDB`](https://pkg.go.dev/github.com/ava-labs/libevm/core/state#StateDB)). This is more akin to `ethdb.Database`, which is often called `disk` in or just `db` geth code.

The specific name in my suggestion is up for discussion, but I do feel quite strongly that `stateDB` is too loaded a term.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+		return db, true, nil
+	}
+
+	// Initialize with min height 1 if deferred initialization is not allowed.
+	if !allowDeferredInit {
+		if err := db.InitBlockDBs(1); err != nil {
+			return nil, false, err
+		}
+		return db, true, nil
+	}
```
ARR4N: It took a non-trivial amount of analysis to convince myself that this was a repeating pattern, and I think it's cleaner to be explicit about the fact that you're trying various means of finding a minimum height.

```suggestion
	minHeightSrcs := [](func() (uint64, bool, error)){
		func() (uint64, bool, error) {
			// If databases already exist, load with existing min height.
			return databaseMinHeight(db.stateDB)
		},
		func() (uint64, bool, error) {
			// Initialize using the minimum block height of existing blocks to migrate.
			return minBlockHeightToMigrate(evmDB)
		},
		func() (uint64, bool, error) {
			return 1, !allowDeferredInit, nil
		},
	}
	for _, fn := range minHeightSrcs {
		h, ok, err := fn()
		if err != nil {
			return nil, false, err
		}
		if !ok {
			continue
		}
		if err := db.InitBlockDBs(h); err != nil {
			return nil, false, err
		}
		return db, true, nil
	}
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	)
+	return db, false, nil
+}
+
+// InitBlockDBs initializes [database.HeightIndex] databases with the specified
+// minimum height.
+// Once initialized, the minimum height cannot be changed without recreating
+// the databases.
+//
+// Returns an error if already initialized.
```
ARR4N, line 188: If you accept a default minimum block height instead of a boolean signalling allowed deferral then this becomes impossible.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	// Don't close stateDB since the caller should be managing it.
+	return errors.Join(
+		db.headerDB.Close(),
+		db.bodyDB.Close(),
+		db.receiptsDB.Close(),
+		db.Database.Close(),
+	)
+}
+
+func (db *Database) initMigrator() error {
```
ARR4N: I think you've ordered methods by exported and then unexported, but this makes it harder on the reader. I had to jump from `InitBlockDBs()` down to here and then back up to `StartMigration()` to understand the setup, and then into `migration.go` to understand the usage.

Grouping identifiers and blocks by usage, and ordering them by either order of usage or order of least to most dependent on other parts, makes it far easier to understand the code.

I haven't dug very far in yet, but I think a good first fix would be to move `initMigrator()` and `StartMigration()` (in that order) to the top of the `migrator.go` file.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	if db.migrator != nil {
+		return nil
+	}
```
ARR4N: I don't think this is necessary at all, but if it were to be kept I think it's better to return an error. While it's very tempting to try and protect against invalid usage, I've experienced many occasions where that actually hides a bug that should have been surfaced instead of silently avoided.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+
+	return nil
+}
+
+// StartMigration begins the background migration of block data from the
+// [ethdb.Database] to the height-indexed block databases.
+//
+// Returns an error if the databases are not initialized.
+// No error if already running.
+func (db *Database) StartMigration() error {
```
ARR4N: Accept a `context.Context` here and plumb it through to avoid `migrator.beginRun()` using `context.Background()`.

[More information](https://google.github.io/styleguide/go/decisions.html#:~:text=Note%3A%20It%20is%20very%20rare%20for%20code%20in%20the%20middle%20of%20a%20callchain%20to%20require%20creating%20a%20base%20context%20of%20its%20own%20using%20context.Background().).

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	if !db.heightDBsReady {
+		return errNotInitialized
+	}
+	db.migrator.start()
+	return nil
+}
+
+func (db *Database) Put(key []byte, value []byte) error {
+	if !db.useHeightIndexedDB(key) {
```
ARR4N: This method is used 4 times here but only introduced ~140 lines later so I'd move it to immediately before `Put()`. It doesn't rely on any other logic and its introduction is motivated by the very next method. The current placement forced me to go searching. The same applies to `heightDBForKey()`.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	return readHashAndData(heightDB, db.Database, key, db.migrator)
+}
+
+func (db *Database) Has(key []byte) (bool, error) {
+	if !db.useHeightIndexedDB(key) {
+		return db.Database.Has(key)
+	}
+
+	_, err := db.Get(key)
+	if err != nil {
```
ARR4N:
```suggestion
	if _, err := db.Get(key); err != nil {
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+// Delete removes the key from the underlying database for non-block data.
+// Block data deletion is a no-op because [database.HeightIndex] does not support deletion.
+func (db *Database) Delete(key []byte) error {
+	if !db.useHeightIndexedDB(key) {
+		return db.Database.Delete(key)
+	}
+
+	num, hash, err := parseBlockKey(key)
+	if err != nil {
+		return err
```
ARR4N: Log a warning here and return `nil` otherwise this goes from being a no-op to being an error that could bubble up and cause other problems.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	db.logger.Warn(
+		"Deleting block data is a no-op",
+		zap.Uint64("height", num),
+		zap.Stringer("hash", hash),
+	)
+	return nil
+}
+
+func (db *Database) Close() error {
+	if db.migrator != nil {
```
ARR4N: This can never be the case when using the public API so the check is unnecessary and may also hide an implementation bug (see my similar comment on `initMigrator()`).

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	case isHeaderKey(key):
+		n = len(evmHeaderPrefix)
+	case isReceiptsKey(key):
+		n = len(evmReceiptsPrefix)
+	default:
+		return false
+	}
+	num := binary.BigEndian.Uint64(key[n : n+blockNumberSize])
+	return num >= db.minHeight
+}
```
ARR4N: There's a lot of duplicated code here and you regularly call `heightDBForKey()` if `useHeightIndexedDB()` returns `true`. Combining them simplifies the implementation and some of the call sites, and the only "waste" on the `false` path is the `hdb` variable. Anywhere that only uses `useHeightIndexedDB()` can swap to `if _, ok := db.heightDBForKey(); ok {` (or `!ok`).

```go
func (db *Database) heightDBForKey(key []byte) (database.HeightIndex, bool) {
	if !db.heightDBsReady {
		return nil, false
	}

	var (
		prefix []byte
		hdb    database.HeightIndex
	)
	switch {
	case isBodyKey(key):
		prefix = evmBlockBodyPrefix
		hdb = db.bodyDB
	case isHeaderKey(key):
		prefix = evmHeaderPrefix
		hdb = db.headerDB
	case isReceiptsKey(key):
		prefix = evmReceiptsPrefix
		hdb = db.receiptsDB
	default:
		return nil, false
	}

	n := len(prefix)
	blockNum := binary.BigEndian.Uint64(key[n : n+blockNumberSize])
	if blockNum < db.minHeight {
		return nil, false
	}
	return hdb, true
}
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+	return bytes.HasPrefix(key, evmHeaderPrefix)
+}
+
+func isReceiptsKey(key []byte) bool {
+	if len(key) != len(evmReceiptsPrefix)+blockNumberSize+blockHashSize {
+		return false
+	}
+	return bytes.HasPrefix(key, evmReceiptsPrefix)
+}
```
ARR4N:
```suggestion
func isKeyForPrefix(key, prefix []byte) bool {
	if len(key) != len(prefix)+blockNumberSize+blockHashSize {
		return false
	}
	return bytes.HasPrefix(key, prefix)
}

func isBodyKey(key []byte) bool {
	return isKeyForPrefix(key, evmBlockBodyPrefix)
}

func isHeaderKey(key []byte) bool {
	return isKeyForPrefix(key, evmHeaderPrefix)
}

func isReceiptsKey(key []byte) bool {
	return isKeyForPrefix(key, evmReceiptsPrefix)
}
```

vms/evm/database/blockdb/database_test.go
```diff
@@ -0,0 +1,526 @@
...
+	"github.com/ava-labs/avalanchego/database/leveldb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	evmdb "github.com/ava-labs/avalanchego/vms/evm/database"
+	heightindexdb "github.com/ava-labs/avalanchego/x/blockdb"
+)
+
+func TestMain(m *testing.M) {
+	customtypes.Register()
+	params.RegisterExtras()
```
ARR4N: Why are these needed?

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	case isHeaderKey(key):
+		n = len(evmHeaderPrefix)
+	case isReceiptsKey(key):
+		n = len(evmReceiptsPrefix)
+	default:
+		return 0, common.Hash{}, errUnexpectedKey
+	}
+	num = binary.BigEndian.Uint64(key[n : n+blockNumberSize])
+	bytes := key[n+blockNumberSize:]
+	if len(bytes) != blockHashSize {
```
ARR4N: This can never happen because all of the `is*Key()` functions check the length. 

I realised this because I thought it was unsafe to take a range into the `key` slice, wrote a test, and it didn't panic. I think this signals that the checks are too far away from here.

vms/evm/database/blockdb/database.go
ARR4N: When I got to `parseBlockKey()` I noticed that there was further overlap; I think you can collapse all 3. I've pushed my [review/experiment branch](https://github.com/ava-labs/avalanchego/compare/2c6713d11a7aed2d849c8eb5d5f6c64366bd14bc...arr4n/dl/evm-blockdb) for you to see what I mean.

Note that everything in there is either `func parseBlockKey()` (very similar logic) or `func (*Database) parseBlockKey()`, which has the added functionality of selecting the correct database while parsing.

vms/evm/database/blockdb/database.go
DracoLi: agreed. how about `metaDB` or `configDB`. this is used for storing database meta so that might be more appropriate.

vms/evm/database/blockdb/database.go
DracoLi: this is the prefix on the `stateDB`, which is different than the `evmDB`, which contains the evm data. The `evmDB` currently has the `ethedb` prefix and clients should be passing a different prefixed db when creating the `blockdb.Database` (See a draft usage of this database usage [here](https://github.com/ava-labs/avalanchego/pull/4627/files#diff-9eec6fc986677ba415f585b88b7487d622715323e58d9097cd52d77138c94ee4R132-R133))

Current convention seems to be having the caller creating the prefixdb. But alternatively, we can also just wrap the stateDB with prefix db in `New` if you think that might be clearer.

vms/evm/database/blockdb/database.go
DracoLi: So for context, I added this to address the state sync use case.

Currently, evm database is created on vm initialization. At that point, if state sync is enabled, we don't know what the height of the first block we will be fetching until we get the state sync summary. While we can just set min height to 1, this allows us to init the block dbs once we know the first block height we need to store is and call `InitBlockDBs` manually (See the [an example here](https://github.com/ava-labs/avalanchego/blob/a250a313adabf2788701a82594d4b9c1b05c44d6/graft/coreth/sync/blocksync/syncer.go#L79-L87))

I'm open to alternative solutions. The main goal here is to being able to set the min height of the dbs to the first non genesis block we need when state sync is enabled to reduce the size of the db index file. Alternatively, we can also not do that and just use min height = 1. This will lead to the db index files being few GBs larger though.

vms/evm/database/blockdb/database.go
DracoLi: This is an interesting way to rewrite this that I have never thought about. thx for this suggestion, I agree this is easier to read.

vms/evm/database/blockdb/database_test.go
DracoLi: This is needed by the `core.GenerateChainWithGenesis` call in `createBlocksToAddr` in `helpers_test.go`.
Without them we get a nil pointer error on `coreth/params.GetExtra` and `coreth/plugin/evm/customtypes.GetHeaderExtra`

vms/evm/database/blockdb/database.go
DracoLi: thx for providing that example. definitely agreed. i think we can clean up a lot by having a `parseKey` function like your draft. I'll steal that idea and update.

vms/evm/database/blockdb/database.go
DracoLi: agreed, its not possible to trigger this under normal usage. I'll remove it

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,528 @@
...
+	"github.com/ava-labs/libevm/rlp"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/heightindexdb/meterdb"
+	"github.com/ava-labs/avalanchego/database/prefixdb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	heightindexdb "github.com/ava-labs/avalanchego/x/blockdb"
```
DracoLi, line 23: Note for reviewers: `x/blockdb` is planned to be moved to `database/heightindexdb` (see #4520). 
I renamed it here to avoid confusing this evm database package (`blockdb`) with the current `x/blockdb` package.

vms/evm/database/blockdb/database.go
DracoLi: yeah, I agree the current way I'm ordering makes top-down reviewing hard. I've reorganized things to have most of the least dependent helpers first, then methods that use them later, which should help readers. 
It's a bit tricky since there are some interconnected methods, please take a look again and let me know what you think.

vms/evm/database/blockdb/database_test.go
DracoLi: I've updated block creation to not use `core.GenerateChainWithGenesis` so we can remove this and also not import from graft/coreeth.
https://github.com/ava-labs/avalanchego/pull/4485/commits/2561499c60de64df05f84bae75509c1523c06767

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
...
+	errInvalidEncodedLength = errors.New("invalid encoded length")
+)
+
+var _ ethdb.Database = (*Database)(nil)
+
+// Database wraps an [ethdb.Database] and routes block headers, bodies, and receipts
+// to separate [database.HeightIndex] databases for blocks at or above the minimum height.
+// All other data uses the underlying [ethdb.Database] directly.
+type Database struct {
+	ethdb.Database
```
alarso16, line 35: I don't think you should embed this interface, seems prone to misuse. Do you have a strong reason to?

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
...
+
+	heightindexdb "github.com/ava-labs/avalanchego/x/blockdb"
+)
+
+var (
+	errAlreadyInitialized   = errors.New("database already initialized")
+	errInvalidEncodedLength = errors.New("invalid encoded length")
+)
+
+var _ ethdb.Database = (*Database)(nil)
```
alarso16: If you're embedding the interface, I don't think you need this

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
...
+
+	return nil
+}
+
+// Key prefixes for block data in [ethdb.Database].
+// This is copied from libevm because they are not exported.
+// Since the prefixes should never be changed, we can avoid libevm changes by
+// duplicating them here.
+const (
+	evmHeaderPrefix    = 'h'
```
alarso16: nit: Everything is prefixed with `evm`, but that seems obvious since you're implementing this in the `evm` package. Maybe remove this unless you're matching libevm?

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
...
+		func() (uint64, bool, error) {
+			// Load existing database min height.
+			return databaseMinHeight(db.metaDB)
+		},
+		func() (uint64, bool, error) {
+			// Use min height 1 unless deferring initialization.
+			return 1, !allowDeferredInit, nil
+		},
+	}
+	for _, fn := range minHeightFn {
```
JonathanOppenheimer, line 163: This seems a little bit overly complicated -- is there a reason for a slice of functions? 

Something like 

```
  minHeight, exists, err := databaseMinHeight(db.metaDB)                                                                                                           
  if err != nil {                                                                                                                                                  
      return nil, false, err                                                                                                                                       
  }                                                                                                                                                                
  if !exists && !allowDeferredInit {                                                                                                                               
      minHeight = 1                                                                                                                                                
      exists = true                                                                                                                                                
  }                                                                                                                                                                
  if exists {                                                                                                                                                      
      if err := db.InitBlockDBs(minHeight); err != nil {                                                                                                           
          return nil, false, err                                                                                                                                   
      }                                                                                                                                                            
      return db, true, nil                                                                                                                                         
  }   
```

would seem to do the same and be more readable.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
...
+	return db, false, nil
+}
+
+// InitBlockDBs initializes [database.HeightIndex] databases with the specified
+// minimum height.
+// Once initialized, the minimum height cannot be changed without recreating
+// the databases.
+//
+// Returns an error if already initialized.
+func (db *Database) InitBlockDBs(minHeight uint64) error {
```
JonathanOppenheimer, line 189: there's no possibility of concurrent calls this function right?

vms/evm/database/blockdb/database.go
JonathanOppenheimer: Also this seems kinda sus -- copied from libevm because they're not exported - what if they change in libevm?

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,425 @@
+// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer, line 1: Some of the functions in this package could use godocs -- `func (b *batch) Put` particularly.

JonathanOppenheimer (commented):
A few comments/questions

vms/evm/database/blockdb/database.go
DracoLi: `ethdb.Database` implements a number of interfaces that we don't need to override. ie `Compacter`, `Snapshotter`, `Iteratee`, etc. Since the goal is only overriding the few methods that will affect block usage, i think embed is a good use case here. Thoughts?

vms/evm/database/blockdb/database.go
DracoLi: yep, that's right 👍  fixed https://github.com/ava-labs/avalanchego/pull/4485/changes/93c1534cc2304e21033435e50dd68220e714d7f6#diff-d2de0a916c8c9953392b7cd5e98c6be0c42d2d3f92615793857091b9a21d0adeL31-L32

vms/evm/database/blockdb/database.go
DracoLi: yeah i'm good with removing the evm prefix for these.

vms/evm/database/blockdb/database.go
DracoLi: I actually had this initially but @ARR4N suggested this func method as its easier to understand (https://github.com/ava-labs/avalanchego/pull/4485#discussion_r2568084924).
The slice of functions do seem a bit weird but I do think its cleaner in terms of communicating the intent, which is finding the min height.

vms/evm/database/blockdb/database.go
DracoLi: no, it will be called on vm init

vms/evm/database/blockdb/database.go
JonathanOppenheimer: Ah okay! That makes a lot of sense - I have a hard time reasoning through the func method, but Arran is far more talented than I am. going to resolve!

vms/evm/database/blockdb/database.go
DracoLi: > Also this seems kinda sus -- copied from libevm because they're not exported - what if they change in libevm?

these should really not be changed as that will mean we will need to migrate all stored block data. alternatively we can export these from libevm but that can make upgrading libevm a bit harder.

vms/evm/database/blockdb/database.go
DracoLi: right - i left some out like `Get` and `Put` since they are part of the Database interface and is pretty self explanatory. But I'm happy to add a simple doc to those so we cover all exported functions.

vms/evm/database/blockdb/database.go
JonathanOppenheimer: Ya understand they can't be for the sake of the database, -- if they'll never be changed in libevm it's good. I just thought it was a bit strange when reading it.

vms/evm/database/blockdb/database.go
DracoLi: https://github.com/ava-labs/avalanchego/pull/4485/commits/93c1534cc2304e21033435e50dd68220e714d7f6

vms/evm/database/blockdb/database.go
JonathanOppenheimer: I didn't see this comment earlier! that's my bad -- will resolve this and the other comment.

JonathanOppenheimer:
~~@DracoLi - nothing else from me - once all the outstanding comments are resolved, I will do another pass. This [comment](https://github.com/ava-labs/avalanchego/pull/4485#discussion_r2568089737) seems relevant~~ 

OOD

vms/evm/database/blockdb/database.go
DracoLi: for new nodes using state sync, we don't know the first height that we will state sync to at vm initialization. We will only know after we fetched the state sync summary afterwards. This is why I went with the deferred option.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+		db:    db,
+		Batch: db.Database.NewBatchWithSize(size),
+	}
+}
+
+// Put inserts the given value into the key-value data store.
+// Block data is written directly to the height-indexed databases, bypassing the batch.
+func (b *batch) Put(key []byte, value []byte) error {
+	if p, ok := b.db.parseKey(key); ok {
+		return p.writeHashAndData(value)
```
JonathanOppenheimer, line 424: `batch.Put` persists block header/body/receipt entries immediately via `p.writeHashAndData` instead of staging them in the batch. 

If a caller batches multi-key updates (e.g. block data + canonical mappings), a later `Reset`/`discard` (or a `Write` failure!) will leave the block data data (in this case) committed without the batch keys, which I believe could produce inconsistent chain state.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+
+// parseKey parses a block key into a parsedBlockKey.
+// It returns false if no block databases for the key prefix exist.
+func (db *Database) parseKey(key []byte) (*parsedBlockKey, bool) {
+	if !db.heightDBsReady {
+		return nil, false
+	}
+
+	var hdb database.HeightIndex
+	switch key[0] {
```
JonathanOppenheimer, line 281: I think this needs some more validation checks. `parseKey` switches on `key[0]` but you don't actually check `len(key) > 0` so calls like `Get(nil)` will panic once height-index DBs are initialized.

The wrapper should handle empty keys by returning false and delegating I think unless you have a better idea (we certainly don't want to panic!)

vms/evm/database/blockdb/helpers_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	evmdb "github.com/ava-labs/avalanchego/vms/evm/database"
+	heightindexdb "github.com/ava-labs/avalanchego/x/blockdb"
+)
+
+func newDatabasesFromDir(t *testing.T, dataDir string) (*Database, ethdb.Database) {
+	t.Helper()
+
+	base, err := leveldb.New(dataDir, nil, logging.NoLog{}, prometheus.NewRegistry())
```
JonathanOppenheimer, line 30: Here you open `LevelDB` instances but never register cleanup: `t.Cleanup`/`Close`, not sure which is preferred, I feel like everyone has different opinions. Either way, this can lead to flakes.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+func (db *Database) Put(key []byte, value []byte) error {
+	if p, ok := db.parseKey(key); ok {
+		return p.writeHashAndData(value)
+	}
+	return db.Database.Put(key, value)
+}
+
+// Has retrieves if a key is present in the key-value data store.
+// Block data is routed to the height-indexed databases.
+func (db *Database) Has(key []byte) (bool, error) {
```
JonathanOppenheimer, line 353: You certainly know _vastly_ more about databases than I do, but for block keys, you call `readBlock` which does the full `Get` and then decode, which seems expensive. Is there a better way to do this? (no idea, maybe not lol)

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+
+// Key prefixes for block data in [ethdb.Database].
+// This is copied from libevm because they are not exported.
+// Since the prefixes should never be changed, we can avoid libevm changes by
+// duplicating them here.
+const (
+	headerPrefix    = 'h'
+	blockBodyPrefix = 'b'
+	receiptsPrefix  = 'r'
+)
```
JonathanOppenheimer, line 232: I was talking to Stephen about this today (unexported stuff from geth). I'm still not much of a fan of doing this, idk. We've had PRs like this: https://github.com/ava-labs/libevm/pull/262

vms/evm/database/blockdb/database.go
JonathanOppenheimer: resolving this

vms/evm/database/blockdb/database.go
JonathanOppenheimer: Resolving this, you used metaDB, which seems reasonable.

vms/evm/database/blockdb/database_test.go
JonathanOppenheimer: I am resolving this.

vms/evm/database/blockdb/database.go
JonathanOppenheimer: 1. Canonical currently says height `H` -> hash `A`.
2. A new candidate hash `B` at height `H` is staged in batch.
3. `batch.Put` writes block data for `B` immediately (overwriting height `H` slot).
4. Batch is discarded/fails, so canonical mapping remains `H`

So now, the DB says canonical hash is `A`, but payload at height `H` is `B`. Then, reads via canonical path can fail (`ErrNotFound`)

vms/evm/database/blockdb/database.go
JonathanOppenheimer: I know this has already been discussed a bunch on this PR.

alarso16: I think grouping this totally makes sense, maybe closer to their first usage? I would personally move the key functions down here.

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+	db.logger.Info(
+		"Initialized height-indexed block databases",
+		zap.Uint64("minHeight", db.minHeight),
+	)
+
+	return nil
+}
+
+// Key prefixes for block data in [ethdb.Database].
+// This is copied from libevm because they are not exported.
```
alarso16, line 225: It's an add-only PR, but I don't think it matters. Couldn't we use whatever prefixes we want, since its a new DB? If we NEED to match the libevm ones (which I believe we do), we should just export them. In `core/rawdb/schema.libevm.go` (new file to avoid merge conflicts):

```go
package rawdb

const (
	HeaderPrefix = headerPrefix
	BlockBodyPrefix = blockBodyPrefix
	ReceiptsPrefix = receiptsPrefix
)
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+	headerPrefix    = 'h'
+	blockBodyPrefix = 'b'
+	receiptsPrefix  = 'r'
+)
+
+var blockPrefixes = []byte{blockBodyPrefix, headerPrefix, receiptsPrefix}
+
+func parseBlockKey(key []byte) (num uint64, hash common.Hash, ok bool) {
+	// Block keys should have 1 byte prefix + blockNumberSize + 32 bytes for the hash
+	if len(key) != 1+blockNumberSize+32 {
```
alarso16, line 238:
```suggestion
	if len(key) != 1+blockNumberSize+common.HashLength {
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+	if !slices.Contains(blockPrefixes, key[0]) {
+		return 0, common.Hash{}, false
+	}
+	num = binary.BigEndian.Uint64(key[1 : 1+blockNumberSize])
+	bytes := key[1+blockNumberSize:]
+	hash = common.BytesToHash(bytes)
+	return num, hash, true
+}
+
+type parsedBlockKey struct {
```
alarso16, line 250: The name `parsedBlockKey` doesn't make clear to me that you need a specific database attached. Could you write a comment describing what this key is used for and what the `db` is?

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+	encoded, err := rlp.EncodeToBytes([][]byte{hash.Bytes(), data})
+	if err != nil {
+		return err
+	}
+	return db.Put(height, encoded)
+}
+
+// parseKey parses a block key into a parsedBlockKey.
+// It returns false if no block databases for the key prefix exist.
+func (db *Database) parseKey(key []byte) (*parsedBlockKey, bool) {
```
alarso16, line 275: I think you should just return whether a nil key if it's not writable
```suggestion
func (db *Database) parseKey(key []byte) *dbKey {
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+// duplicating them here.
+const (
+	headerPrefix    = 'h'
+	blockBodyPrefix = 'b'
+	receiptsPrefix  = 'r'
+)
+
+var blockPrefixes = []byte{blockBodyPrefix, headerPrefix, receiptsPrefix}
+
+func parseBlockKey(key []byte) (num uint64, hash common.Hash, ok bool) {
```
alarso16, line 236: There's only one type of key! It also doesn't parse everything (why not?) This is only used once, so some of these checks become kind of redundant (or at least should be after you fix the panic). I think the key logic needs reworked

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+
+	return &parsedBlockKey{
+		db:   hdb,
+		num:  num,
+		hash: hash,
+	}, true
+}
+
+func (*Database) readBlock(p *parsedBlockKey) ([]byte, error) {
+	data, err := p.db.Get(p.num)
```
alarso16, line 309: This should probably just be a method on `p`: `p.Data()`. If this is the only time this is used (or you strongly disagree), we can leave it

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+
+	var elems [][]byte
+	if err := rlp.DecodeBytes(data, &elems); err != nil {
+		return nil, err
+	}
+	if len(elems) != 2 {
+		err := fmt.Errorf(
+			"invalid hash+data format: expected 2 elements, got %d", len(elems),
+		)
+		return nil, err
```
alarso16, line 322: nit:
```suggestion
		return nil, fmt.Errorf(
			"invalid hash+data format: expected 2 elements, got %d", len(elems),
		)
```

vms/evm/database/blockdb/database.go
```diff
@@ -0,0 +1,436 @@
...
+	return db.Database.Delete(key)
+}
+
+// Close closes the database.
+func (db *Database) Close() error {
+	if !db.heightDBsReady {
+		return db.Database.Close()
+	}
+
+	// Don't close metaDB since the caller should be managing it.
```
alarso16, line 388: This deserve a godoc comment

vms/evm/database/blockdb/helpers_test.go
alarso16: I don't think you actually need `leveldb`. You can probably use `memdb`, and then you don't have to worry about it. Also there are many tests where you don't use this, maybe you can abstract it a little better?

vms/evm/database/blockdb/helpers_test.go
```diff
@@ -0,0 +1,111 @@
...
+	for i, block := range blocks {
+		rawdb.WriteBlock(db, block)
+		if i < len(receipts) {
+			rawdb.WriteReceipts(db, block.Hash(), block.NumberU64(), receipts[i])
+		}
+		rawdb.WriteCanonicalHash(db, block.Hash(), block.NumberU64())
+	}
+}
+
+func requireRLPEqual(t *testing.T, expected, actual interface{}) {
```
alarso16, line 96: Maybe use generics?
```suggestion
func requireRLPEqual[T any](t *testing.T, expected, actual T) {
```

vms/evm/database/blockdb/database_test.go
```diff
@@ -0,0 +1,498 @@
...
+			)
+			require.NoError(t, err)
+			require.Equal(t, tc.wantDBReady, db.heightDBsReady, "database ready mismatch")
+			require.Equal(t, tc.wantMinHeight, db.minHeight, "database min height mismatch")
+		})
+	}
+}
+
+func TestDatabaseGenesisBlockHandling(t *testing.T) {
+	// Verifies that genesis blocks (block 0) only exist in evmDB and not
```
alarso16, line 325: Why is this expected?

alarso16 (commented):
It's somewhat difficult to reason about. At minimum, you should improve the documentation and naming a little (e.g., what is a "min height"? we also at least have a genesis. Do you mean first non-genesis block committed here?). It was also unclear to me until I read through most of this that this wrapped an ethdb.Database with blocksdb, since it's not described anywhere and appears to only contain blockdb logic (and is in blockdb package). Documenting this better would be immensely helpful

vms/evm/database/blockdb/database.go
alarso16, line 236: Is this also copied from libevm?

DracoLi:
closing - I will revisit the integration after SAE

PR #4486, opened 2025-11-10
Update re-execute benchmarks to target latest range

aaronbuchwald:

### DIFF
--- a/input.json
+++ b/input.json
@@ -1,9 +1,0 @@
-{
-    "runner": "avago-runner-i4i-4xlarge-local-ssd",
-    "config": "default",
-    "start-block": "69000001",
-    "end-block": "70000000",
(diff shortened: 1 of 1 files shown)


tests/reexecute/c/vm_reexecute_test.go
```diff
@@ -79,7 +79,9 @@ var (
 	configKey         = "config"
 	defaultConfigKey  = "default"
 	predefinedConfigs = map[string]string{
-		defaultConfigKey: `{}`,
+		defaultConfigKey: `{
+			"allow-missing-tries": true
+		}`,
```
aaronbuchwald, line 84: This explicitly allows archival execution on top of a non-archival database to support creating height sharded archival DBs.

PR #4514, opened 2025-11-14
Adjust free disk space thresholds

yacovm:
## Why this should be merged

The current threshold for free space are very low and as a result the warning may come too late (see https://github.com/ava-labs/avalanchego/issues/4517). 

This commit changes them to something more reasonable. 

## How this works

Changes default thresholds

## How this was tested

CI


StephenButtolph (dismissed):
These are certainly better than the current values.

I do wonder if we should be even more aggressive (or make them a percentage of the volume size rather than hard numbers).

But I'd be happy to merge this at the least

joshua-kim (dismissed):

tests/fixture/tmpnet/process_runtime.go
```diff
@@ -162,6 +162,8 @@ func (p *ProcessRuntime) writeFlags() error {
 	}
 
 	flags.SetDefaults(FlagsMap{
+		config.SystemTrackerRequiredAvailableDiskSpaceKey:         "1GB",
```
yacovm: Looks like Github's runners don't have much disk space in them...

yacovm:
@maru-ava do you have any idea why the bootstrap monitor task fails in CI?

maru-ava:
> @maru-ava do you have any idea why the bootstrap monitor task fails in CI?

I've created a [new PR](https://github.com/ava-labs/avalanchego/pull/4516) from this one that saves the kind logs as an artifact. I'm thinking those logs might be required to identify the problem.

tests/fixture/tmpnet/defaults.go
```diff
@@ -38,11 +38,13 @@ const (
 // Flags suggested for temporary networks. Applied by default.
 func DefaultTmpnetFlags() FlagsMap {
 	return FlagsMap{
-		config.NetworkPeerListPullGossipFreqKey: "250ms",
-		config.NetworkMaxReconnectDelayKey:      "1s",
-		config.HealthCheckFreqKey:               "2s",
-		config.AdminAPIEnabledKey:               "true",
-		config.IndexEnabledKey:                  "true",
+		config.SystemTrackerRequiredAvailableDiskSpaceKey:         "1GB",
```
maru-ava, line 41: This configures nodes deployed by tmpnet, but not nodes deployed by the bootstrap monitor.

tests/fixture/bootstrapmonitor/e2e/e2e_test.go
```diff
@@ -314,6 +314,7 @@ func createBootstrapTester(tc tests.TestContext, clientset *kubernetes.Clientset
 	flags := defaultPodFlags()
 	flags[config.BootstrapIPsKey] = fmt.Sprintf("%s:%d", bootstrapIP, config.DefaultStakingPort)
 	flags[config.BootstrapIDsKey] = bootstrapNodeID.String()
+	flags[config.SystemTrackerWarningThresholdAvailableDiskSpaceKey] = "2GB"
```
maru-ava: This deploys the bootstrap tester, but as per the e2e failure on `STEP: Creating a node to bootstrap from`, the node that the tester will bootstrap _from_ is not being configured with a lower minimum disk space requirement.

tests/fixture/tmpnet/defaults.go
maru-ava: But my commits from #4516 ensure these defaults also apply to nodes deployed by the bootstrap monitor.

tests/fixture/bootstrapmonitor/e2e/e2e_test.go
maru-ava: After picking my commits from #4516, this change can be reverted.

joshua-kim (approved):

PR #4515, opened 2025-11-14
feat: randomize compactions

alarso16:
## Why this should be merged

To unsync compactions across nodes, we ensure that the regular compation interval is maintained, but the first compaction happens at some random point in time.

## How this works

Picks a random time to start compacting, and then compacts regularly at the interval provided in the config. This is hidden behind a config flag

## How this was tested

It hasn't been!

No


samliok (approved):
down to give this a try, do we need to update the realeases.md file since we are adding a config flag?

StephenButtolph:
Decided that this is unlikely to work

PR #4518, opened 2025-11-15
Monitor free disk percentage, not just absolute space

yacovm:
## Why this should be merged

We currently have metrics for available disk space and we fail the health check if we have less than a predefined amount of space left. 

This commit adds metrics for available disk percentage, and fails the health check if we have less than a predefined percentage of space left. 

The motivation is that if one wants to re-use the configuration of the validator to several nodes with different volume sizes, using absolute space is less fitting than using percentages. 

For example, a volume that started with all of its 2TB free and now has 200GB space left is not the same as than a volume that started with 400GB and has half of its space left. 

## How this works

Adds a flag `--system-tracker-disk-warning-threshold-available-space-percentage` that defaults to 3% space left.

Changes `AvailableBytes` on `storage_unix.go`, `storage_openbsd.go`, to return also percentage, not only absolute. 

## How this was tested

Only tested on Linux, however I looked at the `Statfs_t` struct of [OpenBSD](https://github.com/golang/sys/blob/master/unix/ztypes_openbsd_amd64.go#L78-L101) to make sure I picked the right fields.

I took a Fuji node I had and compiled the avalanchego binary from this branch, and re-launched the node and then I created a big file using `fallocate -l 400G bigfile` and observed the metrics: 

image

Soon enough, the node's health check began failing: 

```
curl 'http://localhost:9650/ext/health'  | jq
```

image

I then deleted the file and observed the metrics:

image

And checked that the node returned to be healthy once more:

image

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1143,8 +1143,8 @@ func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warn
 	warningThresholdAvailableDiskSpace = v.GetUint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey)
 	warningThresholdAvailableDiskSpacePercentage = v.GetUint64(SystemTrackerWarnThreshAvailDiskSpacePercentageKey)
 	switch {
-	case warningThresholdAvailableDiskSpacePercentage > 50 || warningThresholdAvailableDiskSpacePercentage == 0:
-		return 0, 0, 0, fmt.Errorf("%q (%d) must be in [1, 50]", SystemTrackerWarnThreshAvailDiskSpacePercentageKey, warningThresholdAvailableDiskSpacePercentage)
+	case warningThresholdAvailableDiskSpacePercentage > 50:
+		return 0, 0, 0, fmt.Errorf("%q (%d) must be in [0, 50]", SystemTrackerWarnThreshAvailDiskSpacePercentageKey, warningThresholdAvailableDiskSpacePercentage)
 	case warningThresholdAvailableDiskSpace < requiredAvailableDiskSpace:
 		return 0, 0, 0, fmt.Errorf("%q (%d) < %q (%d)", SystemTrackerWarningThresholdAvailableDiskSpaceKey, warningThresholdAvailableDiskSpace, SystemTrackerRequiredAvailableDiskSpaceKey, requiredAvailableDiskSpace)
 	default:
--- a/node/node.go
+++ b/node/node.go
@@ -1459,20 +1459,25 @@ func (n *Node) initHealthAPI() error {
 		availableDiskBytes := n.resourceTracker.DiskTracker().AvailableDiskBytes()
 		availableDiskPercentage := n.resourceTracker.DiskTracker().AvailableDiskPercentage()
 
-		var err error
+		var diskSpaceErrors []error
 		if availableDiskBytes < n.Config.RequiredAvailableDiskSpace {
 			n.Log.Fatal("low on disk space. Shutting down...",
 				zap.Uint64("remainingDiskBytes", availableDiskBytes),
 			)
 			go n.Shutdown(1)
-			err = fmt.Errorf("remaining available disk space (%d) is below minimum required available space (%d)", availableDiskBytes, n.Config.RequiredAvailableDiskSpace)
+			err := fmt.Errorf("remaining available disk space (%d) is below minimum required available space (%d)", availableDiskBytes, n.Config.RequiredAvailableDiskSpace)
+			diskSpaceErrors = append(diskSpaceErrors, err)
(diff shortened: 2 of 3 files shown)


utils/resource/no_usage.go
```diff
@@ -10,6 +10,10 @@ var NoUsage User = noUsage{}
 
 type noUsage struct{}
 
+func (noUsage) AvailableDiskPercentage() uint64 {
+	return math.MaxUint64
```
DracoLi: does this need to be swapped?  `AvailableDiskBytes` = `math.MaxUint64` and `AvailableDiskPercentage` = 100?

yacovm: yes, oops!

yacovm: Fixed, thanks

node/node.go
```diff
@@ -1469,8 +1470,13 @@ func (n *Node) initHealthAPI() error {
 			err = fmt.Errorf("remaining available disk space (%d) is below the warning threshold of disk space (%d)", availableDiskBytes, n.Config.WarningThresholdAvailableDiskSpace)
 		}
 
+		if availableDiskPercentage < n.Config.WarningThresholdAvailableDiskSpacePercentage {
+			err = fmt.Errorf("remaining available disk space percentage (%d%%) is below minimum required available space percentage (%d%%)", availableDiskPercentage, n.Config.WarningThresholdAvailableDiskSpacePercentage)
```
rrazvan1:
```suggestion
			err = fmt.Errorf("remaining available disk space percentage (%d%%) is below the warning threshold of disk space percentage (%d%%)", availableDiskPercentage, n.Config.WarningThresholdAvailableDiskSpacePercentage)
```

yacovm: https://github.com/ava-labs/avalanchego/pull/4518/commits/1634be6977cc9a5ca8487add1eee0e5b9697ca1b

node/node.go
```diff
@@ -1469,8 +1470,13 @@ func (n *Node) initHealthAPI() error {
 			err = fmt.Errorf("remaining available disk space (%d) is below the warning threshold of disk space (%d)", availableDiskBytes, n.Config.WarningThresholdAvailableDiskSpace)
 		}
 
+		if availableDiskPercentage < n.Config.WarningThresholdAvailableDiskSpacePercentage {
+			err = fmt.Errorf("remaining available disk space percentage (%d%%) is below the warning threshold of space percentage (%d%%)", availableDiskPercentage, n.Config.WarningThresholdAvailableDiskSpacePercentage)
+		}
```
StephenButtolph, line 1478: Should we be returning both of these errors (rather than overwriting?).

config/config.go
```diff
@@ -1138,14 +1138,17 @@ func getCPUTargeterConfig(v *viper.Viper) (tracker.TargeterConfig, error) {
 	}
 }
 
-func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warningThresholdAvailableDiskSpace uint64, err error) {
+func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warningThresholdAvailableDiskSpace uint64, warningThresholdAvailableDiskSpacePercentage uint64, err error) {
 	requiredAvailableDiskSpace = v.GetUint64(SystemTrackerRequiredAvailableDiskSpaceKey)
 	warningThresholdAvailableDiskSpace = v.GetUint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey)
+	warningThresholdAvailableDiskSpacePercentage = v.GetUint64(SystemTrackerWarnThreshAvailDiskSpacePercentageKey)
 	switch {
+	case warningThresholdAvailableDiskSpacePercentage > 50 || warningThresholdAvailableDiskSpacePercentage == 0:
```
StephenButtolph: Is there a reason that we don't allow people to disable this alert (by setting it to `0`)? We explicitly disallow it, so I guess it was intentional

StephenButtolph (approved):
2 small comments, diff lgtm

config/config.go
yacovm: Changed to allow 0.

https://github.com/ava-labs/avalanchego/pull/4518/commits/364e97d0cdd7eb9e41546913b26c8a8285ec46b4#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R1147

node/node.go
yacovm: Fixed

https://github.com/ava-labs/avalanchego/pull/4518/commits/364e97d0cdd7eb9e41546913b26c8a8285ec46b4#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R1480

PR #4521, opened 2025-11-17
refactor(blockdb): use atomic.Bool for cacheDB closed state

DracoLi:
## Why this should be merged

This cleans up the `cacheDB` close logic by removing an unnecessary mutex and using an atomic.Bool instead, which makes the code simpler and cheaper to reason about while preserving the same behavior on close.

This PR is a follow up to [this comment](https://github.com/ava-labs/avalanchego/pull/4425#discussion_r2524557642) in https://github.com/ava-labs/avalanchego/pull/4425.

## How this works

`cacheDB` now tracks the closed state via an `atomic.Bool` rather than `closeMu`. We still rely on the underlying DB’s own synchronization to protect any in-flight disk I/O; the cache itself doesn’t need extra protection at close. 
If a `Get` races with `Close`, the cache lookup will fail and the underlying DB will return a “db closed” error as expected.

## How this was tested

Unit tests

No, internal implementation detail only


rrazvan1:
> If a Get races with Close, the cache lookup will fail and the underlying DB will return a “db closed” error as expected.

But if the cache has that value, the value will be returned, even though the db is closed.
Is this an issue?

DracoLi:
> But if the cache has that value, the value will be returned, even though the db is closed.
Is this an issue?

Not an issue here.

PR #4524, opened 2025-11-17
chore: ignore local Claude Code agent file

RodrigoVillar:
## Why this should be merged

For developers that use Claude Code (like myself), having an agent file greatly improves the context that Claude Code has and leads to much better prompt results. The Claude Code best practices [page](https://www.anthropic.com/engineering/claude-code-best-practices) suggests that for agent files not meant to be shared, you should add `CLAUDE.local.md` to your `.gitignore` - doing this allows the following:
- Allows developers to use define their own agent files
- Prevents agent files from having to go to AvalancheGo (and undergo a review process)

## How this works

Adds `CLAUDE.local.md` to `.gitignore`

## How this was tested

N/A

No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,15 +60,15 @@ jobs:
           loki_push_url: ${{ secrets.LOKI_PUSH_URL || '' }}
           loki_username: ${{ secrets.LOKI_USERNAME || '' }}
           loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
-  e2e_post_granite:
+  e2e_post_latest:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Run e2e tests
         uses: ./.github/actions/run-monitored-tmpnet-cmd
         with:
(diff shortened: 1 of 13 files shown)


JonathanOppenheimer (approved):
I've never used claude code! Out of curiosity could you share what kind of stuff you put in your agents file? I gotta give it a try perhaps

maru-ava:
I recommend relying on ~/.gitignore committed to a dotfiles repo to ignore files created [edit: or consumed] by one's choice of development tooling.

RodrigoVillar:
@maru-ava Good point; I went with adding it to the AvalancheGo repository to follow the existing convention:

https://github.com/ava-labs/avalanchego/blob/2a10c52709cf1384b4b1a39955f324c5ff11e0d9/.gitignore#L63-L64

Should we remove `.envrc.local` from the `.gitignore` then?

maru-ava:
> Should we remove .envrc.local from the .gitignore then?

...and `.vscode` and `.idea` and... I would be in favor of slimming .gitignore down for sure, but may make sense to raise this in avalanchego slack to check other peoples' view.

PR #4531, opened 2025-11-20
[migrate-coreth][1] Add tooling enabling repo migration

maru-ava:
## Why this should be merged

Adds README documenting proposed procedure and associated tooling.

## How this was tested

CI

N/A

### DIFF
--- a/graft/scripts/rewrite-imports.sh
+++ b/graft/scripts/rewrite-imports.sh
@@ -50,7 +50,7 @@ echo "staging changes..."
 git add -u '*.go'
 
 echo "committing changes..."
-git commit -m "Rewrite ${ORIGINAL_IMPORT} imports to ${TARGET_IMPORT}
+git commit -S -m "Rewrite ${ORIGINAL_IMPORT} imports to ${TARGET_IMPORT}
 
 Rewrites all Go import statements from external package ${ORIGINAL_IMPORT}
 to internal graft subdirectory ${TARGET_IMPORT}."


graft/scripts/rewrite-imports.sh
alarso16, line 1: These changes don't seem complete. For clarity, right now, it finds every import with `github.com/ava-labs/coreth` and replaces it with `github.com/ava-labs/avalanchego/graft/coreth`, and it does this correctly.

However, nothing compiles after this. It seems we also need:
1. Rename the module in `graft/coreth/go.mod`
2. `go mod edit -replace` in avalanchego (so how does releasing work??)

As an alternative to the above, what could be done is providing an optional subdirectory to replace in. This way, coreth could be moved, tested, and compiled, while avalanchego still functions independently.

graft/README.md
```diff
@@ -0,0 +1,88 @@
...
+step, and the commit of all changes and creation of a new branch from
+the current branch before beginning a subsequent step.
+
+ - [ ] Add tasks for subtree merge and import rewrite to graft/Taskfile.yml (as per the example of existing tasks)
+   - These tasks are intended to simplify the repeated invocation that
+     will be required when a repo is being developed in parallel with
+     migration.
+ - [ ] Execute the subtree merge task (it will commit the result automatically)
+ - [ ] Remove files made redundant by the migration
+   - Prioritizing file removal before modification minimizes the changes requiring review
```
JonathanOppenheimer, line 60: can not emphasize this enough haha

graft/README.md
```diff
@@ -0,0 +1,88 @@
...
+ - [ ] Perform required go module changes
+ - [ ] Migrate CI jobs (unit test, e2e, linting, etc)
+ - [ ] Get CI jobs passing
+
+### Rebasing Inflight PRs
+
+Provided a subtree merge was used to perform the graft, a common merge
+base will exist with which to rebase PRs targeted at the original git
+repo. This allows for repo migration and ongoing development to
+proceed in parallel. Once a graft has been finalized, outstanding PR
```
JonathanOppenheimer, line 71: this is very cool

graft/scripts/rewrite-imports.sh
alarso16: After looking at #4534, my concerns here are mostly resolved around go.mods. I think doing this manually is ok, but I would love to know how versioning will work

JonathanOppenheimer:
I would also add a step that all of this temporary supporting files -- scripts, etc. should be removed in a follow up PR, after the graft is complete (unless we want to keep all of the tools for archival purposes somehwere?) 

this could be a 5th PR

graft/scripts/rewrite-imports.sh
maru-ava: My intention here was to limit the script to rewriting imports without worrying about breakage. I think there is likely to be a period of parallel migration work and ongoing development for a given migration, and the script is intended to be run as many times as necessary to catch all the import changes required on both sides (avalanchego and the migrated repo). go module changes, on the other hand, are static and only need to be done once and can then be saved as a commit that gets rebased as often as needed.

To your point about moving/testing/compiling coreth independent of changes to avalanchego, does that actually make sense given the circular dependence between the two?

Regarding releasing for coreth, the mutual dependency likely means there will no longer need to be a need to tag them independently. It's possible, though, and likely required for subnet-evm. iirc the tagging convention for a non-primary go module would be by prefixing it e.g. `subnet-evm-v1.0.0`.

maru-ava:
> I would also add a step that all of this temporary supporting files -- scripts, etc. should be removed in a follow up PR, after the graft is complete (unless we want to keep all of the tools for archival purposes somehwere?)
> 
> this could be a 5th PR

I think the contents of this PR are likely to be something we want to retain for future migrations (e.g. subnet-evm, interop, etc). That was my intent in breaking them out into repo-agnostic form.

JonathanOppenheimer:
> > I would also add a step that all of this temporary supporting files -- scripts, etc. should be removed in a follow up PR, after the graft is complete (unless we want to keep all of the tools for archival purposes somehwere?)
> > this could be a 5th PR
> 
> I think the contents of this PR are likely to be something we want to retain for future migrations (e.g. subnet-evm, interop, etc). That was my intent in breaking them out into repo-agnostic form.

gotcha that makes sense!

graft/README.md
maru-ava: Thanks to @alarso16  for documenting the process! Hence his co-authorship of the PR.

graft/scripts/rewrite-imports.sh
```diff
@@ -0,0 +1,49 @@
...
+  exit 1
+fi
+
+ORIGINAL_IMPORT="$1"
+
+# Extract the last component of the import path
+# e.g., github.com/ava-labs/coreth -> coreth
+PACKAGE_NAME="${ORIGINAL_IMPORT##*/}"
+
+TARGET_IMPORT="github.com/ava-labs/avalanchego/graft/${PACKAGE_NAME}"
```
joshua-kim: Should we support allowing people to graft into an arbitrary destination directory? I don't know if we want `coreth` to live under `avalanchego/graft/coreth`, so it might be nice to support us putting it somewhere like under `vms/evm/coreth`.

alarso16: That seems reasonable. I updated the script to allow this. Is the usage clear?

graft/scripts/subtree-merge.sh
```diff
@@ -0,0 +1,89 @@
...
+  exit 1
+fi
+
+MODULE_PATH="$1"
+
+REPO_ROOT=$( cd "$( dirname "${BASH_SOURCE[0]}" )"; cd ../.. && pwd )
+cd "${REPO_ROOT}"
+
+# Discover version from go.mod if not provided
+if [ $# -eq 2 ]; then
```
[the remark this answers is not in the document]
alarso16: Whoa good catch!

graft/scripts/rewrite-imports.sh
```diff
@@ -0,0 +1,58 @@
...
+rg "${ESCAPED_ORIGINAL}" -t go --files-with-matches --null | \
+  xargs -0 perl -i -pe "s|\\Q${ORIGINAL_IMPORT}\\E|${TARGET_IMPORT}|g"
+
+echo "import rewrite completed successfully"
+
+echo "staging changes..."
+git add -u '*.go'
+
+echo "committing changes..."
+git commit -m "Rewrite ${ORIGINAL_IMPORT} imports to ${TARGET_IMPORT}
```
StephenButtolph: Should this use `-S` to sign this commit? Some people have it setup to auto sign, but it seems like signing here may be a good idea since we expect signed commits.

PR #4535, opened 2025-11-20
[migrate-coreth][5] Migrate CI

maru-ava:
## Why this should be merged

Migrates coreth CI to be able to run in this repo.

I highly recommend reviewing this commit by commit, since there's a lot of unrelated lint fixes in this PR. This helps you see the necessary scope

## How this was tested

CI

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,3 +243,8 @@ To officially support a new platform, one must satisfy the following requirement
 **We and our community welcome responsible disclosures.**
 
 Please refer to our [Security Policy](SECURITY.md) and [Security Advisories](https://github.com/ava-labs/avalanchego/security/advisories).
+
+## Licenses
+
+Unless otherwise stated, all code in this repository is licensed under BSD-3. See our [licensing](LICENSE) for more details. Specifically, [grafted repositories](graft/) may use
+a different license, and one must refer to that project's README and LICENSE for more details
--- a/graft/coreth/README.md
+++ b/graft/coreth/README.md
@@ -92,6 +92,6 @@ To support these changes, there have been a number of changes to the C-Chain blo
 - `ExtDataGasUsed`: amount of gas consumed by the atomic transactions in the block
 - `BlockGasCost`: surcharge for producing a block faster than the target rate
 
-## Releasing
+## Licensing
 
-See [docs/releasing/README.md](docs/releasing/README.md) for the release process.
+As opposed to `avalanchego` (see the project's [README](../../README.md)), this module is licensed under LGPL, unless otherwise stated in each file header.


graft/coreth/scripts/build_test.sh
```diff
@@ -2,14 +2,9 @@
...
-# Avalanche root directory
-CORETH_PATH=$(
-  cd "$(dirname "${BASH_SOURCE[0]}")"
-  cd .. && pwd
-)
 
-# Load the constants
-source "$CORETH_PATH"/scripts/constants.sh
+REPO_ROOT=$( cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )
+source "$REPO_ROOT"/scripts/constants.sh
```
maru-ava, line 7: I didn't see any relevant configuration in the coreth version of this file that wasn't available in the avalanchego version, and figured it was better to use the same file.

.github/workflows/coreth-ci.yml
```diff
@@ -0,0 +1,76 @@
...
+      run:
+        working-directory: ./graft/coreth
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, ubuntu-22.04, ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-go-for-project
+      - name: Check generated codec files are up to date
```
maru-ava: Not clear why these steps are necessary given the lint job - will remove.

.github/workflows/coreth-ci.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./graft/coreth
+      - uses: ./.github/actions/setup-go-for-project
+      - name: Run all lint checks
+        run: ./scripts/run_task.sh lint-all-ci
+      - name: Check go.mod and go.sum are up-to-date
+        run: ./scripts/run_task.sh check-go-mod-tidy
+  links-lint:
```
maru-ava: Not clear this job is necessary, will check if the one in the avalanchego workflow is sufficient.

maru-ava: One job is enough, will remove.

scripts/build_test.sh
```diff
@@ -19,13 +19,15 @@ if [[ -n "${NO_RACE:-}" ]]; then
     race=""
 fi
 
-EXCLUDED_TARGETS="| grep -v /mocks | grep -v proto | grep -v tests/e2e | grep -v tests/load/c | grep -v tests/upgrade | grep -v tests/fixture/bootstrapmonitor/e2e | grep -v tests/reexecute"
-
-if [[ "$(go env GOOS)" == "windows" ]]; then
```
maru-ava, line 24: Discovered while changing this file - we don't test on windows anymore

graft/coreth/.envrc
```diff
@@ -1,10 +0,0 @@
-# - Will execute on entry to the containing directory only if authorized with `direnv allow`
```
maru-ava, line 1: No real reason to include this here but since I do use direnv I was getting annoyed at constantly being asked to allow it to load.

.github/actions/run-monitored-tmpnet-cmd/action.yml
```diff
@@ -5,6 +5,9 @@ inputs:
   run:
     description: "the bash script to run e.g. ./scripts/my-script.sh"
     required: true
+  working_directory:
```
maru-ava, line 8: This was required to allow usage of the action from the graft/coreth path.

scripts/tests.e2e.sh
```diff
@@ -6,6 +6,7 @@ set -euo pipefail
 # ./scripts/tests.e2e.sh
 # ./scripts/tests.e2e.sh --ginkgo.label-filter=x                                       # All arguments are supplied to ginkgo
 # E2E_SERIAL=1 ./scripts/tests.e2e.sh                                                  # Run tests serially
+# E2E_TARGET=./graft/coreth/tests/warp ./scripts/tests.e2e.sh                          # Run the warp e2e tests
```
maru-ava, line 9: I updated this file to support reuse by the warp suite for consistency of execution across e2e suites

vms/evm/emulate/emulate.go
```diff
@@ -9,7 +9,8 @@ package emulate
 
 import (
 	cchain "github.com/ava-labs/avalanchego/graft/coreth/plugin/evm"
-	subnet "github.com/ava-labs/subnet-evm/plugin/evm"
+	// TODO(marun) Commented out until subnet-evm is updated to use github.com/avalanchego/graft/coreth
```
maru-ava: @ARR4N This was a nasty surprise! Thankfully we'll only have this cross-repo circular dependency until subnet-evm gets migrated.

Dockerfile
```diff
@@ -11,6 +11,7 @@ WORKDIR /build
 # Copy and download avalanche dependencies using go mod
 COPY go.mod .
 COPY go.sum .
+COPY graft/coreth ./graft/coreth
```
maru-ava, line 14: Without this the subsequent line fails due to the replace directive pointing to a local path

graft/coreth/core/blockchain_test.go
```diff
@@ -362,6 +362,7 @@ func TestCorruptSnapshots(t *testing.T) {
 }
 
 func TestBlockChainOfflinePruningUngracefulShutdown(t *testing.T) {
+	t.Skip("TODO(marun) needs to be updated for compatibility with migration")
```
alarso16: This shouldn't be flaky. Why does this fail?

maru-ava: Likely an artifact of my not using a long enough timeout early on. Dropped.

vms/evm/emulate/emulate.go
alarso16: All this requires is an updated version of subnet-evm

alarso16: This is fixed

alarso16:
The coreth unit test failures will be fixed with ava-labs/coreth#1421, but requires a tricky rebase on PR 2 after it's merged. So we can worry about that later

.golangci.yml
```diff
@@ -29,6 +29,24 @@ issues:
   # Maximum count of issues with the same text.
   max-same-issues: 0
 
+formatters:
```
alarso16, line 32: This is necessary to allow dynamically appending excluded files at the bottom of the lint file for coreth linting

scripts/build_test.sh
```diff
@@ -19,13 +19,15 @@ if [[ -n "${NO_RACE:-}" ]]; then
     race=""
 fi
 
-EXCLUDED_TARGETS="| grep -v /mocks | grep -v proto | grep -v tests/e2e | grep -v tests/load/c | grep -v tests/upgrade | grep -v tests/fixture/bootstrapmonitor/e2e | grep -v tests/reexecute"
```
maru-ava, line 22: Suggest dropping changes to this file. They were were only necessary when graft/coreth wasn't going to be a go module, but are no longer needed because `go test` doesn't operate across module boundaries. I can handle the removal of windows special-casing separately.

alarso16: Makes sense, done

graft/coreth/accounts/abi/bind/bind_test.go
```diff
@@ -2102,10 +2102,12 @@ var bindTests = []struct {
 // The binding tests have been modified to run in two separate test
 // functions to allow these tests to pass on GitHub Actions.
 func TestGolangBindingsOverload(t *testing.T) {
+	t.Skip("TODO(marun) needs to be updated for compatibility with migration to avalanchego")
```
alarso16, line 2105: These are skipped as a known flaky test. However, they do pass normally

.github/workflows/coreth-ci.yml
```diff
@@ -0,0 +1,62 @@
+name: Coreth
```
maru-ava, line 1: A previous iteration of this work integrated coreth jobs into the main workflow, but now that it is its own go module again I figured a separate workflow was appropriate. We can always revisit in future.

graft/coreth/accounts/abi/bind/bind_test.go
powerslider: so we actually want to merge them skipped and fix the flakiness later?

alarso16: 🤷‍♂️ it's something pretty deep. I spent a decent amount of time trying to find the source, and @ceyonur thinks it's a non-critical flake in the mempool. I'm not sure what the best course of action is

maru-ava: If you want these jobs to be made required post-merge, they'll need to be largely flake-free.

PR #4537, opened 2025-11-21
test: add import compliance tests

### LINKED ISSUE #4541: Remove libevm import restriction test from graft/ post merge.
The test included added in https://github.com/ava-labs/coreth/pull/1348, is a historical artifact, and it, and the script that was a precursor to it should be removed entirely, once its added to AvalancheGo 

To add some historical context on why this test originally existed, here is some historical context from @ARR4N:

> Originally it was [read: this test existed] because we were doing partial migrations of packages by using type aliases to the libevm equivalents.
```
type Foo pkg.Foo
// is different to
type Foo = pkg.Foo
```
> The latter is an alias that is equivalent to dropping in `pkg.Foo` wherever `Foo` is used, whereas the former creates a new type that inherits (lay term; not OOP) the structure, but not the methods, of `pkg.Foo`.
> As an example, the first things we ever migrated were `params.ChainConfig` and `params.Rules` but we still had to use all of the other coreth/params identifiers, so we blocked accidental import of libevm/params.

As the work to do this is largely complete, these are a historical artifact and should be removed.

There is one important exception which is the `libevm/pseudo` package. This should not be imported anywhere outside of `libevm,` without the express approval of @ARR4N, @StephenButtolph or @joshua-kim (as dictated by Arran himself). 

> It requires so much care to avoid catastrophic misuse and there are so few reasons to ever use it that I think it needs someone with a lot of Go experience to weigh in on approving it.

JonathanOppenheimer:
## Why this should be merged
This PR prevents files inside of the `vms/evm` directory from importing packages within the `graft/` directory.
   - This is important because with the removal of coreth, there is not longer a natural boundary between the uplifted and non-uplifted code. We need to enforce such a boundary. 
   - This boundary is important, especially as other people individuals start work on properly uplfiting code post monorepo. Reviewers and developers may not be familiar with why we want this distinction to exist and programmatically enforcing this boundary will prevent mistakes and leakage of legacy EVM code into AvalancheGo.  
  - See this issue for additional context https://github.com/ava-labs/avalanchego/issues/4541
  
Closes #4541, #4643 

## How this works

The boundaries are as follows:

- `graft/coreth` can be imported everywhere in AvalancheGo except `vms/evm`.
  - However, `vms/evm/emulate` is an exception to this rule
- `graft/subnet-evm` can _not_ be imported anywhere in AvalancheGo
  - However, `vms/evm/emulate` is an exception to this rule

## How this was tested
Existing CI.

No


tests/imports_test.go
```diff
@@ -0,0 +1,196 @@
...
+			errorMsg.WriteString(fmt.Sprintf("   • %s\n", file))
+		}
+		errorMsg.WriteString("\n")
+	}
+	require.Fail(t, errorMsg.String())
+}
+
+// TestLibevmImportsAreAllowed ensures that all libevm imports in the graft directory
+// are explicitly allowed via the libevm-allowed-packages.txt file.
+func TestLibevmImportsAreAllowed(t *testing.T) {
```
JonathanOppenheimer: See [here](https://github.com/ava-labs/coreth/blob/master/plugin/evm/imports_test.go) for the source of the code below.

graft/scripts/forbidden-geth-files.txt
```diff
@@ -0,0 +1,57 @@
+# Forbidden geth upstream files that cannot be imported outside the graft directory.
```
JonathanOppenheimer: I considered translating both of these files to `.json` to make the loading and parsing simpler in `loadPatternFile()`.  

However, the text files are very human readable and easy to maintain. Would like any opinions about the tradeoffs here!

graft/scripts/libevm-allowed-packages.txt
```diff
@@ -0,0 +1,49 @@
+# Allowed libevm packages that can be imported in the graft directory.
```
maru-ava: grafts/scripts should only be used for graft-related functionality and is likely to go away at some point once we've migrated everything we care about.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,232 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
maru-ava, line 1: Please find somewhere else for this too, the root of tests/ is for utility functions that get imported, not unit tests like this one.

graft/scripts/libevm-allowed-packages.txt
JonathanOppenheimer: Yeah, this should also be `vms/evm` - I'll generalize to EVM code?

maru-ava: Yeah, vms/evm is a likely candidate. A key consideration of where to locate code is who will own it (i.e. be capable of approving changes). No point in requiring non-evm maintainers to approve evm code.

vms/evm/scripts/libevm-allowed-packages.txt
```diff
@@ -39,6 +40,7 @@
 "github.com/ava-labs/libevm/libevm/stateconf"
 "github.com/ava-labs/libevm/log"
 "github.com/ava-labs/libevm/metrics"
+"github.com/ava-labs/libevm/params"
```
JonathanOppenheimer: This test being broken in the EVM repositories already presenting issues. 

@ARR4N is this okay to import into EVM code? iirc there was some issues with the params package.

graft/scripts/libevm-allowed-packages.txt
JonathanOppenheimer: agree!

vms/evm/scripts/libevm-allowed-packages.txt
JonathanOppenheimer: (I added both existing violation to the allowed package list)

vms/evm/imports_test.go
JonathanOppenheimer: yup, `vm/evm` also seems appropriate

vms/evm/scripts/libevm-allowed-packages.txt
JonathanOppenheimer: No longer relevant.

tests/imports_test.go
JonathanOppenheimer: No longer relevant.

graft/scripts/forbidden-geth-files.txt
JonathanOppenheimer: No longer relevant.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+// See the file LICENSE for licensing terms.
+
+package tests
```
ARR4N:
```suggestion
package evm
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+// TestDoNotImportFromGraft ensures proper import rules for graft packages:
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
+// - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate
```
ARR4N: Why are these the rules? I understand `emulate` being allowed to import, but not the other limitations.

Rule of thumb for all code, not just this instance: comments should explain "why", not "what" nor "how".

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestEnforceGraftImportBoundaries(t *testing.T) {
+	graftRegex := regexp.MustCompile(`^github\.com/ava-labs/avalanchego/graft(/|$)`)
```
ARR4N: (subjective, optional) I've seen enough bugs arising from regular expressions that could have been string comparisons that I tend to avoid them unless absolutely necessary. The functionality in this file could just be:

```go
func isPackageIn(importPath, root string) bool {
	if importPath == root {
		return true
	}
	if len(importPath) < len(root)+1 {
		return false
	}
	return strings.HasPrefix(importPath, root) && importPath[len(root)] == '/'
}
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		errorMsg.WriteString("\n")
+	}
+
+	return errorMsg.String()
+}
+
+// findImportsMatchingPattern is a generalized function that finds all imports
+// matching a given regex pattern in the specified directory.
+// The filterFunc can be used to skip certain files or imports (return true to skip).
```
ARR4N: > return true to skip

This is the opposite to what I'd expect when passing in a filter function:

```go
func myFilter(file string) bool {
   return true // intuition says this is now included
}
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+// TestDoNotImportFromGraft ensures proper import rules for graft packages:
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
```
maru-ava, line 24: Myabe provide more detail as to why  these restrictions are necessary?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+			strings.Contains(path, "graft/*/tempextrastest/") {
+			return true
+		}
+
+		// Skip files in the graft directory itself - they can import from graft
+		if strings.Contains(path, "/graft/") {
+			return true
+		}
+
+		isInEmulate := strings.Contains(path, "/vms/evm/emulate/")
```
maru-ava: (No action required) Maybe localize the following 3 definitions to the conditionals that use them?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	rootDir string,
+	importRegex *regexp.Regexp,
+	filterFunc func(filePath string, importPath string, importSpec *ast.ImportSpec) bool,
+) (map[string]set.Set[string], error) {
+	imports := make(map[string]set.Set[string])
+
+	err := filepath.Walk(rootDir, func(path string, _ os.FileInfo, err error) error {
+		if err != nil || !strings.HasSuffix(path, ".go") {
+			return err
+		}
```
ARR4N:
```suggestion
		if err != nil {
			return err
		}
		if !strings.HasSuffix(path, ".go") {
			return nil
		}
```

Although functionally equivalent, this is much clearer and doesn't force the reader to reason about how the returned `err` will be `nil` in the non-Go instance.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+		// Skip files in the graft directory itself - they can import from graft
+		if strings.Contains(path, "/graft/") {
+			return true
+		}
+
+		isInEmulate := strings.Contains(path, "/vms/evm/emulate/")
+		isInVmsEvm := strings.Contains(path, "/vms/evm/")
+		isCoreth := strings.Contains(importPath, "/graft/coreth")
+		isSubnetEVM := strings.Contains(importPath, "/graft/subnet-evm")
```
maru-ava: (No action required) Not a biggie, but would prefer to see subnet-evm specific stuff as part of the migration not in advance.

But also, why is testing for these independent grafts conflated? Maybe implement 2 tests that use the same helpers, one for each graft location?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	filterFunc func(filePath string, importPath string, importSpec *ast.ImportSpec) bool,
+) (map[string]set.Set[string], error) {
+	imports := make(map[string]set.Set[string])
+
+	err := filepath.Walk(rootDir, func(path string, _ os.FileInfo, err error) error {
+		if err != nil || !strings.HasSuffix(path, ".go") {
+			return err
+		}
+
+		node, err := parser.ParseFile(token.NewFileSet(), path, nil, parser.ParseComments)
```
ARR4N:
```suggestion
		node, err := parser.ParseFile(token.NewFileSet(), path, nil, parser.ImportsOnly)
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to parse %s: %w", path, err)
```
ARR4N:
```suggestion
			return fmt.Errorf("failed to parse %s: %v", path, err)
```

There is no need for the consumer to unwrap the error so don't use `%w`. It should be a considered decision, not the default.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+			if imp.Path == nil {
+				continue
+			}
+
+			importPath := strings.Trim(imp.Path.Value, `"`)
+			if !importRegex.MatchString(importPath) {
+				continue
+			}
+
+			if filterFunc != nil && filterFunc(path, importPath, imp) {
```
ARR4N:
```suggestion
			if filterFunc != nil && !filterFunc(path, importPath, imp) {
```
See comment on this function for rationale.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+// The filterFunc can be used to skip certain files or imports (return true to skip).
+// Returns a map of import paths to the set of files that contain them.
+func findImportsMatchingPattern(
+	rootDir string,
+	importRegex *regexp.Regexp,
+	filterFunc func(filePath string, importPath string, importSpec *ast.ImportSpec) bool,
+) (map[string]set.Set[string], error) {
+	imports := make(map[string]set.Set[string])
+
+	err := filepath.Walk(rootDir, func(path string, _ os.FileInfo, err error) error {
```
ARR4N:
```suggestion
	err := filepath.Walk(rootDir, func(filePath string, _ os.FileInfo, err error) error {
```

The length of variable names should be inversely proportional to the distance across which they're used. This is especially so when there are potentially ambiguous names (like `importPath` vs `filePath`).

When the scope of usage is very tight then even single-character variable names are appropriate as long as their declaration makes their meaning obvious. Longer names in such circumstances add noise (because of length) without any extra signal (because of the declaration caveat).

When the scope is large then you don't want readers to be forced to remember what a variable is, so names become longer to be more descriptive. In this case, by the time I saw `path` at the bottom I had absolutely no idea what it was.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	header := "Graft import rules violated!\n" +
+		"Rules:\n" +
+		"  - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)\n" +
+		"  - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate\n\n"
+	require.Fail(t, formatImportViolations(foundImports, header))
+}
+
+// TestDoNotImportLibevmPseudo ensures that no code in the repository imports
```
maru-ava: (No action required) If it's libevm-internal only, why isn't the package called `internal` so that the restriction is enforced automatically by golang?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+			if filterFunc != nil && filterFunc(path, importPath, imp) {
+				continue
+			}
+
+			if _, exists := imports[importPath]; !exists {
+				imports[importPath] = set.Set[string]{}
+			}
+			fileSet := imports[importPath]
+			fileSet.Add(path)
+			imports[importPath] = fileSet
```
ARR4N:
```suggestion
			imports[importPath].Add(filePath)
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+// TestDoNotImportFromGraft ensures proper import rules for graft packages:
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
+// - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate
+func TestEnforceGraftImportBoundaries(t *testing.T) {
+	graftRegex := regexp.MustCompile(`^github\.com/ava-labs/avalanchego/graft(/|$)`)
+
+	// Find all graft imports in the entire repository
+	foundImports, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
+		// Skip generated files and test-specific files
```
ARR4N:
```suggestion
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		filename := filepath.Base(path)
+		if strings.HasPrefix(filename, "gen_") ||
+			strings.Contains(path, "graft/*/core/main_test.go") ||
+			strings.Contains(path, "graft/*/tempextrastest/") {
+			return true
```
ARR4N:
```suggestion
			return false
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		// Skip files in the graft directory itself - they can import from graft
+		if strings.Contains(path, "/graft/") {
+			return true
```
ARR4N:
```suggestion
			return false
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	foundImports, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
+		// Skip generated files and test-specific files
+		filename := filepath.Base(path)
+		if strings.HasPrefix(filename, "gen_") ||
+			strings.Contains(path, "graft/*/core/main_test.go") ||
+			strings.Contains(path, "graft/*/tempextrastest/") {
+			return true
+		}
+
+		// Skip files in the graft directory itself - they can import from graft
```
ARR4N:
```suggestion
		// The graft directory can import itself.
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		if strings.Contains(path, "/graft/") {
+			return true
+		}
+
+		isInEmulate := strings.Contains(path, "/vms/evm/emulate/")
+		isInVmsEvm := strings.Contains(path, "/vms/evm/")
+		isCoreth := strings.Contains(importPath, "/graft/coreth")
+		isSubnetEVM := strings.Contains(importPath, "/graft/subnet-evm")
+
+		// graft/coreth: illegal in vms/evm except vms/evm/emulate
```
ARR4N:
```suggestion
```

Or change to explain why.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		if isCoreth && isInVmsEvm && !isInEmulate {
+			return false
+		}
+
+		// graft/subnet-evm: illegal everywhere except vms/evm/emulate
```
ARR4N:
```suggestion
```

Or change to explain why.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		isInEmulate := strings.Contains(path, "/vms/evm/emulate/")
+		isInVmsEvm := strings.Contains(path, "/vms/evm/")
+		isCoreth := strings.Contains(importPath, "/graft/coreth")
+		isSubnetEVM := strings.Contains(importPath, "/graft/subnet-evm")
+
+		// graft/coreth: illegal in vms/evm except vms/evm/emulate
+		if isCoreth && isInVmsEvm && !isInEmulate {
+			return false
```
ARR4N:
```suggestion
			return true
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		// graft/subnet-evm: illegal everywhere except vms/evm/emulate
+		if isSubnetEVM && !isInEmulate {
+			return false
```
ARR4N:
```suggestion
			return true
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		return true
```
ARR4N:
```suggestion
		return false
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+// TestDoNotImportFromGraft ensures proper import rules for graft packages:
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
+// - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate
+func TestEnforceGraftImportBoundaries(t *testing.T) {
+	graftRegex := regexp.MustCompile(`^github\.com/ava-labs/avalanchego/graft(/|$)`)
+
+	// Find all graft imports in the entire repository
```
ARR4N:
```suggestion
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	foundImports, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
```
ARR4N:
```suggestion
	violations, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+		if isSubnetEVM && !isInEmulate {
+			return false
+		}
+
+		return true
+	})
+	require.NoError(t, err, "Failed to find graft imports")
+
+	if len(foundImports) == 0 {
+		return // no violations found
```
ARR4N:
```suggestion
		return
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+	header := "Graft import rules violated!\n" +
+		"Rules:\n" +
+		"  - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)\n" +
+		"  - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate\n\n"
+	require.Fail(t, formatImportViolations(foundImports, header))
+}
+
+// TestDoNotImportLibevmPseudo ensures that no code in the repository imports
+// the libevm/pseudo package, which is for internal libevm use only.
+func TestDoNotImportLibevmPseudo(t *testing.T) {
```
ARR4N:
```suggestion
func TestDoNotImportLibEVMPseudo(t *testing.T) {
```

Although the project name is stylised `libevm`, initialisms in Go identifiers should have all initials with the same capitalisation.

ARR4N: Why does this warrant a separate test?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// findImportsMatchingPattern is a generalized function that finds all imports
+// matching a given regex pattern in the specified directory.
+// The filterFunc can be used to skip certain files or imports (return true to skip).
+// Returns a map of import paths to the set of files that contain them.
+func findImportsMatchingPattern(
+	rootDir string,
+	importRegex *regexp.Regexp,
+	filterFunc func(filePath string, importPath string, importSpec *ast.ImportSpec) bool,
```
ARR4N:
```suggestion
	filterFunc func(filePath string, importPath string) bool,
```

It's never used.

vms/evm/imports_test.go
ARR4N, line 1: I've added comments based on the current macro-pattern, but that being said I think it's been overcomplicated by an unnecessary abstraction to support two tests. There's also no need to collect and then format the output when a `t.Error()` will do.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestEnforceGraftImportBoundaries(t *testing.T) {
+	graftRegex := regexp.MustCompile(`^github\.com/ava-labs/avalanchego/graft(/|$)`)
+
+	// Find all graft imports in the entire repository
+	foundImports, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
+		// Skip generated files and test-specific files
+		filename := filepath.Base(path)
+		if strings.HasPrefix(filename, "gen_") ||
+			strings.Contains(path, "graft/*/core/main_test.go") ||
+			strings.Contains(path, "graft/*/tempextrastest/") {
```
ARR4N: `strings.Contains()` doesn't support wild-card matching, but this is captured by the check below anyway.

ARR4N (changes requested):

vms/evm/imports_test.go
```diff
@@ -0,0 +1,168 @@
...
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
+// - graft/subnet-evm cannot be imported anywhere EXCEPT vms/evm/emulate
+func TestEnforceGraftImportBoundaries(t *testing.T) {
+	graftRegex := regexp.MustCompile(`^github\.com/ava-labs/avalanchego/graft(/|$)`)
+
+	// Find all graft imports in the entire repository
+	foundImports, err := findImportsMatchingPattern("..", graftRegex, func(path string, importPath string, _ *ast.ImportSpec) bool {
+		// Skip generated files and test-specific files
+		filename := filepath.Base(path)
+		if strings.HasPrefix(filename, "gen_") ||
```
ARR4N: https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source:~:text=To%20convey%20to%20humans%20and%20machine%20tools%20that%20code%20is%20generated%2C%20generated%20source%20should%20have%20a%20line%20that%20matches%20the%20following%20regular%20expression%20(in%20Go%20syntax)%3A

That said, why should generated files be allowed to cross the import boundary?

ARR4N:
I think the entirety of this PR can be reduced to something like this:

```go
func TestImportViolations(t *testing.T) {
	const root ".."
	err := filepath.Walk(root, func(file string, info fs.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if strings.ToLower(filepath.Ext(file)) != "go" {
			return nil
		}

		node, err := parser.ParseFile(token.NewFileSet(), file, nil, parser.ImportsOnly)
		if err != nil {
			return fmt.Errorf("parser.ParseFile(..., %q, ...): %v", file, err)
		}

		for _, spec := range node.Imports {
			if spec.Path == nil {
				continue
			}
			imp := strings.Trim(spec.Path.Value, `"`)

			inGraft := strings.Contains(file, "/graft/")
			inEmulate := strings.Contains(file, "/vms/evm/emulate/")
			inVMsEVM := strings.Contains(file, "/vms/evm/")
			importsPseudo := isPackageIn(imp, "github.com/ava-labs/libevm/libevm/pseudo")
			importsCoreth := isPackageIn(imp, "github.com/ava-labs/avalanchego/graft/coreth")
			importsSubnetEVM := isPackageIn(imp, "github.com/ava-labs/avalanchego/graft/subnet-evm")

			// TODO: before merging, this needs a comment explaining why each of them are a violation
			violations := []bool{
				importsPseudo,
				!inGraft && importsCoreth && inVMsEVM && !inEmulate,
				!inGraft && importsSubnetEVM && !inEmulate,
			}
			if slices.Contains(violations, true) {
				t.Errorf("File %q imports %q", file, imp)
			}
		}
		return nil
	})

	require.NoErrorf(t, err, "filepath.Walk(%q)", root)
}

func isPackageIn(importPath, root string) bool {
	if importPath == root {
		return true
	}
	if len(importPath) < len(root)+1 {
		return false
	}
	return strings.HasPrefix(importPath, root) && importPath[len(root)] == '/'
}
```

I'm not particularly happy with the relative path to walk nor the way of checking if something is in a sub-dir (they can be fixed together), but otherwise I think this covers all bases.

vms/evm/imports_test.go
maru-ava: avalanchego enforces %w via golangci-lint.

JonathanOppenheimer:
> I'm not particularly happy with the relative path to walk nor the way of checking if something is in a sub-dir (they can be fixed together), but otherwise I think this covers all bases.

I feel bad that you put all of these great comments together and then just gave me a much simpler test, but your diligence is appreciated. I have replaced the file with your suggested test (tweaked slightly)

vms/evm/imports_test.go
JonathanOppenheimer: They shouldn't -- this was a holdover from a prior test.

vms/evm/imports_test.go
JonathanOppenheimer: ah oops... that's just a rookie programming mistake lol

vms/evm/imports_test.go
JonathanOppenheimer: I agree - I replaced the test with the new test you wrote.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: I added a longer comment in the test - does that satisfy your comment?

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: @ARR4N suggested that all of this be combined into a single import violation test. Would you still like to see the tests separated out? I see the benefit of both ways

- single test: less code, everything in one place
- two tests: separation makes more sense semantically 

I could also do a single test with two subtests?

vms/evm/imports_test.go
JonathanOppenheimer: @ARR4N would this be possible?

vms/evm/imports_test.go
JonathanOppenheimer: I added a longer comment explaining why - is it satisfactory?

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: I guess I think of it it as "yes the function says this file will be filtered" 

and filtered = skip, so the function returns true to skip. 

Outdated regardless.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: Outdated

vms/evm/imports_test.go
JonathanOppenheimer: I'm saving this comment! This should be part of a style guide.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
JonathanOppenheimer: Outdated.

vms/evm/imports_test.go
maru-ava: I think the rational belongs in the header rather than inline, but I'm not going to block on it.

JonathanOppenheimer: I'll move it!

vms/evm/imports_test.go
```diff
@@ -0,0 +1,98 @@
...
+		node, err := parser.ParseFile(token.NewFileSet(), file, nil, parser.ImportsOnly)
+		if err != nil {
+			return fmt.Errorf("parser.ParseFile(..., %q, ...): %w", file, err)
+		}
+
+		for _, spec := range node.Imports {
+			if spec.Path == nil {
+				continue
+			}
+			imp := strings.Trim(spec.Path.Value, `"`)
```
maru-ava: (No action required) Maybe use a more descriptive name (e.g. `importPath`) for clarity?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			inGraft := strings.Contains(file, "/graft")
+			inEmulate := strings.Contains(file, "/vms/evm/emulate")
+			inVMsEVM := strings.Contains(file, "/vms/evm")
+			importsPseudo := isPackageIn(imp, "github.com/ava-labs/libevm/libevm/pseudo")
```
maru-ava: (No action required) Maybe `isPackageIn` -> `isImportIn` since `importPath` is being passed in??

vms/evm/imports_test.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+		}
+		return nil
+	})
+
+	require.NoErrorf(t, err, "filepath.Walk(%q)", root)
+	require.Empty(t, violations, "import violations found:\n%s", strings.Join(violations, "\n"))
+}
+
+func isPackageIn(importPath, root string) bool {
```
maru-ava: (No action required) Maybe something like `root` -> `targetedImport` for clarity?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if importPath == root {
+		return true
+	}
+	if len(importPath) < len(root)+1 {
```
maru-ava: (No action required) Maybe document your intent here? it's not obvious to me on first inspection what's going on.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,98 @@
...
+		return false
+	}
+	return strings.HasPrefix(importPath, root) && importPath[len(root)] == '/'
```
maru-ava: (No action required) Given that none of the usage of this helper supplies a value to `targetedImport` that is a single character in length, why does the second part of this conditional make sense? What conditions would it make sense? Maybe document your intent?

vms/evm/imports_test.go
ARR4N: > I'm saving this comment! This should be part of a style guide.

It [is](https://google.github.io/styleguide/go/decisions.html#:~:text=The%20general%20rule%20of%20thumb%20is%20that%20the%20length%20of%20a%20name%20should%20be%20proportional%20to%20the%20size%20of%20its%20scope%20and%20inversely%20proportional%20to%20the%20number%20of%20times%20that%20it%20is%20used%20within%20that%20scope.) ;)

vms/evm/imports_test.go
JonathanOppenheimer: 0 based indexing - this works right? I added a comment.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"io/fs"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+const root = ".."
```
ARR4N: Declarations should be made close to where they're used, not arbitrarily at the top of a file. The "rule" of placing constants at the top of a file is not a thing in Go.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+// grafting
+//
+// coreth can NOT be imported in the vms/evm package, because the goal is that vms/evm should
+// only contain the 'clean' properly uplifted code, that meets AvalancheGo quality standards.
+//
+// subnet-evm can NOT be imported anywhere in AvalancheGo besides /graft, because it must
+// not become a direct dependency of AvalancheGo.
+//
+// both coreth and subnet-evm can be imported in the vms/evm/emulate package, because it is a
+// temporary package that allows consumers to import both coreth and subnet-evm at the same time.
```
ARR4N:
```suggestion
// both coreth and subnet-evm can be imported in the vms/evm/emulate package, because it
// allows consumers to use both coreth and subnet-evm registration at the same time.
```

It's almost certainly not temporary. If it does get removed it will be part of a very significant overhaul.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+//
+// github.com/ava-labs/libevm/libevm/pseudo cannot be imported anywhere, because it requires so
+// much care to avoid catastrophic misuse and there are so few reasons to ever use it anyway
+//
+// TODO(jonathanoppenheimer): remove the emulate functionality once the emulate package is removed.
```
ARR4N:
```suggestion
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+// only contain the 'clean' properly uplifted code, that meets AvalancheGo quality standards.
+//
+// subnet-evm can NOT be imported anywhere in AvalancheGo besides /graft, because it must
+// not become a direct dependency of AvalancheGo.
+//
+// both coreth and subnet-evm can be imported in the vms/evm/emulate package, because it is a
+// temporary package that allows consumers to import both coreth and subnet-evm at the same time.
+//
+// github.com/ava-labs/libevm/libevm/pseudo cannot be imported anywhere, because it requires so
+// much care to avoid catastrophic misuse and there are so few reasons to ever use it anyway
```
ARR4N:
```suggestion
// github.com/ava-labs/libevm/libevm/pseudo cannot be imported anywhere, without review from any of
// @StephenButtolph, @ARR4N, or @joshua-kim. There is almost certainly a better option, and it exists
// only because libevm can't pollute the code base with generic type parameters.
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+//
+// TODO(jonathanoppenheimer): remove the emulate functionality once the emulate package is removed.
+// TODO(jonathanoppenheimer): remove the graft functionality once the graft package will be removed.
+func TestImportViolations(t *testing.T) {
+	var violations []string
```
ARR4N, line 57: Why add the layer of indirection? Just report the violation when it arises.

FWIW this is why forcing use of `require` instead of judicious use of `assert`/`require` as appropriate is a bad idea. If multiple violations occur then it forces whack-a-mole fixing and rerunning of the test every time. Strict adherence to a style guide is _not_ a valid reason to write more complicated code as it's antithetical to the purpose of a guide.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+		for _, spec := range node.Imports {
+			if spec.Path == nil {
+				continue
+			}
+			importPath := strings.Trim(spec.Path.Value, `"`)
+
+			inGraft := strings.Contains(file, "/graft")
+			inEmulate := strings.Contains(file, "/vms/evm/emulate")
+			inVMsEVM := strings.Contains(file, "/vms/evm")
```
ARR4N: `strings.Contains` is subject to false positives. First find the repo root with `filepath.Abs(root)` and then check for prefixes with `strings.HasPrefix(file, filepath.Join(repoRoot, "/graft"))` etc.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+			if slices.Contains(hasViolation, true) {
+				violations = append(violations, fmt.Sprintf("File %q imports %q", file, importPath))
+			}
+		}
+		return nil
+	})
+
+	require.NoErrorf(t, err, "filepath.Walk(%q)", root)
+	require.Empty(t, violations, "import violations found:\n%s", strings.Join(violations, "\n"))
```
ARR4N: (fyi) This comment will become obsolete once the `violations` slice isn't used, but there's no need to repeat the value in the error message as `require` will already do so.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// importPath must be at least len(targetedImport) + 1 to be a subpackage
+	// (e.g., "github.com/foo/x" is len("github.com/foo") + 2, minimum subpackage length)
+	if len(importPath) < len(targetedImport)+1 {
+		return false
+	}
+
+	// Check if importPath is a subpackage by ensuring it has the targetedImport prefix
+	// AND the next character is '/'. This is to prevent false positives where one
+	// package name is a prefix of another like "github.com/foo" vs "github.com/foobar".
```
ARR4N, line 118:
```suggestion
```

This just says what the code does so shouldn't be there as it's redundant and risks being or becoming incorrect. Heuristic: only explain "why" in inline comments, not how nor what. If a how or what is necessary then the code probably needs to be refactored.

[More details](https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html).

ARR4N (commented):
GitHub is forcing me to add a comment here for no reason (well, no reason beyond that it's a terrible product).

vms/evm/imports_test.go
maru-ava: `assert` was banned because in too many instances it was assumed to be equivalent to `require` (fail-fast) and this resulted in buggy tests. For those instances where multiple violations are possible, `t.Log` is always an option.

JonathanOppenheimer: This also will catch all the violations no? Fill up the array first, and then check the array at the end -- so it won't have to be rerun a ton of times.

vms/evm/imports_test.go
JonathanOppenheimer: I think the what is necessary as Maru requested an explanation of the intent behind the code here: https://github.com/ava-labs/avalanchego/pull/4537#discussion_r2578369379.

vms/evm/imports_test.go
JonathanOppenheimer: I've done so!

JonathanOppenheimer: Let me know if this what you had in mind.

vms/evm/imports_test.go
maru-ava: I assume he would prefer that the use of inline assertions so as to avoid the requirement to accumulate failures and output them at the end?

FWIW, I'm sympathetic to revisiting the `assert` ban, maybe something to bring up in retro? I would hope that we now have enough review maturity to avoid misuse going forward.

JonathanOppenheimer: Yah, I understand. If we'd like to revisiting the ban, that totally makes sense, as it would certainly enable additional testing patterns, but I don't think this is the PR to do that in 😂. For retro!

ARR4N (approved):
Feel free to open a PR to move pseudo into `internal/libevm` instead of checking here. If I remember I'll do it myself but it's low priority so I may not get around to it soon.

JonathanOppenheimer:
> Feel free to open a PR to move pseudo into `internal/libevm` instead of checking here. If I remember I'll do it myself but it's low priority so I may not get around to it soon.

I've done so! That PR is available here: https://github.com/ava-labs/libevm/pull/250

I've removed the pseudo check from this PR, and also remove the tests this replaces. This PR is relevant as Austin and I worked through the subnet-evm monorepo PRs, with some assumptions that this PR enforces.

JonathanOppenheimer:
The above PR is now merged in. Thanks to @maru-ava for the suggestion.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+// grafting
+//
+// coreth can NOT be imported in the vms/evm package, because the goal is that vms/evm should
+// only contain the 'clean' properly uplifted code, that meets AvalancheGo quality standards.
+//
+// subnet-evm can NOT be imported anywhere in AvalancheGo besides graft/subnet-evm itself,
+// because it must not become a direct dependency of AvalancheGo or mix directly with coreth.
+//
+// graft/evm is the shared code that both coreth and subnet-evm will depend on, so it must not
+// import from either of them to maintain a one-way dependency direction.
```
maru-ava: nit:  `maintain a one-way dependency direction` -> `avoid circular dependency`

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+//
+// both coreth and subnet-evm can be imported in the vms/evm/emulate package, because it
+// allows consumers to use both coreth and subnet-evm registration at the same time.
+//
+// TODO(jonathanoppenheimer): remove the graft functionality once the graft package will be removed.
```
maru-ava: nit: `will be removed` -> `is removed`

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestImportViolations(t *testing.T) {
+	const root = "../.."
+	repoRoot, err := filepath.Abs(root)
+	require.NoError(t, err)
+
+	graftDir := filepath.Join(repoRoot, "graft")
```
maru-ava, line 51: (No action required) Maybe simplify with composition?

```golang
graftDir := filepath.Join(repoRoot, "graft")
graftEVMDir := filepath.Join(graftDir, "evm")
graftSubnetEVMDir := filepath.Join(graftDir, "subnet-evm")
vmsEVMDir := filepath.Join(repoRoot, "vms", "evm")
emulateDir := filepath.Join(vmsEVMDir, "emulate")
```

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+			inGraft := strings.HasPrefix(absFile, graftDir)
+			inGraftEVM := strings.HasPrefix(absFile, graftEVMDir)
+			inGraftSubnetEVM := strings.HasPrefix(absFile, graftSubnetEVMDir)
+			inEmulate := strings.HasPrefix(absFile, emulateDir)
+			inVMsEVM := strings.HasPrefix(absFile, vmsEVMDir)
+			importsCoreth := isImportIn(importPath, "github.com/ava-labs/avalanchego/graft/coreth")
+			importsSubnetEVM := isImportIn(importPath, "github.com/ava-labs/avalanchego/graft/subnet-evm")
+
+			hasViolation := []bool{
```
maru-ava: Apologies if I missed previous review feedback that justifies this (gerrit ftw), but maybe simplify to a boolean expression that more closely matches the docstring rational?

```golang
hasViolation := (importsCoreth && inVMsEVM && !inEmulate ||
    importsSubnetEVM && !(inGraftSubnetEVM || inEmulate) ||
    (importsCoreth || importsSubnetEVM) && inGraftEVM)
if hasViolation {...}
```

`!inGraft` appeared redundant since inVMsEVM is not in graft.

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestImportViolations ensures proper import rules:
+// - graft/coreth can be imported anywhere EXCEPT vms/evm (but vms/evm/emulate is an exception)
+// - graft/subnet-evm can only be imported within graft/subnet-evm itself and vms/evm/emulate
+// - graft/evm can NOT import from graft/coreth or graft/subnet-evm
```
alarso16, line 26: Should this be true? See #4688. How else can we test code with registrations? Not saying this is incorrect, but we definitely need a conversation about it

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestImportViolations(t *testing.T) {
+	const root = "../.."
+	repoRoot, err := filepath.Abs(root)
+	require.NoError(t, err)
+
+	graftDir := filepath.Join(repoRoot, "graft")
+	graftEVMDir := filepath.Join(repoRoot, "graft", "evm")
+	graftSubnetEVMDir := filepath.Join(repoRoot, "graft", "subnet-evm")
+	emulateDir := filepath.Join(repoRoot, "vms", "evm", "emulate")
+	vmsEVMDir := filepath.Join(repoRoot, "vms", "evm")
```
alarso16, line 54:
```suggestion
	graftDir := filepath.Join(repoRoot, "graft")
	graftEVMDir := filepath.Join(graftDir, "evm")
	graftSubnetEVMDir := filepath.Join(graftDir, "subnet-evm")
	vmsEVMDir := filepath.Join(repoRoot, "vms", "evm")
	emulateDir := filepath.Join(vmsEVMDir "emulate")
```
Why not reuse definitions?

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+				continue
+			}
+			importPath := strings.Trim(spec.Path.Value, `"`)
+
+			inGraft := strings.HasPrefix(absFile, graftDir)
+			inGraftEVM := strings.HasPrefix(absFile, graftEVMDir)
+			inGraftSubnetEVM := strings.HasPrefix(absFile, graftSubnetEVMDir)
+			inEmulate := strings.HasPrefix(absFile, emulateDir)
+			inVMsEVM := strings.HasPrefix(absFile, vmsEVMDir)
+			importsCoreth := isImportIn(importPath, "github.com/ava-labs/avalanchego/graft/coreth")
```
alarso16: nit: constant defined outside of callback?

vms/evm/imports_test.go
alarso16: Why is this a slice? You can just put some `||`. Also all of this boolean logic is a little janky. Think about how these subsets of events interact

vms/evm/imports_test.go
```diff
@@ -0,0 +1,118 @@
...
+			return fmt.Errorf("parser.ParseFile(..., %q, ...): %w", file, err)
+		}
+
+		for _, spec := range node.Imports {
+			if spec.Path == nil {
+				continue
+			}
+			importPath := strings.Trim(spec.Path.Value, `"`)
+
+			inGraft := strings.HasPrefix(absFile, graftDir)
```
alarso16: These don't need redone for each import. Can you move them outside?

alarso16 (changes requested):
This is definitely worth some more thought about exactly what we want to enforce, and the boolean logic is overly complicated (and incorrect I think)

vms/evm/imports_test.go
JonathanOppenheimer: you and Austin suggested the same - thank you!

vms/evm/imports_test.go
JonathanOppenheimer: thank you for the simplifiication!

JonathanOppenheimer: This is resolved with https://github.com/ava-labs/avalanchego/pull/4537#discussion_r2623282192 -- lmk if you still have reservations.

vms/evm/imports_test.go
JonathanOppenheimer: `inGraft` was removed, and `importPath` depends on `spec.Path.Value`, which is the loop variable so it must remain inside the loop.

vms/evm/imports_test.go
JonathanOppenheimer: fixed!

vms/evm/imports_test.go
JonathanOppenheimer: My apologies - I didn't mean to presumptuous. My understanding from standup yesterday and the converstaion is that allowing this would create an annoying circular dependency (if shared evm imports the evm repos and vice versa). 

Should we not enforce this and be okay with the circular dependencies? It's very easy to remove, and I do see how you can test it like that.

vms/evm/imports_test.go
maru-ava: Good catch, the values constructed from `absFile` can be moved outside. Only the ones dependent on `importPath` need to stay.

JonathanOppenheimer: ah sorry I see what you mean -- I move those outside of the loop!

vms/evm/imports_test.go
JonathanOppenheimer: After discussing as a team, we have decided that we would like to prevent circular dependencies between these two, to allow packages in `graft/evm` to be uplifted to `vms/evm` without additional work.

alarso16 (approved):

PR #4538, opened 2025-11-23
feat(gossip): make `BloomFilter` completely thread-safe and introduce `ResetIfNeeded()` method

ARR4N:
## Why this should be merged

Since `gossip.ResetBloomFilterIfNeeded()` wasn't thread-safe, users were required to take out their own lock. However the check for "`IfNeeded`" only requires a read lock (avoiding unnecessarily locking out other goroutines), but this isn't possible externally. Without this, all new SAE transactions received over RPC will be handled sequentially.

Since this change no longer requires consumers to hold a hold while resetting, it could result in a race between `Marshal()` and refilling after a reset, resulting in marshalling an empty filter. The new `BloomFilter.ResetIfNeeded()` method therefore also accepts an iterator that it uses to refill the filter while still holding the lock.

While I could have exposed an `IsResetNeeded() bool` method, this would force every consumer to implement the same logic. I also made `func(*BloomFilter, ...)` functions into methods on the type as part of a general tidying of the file.

## How this works

Adds a `sync.RWMutex` into `BloomFilter`. Methods that were already thread-safe are considered reads while resetting is considered a write.

## How this was tested

Existing UT for general functionality. Actually testing the thread safety is difficult (is it even possible?) so I simply clobber all the methods in an attempt to trigger `go test -race` errors. Refilling has a new UT.

No

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1138,14 +1138,17 @@ func getCPUTargeterConfig(v *viper.Viper) (tracker.TargeterConfig, error) {
 	}
 }
 
-func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warningThresholdAvailableDiskSpace uint64, err error) {
+func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warningThresholdAvailableDiskSpace uint64, warningThresholdAvailableDiskSpacePercentage uint64, err error) {
 	requiredAvailableDiskSpace = v.GetUint64(SystemTrackerRequiredAvailableDiskSpaceKey)
 	warningThresholdAvailableDiskSpace = v.GetUint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey)
+	warningThresholdAvailableDiskSpacePercentage = v.GetUint64(SystemTrackerWarnThreshAvailDiskSpacePercentageKey)
 	switch {
+	case warningThresholdAvailableDiskSpacePercentage > 50:
+		return 0, 0, 0, fmt.Errorf("%q (%d) must be in [0, 50]", SystemTrackerWarnThreshAvailDiskSpacePercentageKey, warningThresholdAvailableDiskSpacePercentage)
 	case warningThresholdAvailableDiskSpace < requiredAvailableDiskSpace:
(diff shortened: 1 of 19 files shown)


network/p2p/gossip/bloom_test.go
```diff
@@ -106,3 +107,33 @@ func TestBloomFilterRefresh(t *testing.T) {
 		})
 	}
 }
+
+func TesetBloomFilterClobber(t *testing.T) {
```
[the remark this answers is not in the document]
ARR4N: I will not allow a glorified spell-check to be considered a contributor to this PR.

network/p2p/gossip/bloom.go
```diff
@@ -16,8 +17,7 @@ import (
 // anticipated at any moment, and a false positive probability of [targetFalsePositiveProbability]. If the
 // false positive probability exceeds [resetFalsePositiveProbability], the bloom filter will be reset.
 //
-// Invariant: The returned bloom filter is not safe to reset concurrently with
-// other operations. However, it is otherwise safe to access concurrently.
+// The returned bloom filter is safe for concurrent usage.
```
StephenButtolph, line 21: Does this refactor encourage misuse of this bloom filter?

Specifically, it seems like this change does not expect for the bloom filter to be refilled atomically with its reset. This code certainly can be cleaned up... But adding this locking behavior doesn't seem useful (when using the bloom filter correctly).

If the bloom filter is accessed in a racy manner (Specifically `Marshal` being called concurrently with `ResetIfNeeded` followed by the expected `Add` calls), wouldn't we send a (potentially empty) bloom filter even though we may have a large number of items still in the set?

ARR4N: Yeah I see what you mean. I think the best way to address that along with my intended improvement is to add a `afterReset func()` argument, which is called while still holding the write lock, and expect refills to be in there. WDYT?

EDIT: I implemented it in the latest commit (7118ad65123872d59d7cc7f89c02da8a626624b5) because it's easy to revert but does a better job of communicating my thinking.

ARR4N: I changed from a general callback to an `iter.Seq[Gossipable]` to reduce the scope to exactly what's needed.

PR #4544, opened 2025-11-25
[bazel] Enable initial build and test

maru-ava:
## Why this should be merged

The inclusion of multiple go modules in this repo and the desire to similarly migrate firewood suggested a tool like [Bazel](https://bazel.build/). Bazel enables a unified dependency graph across both go modules and other languages like rust and solidity which can minimize the cost of this kind of complexity.

## How this works

- See docs/bazel.md for implementation details 
- This PR is intentionally limited in scope to bazel gaining the ability to build avalanchego, run unit tests, and run e2e tests against a bazel-built binary. Future PRs are intended to iteratively apply Bazel where it makes sense to do so.

## How this was tested

CI now includes a `Bazel Tests` workflow that runs e2e and unit test jobs built and run by bazel.

N/A

## TODO

 - [x] Merge #4940
 - [x] Get approval for this PR
 - [x] Merge #4942 into this PR


graft/subnet-evm/tests/genesis.go
```diff
@@ -9,5 +9,5 @@ import (
 
 // Genesis contains the embedded genesis.json file used by tests.
 //
-//go:embed load/genesis.json
+//go:embed genesis.json
```
maru-ava, line 12: Bazel didn't seem to like embedding from other packages

.bazel/patches/com_github_consensys_gnark_crypto_asm_includes.patch
```diff
@@ -0,0 +1,8093 @@
+diff --git a/ecc/bls12-377/fp/BUILD.bazel b/ecc/bls12-377/fp/BUILD.bazel
```
maru-ava: This is...not ideal. Alternate proposals welcome

docs/bazel.md
```diff
@@ -0,0 +1,648 @@
+# Bazel Build System
+
+This document explains how Bazel is configured and used in the
+avalanchego monorepo.
+
+## Quick Start
+
```
ARR4N, line 15: We should recommend usage of `bazelisk` instead of direct `bazel` installation.

(note to self) This file still needs reviewing.

MODULE.bazel
```diff
@@ -0,0 +1,158 @@
...
+    patches = ["//.bazel/patches:com_github_supranational_blst.patch"],
+    path = "github.com/supranational/blst",
+)
+go_deps.gazelle_override(
+    build_file_generation = "off",
+    path = "github.com/supranational/blst",
+)
+
+# firewood: Disable gazelle and provide custom BUILD file for CGO FFI
+go_deps.module_override(
```
ARR4N: Starlark[^1] has basic Python-like syntax so you can loop over lists and dictionaries (including as comprehensions). `MODULE.bazel` is very limited in what's allowed so this is the cleanest pattern I could get:

```starlark
# Patches for external dependencies that need custom BUILD files or fixes
GO_OVERRIDES = {
    "github.com/supranational/blst": {
        "build_gen": "off",
        "patches": ["//.bazel/patches:com_github_supranational_blst.patch"],
    },
    "github.com/ava-labs/firewood-go-ethhash/ffi": {
        "build_gen": "off",
        "patches": ["//.bazel/patches:com_github_ava_labs_firewood_go_ethhash_ffi.patch"],
    },
    "github.com/consensys/gnark-crypto": {
        "patches": ["//.bazel/patches:com_github_consensys_gnark_crypto_asm_includes.patch"],
    },
    "github.com/ava-labs/libevm": {
        "patches": [
            "//.bazel/patches:com_github_ava_labs_libevm_libsecp256k1.patch",
            "//.bazel/patches:com_github_ava_labs_libevm_secp256k1.patch",
        ],
        "gazelle_directives": [
            "gazelle:exclude crypto/secp256k1/**",
        ],
    },
}

[go_deps.module_override(
    patch_strip = 1,
    patches = o.get("patches", []),
    path = path,
) for path, o in GO_OVERRIDES.items()]

[go_deps.gazelle_override(
    build_file_generation = o.get("build_gen", "on"),
    directives = o.get("gazelle_directives", []),
    path = path,
) for path, o in GO_OVERRIDES.items()]
```

[^1]: I left the G 2 years before it was renamed so I might accidentally call it Skylark sometimes.

Taskfile.yml
```diff
@@ -12,6 +12,92 @@ version: '3'
...
+  bazel-build-race:
+    desc: Build avalanchego with Bazel (race detection)
+    cmd: bazel build --config=race //main:avalanchego
+
+  bazel-build-opt:
+    desc: Build avalanchego with Bazel (optimized)
+    cmd: bazel build --compilation_mode=opt //main:avalanchego
+
+  bazel-gazelle-delete:
+    desc: Delete all gazelle-generated BUILD.bazel files (except manually maintained ones)
```
ARR4N: Why would we want to do this?

BUILD.bazel
```diff
@@ -0,0 +1,11 @@
...
+
+# gazelle:prefix github.com/ava-labs/avalanchego
+# gazelle:exclude tools
+# gazelle:exclude .direnv
+# gazelle:proto disable_global
+
+gazelle(
+    name = "gazelle",
+    command = "update",
+)
```
ARR4N, line 20:
```suggestion
gazelle(
    name = "gazelle",
    command = "fix",
)

gazelle(
    name = "gazelle_diff",
    command = "fix",
    extra_args = ["--mode=diff"],
)
```

`fix` [does a little more](https://github.com/bazel-contrib/bazel-gazelle/blob/master/gazelle-reference.md#fix-and-update) than `update`. I find that also having a `diff` version available is very useful for CI.

BUILD.bazel
```diff
@@ -0,0 +1,11 @@
+load("@bazel_gazelle//:def.bzl", "gazelle")
+
+# gazelle:prefix github.com/ava-labs/avalanchego
+# gazelle:exclude tools
+# gazelle:exclude .direnv
+# gazelle:proto disable_global
```
ARR4N, line 9: :sad_homer:

proto definitely out of scope for this PR though.

Taskfile.yml
```diff
@@ -66,6 +152,18 @@ tasks:
...
     cmd: ./scripts/build_xsvm_image.sh
 
+  check-bazel-fmt:
+    desc: Check that BUILD.bazel files are formatted with buildifier
+    cmds:
+      - task: bazel-fmt
+      - task: check-clean-branch
+
+  check-bazel-gazelle-generate:
+    desc: Check that generated BUILD files are up-to-date (requires clean git tree)
```
ARR4N: See my comment on `BUILD.bazel`, which introduces `//:gazelle_diff` for this purpose. It obviates the need for a clean git tree, which will also help the `lint-bazel` task below.

.bazelrc
```diff
@@ -0,0 +1,50 @@
+# Go settings - match scripts/build_test.sh defaults
```
ARR4N, line 4: (note to self) This still requires review.

.bazel/defs.bzl
```diff
@@ -0,0 +1,13 @@
+"""Custom Bazel macros for avalanchego."""
+
+load("@io_bazel_rules_go//go:def.bzl", _go_test = "go_test")
```
ARR4N, line 5: Why the alias as `_go_test`?

I'm 99% sure you can:

```starlark
load("@io_bazel_rules_go//go:def.bzl", "go_test")
```
and then
```starlark
go_test(**kwargs)
```

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
```diff
@@ -0,0 +1,77 @@
+diff --git a/BUILD.bazel b/BUILD.bazel
+new file mode 100644
+--- /dev/null
++++ b/BUILD.bazel
```
ARR4N, line 4: Pre-MODULEs there was a way to just specify a BUILD file :( anywho...

I think these would be easier to read and modify if we jerry-rigged an equivalent solution that had actual BUILD files somewhere like `.bazel/patches/build_files` and then a task that created the patches by diffing against null.

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
```diff
@@ -0,0 +1,77 @@
...
++    name = "firewood_ffi",
++    hdrs = ["firewood.h"],
++    deps = [":firewood_ffi_lib"],
++    linkopts = ["-lm"],
++    visibility = ["//visibility:public"],
++)
++
++go_library(
++    name = "ffi",
++    srcs = [
```
ARR4N: This will break every time they modify the file names. I'd use something like:

```starlark
    srcs = ["firewood.h"] + glob(
        include = ["*.go"],
        exclude = ["*_test.go"],
    )
```

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
```diff
@@ -0,0 +1,77 @@
...
++
++alias(
++    name = "go_default_library",
++    actual = ":ffi",
++    visibility = ["//visibility:public"],
++)
++
++go_test(
++    name = "ffi_test",
++    srcs = [
```
ARR4N: As above re `glob()`.  Applies to all patch files.

.bazel/patches/BUILD.bazel
```diff
@@ -0,0 +1,3 @@
+# Patches for external dependencies that don't work with Bazel out of the box.
```
ARR4N, line 1: What is your thinking re `build_file_generation = "off"` + patch, vs actually patching what is created? There are definitely pros and cons to each so I'm interested in your rationale before weighing in or not.

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
```diff
@@ -0,0 +1,77 @@
...
++)
++
++# Wrapper cc_library to use with cdeps
++cc_library(
++    name = "firewood_ffi",
++    hdrs = ["firewood.h"],
++    deps = [":firewood_ffi_lib"],
++    linkopts = ["-lm"],
++    visibility = ["//visibility:public"],
++)
```
ARR4N, line 28: This whole section makes me unreasonably happy :) I think it's that it hides sooooo much otherwise complex configuration, and does it really elegantly.

.bazel/patches/com_github_consensys_gnark_crypto_asm_includes.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/ecc/bls12-381/fp/BUILD.bazel b/ecc/bls12-381/fp/BUILD.bazel
+--- a/ecc/bls12-381/fp/BUILD.bazel
++++ b/ecc/bls12-381/fp/BUILD.bazel
+@@ -3,4 +3,5 @@
+ go_library(
+     name = "fp",
++    tags = ["no-sandbox"],
```
ARR4N, line 12: I can't find this build tag anywhere in their repo. Why is it necessary?

.bazel/patches/com_github_ava_labs_libevm_secp256k1.patch
```diff
@@ -0,0 +1,31 @@
+--- /dev/null
```
ARR4N: Once we've caught up with geth and managing `libevm` isn't so precarious I'll Bazelify it to make your job here easier.

.bazel/patches/com_github_supranational_blst.patch
```diff
@@ -0,0 +1,149 @@
...
+new file mode 100644
+--- /dev/null
++++ b/bindings/BUILD.bazel
+@@ -0,0 +1,9 @@
++# gazelle:ignore
++# Header library for blst C bindings
++
++cc_library(
++    name = "bindings",
++    hdrs = glob(["*.h"], allow_empty = True),
```
ARR4N: The `allow_empty`s in here seem to be a code smell. My gut tells me that either we enforce that there's something, or we remove the argument entirely.

That said, I feel out of my depth in this particular file.

.bazel/patches/com_github_supranational_blst.patch
```diff
@@ -0,0 +1,149 @@
...
++++ b/bindings/go/BUILD.bazel
+@@ -0,0 +1,83 @@
++# gazelle:ignore
++# Go bindings for blst library
++#
++# Compiler flags are derived from the CGO directives in blst.go:13-16.
++
++load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
++
++COPTS = [
```
ARR4N, line 27: (fyi) I have no idea how to review this section.

.github/workflows/ci.yml
```diff
@@ -42,6 +42,14 @@ jobs:
       - name: test-fuzz
         shell: bash
         run: ./scripts/run_task.sh test-fuzz
+  Bazel_Unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-go-for-project
```
ARR4N: Why is this still necessary? Isn't Bazel taking over this part?

.github/workflows/ci.yml
```diff
@@ -117,6 +125,24 @@ jobs:
...
           loki_username: ${{ secrets.LOKI_USERNAME || '' }}
           loki_password: ${{ secrets.LOKI_PASSWORD || '' }}
+  Bazel_e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run e2e tests with Bazel-built binary
+        uses: ./.github/actions/run-monitored-tmpnet-cmd
+        with:
+          run: ./scripts/run_task.sh bazel-test-e2e-ci
```
ARR4N: (fyi) I'll take your word for it that this is the correct setup. The proof will be in the CI pudding anyway.

graft/coreth/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("@bazel_gazelle//:def.bzl", "gazelle")
+
+# gazelle:prefix github.com/ava-labs/avalanchego/graft/coreth
+# gazelle:map_kind go_test graft_go_test //.bazel:defs.bzl
+
+gazelle(
```
ARR4N: See my comment on the top-level `BUILD.bazel` re `fix` instead of `update` + introduction of `gazelle_diff`. This applies to `//graft/{evm,subnet-evm}` too.

Side note: it's interesting that the `use_repo` directive makes use of `go.work` but that we need separate instances of Gazelle in the different modules.

graft/coreth/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("@bazel_gazelle//:def.bzl", "gazelle")
+
+# gazelle:prefix github.com/ava-labs/avalanchego/graft/coreth
+# gazelle:map_kind go_test graft_go_test //.bazel:defs.bzl
```
ARR4N, line 2: TIL about `map_kind`. Nice!

graft/subnet-evm/tests/genesis.go
ARR4N: If in the future you don't want to move files, you can always add a target in the same package, aliasing or (shadow) copying the original. My first approach would be to see if [copy_file()](https://registry.bazel.build/docs/bazel_lib#lib-copy_file-bzl) is sufficient.

main/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
...
+        "//config",
+        "//graft/coreth/plugin/evm",
+        "//version",
+        "@com_github_spf13_pflag//:pflag",
+        "@org_golang_x_term//:term",
+    ],
+)
+
+go_binary(
+    name = "avalanchego",  # keep
```
ARR4N: Alternatively, I think you could remove the `#keep` and use:

```starlark
alias(
    name = "avalanchego",
    actual = ":main",
)
```

Not something I feel strongly about. Only that `#keep` directives jump out at me as something requiring closer review.

tests/e2e/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+    name = "e2e",
+    srcs = ["ignore.go"],
+    importpath = "github.com/ava-labs/avalanchego/tests/e2e",
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "e2e_test",
+    srcs = ["e2e_test.go"],
+    tags = ["manual"],  # keep
```
ARR4N, line 14: I'm surprised that `#keep` is necessary on tags 🤷

.github/workflows/ci.yml
```diff
@@ -42,6 +42,14 @@ jobs:
...
         shell: bash
         run: ./scripts/run_task.sh test-fuzz
+  Bazel_Unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-go-for-project
+      - name: Run unit tests with Bazel
+        shell: bash
+        run: ./scripts/run_task.sh bazel-test
```
PlatCore: If "Bazel" is "taking over this part", the `bazel` command must be provisioned first (by the Bazel / Bazelisk package, depending on the OS / package manager).
In the (default) `ubuntu-latest` runner there is no Bazel/Bazelisk, so these steps (and the corresponding Bazel_e2e job) should fail. 
Consider installing Bazelisk/Bazel (or using the nix action like the lint job) before running the Bazel tasks.

maru-ava: As per the [passing tests](https://github.com/ava-labs/avalanchego/actions/runs/21743200179/job/62722852109?pr=4942) on #4942, the runners do have bazelisk installed and it installs the bazel version defined in .bazelversion.

.github/workflows/ci.yml
maru-ava: This enables go-task to be started with by run_task.sh with `go run`.  Not a forever decision, but simpler for now. Added a comment documenting this.

.bazel/patches/BUILD.bazel
maru-ava: I used `"off"` where it seemed to be more expedient, but the resulting fragility is probably unjustified. I've proposed switching to generating the patches from readable BUILD sources and added content to docs/bazel.md documenting how to decide when to use either approach.

.bazel/patches/com_github_supranational_blst.patch
maru-ava: Stripped unsupported platforms.

.github/workflows/ci.yml
maru-ava: This job is passing on the subsequent PR.

docs/bazel.md
maru-ava: Updated to reflect usage of bazelisk.

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
maru-ava: Updated to generate the patches.

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
maru-ava: Done

.bazel/patches/com_github_ava_labs_firewood_go_ethhash_ffi.patch
maru-ava: Done

.bazel/patches/com_github_consensys_gnark_crypto_asm_includes.patch
maru-ava: This is a bazel tag, it disables sandboxing on this target to work around relative imports that can't resolve otherwise. See 'Patching Strategies' section of bazel.md.

This is a stop-gap that reduces the patch size required for bazelification. It's not intended to be used for long because it breaks both hermeticity and remote execution.

Updated to add a comment to both uses that documents the rational and points to the readme.

.bazel/patches/com_github_supranational_blst.patch
maru-ava: Good catch, removed.

main/BUILD.bazel
maru-ava: Done

tests/e2e/BUILD.bazel
maru-ava: Good catch, removed.

BUILD.bazel
maru-ava: Switched to fix, but added `gazelle_test` with `--mode=diff` as a proper test target. Is that ok?

.bazel/defs.bzl
maru-ava: Likely an editing artifact, but it's back now because of the need to move `-test.shuffle=on` to a new `go_test` wrapper in this file. I've documented the requirement in a new comment.

graft/coreth/BUILD.bazel
maru-ava: > See my comment on the top-level BUILD.bazel re fix instead of update + introduction of gazelle_diff. This applies to //graft/{evm,subnet-evm} too.

Done

> Side note: it's interesting that the use_repo directive makes use of go.work but that we need separate instances of Gazelle in the different modules.

Good catch. Switched to a single invocation.

Taskfile.yml
maru-ava: Done

Taskfile.yml
maru-ava: For now it's just a helper enabling me to keep the purely generated files in a separate PR. afaik we might want to keep it to ensure that orphaned build files get removed (e.g. in the event of package removal)?

scripts/set_go_version.sh
```diff
@@ -21,6 +21,10 @@ version="$1"
 go work edit -go="$version" go.work
 echo "updated go.work"
 
+# Update MODULE.bazel go_sdk.download version
+sed -i "s/^go_sdk\.download(version = \".*\")$/go_sdk.download(version = \"$version\")/" MODULE.bazel
```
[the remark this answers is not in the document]
maru-ava: Done

MODULE.bazel
maru-ava: Done

tests/e2e/BUILD.bazel
maru-ava: Actually, # keep is what the deletion task is keyed on.

BUILD.bazel
ARR4N: :100:

TIL about `gazelle_test`

tests/e2e/BUILD.bazel
ARR4N: The idea being that if you've made manual changes then you don't want to accidentally wipe them?

Let's keep it as is because I think the deletion task is more part of your current process than something we'll necessarily keep long-term, and if not then we can decide on how to approach this later. My primary concern is that overloading of meaning never ends well.

.bazel/patches/build_files/com_github_ava_labs_firewood_go_ethhash_ffi/BUILD.bazel
ARR4N, line 1: Why do the others need `#gazelle:ignore` but this doesn't?

.bazel/patches/BUILD.bazel
ARR4N, line 1: Reading the complexity of the manual patches reminded me that I used to expose their tests to the current workspace for peace of mind:

```starlark
# Expose the tests / build targets of all patched targets to ensure that they're
# included by CI as they're not covered by default.
test_suite(
    name = "external_tests",
    tests = [
        "@com_github_ava_labs_libevm//crypto/secp256k1:secp256k1_test",
# ...
    ],
)
```

Then a bazel test //... or bazel test //.bazel/patches/build_files:external_tests will run them.

The `test_suite` rule is really useful for collecting bunches of tests that need special treatment. For example, I sometimes declare a `ci` target to collect tests tagged as "manual".

.bazel/defs.bzl
```diff
@@ -0,0 +1,34 @@
...
+# The _go_test alias is required because Starlark forbids a load()
+# binding and def with the same name in the same file.
+load("@io_bazel_rules_go//go:def.bzl", _go_test = "go_test")
+
+def go_test(**kwargs):
+    """go_test wrapper that enables shuffle by default.
+
+    This wrapper exists because .bazelrc --test_arg applies to all test
+    targets including gazelle_test, which doesn't understand Go test
+    flags. By wrapping go_test and injecting shuffle via args, it only
```
ARR4N:
```suggestion
    This wrapper exists because .bazelrc --test_arg applies to test
    targets for all languages / rules, which don't understand Go's test
    flags. By wrapping go_test and injecting shuffle via args, it only
```

scripts/generate_bazel_patches.sh
```diff
@@ -0,0 +1,51 @@
...
+BUILD_FILES_DIR="${REPO_ROOT}/.bazel/patches/build_files"
+PATCHES_DIR="${REPO_ROOT}/.bazel/patches"
+
+for module_dir in "${BUILD_FILES_DIR}"/*/; do
+  module_name="$(basename "${module_dir}")"
+  patch_file="${PATCHES_DIR}/${module_name}.patch"
+
+  # Collect all BUILD.bazel files relative to the module directory
+  patch_content=""
+  while IFS= read -r -d '' build_file; do
```
ARR4N, line 36: Why build it manually instead of using the `diff` command?

scripts/run_bazel.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
ARR4N, line 1: I think it's worth commenting that this is only for automation and that a regular user should just install `bazelisk` and alias it to `bazel`.

.gitignore
```diff
@@ -68,4 +72,4 @@ __debug_*
 
 # polyrepo
 polyrepo.log
-/firewood/
+/firewood/
```
ARR4N:
```suggestion
/firewood/

```

I think you accidentally removed the new line.

MODULE.bazel
```diff
@@ -0,0 +1,145 @@
...
+bazel_dep(name = "rules_go", version = "0.57.0", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "gazelle", version = "0.45.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "rules_proto", version = "7.0.2")
+bazel_dep(name = "protobuf", version = "29.3", repo_name = "com_google_protobuf")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+
+# Go SDK - version should match go.mod and nix/go/default.nix
+go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
+go_sdk.download(version = "1.25.7")
```
ARR4N:
```suggestion
go_sdk.from_file(go_work = "//:go.work")
```

Or `go_sdk.from_file(go_mod = "//:go.mod")`.

[Reference](https://github.com/bazel-contrib/rules_go/blob/master/docs/go/core/bzlmod.md#go-sdks)

Then you can revert the changes to both `check_go_version.sh` and `set_go_version.sh`.

FWIW, I'd originally left a comment on `check_go_version.sh` suggesting extracting the version by running `bazel run @io_bazel_rules_go//go -- version` instead of parsing the file. It's redundant now, but sharing because it's a useful way to run Go inside a Bazel workspace.

Taskfile.yml
```diff
@@ -16,6 +16,94 @@ includes:
...
+
+  bazel-clean:
+    desc: Clean Bazel cache
+    cmd: ./scripts/run_bazel.sh clean
+
+  bazel-clean-all:
+    desc: Clean Bazel cache (full expunge)
+    cmd: ./scripts/run_bazel.sh clean --expunge
+
+  bazel-query:
```
ARR4N: Not specific to this one, but what's the rationale behind the indirection via task? Why wouldn't I just use `bazel query` or `bazel run //:gazelle` directly instead of having to learn the aliases? I'm wary about "training" the team into non-idiomatic usage.

Taskfile.yml
```diff
@@ -75,6 +163,22 @@ tasks:
...
+
+  check-bazel-generate-patches:
+    desc: Checks that generated patches are up-to-date (requires a clean working tree)
+    cmds:
+      - task: bazel-generate-patches
+      - task: check-clean-branch
+
+  check-bazel-gazelle-generate:
+    desc: Check that generated BUILD files are up-to-date
+    cmd: ./scripts/run_bazel.sh test //:gazelle_test //graft/coreth:gazelle_test //graft/evm:gazelle_test //graft/subnet-evm:gazelle_test
```
ARR4N: Why could their `gazelle` targets be unified but not the `gazelle_test` targets?

tests/e2e/BUILD.bazel
maru-ava: The deletion task exists to clean up orphaned `BUILD.bazel` files — afaict, when a go package is removed, `gazelle` won't delete its `BUILD.bazel` file on its own. `gazelle fix --mode=diff` detects drift in existing files but won't flag `BUILD.bazel` files left behind for deleted packages.

That said, I agree it's more of a current-process tool than something permanent. If we find a better approach to orphan detection (e.g., a gazelle plugin or a dedicated lint check), we can replace it.

.bazel/patches/build_files/com_github_ava_labs_firewood_go_ethhash_ffi/BUILD.bazel
maru-ava: Good catch. It should have `# gazelle:ignore` for consistency with the others. Since firewood uses `build_file_generation = "off"`, gazelle won't touch it regardless, but the directive is good defense-in-depth. Done.

fwiw, #5050 will remove this patch.

tests/e2e/BUILD.bazel
maru-ava: ~Also, added a TODO to the deletion task documenting the desire to replace its use with something more targeted.~

edit: Added a TODO to remove the deletion task before merge. proto and mock generation semantics were coloring my judgement.

BUILD.bazel
maru-ava: That's next!

.bazel/patches/BUILD.bazel
maru-ava: Added `test_suite` as suggested, targeting secp256k1. The firewood ffi test needed work and I didn't figure it worth it since migration will enable testing directly, blst was also broken and I figure we can fix it when needed.

scripts/generate_bazel_patches.sh
maru-ava: afaict bazel's internal patch parser expects git-style unified diffs (`diff --git` header, new file mode, no timestamps on ---/+++ lines). Plain `diff -uN` doesn't produce that format, so we'd need sed fixups to add the git headers and strip timestamps. The resulting code is about the same complexity as the current approach. I've added a similar comment to the script for the benefit of future maintainers.

MODULE.bazel
maru-ava: Done

scripts/run_bazel.sh
maru-ava: Added a comment.

fwiw bazel.md already documents that nix shell provides both bazelisk and aliases it to bazel. The easiest way to use the nix shell is to have nix and direnv installed and set `AVALANCHEGO_DIRENV_USE_FLAKE=1`.

Taskfile.yml
maru-ava: Good catch. All gazelle_test targets referenced workspace = "//:BUILD.bazel" and produced identical output. Unified.

Taskfile.yml
maru-ava: Task serves two purposes: trivial local reproducibility of CI jobs and discoverability of common actions. While invoking bazel directly is likely to become more prevalent over time, it will still be valuable to define tasks - aliases, effectively - that both humans and CI can invoke consistently. The alternative would be regressing to hard-coding a specific pattern in CI such that reproducing it would require reading the CI job definition for the exact invocation needed to try and reproduce a failed CI job. 

That said, while I think this makes sense for github actions and a team fresh to bazel, we can always revisit when our circumstances change.

scripts/run_bazel.sh
maru-ava: fwiw the env guard was in response to complaints about the nix shell being bash-centric and not playing well with zsh (default macos shell).

Taskfile.yml
maru-ava: By that logic, though, no reason for `bazel-query` to exist. Deleted. Aside from things called by CI, I'll work on paring down task usage for bazel in preference for direct invocation.

scripts/generate_bazel_patches.sh
ARR4N: We can always revisit it later. I've definitely used `diff -uN` in the past, but I do recall there being a point in time at which it started causing headaches.

PR #4545, opened 2025-11-25
Provide alternative to `gossip.BloomFilter`

StephenButtolph:
## Why this should be merged

`gossip.BloomFilter` is fairly tough to use as it is only partially thread safe (and requires external synchronization primitives with the underlying mempool.

This provides a single implementation that abstracts the usage of a bloom filter away from the VM (unless the VM is doing something weird, like the X-chain).

## How this works

Implements a `gossip.SetWithBloomFilter` implementation that hides the bloom filter from the VM and uses it rather than using `gossip.BloomFilter`.

## How this was tested

Unit tests + existing e2e tests

No

### DIFF
--- a/network/ip_tracker.go
+++ b/network/ip_tracker.go
@@ -552,9 +552,10 @@ func (i *ipTracker) updateMostRecentTrackedIP(node *trackedNode, ip *ips.Claimed
 		return
 	}
 
-	i.bloomAdditions[ip.NodeID] = oldCount + 1
-	bloom.Add(i.bloom, ip.GossipID[:], i.bloomSalt)
-	i.bloomMetrics.Count.Inc()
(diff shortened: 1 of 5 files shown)


network/p2p/gossip/set.go
```diff
@@ -0,0 +1,159 @@
...
+	// that conflicting Gossipable items collide in the bloom filter, so a salt
+	// is generated to eventually resolve collisions.
+	salt ids.ID
+}
+
+func (s *SetWithBloomFilter[T]) Add(v T) error {
+	if err := s.set.Add(v); err != nil {
+		return err
+	}
+	return s.AddToBloom(v.GossipID())
```
StephenButtolph, line 77: TODO: Should this return the error? Or should this just log a warning. The value was added to the inner set and to the bloom filter... So it might make more sense to just log and move on.

PR #4547, opened 2025-11-25
Simplify gossip tests

StephenButtolph:
## Why this should be merged

1. There are test helpers that aren't used externally, we can move them to files that are only accessible for testing.
2. Prefixing everything with `test` isn't really needed if it is defined in a test file.
3. Some of these helpers weren't used anywhere. Like `NoOpGossiper` and `FullSet.Gossip`.
4. We can simplify the usage of the helpers by using more specific types rather than using structs with a single field.

## How this works

Just cleanup, no functional changes.

## How this was tested

CI

No.

### DIFF
--- a/network/p2p/gossip/gossip_test.go
+++ b/network/p2p/gossip/gossip_test.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/exp/maps"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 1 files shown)


network/p2p/gossip/gossip_test.go
```diff
@@ -49,13 +57,60 @@ func TestGossiperShutdown(t *testing.T) {
...
+	return tx(id), err
+}
+
+type setDouble struct {
+	txs   set.Set[tx]
+	bloom *BloomFilter
+	onAdd func(tx tx)
+}
+
+func (t *setDouble) Add(gossipable tx) error {
```
joshua-kim: nit: update the receiver name

maru-ava (approved):

StephenButtolph:
hold off until repo merger is done

PR #4619, opened 2025-11-26
Cleanup GossipEthTxPool

StephenButtolph:
## Why this should be merged

1. `IsSubscribed` is a testing only function that really indicates that the implementation is inherently racy. The subscription should be synchronous while the handling of the subscription should be async.
2. `SubscribeTransactions` currently takes in `false`. This isn't a bug, because the bool is ignored for the legacy pool... But if it weren't ignored, then this would violate the invariant that the bloom filter is a superset of the inner set. So now it is set to `true`.

## How this works

imo, the code would be cleaner if there was a `Close` or `Shutdown` function on the mempool with the goroutine started in `NewGossipEthTxPool`... But that seemed to be a bigger refactor involving node shutdown.

## How this was tested

existing CI

no

### DIFF
--- a/graft/coreth/plugin/evm/eth_gossiper.go
+++ b/graft/coreth/plugin/evm/eth_gossiper.go
@@ -60,7 +60,7 @@ func NewGossipEthTxPool(mempool *txpool.TxPool, registerer prometheus.Registerer
 
 	return &GossipEthTxPool{
 		mempool:    mempool,
-		pendingTxs: make(chan core.NewTxsEvent, pendingTxsBuffer),
+		pendingTxs: pendingTxs,
 		sub:        sub,
 		bloom:      bloom,
 	}, nil
--- a/graft/coreth/plugin/evm/eth_gossiper.go
+++ b/graft/coreth/plugin/evm/eth_gossiper.go
@@ -9,9 +9,9 @@ import (
 	"context"
 	"fmt"
 	"sync"
-	"sync/atomic"
 
 	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/event"
 	"github.com/ava-labs/libevm/log"
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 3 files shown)


graft/coreth/plugin/evm/eth_gossiper.go
```diff
@@ -33,8 +33,13 @@ var (
 	_ gossip.Set[*GossipEthTx]        = (*GossipEthTxPool)(nil)
 
 	_ eth.PushGossiper = (*EthPushGossiper)(nil)
+
+	errSubscribing = fmt.Errorf("subscribing to the mempool failed")
```
ceyonur:
```suggestion
	errSubscribing = errors.New("subscribing to the mempool failed")
```

graft/coreth/plugin/evm/eth_gossiper.go
```diff
@@ -47,48 +53,46 @@ func NewGossipEthTxPool(mempool *txpool.TxPool, registerer prometheus.Registerer
...
-	}
-	g.subscribed.CompareAndSwap(false, true)
-	defer func() {
-		sub.Unsubscribe()
-		g.subscribed.CompareAndSwap(true, false)
-	}()
+// UpdateBloomFilter continuously listens for new pending transactions from the
+// mempool and adds them to the bloom filter. If the bloom filter reaches its
+// capacity, it is reset and all pending transactions are re-added.
+func (g *GossipEthTxPool) UpdateBloomFilter(ctx context.Context) {
```
ceyonur, line 82: nit: maybe can be renamed`BloomFilterUpdateLoop` or `DispatchBloomUpdater` etc etc to indicate the continous loop.

graft/coreth/plugin/evm/eth_gossiper.go
```diff
@@ -47,48 +53,46 @@ func NewGossipEthTxPool(mempool *txpool.TxPool, registerer prometheus.Registerer
...
 
 	for {
 		select {
 		case <-ctx.Done():
 			log.Debug("shutting down subscription")
 			return
-		case pendingTxs := <-g.pendingTxs:
+		case pendingTxs, ok := <-g.pendingTxs:
+			if !ok {
+				log.Debug("pending txs channel closed, shutting down subscription")
```
ceyonur, line 92: nit: I don't think this channel closes normally, should this be a warn/error log?

graft/coreth/plugin/evm/eth_gossiper.go
ceyonur: or really `Subscribe` haha

ARR4N (approved):

StephenButtolph:
Not going to bother with merging. SAE will delete this anyways.

PR #4620, opened 2025-11-26
Skip incrementing bloom count after noop Add

StephenButtolph:
## Why this should be merged

Part of #4621

Currently, our p2p stack takes great care to avoid adding the same entry into a bloom filter multiple times. This is because we use the `Count()` function on the bloom filter to determine when bloom resets should occur to reduce the false positive probability.

However, this has resulted in delicate invariants for the mempool gossip, where the failure case can result in persistent gossip. Namely, if the bloom filter is not a superset of the mempool, transactions can be gossiped persistently through the network.

While this PR doesn't modify this invariant, it does allow for the gossip logic to soften the error case. Ideally, the p2p gossip should only gossip the tx once, even if the mempool locally already has the transaction.

## How this works

Only increments `count` in the bloom filter if `Has` would have previously returned `false`.

## How this was tested

Unit tests.

No.

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -2,30 +2,34 @@ name: 'C-Chain Re-Execution Benchmark'
 description: 'Run C-Chain re-execution benchmark'
 
 inputs:
-  runner_name:
-    description: 'The name of the runner to use and include in the Golang Benchmark name.'
-    required: true
+  task:
+    description: 'Task name to execute from Taskfile.yml. Leave empty to use custom inputs below.'
+    default: ''
+  # Custom inputs (alternative to task-based approach)
   config:
     description: 'The config to pass to the VM for the benchmark. See BenchmarkReexecuteRange for details.'
     default: ''
   start-block:
     description: 'The start block for the benchmark.'
-    default: '101'
+    default: ''
   end-block:
     description: 'The end block for the benchmark.'
-    default: '250000'
+    default: ''
   block-dir-src:
     description: 'The source block directory. Supports S3 directory/zip and local directories.'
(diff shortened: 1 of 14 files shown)


network/ip_tracker.go
```diff
@@ -552,9 +552,10 @@ func (i *ipTracker) updateMostRecentTrackedIP(node *trackedNode, ip *ips.Claimed
 		return
 	}
 
-	i.bloomAdditions[ip.NodeID] = oldCount + 1
-	bloom.Add(i.bloom, ip.GossipID[:], i.bloomSalt)
-	i.bloomMetrics.Count.Inc()
+	if bloom.Add(i.bloom, ip.GossipID[:], i.bloomSalt) {
+		i.bloomAdditions[ip.NodeID] = oldCount + 1
```
StephenButtolph, line 556: Technically not needed, but it seemed reasonable only to increment the count if the bloom also changed - since these bloomAdditions are just to rate limit bloom resets.

utils/bloom/filter.go
```diff
@@ -59,19 +59,29 @@ func New(numHashes, numEntries int) (*Filter, error) {
...
-func (f *Filter) Add(hash uint64) {
+// Add adds the provided hash to the bloom filter. It returns true if the hash
+// was not already present in the bloom filter.
+func (f *Filter) Add(hash uint64) bool {
 	f.lock.Lock()
 	defer f.lock.Unlock()
 
-	_ = 1 % f.numBits // hint to the compiler that numBits is not 0
+	var (
+		_                = 1 % f.numBits // hint to the compiler that numBits is not 0
```
samliok, line 69: why is this hint helpful?

StephenButtolph: It moves the division by zero check out of the loop. (Which is a very minor performance improvement).

PR #4623, opened 2025-11-26
feat(statesync): introduce Finalizer interface for syncer cleanup

### LINKED ISSUE #4603: Implement a cleanup/finalize procedure preventing corruption of synced state due to sync process errors
Based on the `SyncerRegistry` implementation [here](https://github.com/ava-labs/coreth/pull/1083), we need to add a cleanup phase that would be triggered on startup or execution errors in order to eliminate committing incomplete state - something similar to this: https://github.com/ava-labs/hypersdk/blob/main/statesync/client.go#L108

powerslider:
## Why this should be merged

Check https://github.com/ava-labs/avalanchego/issues/4603

## How this works
Add a Finalizer interface to provide explicit cleanup operations for syncers. This ensures cleanup (like flushing batches to disk) is performed reliably even on cancellation or early returns.

- Add `Finalizer` interface to sync/types.go for explicit cleanup.
- Attach `Finalize()` in `CodeQueue` that finalizes code fetching to this new interface.
- Gather finalization logic in a `Finalize()` for StateSyncer to flush in-progress trie batches.
- Implement `Finalize()` for AtomicSyncer to commit pending database changes.
- Add `FinalizeAll()` to `SyncerRegistry` with defer to ensure cleanup runs.
- Remove `OnFailure` callback mechanism (replaced by `Finalizer`).

## How this was tested
existing UT

## Need to be documented?
no

no

resolves #4603

Signed-off-by: Tsvetan Dimitrov (tsvetan.dimitrov23@gmail.com)


graft/coreth/plugin/evm/atomic/sync/syncer.go
```diff
@@ -146,6 +146,16 @@ func (s *Syncer) Sync(ctx context.Context) error {
...
 
+// Finalize commits any pending database changes to disk.
+// This ensures that even if the sync is cancelled or fails, we preserve
+// the progress up to the last fully synced height.
+func (s *Syncer) Finalize() error {
+	if s.db == nil {
+		return nil
+	}
+	return s.db.Commit()
+}
```
ceyonur, line 154: Why is this committing the db? Isn't `onFinish` already doing the same thing? I think this interface is a bit confusing with `onFinish` already here.

graft/coreth/sync/statesync/state_syncer.go
```diff
@@ -301,19 +300,19 @@ func (t *stateSync) removeTrieInProgress(root common.Hash) (int, error) {
 	return len(t.triesInProgress), nil
 }
 
-// onSyncFailure is called if the sync fails, this writes all
-// batches of in-progress trie segments to disk to have maximum
-// progress to restore.
-func (t *stateSync) onSyncFailure() {
+// Finalize checks if there are any in-progress tries and flushes their batches to disk
+// to preserve progress. This is called by the syncer registry on sync failure or cancellation.
```
ceyonur: isn't this also being called in success?

graft/coreth/sync/statesync/state_syncer.go
```diff
@@ -301,19 +300,19 @@ func (t *stateSync) removeTrieInProgress(root common.Hash) (int, error) {
...
+func (t *stateSync) Finalize() error {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
 	for _, trie := range t.triesInProgress {
 		for _, segment := range trie.segments {
```
ceyonur, line 322: (could be an over-cautious comment) 
this looks like should've been cleared on success, but if not then we might end-up in a weird spot.

graft/coreth/plugin/evm/atomic/sync/syncer.go
powerslider: So based on my understanding `onFinish()` and `Finalize()` serve very different goals:

#### `onFinish()`

- Critical Completion Handler for the sync process.
- called by `CallbackLeafSyncer` when a leaf sync task successfully finishes (all leaves fetched and processed).
- its responsibilities include:
   - Commit the final trie state.
   - Insert the trie into the atomic trie database.
   - Accept the trie at the target height.
   - Commit the database.
   - Validate that the synced root matches the expected target root.
- Errors propagates up -> `syncTask()` -> `Sync()` → caller. Basically the sync process fails if this fails.
- The root validation is critical - without it, we could accept a corrupted or incomplete sync.

#### `Finalize()`
- Best-Effort Cleanup Handler
- called `SyncerRegistry.FinalizeAll()` via `defer` - always called after `Sync()` returns (success or failure).
- its responsibilities include:
   - Commit any uncommitted database changes.
- Errors are logged but not returned. Does not affect sync success/failure status.
- Preserves partial progress when sync fails mid-way.

In the success case, `Finalize()` does a redundant `db.Commit()` since `onFinish()` already committed. `Finalize()` only provides unique value when sync fails after processing some leaves that haven't been committed yet (between commit intervals in `onLeafs`).

graft/coreth/sync/statesync/state_syncer.go
powerslider: Yes. The reasons why this is ok and part of the design direction with this centralized best-effort `Finalize()` are the following:
- `Finalize()` in the state syncer flushes any uncommitted segment batches from in-progress tries to disk.
- we are shifiting the current design of dedicated on success and on failure handlers, because we don't want the syncers to self manage their own lifecycles. We want to delegate that to the `SyncerRegistry`.
- `Finalize()` is always called after `Sync()` returns (via `defer` in registry).
- on success: `triesInProgress` is empty -> no-op (tries already removed after completion).
- on failure: `triesInProgress` contains incomplete work -> flushes batches to disk
- we have a single code path, guaranteed cleanup via `defer`.

powerslider: Added an early return for the success path and more clarifications in the method docs.

graft/coreth/sync/statesync/state_syncer.go
powerslider: check my previous reply https://github.com/ava-labs/avalanchego/pull/4623#discussion_r2577001958

graft/coreth/plugin/evm/atomic/sync/syncer.go
powerslider: NOTE: Given that `db.Commit()` on an already-committed database is essentially a no-op (very cheap), the added complexity of tracking when the syncing completes may not be worth it. But if you want explicit clarity about what's happening I can add a `completed` flag that could be set in `onFinish` and then checked this way:
```
func (s *Syncer) Finalize() error {
    if s.completed {
        return nil  // Already committed in onFinish.
    }
    return s.db.Commit()
}
```

graft/coreth/sync/client/leaf_syncer.go
```diff
@@ -159,9 +158,9 @@ func (c *CallbackLeafSyncer) Sync(ctx context.Context) error {
...
 
-	err := eg.Wait()
-	if err != nil {
-		c.config.OnFailure()
+	if err := eg.Wait(); err != nil {
+		return err
 	}
-	return err
+
+	return nil
```
alarso16:
```suggestion
	return eg.Wait()
```

alarso16 (approved):

ceyonur (approved):

ceyonur (approved):

Issue #4569, opened 2025-09-30
Delete `customrawdb` package after migration to `avalanchego`


powerslider:
## Summary 

Delete `plugin/evm/customrawdb` package after confirmed successful migration to `vms/evm/sync` in `avalanchego`.

PR in `avalanchego`: https://github.com/ava-labs/avalanchego/pull/4387

PR #4626, opened 2025-11-27
chore(customrawdb): delete customrawdb package and switch to avalanchego imports

powerslider:
## Why this should be merged
Check https://github.com/ava-labs/avalanchego/issues/4569

## How this works
- Delete `graft/coreth/plugin/evm/customrawdb`.
- Switch dependent code to using `github.com/ava-labs/avalanchego/vms/evm/sync/customrawdb` import instead.
- Fix build and test failures.

## How this was tested
existing UT

## Need to be documented?
no

no

resolves https://github.com/ava-labs/avalanchego/issues/4569

Signed-off-by: Tsvetan Dimitrov ([tsvetan.dimitrov23@gmail.com](mailto:tsvetan.dimitrov23@gmail.com))


graft/coreth/core/state/snapshot/journal.go
```diff
@@ -42,6 +42,13 @@ import (
 	"github.com/ava-labs/libevm/triedb"
 )
 
+var (
```
powerslider: If adding these sentinel errors here adds noise to the PR, I will revert them.

graft/coreth/plugin/evm/customtypes/hashing_test.go
ceyonur, line 1: do we require this change for this PR?

graft/coreth/core/state/snapshot/journal.go
ceyonur: I'd say any cosmetic/style changes here should be reverted as this is a modified-upstream code, and we should not further modify it.

ceyonur (commented):
just some cleanup from this PR. otherwise lgtm

graft/coreth/plugin/evm/customtypes/hashing_test.go
powerslider: no, it was just a fix proposed by Jonathan, but it is not needed here

graft/coreth/core/state/snapshot/journal.go
```diff
@@ -59,20 +59,16 @@ type journalGenerator struct {
 // store. If loading the snapshot from disk is successful, this function also
 // returns a boolean indicating whether or not the snapshot is fully generated.
 func loadSnapshot(diskdb ethdb.KeyValueStore, triedb *triedb.Database, cache int, blockHash, root common.Hash, noBuild bool) (snapshot, bool, error) {
-	// Retrieve the block number and hash of the snapshot, failing if no snapshot
-	// is present in the database (or crashed mid-update).
-	baseBlockHash := customrawdb.ReadSnapshotBlockHash(diskdb)
-	if baseBlockHash == (common.Hash{}) {
```
ceyonur, line 65: should we still check this?

graft/coreth/core/state/snapshot/journal.go
```diff
@@ -59,20 +59,16 @@ type journalGenerator struct {
...
+	// Retrieve the block hash of the snapshot, failing if no snapshot is present.
+	baseBlockHash, err := customrawdb.ReadSnapshotBlockHash(diskdb)
+	if err != nil {
 		return nil, false, errors.New("missing or corrupted snapshot, no snapshot block hash")
 	}
-	if baseBlockHash != blockHash {
-		return nil, false, fmt.Errorf("block hash stored on disk (%#x) does not match last accepted (%#x)", baseBlockHash, blockHash)
-	}
```
ceyonur, line 70: why have we removed this?

graft/coreth/core/state/snapshot/journal.go
powerslider: Since `rawdb.ReadSnapshotRoot` doesn't return an error for missing values, the empty hash check is the only way to detect that the snapshot root is missing. This is the standard go-ethereum pattern for these `rawdb` accessor functions.

In our case in the `customrawdb` package we do return an error and we can utilize that to check against instead of the empty hash check.

One thing I upgraded is wrapping the returned error, so that we can have more context https://github.com/ava-labs/avalanchego/pull/4626/commits/7207b185cf7250f472cd3e4f44b6fe8f3b6f2db8.

graft/coreth/core/state/snapshot/wipe_test.go
```diff
@@ -96,8 +102,14 @@ func TestWipe(t *testing.T) {
...
-		t.Errorf("snapshot block hash marker remained after wipe: %#x", hash)
+	// Verify snapshot markers are removed.
+	blockHash, err = customrawdb.ReadSnapshotBlockHash(db)
+	switch {
+	case errors.Is(err, database.ErrNotFound): // Expected: marker was deleted.
+	case err != nil:
+		t.Errorf("unexpected error reading snapshot block hash after wipe: %v", err)
+	case blockHash != (common.Hash{}):
+		t.Errorf("snapshot block hash marker remained after wipe: %#x", blockHash)
 	}
```
alarso16, line 113: This doesn't check the blockhash if the error is `database.ErrNotFound`. Is that expected? Just looks like a bug at first glance

alarso16: In either case, this definitely seems like it's a symptom of racy code. Will you add a comment describing why the error may or may not be nil?

alarso16: Ahh, I now understand this. resolving

PR #4630, opened 2025-12-01
Add Cancun Solidity Compatibility Notes 

ceyonur:
## Why this should be merged

Avalanche EVMS do not currently support Pectra, so we should add this compatibility note to avoid compiling contract for pectra.

## How this works

Adds to Readme Compatibility section and `cancun` version to solc commands

## How this was tested

CI

No


JonathanOppenheimer:
This PR has been synced to subnet-evm in https://github.com/ava-labs/subnet-evm/pull/1880

Issue #4621, opened 2025-11-26
Standardize Mempool Gossip


StephenButtolph:
**Context and scope**

Currently, each VM must manage their own bloom filter manually to use the `gossip` package's pull logic.

There should be a standard implementation that fully manages the bloom filter without requiring management by each VM.

StephenButtolph:
This was merged with #4819

PR #4632, opened 2025-12-01
Implement gossip.SetWithBloomFilter

StephenButtolph:
## Why this should be merged

Part of #4621

This allows VM implementations an alternative to the `gossip.BloomFilter` implementation. By using this implementation, users do not need to manage a bloom filter on their own.

## How this works

Provides a generic implementation similar to the implementations in the X-chain and the P-chain.

To speed up development, this PR does not include migrating the existing VMs to this new implementation.

## How this was tested

Updated the integration test of the gossip package to utilize the simplified API.

No.

### DIFF
--- a/network/p2p/gossip/set.go
+++ b/network/p2p/gossip/set.go
@@ -5,6 +5,7 @@ package gossip
 
 import (
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"sync"
 
@@ -22,7 +23,7 @@ var (
 	_ Set[Gossipable]             = (*BloomSet[Gossipable])(nil)
 	_ PullGossiperSet[Gossipable] = (*BloomSet[Gossipable])(nil)
 
-	ErrBloomReset = fmt.Errorf("bloom reset")
+	ErrBloomReset = errors.New("bloom reset")
 )
 
 type BloomSetConfig struct {


network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,130 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer, line 1: I saw you added the locks, but I don't think these are tested anywhere. 

Something like `func TestSetWithBloomFilter_Concurrent`?

network/p2p/gossip/bloom.go
```diff
@@ -18,6 +18,8 @@ import (
 //
 // Invariant: The returned bloom filter is not safe to reset concurrently with
 // other operations. However, it is otherwise safe to access concurrently.
+//
+// Deprecated: [SetWithBloomFilter] should be used to manage bloom filters.
```
JonathanOppenheimer: What's the idea behind leaving around the deprecated functions? out of scope for this PR?

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,162 @@
...
+	l        sync.RWMutex
+	maxCount int
+	bloom    *bloom.Filter
+	// salt is provided to eventually unblock collisions in Bloom. It's possible
+	// that conflicting Gossipable items collide in the bloom filter, so a salt
+	// is generated to eventually resolve collisions.
+	salt ids.ID
+}
+
+func (s *SetWithBloomFilter[T]) Add(v T) error {
```
JonathanOppenheimer: I see the addToBloom guarantee on the error, but is there another guarantee - if adding to the set fails, it can't be in the bloom filter? (as it won't be added) 

We can say 

"If adding to the bloom filter fails after successfully adding to the set,  the value remains in the set and is still added to the bloom filter" 

right

network/p2p/gossip/bloom.go
StephenButtolph: They are used in all of our mempool implementations... I spent a long time looking into updating the coreth mempool... At the end of the day, I determined that separating it out into separate PRs made the most sense to unblock SAE.

network/p2p/gossip/set.go
StephenButtolph: sure

JonathanOppenheimer: oh lol I was more looking to confirm my understanding was correct, not to add the comment

JonathanOppenheimer: it doesn't have to be a comment

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+// NewSetWithBloomFilter wraps set with a bloom filter. It is expected for all
+// future additions to the provided set to go through the returned set, or be
+// followed by a call to AddToBloom.
+func NewSetWithBloomFilter[T Gossipable](
+	set Set[T],
+	registerer prometheus.Registerer,
+	namespace string,
+	minTargetElements int,
+	targetFalsePositiveProbability,
+	resetFalsePositiveProbability float64,
```
joshua-kim: Related to our offline conversation but since this is a new api, would it be reasonable to break the symmetry with the old `NewBloomFilter` api and make this accept a config struct here + define a default in this package?

go.mod
```diff
@@ -21,9 +21,9 @@ require (
 	github.com/DataDog/zstd v1.5.2
 	github.com/StephenButtolph/canoto v0.17.3
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/avalanchego/graft/coreth v0.16.0-rc.0
+	github.com/ava-labs/avalanchego/graft/coreth v0.0.0-20251201173339-98b2978e465a
```
StephenButtolph, line 24: I had to update this to allow subnet-evm to import avalanchego + coreth post-graft

network/p2p/gossip/gossip.go
```diff
@@ -206,17 +219,27 @@ func NewPullGossiper[T Gossipable](
 	}
 }
 
+// PullGossiperSet exposes the current bloom filter and allows adding new items
+// that were not included in the filter.
+type PullGossiperSet[T Gossipable] interface {
+	// Add adds a value to the set. Returns an error if v was not added.
+	Add(v T) error
```
joshua-kim, line 229: nit: I noticed we use this throughout the PR but did we intend to use `v` here? Should this be `t`?

network/p2p/gossip/gossip_test.go
```diff
@@ -68,39 +67,56 @@ func (marshaller) UnmarshalGossip(bytes []byte) (tx, error) {
 }
 
 type setDouble struct {
+	l     sync.RWMutex
```
joshua-kim: nit: it's used in a small scope/struct so I think `l` can work, but isn't `mu` (Go-ey) or `lock` (used more in our codebase) what we would expect here?

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+type SetWithBloomFilter[T Gossipable] struct {
+	set Set[T]
+
+	minTargetElements              int
+	targetFalsePositiveProbability float64
+	resetFalsePositiveProbability  float64
+
+	metrics *bloom.Metrics
+
+	l        sync.RWMutex
```
joshua-kim: Same nit w.r.t naming

network/p2p/gossip/bloom.go
```diff
@@ -72,12 +75,8 @@ func (b *BloomFilter) Has(gossipable Gossipable) bool {
 	return bloom.Contains(b.bloom, h[:], b.salt[:])
 }
 
-func (b *BloomFilter) Marshal() ([]byte, []byte) {
-	bloomBytes := b.bloom.Marshal()
-	// salt must be copied here to ensure the bytes aren't overwritten if salt
-	// is later modified.
-	salt := b.salt
-	return bloomBytes, salt[:]
+func (b *BloomFilter) BloomFilter() (*bloom.Filter, ids.ID) {
```
ARR4N, line 78: For my understanding: why `ids.ID`? It seems like an odd overloading of the meaning of a type.

network/p2p/gossip/gossip.go
```diff
@@ -336,10 +359,16 @@ func NewPushGossiper[T Gossipable](
 	}, nil
 }
 
+// PushGossiperSet exposes whether hashes are still included in a set.
+type PushGossiperSet interface {
```
ARR4N, line 369: (subjective) I understand (and agree with) the abstraction of the different set types, but their names feel unintuitive to me. Although they're named for how they're used, I think it would make more sense for them to be named for what they do. This, for example, could be a `MembershipVerber` (perhaps `Reporter` or `Checker`).

If you plan on coming back to this code later then I think it's better to just add a TODO so we don't block SAE on this.

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+	_ PullGossiperSet[Gossipable] = (*SetWithBloomFilter[Gossipable])(nil)
+)
+
+type Set[T Gossipable] interface {
+	HandlerSet[T]
+	PushGossiperSet
+	// Len returns the number of items in the set.
+	//
+	// This value should always be at least as large as the number of items that
+	// can be iterated over with a call to Iterate.
```
ARR4N: Please elaborate on scenarios in which one might expect `Iterate` to yield fewer items.

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+	Len() int
+}
+
+// NewSetWithBloomFilter wraps set with a bloom filter. It is expected for all
+// future additions to the provided set to go through the returned set, or be
+// followed by a call to AddToBloom.
```
ARR4N:
```suggestion
// NewSetWithBloomFilter wraps the [Set] with a bloom filter. It is expected for all
// future additions to the provided set to go through the returned [SetWithBloomFilter], or be
// followed by a call to [SetWithBloomFilter.AddToBloom].
```

Just to disambiguate.

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+	// salt is provided to eventually unblock collisions in Bloom. It's possible
+	// that conflicting Gossipable items collide in the bloom filter, so a salt
+	// is generated to eventually resolve collisions.
+	salt ids.ID
+}
+
+// Add adds v to the set and bloom filter.
+//
+// If adding to the bloom filter fails after successfully adding to the set, the
+// value remains in the set and is still added to the bloom filter.
```
ARR4N: > value remains in the set and is still added to the bloom filter.

Remains in the set but is _not_ added to the bloom filter?

In these circumstances, is a consumer expected to retry anything? If so, we should probably differentiate between the two error paths with wrapped sentinel errors.

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+// If an error is returned, the bloom filter and salt are unchanged.
+func (s *SetWithBloomFilter[T]) resetBloomFilter() error {
+	targetElements := max(2*s.set.Len(), s.minTargetElements)
+	numHashes, numEntries := bloom.OptimalParameters(
+		targetElements,
+		s.targetFalsePositiveProbability,
+	)
+	newBloom, err := bloom.New(numHashes, numEntries)
+	if err != nil {
+		return fmt.Errorf("creating new bloom: %w", err)
```
ARR4N: Once I've convinced you about `assert` and `require` I'll move on to not using `%w` by default, but I need to prioritise my battles ;)

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (s *SetWithBloomFilter[T]) shouldReset() bool {
+	return s.bloom.Count() > s.maxCount
+}
+
+// resetBloomFilter attempts to generate a new bloom filter and fill it with the
+// current entries in the set.
+//
+// If an error is returned, the bloom filter and salt are unchanged.
+func (s *SetWithBloomFilter[T]) resetBloomFilter() error {
```
ARR4N: Please comment that `shouldReset` requires at least a read lock while `resetBloomFilter` requires holding a write lock.

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+	type (
+		op struct {
+			add    *tx
+			remove *tx
+		}
+		test struct {
+			name                          string
+			resetFalsePositiveProbability float64
+			ops                           []op
+			expected                      []tx
```
ARR4N:
```suggestion
			expectedInFilter                      []tx
```

I was really confused when I saw `add(0,1,2)` then `remove(0,1,2)` then `expect(0,1,2)` because I didn't appreciate that it wasn't the set itself.

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+			expected: []tx{
+				{2},
+			},
+			expectedResetCount: 2,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require := require.New(t)
```
ARR4N: I'm trying to discourage use of this pattern because of the number of times I've seen the result being used  in sub-tests and therefore with a closure over the wrong `T`.

network/p2p/gossip/gossip_test.go
```diff
@@ -593,7 +608,9 @@ func TestPushGossiper(t *testing.T) {
 
 			gossiper, err := NewPushGossiper[tx](
 				marshaller,
-				fullSet[tx]{},
+				hasFunc(func(ids.ID) bool {
+					return true // Never remove the items from the set
+				}),
```
ARR4N, line 599: I like this; it's much clearer.

graft/coreth/plugin/evm/tx_gossip_test.go
```diff
@@ -107,31 +107,20 @@ func TestEthTxGossip(t *testing.T) {
 	}
 
 	// Ask the VM for any new transactions. We should get nothing at first.
-	emptyBloomFilter, err := gossip.NewBloomFilter(
-		prometheus.NewRegistry(),
-		"",
-		config.TxGossipBloomMinTargetElements,
-		config.TxGossipBloomTargetFalsePositiveRate,
-		config.TxGossipBloomResetFalsePositiveRate,
+	requestBytes, err := gossip.MarshalAppRequest(
```
ARR4N, line 110: (out of scope) Just noting that these changes are identical to the `atomic` ones, which suggests there may be insufficient abstraction.

network/p2p/gossip/bloom_test.go
```diff
@@ -81,19 +81,21 @@ func TestBloomFilterRefresh(t *testing.T) {
...
 					resetCount++
 				}
-				bloom.Add(item)
+				require.Equal(initialBloomBytes, bloomFilter.Marshal())
+				require.Equal(initialSaltBytes, salt[:])
 
-				require.Equal(initialBloomBytes, bloomBytes)
-				require.Equal(initialSaltBytes, saltBytes)
+				// If the bloom filter wasn't reset, adding an item may modify
+				// the returned bloom filter.
```
ARR4N:
```suggestion
				// the returned bloom filter, so this must be done after the checks.
```

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+			var s setDouble
+			bs, err := NewSetWithBloomFilter(
+				&s,
+				prometheus.NewRegistry(),
+				"",
+				minTargetElements,
+				targetFalsePositiveProbability,
+				tt.resetFalsePositiveProbability,
+			)
+			require.NoError(err)
```
ARR4N:
```suggestion
			require.NoError(err, "NewSetWithBloomFilter()")
```

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+			for _, op := range tt.ops {
+				if op.add != nil {
+					require.NoError(bs.Add(*op.add))
```
ARR4N:
```suggestion
					require.NoErrorf(bs.Add(*op.add), "%T.Add(...)", bs)
```

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+				if op.remove != nil {
+					s.txs.Remove(*op.remove)
+				}
+			}
+
+			// Add one to expectedResetCount to account for the initial creation
+			// of the bloom filter.
+			require.Equal(tt.expectedResetCount+1, testutil.ToFloat64(bs.metrics.ResetCount))
```
ARR4N:
```suggestion
			require.Equal(tt.expectedResetCount+1, testutil.ToFloat64(bs.metrics.ResetCount), "number of resets")
```

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+			b, h := bs.BloomFilter()
+			for _, expected := range tt.expected {
+				require.True(bloom.Contains(b, expected[:], h[:]))
```
ARR4N:
```suggestion
				require.Truef(bloom.Contains(b, expected[:], h[:]), "%T.Contains(%#x)", bloom, expected)
```

Also, `assert` is more appropriate for testing set membership, to provide a full picture and avoid whack-a-mole debugging. Just saying 😉

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		})
+	}
+}
+
+// TestSetWithBloomFilter_Concurrent tests that SetWithBloomFilter is ensures
```
ARR4N:
```suggestion
// TestSetWithBloomFilter_Concurrent tests that SetWithBloomFilter ensures
```

network/p2p/gossip/set_test.go
```diff
@@ -0,0 +1,178 @@
...
+// that the returned bloom filter is a super set of the items in the Set at the
+// time it is called, even under concurrent resets.
```
ARR4N:
```suggestion
// time it is called, even under concurrent resets, because resets of the filter are
// accompanied by refilling from the Set.
```

ARR4N (approved):
No need for me to re-review the requested changes.

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,167 @@
...
+		minTargetElements:              minTargetElements,
+		targetFalsePositiveProbability: targetFalsePositiveProbability,
+		resetFalsePositiveProbability:  resetFalsePositiveProbability,
+
+		metrics: metrics,
+	}
+	return m, m.resetBloomFilter()
+}
+
+type SetWithBloomFilter[T Gossipable] struct {
```
joshua-kim: Name is a bit long - maybe `BloomSet`?

network/p2p/gossip/set_test.go
joshua-kim: I'm curious on the rationale for these logs in `require` - isn't this generally self documenting?

network/p2p/gossip/bloom.go
joshua-kim: I don't think there was a reason in the old code - I think we just went with hash = `ids.ID`. It would probably make more sense to make a type alias for salt or just use plain `[]byte`

network/p2p/gossip/gossip.go
StephenButtolph: I felt like `v` for value was pretty normal

network/p2p/gossip/bloom.go
StephenButtolph: I returned an array rather than a slice to avoid needing to do the copy.. It could just be `[32]byte`... We just typically use `ids.ID` throughout the repo for these... We could probably clean it up later

network/p2p/gossip/gossip.go
StephenButtolph: I added TODOs

network/p2p/gossip/set.go
StephenButtolph: I reworded it to say it should be equal... Really it just _must_ not be smaller.

network/p2p/gossip/set.go
StephenButtolph: I unexported `AddToBloom`, so I changed some of this wording.

network/p2p/gossip/set.go
StephenButtolph: I added an exported error to check for the case that the set addition succeeded but the bloom reset failed.

graft/coreth/plugin/evm/tx_gossip_test.go
StephenButtolph: Yeah there is probably a nice test helper that should be made at some point. I really didn't want to refactor all of the testing here... Just wanted to change the tests as minimally as possible to work with the new interface

network/p2p/gossip/set_test.go
StephenButtolph: I also don't find the logs useful, but adding them doesn't matter to me

network/p2p/gossip/set_test.go
StephenButtolph: I know people talk about whack-a-mole debugging... I have never experienced it myself

network/p2p/gossip/set.go
```diff
@@ -0,0 +1,191 @@
...
+	"sync"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/bloom"
+)
+
+const (
+	DefaultMinTargetElements              = 1000
+	DefaultTargetFalsePositiveProbability = .01
+	DefaultResetFalsePositiveProbability  = .05
```
joshua-kim, line 19: Should we expose these under a single `DefaultBloomSetConfig`? It seems like it would make discoverability easier

joshua-kim (approved):

network/p2p/gossip/set.go
StephenButtolph: I did it this way so that the defaults can't be changed (As we can't expose a `const DefaultBloomSetConfig`).

PR #4635, opened 2025-12-01
fix: compatibility links

JonathanOppenheimer:
## Why this should be merged
The links are broken

## How this works
I fixed the links

## How this was tested
I clicked the links

No


JonathanOppenheimer:
> Seems like you lost some additional links?

those docs don't exist anymore

Issue #4508, opened 2025-11-14
Epic: Multiple metrics in reexecution test


RodrigoVillar:
At one point, we had multiple metrics in the reexecution tests (#4369) but we had to remove this as this resulted in our github-action-benchmark being corrupted (#4396). We should strive for our metrics to be as granular as possible and so this epic is catered towards that.

RodrigoVillar:
For adding Grafana visualizations, I noticed that in our HyperSDK dashboard, we have MeterVM metrics there (https://github.com/ava-labs/avalanche-monitoring/blob/main/grafana/dashboards/morpheusvm/performance.json). It may make sense to rip those metrics and place them into the C-Chain dashboard.

RodrigoVillar:
Closing this now that #4673 is merged in.

PR #4640, opened 2025-12-02
chore(reexecute/c): remove go bench from benchmark

RodrigoVillar:
## Why this should be merged

As part of #4508, this PR removes usage of `go bench` from the reexecution test. This offers the following benefits:

- Although the reexecution test is housed within a go benchmark, we only run it once, which is not idiomatic as go benchmarks are meant to be iterated over multiple times.
- This sets up the foundation to reintroduce multiple metrics (e.g. block acceptance time, block verification time, etc.) to our dashboard as adding additional metrics is as simple as calling `addResult()` on the new benchmark tool.

## How this works

- Removes usage of `*testing.B`
- Output of the reexecution statistics file is handled by the test itself rather than `./scripts/benchmark_cchain_range.sh`
- Updates GAB (github-action-benchmark) action to use custom benchmarks rather than `go bench`

## How this was tested

CI + ran `task reexecute-cchain-range START_BLOCK=1 END_BLOCK=100_000 BENCHMARK_OUTPUT_FILE=<INSERT_CUSTOM_PATH_HERE>` locally and got the following:

```json
[
  {
    "name": "BenchmarkReexecuteRange/[1,100000]-Config-default-Runner-dev - mgas/s",
    "value": "254.90228109633716",
    "unit": "mgas/s"
  }
]
```

Furthermore, results are logged to the terminal as follows:

```
[12-09|13:06:35.049] INFO c-chain-reexecution c/vm_reexecute.go:637 BenchmarkReexecuteRange/[1,100000]-Config-default-Runner-dev {"result": "262.43886323313035 mgas/s"}
```

To test that benchmark results display correctly on the GAB dashboard, I created a sample benchmark (https://github.com/RodrigoVillar/sample-reexecution-custom-benchmark/blob/main/main.go) which is a copy-and-paste of the benchmarking code in this PR and plotted some points into my own dashboard. The results of the sample benchmark can be seen below:

Screenshot 2025-12-04 at 09 35 04

Attached is a link to the sample benchmark dashboard: https://rodrigovillar.github.io/sample-reexecution-custom-benchmark/dev/bench/

No

## TODOs after merging

- [x] Prune `gh-pages` branch of data points which use `go bench`

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -20,7 +20,6 @@ set -euo pipefail
 : "${END_BLOCK:?END_BLOCK must be set}"
 
 go run github.com/ava-labs/avalanchego/tests/reexecute/c \
-  --benchmark-output-file="${BENCHMARK_OUTPUT_FILE}" \
   --block-dir="${BLOCK_DIR}" \
   --current-state-dir="${CURRENT_STATE_DIR}" \
   ${RUNNER_NAME:+--runner="${RUNNER_NAME}"} \
@@ -28,6 +27,7 @@ go run github.com/ava-labs/avalanchego/tests/reexecute/c \
   --start-block="${START_BLOCK}" \
   --end-block="${END_BLOCK}" \
   ${LABELS:+--labels="${LABELS}"} \
+  ${BENCHMARK_OUTPUT_FILE:+--benchmark-output-file="${BENCHMARK_OUTPUT_FILE}"} \
   ${METRICS_SERVER_ENABLED:+--metrics-server-enabled="${METRICS_SERVER_ENABLED}"} \
   ${METRICS_SERVER_PORT:+--metrics-server-port="${METRICS_SERVER_PORT}"} \
   ${METRICS_COLLECTOR_ENABLED:+--metrics-collector-enabled="${METRICS_COLLECTOR_ENABLED}"}


tests/reexecute/c/vm_reexecute.go
```diff
@@ -561,14 +574,70 @@ func startCollector(tb testing.TB, log logging.Logger, name string, labels map[s
...
+// When results are added, the unit is appended to this base name.
+func newBenchmarkTool(name string) *benchmarkTool {
+	return &benchmarkTool{
+		name:    name,
+		results: make([]benchmarkResult, 0),
+	}
+}
+
+// addResult adds a new benchmark result with the given value and unit.
+// The result name is constructed by appending the unit to the benchmark name
```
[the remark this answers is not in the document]
RodrigoVillar: Done: 3588978

scripts/benchmark_cchain_range.sh
```diff
@@ -19,7 +19,8 @@ set -euo pipefail
 : "${START_BLOCK:?START_BLOCK must be set}"
 : "${END_BLOCK:?END_BLOCK must be set}"
 
-cmd="go test -timeout=0 -v -benchtime=1x -bench=BenchmarkReexecuteRange -run=^$ github.com/ava-labs/avalanchego/tests/reexecute/c \
+cmd="go run github.com/ava-labs/avalanchego/tests/reexecute/c \
+  --benchmark-output-file=\"${BENCHMARK_OUTPUT_FILE}\" \
```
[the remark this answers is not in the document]
RodrigoVillar: We follow this pattern for all other strings in this script.

scripts/benchmark_cchain_range.sh
```diff
@@ -31,8 +32,4 @@ cmd="go test -timeout=0 -v -benchtime=1x -bench=BenchmarkReexecuteRange -run=^$
   ${METRICS_SERVER_PORT:+--metrics-server-port=\"${METRICS_SERVER_PORT}\"} \
   ${METRICS_COLLECTOR_ENABLED:+--metrics-collector-enabled=\"${METRICS_COLLECTOR_ENABLED}\"}"
 
-if [ -n "${BENCHMARK_OUTPUT_FILE:-}" ]; then
-  eval "$cmd" | tee "${BENCHMARK_OUTPUT_FILE}"
-else
-  eval "$cmd"
-fi
+eval "$cmd"
```
aaronbuchwald: Should we remove this and just run the command now?

scripts/benchmark_cchain_range.sh
aaronbuchwald: How should this be handled if the benchmark output file is not defined?

scripts/benchmark_cchain_range.sh
RodrigoVillar: Yup, removed `eval` here: 955273e

scripts/benchmark_cchain_range.sh
RodrigoVillar: In `vm_reexecute.go`, we don't write to a file if the benchmark output filepath is empty, but you're right that if a user currently doesn't set `BENCHMARK_OUTPUT_FILE`, that we'll have:

```bash
go run github.com/ava-labs/avalanchego/tests/reexecute/c --benchmark-output-file= --current-state-dir=...`
```

Changed it so that if a user doesn't define `BENCHMARK_OUTPUT_FILE`, then `--benchmark-output-file` isn't set: f682b20

aaronbuchwald (approved):

Issue #4582, opened 2025-09-23
Add support for dynamic state syncing to the sync client


powerslider:
### Goal 
   - Extend the current static state sync to a dynamic model where sync targets can change during synchronization and block execution continues after finalization with a controlled queue.

### Key ideas
  - Start all syncers concurrently and independently.
  - Allow `UpdateSyncTarget` to advance the sync target while syncers are running.
  - Disallow further target updates once all syncers reach a consistent target.
  - Finalize VM state (blockchain, atomic, shared memory, indices) before any post-finalization block execution.
  - Buffer blocks that arrive during dynamic sync in a `BlockQueue` and execute them after finalization, with batching and race-safe cut-offs.
  - Drive behavior with a clear state machine to handle edge cases and consistency

### Non-goals
- No changes to consensus rules or block validation semantics.
- No redesign of underlying syncers’ low-level protocols - we adapt them via an interface for dynamic target updates.

### Full Reference Documentation
https://www.notion.so/Dynamic-State-Sync-Documentation-2b110c29114e811490fcfcbf184f7f2d

PR #4646, opened 2025-12-03
feat(vmsync): dynamic state sync with coordinator, pivot cadence, and engine-driven target updates

powerslider:
## Why this should be merged

Static state sync blocks until completion, causing the node to fall further behind as new blocks arrive. Dynamic state sync addresses this by:

- Queueing block operations during sync for deferred execution.
- Updating sync targets as new blocks are accepted.
- Processing queued blocks after sync completes.

Check #4582

## How this works

Check the full reference documentation [here](https://www.notion.so/Dynamic-State-Sync-Documentation-2b110c29114e811490fcfcbf184f7f2d).

## How this was tested

new unit tests (will add e2e later)

for now no

### DIFF
--- a/graft/coreth/sync/session/session.go
+++ b/graft/coreth/sync/session/session.go
@@ -1,0 +1,181 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package session
+
+import (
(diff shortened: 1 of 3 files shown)


powerslider:
Closed in favor of https://github.com/ava-labs/avalanchego/pull/5051

PR #4507, opened 2025-11-14
Simplify testing with arbitrary versions of firewood

maru-ava:
## Why this should be merged

This change adds support for using an experimental tool called `polyrepo` to clone firewood and coreth and integrate them with the local avalanchego clone so that avalanchego tasks can be run against a specific combination of the 3 repos.

To test-drive:
 - Clone this branch
 - Install nix
 - Use polyrepo to sync and integrate firewood for testing
   - ./bin/polyrepo sync [optional: firewood@revision, defaults to the vendored version)
     - Clones firewood and configures its use with the local avalanchego clone
 - Ensure the shell is configured for AWS access (required for S3 downloads)
   - Okta -> AWS Access Portal -> Experimental -> Access Keys
     - Note: these credentials will expire after 1 hour
   - export AWS_ACCESS_KEY_ID=<AWS_ACCESS_KEY_ID>
   - export AWS_SECRET_ACCESS_KEY=<AWS_SECRET_ACCESS_KEY>
   - export AWS_SESSION_TOKEN=<AWS_SESSION_TOKEN>
   - export AWS_REGION=us-east-2
 - Run a reexecution test
   - ./scripts/run_task.sh c-chain-reexecution-firewood-101-250k

## How this was tested

CI

## Need to be documented in RELEASES.md?

N/A

PR #4650, opened 2025-12-03
[ci] Support running re-execution benchmark with arbitrary version of Firewood

Elvis339:
## Why this should be merged

Enables Firewood to track performance over time by running C-Chain reexecution benchmarks with custom Firewood builds. This establishes the infrastructure for catching performance regressions before they reach production.

https://github.com/ava-labs/firewood/issues/1494

## How this works

1. Firewood triggers the existing C-Chain reexecution workflow via GitHub API with `with-dependencies` parameter
2. Workflow sets up dependencies via before calling the benchmark action
3. The task uses [polyrepo](https://github.com/ava-labs/avalanchego/tree/maru/polyrepo/tests/fixture/polyrepo/core) to sync Firewood at the specified ref
4. Runs C-Chain reexecution benchmark with the custom build
5. Uploads results as artifact for Firewood to download and track separately

The root action remains unaware of external dependency configuration - setup is handled by the workflow before invocation.

The same functionality is available locally for development:

```
nix develop
FIREWOOD_REF=abc123 ./scripts/run_task.sh run-polyrepo
./scripts/run_task.sh test-cchain-reexecution -- firewood-101-250k
```

## Changes

- Update workflows to call setup task before invoking benchmark action
- Root action no longer accepts dependency refs (workflow responsibility)
- Always upload benchmark artifacts (enables downstream tracking)
- Update documentation in [C-Chain Re-Execution README](tests/reexecute/c/README.md)

## How this was tested

gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --ref es/enable-firewood-dev-workflow \
  -f test=firewood-101-250k \
  -f with-dependencies="firewood=v0.0.15,libevm=v1.13.15-0.20251210210615-b8e76562a300" \
  -f runner=avalanche-avalanchego-runner-2ti \
  -f timeout-minutes=60

No

### DIFF
--- a/vms/README.md
+++ b/vms/README.md
@@ -4,7 +4,7 @@
 
 The Avalanche Network is composed of multiple validator sets and blockchains. A validator set defines a group of validators and their specified weights in the consensus process. A subnet is a validator set working together to achieve consensus on a set of blockchains. Every blockchain is validated by one subnet, and one subnet can validate many blockchains.
 
-There is a special subnet inherent to the Avalanche Network called the Primary Network. The Primary Network is validated by every node on the Avalanche network. All subnets' validator sets are required to be a subset of the Primary Network's validator set. That is, if a validator belongs to a subnet then it also belongs to the Primary Network. The Primary Network validates three blockchains that are inherent to the Avalanche Network: the P-Chain, C-Chain, and X-Chain.
+There is a special subnet inherent to the Avalanche Network called the Primary Network. The Primary Network validates three blockchains that are inherent to the Avalanche Network: the P-Chain, C-Chain, and X-Chain.
 
 For each blockchain, consensus is driven by the consensus engine. For each subnet, the P-Chain, or Platform Chain, defines the validator set and the set of blockchains that are validated by the subnet.
 


.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -82,8 +82,6 @@ jobs:
       id-token: write
       contents: write
     runs-on: ${{ matrix.runner }}
-    container:
```
Elvis339, line 85: With `ghcr.io/actions/actions-runnner:2.325.0` container directive the Firewood build fails due to `io-uring` permission error:

```
[11-27|11:07:45.868] CRIT <2q9e4r6Mu3U68nU1fYjgbR6JvwrRx36CohpAX5UQxse55x1Q5 Chain> triedb/firewood/database.go:91 firewood: error creating database        error="File IO error: Operation not permitted (os error 1) at offset 0 of file '/tmp/reexecution-data-20251127-110654/current-state/chain-data-dir/firewood/firewood.db' IO-uring setup"
```

graft/coreth/scripts/build_firewood.sh
```diff
@@ -0,0 +1,97 @@
...
+print_usage() {
+  printf "Usage: build_firewood.sh [FIREWOOD_COMMIT] [--dirty]
+
+  Build Firewood FFI from source for testing with Coreth.
+
+  Arguments:
+    FIREWOOD_COMMIT   Git commit or branch to build (default: main)
+
+  Options:
+    --dirty           Update go.mod after building (dirties working tree)
```
RodrigoVillar: Is there a case when we wouldn't want to set `--dirty`?

Elvis339: Maybe I just didn't want default to dirty the working tree.

RodrigoVillar: If a user calls `./build_firewood.sh X` (where `X = FIREWOOD_COMMIT`), then this will build Firewood but won't replace the `firewood-go-ethhash` dependency in `coreth` with the newly built Firewood. In other words, this build will have no utility as `coreth` will be using it's own `firewood-go-ethhash` version. 

Unless I'm missing context, we should remove the `--dirty` flag and make the `go mod edit -replace` step mandatory.

Elvis339: I understand your positio this was actually my initial thinking too. However, I received similar feedback ~3 months ago that changed my mind: https://github.com/ava-labs/avalanchego/pull/4341#discussion_r2397033062

The concern is that automatically dirtying the working tree creates risk of accidentally committing the go.mod change. By defaulting to "just build and print the path", the user explicitly opts-in to modifying their working tree with `--dirty`.
For CI, we always pass `--dirty` since it's ephemeral. For local dev, the user can either:
1. Run with --dirty and remember to revert
2. Capture the path and manually run go mod edit -replace

Taskfile.yml
```diff
@@ -328,6 +328,14 @@ tasks:
           CURRENT_STATE_DIR_SRC: 'cchain-current-state-firewood-100'
           CONFIG: firewood
 
+  build-firewood:
+    desc: Build Firewood FFI from source for testing with Coreth
+    dir: graft/coreth
+    vars:
+      FIREWOOD_REF: '{{.FIREWOOD_REF | default "main"}}'
```
maru-ava: Please avoid defining standard configuration for shell scripts as vars. The exception could be for scripts that have different defaults between CI and non-CI usage (e.g. fuzzing), and that is not the case here. 

In that regard, a majority of the reexec tasks need a refactor.

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
```diff
@@ -0,0 +1,139 @@
...
+
+name: Firewood Reexecution Benchmark
+
+on:
+  workflow_dispatch:
+    inputs:
+      firewood:
+        description: 'Firewood commit/branch/tag to build and test'
+        required: true
+      libevm:
```
maru-ava: What's the use case for varying libevm for this kind of testing?

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
```diff
@@ -0,0 +1,139 @@
...
+jobs:
+  determine-build-type:
+    runs-on: ubuntu-latest
+    outputs:
+      from-source: ${{ steps.check.outputs.from-source }}
+    steps:
+      - name: Check Firewood input
+        id: check
+        run: |
+          # Semantic versions (vX.Y.Z) are published to Go module registry
```
maru-ava: (No action required) Why is it desirable to key on the semantic version rather than attempting to get the desired revision for firewood-go-ethhash and if that fails, build with nix?

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: Install Nix
+        if: needs.determine-build-type.outputs.from-source == 'true'
+        uses: cachix/install-nix-action@02a151ada4993995686f9ed4f1be7cfbb229e56f # v31
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update libevm
+        if: inputs.libevm != ''
+        shell: bash
+        run: |
+          echo "Updating libevm to: ${{ inputs.libevm }}"
```
maru-ava: I thought I'd been clear of the need to do as little work as possible in CI to ensure local reproducibility. As written, the only way this logic can be tested is by cycling through CI.

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
```diff
@@ -0,0 +1,139 @@
...
+          echo "Fetching published Firewood: ${{ inputs.firewood }}"
+          go get github.com/ava-labs/firewood-go-ethhash/ffi@${{ inputs.firewood }}
+      - name: Build Firewood (from source)
+        if: needs.determine-build-type.outputs.from-source == 'true'
+        shell: bash
+        run: ./scripts/run_task.sh build-firewood FIREWOOD_REF="${{ inputs.firewood }}" DIRTY=true
+        env:
+          TMPDIR: /tmp
+          NIX_BUILD_TOP: /tmp
+      # Commit go.mod changes locally to keep working tree clean
```
maru-ava: Maybe avoid having to commit by setting `gh-repository` for github-action-benchmark so that it can use a separate clone rather than a branch switch?

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
Elvis339: The suggestion is actually more robust, implemented it: https://github.com/ava-labs/avalanchego/pull/4650/commits/6f21231e66016515be3863f07cc99dc8871360d6

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
Elvis339: The workflow is a temporary bridge until we move Firewood benchmarks over to the Polyrepo tooling you’ve been building. The plan is to hand orchestration, not grow it into a long‑term interface, which is why I opted to keep the “version vs source” detection and optional libevm update directly in the YAML instead of adding another Task/script layer that would also be short‑lived. The actual work (go get and ./scripts/run_task.sh build-firewood) is already runnable locally today. 

Could you explain what we’d gain by pushing this glue into a Task or script first? If there’s a specific local flow or testing pattern you might want to preserve for this transition period, I’m happy to adjust, but I just don’t see a clear advantage that outweighs the extra surface area for something we intend to tear down soon.

Taskfile.yml
Elvis339: Right, makes sense: https://github.com/ava-labs/avalanchego/pull/4650/commits/e4376589182faf6dcfdfe7b24298bd3e1a4e73c8

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
Elvis339: The Firewood team typically specifies a particular libevm version. Since they're the primary consumers of this workflow, I exposed it as an optional input for their use.

[Source](https://platwizards.slack.com/archives/C075H83CMSQ/p1764269385567189?thread_ts=1764243336.505679&cid=C075H83CMSQ)

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
Elvis339: github-action-benchmark only does branch switching when auto-push: true. Since push-github-action-benchmark: false I might not even need [the commit](https://github.com/ava-labs/avalanchego/pull/4650/commits/15b52ee41792a614c5b2d2c8e554091c44974711) step, triggering benchmark to verify this works. 

Result: https://github.com/ava-labs/avalanchego/actions/runs/20070059594/job/57569292678

.github/workflows/c-chain-reexecution-benchmark-firewood.yml
maru-ava: While the near-term goal is to migrate firewood to the monorepo, libevm will remain forever independent since it is a fork of go-ethereum and keeping it as close as possible to upstream will simplify fork maintenance. That means the logic of this particular step will be long-lived.

But the goal of making CI locally reproducible isn't just to simplify long term maintenance. It's also to simplify initial implementation. What you've defined here can only be iterated on in CI - every change, trigger a CI job to test it. For things that can _only_ be done in CI, that's fine, but everything else should be externalized to simplify both implementation and review. Your reviewers benefit from being able to run things locally too.

Maybe create a task that accepts a version of libevm and firewood and ensures the working directory is capable of testing with them? And then, rather than creating a whole new workflow, update an existing reexec workflow to accept a firewood revision and if provided conditionally install nix and run the new task before calling the custom reexec action?

Elvis339: You're right that local reproducibility simplifies both implementation and review I don't disagree with the principle. The question is whether it's worth the investment here.

It seems like we're not aligned on the end state. The plan for [Stage 2](https://github.com/ava-labs/firewood/issues/1494) is that Polyrepo handles dependency stitching from Firewood's CI, and Firewood owns the full benchmark flow with [PerfProbe](https://github.com/ava-labs/perf-probe) for statistical analysis. When that's in place, this AvalancheGo workflow gets deleted entirely—libevm handling included.

**Firewood CI → Polyrepo orchestrates dependencies → Run reexecution in Firewood** (not [this workflow](https://github.com/ava-labs/avalanchego/pull/4650/files#diff-dc50679ead35a388845336eaab00d61134127f673c49db5d3100ac1225fc938d))

That's why I've been treating this as disposable scaffolding rather than investing in local reproducibility. Having said that, how do you see libevm fitting into Stage 2 with Polyrepo? Is there something about its independence that would require AvalancheGo to keep owning this logic?

maru-ava: polyrepo was a pre-monorepo strategy. 4 weeks ago I posted [examples](https://github.com/ava-labs/firewood/pull/1454) of [its use](https://github.com/ava-labs/avalanchego/pull/4507). Now, I'm not clear what the value would be of going further than enabling firewood to trigger the reexec job as you've proposed. Moving firewood to a monorepo where avalanchego depends on it directly is the more likely outcome.

Elvis339: Talked offline decided to use Polyrepo in stage 1.

tests/reexecute/c/README.md
```diff
@@ -247,6 +247,42 @@ To execute a benchmark with any of these options, you must use a compatible `CUR
...
+## Testing with Custom Dependency Versions
+
+The benchmarks support testing with custom versions of `libevm` and `firewood` dependencies. This is useful for:
+- Testing unreleased versions before merging
+- Benchmarking performance changes in dependencies
+- Reproducing issues with specific dependency versions
+
+### Local Usage
+
+**Prerequisite**: You must be in a nix shell (`nix develop`) for firewood builds.
```
maru-ava: Given that polyrepo invokes the ffi nix shell for the firewood build, what other step requires a nix shell?

maru-ava:
Functionally this looks ok. Doc question and lint failure will need to be addressed before merge. Also, in addition to the command that you ran to verify that this is working, please include a passing test run that resulted from that command (this is general best practice when manual testing is required).

tests/reexecute/c/README.md
Elvis339: Running `./scripts/run_task.sh c-chain-reexecution-*` requires nix shell because it's missing `s5cmd`. 

For example this task https://github.com/ava-labs/avalanchego/blob/fec393b84dc0baa8d3e335523eabbd91168936fb/Taskfile.yml#L295 invokes https://github.com/ava-labs/avalanchego/blob/fec393b84dc0baa8d3e335523eabbd91168936fb/Taskfile.yml#L234 which invokes https://github.com/ava-labs/avalanchego/blob/fec393b84dc0baa8d3e335523eabbd91168936fb/Taskfile.yml#L145 which invokes 
https://github.com/ava-labs/avalanchego/blob/master/scripts/copy_dir.sh and requires `s5cmd` available in nix shell.

maru-ava: To my point, none of that is firewood-specific. Maybe update this text for accuracy?

maru-ava:
I guess the 10 input params is an indication that this remote execution can't be much more than a stop-gap pending firewood migration to a monorepo?

tests/reexecute/c/README.md
Elvis339: Right, good catch. https://github.com/ava-labs/avalanchego/pull/4650/commits/f79c8d5ebe5350ac1994feb6b425857fe0115d46

Elvis339:
> I guess the 10 input params is an indication that this remote execution can't be much more than a stop-gap pending firewood migration to a monorepo?

I wasn't aware of the 10 input param limit first time hitting this constraint.
As I see it, we have two options:
1. Keep as stop-gap - Accept the current limitations until Firewood is grafted into the monorepo
 2. Move orchestration to Firewood remove this from avalanchego and have polyrepo orchestrate the benchmarks from Firewood's side (as we discussed offline)

That said, both options are effectively stop-gaps. (1) is temporary until Firewood joins the monorepo, and (2) would still require cross-repo coordination that becomes unnecessary once Firewood is grafted.

If you're okay with keeping (1) as a stop-gap for now with a GH issue to track the long-term solution, that would unblock the Firewood team to start collecting performance metrics sooner.

maru-ava:
> If you're okay with keeping (1) as a stop-gap for now with a GH issue to track the long-term solution, that would unblock the Firewood team to start collecting performance metrics sooner.

I'm fine with that, as per my use of the term 'stop-gap' - a temporary solution pending a better one.

Elvis339:
Executed:
```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --ref es/enable-firewood-dev-workflow \
  -f task=c-chain-reexecution-firewood-101-250k \
  -f firewood-ref=v0.0.15 \
  -f runner=avalanche-avalanchego-runner-2ti \
  -f timeout-minutes=60
```
  
  Result: https://github.com/ava-labs/avalanchego/actions/runs/20240259592

.github/workflows/c-chain-reexecution-benchmark-container.yml
maru-ava: Why are we using this image in the first place? I get that it wasn't working for firewood, but it would be good to know why it's ok to stop using it.

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -25,12 +25,15 @@ on:
       runner:
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
         required: true
-      push-post-state:
```
maru-ava, line 28: I see that the job definitions have been updated to use a matrix field instead of this import parameter, what are the implications of this change?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -103,13 +104,23 @@ jobs:
...
           push-github-action-benchmark: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master') }}
-          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-role: ${{ matrix.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          push-post-state: ${{ github.event.inputs.push-post-state }}
+          push-post-state: ${{ matrix.push-post-state || '' }}
           runner_name: ${{ matrix.runner }}
+      - name: Upload benchmark results
+        if: inputs.firewood != ''
```
maru-ava: Maybe add a ~command~ comment here indicating why a value for this parameter should prevent uploading?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -25,12 +25,15 @@ on:
       runner:
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
         required: true
-      push-post-state:
-        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      firewood:
```
maru-ava: Maybe use more descriptive names (e.g. firewood-ref)? I think 'firewood' or 'libevm' by themselves are ambiguous.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: The main implication is that `push-post-state` can't be set from manual workflow_dispatch anymore you'd have to add it to the JSON config file instead.

Scheduled runs still work fine since they read from the JSON anyway.

We had to make this trade-off because GitHub has a 10 input limit for workflow_dispatch, and we needed room for `firewood-ref` and `libevm-ref`. Figured those are more useful for manual runs (testing custom versions), while `push-post-state` is typically used in scheduled/automated runs anyway.

If you ever need to manually push state after a run, you can still do it with:
```bash
task export-dir-to-s3 SRC=/path/to/current-state DST=s3://bucket/destination
```

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Added it here: https://github.com/ava-labs/avalanchego/blob/a89c44e1006ea0672662bedb065462d0802a67a4/.github/actions/c-chain-reexecution-benchmark/action.yml#L196 but forgot to add it to concrete implementations, thanks for pointing it out. 

https://github.com/ava-labs/avalanchego/pull/4650/commits/d42815596382538b5f4588f4839524c415b49eb5 https://github.com/ava-labs/avalanchego/pull/4650/commits/87b3d4c459c1ab041a07cc769e717b97fe08a972

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4650/changes/d42815596382538b5f4588f4839524c415b49eb5

Elvis339:
Re: https://github.com/ava-labs/avalanchego/pull/4650#discussion_r2620210083

> Why are we using this image in the first place? I get that it wasn't working for firewood, but it would be good to know why it's ok to stop using it.

The image was originally used because it was minimal and we didn't want to maintain our own ubuntu base was missing sudo and the initial ARC implementation needed some baseline. SREs and Infra have been iterating on it since then.
As for why it's ok to remove - I ran into io_uring not being available when trying to build Firewood (which uses io_uring for async I/O). Removed the container and it worked. The self-hosted runners already have all the dependencies we need, so the container was just adding constraints without much benefit for our use case.

.github/workflows/c-chain-reexecution-benchmark-container.yml
maru-ava: @aaronbuchwald Thoughts?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -103,13 +104,25 @@ jobs:
...
           push-github-action-benchmark: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master') }}
-          aws-role: ${{ github.event.inputs.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
+          aws-role: ${{ matrix.push-post-state != '' && secrets.AWS_S3_RW_ROLE || secrets.AWS_S3_READ_ONLY_ROLE }}
           aws-region: 'us-east-2'
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          push-post-state: ${{ github.event.inputs.push-post-state }}
+          push-post-state: ${{ matrix.push-post-state || '' }}
           runner_name: ${{ matrix.runner }}
+      - name: Upload benchmark results
+        # Skip when using custom firewood - Firewood downloads the artifact and
```
maru-ava: Maybe ensure the comment clearly matches the conditional? e.g. `Only upload when not using a custom version of firewood` is likely a better match for `inputs.firewood-ref != ''`

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -25,12 +25,15 @@ on:
...
         description: 'Runner to execute the benchmark. Input to the runs-on field of the job.'
         required: true
-      push-post-state:
-        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      firewood-ref:
+        description: 'Firewood commit/branch/tag to build and test (optional). If provided, builds custom firewood and uploads benchmark artifact.'
```
maru-ava: Maybe include mention in this description that providing a firewood ref will preclude comparing results?

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: https://github.com/ava-labs/avalanchego/pull/4650/changes/64935989af1a07a66597fe07d9cd9693f4c74194

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Thank you for pointing it out.

https://github.com/ava-labs/avalanchego/pull/4650/changes/64935989af1a07a66597fe07d9cd9693f4c74194

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: I haven't used this much recently, but would really prefer to keep it easily accessible as I've already done the work to make it so we can easily push new DBs from CI rather than spinning up new instances to do the same thing.

It seems preferable to me to separate concerns here, so that the re-execution custom action does not need to know anything about what it takes to create an AvalancheGo directory with custom local dependencies correctly.

Would it make sense to remove the separate repo refs and instead move that to the prior step in a workflow file?

ie.

1. checkout avalanchego
2. optionally check out and update local dependencies so that building/running AvalancheGo works correctly
3. run the re-execution action within the avalanchego working directory (`go build` and `go run` should just work)

scripts/run_polyrepo.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+POLYREPO_REVISION=6239973c9b
+echo "Running polyrepo@${POLYREPO_REVISION} via go run..."
+go run github.com/ava-labs/avalanchego/tests/fixture/polyrepo@"${POLYREPO_REVISION}" "${@}"
```
aaronbuchwald: nit: add a newline here to satisfy the linting gods

aaronbuchwald:
I see why the workflows need to support the libevm, coreth, etc. refs to pull them in and make them configurable, but is it necessary for the re-execution benchmark itself to know about them?

The re-execution benchmarks currently only assume that it can run `go build` and `go run` within the AvalancheGo directory, so it seems unnecessary to me to integrate the two.

If there were instead a separate script/task or similar that took in the dependencies to update, we could keep these concerns separate.

maru-ava:
> I see why the workflows need to support the libevm, coreth, etc. refs to pull them in and make them configurable, but is it necessary for the re-execution benchmark itself to know about them?

The goal is to have a single discoverable entrypoint for any given task that runs in CI, so while it is entirely possible to configure a reexec test separately, that's the task that then should be executed by CI.

Note that these changes are not likely to be long-lived assuming firewood ends up being migrated to the avalanchego repo.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: > Would it make sense to remove the separate repo refs and instead move that to the prior step in a workflow file?
ie.
checkout avalanchego
optionally check out and update local dependencies so that building/running AvalancheGo works correctly
run the re-execution action within the avalanchego working directory (go build and go run should just work)


As Maru pointed out below the design ensures CI and local have the same invocation pattern `task reexecute-cchain-range-with-copied-data FIREWOOD_REF=abc123` works same in both environments. Moving dependency setup to workflow steps would break this and make it harder to reproduce CI runs locally. 

> I haven't used this much recently, but would really prefer to keep it easily accessible as I've already done the work to make it so we can easily push new DBs from CI rather than spinning up new instances to do the same thing.

GitHub's 10-input limit forced me to remove push-post-state to prioritize libevm-ref and firewood-ref for manual testing.
For CI-triggered push without spinning up a local instance, a couple options:
1. Separate workflow - Create a lightweight workflow just for S3 export that calls task export-dir-to-s3
2. JSON config - Add push-post-state to the schedule JSON config so it runs automatically on scheduled benchmarks

Would either of these work, or could you propose an alternative?

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: I believe this was originally added when we first started supporting runs with ARC in conjunction with installing ARC dependencies. I agree w/ Maru we should know why this was added (as well as ARC dependency installation) and see why it's now ok to remove either or both.

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: I'd strongly prefer to keep this. Could we combine the firewood ref and libevm ref into a single comma separated value that declares what dependencies must be overridden and to what?

Elvis339:
> I see why the workflows need to support the libevm, coreth, etc. refs to pull them in and make them configurable, but is it necessary for the re-execution benchmark itself to know about them?
> 
> The re-execution benchmarks currently only assume that it can run `go build` and `go run` within the AvalancheGo directory, so it seems unnecessary to me to integrate the two.
> 
> If there were instead a separate script/task or similar that took in the dependencies to update, we could keep these concerns separate.

The benchmark itself (reexecute-cchain-range task) has zero knowledge of libevm/firewood - it just runs go run against whatever's in the directory.

The dependency setup is a separate step that runs before the benchmark, composed in a parent task:

1. if LIBEVM_REF → go get libevm@ref
2. if FIREWOOD_REF → polyrepo sync firewood@ref
. reexecute-cchain-range ← benchmark runs here, knows nothing about steps 1-2

So the concerns are separated - just composed together in one task entry point.

```
 cmds:
   - cmd: |  # Step 1: Optional libevm setup
       if [[ -n "{{.LIBEVM_REF}}" ]]; then
         go get github.com/ava-labs/libevm@{{.LIBEVM_REF}}
       fi
   - cmd: |  # Step 2: Optional firewood setup
       if [[ -n "{{.FIREWOOD_REF}}" ]]; then
         ./scripts/run_polyrepo.sh sync firewood@{{.FIREWOOD_REF}}
       fi
   - task: import-cchain-reexecute-range  # Step 3: Import data
   - task: reexecute-cchain-range         # Step 4: Run benchmark
```

scripts/run_polyrepo.sh
Elvis339: Done, 🙏 to the linting gods.

Elvis339:
After: https://github.com/ava-labs/avalanchego/pull/4650/commits/1416e52dc659c9258f1451aec9d854692a535bff

Executed:
```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --ref es/enable-firewood-dev-workflow \
  -f task=c-chain-reexecution-firewood-101-250k \
  -f with="firewood=v0.0.15,libevm=v1.13.15-0.20251210210615-b8e76562a300" \
  -f runner=avalanche-avalanchego-runner-2ti \
  -f timeout-minutes=60
 ```
 
 Results: 
 - `define-matrix` [job](https://github.com/ava-labs/avalanchego/actions/runs/20391111752/job/58600407807#step:3:39) is parsing `with: firewood=xyz,libevm=xyz` correctly.
    - It's passed to child [task](https://github.com/ava-labs/avalanchego/actions/runs/20391111752/job/58600410364#step:4:532)
    
Note: the failed workflow is due to change in FFI API. (0.0.15 vs 0.0.18)

File: .github/workflows/c-chain-reexecution-benchmark-container.yml (hunk 9 above)
Comment (Elvis339) [reply .github/workflows/c-chain-reexecution-benchmark-container.yml:28]: Alright, I've implemented it as you suggested. The workflow now uses a single comma-separated with input that declares which dependencies to override and their values:
- Single dependency: with=firewood=abc123
- Multiple dependencies: with="firewood=abc123,libevm=v1.2.3"

The workflow parses this input and extracts the firewood-ref and libevm-ref values, which are then passed to the action. This keeps the interface simple and flexible.

This action serves as temporary glue until we have a proper monorepo structure that can handle dependency versioning more natively.

Commit: https://github.com/ava-labs/avalanchego/pull/4650/commits/1416e52dc659c9258f1451aec9d854692a535bff
Test: https://github.com/ava-labs/avalanchego/pull/4650#issuecomment-3677519038

File: .github/actions/c-chain-reexecution-benchmark/action.yml [hunk 15]
```diff
@@ -180,7 +190,13 @@ runs:
         GH_RUN_NUMBER: ${{ inputs.run_number }}
         GH_RUN_ATTEMPT: ${{ inputs.run_attempt }}
         GH_JOB_ID: ${{ inputs.job }}
+        # Required for Nix builds in CI self-hosted runners (e.g., building Firewood via polyrepo)
+        TMPDIR: /tmp
+        NIX_BUILD_TOP: /tmp
+    # Skip when using custom firewood - Firewood downloads the artifact and
+    # publishes to their own gh-pages for tracking
     - name: Compare Benchmark Results
+      if: inputs.firewood-ref == ''
```
aaronbuchwald: Why does Firewood need to override this to write to a different location?

I don't see a reason Firewood needs to do this, but if it decides to there are other ways to control its own benchmark data without a special exemption here.

If it wants to control where it writes its own benchmark data, then it can simply use the existing task definitions or the custom action within its own repo to write to its desired branch.

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -25,12 +25,15 @@ on:
...
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      with:
+        description: 'Dependencies to use (e.g., "firewood=abc123" or "firewood=abc,libevm=xyz")'
+        default: ''
+      push-post-state:
+        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
+        default: ''
```
aaronbuchwald: Can we move this back to its original position to reduce the diff?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -25,12 +25,15 @@ on:
...
         required: true
-      push-post-state:
-        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      with:
+        description: 'Dependencies to use (e.g., "firewood=abc123" or "firewood=abc,libevm=xyz")'
+        default: ''
```
aaronbuchwald: Could we rename this to make it explicit ie. `with-dependencies` or `with-dependency-overrides`

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -113,3 +121,13 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           push-post-state: ${{ github.event.inputs.push-post-state }}
           runner_type: ${{ matrix.runner }}
+      - name: Upload benchmark results
```
aaronbuchwald: What's the rationale to write benchmark data to a separate location? It seems like this should not be the responsibility of AvalancheGo.

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
```diff
@@ -25,12 +25,15 @@ on:
...
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      with:
+        description: 'Dependencies to use (e.g., "firewood=abc123" or "firewood=abc,libevm=xyz")'
+        default: ''
+      push-post-state:
+        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
+        default: ''
```
aaronbuchwald: Can we reduce the diff here by moving push-post-state back to its original location?

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
```diff
@@ -25,12 +25,15 @@ on:
...
         required: true
-      push-post-state:
-        description: 'S3 location to push post-execution state directory. Skips this step if left unpopulated.'
-        default: ''
       timeout-minutes:
         description: 'Timeout in minutes for the job.'
         default: 30
+      with:
+        description: 'Dependencies to use (e.g., "firewood=abc123" or "firewood=abc,libevm=xyz")'
+        default: ''
```
aaronbuchwald: Same comment here, could we use an explicit input name?

aaronbuchwald (changes requested):

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Firewood needs to track its own performance regressions, but can't run C-Chain reexecution benchmarks (AvalancheGo setup, AWS credentials that live here).

The artifact upload isn't writing benchmark data to a separate location it's providing a neutral handoff point.

Why this is necessary:
1. Firewood cannot run C-Chain reexecution benchmarks (requires AvalancheGo setup and AWS credentials)
    - edit: With Polyrepo Firewood could run reexecution there but, the goal is to migrate to monorepo and this action wouldn't cause much diff then.
2. AvalancheGo shouldn't track Firewood's performance metrics (that's Firewood's domain)

The artifact is just raw JSON output. Firewood downloads it and feeds it to its own github-action-benchmark on its own gh-pages. AvalancheGo doesn't store, track, or visualize Firewood's data.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: When firewood-ref is set, we're benchmarking a specific Firewood commit. If we ran github-action-benchmark here, it would push to AvalancheGo's gh-pages and compare against AvalancheGo's baseline mixing Firewood data into AvalancheGo's tracking.

This whole setup mimics monorepo behavior since AvalancheGo is where dependencies converge (Coreth, libevm, Firewood).

The rationale is to skip comparison here (Firewood handles its own tracking) vs. polluting AvalancheGo's benchmark history with Firewood-specific runs that can't be correlated to Firewood commits.

Elvis339: Or are you suggesting generic approach where instead of checking firewood-ref, we add a generic input like skip-benchmark-comparison: false that any caller can set. Then Firewood's trigger can pass skip-benchmark-comparison: true without AvalancheGo needing to know about Firewood specifically?

I don't understand this:
> If it wants to control where it writes its own benchmark data, then it can simply use the existing task definitions or the custom action within its own repo to write to its desired branch.

Which task definitions are you referring to? This action produces the benchmark output - the artifact upload just makes that output available for download. Could you clarify what alternative you have in mind?

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4650/commits/73d405729499160905bb913d05fc60672cee6034

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
Elvis339: https://github.com/ava-labs/avalanchego/pull/4650/commits/73d405729499160905bb913d05fc60672cee6034

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4650/commits/4c4d3eb70aab1d6c3faa300306b6710d750480c4

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4650/commits/4c4d3eb70aab1d6c3faa300306b6710d750480c4

Taskfile.yml
```diff
@@ -273,7 +278,7 @@ tasks:
       - task: import-cchain-reexecute-range
         vars:
           BLOCK_DIR_SRC: '{{.S3_BOOTSTRAP_BUCKET}}/{{.BLOCK_DIR_SRC}}/**'
-          CURRENT_STATE_DIR_SRC: '{{if .CURRENT_STATE_DIR_SRC}}{{.S3_BOOTSTRAP_BUCKET}}/{{.CURRENT_STATE_DIR_SRC}}/**{{else}}genesis{{end}}'
+          CURRENT_STATE_DIR_SRC: '{{if .CURRENT_STATE_DIR_SRC}}{{.S3_BOOTSTRAP_BUCKET}}/{{.CURRENT_STATE_DIR_SRC}}/**{{else}}{{.GENESIS_SENTINEL}}{{end}}'
```
Elvis339: Using `"genesis"` instead of empty string because empty string would try to download from an invalid S3 path unless handled in downstream script. Also "genesis" actually tells you what's happening while empty string is ambiguous (is it null? an error?).

Defining it as `GENESIS_SENTINEL` variable makes it grepable a single source of truth, easy to find all usages.

This logic lives in the Taskfile because that's where we orchestrate tasks. The import script just imports from whatever source you give it. It doesn't need to know about optional parameters or special cases. Keeps scripts simple, keeps orchestration visible in one place.

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: @aaronbuchwald, just wanted to confirm I noticed you approved #4761 which removes the push-post-state to S3 functionality. You mentioned earlier you'd strongly prefer to keep it. Are you good with this removal, or should we add it back?

https://github.com/ava-labs/avalanchego/blob/master/.github/actions/c-chain-reexecution-benchmark/action.yml

maru-ava: Not removed, moved to the script: https://github.com/ava-labs/avalanchego/pull/4761/changes#diff-9768d8f505409dc874c5c244c11ff0428f0dd2e5e22d01be33527e9bb90a03c3R179

maru-ava: The input is still required to configure it: https://github.com/ava-labs/avalanchego/blob/master/.github/actions/c-chain-reexecution-benchmark/action.yml#L140

.github/workflows/c-chain-reexecution-benchmark-container.yml
aaronbuchwald: 1. Firewood can use the same AWS credentials or set up its own.
2. AvalancheGo uses Firewood and this is in fact a benchmark of AvalancheGo using Firewood, so it is relevant to include within AvalancheGo.

If Firewood wants more control, AvalancheGo already provides the custom action that Firewood can use from its own repo. This seems like a much more reasonable place to draw the line and allow Firewood to use rather than have AvalancheGo push data for Firewood to ingest.

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: Going to merge this conversation into one comment on the current changes rather than continuing across two threads.

Taskfile.yml
Elvis339: Removed from this PR, will open new.

Elvis339:
> I do not think that the re-execution benchmark action itself should be responsible for coupling firewood/libevm refs when any workflow invoking this can simply update the dependencies before invoking this action as a step. Separately, the action should not support a special case of archiving data just for Firewood.
> 
> These are simple pre- and post- execution steps, which can be easily added as a step before or after the action gets invoked within a workflow.
> 
> If Firewood wants to execute this separately, why not just define its own workflow to do so? If it wants to execute it within AvalancheGo, then we should just write the data to the default location using Firewood as the config.

Resolved offline. Root action now:
- Has no knowledge of firewood/libevm refs (workflow handles setup via task)
- Pushes to GitHub Action Benchmark (controlled by input)
- Always uploads artifacts

Dependency setup externalized to `task setup-reexecution-deps`, called by workflow before action.

Taskfile.yml
```diff
@@ -178,6 +178,23 @@ tasks:
     desc: Sets up C-Chain test data (blocks from S3, state from S3 or empty for genesis).
     cmd: ./scripts/setup_cchain_data.sh
 
+  setup-reexecution-deps:
+    desc: Setup custom deps for reexecution benchmarks. Use LIBEVM_REF and/or FIREWOOD_REF env vars.
+    cmds:
```
maru-ava: (No action required) Maybe update the polyrepo script to accept LIBEVM_REF and FIREWOOD_REF so that this task is just providing those env vars to the script?

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
```diff
@@ -84,6 +93,17 @@ jobs:
     runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v4
+      - name: Install Nix
```
maru-ava, line 98: Maybe make nix installation optional as well?

Also, what is the implication of calling this action twice (once here, once in the reexec action)?

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -152,6 +152,16 @@ runs:
...
         GH_JOB_ID: ${{ inputs.job }}
+        # Required for Nix builds on self-hosted runners
+        TMPDIR: /tmp
+        NIX_BUILD_TOP: /tmp
+    - name: Upload Benchmark Artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: benchmark-output-${{ inputs.run-id }}-${{ inputs.run-attempt }}-${{ inputs.job }}-${{ inputs.test }}-${{ inputs.runner_type }}
+        path: ${{ env.BENCHMARK_OUTPUT_FILE }}
+        retention-days: 30
```
maru-ava: Maybe omit this? afaik 30 days is the default

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -152,6 +152,16 @@ runs:
         GH_RUN_NUMBER: ${{ inputs.run-number }}
         GH_RUN_ATTEMPT: ${{ inputs.run-attempt }}
         GH_JOB_ID: ${{ inputs.job }}
+        # Required for Nix builds on self-hosted runners
+        TMPDIR: /tmp
+        NIX_BUILD_TOP: /tmp
+    - name: Upload Benchmark Artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: benchmark-output-${{ inputs.run-id }}-${{ inputs.run-attempt }}-${{ inputs.job }}-${{ inputs.test }}-${{ inputs.runner_type }}
```
maru-ava: What's the rational for embedding all these values in the name?

tests/reexecute/c/README.md
```diff
@@ -250,6 +250,47 @@ To execute a benchmark with any of these options, you must use a compatible `CUR
...
+- Benchmarking performance changes in dependencies
+- Reproducing issues with specific dependency versions
+
+### Local Usage
+
+**Prerequisite**: You must be in a nix shell (`nix develop`) to run benchmark tasks.
+
+First, set up dependencies using the `setup-reexecution-deps` task, then run the benchmark:
+
+```bash
```
maru-ava, line 268: (No action required) Maybe include a single example with both versions and document that both or either one can be provided? Same comment for the CI invocation examples.

tests/reexecute/c/README.md
```diff
@@ -250,6 +250,47 @@ To execute a benchmark with any of these options, you must use a compatible `CUR
...
+# Test with custom libevm version
+LIBEVM_REF=v1.2.3 ./scripts/run_task.sh setup-reexecution-deps
+./scripts/run_task.sh test-cchain-reexecution -- firewood-101-250k
+
+# Test with both custom versions
+FIREWOOD_REF=abc123def LIBEVM_REF=v1.2.3 ./scripts/run_task.sh setup-reexecution-deps
+./scripts/run_task.sh test-cchain-reexecution -- firewood-101-250k
+```
+
+### How It Works
```
maru-ava: (No action required) Maybe focus this doc additions on usage rather than duplicating implementation details?

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: The default is 90 days per [GitHub docs](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization). 

I'm keeping 30 days explicit because: (1) it documents intent rather than relying on assumed defaults, and (2) 30 days is enough for debugging while avoiding unnecessary artifact accumulation given our run volume.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Artifact names must be unique within a workflow run. Matrix jobs run in parallel and would otherwise conflict with a 409 error. These values ensure each matrix combination produces a uniquely named artifact.

Elvis339: The identifiers ensure uniqueness:
- run-id + run-attempt → unique per workflow execution
- job → matrix job identifier
- test + runner_type → differentiates matrix combinations

Taskfile.yml
Elvis339: That's so much better, thanks for recommending this.

https://github.com/ava-labs/avalanchego/pull/4650/commits/954121b3b7007ed46c53eff5111a058f02dfebfd

Edit: new commit, updated doc and how polyrepo is being ran https://github.com/ava-labs/avalanchego/pull/4650/commits/1f06a146081baca13ef0810274a3f32f28dd72e4

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: While documenting intent might be a good principle, what _is_ your intent in this case? We don't set retention days for other uses of this action, what makes this one special?

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: Artifacts are already effectively namespaced by PR run (i.e. run-id and run-attempt). That's why we can use names like `upgrade-tmpnet-data` and not have collisions across PR runs.

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Fair point. Removing for consistency with other artifact uploads in the repo. https://github.com/ava-labs/avalanchego/pull/4650/changes/221157a1523fb61eef592225f48ee3cc2732972f

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Simplified the name. Thanks for explaining, the tests indeed pass. 

https://github.com/ava-labs/avalanchego/pull/4650/changes/6aef9dfae2a4e9096bf8124513fbf76c62697fea

tests/reexecute/c/README.md
Elvis339: https://github.com/ava-labs/avalanchego/pull/4650/commits/d9b271708398df52009b976aa814d69f8a8d0318

Edit: new commit, updated doc and how polyrepo is being ran https://github.com/ava-labs/avalanchego/pull/4650/commits/1f06a146081baca13ef0810274a3f32f28dd72e4

tests/reexecute/c/README.md
Elvis339: https://github.com/ava-labs/avalanchego/pull/4650/commits/d9b271708398df52009b976aa814d69f8a8d0318

Edit: new commit, updated doc and how polyrepo is being ran https://github.com/ava-labs/avalanchego/pull/4650/commits/1f06a146081baca13ef0810274a3f32f28dd72e4

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
Elvis339: Made Nix installation conditional - only runs when libevm-ref or firewood-ref is specified.
Why workflow installs Nix before the action:
- Polyrepo builds Firewood via nix build
- The nix develop shell provides Go for go get libevm@<ref>

Double install implication: The action's Nix install is idempotent it detects the existing installation and skips:
https://github.com/ava-labs/avalanchego/actions/runs/21261969343/job/61191483101#step:6:53

The run failed but unrelated to Nix - the github-action-benchmark step tries to checkout gh-pages but fails because go.mod/go.sum were modified by the dependency setup.

Edit: fixed https://github.com/ava-labs/avalanchego/pull/4650/commits/e3ab8ad7ae898eb909d35c76e34961f41fa9c4c5 https://github.com/ava-labs/avalanchego/actions/runs/21262297545/job/61192576708

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -56,6 +56,9 @@ inputs:
   push-github-action-benchmark:
     description: 'Whether to push the benchmark result to GitHub.'
     required: true
+  skip-benchmark-comparison:
```
Elvis339, line 59: When custom dependencies (firewood-ref or libevm-ref) are used,
go.mod/go.sum are modified. This causes github-action-benchmark
to fail when checking out gh-pages.

New input skips comparison step only when custom deps are present,
preserving summary display for normal PR runs.

maru-ava: Would it make sense to make this skip automatic with an appropriate warning if the tree is dirty?

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
maru-ava: Maybe add a comment documenting the idempotency for the benefit of future maintainers?

Taskfile.yml
```diff
@@ -174,6 +174,10 @@ tasks:
     desc: Runs shellcheck to check sanity of shell scripts
     cmd: ./scripts/shellcheck.sh
 
+  polyrepo:
```
maru-ava: Please use an action name e.g. `run-polyrepo`.

scripts/run_polyrepo.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -euo pipefail
+
+# Run polyrepo tool for managing local dependencies.
+#
+# Usage:
+#   ./scripts/run_polyrepo.sh [polyrepo args...]
+#
+# Environment variables (optional):
+#   LIBEVM_REF - Git ref for libevm (runs: go get && go mod tidy)
```
maru-ava: Maybe use env vars for both args so that the caller has a consistent way of specifying both values?

.github/workflows/c-chain-reexecution-benchmark-gh-native.yml
Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4650/commits/dac526253797b459e46a195e89cfdc396e15d8f4

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: The inputs already tell us when the tree will be dirty if `firewood-ref` or `libevm-ref` is provided, `polyrepo/go get` will modify `go.mod`. A runtime `git status` check would just confirm what we already know statically.

Putting the detection logic inside the action would also conflict with Aaron's constraint he was okay with the action accepting `skip-benchmark-comparison` as an input, but the decision of when to skip should be made by the caller, not the action itself.

maru-ava: @aaronbuchwald Why is it desirable to have to pass in a synthesized variable instead of just detecting this condition locally?

.github/workflows/c-chain-reexecution-benchmark-container.yml
```diff
@@ -107,7 +129,8 @@ jobs:
           prometheus-push-url: ${{ secrets.PROMETHEUS_PUSH_URL || '' }}
           prometheus-username: ${{ secrets.PROMETHEUS_USERNAME || '' }}
           prometheus-password: ${{ secrets.PROMETHEUS_PASSWORD || '' }}
-          push-github-action-benchmark: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master') }}
+          push-github-action-benchmark: ${{ (github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.repository == 'ava-labs/avalanchego' && github.ref_name == 'master')) && matrix.firewood-ref == '' }}
+          skip-benchmark-comparison: ${{ matrix.firewood-ref != '' || matrix.libevm-ref != '' }}
```
maru-ava, line 134: Why are these values the trigger instead of the dirty tree state?

Taskfile.yml
Elvis339: Done: 

https://github.com/ava-labs/avalanchego/pull/4650/commits/a29c22c3e51d0f9bb78d4a5c8c9025f60104da66

.github/workflows/c-chain-reexecution-benchmark-container.yml
Elvis339: Two separate concerns:
- `push-github-action-benchmark` - Controls whether results get pushed to `gh-pages` for historical tracking. Only enabled for scheduled runs on master with default deps.
- `skip-benchmark-comparison` - Controls whether the comparison step runs at all. When custom deps are used, go.mod/go.sum are modified, causing the benchmark action to fail on `gh-pages` checkout (dirty tree error).

Why input values instead of dirty tree check? 
We know statically that if `firewood-ref` or `libevm-ref` is provided, the tree will be dirty. Runtime check would confirm what inputs already tell us.

I'm okay with adding automatic dirty tree detection inside `c-chain-reexecution-benchmark` action if Aaron agrees. From my conversation with him, I inferred the action should stay focused on running benchmarks and not infer/handle dependency state the caller knows it modified deps, so the caller decides whether to skip comparison.

scripts/run_polyrepo.sh
Elvis339: That was my initial implementation :D 

Re-added: https://github.com/ava-labs/avalanchego/pull/4650/commits/a29c22c3e51d0f9bb78d4a5c8c9025f60104da66

.github/actions/c-chain-reexecution-benchmark/action.yml
Elvis339: Adding context on the design. Polyrepo is a composable layer that sits before the benchmark action, setting up deps. The base action runs benchmarks on whatever code is present, then uploads artifacts (Firewood will track performance on their own gh-pages).

Aaron's constraint as I understood it is that the action should run reexecution for the environment that's present not be responsible for in-place modification or inferring dependency state.

This matters because if the action detects dirty tree and decides to skip, it's making decisions about dependency state without knowing why the tree is dirty. The caller (workflow) knows it ran polyrepo → knows the tree is dirty → passes skip-benchmark-comparison: true.

This keeps concerns separated dep setup in workflow, benchmarking in action.

I might be wrong in how I inferred Aaron's constraint happy for him to correct me. Open to doing whatever to progress with this PR.

Elvis339:
Tested with custom firewood dependency:
https://github.com/ava-labs/avalanchego/actions/runs/21286107306/job/61267590660

```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --repo ava-labs/avalanchego \
  --ref es/enable-firewood-dev-workflow \
  -f config="firewood" \
  -f with-dependencies="firewood=v0.0.18" \
  -f start-block="1" \
  -f end-block="10000" \
  -f timeout-minutes="1440"
```
Expected:
- Dependency setup
- Benchmark summary skipped

Tested without custom deps and Firewood:
https://github.com/ava-labs/avalanchego/actions/runs/21286367028/job/61268454476

```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --repo ava-labs/avalanchego \
  --ref es/enable-firewood-dev-workflow \
  -f test="hashdb-101-250k" \
  -f timeout-minutes="60"
```

Expected:
- Nix/dependency configuration skipped
- Benchmark summary displayed

.github/actions/c-chain-reexecution-benchmark/action.yml
aaronbuchwald: The original constraint is that the custom action should not need to know about the special case of firewood or overwriting dependencies. This seems a better separation of responsibilities to me than to pass in firewood-ref, libevm-ref, etc. which imo is better handled as a separate step.

I am not sure why `skip-benchmark-comparison` needs to be added as opposed to simply setting `push-github-action-benchmark` to false.

Elvis339: The issue is that github-action-benchmark fails entirely when the tree is dirty - it tries to checkout gh-pages and errors on the dirty working tree. Setting push-github-action-benchmark: false still runs the step, which fails. We need to skip the step entirely.

PR #4652, opened 2025-12-03
refactor(vmsync): extract strategy pattern for static state sync via `Executor` interface

### LINKED ISSUE #4651: Transform sync client to execute based on static or dynamic sync strategy
## Summary

Simplify `vmsync/client.go` to be a thin adapter layer that delegates to either a static or dynamic sync strategy based on configuration.

## Current Problems

| Problem                | Description                                                                   |
| ---------------------- | ----------------------------------------------------------------------------- |
| Mixed responsibilities | Client handles orchestration, syncer creation, finalization, and engine hooks |
| Inline branching       | Static vs dynamic logic interleaved throughout `acceptSyncSummary`            |
| Dynamic-only methods   | `OnEngineX` methods are no-ops for static sync but exposed on Client          |
| Long methods           | `acceptSyncSummary` is 65+ lines with multiple concerns                       |
| Tight coupling         | Finalization logic is embedded in client                                      |

NOTE: part of the goal is to prepare the client for the rollout of dynamic state sync support. Check https://github.com/ava-labs/avalanchego/pull/4646.

powerslider:
## Why this should be merged

Check #4651 

## How this works

Restructure the `vmsync` package to separate concerns and prepare for future dynamic state sync support.

- Add `Executor` interface with `Execute` method for sync execution.
- Implement `staticExecuhoc` for current sequential sync behavior.
- Simplify `client.go` to orchestrate sync lifecycle via strategy.

This refactoring enables adding dynamic state sync as a separate `Executor` implementation without modifying existing static sync logic.

## How this was tested

existing UT

no

resolves #4651

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -114,7 +114,7 @@ runs:
         echo "Grafana: ${metrics_url}"
         echo "🔗 [View Grafana Dashboard](${metrics_url})" >> "$GITHUB_STEP_SUMMARY"
       env:
-        GRAFANA_URL: https://grafana-poc.avax-dev.network/d/Gl1I20mnk/c-chain?orgId=1&refresh=10s&var-filter=is_ephemeral_node%7C%3D%7Cfalse&var-filter=gh_repo%7C%3D%7C${{ inputs.repository_owner }}%2F${{ inputs.repository_name }}&var-filter=gh_run_id%7C%3D%7C${{ inputs.run_id }}&var-filter=gh_run_attempt%7C%3D%7C${{ inputs.run_attempt }}
+        GRAFANA_URL: https://grafana-poc.avax-dev.network/d/Gl1I20mnk/c-chain?orgId=1&refresh=10s&var-filter=is_ephemeral_node%7C%3D%7Cfalse&var-filter=gh_repo%7C%3D%7C${{ inputs.repository-owner }}%2F${{ inputs.repository-name }}&var-filter=gh_run_id%7C%3D%7C${{ inputs.run-id }}&var-filter=gh_run_attempt%7C%3D%7C${{ inputs.run-attempt }}
         GH_JOB_ID: ${{ inputs.job }}
     - name: Warn that collection of metrics and logs will not be performed
       if: inputs.prometheus-username == ''
@@ -174,11 +174,11 @@ runs:
         PROMETHEUS_PUSH_URL: ${{ inputs.prometheus-push-url }}
         PROMETHEUS_USERNAME: ${{ inputs.prometheus-username }}
         PROMETHEUS_PASSWORD: ${{ inputs.prometheus-password }}
-        GH_REPO: ${{ inputs.repository_owner }}/${{ inputs.repository_name }}
+        GH_REPO: ${{ inputs.repository-owner }}/${{ inputs.repository-name }}
         GH_WORKFLOW: ${{ inputs.workflow }}
-        GH_RUN_ID: ${{ inputs.run_id }}
-        GH_RUN_NUMBER: ${{ inputs.run_number }}
-        GH_RUN_ATTEMPT: ${{ inputs.run_attempt }}
+        GH_RUN_ID: ${{ inputs.run-id }}
+        GH_RUN_NUMBER: ${{ inputs.run-number }}
+        GH_RUN_ATTEMPT: ${{ inputs.run-attempt }}
         GH_JOB_ID: ${{ inputs.job }}
     - name: Compare Benchmark Results
       uses: benchmark-action/github-action-benchmark@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,14 +43,18 @@ jobs:
         shell: bash
         run: ./scripts/run_task.sh test-fuzz
   e2e:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, custom-arm64-noble]
     steps:
       - uses: actions/checkout@v4
       - name: Run e2e tests
         uses: ./.github/actions/run-monitored-tmpnet-cmd
         with:
           run: ./scripts/run_task.sh test-e2e-ci
-          artifact_prefix: e2e
+          artifact_prefix: e2e-${{ matrix.os }}
           filter_by_owner: avalanchego-e2e
           prometheus_url: ${{ secrets.PROMETHEUS_URL || '' }}
           prometheus_push_url: ${{ secrets.PROMETHEUS_PUSH_URL || '' }}
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ AvalancheGo support tiers:
 | Architecture | Operating system | Support tier  |
 | :----------: | :--------------: | :-----------: |
 |    amd64     |      Linux       |       1       |
-|    arm64     |      Linux       |       2       |
+|    arm64     |      Linux       |       1       |
 |    arm64     |      Darwin      |       2       |
 |    amd64     |      Darwin      | Not supported |
 |    amd64     |     Windows      | Not supported |
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -16,6 +16,7 @@ This version is backwards compatible to [v1.14.0](https://github.com/ava-labs/av
 ### Fixes
 
 - Update go version to 1.24.11
+- Add Linux arm64 to tier one support status
 
 ### EVM
 
--- a/flake.nix
+++ b/flake.nix
@@ -71,9 +71,6 @@
 
             # s5cmd for rapid s3 interactions
             s5cmd
-          ] ++ lib.optionals stdenv.isDarwin [
-            # macOS-specific frameworks
-            darwin.apple_sdk.frameworks.Security
           ];
 
           # Add scripts/ directory to PATH so kind-with-registry.sh is accessible
--- a/network/p2p/network.go
+++ b/network/p2p/network.go
@@ -167,6 +167,13 @@ func (p *Peers) Has(nodeID ids.NodeID) bool {
 	return p.set.Contains(nodeID)
 }
 
+func (p *Peers) Len() int {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	return p.set.Len()
+}
+
 // Sample returns a pseudo-random sample of up to limit Peers
 func (p *Peers) Sample(limit int) []ids.NodeID {
 	p.lock.RLock()
--- a/network/p2p/network_test.go
+++ b/network/p2p/network_test.go
@@ -574,6 +574,7 @@ func TestPeersSample(t *testing.T) {
 			sampleable := set.Set[ids.NodeID]{}
 			sampleable.Union(tt.connected)
 			sampleable.Difference(tt.disconnected)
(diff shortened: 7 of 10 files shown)


graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -42,11 +43,11 @@ type BlockAcceptor interface {
...
 }
 
-// EthBlockWrapper can be implemented by a concrete block wrapper type to
-// return *types.Block, which is needed to update chain pointers at the
-// end of the sync operation.
-type EthBlockWrapper interface {
-	GetEthBlock() *types.Block
+// SyncStrategy defines how state sync is executed.
+// Implementations handle the sync lifecycle differently based on sync mode.
+type SyncStrategy interface {
```
alarso16: The name `SyncStrategy` indicates to me that this interface has something to do with a strategy, which is just not true. It just syncs. Additionally, the name `Start` indicates that it only starts the syncers, and then returns. Your comments are correct and the code works here, I just don't like the names

May I suggest `SyncOperation` and `Sync`? I honestly don't eve know if this needs to be a whole interface (does it?) You could also just define:
```
type SyncOperation func(context.Context, message.Syncable) error
```

graft/coreth/plugin/evm/vmsync/finalizer.go
```diff
@@ -0,0 +1,152 @@
...
+	state              *chain.State
+	acceptor           BlockAcceptor
+	verDB              *versiondb.Database
+	metadataDB         database.Database
+	extender           syncpkg.Extender
+	lastAcceptedHeight uint64
+}
+
+// newFinalizer creates a new finalizer with the given dependencies.
+func newFinalizer(
```
alarso16: Is this creator going to have more work to do? I'm not sure abstracting this into a struct is productive

graft/coreth/plugin/evm/vmsync/client.go
alarso16: Good to know this is a valid pattern name, but I do think it's a little vague. Don't love the name

powerslider: This implements the strategy pattern. Check [here](https://refactoring.guru/design-patterns/strategy). The whole idea is to have a `static` strategy and a `dynamic` strategy. The latter will be introduce in another PR. This lays the groundwork for that. Regarding using the name `Strategy`, I was influenced by the name of the pattern, but I don't mind changing it to sync operation or sync mode, etc.

graft/coreth/plugin/evm/vmsync/finalizer.go
powerslider: I found a better way to do this. You can check https://github.com/ava-labs/avalanchego/pull/4652/changes/ed5e8e78544f00819f0cdebaac23440e263e7770. I really wanted to have a way to call the finalization logic in each strategy, but now I returned the logic in the client itself and added a `Commiter` interface for the strategies to depend on. It fits nicely since all the parameters needed for this process are accessible from `ClientConfig` on the client.

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -42,13 +58,20 @@ type BlockAcceptor interface {
...
 }
 
+// SyncStrategy defines how state sync is executed.
+// Implementations handle the sync lifecycle differently based on sync mode.
+type SyncStrategy interface {
+	// Start begins the sync process and blocks until completion or error.
+	Start(ctx context.Context, summary message.Syncable) error
+}
+
+var _ Committer = (*client)(nil)
```
alarso16: Not clear to me why this is abstracted. Are we going to have multiple implementations soon?

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -268,82 +210,88 @@ func (client *client) acceptSyncSummary(proposedSummary message.Syncable) (block
...
-// finishSync is responsible for updating disk and memory pointers so the VM is prepared
-// for bootstrapping. Executes any shared memory operations from the atomic trie to shared memory.
-func (client *client) finishSync() error {
-	stateBlock, err := client.State.GetBlock(context.TODO(), ids.ID(client.summary.GetBlockHash()))
+// Error returns a non-nil error if one occurred during the sync.
+func (c *client) Error() error { return c.err }
+
+// Commit implements Committer. It updates disk and memory pointers so the VM
+// is prepared for bootstrapping. Executes any shared memory operations from
+// the atomic trie to shared memory.
```
alarso16, line 271: I think you should also say it preps the blockchain for execution (see `ResetToSyncedBlock`)

alarso16 (commented):
Looks correct, left some nits

graft/coreth/plugin/evm/vmsync/client.go
powerslider: The interface decouples `staticStrategy` from the concrete `*client`, enabling unit tests to mock the commit behavior without constructing a full client with all its dependencies (eth.Ethereum, databases, etc.). It's a single-method interface following Go idioms, and costs almost nothing while making the strategy independently testable. When `dynamicStrategy` is added, it will use the same `Committer` interface.

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -268,82 +210,89 @@ func (client *client) acceptSyncSummary(proposedSummary message.Syncable) (block
...
-func (client *client) finishSync() error {
-	stateBlock, err := client.State.GetBlock(context.TODO(), ids.ID(client.summary.GetBlockHash()))
+// Error returns a non-nil error if one occurred during the sync.
+func (c *client) Error() error { return c.err }
+
+// Commit implements Committer. It resets the blockchain to the synced block,
+// preparing it for execution, and updates disk and memory pointers so the VM
+// is ready for bootstrapping. Also executes any shared memory operations from
+// the atomic trie to shared memory.
+func (c *client) Commit(ctx context.Context, summary message.Syncable) error {
```
ceyonur: nit: this looks to me that it does more than just committing but more like Finalizing/Finishing. But I don't have the full picture so not sure if we plan to use this interface elsewhere.

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -268,82 +210,89 @@ func (client *client) acceptSyncSummary(proposedSummary message.Syncable) (block
...
+		return fmt.Errorf("failed to commit db: %w", err)
 	}
 
-	log.Info("Starting state sync", "summary", proposedSummary)
+	return nil
+}
 
-	// create a cancellable ctx for the state sync goroutine
+// startAsync launches the sync executor in a background goroutine.
+func (c *client) startAsync(executor Executor, summary message.Syncable) block.StateSyncMode {
```
ceyonur, line 234: currently this seems a bit overkill to work with interfaces as it's a unexported helper. But then the whole `Executor` interface seems an overkill. I assume we want use this interface in the future?

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -356,52 +305,97 @@ func (client *client) finishSync() error {
...
+		codeQueue, c.config.RequestSize,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create EVM state syncer: %w", err)
+	}
+
+	syncers := []syncpkg.Syncer{blockSyncer, codeSyncer, stateSyncer}
+
+	if c.config.Extender != nil {
+		atomicSyncer, err := c.config.Extender.CreateSyncer(c.config.Client, c.config.VerDB, summary)
```
ceyonur: nit: rename and try to avoid `atomic` as that info is not immediatly available (`Extender` is a generic interface).

ceyonur: (the idea is to use this extender to maybe create multiple syncers or selectively create, so IMO it should be generic)

graft/coreth/plugin/evm/vmsync/client.go
```diff
@@ -356,52 +305,97 @@ func (client *client) finishSync() error {
...
 		return err
 	}
-	return client.VerDB.Commit()
+	return c.config.VerDB.Commit()
 }
 
-// Error returns a non-nil error if one occurred during the sync.
-func (client *client) Error() error { return client.err }
+// newSyncerRegistry creates a registry with all required syncers for the given summary.
+func (c *client) newSyncerRegistry(summary message.Syncable) (*SyncerRegistry, error) {
```
ceyonur, line 353: (no action required): there are bunch of code moving around. the diff could have been kept less than what it shows. it kind-of hinders the actual meaningful changes (and actually could hide potentially invalid changes as well)

graft/coreth/plugin/evm/vmsync/client.go
powerslider: This will be needed for my next PR when I introduce the dynamic `Executor`. That is why I need it here.

graft/coreth/plugin/evm/vmsync/client.go
powerslider: This was called `Finalizer` before, but it clashed with the `synckg.Finalizer` interface, so I changed it to be more focused. In this case commiting to storage is "finalizing", so that is why I named it to `Commiter`. I think it is fine for now. In the future if we add more finalization logic, we could rename it to something else, but I'd rather defer that to that future moment.

graft/coreth/plugin/evm/vmsync/client.go
powerslider: I totally understand your point, but I think that it will be misleading to say `extenderSyncer`, because in essence we are instantiating the `atomic` syncer here and I think that we need to mention that in the error handling in order to have proper troubleshooting. Given that what would you suggest as a way to not pollute the abstraction, but also mark that this is the `atomic` syncer in error handling and logging?

graft/coreth/plugin/evm/vmsync/client.go
powerslider: I think that this is acceptable in this PR since we are actually thoroughly refactoring the client and moving around code is normal in this scenario. I can't reduce the diff in this specific case since the whole idea of the PR was to change the responsibilities of the sync client from an execution responsible component to an orchestration driven component.

graft/coreth/plugin/evm/vmsync/client.go
powerslider: An idea that pops into my mind is to have something like a `Name()` method in the `Extender` interface that can expose the implementation name/ID of this component without explicitly writing it as a string or var name anywhere? - basically owned identity which is the same approach we use with the `Syncer` interface.

ceyonur: can we wrap the error in the `atomic.CreateSyncer` like with `atomic.CreateSyncer failed: %w`?

graft/coreth/plugin/evm/vmsync/client.go
ceyonur: I was referring moving the code from `registerSyncers` vs `newSyncerRegistry`, which two are very similar and diff can be more obvious if we haven't moved it to the below (here).

graft/coreth/plugin/evm/vmsync/client.go
ceyonur: This is a pattern that's being used in strevm https://github.com/ava-labs/strevm/blob/main/saexec/execution.go#L102

graft/coreth/plugin/evm/vmsync/client.go
ceyonur: (as said no action required)

PR #4660, opened 2025-12-04
Add Read Only SharedMemory

joshua-kim:
## Why this should be merged

As part of https://github.com/ava-labs/avalanchego/pull/3732/files, when replaying blocks to migrate state from the old db format to firewood we need to replay atomic txs. Atomic txs cannot be replayed however, because duplicate writes to atomic memory are not permitted. One way to get around this is to allow reads from atomic memory while replaying the transactions.

## How this works

Drops writes to shared memory.

## How this was tested

Tested through linked PR

No

### DIFF
--- a/chains/atomic/shared_memory.go
+++ b/chains/atomic/shared_memory.go
@@ -4,13 +4,18 @@
 package atomic
 
 import (
+	"errors"
+
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/database/versiondb"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils"
 )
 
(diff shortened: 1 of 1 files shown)


chains/atomic/shared_memory.go
```diff
@@ -163,3 +168,52 @@ func (sm *sharedMemory) Apply(requests map[ids.ID]*Requests, batches ...database
...
+		limit,
+	)
+}
+
+// Apply drops atomic write requests.
+func (ReadOnly) Apply(
+	_ map[ids.ID]*Requests,
+	bs ...database.Batch,
+) error {
+	// TODO hacky
```
maru-ava, line 213: Maybe elaborate on or remove this TODO?

maru-ava (commented):
> Tested through linked PR

Why is this broken out if it can't be tested independently?

PR #4669, opened 2025-12-09
refactor: Firewood Revisions

alarso16:
## Why this should be merged

Now that using `Revision` is concurrent safe and guaranteed to not cause undefined behavior, it should be used a more canonical way of accessing state. 

## How this works

Uses Firewood's `Revision` instead of relying on a one-Revision cache internal to Rust. The Revision has a finalizer attached to release all related memory when there are no more references.

## How this was tested

CI and re-execution tests

No

### DIFF
--- a/graft/coreth/core/txpool/legacypool/legacypool.go
+++ b/graft/coreth/core/txpool/legacypool/legacypool.go
@@ -446,6 +446,11 @@ func (pool *LegacyPool) Close() error {
 	if pool.journal != nil {
 		pool.journal.close()
 	}
+
+	// remove all references to state to allow GC to reclaim memory
+	pool.pendingNonces = nil
+	pool.currentState = nil
+
 	log.Info("Transaction pool stopped")
 	return nil
 }
--- a/graft/evm/firewood/triedb.go
+++ b/graft/evm/firewood/triedb.go
@@ -486,16 +486,16 @@ func (t *TrieDB) removeProposalFromMap(pCtx *ProposalContext) {
 // Reader retrieves a node reader belonging to the given state root.
 // An error will be returned if the requested state is not available.
 func (t *TrieDB) Reader(root common.Hash) (database.Reader, error) {
-	if _, err := t.fwDisk.GetFromRoot(ffi.Hash(root), []byte{}); err != nil {
+	revision, err := t.fwDisk.Revision(ffi.Hash(root))
+	if err != nil {
 		return nil, fmt.Errorf("firewood: unable to retrieve from root %s: %w", root.Hex(), err)
 	}
-	return &reader{db: t, root: ffi.Hash(root)}, nil
+	return &reader{revision: revision}, nil
 }
 
 // reader is a state reader of Database which implements the Reader interface.
 type reader struct {
-	db   *TrieDB
-	root ffi.Hash // The root of the state this reader is reading.
+	revision *ffi.Revision
 }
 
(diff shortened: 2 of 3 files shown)


alarso16:
> Revisions hanging around will prevent the revision manager from reaping the nodestores, and this may not be good from a memory perspective. Last I checked, finalizers were considered expensive and could run at an arbitrary time after becoming unreachable. What's preventing us from explicitly dropping the revision?

We don't know when it will no longer be used. There's very strict interfaces to adhere to, and that includes that we cannot assume there will be any behavior to indicate the `Reader` is no longer used.

> Also, if we're going with finalizers, is there any particular reason we're not using `runtime.AddCleanup`?

No particular reason! In our case, they should be equivalent. We know that they are properly being cleaned because closing the database doesn't hang. We could use `runtime.AddCleanup` in Firewood, and that's worth thinking about

AminR443:
Sounds good to me. I think we should consider using `runtime.AddCleanup`, since Go’s docs suggest it as the new go-to option:

> New Go code should consider using [[AddCleanup](https://pkg.go.dev/runtime#AddCleanup)](https://pkg.go.dev/runtime#AddCleanup) instead, which is much less error-prone than SetFinalizer.

I'll create an issue for that in firewood repo.

graft/coreth/core/txpool/legacypool/legacypool.go
```diff
@@ -446,6 +446,11 @@ func (pool *LegacyPool) Close() error {
 	if pool.journal != nil {
 		pool.journal.close()
 	}
+
+	// remove all references to state to allow GC to reclaim memory
+	pool.pendingNonces = nil
```
alarso16, line 451: These keep around a `StateDB`, which holds onto the `*firewood.accountTrie`, so the database closure will hang unless we remove these references. Another (worse) option is that on `state.Trie.Copy()`, we don't copy the `Reader`, and instead use `GetFromRoot` then. This seems too messy, so I have this small line change instead.

graft/evm/firewood/triedb.go
```diff
@@ -493,16 +493,16 @@ func (db *Database) removeProposalFromMap(pCtx *ProposalContext) {
 // Reader retrieves a node reader belonging to the given state root.
 // An error will be returned if the requested state is not available.
 func (db *Database) Reader(root common.Hash) (database.Reader, error) {
-	if _, err := db.fwDisk.GetFromRoot(ffi.Hash(root), []byte{}); err != nil {
+	revision, err := db.fwDisk.Revision(ffi.Hash(root))
+	if err != nil {
 		return nil, fmt.Errorf("firewood: unable to retrieve from root %s: %w", root.Hex(), err)
 	}
-	return &reader{db: db, root: ffi.Hash(root)}, nil
+	return &reader{revision: revision}, nil
```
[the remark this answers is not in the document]
alarso16: Automatically cleaned...

graft/coreth/core/txpool/legacypool/legacypool.go
RodrigoVillar: I wonder if it's possible to remove `*state.StateDB` from the `noncer` - doing so could allow us to remove the `pool.pendingNonces = nil` statement (which seems odd to me IMO). After all, the `noncer` is only used by the legacy pool which itself already has a reference to `*state.StateDB`.

This is out of scope for the PR though, so approving nonetheless.

maru-ava:
Infra blips :/

PR #4673, opened 2025-12-10
chore(reexecute/c): add back scraping of multiple metrics

RodrigoVillar:
## Why this should be merged

With #4640 merged in, we now have the ability to register multiple benchmark results without corrupting the Github-Action-Benchmark dashboard. This PR therefore re-registers the MeterVM metrics which gives us a more precise assessment for the re-execution test (by reporting sub-metrics)  rather than just printing the average mGas/s for the entire benchmark.

## How this works

This PR is basically a copy-and-paste of #4369.

## How this was tested

CI

No

### DIFF
--- a/tests/reexecute/c/metrics.go
+++ b/tests/reexecute/c/metrics.go
@@ -45,6 +45,12 @@ var (
 	}
 )
 
+type topLevelMetric struct {
+	name  string
+	query string
+	kind  metricKind
+}
+
 func getMetricValue(registry prometheus.Gatherer, metric topLevelMetric) (float64, error) {
 	metricFamilies, err := registry.Gather()
 	if err != nil {
@@ -68,12 +74,6 @@ func getMetricValue(registry prometheus.Gatherer, metric topLevelMetric) (float6
 	}
 
 	return 0, fmt.Errorf("metric %s not found", query)
-}
-
-type topLevelMetric struct {
-	name  string
-	query string
-	kind  metricKind
 }
 
 func getTopLevelMetrics(tc tests.TestContext, tool *benchmarkTool, registry prometheus.Gatherer, elapsed time.Duration) {


tests/reexecute/c/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, mf := range metricFamilies {
+		switch metric.kind {
+		case counter:
+			if mf.GetName() == query {
+				return mf.GetMetric()[0].Counter.GetValue(), nil
+			}
+		case gauge:
+			if mf.GetName() == query {
+				return mf.GetMetric()[0].Gauge.GetValue(), nil
+			}
```
aaronbuchwald, line 70: Hmm seems that if the kind is mismatched with the actual registered prometheus metric, then it will fail late here.

Is there any way to make this fail earlier / louder / with a better error message if that's the case?

tests/reexecute/c/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+	totalGGas := totalGas / ggas
+	msPerGGas := (float64(elapsed) / nsPerMs) / totalGGas
+	tool.addResult(msPerGGas, "ms/ggas")
+
+	for _, metric := range meterVMMetrics {
+		metricVal, err := getMetricValue(registry, metric)
+		r.NoError(err)
+
+		metricValMS := (metricVal / nsPerMs) / totalGGas
```
aaronbuchwald: the type doesn't make the conversion from ns to ms obvious, what's the best way to make that explicit? Ie include in the type somehow or just add a comment that we assume all of these metrics follow the expected format.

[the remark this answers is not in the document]
RodrigoVillar: Good call; done: 43f9633

RodrigoVillar: I went with renaming some of the variables + adding a comment to make this clear: 43f9633

tests/reexecute/c/metrics.go
RodrigoVillar: Not sure what you mean by "it will fail". If we mismatch metric kinds (e.g. say it's a counter but it's actually a gauge), then `getMetricValue()` will still return, but the value returned will be `0` - I could require that values returned from `getMetricValue()` are nonzero if that's what you mean for the function to succeed.

aaronbuchwald: Ah I was thinking that there's no guarantee `Gauge` or `Counter` field will be non-nil, but it looks like `GetValue` checks for nil and returns 0 if it's not populated.

tests/reexecute/c/metrics.go
aaronbuchwald: Hmm this is a bit more explicit, but I meant more that we could document in the custom metric type exactly how it will be normalized, so that it's more clear. This is fine for now, just a suggestion to make it more explicit.

aaronbuchwald (approved):

tests/reexecute/c/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+		default:
+			return 0, fmt.Errorf("metric type unknown: %d", metric.kind)
+		}
+	}
+
+	return 0, fmt.Errorf("metric %s not found", query)
+}
+
+type topLevelMetric struct {
```
maru-ava: I think this probably belongs above any function definitions, especially given its use in the initial var block.

RodrigoVillar: Good call; done: 4144d9e

PR #4674, opened 2025-12-10
[migrate-subnet-evm][6] Automated Lint 

JonathanOppenheimer:
## Why this should be merged

This PR ensures that the lint CI passes. 

## How this works

From the repo root:

```bash
./scripts/run_task.sh lint-fix
cd graft/subnet-evm
task generate-mocks
task generate-codec 
./scripts/run_task.sh lint-fix
```

## How this was tested

CI

No


graft/subnet-evm/core/gen_genesis.go
StephenButtolph, line 1: Are we sure we should be modifying this?

JonathanOppenheimer: These changes are automatic on `task generate-codec`. Without them CI fails.

PR #4688, opened 2025-12-12
refactor!: Create a shared EVM module

alarso16:
## Why this should be merged

This also shared go-ethereum code to be used by both coreth and subnet-evm without violating licensing issues. Furthermore, it allows testing with both sets of registrations.

This maybe shouldn't be merged in its current form. Likely, it would be easier to just create the module and CI in one PR, and then add a simple implementation in a different PR

## How this works

Commit 1 preps a new module and CI for shared evm code.

Commit 2 moves the RPC package, and adds abstractions for testing ease. The dubious part is definitely the global for knowing whether it's a coreth or subnet-evm test for when we need to check constants.

## How this was tested

Well it passes CI! To be convinced that both registrations are run and either will cause a CI failure, I recommend checking this out locally and altering the constant set in `graft/evm-shared/rpc/subscription_test.go`.

No?


ARR4N:
Code originally from `geth`, used by both `coreth` and `subnet-evm` should be sourced from `libevm` so there's only a single instance and to allow for simpler upstream syncing.

alarso16:
Talked offline, this is still the temporary middle place before properly deduplicating with libevm

alarso16:
Instead of importing coreth and subnet-evm into this evm directory, we should instead find a way of to move this into libevm.

PR #4690, opened 2025-12-15
chore: Create shared `evm` module

alarso16:
## Why this should be merged

This creates all the groundwork to de-duplicate between coreth and subnet-evm in a shared repository. This includes CI and a trivial refactor to prove it works

## How this works

Adds all coreth/subnet-evm CI (copy-paste) and lots of `go mod tidy`s

## How this was tested

See #4688 for a complicated use case

No

### DIFF
--- a/graft/evm/scripts/lint_setup.sh
+++ b/graft/evm/scripts/lint_setup.sh
@@ -38,9 +38,8 @@ function setup_lint {
   fi
 
   local upstream_folders_file="./scripts/upstream_files.txt"
-  # check that the file exists
   if [[ ! -f "$upstream_folders_file" ]]; then
-    echo "upstream_folders_file not found: $upstream_folders_file"
+    # if the file doesn't exist, assume no upstream files
     upstream_folders=()
   else
     # Read the upstream_folders file into an array


graft/evm/.golangci.yml
alarso16, line 1: In the future, we should de-duplicate with coreth and subnet-evm (but certainly out of scope)

graft/evm/scripts/shellcheck.sh
alarso16, line 1: This should probably be de-duplicated with coreth and subnet-evm

graft/evm/Taskfile.yml
alarso16, line 1: This was copy-pasted from coreth with paths adjusted

graft/evm/scripts/shellcheck.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
JonathanOppenheimer: Maybe out of scope for this PR, but I want to start deduplicating these scripts as well -- Maru pointed it out, and we can start transitioning to the AvalancheGo version of these files. 

Just wanted to let you know I will probably make a PR removing this shortly.

maru-ava: Please remove this in favor of what appears in ./scripts/shellcheck.sh. We should always be using the shellcheck provided in the nix shell rather than maintaining this legacy method.

graft/evm/.golangci.yml
```diff
@@ -0,0 +1,77 @@
+version: "2"
```
JonathanOppenheimer, line 1: What is the point of a custom .golangci.yml without an upstream files? (we only use this linter on upstream code right?)

graft/evm/Taskfile.yml
```diff
@@ -0,0 +1,99 @@
+# https://taskfile.dev
```
JonathanOppenheimer, line 1: All of these tasks seem overkill -- YAGNI?

graft/evm/scripts/coverage.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
JonathanOppenheimer, line 1: Same comment here as on shellcheck

graft/evm/scripts/shellcheck.sh
alarso16: This was an oversight - looks like these are already being linted by avalanchego's linter (which is fine by me)

graft/evm/Taskfile.yml
alarso16: true, we can copy paste later

graft/evm/scripts/coverage.sh
alarso16: Should we de-dupe in this PR? There's quite a few things like this

graft/evm/.golangci.yml
alarso16: There will be upstream code soon, right?

graft/evm/scripts/shellcheck.sh
alarso16: This one we don't need, so I left a comment on the one we actually do need

graft/evm/scripts/coverage.sh
JonathanOppenheimer: Up to you - most (if not all) of the `graft/evm/scripts` will need to be de-duplicated. I'd say all of them are definitely out of scope, but if you're creating a new script in this PR anyway, maybe you could deduplicate it instead.

graft/evm/scripts/shellcheck.sh
JonathanOppenheimer: This is out of date.

graft/evm/.golangci.yml
JonathanOppenheimer: should we create an issue?

graft/evm/.golangci.yml
JonathanOppenheimer: Yeah, I was going to say something also about YAGNI, but we probably are actually going to need it if you move the RPC stuff? You also do literlaly need the text file, but I saw you made it so you didn't anymore.

graft/evm/scripts/coverage.sh
JonathanOppenheimer: If you're not against it, I would.

JonathanOppenheimer: build_test too

graft/evm/scripts/lint_setup.sh
```diff
@@ -38,8 +38,14 @@ function setup_lint {
   fi
 
   local upstream_folders_file="./scripts/upstream_files.txt"
-  # Read the upstream_folders file into an array
-  mapfile -t upstream_folders <"$upstream_folders_file"
+  # check that the file exists
```
maru-ava: From the department of redundant redundancies, maybe avoid comments that can be trivially inferred from the code? I'm more interested in learning the circumstances under which this file would not exist and what the impact of that absence will be.

graft/evm/Taskfile.yml
```diff
@@ -0,0 +1,46 @@
...
+
+  coverage:
+    desc: Display test coverage statistics from coverage.out file
+    cmd: ./scripts/coverage.sh # ci.yml
+
+  lint:
+    desc: Run golangci-lint and check for allowed Ethereum imports in Go code
+    cmd: ./scripts/lint.sh
+
+  lint-all-ci:
```
maru-ava, line 39: (No action required) If there is no ci-specific behavior required, maybe this job doesn't need to exist?

graft/evm/scripts/coverage.sh
alarso16: Neither of these are immediately trivial. This seems to clutter the PR. I'll make an issue for it

graft/evm/Taskfile.yml
alarso16: For now, yes. I but I do think we will be adding more (e.g. mockgen is up-to-date) later. Not sure it's worth changing the workflow file when we add more linters. But if it's confusing or a redundant reedundancy, I'll remove it

graft/evm/scripts/lint_setup.sh
alarso16: Good point, I'll clarify behavior and remove the obvious part

graft/evm/scripts/coverage.sh
JonathanOppenheimer: Sounds good

graft/evm/.golangci.yml
alarso16: We can just add it as needed. This is reasonable to add right now I think

maru-ava (approved):

graft/evm/go.mod
```diff
@@ -0,0 +1,13 @@
...
+
+require github.com/ava-labs/libevm v1.13.15-0.20251016142715-1bccf4f2ddb2
+
+require (
+	github.com/holiman/uint256 v1.2.4 // indirect
+	golang.org/x/crypto v0.45.0 // indirect
+	golang.org/x/sys v0.38.0 // indirect
+)
+
+replace github.com/ava-labs/avalanchego => ../../
```
StephenButtolph, line 13: Because we don't currently depend on avalanchego, this replace looks weird... But this is presumably in place for when this does depend on avalanchego.

StephenButtolph (approved):

graft/evm/go.mod
alarso16: Ah true! This is leftover from a rebase where it does... in any case it will soon

PR #4692, opened 2025-12-15
fix: Implement correct balance updating logic

### LINKED ISSUE #4719: Additional State Upgrade Config Validation
Upgrade configurations allow for specifying a `BalanceChange` to be applied to an account as part of a network state upgrade.

Currently, the `BalanceChange` value is interpreted as a `uint256` and always added to the account balance, but there is no validation that the configured value provided was not negative.

https://github.com/ava-labs/subnet-evm/blob/master/stateupgrade/state_upgrade.go#L34

Whether or not balance deductions are supported, proper overflow & underflow checks should also be added.

JonathanOppenheimer:
## Why this should be merged
Previously `upgradeAccount` would naively assume that any `upgrade.BalanceChange` was positive, even though the underlying BalanceChange was a `Big.Int`, which can be negative. This meant that negative balance updates would be interpreted as positive updates, which is strange, and arguably incorrect behavior. This PR correctly implements the balance changing behavior, and adds checks to ensure account balances are not updated to non-sensical values. Closes https://github.com/ava-labs/avalanchego/issues/4719

## How this was tested
I added a new test file, `state_upgrade_test.go.` This behavior was seemingly not tested prior.  

## Need to be documented?
I don't think so - we are not exposing any additional API methods, we're just fixing an existing one. 

No


JonathanOppenheimer:
Relevant comment: https://github.com/ava-labs/subnet-evm/pull/1745#discussion_r2429918750

PR #4695, opened 2025-12-15
test(reexecute): add firewood chaos test [2/2]

RodrigoVillar:
## Why this should be merged

As mentioned in https://github.com/ava-labs/firewood/issues/1371, we need a crash test to make sure that Firewood is safe against application crashes. This PR fulfills this by adding the `chaos` executable, which does the following:

- Starts the reexecution test as a child process which uses Firewood
- After waiting for `x` amount of time, the executable kills the test
- Starts the reexecution test again, picking up from where the child process crashed

The test is considered a success if the reexecution test is able to start up again.

Closes https://github.com/ava-labs/firewood/issues/1588

## How this works

- Adds the `tests/reexecute/chaos` executable
- Adds firewood chaos test workflow which is scheduled and can also be invoked via a manual dispatch
- Extends reexecution script to add invocation of chaos test

## How this was tested

CI

https://github.com/ava-labs/avalanchego/actions/runs/20284273056/job/58254137735?pr=4695

No

### DIFF
--- a/tests/reexecute/chaos/main.go
+++ b/tests/reexecute/chaos/main.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package main


tests/reexecute/chaos/deps.go
RodrigoVillar, line 1: Refactoring `newMainnetCChainVM()` so that it's usable by `chaos` and `c` will be done in a separate PR prior to this PR being merged in.

.github/workflows/firewood-chaos-test.yml
```diff
@@ -0,0 +1,108 @@
...
+        description: 'Maximum wait time before killing the process (e.g., 150s, 3m).'
+        default: '150s'
+      runner:
+        description: 'Runner to execute the chaos test. Input to the runs-on field of the job.'
+        required: true
+      timeout-minutes:
+        description: 'Timeout in minutes for the job.'
+        default: '60'
+  # XXX: remove this before merging
+  pull_request:
```
RodrigoVillar, line 39: This is for testing the PR - will remove prior to merging.

tests/reexecute/chaos/main.go
```diff
@@ -0,0 +1,218 @@
...
+		r.FailNow("reexecution test terminated prior to crash test", zap.Error(waitErr))
+	default:
+		pgid, err := syscall.Getpgid(cmd.Process.Pid)
+		r.NoError(err)
+
+		log.Debug("killing reexecution test")
+
+		r.NoError(syscall.Kill(-pgid, syscall.SIGKILL))
+
+		waitErr := <-done
```
rkuris: This should timeout after a few minutes

tests/reexecute/chaos/main.go
```diff
@@ -0,0 +1,218 @@
...
+
+	log.Debug("read VM", zap.Uint64("latest height", lastAcceptedBlock.Height()))
+
+	cmd = createReexecutionCmd(blockDir, currentStateDir, lastAcceptedBlock.Height()+1, endBlock)
+
+	// 5. Restart the reexecution test from the recovered height to verify state consistency
+	r.NoError(cmd.Run())
+}
+
+// openDB attempts to open a LevelDB database with retry logic and exponential backoff.
```
rkuris: nit: This isn't an exponential backoff, it's a linear backoff.

RodrigoVillar: Good call; done: 1893eff

tests/reexecute/chaos/main.go
RodrigoVillar: Added a context here to fail if we exceed the timeout: 257b336

tests/reexecute/chaos/deps.go
RodrigoVillar: Done here: https://github.com/ava-labs/avalanchego/pull/4758

.github/workflows/chaos-test.json
```diff
@@ -0,0 +1,30 @@
+{
```
maru-ava: This is specifically a reexec chaos test, please ensure the naming reflects this unambiguously.

Taskfile.yml
```diff
@@ -117,6 +117,45 @@ tasks:
     cmds:
       - cmd: bash -x ./scripts/copy_dir.sh {{.SRC}} {{.DST}}
 
+  firewood-chaos-test:
```
maru-ava: As per the example of #4761, var indirection as proposed below is best avoided. Instead, please prefer passing args directly so that any changes to the inputs don't require task modification:

```yaml
  test-firewood-chaos:
    desc: ...
    cmd: go run ./tests/reexecute/chaos {{.CLI_ARGS}}
```

.github/workflows/chaos-test.json
```diff
@@ -0,0 +1,30 @@
...
+    "pull_request": {
+        "include": [
+            {
+                "start-block": "101",
```
maru-ava: Also as per the example of #4761, please ensure that configuration required to run a test locally is defined outside of github actions - ideally execution would be something like `task run-test test-name`. That could mean that only `test`, `runner` and `timeout-minutes` would be defined in this file, and given that both the runner and the timeout are static, the value of this file would seem questionable.

.github/workflows/firewood-chaos-test.yml
```diff
@@ -0,0 +1,108 @@
+name: Firewood Chaos Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      start-block:
```
maru-ava, line 10: (No action required) It may make sense to copy-paste these fields directly from one of the existing workflows rather than customizing them and maintaining them separately. A shame there is no built-in templating mechanism for github actions.

maru-ava (commented):
Apart from the implementation considerations mentioned inline, I'm ok merging what you've proposed provided its scope remains limited. Fault injection will be a priority post-monorepo and tools like antithesis and chaos mesh provide a much richer set of capabilities that I would rather not see us attempt to duplicate.

RodrigoVillar:
Will wait for #4761 to be merged in before making any of the requested changes here.

Taskfile.yml
RodrigoVillar: Followed the example of #4761 and moved the logic of `test-firewood-chaos` into it's own script: 00a4852

.github/workflows/chaos-test.json
RodrigoVillar: Done

.github/workflows/firewood-chaos-test.yml
RodrigoVillar: Done - I copied and pasted from `.github/actions/c-chain-reexecution-benchmark/action.yml` while removing fields that weren't relevant to the chaos test.

.github/workflows/chaos-test.json
RodrigoVillar: Also went ahead and removed the JSON file as suggested: 9b24d42

scripts/tests.firewood_chaos.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
```
maru-ava: Would it make sense to augment the cchain test script to support invoking the chaos binary? Duplication in the github action might be unavoidable, but I'm not sure why that needs to extend to scripting.

.github/workflows/firewood-chaos-test.yml
```diff
@@ -0,0 +1,105 @@
+name: Firewood Chaos Test
```
maru-ava, line 1: Maybe use a more specific name (e.g. `firewood-chaos-test.yml`) to avoid ambiguity?

RodrigoVillar: Done: ef82e6a

scripts/tests.firewood_chaos.sh
RodrigoVillar: Good call; extended the reexecution script to enable invocation of the chaos test: c793199

scripts/benchmark_cchain_range.sh
```diff
@@ -150,6 +189,9 @@ elif [[ -z "${BLOCK_DIR:-}" || -z "${CURRENT_STATE_DIR:-}" ]]; then
     echo "  Block range:"
     [[ -n "${START_BLOCK:-}" ]] && echo "    START_BLOCK: ${START_BLOCK}" || echo "    START_BLOCK: (not set)"
     [[ -n "${END_BLOCK:-}" ]] && echo "    END_BLOCK: ${END_BLOCK}" || echo "    END_BLOCK: (not set)"
+    echo "  Timeouts (chaos tests):"
```
maru-ava: (No action required) Maybe only output these if running a chaos test?

scripts/benchmark_cchain_range.sh
```diff
@@ -116,12 +131,36 @@ if [[ -n "$TEST_NAME" ]]; then
...
+            CONFIG="${CONFIG:-firewood-archive}"
+            ;;
         *)
             error "Unknown test '$TEST_NAME'"
             ;;
     esac
 fi
 
+# Detect if this is a chaos test
+IS_CHAOS_TEST=false
```
maru-ava: Rather than true / false, please rely on the bashism of empty output being false, non-empty being true. That allows checking with `-n` and `-z` instead of string comparison and `CHAOS_MODE=1` to enable.

scripts/benchmark_cchain_range.sh
```diff
@@ -116,12 +131,36 @@ if [[ -n "$TEST_NAME" ]]; then
             END_BLOCK="${END_BLOCK:-40000000}"
             CONFIG="${CONFIG:-firewood}"
             ;;
+        chaos-101-250k)
+            BLOCK_DIR_SRC="${BLOCK_DIR_SRC:-cchain-mainnet-blocks-1m-ldb}"
+            CURRENT_STATE_DIR_SRC="${CURRENT_STATE_DIR_SRC:-cchain-current-state-firewood-100}"
+            START_BLOCK="${START_BLOCK:-101}"
+            END_BLOCK="${END_BLOCK:-250000}"
+            MIN_WAIT_TIME="${MIN_WAIT_TIME:-120s}"
```
maru-ava: It doesn't appear that the wait time configuration needs to be defined per test configuration, so maybe just accept them outside of test selection?

scripts/benchmark_cchain_range.sh
```diff
@@ -116,12 +131,36 @@ if [[ -n "$TEST_NAME" ]]; then
             END_BLOCK="${END_BLOCK:-40000000}"
             CONFIG="${CONFIG:-firewood}"
             ;;
+        chaos-101-250k)
```
maru-ava: This test would appear to be duplicating `firewood-101-250k`, maybe that suggests that chaos test configuration could reuse tests available for regular reexecution testing? 

e.g. `CHAOS_MODE=1 MAX_WAIT_TIME=.. MIN_WAIT_TIME=.. ./scripts/benchmark_cchain_range.sh firewood-101-250k`

That could suggest creating `firewood-archive-101-250k` so that it could be invoked as a regular and chaos reexecution test.

scripts/benchmark_cchain_range.sh
```diff
@@ -158,25 +200,52 @@ if [[ -z "${START_BLOCK:-}" || -z "${END_BLOCK:-}" ]]; then
...
+    if [[ -z "${CONFIG:-}" ]]; then
+        error "CONFIG is required for chaos tests"
+    fi
+    if [[ -z "${MIN_WAIT_TIME:-}" || -z "${MAX_WAIT_TIME:-}" ]]; then
+        error "MIN_WAIT_TIME and MAX_WAIT_TIME are required for chaos tests"
+    fi
+fi
+
+if [[ "$IS_CHAOS_TEST" == "true" ]]; then
+    echo "=== Firewood Chaos Test: ${TEST_NAME:-custom} ==="
```
maru-ava, line 197: Maybe avoid duplicating this informational stanza i.e. output it configurably for both chaos and non-chaos outside of the block?

scripts/benchmark_cchain_range.sh
RodrigoVillar: Done

scripts/benchmark_cchain_range.sh
RodrigoVillar: Switching from using true/false to non-empty/empty.

scripts/benchmark_cchain_range.sh
RodrigoVillar: Done

scripts/benchmark_cchain_range.sh
RodrigoVillar: Done

scripts/benchmark_cchain_range.sh
RodrigoVillar: Great call - done.

scripts/benchmark_cchain_range.sh
```diff
@@ -158,25 +186,49 @@ if [[ -z "${START_BLOCK:-}" || -z "${END_BLOCK:-}" ]]; then
     error "START_BLOCK and END_BLOCK are required"
 fi
 
-echo "=== C-Chain Re-execution: ${TEST_NAME:-custom} ==="
+# Chaos tests require additional validation
+if [[ -n "${CHAOS_MODE:-}" ]]; then
+    if [[ -z "${MIN_WAIT_TIME:-}" || -z "${MAX_WAIT_TIME:-}" || -z "${CONFIG:-}" ]]; then
```
maru-ava, line 193: Maybe simplify by including this check in the conditional block below?

scripts/benchmark_cchain_range.sh
```diff
@@ -11,6 +11,8 @@ set -euo pipefail
 # without a test name and without required env vars.
 #
 # Test names configure defaults for S3 sources and block ranges.
+# If running in chaos mode, test names also configure defaults for the VM Config
```
maru-ava, line 14: (No action required) Are only firewood test setups compatible with chaos mode, or can tests that define a non-firewood CONFIG (default or archive)  also be used?

RodrigoVillar: Only Firewood test setups are compatible with chaos mode, as passing in a non-Firewood config (e.g. `default`) will cause the chaos executable to fail (as `firewood` and `firewood-archive` are the only defined config options). In the context of Firewood, I didn't see any additional value in defining non-Firewood options.

scripts/benchmark_cchain_range.sh
RodrigoVillar: Done: c29a101

scripts/benchmark_cchain_range.sh
maru-ava: Would it make sense to document that in the usage message?

RodrigoVillar: The available `CONFIG` options are already documented:

https://github.com/ava-labs/avalanchego/blob/c29a101a1d63e66b384d91d3fa267550d9d9c684/scripts/benchmark_cchain_range.sh#L40

To your point though, I've updated the documentation to make this more explicit: ada0690

PR #4696, opened 2025-12-15
refactor: Share Firewood for EVM

alarso16:
## Why this should be merged

We should only have one Firewood implementation for coreth and subnet-evm, since all behavior is the same between them. This shares with minimal refactoring to ensure that types don't clash.

## How this works

A move and import changes

## How this was tested

CI

No

### DIFF
--- a/graft/evm/firewood/state.go
+++ b/graft/evm/firewood/state.go
@@ -10,29 +10,29 @@ import (
 	"github.com/ava-labs/libevm/core/state"
 )
 
-var _ state.Database = (*firewoodState)(nil)
+var _ state.Database = (*stateAccessor)(nil)
 
-type firewoodState struct {
+type stateAccessor struct {
 	state.Database
 	triedb *TrieDB
 }
 
-func NewStateWrapper(db state.Database, fw *TrieDB) state.Database {
-	return &firewoodState{
+func NewStateAccessor(db state.Database, fw *TrieDB) state.Database {
+	return &stateAccessor{
 		Database: db,
 		triedb:   fw,
 	}
 }
 
 // OpenTrie opens the main account trie.
-func (f *firewoodState) OpenTrie(root common.Hash) (state.Trie, error) {
-	return newAccountTrie(root, f.triedb)
+func (s *stateAccessor) OpenTrie(root common.Hash) (state.Trie, error) {
+	return newAccountTrie(root, s.triedb)
 }
 
 // OpenStorageTrie opens a wrapped version of the account trie.
 //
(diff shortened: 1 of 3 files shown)


graft/evm/firewood/state.go
alarso16, line 1: This is a move from `firewood_database.go`

graft/coreth/triedb/firewood/database.go
```diff
@@ -1,601 +0,0 @@
-// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer, line 1: this is the duplicate coreth version of what is now `evm/firewood/triedb.go`

graft/evm/firewood/account_trie.go
```diff
@@ -17,17 +18,19 @@ import (
...
+//     values, and we thus rely on the `accountTrie`.
 //  2. The `Hash` method actually creates the proposal, since Firewood cannot calculate
 //     the hash of the trie without committing it. It is immediately dropped, and this
 //     can likely be optimized.
 //
 // Note this is not concurrent safe.
-type AccountTrie struct {
-	fw           *Database
+type accountTrie struct {
+	fw           *TrieDB
```
JonathanOppenheimer, line 33: excuse my ignorance, what motivated this change?

graft/coreth/core/extstate/database_test.go
```diff
@@ -68,11 +65,10 @@ type merkleTrie struct {
 func newFuzzState(t *testing.T) *fuzzState {
 	r := require.New(t)
 
-	hashState := NewDatabaseWithConfig(
+	hashState := state.NewDatabaseWithConfig(
 		rawdb.NewMemoryDatabase(),
-		&triedb.Config{
-			DBOverride: hashdb.Defaults.BackendConstructor,
-		})
```
RodrigoVillar, line 75: q: why do we no longer use set the `DBOverride` here?

alarso16: It's not necessary for these tests. The coreth/subnet-evm versions of hashdb have additional support for frequent re-orgs, but that's not necessary for these tests. Furthermore, I really do not want to inject a dependency between coreth and this module

RodrigoVillar: Ah gotcha, thanks for clarifying.

graft/evm/firewood/account_trie.go
alarso16: Clarity between `state.Database` and `triedb.DBOverride`

PR #4697, opened 2025-12-15
perf: Share proposal between hash and commit

alarso16:
## Why this should be merged

This is about a 30% performance improvement for bootstrapping nodes.

## How this works

See ava-labs/libevm#240 for comments from an initial review. All proposals created at hash time are stored in a map, and some shutdown recovery is added for block hashes and heights.

## How this was tested

UT and re-execution of state.

No

### DIFF
--- a/graft/coreth/core/genesis.go
+++ b/graft/coreth/core/genesis.go
@@ -38,7 +38,6 @@ import (
 	"github.com/ava-labs/avalanchego/graft/coreth/params"
 	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/customtypes"
 	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/upgrade/ap3"
-	"github.com/ava-labs/avalanchego/graft/evm/firewood"
 	"github.com/ava-labs/avalanchego/graft/evm/triedb/pathdb"
 	"github.com/ava-labs/avalanchego/vms/evm/acp226"
 	"github.com/ava-labs/libevm/common"
@@ -339,18 +338,8 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	if _, err := statedb.Commit(0, false, stateconf.WithTrieDBUpdateOpts(triedbOpt)); err != nil {
 		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
 	}
-
-	// Firewood requires `Update` and `Commit`, even if the state is empty.
-	_, isFirewood := triedb.Backend().(*firewood.TrieDB)
-	if root == types.EmptyRootHash && isFirewood {
-		// Ensure the Firewood TrieDB is aware of the genesis block.
-		if err := triedb.Update(types.EmptyRootHash, types.EmptyRootHash, 0, nil, nil, triedbOpt); err != nil {
-			panic(fmt.Sprintf("unable to update firewood triedb with genesis block: %v", err))
-		}
-	}
-
 	// Commit newly generated states into disk if it's not empty.
-	if root != types.EmptyRootHash || isFirewood {
+	if root != types.EmptyRootHash {
 		if err := triedb.Commit(root, true); err != nil {
 			panic(fmt.Sprintf("unable to commit genesis block: %v", err))
 		}
--- a/graft/evm/firewood/triedb.go
+++ b/graft/evm/firewood/triedb.go
@@ -294,6 +294,13 @@ func (t *TrieDB) Update(root, parent common.Hash, height uint64, _ *trienode.Mer
 	defer t.proposals.Unlock()
 
 	p, ok := t.possible[possibleKey{parentBlockHash: parentBlockHash, root: root}]
+	// It's possible that we are committing a proposal on top of an empty genesis block in testing.
+	// In this case, we can still find the proposal by looking for the empty block hash
+	if !ok && height == 1 && parent == types.EmptyRootHash {
+		p, ok = t.possible[possibleKey{parentBlockHash: common.Hash{}, root: root}]
+		p.height = 1
+		p.parent.blockHashes[parentBlockHash] = struct{}{}
+	}
 	if !ok {
 		return fmt.Errorf("%w for block %d, root %s, hash %s", errNoProposalFound, height, root.Hex(), blockHash.Hex())
 	}
--- a/graft/subnet-evm/core/genesis.go
+++ b/graft/subnet-evm/core/genesis.go
@@ -34,7 +34,6 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/ava-labs/avalanchego/graft/evm/firewood"
 	"github.com/ava-labs/avalanchego/graft/evm/triedb/pathdb"
 	"github.com/ava-labs/avalanchego/graft/subnet-evm/core/extstate"
 	"github.com/ava-labs/avalanchego/graft/subnet-evm/params"
@@ -383,18 +382,8 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	if _, err := statedb.Commit(0, false, stateconf.WithTrieDBUpdateOpts(triedbOpt)); err != nil {
 		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
 	}
-
-	// Firewood requires `Update` and `Commit`, even if the state is empty.
-	_, isFirewood := triedb.Backend().(*firewood.TrieDB)
-	if root == types.EmptyRootHash && isFirewood {
-		// Ensure the Firewood TrieDB is aware of the genesis block.
-		if err := triedb.Update(types.EmptyRootHash, types.EmptyRootHash, 0, nil, nil, triedbOpt); err != nil {
-			panic(fmt.Sprintf("unable to update firewood triedb with genesis block: %v", err))
-		}
-	}
-
 	// Commit newly generated states into disk if it's not empty.
-	if root != types.EmptyRootHash || isFirewood {
+	if root != types.EmptyRootHash {
 		if err := triedb.Commit(root, true); err != nil {
 			panic(fmt.Sprintf("unable to commit genesis block: %v", err))
 		}


graft/coreth/core/blockchain.go
```diff
@@ -231,12 +231,12 @@ func (c *CacheConfig) triedbConfig() *triedb.Config {
...
 		config.DBOverride = firewood.Config{
-			ChainDataDir:         c.ChainDataDir,
-			CleanCacheSize:       c.TrieCleanLimit * 1024 * 1024,
-			FreeListCacheEntries: firewood.Defaults.FreeListCacheEntries,
-			Revisions:            uint(c.StateHistory), // must be at least 2
-			ReadCacheStrategy:    ffi.CacheAllReads,
-			ArchiveMode:          !c.Pruning,
+			DatabasePath:         c.ChainDataDir,
+			CacheSizeBytes:       uint(c.TrieCleanLimit) * 1024 * 1024,
+			FreeListCacheEntries: 40_000,               // same as default
```
powerslider: Can't we use a constant again here if one is available?

graft/coreth/core/genesis.go
```diff
@@ -338,8 +339,13 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	if _, err := statedb.Commit(0, false, stateconf.WithTrieDBUpdateOpts(triedbOpt)); err != nil {
 		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
 	}
+	if root == types.EmptyRootHash && isFirewood(triedb) {
+		if err := triedb.Update(root, root, 0, nil, nil, triedbOpt); err != nil {
+			panic(fmt.Sprintf("unable to update genesis block in triedb: %v", err))
```
powerslider: Mostly a question: Why is throwing a panic acceptable here? I understand that it is a terminal condition, but maybe we should return a typed error and handle this elsewhere. If it is acceptable according to geth standards to leave it like this, I am ok to not insist on improving this.

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,565 @@ import (
...
+func (*TrieDB) Size() (common.StorageSize, common.StorageSize) {
+	return 0, 0
+}
+
+// Reference is no-op because proposals are only referenced when created.
+// Additionally, internal nodes do not need tracked by consumers.
+func (*TrieDB) Reference(common.Hash, common.Hash) {}
+
+// Dereference is no-op because proposals will be removed automatically.
+// Additionally, internal nodes do not need tracked by consumers.
```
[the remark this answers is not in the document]
alarso16: Seems excessive

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,565 @@ import (
...
+	// in the case of duplicate state roots.
+	// The root of the tree is stored here, and represents the top-most layer on disk.
+	tree *proposal
+	// possible temporarily holds proposals created during a trie update.
+	// This is cleared after the update is complete and the proposals have been sent to the database.
+	possible map[possibleKey]*proposal
+}
+
+type possibleKey struct {
+	parentBlockHash, root common.Hash //nolint:unused // It is used as a map key
```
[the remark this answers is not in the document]
alarso16: The key is never explicitly read, the linter is wrong.

rkuris (approved):

graft/coreth/core/genesis.go
```diff
@@ -338,8 +339,13 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	if _, err := statedb.Commit(0, false, stateconf.WithTrieDBUpdateOpts(triedbOpt)); err != nil {
 		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
 	}
+	if root == types.EmptyRootHash && isFirewood(triedb) {
```
JonathanOppenheimer: Can you explain the empty genesis case a little more? This isn't the case for hashdb or pathdb and as I understand it, this is because they only track state roots, while firewood tracks state roots and block hashes.

Sure this is clear for others, just curious

graft/evm/firewood/recovery.go
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: Is it possible to test this code? units tests for things like                                                                                                                                                   

- multiple children with same state root
- other edge cases like this 
- concurrent operations

could be useful

graft/coreth/core/blockchain.go
JonathanOppenheimer: is there a reason you changed this to an embedded constant -- I think having it has a default is better.

JonathanOppenheimer (commented):
You know this code better than I do

graft/coreth/core/genesis.go
alarso16: It's definitely not obvious, it took me a bit to figure out exactly what was wrong. We need to know the block hash, otherwise I have to do some janky stuff like "how do I find the parent of an empty block?". There's still corresponding block hashes, so it's kind of necessary to tell it for tracking.

Honestly, handling an empty genesis block is really strange anyway. If there's nothing in the genesis block, then all blocks afterward will always be empty, which is super weird. The only reason I need this is at all is to make it pass the tests, which isn't a very good reason

There's two options here:
1. Add a comment clarifying why it's necessary at all
2. Remove this, and remove all tests that make an empty genesis block, since that will never happen in production

graft/evm/firewood/recovery.go
alarso16: It is possible, I have started working on some tests, but really what we rely on for testing all of this is the blockchain tests. They do cover all of those things you are suggesting, but it's not great that we rely on the integration tests to ensure it works. Makes tracking down errors pretty tricky.

One thing I talked with @michaelkaplan13 about is that since most of this code will not be used in production, it might be better to only spend time guaranteeing the post-SAE behavior (no proposal tree). Any thoughts on this?

graft/coreth/core/blockchain.go
alarso16: This was an @ARR4N suggestion. Because creating a config without a database path should return an error, having a "default config" isn't actually logical, since you always have to alter it

graft/coreth/core/genesis.go
JonathanOppenheimer: Ah thanks for the explanation! If this isn't something that will ever happen in production, I don't think we should modify our prod code to get some tests to pass, so I'd vote for option 2? Up to your discretion

graft/coreth/core/blockchain.go
JonathanOppenheimer: makes sense

graft/evm/firewood/recovery.go
JonathanOppenheimer: No that makes a lot of sense -- if this isn't going to be used in prodw, it would be a time waster. I do agree that yeah without unit tests it makes it harder to track down errors. Maybe just have some sort of tracker to make sure that we do guarantee the post SAE behavior?

Elvis339:
Hey Austin

Found a bug while running C-Chain re-execution benchmarks to measure the performance impact of your proposal-sharing changes.

The PR CI passes because it runs with `config: "default"` (hashdb) using a hashdb snapshot. It never exercises the Firewood code path.

When running with `config: "firewood"` starting from a restored Firewood snapshot at block 10M, it fails immediately on the first block https://github.com/ava-labs/avalanchego/actions/runs/20661797719/job/59326295398#step:4:831:
```
failed to verify block at height 10000001: no proposal found for block 10000001,
root 0x5f7138331cc9dd9d47cd78fb5bf632b0961a28d29d394f19e6d32f2a9e1f4266,
hash 0x90649defd375c9784a95bb01e752c2a06341213c59e25dedfe8aee7a7e8b8d3d
```

The proposal-sharing logic expects `Hash()` to populate `t.possible` before `Update()` is called. When loading from a Firewood snapshot and executing the first block, something in the initialization path skips `Hash()` so `t.possible` is empty when `Update()` runs.

To reproduce:
```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
    --repo ava-labs/avalanchego \
    --ref es/candidate-firewood \
    -f block-dir-src="cchain-mainnet-blocks-10m-20m-ldb" \
    -f current-state-dir-src="cchain-current-state-firewood-10m" \
    -f config="firewood" \
    -f start-block="10000001" \
    -f end-block="10050000"
```


To save you time digging through git history: `es/candidate-firewood` is your PR merged on top of `es/baseline-firewood`, which contains master plus benchmarking infrastructure (pprof, FFI metrics, cross-stack profiling). The baseline branch with the same Firewood config and snapshot works fine. https://github.com/ava-labs/avalanchego/actions/runs/20661268516/job/59324164005

Let me know if you need help reproducing or testing a fix.

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,565 @@ import (
...
 		}
-	} else {
-		// Find any proposal with the given parent root.
-		// Since we are only using the proposal to find the root hash,
-		// we can use the first proposal found.
-		proposals, ok := t.proposalMap[parentRoot]
-		if !ok || len(proposals) == 0 {
-			return common.Hash{}, fmt.Errorf("firewood: no proposal found for parent root %s", parentRoot.Hex())
+		for parentHash := range t.proposals.tree.blockHashes {
+			t.possible[possibleKey{parentBlockHash: parentHash, root: root}] = p
```
Elvis339: This looks like a potential data race.

`createProposals` takes `RLock` but writes to `t.possible`. If two goroutines call `Hash()` concurrently on different `accountTries` sharing the same TrieDB, both would hold read locks while writing to the same map.

alarso16:
> Found a bug while running C-Chain re-execution benchmarks to measure the performance impact of your proposal-sharing changes.

This isn't exactly a bug... I added specific recovery logic in this PR. On `master`, when the database opens, we set the `height` of the trie to 0, and use this as a special value, indicating either the genesis block is committed or the database was just opened.

This PR makes it safer by tracking the most recently committed height/block hash in leveldb, so upon recovery, we can properly populate these fields in the proposal trie. However, your snapshot does not have that, and I didn't realize this would be a breaking change.

Now that you bring this up, I do think this will also be an issue for statesync as well. Let me rework a few things to see if I can find a less fragile way of crash recovery

alarso16:
After some investigation, I've discovered a couple ideas:

1. Leave it as it is. Really, this should have already been included in Firewood, but wasn't purely necessary. For this performance change, it practically is necessary. For statesync, we can just add another method that would be called in `core.BlockChain.ResetToStateSyncedBlock`. This would require re-generating the re-execution state
2. Find some hacky way to use `common.Hash{}` and `0` as magic values for block hash and height, respectively. Although this would work, the code would be much harder to reason about and is pretty bug prone. Really, the only advantage is that it doesn't corrupt the state for the re-execution tests and doesn't expand the state.
3. Use `core.BlockChain` state to populate these fields independently of the Firewood state. This would avoid your issue and remove a race if there's an unclean shutdown (Firewood committed, trackers not committed). This just requires adding another method that must be called in `NewBlockChain`, making Firewood difficult to use outside of this (like when we test `TrieDB` independently). 

I'm leaning towards option 3.

Elvis339:
> After some investigation, I've discovered a couple ideas:
> 
> 1. Leave it as it is. Really, this should have already been included in Firewood, but wasn't purely necessary. For this performance change, it practically is necessary. For statesync, we can just add another method that would be called in `core.BlockChain.ResetToStateSyncedBlock`. This would require re-generating the re-execution state
> 2. Find some hacky way to use `common.Hash{}` and `0` as magic values for block hash and height, respectively. Although this would work, the code would be much harder to reason about and is pretty bug prone. Really, the only advantage is that it doesn't corrupt the state for the re-execution tests and doesn't expand the state.
> 3. Use `core.BlockChain` state to populate these fields independently of the Firewood state. This would avoid your issue and remove a race if there's an unclean shutdown (Firewood committed, trackers not committed). This just requires adding another method that must be called in `NewBlockChain`, making Firewood difficult to use outside of this (like when we test `TrieDB` independently).
> 
> I'm leaning towards option 3.

I see the issue. `proposalTree.Block == 0` is a magic value that lets the first proposal skip the parent hash check. Your PR stores height/blockHashes to leveldb instead, but my snapshot doesn't have that data which breaks in initialization path.

It does seem like Option 3 is cleanest. BlockChain already knows the committed height/hash after loadLastState(), no need to duplicate it in leveldb (and risk a two-phase commit race).

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,574 @@ import (
...
+		count int         // number of proposals created.
+		root  common.Hash // The resulting root hash, should match between proposals
 	)
-	start := time.Now()
-	if t.proposalTree.Root == parentRoot {
+	if t.proposals.tree.root == parentRoot {
 		// Propose from the database root.
-		p, err = t.fwDisk.Propose(keys, values)
+		p, err := t.createProposal(t.proposals.tree, keys, values)
+		root = p.root
```
Elvis339: If createProposal returns (nil, err), accessing p.root will panic.

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,574 @@ import (
...
+		// If the block hash is already tracked, we can skip proposing this again.
+		if _, ok := p.blockHashes[block]; ok {
+			log.Debug("proposal already exists", "root", root.Hex(), "parentBlock", parentBlock.Hex(), "block", block.Hex())
+			return true
+		}
+
+		// We have an identical proposal, but should ensure the hash is tracked with this proposal.
+		if _, ok := p.parent.blockHashes[parentBlock]; ok {
+			log.Debug("proposal already exists, updating hash", "root", root.Hex(), "parentBlock", parentBlock.Hex(), "block", block.Hex())
+			p.blockHashes[block] = struct{}{}
```
Elvis339, line 353: No action required: function is named `exists` indicating pure function but it mutates `p.blockHashes`. If you agree I think it makes sense to rename to `existsOrTrack` or move the mutation to the caller. Up to you.

graft/evm/firewood/triedb.go
```diff
@@ -29,566 +29,574 @@ import (
...
+	Firewood *ffi.Database
+
+	// kvStore is used for storing recovery information.
+	kvStore ethdb.Database
+
+	// possible temporarily holds proposals created during a trie update.
+	// This is cleared after the update is complete and the proposals have been sent to the database.
+	// It's unexpected for mulitple updates to this to occur simultaneously, but a lock is used to ensure safety.
+	possible     map[possibleKey]*proposal
+	possibleLock sync.Mutex
```
Elvis339: Having two locks for related state increases cognitive load.

Curious about the decision to separate possibleLock from proposals.RWMutex would embedding possible in the proposals struct with a single lock work here? Or is there a concurrent access pattern I'm missing that requires the separation?

Elvis339 (commented):
I profiled master on C-Chain re-execution (blocks 10M-10.05M):
Commit: 18.5% of CPU
(*Database).Propose: 1.76% of CPU
The proposed optimization removes one of two proposal creations, so max improvement is ~0.9%.
Where did the 30% improvement come from in your testing? Different workload? Or maybe the win is in memory, not CPU?

Just want to understand where to look.

To reproduce download pprof artifacts: https://github.com/ava-labs/avalanchego/actions/runs/20661268516/artifacts/5036934243
Run: `go tool pprof -top pprof-profiles/cpu.prof | grep -E "(firewood|ffi.*Propose)" `

graft/evm/firewood/triedb.go
alarso16: There's three entrypoints essentially:
1. On `block.Verify`, we will execute the transactions and hash (`getProposalHash`), and assuming the hashes look right, we will call `Update`)
2. On `block.Accept`, we queue the state change to asynchronously call `Commit`.
3. If we don't properly block the APIs that try to access `StateDB.DumpToCollector`, `getProposalHash` may be called unintentionally.

The concern is between 3 and 1/2. It would be possible to DoS the execution through an API, since it will hold onto the lock while trying to hash the trie

graft/evm/firewood/triedb.go
alarso16: I think I'd prefer to leave it. The proposal does exist, and if we alter the `proposals` state, then we still drop the proposal we are holding on to.

However, if this is a readability issue, I'm happy to change it. Any other opinions?

graft/evm/firewood/triedb.go
alarso16: Ah good catch

graft/evm/firewood/triedb.go
Elvis339: Cool, then I think renaming function from `exists` to `existsOrTrack` (or something else) clearly communicates the intent.

graft/evm/firewood/triedb.go
Elvis339: Got it thanks.

alarso16:
> After some investigation, I've discovered a couple ideas

I made a draft showing how number 3 could be implemented: #4835. It's definitely messier, but wouldn't break state. However, I do think that we will be making a breaking change to firewood state before the v0.1.0 release

graft/coreth/core/genesis.go
alarso16: I found a cleaner way, so this should be good

graft/evm/firewood/triedb.go
```diff
@@ -133,53 +147,68 @@ func New(config TrieDBConfig) (*TrieDB, error) {
 
 	fw, err := ffi.New(path, options...)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("opening firewood database: %w", err)
 	}
 
-	currentRoot, err := fw.Root()
+	intialRoot, err := fw.Root()
```
JonathanOppenheimer:
```suggestion
	initialRoot, err := fw.Root()
```

graft/evm/firewood/triedb.go
```diff
@@ -88,39 +117,24 @@ func DefaultConfig(dir string) TrieDBConfig {
...
 }
 
+// BackendConstructor implements the [triedb.DBConstructor] interface.
+// It creates a new Firewood database with the given configuration.
+// Any error during creation will cause the program to exit.
 func (c TrieDBConfig) BackendConstructor(ethdb.Database) triedb.DBOverride {
 	db, err := New(c)
 	if err != nil {
-		log.Crit("firewood: error creating database", "error", err)
+		log.Crit("firewood: creating database", "error", err)
```
JonathanOppenheimer: this is sort of a nit, but you removed the `firewood:` prefix in some places, but not other. all the error messages in this PR could probably use a pass for consistency.

graft/evm/firewood/triedb.go
```diff
@@ -188,420 +217,380 @@ func (*TrieDB) Scheme() string {
...
-// proposed from the given parent root.
-func (t *TrieDB) getProposalHash(parentRoot common.Hash, keys, values [][]byte) (common.Hash, error) {
-	// This function only reads from existing tracked proposals, so we can use a read lock.
-	t.proposalLock.RLock()
-	defer t.proposalLock.RUnlock()
+	// Must prevent a simultaneous `Commit`, as it alters the proposal tree/disk state.
+	t.proposals.RLock()
+	defer t.proposals.RUnlock()
+	t.possibleLock.Lock()
+	defer t.possibleLock.Unlock()
```
JonathanOppenheimer: this locking seems kinda sus -- is this safe? -- can you check if a proposal already exists or does that not make sense

graft/evm/firewood/triedb.go
```diff
@@ -133,53 +147,68 @@ func New(config TrieDBConfig) (*TrieDB, error) {
...
 	}
 
-	// Check that the directory exists
-	dir := filepath.Dir(path)
 	switch info, err := os.Stat(dir); {
 	case os.IsNotExist(err):
 		log.Info("Database directory not found, creating", "path", dir)
-		if err := os.MkdirAll(dir, 0o755); err != nil {
-			return fmt.Errorf("error creating database directory: %w", err)
+		if err := os.MkdirAll(dir, 0o750); err != nil {
```
JonathanOppenheimer: isn't this a breaking change

graft/evm/firewood/triedb.go
```diff
@@ -29,38 +29,67 @@ import (
...
-// ProposalContext represents a proposal in the Firewood database.
-// This tracks all outstanding proposals to allow dereferencing upon commit.
-type ProposalContext struct {
-	Proposal *ffi.Proposal
-	Hashes   map[common.Hash]struct{} // All corresponding block hashes
-	Root     common.Hash
-	Block    uint64
-	Parent   *ProposalContext
-	Children []*ProposalContext
+type possibleKey struct {
```
JonathanOppenheimer, line 74: it seems kinda weird to declare a `struct` that's never used, just for a key. does adding something like this provide value or nah:

```
func newPossibleKey(parentBlockHash, root common.Hash) possibleKey {
  return possibleKey{parentBlockHash: parentBlockHash, root: root}
}
```

graft/evm/firewood/triedb.go
```diff
@@ -188,420 +217,380 @@ func (*TrieDB) Scheme() string {
...
+	t.proposals.byStateRoot[root] = append(t.proposals.byStateRoot[root], p)
+	p.blockHashes[blockHash] = struct{}{}
 
 	return nil
 }
 
+// Check if this proposal already existsOrTrack.
+// During reorgs, we may have already tracked this block hash.
+// Additionally, we may have coincidentally created an identical proposal with a different block hash.
+func (ps *proposals) existsOrTrack(root, block, parentBlock common.Hash) bool {
```
JonathanOppenheimer: This doesn't seem to track anything unless I'm misunderstanding. unless you're saying because it's already track_**ed**_ it already exists? I think this could be renamed

graft/evm/firewood/triedb.go
```diff
@@ -29,38 +29,67 @@ import (
...
+
+	// The underlying Firewood database, used for storing proposals and revisions.
+	// This is exported as read-only, with knowledge that the consumer will not close it
+	// and the latest state can be modified at any time.
+	Firewood *ffi.Database
+
+	// possible temporarily holds proposals created during a trie update.
+	// This is cleared after the update is complete and the proposals have been sent to the database.
+	// It's unexpected for multiple updates to this to occur simultaneously, but a lock is used to ensure safety.
+	possible     map[possibleKey]*proposal
```
JonathanOppenheimer: should there be a max size before this gets cleared?

graft/evm/firewood/triedb.go
```diff
@@ -133,53 +147,68 @@ func New(config TrieDBConfig) (*TrieDB, error) {
...
+			},
 		},
+		possible: make(map[possibleKey]*proposal),
 	}, nil
 }
 
-func validatePath(path string) error {
-	if path == "" {
-		return errors.New("firewood database file path must be set")
+func validateDir(dir string) error {
```
JonathanOppenheimer, line 182: go doc?

JonathanOppenheimer:
I'm not sure about test coverage for this -- I would assume this refactor would introduce new things to test -- some ideas include -- like if `Hash()` gets called concurrently, or if any of the functions (e.g. `Update()`) are called in correctly.

graft/evm/firewood/triedb.go
alarso16: It was an attempt at optimization, but I don't think it works. I'll simplify for now

graft/evm/firewood/triedb.go
alarso16: Elvis recommended this name. I'll change this back to `exists`

graft/evm/firewood/triedb.go
alarso16: It's named for clarity, but It's being used correctly. The linter (I believe) is just wrong this time

graft/evm/firewood/triedb.go
alarso16: oops

graft/evm/firewood/triedb.go
alarso16: I meant to remove it everywhere, should be better now

graft/evm/firewood/triedb.go
alarso16: What should the max size be? There only ever should be a few, but I don't want to draw a line.

JonathanOppenheimer: fair!

graft/evm/firewood/triedb.go
JonathanOppenheimer: got it

alarso16:
These tests are not complete, but I do have an outstanding issue for guaranteeing minimum behavior. This just ensure that the basics are laced together correctly, and the specific multiple hashing thing you suggested

JonathanOppenheimer:
> These tests are not complete, but I do have an outstanding issue for guaranteeing minimum behavior. This just ensure that the basics are laced together correctly, and the specific multiple hashing thing you suggested

fire - they can be built upon!!

graft/evm/firewood/triedb.go
```diff
@@ -29,38 +29,68 @@ import (
...
+	errNoProposalFound = errors.New("no proposal found")
 )
 
-type proposable interface {
-	// Propose creates a new proposal from the current state with the given keys and values.
-	Propose(keys, values [][]byte) (*ffi.Proposal, error)
+// TrieDB is a triedb.DBOverride implementation backed by Firewood.
+// It acts as a HashDB for backwards compatibility with most of the blockchain code.
+type TrieDB struct {
+	proposals
```
powerslider, line 50:
```suggestion
    proposalTree proposals
```
Reasoning: https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs

graft/evm/firewood/triedb.go
```diff
@@ -188,420 +219,374 @@ func (*TrieDB) Scheme() string {
...
-		return fmt.Errorf("firewood: parent root %s does not match proposal tree root %s for root %s at height %d", parentRoot.Hex(), t.proposalTree.Root.Hex(), root.Hex(), block)
+	// The rest of the operations except key-value arranging must occur with a lock
+	t.proposals.Lock()
+	defer t.proposals.Unlock()
+
+	p, ok := t.possible[possibleKey{parentBlockHash: parentBlockHash, root: root}]
+	// Now, all unused proposals have no other references, since we didn't store them
+	// in the proposal map or tree, so they will be garbage collected.
+	// Any proposals with a different root were mistakenly created, so they can be freed as well.
+	clear(t.possible)
```
powerslider: ok, this one might be a stretch, but here is what I notice: If `Update` fails for any transient reason (missing payload, wrong root/height, or a race), all proposals created in the preceding `Hash` calls are discarded and cannot be recovered. This makes error handling brittle and can cause cascading failures after a single bad update call.

A possible solution could be to remove the `clear` from here and:

-  to clear proposals since we are already tracking a matching one: https://github.com/ava-labs/avalanchego/pull/4697/changes#r2699054751 
-  and for the path of not finding a matching one you could clear after that those that have a different root: https://github.com/ava-labs/avalanchego/pull/4697/changes#r2699076247

graft/evm/firewood/triedb.go
```diff
@@ -188,420 +219,374 @@ func (*TrieDB) Scheme() string {
...
 	}
 
-	log.Debug("firewood: proposing from database root", "root", root.Hex(), "height", block)
-	p, err := createProposal(t.Firewood, root, keys, values)
-	if err != nil {
-		return err
+	// If we have already created an identical proposal, we can skip adding it again.
+	if t.proposals.exists(root, blockHash, parentBlockHash) {
+		return nil
 	}
```
powerslider, line 313:
```suggestion
	if t.proposals.exists(root, blockHash, parentBlockHash) {
		// All unused proposals can now be cleared since we already tracked a matching one.
		clear(t.possible)
		return nil
	}
```

graft/evm/firewood/triedb.go
```diff
@@ -188,420 +219,374 @@ func (*TrieDB) Scheme() string {
...
+	case p.height != height:
+		return fmt.Errorf("height mismatch, expected %d, got %d", height, p.height)
 	}
-	t.proposalMap[root] = append(t.proposalMap[root], pCtx)
-	t.proposalTree.Children = append(t.proposalTree.Children, pCtx)
+
+	// Track the proposal context in the tree and map.
+	p.parent.children = append(p.parent.children, p.proposalMeta)
+	t.proposals.byStateRoot[root] = append(t.proposals.byStateRoot[root], p)
+	p.blockHashes[blockHash] = struct{}{}
```
powerslider, line 326:
```suggestion
	p.blockHashes[blockHash] = struct{}{}
	// Now all unused proposals have no other references, 
	// since we didn't store them in the proposal map or tree and any proposals 
	// with a different root can be freed.
	clear(t.possible)
```

graft/coreth/core/genesis.go
```diff
@@ -343,6 +344,10 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 		if err := triedb.Commit(root, true); err != nil {
 			panic(fmt.Sprintf("unable to commit genesis block: %v", err))
 		}
+	} else {
+		if t, ok := triedb.Backend().(*firewood.TrieDB); ok {
+			t.SetHashAndHeight(block.Hash(), 0)
+		}
```
ceyonur: I'm confused, why do we call this here only if the root is empty? Shouldn't we call this nevertheless?

graft/coreth/core/genesis.go
alarso16: I removed this

graft/coreth/core/blockchain.go
alarso16: This isn't a part of this refactor (already merged ;))

graft/evm/firewood/triedb.go
alarso16: It says to avoid embedded public fields specifically. Does it make any difference since it's private?

graft/evm/firewood/triedb.go
alarso16: Seems reasonable, I added a test to check this too

graft/coreth/core/genesis.go
alarso16: We could, but it's not actually necessary. What this is doing is setting the block hash and height for internal tracking for the diamond verified block case. If the genesis `root == types.EmptyRootHash`, the block hash isn't `common.Hash{}`, but we don't actually need to "Commit", since there's no key value changes. We don't need to in the normal (non-empty) genesis case because we assume that the first block hash will be `common.Hash{}` and the first height will be 0.

The real reason this is designed is for the two interesting cases:
1. Startup recovery. If the node was turned off and on, we need some way to load this state
2. State sync finish. For the first block after state sync finishes, we won't know the block height or hash

The other option for this specifically is that we could still `triedb.Commit` in this case, but I thought this was more intuitive. If you disagree, I can change it to something like this.

```go
_, isFirewood := triedb.Backend().(*firewood.TrieDB)
if root != types.EmptyRootHash || isFirewood {
	if err := triedb.Commit(root, true); err != nil {
		panic(fmt.Sprintf("unable to commit genesis block: %v", err))
	}
}
```

ceyonur: Isn't this only useful once we commit the triedb? If the `root == types.EmptyRootHash ` we don't commit anything. I'm also confused with:
>If the genesis root == types.EmptyRootHash, the block hash isn't common.Hash{}

I don't think we ever have an empty block/header hash for genesis.

alarso16: > Isn't this only useful once we commit the triedb? If the root == types.EmptyRootHash  we don't commit anything.

We don't commit any key-value changes, but I need to somehow inform the `TrieDB` that the block hash of the most recently committed block isn't `common.Hash{}`, so we can accurately parent the following blocks.

> I don't think we ever have an empty block/header hash for genesis.

The genesis block's parent hash is `common.Hash{}`, so we kind of do have a block with the empty hash.

graft/evm/firewood/triedb.go
alarso16: Discussed offline, we agree to disagree since `proposals` is also an appropriate field name. Used cautiously, but I can be persuaded

graft/coreth/core/genesis.go
ceyonur: >The genesis block's parent hash is common.Hash{}, so we kind of do have a block with the empty hash.

It's because genesis block does not have any parent, but it does not imply the genesis block hash is `common.Hash{}`

alarso16: Discussed offline - this usage is unintuitive, so I will doc some details about when it is appropriate and switch to committing instead

graft/coreth/core/genesis.go
```diff
@@ -338,8 +339,18 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	if _, err := statedb.Commit(0, false, stateconf.WithTrieDBUpdateOpts(triedbOpt)); err != nil {
 		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
 	}
+
+	// Firewood requires `Update` and `Commit`, even if the state is empty.
```
ceyonur: could we provide a bit more for why it requires?

alarso16: I deleted it. Now it's abstracted in the narrowest special case to get the empty block tests to pass

PR #4740, opened 2025-12-15
test: use generated ABI

JonathanOppenheimer:
## Why this should be merged

Currently, the committed ABI files are 2 years old, and there is no clear path of how they were generated. This PR deletes those files, and commits the files generated by `compile.go` instead, establishing a source of truth for the generated `.abi` files. Additionally the freshness of these files is added to and checked by `check-generate-bindings` which run as part of CI. 

## How this was tested
CI

## Need to be documented?
No

No

### DIFF
--- a/graft/subnet-evm/Taskfile.yml
+++ b/graft/subnet-evm/Taskfile.yml
@@ -93,9 +93,8 @@ tasks:
   generate-bindings:
     desc: Generates Go bindings from Solidity contracts in precompile
     cmds:
-      - cmd: |
-          # Find and remove generated binding files in precompile directories only
-          find ./precompile -type f -name 'gen_*binding.go' -exec grep -l '^// Code generated - DO NOT EDIT\.$' {} \; | xargs -r rm
+      - cmd: find ./precompile -type f -name 'gen_*binding.go' -exec grep -l '^// Code generated - DO NOT EDIT\.$' {} \; | xargs -r rm
+      - cmd: find ./precompile -type f -name '*.abi' | xargs -r rm
       - cmd: go generate ./precompile/...
 
   lint:


JonathanOppenheimer:
See this PR for further context: https://github.com/ava-labs/subnet-evm/pull/1912

powerslider (approved):

maru-ava:
Would it make sense to add a task that generates these files and another for validating that the files are up-to-date, as per the example of other generated files? It's not clear to me what the value of doing a one-time update would be that is not validated by CI. What am I missing?

JonathanOppenheimer:
> Would it make sense to add a task that generates these files and another for validating that the files are up-to-date, as per the example of other generated files? It's not clear to me what the value of doing a one-time update would be that is not validated by CI. What am I missing?

This is actually already checked but admittedly not clear through just the text of this PR. The `check-generate-bindings` job does this:

1. `go generate ./precompile/...` runs all go:generate directives, including the solc ones that produce `.abi` files
2.  check-clean-branch checks if any files changed in git (which includes the `.abi` files)

The only difference is that `.abi` files aren't deleted before regeneration like the `gen_*binding.go` files are, as they're directly overwritten with solc `--overwrite`

JonathanOppenheimer:
I will add explicit deletion to catch orphans.

JonathanOppenheimer:
> Thank you for the generation cleanup. I guess this begs the question of why _all_ of our `go generate` tasks aren't deleting before generating?

They absolutely should be -- if they were, this issue would have been caught a long time ago, as these `.abi` files have been orphaned for 2+ years. I will file an issue for this.

JonathanOppenheimer:
Issue: https://github.com/ava-labs/avalanchego/issues/4778

PR #4743, opened 2025-12-16
debug: Do not merge

rkuris:
## Why this should be merged

Adds logging which confirms that the data is the same while debugging a test case. Not for general use, but we might be able to lower the logging level to trace or something for a more permanent debugging fixture, as this did prove that the data was the same.


JonathanOppenheimer:
@rkuris feel free to reopen

PR #4755, opened 2025-12-16
Remove `*VM` from `AvaxAPI`

StephenButtolph:
## Why this should be merged

The AvaxAPI, while gross, is still needed with SAE. However, SAE won't have access to the `vm.VM` struct. Therefore, 

## How this works

Removes `vm *VM` and replaces it with its inner pieces.

The only non-trivial change is around allowing a `nil` blockchain. In SAE, we will not be using `core.Blockchain`.

## How this was tested

CI

No.


graft/coreth/plugin/evm/atomic/vm/api.go
```diff
@@ -34,11 +41,26 @@ var (
...
 // AvaxAPI offers Avalanche network related API methods
-type AvaxAPI struct{ vm *VM }
+type AvaxAPI struct {
+	// If non-nil, bc is used to prevent returning any transactions as accepted
+	// prior to having advanced the chain state to the block containing the
+	// transaction.
+	bc *core.BlockChain
+
+	Context      *snow.Context
+	Mempool      *txpool.Mempool
```
JonathanOppenheimer, line 51: This was indirectly renamed from atomic mempool, and I looked to see if there were any mempools which are not atomic, and I learned that the p and x chain mempools are not atomic. I don't really have anything to say here, as I think changing the name back to atomic mempool is unnecessary

graft/coreth/plugin/evm/atomic/vm/api.go
```diff
@@ -196,41 +213,58 @@ type FormattedTx struct {
...
-		lastAccepted := service.vm.InnerVM.Ethereum().BlockChain().LastAcceptedBlock()
-		if height > lastAccepted.NumberU64() {
-			return nil
+	reply.BlockHeight = height
+	return err
+}
+
+// getAtomicTx returns the requested transaction, status, and height.
+// If the status is [atomic.Unknown], then the returned transaction will be nil.
+func (service *AvaxAPI) getAtomicTx(txID ids.ID) (*atomic.Tx, atomic.Status, *json.Uint64, error) {
```
JonathanOppenheimer, line 246: this is moved from `graft/coreth/plugin/evm/atomic/vm/vm.go`

michaelkaplan13 (approved):
Changes LGTM.

Will be nice to have `strevm` included in `avalanchego` in the future to further reduce dependent cross-repo PRs.

graft/coreth/plugin/evm/atomic/vm/api.go
```diff
@@ -155,38 +182,28 @@ func (service *AvaxAPI) IssueTx(_ *http.Request, args *api.FormattedTx, response
 	// we push it to the network for inclusion. If the tx was previously added
 	// to the mempool through p2p gossip, this will ensure this node also pushes
 	// it to the network.
-	service.vm.AtomicTxPushGossiper.Add(tx)
+	service.PushGossiper.Add(tx)
```
JonathanOppenheimer, line 185:
```suggestion
        if service.PushGossiper != nil {
	    service.PushGossiper.Add(tx)
        }
```

JonathanOppenheimer: Stephen explained to me in the office why this is not correct, as the push gossiper will not be initialized later so this will never walk.

PR #4758, opened 2025-12-17
refactor(reexecute): export NewMainnetCChainVM() [1/2]

RodrigoVillar:
## Why this should be merged

A prerequiste of #4695 is that we need to create a Coreth instance to read the last accepted block. We can do this with `newCChainMainnetVM()`, but this is currently unexported and lives in the `reexecute/c` package.

## How this works

Exports `NewMainnetCChainVM()` and moves it to `reexecute` for usage by `reexecute/c` and the future `reexecute/chaos` packages.

## How this was tested

CI

No

### DIFF
--- a/tests/reexecute/c/vm_reexecute.go
+++ b/tests/reexecute/c/vm_reexecute.go
@@ -22,35 +22,15 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/avalanchego/api/metrics"
-	"github.com/ava-labs/avalanchego/chains/atomic"
-	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/database/leveldb"
-	"github.com/ava-labs/avalanchego/database/prefixdb"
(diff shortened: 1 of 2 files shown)


tests/reexecute/c/vm_reexecute.go
```diff
@@ -289,87 +269,6 @@ func benchmarkReexecuteRange(
 	}
 }
 
-func newMainnetCChainVM(
```
maru-ava, line 292: Why does this have to move to be exported?

RodrigoVillar: This function lives in an executable package (`main`) - functions in executable packages can't be exported for reuse: 

https://github.com/ava-labs/avalanchego/blob/0aaf6821de3d8eae69c406a7db5568cf45d9096b/tests/reexecute/c/vm_reexecute.go#L1-L4

maru-ava (approved):

alarso16 (approved):

PR #4761, opened 2025-12-17
[ci] Migrate reexec tasks to a script to simplify maintenance

maru-ava:
## Why this should be merged

Doing anything non-trivial in tasks is not recommended. This change moves the reexec logic to a new script for ease of maintenance and to simplify usage.

## How this was tested

- [x] CI

N/A

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -7,7 +7,8 @@ set -euo pipefail
 # Usage:
 #   ./benchmark_cchain_range.sh [test-name]
 #
-# Running without arguments will output available tests.
+# To see available tests: use `help` as the test name or invoke
+# without a test name and without required env vars.
 #
 # Test names configure defaults for S3 sources and block ranges.
 # All defaults can be overridden via environment variables.
@@ -20,7 +21,7 @@ set -euo pipefail
 #     CURRENT_STATE_DIR: Path to local current state directory.
 #
 #   Required:
-#     START_BLOCK: The starting block height (exclusive).
+#     START_BLOCK: The starting block height (inclusive).
 #     END_BLOCK: The ending block height (inclusive).
 #
 #   Optional:
--- a/tests/reexecute/c/README.md
+++ b/tests/reexecute/c/README.md
@@ -2,6 +2,8 @@
 
 The C-Chain benchmarks support re-executing a range of mainnet C-Chain blocks against a provided snapshot of the current state as of some initial state.
 
+Benchmarks can be invoked by running `./scripts/run_task.sh test-cchain-reexecution` or `./scripts/benchmark_cchain_range.sh`.
+
 ## Prerequisites
 
 Configuring your AvalancheGo dev environment requires:


.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -50,10 +50,6 @@ inputs:
     description: 'The password for the Prometheus instance.'
     required: true
     default: ''
-  workspace:
```
maru-ava, line 53: I noticed this was unused.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -126,54 +122,30 @@ runs:
         role-to-assume: ${{ inputs.aws-role }}
         aws-region: ${{ inputs.aws-region }}
         role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
-    - name: Validate inputs
```
maru-ava, line 129: Validation moved to the script so its defined once for use locally and in CI and it can be subject to shellcheck linting.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -126,54 +122,30 @@ runs:
...
+    - name: Set benchmark output file
       shell: bash
-      run: |
-        TIMESTAMP=$(date '+%Y%m%d-%H%M%S')
-        echo "EXECUTION_DATA_DIR=/tmp/reexecution-data-${TIMESTAMP}" >> "$GITHUB_ENV"
-        echo "BENCHMARK_OUTPUT_FILE=${GITHUB_WORKSPACE}/benchmark-output.json" >> "$GITHUB_ENV"
+      run: echo "BENCHMARK_OUTPUT_FILE=${GITHUB_WORKSPACE}/benchmark-output.json" >> "$GITHUB_ENV"
     - name: Run C-Chain Re-execution Benchmark
       shell: nix develop --impure --command bash -x {0}
-      run: |
```
maru-ava, line 153: Moved this configuration to the script

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -126,54 +122,30 @@ runs:
...
-
-          if [[ ${#missing[@]} -gt 0 ]]; then
-            echo "::error::When 'task' is empty, the following inputs are required: ${missing[*]}"
-            exit 1
-          fi
-        fi
-    - name: Set task env
+    - name: Set benchmark output file
       shell: bash
-      run: |
```
maru-ava, line 147: EXECUTION_DATA_DIR is now set in the script

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -188,10 +160,3 @@ runs:
         summary-always: true
         github-token: ${{ inputs.github-token }}
         auto-push: ${{ inputs.push-github-action-benchmark }}
-    - name: Push Post-State to S3
```
maru-ava, line 191: @aaronbuchwald I might have been overzealous here - was your intent to only push to s3 if the benchmark comparison succeeded?

scripts/reexecute_cchain.sh
```diff
@@ -0,0 +1,137 @@
...
+[[ -z "${S3_BLOCK_DIR:-}" ]] && missing+=("S3_BLOCK_DIR")
+[[ -z "${S3_STATE_DIR:-}" ]] && missing+=("S3_STATE_DIR")
+[[ -z "${START_BLOCK:-}" ]] && missing+=("START_BLOCK")
+[[ -z "${END_BLOCK:-}" ]] && missing+=("END_BLOCK")
+
+if [[ ${#missing[@]} -gt 0 ]]; then
+    error "Missing required env vars: ${missing[*]}"
+fi
+
+TIMESTAMP=$(date '+%Y%m%d-%H%M%S')
```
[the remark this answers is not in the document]
maru-ava: The format is as expected.

.github/actions/c-chain-reexecution-benchmark/action.yml
```diff
@@ -126,54 +122,30 @@ runs:
...
-          ./scripts/run_task.sh reexecute-cchain-range-with-copied-data \
-            CONFIG=${{ inputs.config }} \
-            EXECUTION_DATA_DIR=${{ env.EXECUTION_DATA_DIR }} \
-            BLOCK_DIR_SRC=${{ inputs.block-dir-src }} \
-            CURRENT_STATE_DIR_SRC=${{ inputs.current-state-dir-src }} \
-            START_BLOCK=${{ inputs.start-block }} \
-            END_BLOCK=${{ inputs.end-block }} \
-            BENCHMARK_OUTPUT_FILE="${{ env.BENCHMARK_OUTPUT_FILE }}"
-        fi
+      run: ./scripts/reexecute_cchain.sh "${{ inputs.test || 'custom' }}"
```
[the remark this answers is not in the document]
maru-ava: This isn't a change in behavior.

scripts/reexecute_cchain.sh
```diff
@@ -0,0 +1,137 @@
...
+# Usage: ./scripts/reexecute_cchain.sh <test-name>
+#
+# Preset tests handle S3 import and execution automatically.
+# Environment variables (BENCHMARK_OUTPUT_FILE, METRICS_*, etc.) are passed through.
+#
+# Optional env vars:
+#   PUSH_POST_STATE - S3 destination to push current-state after execution (requires AWS credentials)
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+S3_BOOTSTRAP_BUCKET="${S3_BOOTSTRAP_BUCKET:-s3://avalanchego-bootstrap-testing}"
```
aaronbuchwald: If I'm not mistaken, we switched to turning each of these sets of parameters into tasks to comply with the notion that tasks should be fully encompassed within the task instead of passing in a bunch of parameters.

I'd argue that a wide range of different sets of options like this is not as common to place within a bash script.

Would you still push to keep each of these baked into one configuration options within this script if we are going to move it out of the taskfile?

Taskfile.yml
```diff
@@ -142,30 +125,6 @@ tasks:
...
-          DST: '{{.EXECUTION_DATA_DIR}}/current-state'
-
-  import-s3-to-dir:
-    desc: Imports an S3 path to a local directory.
-    vars:
-      SRC: '{{.SRC}}'
-      DST: '{{.DST}}'
-    cmds:
-      - cmd: bash -x ./scripts/copy_dir.sh {{.SRC}} {{.DST}}
-
```
RodrigoVillar, line 169: q: why was this removed (similar question as well for `export-dir-to-s3`)?

aaronbuchwald:
Could you please validate the change by stepping through the README and updating it as needed?

Taskfile.yml
maru-ava: The tasks existed to be consumed by other tasks or to be called from CI. The new script calls the copy_dir.sh script directly.

scripts/reexecute_cchain.sh
maru-ava: I intended for tasks to be both discoverable and to serve as entrypoints into CI so that reproducing CI was as simple as possible. The goal was not to implement anything non-trivial directly in the tasks, though, and I think that the reexec task logic is better maintained outside of the taskfile as I've proposed in this PR.   

I think creating a task per reexec test configuration makes about as much sense as creating a task per e2e test. It's still possible to target the execution of an e2e test but it's the e2e test suite that's responsible for enabling that rather than something we need to do in the taskfile. I think it similarly makes sense for there to be an entrypoint in the taskfile for executing reexec tests, and that the entrypoint be responsible for defining both common test configurations and enabling ad-hoc configuration.

scripts/reexecute_cchain.sh
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env bash
```
maru-ava: I'm not wedded to this structure, and it occurred to me that it might as well make sense to add this functionality directly to the benchmark_cchain_range.sh script.

Taskfile.yml
maru-ava: Ah, I see why you asked. You've proposed using `import-cchain-reexecute-range` in #4695. I think the right place for that would be a script, would you like to factor that out of the script proposed here for reuse?

scripts/reexecute_cchain.sh
```diff
@@ -0,0 +1,137 @@
...
+
+set -euo pipefail
+
+# C-Chain Re-execution Script
+# Usage: ./scripts/reexecute_cchain.sh <test-name>
```
RodrigoVillar: Running `START_BLOCK=1 END_BLOCK=200_000 task test-c-chain-reexecution -- custom` locally fails with the following error message:

```
task: [test-c-chain-reexecution] ./scripts/reexecute_cchain.sh custom
Error: Missing required env vars: S3_BLOCK_DIR S3_STATE_DIR
```

Ideally, I shouldn't have to set the S3 variables if I'm using my own current state/block directory.

Taskfile.yml
RodrigoVillar: I think rewriting `import-cchain-reexecute-range` into a script would be ideally while keeping around the ability to invoke it via a task.

scripts/reexecute_cchain.sh
maru-ava: If you're using your own state/block directory, you can call the `benchmark_cchain_range.sh` script directly. But to your point, maybe there should be a single more flexible script?

maru-ava:
> Could you please validate the change by stepping through the README and updating it as needed?

Done

Taskfile.yml
maru-ava: What would be the value of invoking it as a task?

scripts/reexecute_cchain.sh
RodrigoVillar: Ideally yes. To my understanding, `reexecute_cchain_range.sh` is an extension of `benchmark_cchain_range.sh` and so I think it would make sense to have one script rather than have two scripts which can be confusing (e.g. I ran `reexecute_cchain_range.sh` rather than `benchmark_cchain_range.sh`)

maru-ava: Makes sense, I'll tackle that first thing tomorrow.

Taskfile.yml
RodrigoVillar: My understanding of `task`s was that it should be the common gateway for invoking common commands. For example, I've become used to running `task lint` instead `./scripts/lint.sh` because of its convenience. Similarly, running `task reexecute-cchain-range` is much easier than having to invoke `./scripts/benchmark_cchain_range.sh`.

My preference of being able to invoke the reexecution test via a `task` command is based on the above, but I could certainly be mistaken.

maru-ava: Sure, but you're not configuring lint you're just running it. My intention with this PR is stop the counter-productive practice of defining an _API_ with a task via the use of vars. If you want a task entrypoint but are willing to configure it the same way as the underlying script - i.e. via CLI flags or env vars - that's fine.

RodrigoVillar: Makes sense. I guess my only question is how commands such as `import-cchain-reexecute-range` will be made composable, but will defer until this PR is updated.

maru-ava: I don't think there's a good case for task composition if it requires duplicating parameters with task vars. Better to just call the script/command directly.

scripts/reexecute_cchain.sh
maru-ava: Updated the existing script and added a new script/task for the import. PTAL

maru-ava:
Not ideal that we have a README with complex instructions that have to be manually verified when things change. Mental note: prefer an executable script in future with inline comments instead of a static README.

scripts/import_cchain_data.sh
RodrigoVillar, line 1: This looks great!

scripts/benchmark_cchain_range.sh
```diff
@@ -2,23 +2,170 @@
...
+    # Validate required variables for test mode
+    missing=()
+    [[ -z "${BLOCK_DIR_SRC:-}" ]] && missing+=("BLOCK_DIR_SRC")
+    [[ -z "${CURRENT_STATE_DIR_SRC:-}" ]] && missing+=("CURRENT_STATE_DIR_SRC")
+    [[ -z "${START_BLOCK:-}" ]] && missing+=("START_BLOCK")
+    [[ -z "${END_BLOCK:-}" ]] && missing+=("END_BLOCK")
+
+    if [[ ${#missing[@]} -gt 0 ]]; then
+        error "Missing required env vars: ${missing[*]}"
+    fi
```
RodrigoVillar: When running `task test-cchain-reexecution` with just `BLOCK_DIR` and `CURRENT_STATE_DIR` set, I get the following error message:

```
Without test-name, expects env vars: BLOCK_DIR, CURRENT_STATE_DIR, START_BLOCK, END_BLOCK
```

Ideally, it should tell me that just `START_BLOCK` and `END_BLOCK` are missing.

Taskfile.yml
RodrigoVillar: Gotcha, calling the script sounds good to me.

scripts/benchmark_cchain_range.sh
```diff
@@ -2,23 +2,170 @@
...
+# Optional env vars (all modes):
+#   CONFIG: VM config preset (default, archive, firewood).
+#   LABELS: Comma-separated key=value pairs for metric labels.
+#   BENCHMARK_OUTPUT_FILE: If set, benchmark output is also written to this file.
+#   METRICS_SERVER_ENABLED: If set, enables the metrics server.
+#   METRICS_SERVER_PORT: If set, determines the port the metrics server will listen to.
+#   METRICS_COLLECTOR_ENABLED: If set, enables the metrics collector.
+#   PUSH_POST_STATE: S3 destination to push current-state after execution.
+#
+# For 'custom' test or S3 import with other tests:
```
RodrigoVillar: (nit / feel-free-to-ignore): could we say something along the lines of "For all other tests which fetch from S3" to make it clear that any predefined/custom test uses S3?

tests/reexecute/c/README.md
```diff
@@ -56,26 +54,29 @@ Running the re-execution test in CI will implicitly set `METRICS_COLLECTOR_ENABL
...
+```
 
-### Using Custom Parameters
+These tests automatically download the required data from S3 and run the benchmark with the appropriate configuration.
 
-For custom benchmark runs:
+### Using Custom Parameters
 
-[Taskfile](https://taskfile.dev/) supports reading arguments via both environment variables and named arguments on the command line, so we'll set `EXECUTION_DATA_DIR` and use the defaults for the remainder of the parameters:
+For custom benchmark runs, use the `custom` test with environment variables:
```
RodrigoVillar: nit: this seems like `./scripts/benchmark_cchain_range.sh` and `./scripts/benchmark_cchain_range.sh custom` do the same thing

scripts/benchmark_cchain_range.sh
maru-ava: Done.

~Though 'runs a predefined test that automatically imports data from S3' on L12 already tried to convey this.~

Edit: Rewrite this section as part of removing `custom`

scripts/benchmark_cchain_range.sh
maru-ava: Updated to indicate the status of required env vars.

tests/reexecute/c/README.md
maru-ava: Good catch, removed `custom`

maru-ava:
Rebased

tests/reexecute/c/README.md
RodrigoVillar, line 1: (out-of-scope) While I'm a huge proponent of "documentation is king", this should really live as a man page instead of a markdown file.

scripts/benchmark_cchain_range.sh
RodrigoVillar: This looks great!

scripts/benchmark_cchain_range.sh
```diff
@@ -2,24 +2,166 @@
...
+#
+# Environment variables:
+#   Data sources (provide S3 sources OR local paths):
+#     BLOCK_DIR_SRC: S3 object key for blocks (triggers S3 import).
+#     CURRENT_STATE_DIR_SRC: S3 object key for state (triggers S3 import).
+#     BLOCK_DIR: Path to local block directory.
+#     CURRENT_STATE_DIR: Path to local current state directory.
+#
+#   Required:
+#     START_BLOCK: The starting block height (exclusive).
```
RodrigoVillar:
```suggestion
#     START_BLOCK: The starting block height (inclusive).
```

This should be inclusive as the start block is executed.

scripts/reexecute_cchain.sh
RodrigoVillar: LGTM.

scripts/benchmark_cchain_range.sh
```diff
@@ -2,24 +2,166 @@
...
-: "${BLOCK_DIR:?BLOCK_DIR must be set}"
-: "${CURRENT_STATE_DIR:?CURRENT_STATE_DIR must be set}"
-: "${START_BLOCK:?START_BLOCK must be set}"
-: "${END_BLOCK:?END_BLOCK must be set}"
+# C-Chain Re-execution Benchmark Script
+#
+# Usage:
+#   ./benchmark_cchain_range.sh [test-name]
+#
+# Running without arguments will output available tests.
```
RodrigoVillar: Running without arguments now runs the custom test.

tests/reexecute/c/README.md
```diff
@@ -2,8 +2,6 @@
 
 The C-Chain benchmarks support re-executing a range of mainnet C-Chain blocks against a provided snapshot of the current state as of some initial state.
 
-AvalancheGo provides a [Taskfile](https://taskfile.dev/) with commands to manage the import/export of data required for re-execution (block range and current state) and triggering a benchmark run.
```
RodrigoVillar, line 5: q: do you think we should mention somewhere in the README that the reexecution script can also be invoked via the taskfile?

scripts/benchmark_cchain_range.sh
maru-ava: Done

scripts/benchmark_cchain_range.sh
maru-ava: Only if the required env vars are provided, otherwise the usage message gets printed with the available tests. To your point, though, I've updated this line to hopefully to be more informative.

tests/reexecute/c/README.md
maru-ava: Added mention of both methods of execution to this line.

aaronbuchwald (approved):

PR #4762, opened 2025-12-17
build: update go version to v1.24.11

JonathanOppenheimer:
## Why this should be merged

See: https://github.com/ava-labs/precompile-evm/pull/99#discussion_r2604525896. These updates also fixed some network and crypto vulnerabilities. See https://go.dev/doc/devel/release

## How this was tested
CI

Yes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,8 +12,6 @@ module github.com/ava-labs/avalanchego
 //   - graft/evm/go.mod
 //   - graft/coreth/go.mod
 //   - graft/subnet-evm/go.mod
-//   - CONTRIBUTING.md
-//   - README.md
 //
 // and
 //   - CONTRIBUTING.md


go.mod
maru-ava: Maybe make this explicit so that reviewers have a definitive list of files to check?

graft/subnet-evm/README.md
```diff
@@ -73,7 +73,7 @@ To support these changes, there have been a number of changes to the SubnetEVM b
 
 ### Clone Subnet-evm
 
-First install Go 1.24.9 or later. Follow the instructions on the [go docs](https://go.dev/doc/install). You can verify by running `go version`.
```
maru-ava, line 76: Maybe either: 
 - remove a specific version from this document with a link to the relevant section of CONTRIBUTING.md?
 or
 - Add an indication to go.mod comment that this doc needs to be updated?

graft/subnet-evm/Dockerfile
```diff
@@ -4,7 +4,7 @@
 ARG AVALANCHEGO_NODE_IMAGE="invalid-image"
 
 # ============= Compilation Stage ================
-FROM --platform=$BUILDPLATFORM golang:1.24.9-bookworm AS builder
+FROM --platform=$BUILDPLATFORM golang:1.24.11-bookworm AS builder
```
maru-ava, line 7: Please:
 - Add this file to the list of files in the go.mod comment
 - File an issue to update this image to accept a GO_VERSION arg as per the example of avalanchego's Dockerfile so that we avoid having to maintain the version in this file, and make mention of removing the file from the go.mod comment once this is done.

graft/coreth/RELEASES.md
```diff
@@ -7,6 +7,7 @@
   - Also removes blockHook extension in `ethclient` package.
 - Enables Firewood to run with pruning disabled.
   - This change modifies the filepath of Firewood and any nodes using Firewood will need to resync.
+- Update go version to 1.24.11
```
maru-ava: Why is a note like this something we have to maintain for coreth but not avalanchego? If the practice is to continue, maybe it needs a note in the go.mod comment?

JonathanOppenheimer: Oh honestly we should probably get rid of this file entirely? In my opinion we should maintain a single RELEASES.MD perhaps with an EVM section.

JonathanOppenheimer: I will do that in this PR, as I don't think it's a huge change, but if you'd prefer it in a different PR that's totally fair.

maru-ava: This PR is fine

graft/subnet-evm/Dockerfile
JonathanOppenheimer: https://github.com/ava-labs/avalanchego/issues/4764

graft/subnet-evm/README.md
JonathanOppenheimer: I will implement the first option.

go.mod
```diff
@@ -5,16 +5,26 @@ module github.com/ava-labs/avalanchego
...
 
 // - Changes to the minimum golang version must also be replicated in:
+//
+// The following go.mod files:
+//   - go.mod (here)
+//   - tools/go.mod
+//   - graft/evm/go.mod
+//   - graft/coreth/go.mod
+//   - graft/subnet-evm/go.mod
 //   - CONTRIBUTING.md
```
maru-ava: Looks like these 2 lines can be removed

JonathanOppenheimer: oops thanks!

PR #4768, opened 2025-12-18
[nix] Update to 25.11 and stop recommending determinate systems installer

maru-ava:
## Why this should be merged

- Update to 25.11. 25.05 is EOL at EOY

- Update protobuf generation to account for the updated binary versions provided by nix

- The nix installer from Determinate Systems will no longer be installing vanilla upstream nix as of Jan 1 2026. Without a support contract from Determinate we're better off using upstream nix.

## How this was tested

CI

N/A


graft/subnet-evm/scripts/dev_shell.sh
```diff
@@ -1,18 +0,0 @@
-#!/usr/bin/env bash
```
maru-ava, line 1: As per the doc update, this script is replaced by `nix develop` anywhere in the repo.

scripts/protobuf_codegen.sh
```diff
@@ -10,23 +10,26 @@ fi
 # the versions here should match those of the binaries installed in the nix dev shell
 
 ## ensure the correct version of "buf" is installed
-BUF_VERSION='1.52.1'
-if [[ $(buf --version | cut -f2 -d' ') != "${BUF_VERSION}" ]]; then
-  echo "could not find buf ${BUF_VERSION}, is it installed + in PATH?"
+BUF_VERSION='1.59.0'
+ACTUAL_BUF_VERSION=$(buf --version 2>/dev/null | cut -f2 -d' ' || echo "not found")
```
maru-ava, line 14: Since the protobuf version often changes when upgrading nix releases, these changes ensuring the actual version is output when there is a version mismatch to simplify discovering the new version that needs to be configured in this file.

nix/go/default.nix
```diff
@@ -28,7 +28,7 @@ let
     "darwin-arm64" = "a9c90c786e75d5d1da0547de2d1199034df6a4b163af2fa91b9168c65f229c12";
   };
 
-  targetSystem = parseSystem pkgs.system;
+  targetSystem = parseSystem pkgs.stdenv.hostPlatform.system;
```
maru-ava, line 31: `system` -> `stdenv.hostPlatform` was prompted by a deprecation warning

scripts/install_nix.sh
```diff
@@ -0,0 +1,34 @@
...
+      echo "experimental-features = nix-command flakes" >> ~/.config/nix/nix.conf
+    fi
+    ;;
+  *)
+    echo "Unsupported OS: $(uname -s)"
+    exit 1
+    ;;
+esac
+
+echo "Nix installation complete."
```
JonathanOppenheimer, line 34: does this require a shell restart, a source or anything to activate nix? or not necessary?

maru-ava: The installation scripts already provide an indication of next steps.

StephenButtolph (approved):

PR #4769, opened 2025-12-18
refactor: Remove inbound message interface

alarso16:
## Why this should be merged

An immediate follow-up to #4766.

## How this works

Removes an interface with one implementation

## How this was tested

CI

No


message/messages.go
```diff
@@ -28,71 +28,31 @@ const (
...
-func (m *inboundMessage) Message() fmt.Stringer {
-	return m.message
-}
-
-func (m *inboundMessage) Expiration() time.Time {
-	return m.expiration
-}
-
-func (m *inboundMessage) OnFinishedHandling() {
+func (m *InboundMessage) OnFinishedHandling() {
```
alarso16, line 47: Didn't want to get rid of this because of the nil case

PR #4770, opened 2025-12-18
feat!: Use disk percentage instead of bytes for health

alarso16:
## Why this should be merged

Configuring some constant size for disk warnings is impractical, since most setups use auto-scaling on cloud providers, which is percentage based.

## How this works

Deprecates byte-based disk warnings, replaces with percentage based warnings.

## How this was tested

CI

Yes

### DIFF
--- a/tests/fixture/tmpnet/defaults.go
+++ b/tests/fixture/tmpnet/defaults.go
@@ -38,8 +38,8 @@ const (
 // Flags suggested for temporary networks. Applied by default.
 func DefaultTmpnetFlags() FlagsMap {
 	return FlagsMap{
-		config.SystemTrackerRequiredAvailableDiskSpacePercentageKey: "1",
-		config.SystemTrackerWarningAvailableDiskSpacePercentageKey:  "1",
+		config.SystemTrackerRequiredAvailableDiskSpacePercentageKey: "0",
+		config.SystemTrackerWarningAvailableDiskSpacePercentageKey:  "0",
 		config.NetworkPeerListPullGossipFreqKey:                     "250ms",
 		config.NetworkMaxReconnectDelayKey:                          "1s",
 		config.HealthCheckFreqKey:                                   "2s",


config/config.go
```diff
@@ -1138,17 +1141,20 @@ func getCPUTargeterConfig(v *viper.Viper) (tracker.TargeterConfig, error) {
...
-		return 0, 0, 0, fmt.Errorf("%q (%d) < %q (%d)", SystemTrackerWarningThresholdAvailableDiskSpaceKey, warningThresholdAvailableDiskSpace, SystemTrackerRequiredAvailableDiskSpaceKey, requiredAvailableDiskSpace)
+	case warnThresholdAvailableDiskSpacePercentage > 50:
+		return 0, 0, fmt.Errorf("%q (%d) must be in [0, 50]", SystemTrackerWarnThreshAvailDiskSpacePercentageKey, warnThresholdAvailableDiskSpacePercentage)
+	case warnThresholdAvailableDiskSpacePercentage < requiredAvailableDiskSpacePercentage:
+		return 0, 0, fmt.Errorf("%q (%d) < %q (%d)", SystemTrackerWarnThreshAvailDiskSpacePercentageKey, warnThresholdAvailableDiskSpacePercentage, SystemTrackerRequiredAvailableDiskSpacePercentageKey, requiredAvailableDiskSpacePercentage)
 	default:
-		return requiredAvailableDiskSpace, warningThresholdAvailableDiskSpace, warningThresholdAvailableDiskSpacePercentage, nil
+		return requiredAvailableDiskSpacePercentage, warnThresholdAvailableDiskSpacePercentage, nil
 	}
 }
```
StephenButtolph, line 1168: There is just too much text here haha:
```golang
func getDiskSpaceConfig(v *viper.Viper) (
	requiredAvailableDiskSpacePercentage uint64,
	warningAvailableDiskSpacePercentage uint64,
	err error,
) {
	var (
		warnKey     = SystemTrackerWarnThreshAvailDiskSpacePercentageKey
		requiredKey = SystemTrackerRequiredAvailableDiskSpacePercentageKey

		warn     = v.GetUint64(warnKey)
		required = v.GetUint64(requiredKey)
	)
	switch {
	case warn > 50:
		return 0, 0, fmt.Errorf("%q (%d) must be in [0, 50]", warnKey, warn)
	case warn < required:
		return 0, 0, fmt.Errorf("%q (%d) < %q (%d)", warnKey, warn, requiredKey, required)
	default:
		return required, warn, nil
	}
}
```
Seems a lot easier to read

config/keys.go
```diff
@@ -70,145 +70,146 @@ const (
...
+	CPUMaxNonVdrUsageKey                                 = "throttler-inbound-cpu-max-non-validator-usage"
+	CPUMaxNonVdrNodeUsageKey                             = "throttler-inbound-cpu-max-non-validator-node-usage"
+	SystemTrackerFrequencyKey                            = "system-tracker-frequency"
+	SystemTrackerProcessingHalflifeKey                   = "system-tracker-processing-halflife"
+	SystemTrackerCPUHalflifeKey                          = "system-tracker-cpu-halflife"
+	SystemTrackerDiskHalflifeKey                         = "system-tracker-disk-halflife"
+	SystemTrackerRequiredAvailableDiskSpaceKey           = "system-tracker-disk-required-available-space"
+	SystemTrackerRequiredAvailableDiskSpacePercentageKey = "system-tracker-disk-required-available-space-percentage"
+	SystemTrackerWarningThresholdAvailableDiskSpaceKey   = "system-tracker-disk-warning-threshold-available-space"
+	SystemTrackerWarnThreshAvailDiskSpacePercentageKey   = "system-tracker-disk-warning-threshold-available-space-percentage"
```
StephenButtolph: I know this isn't changed in this PR... But this variable name is already extremely verbose... Are we actually saving anything by shortening `Threshold` to `Thresh` and `Available` to `Avail`?

Also, if `Threshold` isn't required in the `require` keys... I don't think it should be needed in the `warning` keys.

Lets change this to:
```
SystemTrackerWarningAvailableDiskSpacePercentageKey   = "system-tracker-disk-warning-available-space-percentage"
```

We don't need to worry about a deprecation notice for this change, because we haven't actually released this flag publicly yet.

node/node.go
```diff
@@ -1460,19 +1460,15 @@ func (n *Node) initHealthAPI() error {
 		availableDiskPercentage := n.resourceTracker.DiskTracker().AvailableDiskPercentage()
 
 		var diskSpaceErrors []error
-		if availableDiskBytes < n.Config.RequiredAvailableDiskSpace {
-			n.Log.Fatal("low on disk space. Shutting down...",
-				zap.Uint64("remainingDiskBytes", availableDiskBytes),
+
+		if availableDiskPercentage < n.Config.RequiredAvailableDiskSpacePercentage {
+			n.Log.Fatal("low on disk space percentage. Shutting down...",
```
StephenButtolph: I don't think the percentage comment is really needed in the log.
```suggestion
			n.Log.Fatal("low on disk space. Shutting down...",
```

node/node.go
```diff
@@ -1460,19 +1460,15 @@ func (n *Node) initHealthAPI() error {
...
+				zap.Uint64("remainingDiskPercentage", availableDiskPercentage),
```
StephenButtolph, line 1467: Can we log the exact remaining bytes as well? We still have the information

config/node/config.go
```diff
@@ -205,9 +205,10 @@ type Config struct {
 
 	DiskTargeterConfig tracker.TargeterConfig `json:"diskTargeterConfig"`
 
-	RequiredAvailableDiskSpace                   uint64 `json:"requiredAvailableDiskSpace"`
-	WarningThresholdAvailableDiskSpace           uint64 `json:"warningThresholdAvailableDiskSpace"`
-	WarningThresholdAvailableDiskSpacePercentage uint64 `json:"warningThresholdAvailableDiskSpacePercentage"`
+	RequiredAvailableDiskSpace           uint64 `json:"requiredAvailableDiskSpace"`
+	RequiredAvailableDiskSpacePercentage uint64 `json:"requiredAvailableDiskSpacePercentage"`
+	WarningThresholdAvailableDiskSpace   uint64 `json:"warningThresholdAvailableDiskSpace"`
+	WarningAvailableDiskSpacePercentage  uint64 `json:"warningThresholdAvailableDiskSpacePercentage"`
```
StephenButtolph:
```suggestion
	WarningAvailableDiskSpacePercentage  uint64 `json:"warningAvailableDiskSpacePercentage"`
```

node/node.go
```diff
@@ -1460,20 +1460,18 @@ func (n *Node) initHealthAPI() error {
 		availableDiskPercentage := n.resourceTracker.DiskTracker().AvailableDiskPercentage()
 
 		var diskSpaceErrors []error
-		if availableDiskBytes < n.Config.RequiredAvailableDiskSpace {
+
```
StephenButtolph: A bit pedantic, but we don't need a slice anymore. There is only ever at most 1 error appended to the slice.

RELEASES.md
```diff
@@ -11,6 +11,7 @@ This version is backwards compatible to [v1.14.0](https://github.com/ava-labs/av
 ### Fixes
 
 - Update go version to 1.24.11
+- Deprecate `RequiredAvailableDiskSpace` and `WarningThresholdAvailableDiskSpace` config options.
```
StephenButtolph: Typically we put a specific group for config changes. Can we move these there? Also can we comment on the new configs?

Last nit is that we should probably say what the actual flag is here `--system-tracker-disk-required-available-space` and `--system-tracker-disk-warning-threshold-available-space` rather than `RequiredAvailableDiskSpace` and `WarningThresholdAvailableDiskSpace`

alarso16: Ah of course. It's not very detailed, but I did fix both of these

node/node.go
alarso16: Ahhh, duh.

config/node/config.go
```diff
@@ -205,9 +205,10 @@ type Config struct {
 
 	DiskTargeterConfig tracker.TargeterConfig `json:"diskTargeterConfig"`
 
-	RequiredAvailableDiskSpace                   uint64 `json:"requiredAvailableDiskSpace"`
-	WarningThresholdAvailableDiskSpace           uint64 `json:"warningThresholdAvailableDiskSpace"`
-	WarningThresholdAvailableDiskSpacePercentage uint64 `json:"warningThresholdAvailableDiskSpacePercentage"`
+	RequiredAvailableDiskSpace           uint64 `json:"requiredAvailableDiskSpace"`
```
StephenButtolph:
```suggestion
```

config/node/config.go
```diff
@@ -205,9 +205,10 @@ type Config struct {
...
+	RequiredAvailableDiskSpacePercentage uint64 `json:"requiredAvailableDiskSpacePercentage"`
+	WarningThresholdAvailableDiskSpace   uint64 `json:"warningThresholdAvailableDiskSpace"`
```
StephenButtolph:
```suggestion
```

config/keys.go
```diff
@@ -70,145 +70,146 @@ const (
...
+	InboundThrottlerDiskMaxRecheckDelayKey               = "throttler-inbound-disk-max-recheck-delay"
+	CPUVdrAllocKey                                       = "throttler-inbound-cpu-validator-alloc"
+	CPUMaxNonVdrUsageKey                                 = "throttler-inbound-cpu-max-non-validator-usage"
+	CPUMaxNonVdrNodeUsageKey                             = "throttler-inbound-cpu-max-non-validator-node-usage"
+	SystemTrackerFrequencyKey                            = "system-tracker-frequency"
+	SystemTrackerProcessingHalflifeKey                   = "system-tracker-processing-halflife"
+	SystemTrackerCPUHalflifeKey                          = "system-tracker-cpu-halflife"
+	SystemTrackerDiskHalflifeKey                         = "system-tracker-disk-halflife"
+	SystemTrackerRequiredAvailableDiskSpaceKey           = "system-tracker-disk-required-available-space"
+	SystemTrackerRequiredAvailableDiskSpacePercentageKey = "system-tracker-disk-required-available-space-percentage"
```
alarso16, line 195: This is the added key (for review ease)

config/config.go
```diff
@@ -1138,17 +1141,25 @@ func getCPUTargeterConfig(v *viper.Viper) (tracker.TargeterConfig, error) {
...
 
-func getDiskSpaceConfig(v *viper.Viper) (requiredAvailableDiskSpace uint64, warningThresholdAvailableDiskSpace uint64, warningThresholdAvailableDiskSpacePercentage uint64, err error) {
-	requiredAvailableDiskSpace = v.GetUint64(SystemTrackerRequiredAvailableDiskSpaceKey)
-	warningThresholdAvailableDiskSpace = v.GetUint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey)
-	warningThresholdAvailableDiskSpacePercentage = v.GetUint64(SystemTrackerWarnThreshAvailDiskSpacePercentageKey)
+func getDiskSpaceConfig(v *viper.Viper) (
```
JonathanOppenheimer, line 1148: should we add any tests for this function? (like for warn < required or warn > 50).

could also make these sentinel errors rather than formatted strings

node/node.go
```diff
@@ -1459,26 +1459,20 @@ func (n *Node) initHealthAPI() error {
 		availableDiskBytes := n.resourceTracker.DiskTracker().AvailableDiskBytes()
 		availableDiskPercentage := n.resourceTracker.DiskTracker().AvailableDiskPercentage()
```
JonathanOppenheimer, line 1460: You can trace the available disk percentage to the origin:
https://github.com/ava-labs/avalanchego/blob/ba52a5a9efa78257f4d8a1891753873eb871fc50/utils/storage/storage_unix.go#L16

The percentage remaining is just an integer / 100 (so there's no fractional percents) -- so if we had a situation where there was a 1 TB disk with like 35GB free (3.5%), that would be reported as 3% and the node would shut down. I'm pretty sure this should be propagated as a float throughout the entire way? (unless we don't care about these sort of rounding errors)

Also more generally, I thought more about this and I actually don't like a pure % based system. If I have a 10TB disk, 1% is 100GB, which is A TON OF space. Maybe you guys discussed this more extensively, but I feel like there should be some sort of check within the available disk percentage that checks the actual remaining space in bytes, and if it's above some reasonable amount we don't warn / shutdown.

JonathanOppenheimer: Maybe I don't understand the cloud scaling thing, but do you sort of know what I'm getting at -- naively doing the percent would lead to shut downs and warnings that seem premature on large disks. 

(e.g. an extra 90GB is an extra 3 days of runway)

alarso16: 1. I'm a float hater, we shouldn't add a float into prod code
2. It is a lot as a constant, but 1% at that rate of growth should still the autoscaler plenty of time to deal with any issues. Having less than a couple percent open disk space is rather dangerous, even if the number looks large

alarso16: I don't plan on addressing this

tests/fixture/tmpnet/defaults.go
```diff
@@ -38,20 +38,22 @@ const (
...
+		config.AdminAPIEnabledKey:                                   "true",
+		config.IndexEnabledKey:                                      "true",
 	}
 }
 
 // Flags suggested for e2e testing
 func DefaultE2EFlags() FlagsMap {
 	return FlagsMap{
-		config.ProposerVMUseCurrentHeightKey: "true",
+		config.SystemTrackerRequiredAvailableDiskSpacePercentageKey: "1",
```
maru-ava: These additions are unnecessary - DefaultTmpnetFlags are always applied and DefaultE2EFlags only to e2e testing.

config/config.go
alarso16: I added the bare minimum to test the functionality guaranteed. The tests feel like overkill and the errors are objectively more confusing, but it's probably fine

JonathanOppenheimer: Oh I was legitimately asking if it should be tested. If you think the tests are overkill, and the errors more confusing, they should both be removed. (although I dont think there's a downside imo testing things that are seemingly obvious as who knows what might happen down the line)

tests/fixture/tmpnet/defaults.go
alarso16: The e2e test failed without them 🤷‍♂️, the nodes never reported healty

alarso16: Specifically, [this commit](https://github.com/ava-labs/avalanchego/pull/4770/changes/477a58ca974aa5fa0f5cab642f1c46d8b73458a8) wasn't sufficient (you can see the CI failure)

alarso16: Seems that the runners are rather unpredictable, and adding this didn't actually prevent the failure. We can just disable this check completely, since the tmpnet better not use very much disk space...

maru-ava (approved):

PR #4775, opened 2025-12-19
fix(ci): rename RUNNER_NAME to avoid GitHub env var conflict

Elvis339:
## Why this should be merged

Every benchmark test run creates a new sub-dashboard instead of adding to the existing dashboard (ref: https://ava-labs.github.io/avalanchego/dev/bench/).

The runner name is changing for every test run because GitHub's built-in RUNNER_NAME env var (which contains the ephemeral ARC runner name like avalanche-avalanchego-runner-2ti-z284t-runner-88vxw) is overwriting the static runner name we pass.

This was introduced in #4443 when the approach changed from passing RUNNER_NAME as a task argument (highest precedence) to setting it as an environment variable (lower precedence than GitHub's built-in).

## How this works

Renames the custom environment variable from `RUNNER_NAME` to `BENCHMARK_RUNNER_NAME` to avoid conflicting with GitHub's built-in `RUNNER_NAME` variable.

## How this was tested

- CI: Compared against the working behavior before [#4443](https://github.com/ava-labs/avalanchego/actions/runs/19718501681/job/56496196216#step:5:599) and now https://github.com/ava-labs/avalanchego/actions/runs/20373088623/job/58544395265?pr=4775#step:5:599

No

### DIFF
--- a/.github/actions/c-chain-reexecution-benchmark/action.yml
+++ b/.github/actions/c-chain-reexecution-benchmark/action.yml
@@ -21,8 +21,8 @@ inputs:
   current-state-dir-src:
     description: 'The current state directory. Supports S3 directory/zip and local directories.'
     default: ''
-  runner_name:
-    description: 'The name of the runner to use and include in the Golang Benchmark name.'
+  runner_type:
+    description: 'The type/label of the runner to include in the benchmark name for grouping results.'
     required: true
   aws-role:
     description: 'AWS role to assume for S3 access.'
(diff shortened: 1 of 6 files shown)


Elvis339:
https://github.com/ava-labs/avalanchego/actions/runs/20373088623/job/58544395265?pr=4775#step:5:593

Screenshot 2025-12-19 at 18 35 52

.github/actions/c-chain-reexecution-benchmark/action.yml
maru-ava: Ideally the input name matches to avoid ambiguity. And maybe 'runner_type' would be more appropriate given that we're talking about the type of runner rather than the name of an individual runner?

Elvis339: Done: https://github.com/ava-labs/avalanchego/pull/4775/changes/a3a81b0a641442c3bc851f13f0823c5fc096559e

To keep things consistent, the change ended up larger than I planned.

RodrigoVillar (approved):

maru-ava (approved):

PR #4783, opened 2025-12-19
ci: arm64 to tier 1 support

### LINKED ISSUE #4773: Upgrade linux arm64 to Tier 1 supported
The only thing holding linux arm64 from Tier 1 support is running the e2e suite on a linux arm64 runner.

- Update CI so the e2e job runs runs on the current default runner and a linux arm64 runner
- Update https://github.com/ava-labs/avalanchego?tab=readme-ov-file#supported-platforms

JonathanOppenheimer:
## Why this should be merged
Closes #4773 

## How this was tested
CI

Yes

### DIFF
--- a/firewood/syncer/proof.go
+++ b/firewood/syncer/proof.go
@@ -1,0 +1,51 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package firewood
+
+import (
+	"errors"
+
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
maru-ava: Thanks for doing this! This prefix will need to be differentiated per runner or arch to avoid the name clash that's failing the job.

JonathanOppenheimer: Got it! Is appending the `${{ matrix.os }}` a reasonable name?

maru-ava (approved):

Issue #1508, opened 2023-05-15
Cache BLS Serialized form in PVM


joshua-kim:
Followup to https://github.com/ava-labs/avalanchego/pull/1445.

We can cache the serialized form of our bls keys in memory to avoid re-serializing keys (which is expensive) every time `GetValidatorSet` is called.

joshua-kim:
Duplicate of https://github.com/ava-labs/avalanchego/issues/3902

PR #1467, opened 2023-05-03
Some more GetCanonicalValidatorSet benchmarks

abi87:

PR #1445, opened 2023-04-28
Cache BLS Keys

joshua-kim:
## Why this should be merged

Since the P-Chain communicates with the `warp` contract over GRPC, we can re-use the bytes that are coming out of grpc to pre-populate the serialized form of a bls key, avoiding re-calculating it later which turns out to be roughly half of the cpu cycles spent in `GetCanonicalValidatorSet` (aside from sorting).

## How this works

Pre-populates a new field that holds the cached representation of the bls key.

## How this was tested

Added a benchmark test. Seeing roughly a 20% performance boost:

Before:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/vms/platformvm/warp
BenchmarkGetCanonicalValidatorSet
BenchmarkGetCanonicalValidatorSet/0
BenchmarkGetCanonicalValidatorSet/0-10  	 6599539	       180.4 ns/op
BenchmarkGetCanonicalValidatorSet/1
BenchmarkGetCanonicalValidatorSet/1-10  	 2472804	       485.5 ns/op
BenchmarkGetCanonicalValidatorSet/10
BenchmarkGetCanonicalValidatorSet/10-10 	  342585	      3498 ns/op
BenchmarkGetCanonicalValidatorSet/100
BenchmarkGetCanonicalValidatorSet/100-10         	   33254	     36251 ns/op
BenchmarkGetCanonicalValidatorSet/1000
BenchmarkGetCanonicalValidatorSet/1000-10        	    3010	    388901 ns/op
BenchmarkGetCanonicalValidatorSet/10000
BenchmarkGetCanonicalValidatorSet/10000-10       	     260	   4471217 ns/op
PASS
```

After:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/avalanchego/vms/platformvm/warp
BenchmarkGetCanonicalValidatorSet
BenchmarkGetCanonicalValidatorSet/0
BenchmarkGetCanonicalValidatorSet/0-10  	 6491647	       186.2 ns/op
BenchmarkGetCanonicalValidatorSet/1
BenchmarkGetCanonicalValidatorSet/1-10  	 3459529	       349.5 ns/op
BenchmarkGetCanonicalValidatorSet/10
BenchmarkGetCanonicalValidatorSet/10-10 	  557814	      2129 ns/op
BenchmarkGetCanonicalValidatorSet/100
BenchmarkGetCanonicalValidatorSet/100-10         	   49543	     24245 ns/op
[...]

PR #4784, opened 2025-12-19
perf: Optimize BLS key usage

### LINKED ISSUE #3902: Optimize BLS key usage
There are three different representations of BLS public keys used by avalanchego:
1. In-memory blst pointer
2. bytes in the uncompressed format
3. bytes in the compressed format

We currently convert (2) and (3) to (1) and pass (1) around. However, there are instances where we repeatedly convert between the formats.

For example, during ICM verification, we may:
1. Read (2) from disk [here](https://github.com/ava-labs/avalanchego/blob/2ef0cbeda604c3fecb70e3faa18dc58377f47c04/vms/platformvm/state/state.go#L1487)
2. Convert to (1) to [apply the validator diff](https://github.com/ava-labs/avalanchego/blob/2ef0cbeda604c3fecb70e3faa18dc58377f47c04/vms/platformvm/state/state.go#L1493)
3. Convert to (2) to [send over the grpc server](https://github.com/ava-labs/avalanchego/blob/2ef0cbeda604c3fecb70e3faa18dc58377f47c04/snow/validators/gvalidators/validator_state_server.go#L76)
4. Convert to (1) to [convert back from the grpc format](https://github.com/ava-labs/avalanchego/blob/master/snow/validators/gvalidators/validator_state_client.go#L84)
5. Convert to (2) to [convert to the canonical warp format](https://github.com/ava-labs/avalanchego/blob/2ef0cbeda604c3fecb70e3faa18dc58377f47c04/vms/platformvm/warp/validator.go#L90)

This PR #1445 did some of the work (removes 4,5) but we should probably do a holistic pass over how we represent public keys.

JonathanOppenheimer:
## Why this should be merged

This PR eliminates redundant BLS public key conversions throughout the validator codebase, which is a somewhat significant performance increase. 

This is measured in the following benchmark: 

`go test -bench=. -benchmem ./vms/platformvm/warp | grep "BenchmarkGetCanonicalValidatorSet"`

A summary of results:

  | Size              | Metric  | Unoptimized | Optimized | Improvement |
  |-------------------|---------|-------------|-----------|-------------|
  | 100 validators    | Time/op | 1.75ms      | 1.04ms    | 41% faster  |
  | 100 validators    | Memory  | 219KB       | 159KB     | 27% less    |
  | 100 validators    | Allocs  | 2,624       | 1,975     | 25% fewer   |
  | 1,000 validators  | Time/op | 17.4ms      | 12.3ms    | 29% faster  |
  | 1,000 validators  | Memory  | 2.15MB      | 1.58MB    | 27% less    |
  | 1,000 validators  | Allocs  | 26,232      | 19,749    | 25% fewer   |
  | 10,000 validators | Time/op | 182ms       | 130ms     | 29% faster  |
  | 10,000 validators | Memory  | 21.5MB      | 15.8MB    | 27% less    |
  | 10,000 validators | Allocs  | 262,323     | 197,489   | 25% fewer   |

## How this works

For prior context see 
- https://github.com/ava-labs/avalanchego/pull/1445
- https://github.com/ava-labs/avalanchego/pull/1467
- https://github.com/ava-labs/avalanchego/issues/1508
- https://github.com/ava-labs/avalanchego/issues/3902

Closes https://github.com/ava-labs/avalanchego/issues/3902

1. Warp struct 
    - Removed redundant `PublicKey` field from Warp struct, keeping only `PublicKeyBytes`
    - Updated `FlattenValidatorSet` to use `PublicKeyBytes` directly 

2. `GetValidatorOutput`
    - Added `PublicKeyBytes` field to store uncompressed bytes (96 bytes)
    - Populated in key locations: `ApplyValidatorPublicKeyDiffs`, `gRPC` client, validator set `Map()` function
    - This avoids the pointer -> bytes conversions 

There's no breaking changes in this PR, as the cache fields aren't serialized, so no concerns there either. 

## How this was tested
CI, the `BenchmarkGetCanonicalValidatorSet` 

No


snow/validators/warp_test.go
```diff
@@ -53,7 +52,11 @@ func TestWarpJSON(t *testing.T) {
 
 	var parsedW Warp
 	require.NoError(t, json.Unmarshal(wJSON, &parsedW))
-	require.Equal(t, w, parsedW)
+
+	// Compare only public fields, not cache
```
JonathanOppenheimer, line 56: not sure how to avoid this

snow/validators/warp_test.go
```diff
@@ -97,7 +99,15 @@ func TestWarpSetJSON(t *testing.T) {
 
 	var parsedWS WarpSet
 	require.NoError(t, json.Unmarshal(wsJSON, &parsedWS))
-	require.Equal(t, ws, parsedWS)
+
+	// Compare only public fields, not cache
```
JonathanOppenheimer, line 103: or this

vms/platformvm/warp/validator_test.go
```diff
@@ -313,32 +310,68 @@ func BenchmarkGetCanonicalValidatorSet(b *testing.B) {
 	pChainHeight := uint64(1)
 	subnetID := ids.GenerateTestID()
 	numNodes := 10_000
-	getValidatorOutputs := make([]*validators.GetValidatorOutput, 0, numNodes)
+
+	// Create validators with both PublicKey and PublicKeyBytes populated (optimized)
```
JonathanOppenheimer: Now that you can see the results yourself, should I remove/revert this? I don't really see any useful reason to have this

snow/validators/state_test.go
```diff
@@ -232,7 +232,16 @@ func TestCachedState_GetWarpValidatorSet_Inactive(t *testing.T) {
 
 			got, err := cached.GetWarpValidatorSet(t.Context(), height, subnetID)
 			require.ErrorIs(err, test.wantErr)
-			require.Equal(test.want, got)
+
+			// Compare only the public fields, not cache fields
```
JonathanOppenheimer: This one too as well

snow/validators/validator.go
```diff
@@ -25,9 +25,10 @@ type Validator struct {
 // GetValidatorOutput is a struct that contains the publicly relevant values of
 // a validator of the Avalanche Network for the output of GetValidator.
 type GetValidatorOutput struct {
-	NodeID    ids.NodeID
-	PublicKey *bls.PublicKey
-	Weight    uint64
+	NodeID         ids.NodeID
+	PublicKey      *bls.PublicKey
```
yacovm: So, why should we still have the `PublicKeyBytes []byte` field if all we use is the `PublicKey *bls.PublicKey` field? 

We barely use the public key bytes now:

image

Why don't we just only populate the `PublicKey *bls.PublicKey` ? 

Then we don't need that `sync.Once` lazy initialization and we can initialize it only once at creation. 

Thoughts?

JonathanOppenheimer: No this is fair, if we want to keep the structure minimal, it doesn't make sense to add another field that is rarely used.

vms/platformvm/warp/validator_test.go
JonathanOppenheimer: this is outdated, now but you can still run the benchmark yourself on this branch vs. master.

snow/validators/validator.go
JonathanOppenheimer: This aligns the two but the warp structure does still rely on a `PublicKeyBytes []byte`

snow/validators/warp.go
```diff
@@ -51,11 +52,22 @@ func (w *WarpSet) UnmarshalJSON(b []byte) error {
...
 
 type Warp struct {
-	PublicKey *bls.PublicKey
 	// PublicKeyBytes is expected to be in the uncompressed form.
 	PublicKeyBytes []byte
 	Weight         uint64
 	NodeIDs        []ids.NodeID
+
+	// Cache for converted public key (lazy initialization)
+	publicKeyCache *bls.PublicKey
```
yacovm, line 61: I think we can just have the public key as an object and not bother with this lazy initialization. Since we only initialize the PK in a single place anyway.

JonathanOppenheimer: Done!

snow/validators/validator.go
JonathanOppenheimer: Discussed over slack, decided to try removing the lazy initialization for warp.

JonathanOppenheimer:
@yacovm the eager initialization did indeed tank the performance:

  **BenchmarkFlattenValidatorSet (Validator Set Creation)**

  | Size | Master (ns/op) | Optimized (ns/op) | Diff % | Master (B/op) | Optimized (B/op) | Master (allocs/op) | Optimized (allocs/op) |
  |------|----------------|-------------------|--------|---------------|------------------|--------------------|-----------------------|
  | 1    | 304.2          | 386.9             | +27.2% | 288           | 384              | 5                  | 6                     |
  | 2    | 459.6          | 625.0             | +36.0% | 576           | 768              | 9                  | 11                    |
  | 4    | 783.8          | 1111              | +41.7% | 1152          | 1536             | 17                 | 21                    |
  | 8    | 1367           | 2089              | +52.8% | 2304          | 3072             | 33                 | 41                    |
  | 16   | 2972           | 4350              | +46.4% | 5544          | 7080             | 68                 | 84                    |
  | 32   | 5909           | 8745              | +48.0% | 11048         | 14120            | 132                | 164                   |
  | 64   | 12103          | 17868             | +47.6% | 21928         | 28072            | 260                | 324                   |
  | 128  | 25477          | 37114             | +45.7% | 43560         | 55849            | 516                | 644                   |
  | 256  | 53187          | 76320             | +43.5% | 87593         | 112170           | 1028               | 1284                  |
  | 512  | 116610         | 155932            | +33.7% | 175532        | 224684           | 2052               | 2564                  |
  | 1024 | 241062         | 333267            | +38.2% | 350808        | 449116           | 4102               | 5126                  |

  **BenchmarkGetCanonicalValidatorSet (Full Validator Set Retrieval)**

  | Size  | Master (ns/op) | Optimized (ns/op) | Diff % | Master (B/op) | Optimized (B/op) | Master (allocs/op) | Optimized (allocs/op) |
  |-------|----------------|-------------------|--------|---------------|------------------|--------------------|-----------------------|
  | 0     | 97.27          | 98.02             | +0.8%  | 0             | 0                | 0                  | 0                     |
  | 1     | 321.0          | 408.4             | +27.2% | 288           | 384              | 5                  | 6                     |
  | 10    | 2186           | 3073              | +40.6% | 3336          | 4296             | 44                 | 54                    |
  | 100   | 22286          | 30736             | +37.9% | 32393         | 41993            | 404                | 504                   |
  | 1000  | 252871         | 340886            | +34.8% | 342809        | 438812           | 4006               | 5006                  |
  | 10000 | 2939377        | 3850643           | +31.0% | 3318845       | 4278849          | 40035              | 50036                 |

  
  However maybe we are okay with these tradeoffs, as we do
  - More upfront work and memory during validator set creation
  - For faster, simpler access during signature operations

JonathanOppenheimer:
I can also revert to 4f55cf3 for the original performance gains. I guess the larger question is does IRL usage mimic the benchmark or nah.

JonathanOppenheimer:
I added a new benchmark `BenchmarkAggregatePublicKeys` to give some visibility into the trade offs with choices in various scenarios. I compared the changes on 3de36f8 (which is my first attempt at optimization which leads to performance increases on the other two benchmarks) to the current tip of this branch, which is post Yacov's suggestions, which more closely resembles master without lazy initialization. 

  | Size   | Pre-Optimization (3de36f8)ns/op | Optimized (Current)ns/op | Improvement | Pre-OptB/op | OptimizedB/op | Pre-Optallocs/op | Optimizedallocs/op |
  |--------|---------------------------------|--------------------------|-------------|-------------|---------------|------------------|--------------------|
  | 1      | 922.6                           | 916.5                    | -0.7%    | 576         | 576           | 10               | 10                 |
  | 10     | 12,855                          | 12,662                   | -1.5%     | 3,528       | 3,528         | 37               | 37                 |
  | 100    | 63,620                          | 63,993                   | +0.6%       | 5,626       | 5,625         | 49               | 49                 |
  | 1,000  | 547,311                         | 536,762                  | -1.9%     | 12,967      | 12,923        | 49               | 49                 |
  | 10,000 | 5,226,422                       | 5,230,765                | +0.1%       | 90,824      | 86,652        | 92               | 49                 |

JonathanOppenheimer:
Performance Results (ns/op)

  | Size   | Baseline (4f55cf3)PublicKeyBytes + Lazy | After First Opt (3de36f8)No PublicKeyBytes + Lazy | Final (Current)No PublicKeyBytes + Eager | Baseline→First | First→Final | Total Improvement |
  |--------|-----------------------------------------|---------------------------------------------------|------------------------------------------|----------------|-------------|-------------------|
  | 1      | 901.2                                   | 922.6                                             | 916.5                                    | -2.4%        | +0.7%     | -1.7%           |
  | 10     | 13,063                                  | 12,855                                            | 12,662                                   | +1.6%        | +1.5%     | +3.1%           |
  | 100    | 63,658                                  | 63,620                                            | 63,993                                   | +0.1%        | -0.6%     | -0.5%           |
  | 1,000  | 545,207                                 | 547,311                                           | 536,762                                  | -0.4%        | +1.9%     | +1.5%           |
  | 10,000 | 5,083,804                               | 5,226,422                                         | 5,230,765                                | -2.8%       | -0.1%     | -2.9%           |

JonathanOppenheimer:
My vote: revert all changes to [4f55cf3](https://github.com/ava-labs/avalanchego/pull/4784/commits/4f55cf365c212c570d6a8b54b9a0bd9d72f7cddd), but keep the new benchmark.

yacovm:
We should optimize the transaction validation path, so let's revert my idea 👍

JonathanOppenheimer:
> We should optimize the transaction validation path, so let's revert my idea 👍

Reverted!

snow/validators/state_test.go
```diff
@@ -232,7 +232,16 @@ func TestCachedState_GetWarpValidatorSet_Inactive(t *testing.T) {
...
+			require.Equal(test.want.TotalWeight, got.TotalWeight)
+			require.Len(got.Validators, len(test.want.Validators))
+			for i, wantVdr := range test.want.Validators {
+				gotVdr := got.Validators[i]
```
yacovm: You can do this too:

```
				gotVdr.PublicKey = nil
				wantVdr.PublicKey = nil
				require.Equal(wantVdr, gotVdr)
```

vms/platformvm/state/l1_validator.go
```diff
@@ -180,10 +180,10 @@ func (v L1Validator) effectiveNodeID() ids.NodeID {
 }
 
 func (v L1Validator) effectivePublicKey() *bls.PublicKey {
-	if v.IsActive() {
-		return bls.PublicKeyFromValidUncompressedBytes(v.PublicKey)
+	if !v.IsActive() {
```
yacovm, line 183: not that it's important but why the order change?

yacovm (approved):

snow/validators/state_test.go
JonathanOppenheimer: Screenshot 2025-12-31 at 11 25 35 AM

We have to set the bytes directly, which will fail the test.

vms/platformvm/state/l1_validator.go
JonathanOppenheimer: It just matches the early return pattern others have stressed.

joshua-kim:
Which conventional commits spec are we using? Under some this isn't a `feat` and this is considered `perf`

JonathanOppenheimer:
> Which conventional commits spec are we using? Under some this isn't a `feat` and this is considered `perf`

you're right - it should be perf. I just started doing it a couple months ago, and have been using these: 

> types other than fix: and feat: are allowed, for example [@commitlint/config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional) (based on the [Angular convention](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines)) recommends build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, and others.

snow/validators/warp.go
alarso16, line 131: Isn't this always true because you got rid of the field? So it refers to the function

alarso16 (commented):
EVM code looks fine, but I would check for more mistakes like the one I noticed

snow/validators/warp.go
yacovm: This is a field of `GetValidatorOutput`, not of `Warp`.

alarso16 (approved):
I've been corrected, seems fine

JonathanOppenheimer:
Closing due to staleness - linked to issue so can always find if necessary.

PR #4786, opened 2025-12-22
refactor: share EVM rpc package

JonathanOppenheimer:
## Why this should be merged

The subnet-evm `rpc` package was practically identical to the coreth `rpc` package. It makes no sense to maintain two copies of this code. This also makes moving to libevm's `rpc` package far easier, to look in one location and see what needs to be libevmified. 

## How this works

  - Created `graft/evm/rpc/` directory structure
  - Copied all source files from `coreth/rpc` to the shared location
  - Updated graft/evm/go.mod with required dependencies:
    - `github.com/gorilla/websocket v1.5.0`
    - `github.com/deckarep/golang-set/v2 v2.1.0`
    - `golang.org/x/time v0.12.0`
  - Fixed test imports and restored `TestNotify` to original libevm version
  - Removed `test_main.go` for registration during testing.   
  
  <details>
  <summary>Outdated information</summary>
  I also added `main_test.go` with a `TestMain` that automatically runs all RPC tests twice - once with C-Chain type registration and once with Subnet-EVM type registration.

This works by putting `TestMain` in `package rpc_test` (an external test package) instead of `package rpc`, which lets us import the `emulate` package without creating a circular dependency. We can then use `emulate.CChain()` and `emulate.SubnetEVM()` to properly register/cleanup types between test runs.

Thus when you run `go test` (which also automatically happens via CI -- you can check for yourself!), it:

1.   Runs all tests with C-Chain types registered via `emulate.CChain()`
2.   Cleans up the registration
3.   Runs all tests again with Subnet-EVM types via `emulate.SubnetEVM()`
4.   Reports if either variant fails
</details>
  

## How this was tested
CI 

No

### DIFF
--- a/graft/coreth/go.sum
+++ b/graft/coreth/go.sum
@@ -5,8 +5,8 @@ connectrpc.com/grpcreflect v1.3.0 h1:Y4V+ACf8/vOb1XOc251Qun7jMB75gCUNw6llvB9csXc
 connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JIovCtDYs=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
 github.com/DataDog/zstd v1.5.2 h1:vUG4lAyuPCXO0TLbXvPv7EB7cNK1QV/luu55UHLrrn8=
--- a/graft/evm/go.mod
+++ b/graft/evm/go.mod
@@ -20,7 +20,7 @@ require (
 )
 
 require (
-	github.com/BurntSushi/toml v1.3.2 // indirect
+	github.com/BurntSushi/toml v1.5.0 // indirect
 	github.com/DataDog/zstd v1.5.2 // indirect
 	github.com/Microsoft/go-winio v0.6.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -43,8 +43,8 @@ connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JI
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
--- a/graft/evm/go.sum
+++ b/graft/evm/go.sum
@@ -1,8 +1,8 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
 github.com/DataDog/zstd v1.5.2 h1:vUG4lAyuPCXO0TLbXvPv7EB7cNK1QV/luu55UHLrrn8=
--- a/graft/subnet-evm/go.sum
+++ b/graft/subnet-evm/go.sum
@@ -5,8 +5,8 @@ connectrpc.com/grpcreflect v1.3.0 h1:Y4V+ACf8/vOb1XOc251Qun7jMB75gCUNw6llvB9csXc
 connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JIovCtDYs=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
 github.com/DataDog/zstd v1.5.2 h1:vUG4lAyuPCXO0TLbXvPv7EB7cNK1QV/luu55UHLrrn8=


JonathanOppenheimer:
> Don't we need to register the types for our tests? I think that was the only thing preventing us from using the geth rpc package

The only place we registered was in an additional main file just in subnet-evm, and that was for test retrying for flakes. I removed that as we should not be doing test retries in the monorepo, and should just skip flaky tests entirely if they're flaky. Is there registration anywhere else?

alarso16:
> > Don't we need to register the types for our tests? I think that was the only thing preventing us from using the geth rpc package
> 
> The only place we registered was in an additional main file just in subnet-evm, and that was for test retrying for flakes. I removed that as we should not be doing test retries in the monorepo, and should just skip flaky tests entirely if they're flaky. Is there registration anywhere else?

No, we didn't just retry failed tests. They also registered `customtypes` in each respective repo, so that's why the contents of the block changed when you moved them over. I'm not sure if we still need to be testing that functionality

JonathanOppenheimer:
> > > Don't we need to register the types for our tests? I think that was the only thing preventing us from using the geth rpc package
> > 
> > 
> > The only place we registered was in an additional main file just in subnet-evm, and that was for test retrying for flakes. I removed that as we should not be doing test retries in the monorepo, and should just skip flaky tests entirely if they're flaky. Is there registration anywhere else?
> 
> No, we didn't just retry failed tests. They also registered `customtypes` in each respective repo, so that's why the contents of the block changed when you moved them over. I'm not sure if we still need to be testing that functionality

I don't think this is true. There are two places in this PR where I removed ` customtypes.Register()`.

1. [For a test retry loop](https://github.com/ava-labs/avalanchego/blob/10f374649645f3093023633d60825875c5645589/graft/coreth/rpc/main_test.go). As explained above, I this this is unnecessary. 
2. In `TestMain` of subnet-evm's [subscription test](https://github.com/ava-labs/avalanchego/blob/10f374649645f3093023633d60825875c5645589/graft/subnet-evm/rpc/subscription_test.go). This differed from coreth, and only appeared to be present as we were still importing the old custom types instead of the libevm types package. I removed it to align the two, as I assume subnet-evm just didn't get updated. 

Are there other important places in which I removed it?

ARR4N:
If `customtypes.Register()` isn't called then the test isn't testing `coreth` / `subnet-evm` functionality in its entirety. I'd recommend using `evm.RegisterLibEVMExtras()` (which calls `customtypes.Register()`) instead of the individual registration as this guarantees that _all_ hooks/payloads are registered.

This can alternately be done with `emulate.CChain()` and `emulate.SubnetEVM()` if you need to test both in the same package (i.e. the same test binary). If there is commonality between the testing then the best approach is probably something like:

```go
func testCChainAndSubnetEVM(t *testing.T, testFn func(*testing.T)) {
  fn := func() error {
    testFn(t)
    return nil
  }

  t.Run("cchain", func(t *testing.T) {
    _ = emulate.CChain(fn)
  })
  t.Run("subnetEVM", func(t *testing.T) {
    _ = emulate.SubnetEVM(fn)
  })
}

func TestFoo(t *testing.T) {
  testCorethAndSubnetEVM(t, testFoo)
}

func testFoo(t *testing.T) {
  // this is where the actual testing occurs; i.e. just rename any existing TestFoo to testFoo
}

func TestBar(t *testing.T) {
  testCorethAndSubnetEVM(t, testBar)
}

func testBar(t *testing.T) {
  // as with testFoo()
}
```

The linter is going to tell you to call `t.Helper()` but it's better not to in this case otherwise all errors will be reported as being on the single line of the `TestFoo()` implementation, instead of on the specific line in `testFoo()`.

alarso16:
I believe we have decided to enforce that we don't want cyclic dependencies between `coreth` and `evm`, so we would probably have to "uplift" the customtypes packages first. Also, I have tried this in #4688, and trying to register all EVM imports in this package causes a cyclic dependency (unless you make a test package?)

JonathanOppenheimer:
> I believe we have decided to enforce that we don't want cyclic dependencies between `coreth` and `evm`, so we would probably have to "uplift" the customtypes packages first. Also, I have tried this in #4688, and trying to register all EVM imports in this package causes a cyclic dependency (unless you make a test package?)

Yeah I am trying this right now and was about to comment about the cyclic imports. The shared `graft/evm/rpc` package is a dependency used by coreth, which creates a circular dependency if we try to use `vms/evm/emulate` (which depends on coreth) in the RPC tests. This can be avoided if we just call the custom types registration directly or put it in a seperate test package. 

I definitely think this + in conjunction with some sort of test main would be a good model for how to combine packages while testing separately.  I've committed a new `main_test.go` which does this to some extent although simpler. What are the thoughts on this approach?

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// The EVM_VARIANT environment variable can override this behavior:
+//   - "cchain": Run tests only with C-Chain type registration
+//   - "subnetEVM": Run tests only with Subnet-EVM type registration
+//   - "both" or unset: Run tests twice with both variants (default)
+//
+// This uses the emulate package's temporary registration mechanism, which properly
+// cleans up after each test run, allowing us to test both variants in sequence.
+func TestMain(m *testing.M) {
+	variant := os.Getenv("EVM_VARIANT")
```
alarso16: I think being able to set this is unnecessary. Running for both coreth and subnet-evm is required for the code to be functional

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if variant == "" {
+		variant = "both"
+	}
+
+	var exitCode int
+
+	switch variant {
```
alarso16: A lot of this code will be used for any other packages that need to import the registrations. Can you extract this into a test helper package? maybe something like `libevmtest.RunWithAll(*testing.M)`

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: Here is the critical file to be reviewed.

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+		fmt.Println("=== Running RPC tests with both C-Chain and Subnet-EVM ===")
+
+		fmt.Println("--- Running C-Chain variant ---")
+		cchainCode := runWith(m, emulate.CChain)
+
+		fmt.Println("\n--- Running Subnet-EVM variant ---")
+		subnetCode := runWith(m, emulate.SubnetEVM)
+
+		if cchainCode != 0 {
+			fmt.Fprintf(os.Stderr, "\n✗ C-Chain tests failed with exit code %d\n", cchainCode)
```
alarso16: Is the code informative?

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+			exitCode = cchainCode
+		} else if subnetCode != 0 {
```
alarso16: This definitely should also be printed if the coreth tests fails

graft/evm/rpc/subscription_test.go
```diff
@@ -296,8 +288,16 @@ func TestNotify(t *testing.T) {
...
+
+	// Check if the output matches either C-Chain or Subnet-EVM format
+	if have != wantCChain && have != wantSubnetEVM {
```
alarso16: I'm not sure if this is ok. We might need to set an atomic global to know whether it's coreth or subnet-evm. What if we always return the subnet-evm value? The test would pass, but the code wouldn't work

graft/evm/rpc/main_test.go
JonathanOppenheimer: I will remove the functionality for a user to be able to specify which variant of the test they want to run.

graft/evm/rpc/main_test.go
JonathanOppenheimer: yup!

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: Here is the critical file to review

graft/evm/rpc/notify_test.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: This test had to be moved to a separate file because of an import cycle.

graft/evm/rpc/testing.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: This is also new

graft/evm/rpc/main_test.go
JonathanOppenheimer: Wdym by informative? It's just the code.

graft/evm/rpc/subscription_test.go
JonathanOppenheimer: Yup, we can check the atomic global you recommended to see which json to match against.

graft/evm/rpc/main_test.go
alarso16: Sorry, to be clear I mean the value returned by `m.Run`. Don't we only care if it's 0 or non-zero?

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,87 @@
...
+	UnknownVariant Variant = iota
+	// CChainVariant indicates C-Chain types are registered.
+	CChainVariant
+	// SubnetEVMVariant indicates Subnet-EVM types are registered.
+	SubnetEVMVariant
+)
+
+// currentVariant tracks which EVM variant is currently registered.
+// This allows tests to check which variant they're running under.
+var currentVariant atomic.Int32
```
alarso16: This needs to be at a per-package level - otherwise it's prone to misuse. By default, ` go test ./...` runs up to `GOMAXPROCS` packages in parallel (which is normally not 1). That means multiple readers and writers

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,87 @@
...
+
+// CurrentVariant returns which EVM variant is currently registered.
+// This is safe to call from tests running in parallel.
+func CurrentVariant() Variant {
+	return Variant(currentVariant.Load())
```
alarso16: You can make this typed with avalanchego's `utils.Atomic`

graft/evm/rpc/main_test.go
JonathanOppenheimer: Oh yeah you're just saying we don't need to print it. Yeah I can remove the print

graft/evm/libevmtest/testmain.go
alarso16: It might actually make more sense to provide the atomic to the runner so then you use it like this:
```
var currentVariant utils.Atomic[Variant]

func TestMain(m *testing.M) {
     RunnerWithAll(m, &currentVariant)
}
```

Not totally positive about this yet. Really, you only want to expose the `Load` functionality to the test user, but this would work

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,87 @@
...
+	} else {
+		fmt.Println("Subnet-EVM tests passed")
+	}
+
+	if cchainCode != 0 {
+		return cchainCode
+	}
+	if subnetCode != 0 {
+		return subnetCode
+	}
```
alarso16:
```suggestion
	if cchainCode != 0 || subnetCode != 0 {
	    return 1
	}
```
Don't just apply this, but is there any reason we care about the specific code?

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,87 @@
...
+
+	fmt.Println("\nAll tests passed with both variants")
```
alarso16:
```suggestion
```
I think this is a little too verbose

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,87 @@
...
+// RunWithAll runs all tests with both C-Chain and Subnet-EVM type registration.
+// Tests are run twice - once with C-Chain types and once with Subnet-EVM types.
+// Both test runs must pass for the overall result to be successful.
+//
+// This function sets a global variable that tests can check via CurrentVariant()
+// to determine which variant is currently registered.
+func RunWithAll(m *testing.M) int {
+	fmt.Println("=== Running tests with both C-Chain and Subnet-EVM ===")
+
+	fmt.Println("\n--- Running C-Chain variant ---")
```
alarso16: Why do you prefix all these with newlines?

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: Not really, -- more work could be done to determine which of the two test variants failed, by returning extra information in these if cases.

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: So there can be a blank line in between these two lines.

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: This seems like it would work to me.

JonathanOppenheimer: We're rolling with it

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: I am going to apply this but leave the comment open if you have further comments.

JonathanOppenheimer: I just learned Go does not have a ternary operator. Lame. 5 lines could be one!

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: it's just for style

alarso16: I think you shouldn't. Run `go test -v ./...` to see how go does it (there's no blank lines)

graft/evm/go.mod
```diff
@@ -3,63 +3,129 @@ module github.com/ava-labs/avalanchego/graft/evm
...
 )
 
 require (
-	github.com/DataDog/zstd v1.4.5 // indirect
-	github.com/StackExchange/wmi v1.2.1 // indirect
+	github.com/DataDog/zstd v1.5.2 // indirect
+	github.com/Microsoft/go-winio v0.6.1 // indirect
+	github.com/StephenButtolph/canoto v0.17.3 // indirect
 	github.com/VictoriaMetrics/fastcache v1.12.1 // indirect
+	github.com/ava-labs/avalanchego/graft/coreth v0.0.0-20251203215505-70148edc6eca // indirect
```
alarso16: Open question: Will this indirect dependency cause a giant headache down the road?

JonathanOppenheimer: I can't foresee any issue - do you have something in mind. It's because of importing emulate which imports each of the two repositories.

alarso16: The reason that we didn't want to allow `evm` to import `coreth` is because of the dependency tree. Even if it doesn't import it directly, this doesn't seem to avoid the problem

JonathanOppenheimer: I thought we didn't want to allow `graft/evm` to import `coreth` because it would prevent a direct uplift to `vms/evm`. An indirect dependency (esp that is a result of importing `vms/evm` itself) doesn't prevent tthat. What are your concerns regarding the dependency tree?

graft/evm/rpc/notify_test.go
alarso16, line 1: Why is this in a separate file? You should be able to just keep the test there, since any conflicts when updating will be easier to see

graft/evm/rpc/notify_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	switch currentVariant.Get() {
+	case libevmtest.CChainVariant:
+		require.Equal(t, wantCChain, have)
+	case libevmtest.SubnetEVMVariant:
+		require.Equal(t, wantSubnetEVM, have)
+	default:
+		require.FailNow(t, "Unknown or unregistered variant (this test must run via TestMain)")
```
alarso16: The comment in the parentheses doesn't really make sense. I believe any way that you run tests will go through `TestMain`
```suggestion
		require.FailNow(t, "libevm extras not registered (use libevmtest to aid tracking registrations)")
```
I don't love this wording still, but maybe this makes more sense

graft/evm/rpc/notify_test.go
JonathanOppenheimer: Because it has to be in a separate package, `rpc_test` to prevent import cycles.

JonathanOppenheimer: and if we change the whole old test file to be in `rpc_test` it doesn't work as a tonnn more stuff would either need to get exported to be accessible, or moved. This was the minimal change I could make.

JonathanOppenheimer: This is also why https://github.com/ava-labs/avalanchego/blob/72ea713ecc8af00b4264ec28bfe32cb054d622f4/graft/evm/rpc/testing.go exists now.

alarso16: Ahhh, I didn't realize that you didn't change all of the tests to run in that package.

alarso16 (commented):
This works, but some of the package management stuff feels hacky. I don't have any better solutions, so maybe get an opinion from @ARR4N or @ceyonur. I'm not confident enough that this is the route we want to take to give a real approval

My concerns are:
- Importing coreth and subnet-evm was intended to be avoided in this package, but it has to be done to enable registrations. Part of this messiness is because we decided not to directly import coreth and subnet-evm, but it would be simpler if we could (see @ARR4N's suggestion above)
- This makes the dependency tree between the four modules a strongly connected graph, which is less than ideal. Would it be better to move the coreth/subnet-evm type registrations into the `evm` package first?
- Running two separate packages of tests in the same folder is quite gross (tbh didn't even know Go supports it)

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,71 @@
...
+	SubnetEVMVariant
+)
+
+// RunWithAll runs all tests with both C-Chain and Subnet-EVM type registration.
+// Tests are run twice - once with C-Chain types and once with Subnet-EVM types.
+// Both test runs must pass for the overall result to be successful.
+//
+// The currentVariant atomic is set before each test run, allowing tests to check
+// which variant they're running under via currentVariant.Get().
+func RunWithAll(m *testing.M, currentVariant *utils.Atomic[Variant]) int {
```
ARR4N: There's no need for an atomic because the emulation functions hold a global mutex. Additionally, if you just create a `currentVariant` global in this package and expose a `CurrentVariant() Variant` function then you don't force every user to do it.

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,71 @@
...
+func runWith(m *testing.M, currentVariant *utils.Atomic[Variant], emulateFn func(func() error) error, variant Variant) int {
+	var code int
+	currentVariant.Set(variant)
+	_ = emulateFn(func() error {
+		code = m.Run()
+		return nil
+	})
+	currentVariant.Set(UnknownVariant)
+	return code
+}
```
ARR4N:
```suggestion
var current Variant

// CurrentVariant returns the currently emulated EVM variant. Note that
// if emulation was initiated outside of this package then [UnknownVariant]
// will be returned.
func CurrentVariant() Variant {
    return current
}

// runWith executes tests with the specified emulation function and variant.
func runWith(m *testing.M, emulateFn func(func() error) error, variant Variant) int {
	var code int
	_ = emulateFn(func() error {
	    current = variant
	    defer func() { current = UnknownVariant }()
		code = m.Run()
		return nil
	})
	return code
}
```

graft/evm/libevmtest/testmain.go
```diff
@@ -0,0 +1,71 @@
...
+		fmt.Fprintln(os.Stderr, "Subnet-EVM tests failed")
+	} else {
+		fmt.Println("Subnet-EVM tests passed")
+	}
+
+	if cchainCode != 0 || subnetCode != 0 {
+		return 1
+	}
+
+	return 0
```
ARR4N:
```suggestion

	return cChainCode | subnetCode
```

graft/evm/libevmtest/testmain.go
ARR4N: One downside of using `testing.M` instead of `testing.T` is that you can't "namespace" the tests with `t.Run()`. That said, it's beneficial to not have to call a separate function for every single test, so I understand why you're doing it like this.

alarso16: I thought Go parallelized packages during testing. Wouldn't the global make a race if there's multiple packages?

maru-ava: Each package runs in its own process. A blessing and a curse, it complicates sharing expensive fixture across packages. That's why we use ginkgo for use cases requiring expensive fixture.

JonathanOppenheimer: If that's the case, I will revert the changes back to a `currentVariant` global!

JonathanOppenheimer: Yeah there's a very large number of tests, so I think doing otherwise would be bloat inducing.

graft/evm/libevmtest/testmain.go
JonathanOppenheimer: Alright I removed them.

graft/evm/go.mod
JonathanOppenheimer: This was discussed on call / in slack -- this is okay.

graft/evm/scripts/upstream_files.txt
```diff
@@ -0,0 +1,5 @@
+rpc/*
+
+!rpc/main_test.go
+!rpc/notify_test.go
```
alarso16: `notify_test.go` and `testing.go` are upstream code, so maybe it would be better to have them licensed as such

JonathanOppenheimer: good point, I fixed this

ARR4N (commented):
`libevmtest` package LGTM

> [!WARNING]
> Reminder: the `emulate` package MUST NOT be used in production in a node. It is only for tests and consumers like RPC clients and data ingestion.

JonathanOppenheimer:
> `libevmtest` package LGTM
> 
> Warning
> 
> Reminder: the `emulate` package MUST NOT be used in production in a node. It is only for tests and consumers like RPC clients and data ingestion.

Would you like that a comment somewhere?

alarso16:
> Would you like that a comment somewhere?

I personally don't think that's necessary. Because `libevmtest` contains `test` in the name, the linter enforces that it can only be imported in test code

JonathanOppenheimer:
> > Would you like that a comment somewhere?
> 
> I personally don't think that's necessary. Because `libevmtest` contains `test` in the name, the linter enforces that it can only be imported in test code

Got it makes sense. @ARR4N is there anything else from this PR you'd like to see.

graft/evm/rpc/notify_test.go
```diff
@@ -0,0 +1,70 @@
...
+	out := new(bytes.Buffer)
+	id := rpc.ID("test")
+	notifier := rpc.NewTestNotifier(out, id)
+
+	msg := &types.Header{
+		ParentHash: common.HexToHash("0x01"),
+		Number:     big.NewInt(100),
+	}
+	require.NoError(t, notifier.Notify(id, msg))
+	have := strings.TrimSpace(out.String())
```
ARR4N: `got` and `want` are the idiomatic Go names.

graft/evm/rpc/notify_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestNotify(t *testing.T) {
+	out := new(bytes.Buffer)
+	id := rpc.ID("test")
+	notifier := rpc.NewTestNotifier(out, id)
+
+	msg := &types.Header{
+		ParentHash: common.HexToHash("0x01"),
+		Number:     big.NewInt(100),
+	}
+	require.NoError(t, notifier.Notify(id, msg))
```
ARR4N: If you trace the path of what this does, relative to what's being asserted, it only tests that `msg` is properly JSON encoded. Everything else is just a convoluted way of getting the JSON into the `bytes.Buffer`, and the `Notifier.send()` method is already tested by the upstream code. Since we already know that `libevm` properly JSON encodes after registration I think `TestNotify()` is redundant.

graft/evm/rpc/main_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"os"
+	"testing"
+
+	"github.com/ava-labs/avalanchego/graft/evm/libevmtest"
+)
+
+// TestMain runs all tests in the rpc package with proper EVM type registration.
+// Tests are run twice - once with C-Chain types and once with Subnet-EVM types.
+// Both test runs must pass for the overall result to be successful.
+func TestMain(m *testing.M) {
```
ARR4N: Was this solely to ensure that `TestNotify()` was tested with both variants? If so, then why use `testing.M` instead of just the local `testing.T` as I originally described? If not, what else are you trying to test under both emulation variants?

JonathanOppenheimer: This does actually test `client_opt_test.go` as well, which was already in `rpc_test` for some reason. -- `testing.M` was used to prevent having multiple test wrappers. I have kept it this way in 795c9d3, so we can test all the `RPC` tests under registration.

graft/evm/rpc/notify_test.go
JonathanOppenheimer: Got it! We can delete this

graft/evm/rpc/notify_test.go
JonathanOppenheimer: This is a `geth` test, but it's been removed in entirety anway.

graft/evm/rpc/subscription_test.go
```diff
@@ -288,7 +288,7 @@ func TestNotify(t *testing.T) {
...
...00000000000000","stateRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","transactionsRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","receiptsRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","difficulty":null,"number":"0x64","gasLimit":"0x0","gasUsed":"0x0","timestamp":"0x0","extraData":"0x","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","baseFeePerGas":null,"withdrawalsRoot":null,"blobGasUsed":null,"excessBlobGas":null,"parentBeaconBlockRoot":null,"hash":"0xe5fb877dde471b45b9742bb4bb4b3d74a761e2fb7cb849a3d2b687eed90fb604"}}}`
```
JonathanOppenheimer, line 291: @ceyonur suggested bring back the original TestNotify, rather than deleting it entirely, when testing without registration.

graft/evm/rpc/testing.go
```diff
@@ -9,7 +9,7 @@
 //
```
JonathanOppenheimer, line 9: GitHub is drunk showing this move...

JonathanOppenheimer:
After extensive discussion, we have decided not to test under registration.

graft/coreth/go.mod
```diff
@@ -8,15 +8,13 @@ module github.com/ava-labs/avalanchego/graft/coreth
 go 1.24.12
 
 require (
-	github.com/ava-labs/avalanchego v1.14.1-0.20251120155522-df4a8e531761
+	github.com/ava-labs/avalanchego v1.14.1-antithesis-docker-image-fix
```
alarso16: Why is this updated?

graft/coreth/go.sum
```diff
@@ -5,8 +5,8 @@ connectrpc.com/grpcreflect v1.3.0 h1:Y4V+ACf8/vOb1XOc251Qun7jMB75gCUNw6llvB9csXc
 connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JIovCtDYs=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
-github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
+github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
```
alarso16: downgrade?

graft/evm/rpc/client_opt_test.go
alarso16, line 1: nit: This increases diff from geth. I'm not sure why it's like that, but you should probably leave it

graft/evm/scripts/upstream_files.txt
```diff
@@ -1,4 +1,6 @@
+rpc/*
 core/*
 triedb/*
 
-!core/state/snapshot/snapshot_ext.go
\ No newline at end of file
+!core/state/snapshot/snapshot_ext.go
+!rpc/main_test.go
```
alarso16: I don't think this file exists
```suggestion
```

graft/evm/go.mod
```diff
@@ -92,4 +99,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/ava-labs/avalanchego => ../../
+replace (
+	github.com/ava-labs/avalanchego => ../../
+	github.com/ava-labs/avalanchego/graft/coreth => ../coreth
```
alarso16: You don't have to import coreth or subnet-evm

graft/subnet-evm/accounts/abi/bind/precompilebind/precompile_bind_test.go
```diff
@@ -694,6 +694,11 @@ func TestPrecompileBind(t *testing.T) {
 	out, err = replacer.CombinedOutput()
 	require.NoError(t, err, "failed to replace binding test dependency to current source tree: %v\n%s", err, out)
 
+	replacer = exec.Command(gocmd, "mod", "edit", "-x", "-require", "github.com/ava-labs/avalanchego/graft/coreth@v0.0.0", "-replace", "github.com/ava-labs/avalanchego/graft/coreth="+filepath.Join(pwd, "..", "..", "..", "..", "..", "coreth"))
```
alarso16: I don't think this is necessary

JonathanOppenheimer: removed

graft/evm/go.mod
JonathanOppenheimer: removed

graft/evm/rpc/client_opt_test.go
JonathanOppenheimer: reverted.

graft/coreth/go.mod
JonathanOppenheimer: This is a result of running `go mod tidy`

graft/coreth/go.sum
JonathanOppenheimer: This is a result of running `go mod tidy`

graft/coreth/go.mod
```diff
@@ -183,6 +182,8 @@ tool (
 	github.com/onsi/ginkgo/v2/ginkgo
 )
 
-replace github.com/ava-labs/avalanchego => ../../
-
-replace github.com/ava-labs/avalanchego/graft/evm => ../evm
+replace (
+	github.com/ava-labs/avalanchego => ../../
+	github.com/ava-labs/avalanchego/graft/evm => ../evm
+	github.com/ava-labs/avalanchego/graft/subnet-evm => ../subnet-evm
```
alarso16: subnet-evm isn't imported (even transitively), so you can delete this

JonathanOppenheimer: Removed

graft/coreth/go.sum
JonathanOppenheimer: fixed

graft/coreth/go.mod
JonathanOppenheimer: fixed

Issue #4757, opened 2025-12-16
Fix `TestWaitForEvent` flake


StephenButtolph:
See: https://github.com/ava-labs/avalanchego/actions/runs/20283037548/job/58249847057?pr=4755#step:4:294
```
--- FAIL: TestWaitForEvent (0.27s)
    --- FAIL: TestWaitForEvent/WaitForEvent_build_block_after_re-org (2.04s)
        vm_test.go:2995: 
            	Error Trace:	/Users/runner/work/avalanchego/avalanchego/graft/subnet-evm/plugin/evm/vm_test.go:2995
            	            				/Users/runner/work/avalanchego/avalanchego/graft/subnet-evm/plugin/evm/vm_test.go:3152
            	Error:      	Received unexpected error:
            	            	context deadline exceeded
            	Test:       	TestWaitForEvent/WaitForEvent_build_block_after_re-org
+-----------------+-------------------------+--------+-------+
|    DATABASE     |        CATEGORY         |  SIZE  | ITEMS |
+-----------------+-------------------------+--------+-------+
| Key-Value store | Headers                 | 0.00 B |     0 |
| Key-Value store | Bodies                  | 0.00 B |     0 |
| Key-Value store | Receipt lists           | 0.00 B |     0 |
| Key-Value store | Difficulties            | 0.00 B |     0 |
| Key-Value store | Block number->hash      | 0.00 B |     0 |
| Key-Value store | Block hash->number      | 0.00 B |     0 |
| Key-Value store | Transaction index       | 0.00 B |     0 |
| Key-Value store | Bloombit index          | 0.00 B |     0 |
| Key-Value store | Contract codes          | 0.00 B |     0 |
| Key-Value store | Hash trie nodes         | 0.00 B |     0 |
| Key-Value store | Path trie state lookups | 0.00 B |     0 |
| Key-Value store | Path trie account nodes | 0.00 B |     0 |
| Key-Value store | Path trie storage nodes | 0.00 B |     0 |
| Key-Value store | Trie preimages          | 0.00 B |     0 |
| Key-Value store | Account snapshot        | 0.00 B |     0 |
| Key-Value store | Storage snapshot        | 0.00 B |     0 |
| Key-Value store | Beacon sync headers     | 0.00 B |     0 |
| Key-Value store | Clique snapshots        | 0.00 B |     0 |
| Key-Value store | Singleton metadata      | 0.00 B |     0 |
[...]

PR #4789, opened 2025-12-24
Fix flake in `TestWaitForEvent/WaitForEvent_build_block_after_re-org`

yacovm:
## Why this should be merged

Fixes the flake by giving more time. The wait for event internally waits for 2 seconds and the deadline for the context is also 2 seconds which is borderline in times where the machine is overloaded and the context may expire too soon even though it should've succeeded.

## How this works

Increases the timeout.

## How this was tested

Commented out the other tests and ran the test 1000 times in parallel in a loop. Can be seen [here](https://github.com/yacovm/avalanchego/commit/ef3c797adace0a460e2feac185e50f6a3c41f7cf).


JonathanOppenheimer:
Closes https://github.com/ava-labs/avalanchego/issues/4757

JonathanOppenheimer (changes requested):
Can you change the test timeout in `coreth` as well?

PR #4790, opened 2025-12-25
feat(vm_reexecute): add pprof profiling support

Elvis339:
## Why this should be merged

Adds `--pprof-dir <path>` flag to C-chain re-execution benchmark for CPU and memory profiling.

Part of cross-stack profiling effort: https://github.com/ava-labs/firewood/issues/1582

pprof shows time spent in CGO calls as black boxes we see that `ffi.(*Revision).Get` takes X% of CPU but not what happens inside Firewood it's still valuable because:
- Shows Go-side hotspots (GC, allocations, serialization, AvalancheGo)
- Identifies which Go code paths trigger FFI calls
- Measures total time per FFI function

Combined with Rust-side profiling, we get full visibility across the boundary.

## How this works

When `--pprof-dir <path>` is passed:
1. Creates directory at `<path>` directory
2. Starts CPU profiling
3. Writes `cpu.profile`, `mem.profile` and `lock.profile` on exit

## How this was tested

1. `nix develop` (make sure you have AWS credentials set, c-chain-reexecute requires S3 access)
2. Edit line 97 of `vm_reexecution` and set output dir then run `./scripts/run_task.sh c-chain-reexecution-firewood-101-250k`
3. `go tool pprof /tmp/pprof/cpu.profile`
    - `top 20`
    - `list Revision.*Get`

NOTE: `--pprof-dir` is intentionally not exposed in `benchmark_cchain_range.sh` yet the profiling API may evolve as other profiling needs arise https://github.com/ava-labs/firewood/issues/1582. The flag is still available when running the test directly via go run.

```
ROUTINE ======================== github.com/ava-labs/firewood-go-ethhash/ffi.(*Revision).Get in /Users/elvis.sabanovic/go/pkg/mod/github.com/ava-labs/firewood-go-ethhash/ffi@v0.0.18/revision.go
      20ms      6.27s (flat, cum)  1.06% of Total
         .          .     61:func (r *Revision) Get(key []byte) ([]byte, error) {
         .          .     62:   if r.handle == nil {
         .          .     63:           return nil, ErrDroppedRevision
         .          .     64:   }
         .          .     65:
         .       10ms     66:   var pinner runtime.Pinner
      10ms       10ms     67:   defer pinner.Unpin()
         .          .     68:
      10ms      110ms     69:   return getValueFromValueResult(C.fwd_get_from_revision(
         .          .     70:           r.handle,
         .          .     71:           newBorrowedBytes(key, &pinner),
         .      6.14s     72:   ))
         .          .     73:}
         .          .     74:
         .          .     75:// Iter creates an iterator starting from the provided key on revision.
         .          .     76:// pass empty slice to start from beginning
         .          .     77:// It returns ErrDroppedRevision if Drop has already been called.
ROUTINE ======================== github.com/ava-labs/firewood-go-ethhash/ffi.(*Revision).Get.func1 in /Users/elvis.sabanovic/go/pkg/mod/github.com/ava-labs/firewood-go-ethhash/ffi@v0.0.18/revision.go
         0      6.14s (flat, cum)  1.03% of Total
         .          .     69:   return getValueFromValueResult(C.fwd_get_from_revision(
         .          .     70:           r.handle,
         .       30ms     71:           newBorrowedBytes(key, &pinner),
         .      6.11s     72:   ))
         .          .     73:}
         .          .     74:
         .          .     75:// Iter creates an iterator starting from the provided key on revision.
         .          .     76:// pass empty slice to start from beginning
         .          .     77:// It returns ErrDroppedRevision if Drop has already been called.
```

No

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -11,9 +11,9 @@ set -euo pipefail
 #   RUNNER_TYPE (optional): Runner type/label to include in benchmark naming.
 #   LABELS (optional): Comma-separated key=value pairs for metric labels.
 #   BENCHMARK_OUTPUT_FILE (optional): If set, benchmark output is also written to this file.
-#   METRICS_SERVER_ENABLED (optional, bool): If set, starts HTTP server exposing /metrics endpoint
+#   METRICS_SERVER_ENABLED (optional): If set, enables the metrics server.
 #   METRICS_SERVER_PORT (optional): If set, determines the port the metrics server will listen to.
-#   METRICS_COLLECTOR_ENABLED (optional, bool): If set, starts Prometheus agent to collect and forward metrics to remote instance
+#   METRICS_COLLECTOR_ENABLED (optional): If set, enables the metrics collector.
 
 : "${BLOCK_DIR:?BLOCK_DIR must be set}"
 : "${CURRENT_STATE_DIR:?CURRENT_STATE_DIR must be set}"


scripts/benchmark_cchain_range.sh
RodrigoVillar, line 1: I would revert these changes, given that #4761 will overwrite most of these.

tests/reexecute/c/vm_reexecute.go
```diff
@@ -135,6 +138,10 @@ func main() {
 	tc.SetDefaultContextParent(context.Background())
 	defer tc.RecoverAndExit()
 
+	if pprofEnabled {
+		defer setupPprof(tc)()
+	}
```
RodrigoVillar: This should go inside of `benchmarkReexecuteRange()`.

tests/reexecute/c/vm_reexecute.go
```diff
@@ -553,3 +561,37 @@ func parseCustomLabels(labelsStr string) (map[string]string, error) {
 	}
 	return labels, nil
 }
+
+func setupPprof(tc tests.TestContext) func() {
+	r := require.New(tc)
+	logger := tests.NewDefaultLogger("pprof")
```
RodrigoVillar: I think we can just use `tc.Log()` instead of creating a new logger here.

tests/reexecute/c/vm_reexecute.go
```diff
@@ -553,3 +561,37 @@ func parseCustomLabels(labelsStr string) (map[string]string, error) {
...
+
+	cwd, err := os.Getwd()
+	r.NoError(err, "failed to get current working directory")
+
+	profileDir := filepath.Join(cwd, "pprof")
+	r.NoError(os.MkdirAll(profileDir, perms.ReadWriteExecute), "failed to create profile directory")
+
+	cpuFile, err := os.Create(filepath.Join(profileDir, "cpu.prof"))
+	r.NoError(err, "failed to create CPU profile")
```
RodrigoVillar: I think it would be easier + cleaner if the reexecution test defined a `pprof` outfile file rather than defining a `pprofEnabled` boolean argument.

tests/reexecute/c/vm_reexecute.go
```diff
@@ -553,3 +561,37 @@ func parseCustomLabels(labelsStr string) (map[string]string, error) {
...
+		memFile, err := os.Create(filepath.Join(profileDir, "mem.prof"))
+		if err != nil {
+			logger.Error("failed to create memory profile", zap.Error(err))
+			return
+		}
+		defer memFile.Close()
+		if err := pprof.WriteHeapProfile(memFile); err != nil {
+			logger.Error("failed to write memory profile", zap.Error(err))
+			return
+		}
```
RodrigoVillar: q: will it be frequent that we write the CPU profile but not the memory profile (and vice versa)? Asking because similar to my comment about changing `pprofEnabled` from being a boolean to defining the filepath of the CPU profile, this should be defined by the client. However, if we always write the CPU and memory profile together, then `pprofEnabled` should be a directory argument rather than a boolean.

tests/reexecute/c/vm_reexecute.go
Elvis339: I noticed there's profiler in utils, so I refactored initial implementation.

https://github.com/ava-labs/avalanchego/pull/4790/commits/7f0d9331ce8af4e8cf85ad1e0baa5ccd330f1e9e

scripts/benchmark_cchain_range.sh
Elvis339: Changes in #4761 don't seem to conflict with changes here, even if they were to conflict I can resolve conflicts. Do you see any other reason for reverting beside git conflict?

tests/reexecute/c/vm_reexecute.go
Elvis339: Same comment: https://github.com/ava-labs/avalanchego/pull/4790#discussion_r2661824864

tests/reexecute/c/vm_reexecute.go
Elvis339: Thanks for the suggestion! I wouldn't say it's necessarily easier or cleaner than a boolean flag (similar line count, same behavior), but I do appreciate the semantic clarity the path serving as both the "enabled" signal and the configuration is more idiomatic for CLI tools.

Implemented in https://github.com/ava-labs/avalanchego/pull/4790/commits/1b5ea63e7f35b9c14bac123978070d64caf7bbee

tests/reexecute/c/vm_reexecute.go
Elvis339: Curious what use case you're thinking of when would you want CPU but not memory (or vice versa)?

There's no real cost to writing all three together, and imagine running a 45-minute block range (9M to 10M) only to realize you needed the memory profile but didn't capture it. Easier to just grab everything and analyze what you need.
Is there a scenario where separating them would be beneficial?

RodrigoVillar: > Curious what use case you're thinking of when would you want CPU but not memory (or vice versa)?

I don't have a use case, just wanted to make sure that enabling all profiles would be the expected behavior.

scripts/benchmark_cchain_range.sh
RodrigoVillar: PRs should be limited in scope to what they're intending to change. While I'm very much open to improving the reexecution documentation (e.g. #4803), I would prefer that such changes go into a separate PR.

Elvis339: Reverted. Happy to contribute to #4803 when you get to it a man page for the reexecution tooling would be a nice improvement.

Elvis339:
> Currently, the changes this PR introduces are only invokable via `go run`, which is not the recommended way to running the reexecution test (with preference going to `./scripts/benchmark_cchain_range.sh` or using `task`). Can this PR also include a way to invoke profiling both the preferred options above?
> 
> This may require coordination with #4761

See the note in the PR description the `--pprof-dir` is intentionally not exposed in the scripts yet.
This PR is a prerequisite for cross-stack profiling (a larger change that will consolidate how profiling is exposed across the tooling). Breaking it down for simpler, incremental review.
Currently I'm the only one experimenting with these benchmarks with debug flags and I'm compiling vm_reexecutable myself anyway so having it available via cmd arg is fine for now. Once the full profiling story lands, we can expose it properly in the scripts.

RodrigoVillar:
> > Currently, the changes this PR introduces are only invokable via `go run`, which is not the recommended way to running the reexecution test (with preference going to `./scripts/benchmark_cchain_range.sh` or using `task`). Can this PR also include a way to invoke profiling both the preferred options above?
> > This may require coordination with #4761
> 
> See the note in the PR description the `--pprof-dir` is intentionally not exposed in the scripts yet. This PR is a prerequisite for cross-stack profiling (a larger change that will consolidate how profiling is exposed across the tooling). Breaking it down for simpler, incremental review. Currently I'm the only one experimenting with these benchmarks with debug flags and I'm compiling vm_reexecutable myself anyway so having it available via cmd arg is fine for now. Once the full profiling story lands, we can expose it properly in the scripts.

Ah I now see #4791

Elvis339:
Shown useful for perf validation 

https://github.com/ava-labs/avalanchego/pull/4697

PR #4793, opened 2025-12-29
fix: Always register Firewood metrics when enabled

alarso16:
## Why this should be merged

Firewood metrics aren't particularly expensive, and there's a ticket to expose more expensive metrics within Firewood. It should always be registered.

## How this works

`ffi.StartMetrics` can only be called once per process, otherwirse it returns an error. This is a problem for tests, so this is moved to an init function, guaranteeing it will only happen once.

## How this was tested

Locally, to ensure metrics are still scrapable.
image

No.

### DIFF
--- a/graft/evm/firewood/metrics.go
+++ b/graft/evm/firewood/metrics.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package firewood


graft/evm/firewood/metrics.go
```diff
@@ -0,0 +1,15 @@
...
+package firewood
+
+import (
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
+	"github.com/ava-labs/libevm/log"
+)
+
+func init() {
+	if err := ffi.StartMetrics(); err != nil {
+		log.Crit("starting firewood metrics", "error", err)
```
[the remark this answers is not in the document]
JonathanOppenheimer: this makes sense

graft/coreth/plugin/evm/vm.go
```diff
@@ -513,12 +513,9 @@ func (vm *VM) initializeMetrics() error {
...
 
-	if vm.config.MetricsExpensiveEnabled && vm.config.StateScheme == customrawdb.FirewoodScheme {
-		if err := ffi.StartMetrics(); err != nil {
-			return fmt.Errorf("failed to start firewood metrics collection: %w", err)
-		}
-		if err := vm.ctx.Metrics.Register("firewood", ffi.Gatherer{}); err != nil {
-			return fmt.Errorf("failed to register firewood metrics: %w", err)
+	if vm.config.StateScheme == customrawdb.FirewoodScheme {
+		if err := vm.ctx.Metrics.Register(customrawdb.FirewoodScheme, ffi.Gatherer{}); err != nil {
+			return fmt.Errorf("registering firewood metrics: %w", err)
```
JonathanOppenheimer, line 518: shouldn't this also be failed to register?

graft/subnet-evm/plugin/evm/vm.go
```diff
@@ -584,12 +584,9 @@ func (vm *VM) initializeMetrics() error {
...
 
-	if vm.config.MetricsExpensiveEnabled && vm.config.StateScheme == customrawdb.FirewoodScheme {
-		if err := ffi.StartMetrics(); err != nil {
-			return fmt.Errorf("failed to start firewood metrics collection: %w", err)
-		}
-		if err := vm.ctx.Metrics.Register("firewood", ffi.Gatherer{}); err != nil {
-			return fmt.Errorf("failed to register firewood metrics: %w", err)
+	if vm.config.StateScheme == customrawdb.FirewoodScheme {
+		if err := vm.ctx.Metrics.Register(customrawdb.FirewoodScheme, ffi.Gatherer{}); err != nil {
+			return fmt.Errorf("registering firewood metrics: %w", err)
```
JonathanOppenheimer, line 589: same here - you clearly did this on purpose so it's intended?

ARR4N: It's an error so failure is implied. By the time it bubbles up to the top the message will be something like "failed to X: error when Y: unable to Z: permission denied" instead of "X: Y: Z: permission denied".

There's no established guidance on it anywhere, just something I recommend based on the rationale of maximising signal:noise.

JonathanOppenheimer: ah that makes sense - I will resolve all of these!

graft/evm/firewood/metrics.go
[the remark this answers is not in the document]
alarso16: It's a pet peeve of @ARR4N. There was a critical log, that's clearly a failure. So what's the point in saying it explicitly?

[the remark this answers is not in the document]
JonathanOppenheimer: Yup, he had commented as such. All good w/ me

maru-ava (approved):

PR #4794, opened 2025-12-30
Prep Firewood x/sync for EVM

alarso16:
## Why this should be merged

While writing the code for the EVM syncer, I noticed a few things that needed fixed.

## How this works

Abstracts some helpers and corrects some typos.

## How this was tested

Unit test

No

### DIFF
--- a/firewood/syncer/handler.go
+++ b/firewood/syncer/handler.go
@@ -12,11 +12,11 @@ import (
 // NewGetRangeProofHandler returns a handler that services GetRangeProof requests
 // using the provided Firewood database for p2p connections.
 func NewGetRangeProofHandler(db *ffi.Database) *sync.GetRangeProofHandler[*RangeProof, struct{}] {
-	return sync.NewGetRangeProofHandler(&database{db}, rangeProofMarshaler{})
+	return sync.NewGetRangeProofHandler(&database{db: db}, rangeProofMarshaler{})
 }
 
 // NewGetChangeProofHandler returns a handler that services GetChangeProof requests
 // using the provided Firewood database for p2p connections.
 func NewGetChangeProofHandler(db *ffi.Database) *sync.GetChangeProofHandler[*RangeProof, struct{}] {
-	return sync.NewGetChangeProofHandler(&database{db}, rangeProofMarshaler{}, changeProofMarshaler{})
+	return sync.NewGetChangeProofHandler(&database{db: db}, rangeProofMarshaler{}, changeProofMarshaler{})
 }
--- a/firewood/syncer/syncer.go
+++ b/firewood/syncer/syncer.go
@@ -49,7 +49,7 @@ func New(config Config, db *ffi.Database, targetRoot ids.ID, rangeProofClient *p
 		config.SimultaneousWorkLimit = defaultSimultaneousWorkLimit
 	}
 	return xsync.NewSyncer(
-		&database{db},
+		&database{db: db},
 		xsync.Config[*RangeProof, struct{}]{
 			RangeProofMarshaler:   rangeProofMarshaler{},
 			ChangeProofMarshaler:  changeProofMarshaler{},


firewood/syncer/proof.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package firewood
+package syncer
```
alarso16, line 4: package name was wrong (oops)

firewood/syncer/proof.go
```diff
@@ -13,17 +13,17 @@ import (
 )
 
 var (
-	_ sync.Marshaler[*RangeProof] = rangeProofMarshaler{}
-	_ sync.Marshaler[struct{}]    = changeProofMarshaler{}
+	_ sync.Marshaler[*RangeProof] = RangeProofMarshaler{}
```
alarso16: Needs exported to create the test handlers

firewood/syncer/syncer.go
```diff
@@ -21,24 +22,27 @@ import (
...
+	_ xsync.DB[*RangeProof, struct{}] = (*Database)(nil)
 
 	defaultSimultaneousWorkLimit = 8
 )
 
-// database wraps a Firewood FFI database to implement the xsync.DB interface.
-type database struct {
-	db *ffi.Database
+// Database wraps a Firewood FFI Database to implement the xsync.DB interface.
+type Database struct {
```
alarso16: Needs exported to create the test handlers

firewood/syncer/syncer.go
```diff
@@ -93,12 +104,18 @@ func (*database) VerifyRangeProof(_ context.Context, proof *RangeProof, start ma
...
 }
 
-func (db *database) CommitRangeProof(_ context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
+func (db *Database) CommitRangeProof(ctx context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
 	_, err := db.db.VerifyAndCommitRangeProof(proof.rp, start, end, ffi.Hash(proof.root), uint32(proof.maxLength))
 	if err != nil {
 		return maybe.Nothing[[]byte](), err
 	}
 
+	if db.rangeProofCallback != nil {
```
alarso16: Notice this is after the proof has been committed.

firewood/syncer/syncer_test.go
```diff
@@ -1,20 +1,23 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package firewood
+package syncer_test
```
alarso16: Had to create a test package to avoid import cycles, is there a better way to do this?

firewood/syncer/syncer.go
```diff
@@ -21,24 +22,27 @@ import (
...
+	db                 *ffi.Database
+	rangeProofCallback func(context.Context, *ffi.RangeProof) error
 }
 
 type Config struct {
 	SimultaneousWorkLimit int
 	Log                   logging.Logger
 	StateSyncNodes        []ids.NodeID
 	Registerer            prometheus.Registerer
+	RangeProofCallback    func(context.Context, *ffi.RangeProof) error
```
joshua-kim: What is this for? Are we using this in prod code/tests? The test added as part of this PR is testing that it's invoked, but it's not clear to me why it is important for us to make that assertion.

firewood/syncer/syncertest/network.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
+
+	"github.com/ava-labs/avalanchego/firewood/syncer"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/network/p2p/p2ptest"
+	"github.com/ava-labs/avalanchego/x/sync"
+)
+
+func NewTestRangeProofHandler(t *testing.T, db *ffi.Database) *p2p.Client {
```
joshua-kim: 1. Naming is misleading - this returns a client
2. I think the easiest way to get rid of the circular dependency problem would be to define `func NewGetRangeProofHandler(...)` in `syncer`, and expect tests to define the database for the handler themselves. If we do this we should be able to unstage the diff exporting the db + marshalers

firewood/syncer/syncer.go
alarso16: I need to do another operation external to the Firewood database during syncing. The range proof contains all the code hashes, so we use this to request the code separately

alarso16: This will be done in a separate PR - there's a lot of options how to do this anyway

firewood/syncer/handler.go
```diff
@@ -0,0 +1,18 @@
...
+package syncer
+
+import (
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
+
+	"github.com/ava-labs/avalanchego/x/sync"
+)
+
+func NewGetRangeProofHandler(db *ffi.Database) *sync.GetRangeProofHandler[*RangeProof, struct{}] {
+	return sync.NewGetRangeProofHandler(&database{db}, rangeProofMarshaler{})
```
JonathanOppenheimer: Is there a reason you didn't do NewDatabase(db), and instead directly wrapped?

firewood/syncer/handler.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+func NewGetChangeProofHandler(db *ffi.Database) *sync.GetChangeProofHandler[*RangeProof, struct{}] {
+	return sync.NewGetChangeProofHandler(&database{db}, rangeProofMarshaler{}, changeProofMarshaler{})
```
JonathanOppenheimer: same here

firewood/syncer/handler.go
JonathanOppenheimer: wait nvm

firewood/syncer/handler.go
JonathanOppenheimer: nvm

firewood/syncer/syncer_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package firewood
+package syncer
```
JonathanOppenheimer, line 4: do you want to put this in a test package?

firewood/syncer/handler.go
```diff
@@ -0,0 +1,18 @@
...
+// See the file LICENSE for licensing terms.
+
+package syncer
+
+import (
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
+
+	"github.com/ava-labs/avalanchego/x/sync"
+)
+
```
JonathanOppenheimer, line 11: do you want to add function doc comments to these two?

JonathanOppenheimer (approved):

firewood/syncer/handler.go
alarso16: The database doesn't need exposed. Any user only needs the handler

firewood/syncer/syncer_test.go
alarso16: I do not! My understanding is that putting something in a test package should be avoided when possible

firewood/syncer/syncer.go
```diff
@@ -49,7 +49,7 @@ func NewSyncer(config Config, db *ffi.Database, targetRoot ids.ID, rangeProofCli
 		config.SimultaneousWorkLimit = defaultSimultaneousWorkLimit
 	}
 	return xsync.NewSyncer(
-		&database{db: db},
+		&database{db},
```
joshua-kim: nit: keep keys for struct initialization - it makes it confusing especially as fields are added because order starts to matter

e.g

```
f := foo{1, 2}
```

might be different than
```
f := foo{1, 2}
```

if both fields were of the same type and were reordered

Issue #4776, opened 2025-12-19
Avoid using `runtime.caller` throughout the EVM codebase to get file paths


JonathanOppenheimer:
I'm fine with this cleanup, but please file an issue to avoid this practice entirely by embedding (as per the comment of the now deleted GetRepoRootPath).

_Originally posted by @maru-ava in https://github.com/ava-labs/avalanchego/pull/4742#discussion_r2635522411_

PR #4795, opened 2025-12-31
refactor: directly embed files in `graft`

JonathanOppenheimer:
## Why this should be merged

Closes https://github.com/ava-labs/avalanchego/issues/4776 -- this is a pretty nice clean up IMO. Credit to Maru for raising the issue. 

## How this works
One thing to note that is that embed declaration only work for files at the same level directory. See https://github.com/golang/go/issues/46056. So I created a single file to embed the shared genesis file and then imported it. 

## How this was tested
CI

No

### DIFF
--- a/graft/subnet-evm/tests/load/load_test.go
+++ b/graft/subnet-evm/tests/load/load_test.go
@@ -44,10 +44,17 @@ func TestE2E(t *testing.T) {
 var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
 	require := require.New(ginkgo.GinkgoT())
 
-	var env *e2e.TestEnvironment
+	var (
+		env      *e2e.TestEnvironment
+		repoRoot string
+	)
 
 	ginkgo.BeforeAll(func() {
 		tc := e2e.NewTestContext()
+
+		var err error
+		repoRoot, err = e2e.GetRepoRootPath("tests/load")
+		require.NoError(err)
 
 		nodes := utils.NewTmpnetNodes(nodeCount)
 		env = e2e.NewTestEnvironment(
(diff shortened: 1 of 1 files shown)


graft/subnet-evm/tests/load/load_test.go
```diff
@@ -88,7 +84,7 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
 
 		log.Info("Running load simulator...", "rpcEndpoints", commaSeparatedRPCEndpoints)
 		cmd := exec.Command("./scripts/run_simulator.sh")
-		cmd.Dir = filepath.Join(testDir, "../..")
+		cmd.Dir = "../.."
```
StephenButtolph: Why is this line changed?

JonathanOppenheimer: The `testDir` variable doesn't exist anymore

JonathanOppenheimer: Would you like me to add it back, so it is just used here? It's no longer necessary prior as the genesis is directly embedded and doesn't have to be referenced by path.

alarso16 (approved):

graft/subnet-evm/tests/load/load_test.go
maru-ava: What guarantees that this relative path always ends up as expected if it's not rooted at an absolute path?

JonathanOppenheimer: that is fair! I'll add it back.

graft/subnet-evm/tests/load/load_test.go
```diff
@@ -45,13 +44,17 @@ func TestE2E(t *testing.T) {
...
+		env      *e2e.TestEnvironment
+		repoRoot string
+	)
 
 	ginkgo.BeforeAll(func() {
 		tc := e2e.NewTestContext()
-		genesisPath := filepath.Join(testDir, "genesis.json")
+
+		var err error
+		repoRoot, err = e2e.GetRepoRootPath("tests/load")
```
maru-ava, line 56: I'm ok with continuing to use GetRepoRootPath for now, but ideally we remove the need for it going forward (i.e. by checking for the run_simulator.sh script via a relative path and if missing require an absolute path to be provided by a flag).

JonathanOppenheimer: I will do a follow up PR to accomplish this! and remove the function entirely so no one else can use it.

JonathanOppenheimer: Here is the PR https://github.com/ava-labs/avalanchego/pull/4813

Issue #4699, opened 2025-12-11
Refactor warp e2e test to avoid a single sequential function


JonathanOppenheimer:
[No action needed] It's outside the scope of this PR since the tests were already set up this way, but I think there is opportunity to structure these in a more standard way rather than a big sequential `testFunc`

Something like:
```
    testCombinations := []struct {
		name            string
		sendingSubnet   *Subnet
		receivingSubnet *Subnet
	}{
		{"SubnetA -> SubnetB", subnetA, subnetB},
		{"SubnetA -> SubnetA", subnetA, subnetA},
		{"SubnetA -> C-Chain", subnetA, cChainSubnetDetails},
		{"C-Chain -> SubnetA", cChainSubnetDetails, subnetA},
		{"C-Chain -> C-Chain", cChainSubnetDetails, cChainSubnetDetails},
	}
    var _ = ginkgo.Describe("[Warp]", func() {
	for _, combination := range testCombinations {
		combination := combination
		
		ginkgo.Describe(combination.name, ginkgo.Ordered, func() {
			var w *warpTest
			
			ginkgo.BeforeAll(func() {
				tc := e2e.NewTestContext()
				w = newWarpTest(tc.DefaultContext(), combination.sendingSubnet, combination.receivingSubnet)
			})

			ginkgo.It("should send warp message from sending subnet", func() {
				w.sendMessageFromSendingSubnet()
			})

			ginkgo.It("should aggregate signatures via API", func() {
				w.aggregateSignaturesViaAPI()
			})

			ginkgo.It("should deliver addressed call payload to receiving subnet", func() {
				w.deliverAddressedCallToReceivingSubnet()
			})

			ginkgo.It("should deliver block hash payload", func() {
				w.deliverBlockHashPayload()
			})

			ginkgo.It("should verify warp bindings", func() {
				w.warpBindingsTest()
			})

			ginkgo.It("should handle warp load testing", func() {
				w.warpLoad()
			})
		})
	}
})
```

The above is AI assisted and not thoroughly reviewed (read likely not best), but generally should provide better failure isolation and breaks each specific functionality tested into it's own `It` which is generally best practice.

I mostly add this comment to be on the lookout for in the future.

_Originally posted by @michaelkaplan13 in https://github.com/ava-labs/subnet-evm/pull/1907#discussion_r2612044181_

PR #4798, opened 2025-12-31
refactor: warp e2e tests no longer in single function

JonathanOppenheimer:
## Why this should be merged

This  refactors the warp e2e tests in both subnet-evm and coreth to use a more standard ginkgo test structure (close to the one suggested by @michaelkaplan13 ). This provides better failure isolation by breaking each test functionality into its own `It` block within ordered `Describe` blocks for each test combination.

Closes https://github.com/ava-labs/avalanchego/issues/4699

## How this works

Explicitly defined test combinations. 

## How this was tested
CI 

No

### DIFF
--- a/graft/coreth/tests/warp/warp_test.go
+++ b/graft/coreth/tests/warp/warp_test.go
@@ -145,15 +145,25 @@ var _ = ginkgo.Describe("[Warp]", func() {
 				w = newWarpTest(combination.sendingSubnet(), combination.receivingSubnet())
 			})
 
-			ginkgo.It("should send warp message from sending subnet", w.sendMessageFromSendingSubnet)
+			ginkgo.It("should send warp message from sending subnet", func() {
+				w.sendMessageFromSendingSubnet()
+			})
 
-			ginkgo.It("should aggregate signatures via API", w.aggregateSignaturesViaAPI)
+			ginkgo.It("should aggregate signatures via API", func() {
+				w.aggregateSignaturesViaAPI()
+			})
 
-			ginkgo.It("should deliver addressed call payload to receiving subnet", w.deliverAddressedCallToReceivingSubnet)
+			ginkgo.It("should deliver addressed call payload to receiving subnet", func() {
+				w.deliverAddressedCallToReceivingSubnet()
+			})
 
-			ginkgo.It("should deliver block hash payload", w.deliverBlockHashPayload)
+			ginkgo.It("should deliver block hash payload", func() {
+				w.deliverBlockHashPayload()
+			})
 
-			ginkgo.It("should handle warp load testing", w.warpLoad)
+			ginkgo.It("should handle warp load testing", func() {
+				w.warpLoad()
+			})
 		})
 	}
 })
--- a/graft/subnet-evm/tests/warp/warp_test.go
+++ b/graft/subnet-evm/tests/warp/warp_test.go
@@ -177,17 +177,29 @@ var _ = ginkgo.Describe("[Warp]", func() {
 				w = newWarpTest(combination.sendingSubnet(), combination.receivingSubnet())
 			})
 
-			ginkgo.It("should send warp message from sending subnet", w.sendMessageFromSendingSubnet)
+			ginkgo.It("should send warp message from sending subnet", func() {
+				w.sendMessageFromSendingSubnet()
+			})
 
-			ginkgo.It("should aggregate signatures via API", w.aggregateSignaturesViaAPI)
+			ginkgo.It("should aggregate signatures via API", func() {
+				w.aggregateSignaturesViaAPI()
+			})
 
-			ginkgo.It("should deliver addressed call payload to receiving subnet", w.deliverAddressedCallToReceivingSubnet)
+			ginkgo.It("should deliver addressed call payload to receiving subnet", func() {
+				w.deliverAddressedCallToReceivingSubnet()
+			})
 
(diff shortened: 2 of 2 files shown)


Elvis339:
No action required: the context created in `BeforeAll` is used for `newWarpTest` initialization, including `IssueTxsToActivateProposerVMFork` which waits for transactions. If `BeforeAll` has a shorter timeout than `It` blocks, this could *potentially* cause timeout issues. Maybe consider having `newWarpTest` create its own context like the other methods do. Not blocking if this hasn't caused issues in practice just my two cents.

JonathanOppenheimer:
> No action required: the context created in `BeforeAll` is used for `newWarpTest` initialization, including `IssueTxsToActivateProposerVMFork` which waits for transactions. If `BeforeAll` has a shorter timeout than `It` blocks, this could _potentially_ cause timeout issues. Maybe consider having `newWarpTest` create its own context like the other methods do. Not blocking if this hasn't caused issues in practice just my two cents.

Nah this is a great catch! I will make this change

graft/coreth/tests/warp/warp_test.go
```diff
@@ -121,32 +121,56 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		sendingSubnet   func() *Subnet
+		receivingSubnet func() *Subnet
+	}
 
-		ginkgo.GinkgoLogr.Info("Delivering block hash payload to receiving subnet")
-		w.deliverBlockHashPayload()
+	testCombinations := []testCombination{
+		// TODO: Uncomment these tests when we have a way to run them in CI, currently we should not depend on Subnet-EVM
+		// as Coreth and Subnet-EVM have different release cycles. The problem is that once we update AvalancheGo (protocol version),
+		// we need to update Subnet-EVM to the same protocol version. Until then all Subnet-EVM tests are broken, so it's blocking Coreth development.
```
maru-ava, line 133: Why isn't this test just part of subnet-evm? It might have made sense to keep it in coreth to ensure pre-merge validation but now that everything is in the same repo...

JonathanOppenheimer: It's in a follow up PR here: https://github.com/ava-labs/avalanchego/pull/4799

I wanted to keep this PR focused, on just in-line + clean up.

graft/coreth/tests/warp/warp_test.go
```diff
@@ -121,32 +121,56 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-		w.warpLoad()
+	for _, combination := range testCombinations {
+		ginkgo.Describe(combination.name, ginkgo.Ordered, func() {
+			var w *warpTest
+
+			ginkgo.BeforeAll(func() {
+				w = newWarpTest(combination.sendingSubnet(), combination.receivingSubnet())
+			})
+
+			ginkgo.It("should send warp message from sending subnet", func() {
```
maru-ava: Given that `It` is already going to emit informative logging (e.g. `It should send warp message from sending subnet`), these additional log statements appear entirely duplicative. Maybe prefer the following pattern?

```golang
ginkgo.It("should send warp message from sending subnet", w.aggregateSignaturesViaAPI)
```

JonathanOppenheimer: Oh I actually did not know that. I have changed these -- it's a lot simpler.

graft/subnet-evm/tests/warp/warp_test.go
```diff
@@ -154,33 +154,59 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-		w.warpLoad()
+	for _, combination := range testCombinations {
+		ginkgo.Describe(combination.name, ginkgo.Ordered, func() {
+			var w *warpTest
+
+			ginkgo.BeforeAll(func() {
+				w = newWarpTest(combination.sendingSubnet(), combination.receivingSubnet())
+			})
+
+			ginkgo.It("should send warp message from sending subnet", func() {
```
maru-ava: Same comment re: simplifying

JonathanOppenheimer: I didn't stage all the changes correctly, but it's there now!

graft/coreth/tests/warp/warp_test.go
```diff
@@ -145,15 +145,25 @@ var _ = ginkgo.Describe("[Warp]", func() {
 				w = newWarpTest(combination.sendingSubnet(), combination.receivingSubnet())
 			})
 
-			ginkgo.It("should send warp message from sending subnet", w.sendMessageFromSendingSubnet)
+			ginkgo.It("should send warp message from sending subnet", func() {
```
maru-ava, line 148: Was there a problem with using the test functions directly? I can only imaging it must have been the way ginkgo's test runner does discovery.

JonathanOppenheimer: Yeah without this it will still be `nil`, rather than how it should be after `BeforeAll` initializes it.

JonathanOppenheimer: Here is the failed CI run: https://github.com/ava-labs/avalanchego/actions/runs/20755108951/job/59594872320. We still remove the extra comment though so?

maru-ava: Still a win, yeah.

PR #4799, opened 2025-12-31
test: enable c-chain <-> subnet e2e warp tests

JonathanOppenheimer:
## Why this should be merged

Previously

> Uncomment these tests when we have a way to run them in CI, currently we should not depend on Subnet-EVM as Coreth and Subnet-EVM have different release cycles. The problem is that once we update AvalancheGo (protocol version), we need to update Subnet-EVM to the same protocol version. Until then all Subnet-EVM tests are broken, so it's blocking Coreth development. It's best to not run these tests until we have a way to run them in CI.

However, with the mono-repo, coreth and subnet-evm now have identical release cycles (in fact, they quite literally have to!). So this is no longer a concern. Additionally the warp test is run in CI, so that is also no longer a concern. 

This should not be merged until it's base PR (https://github.com/ava-labs/avalanchego/pull/4798) is merged. In addition https://github.com/ava-labs/avalanchego/pull/4795, should be merged so the genesis.json file does not need to be copied. 

This should be merged as it adds some important test coverage IMO. 

## How this works

Uncomments the tests, copy and pastes necessary code from subnet-evm. 

## How this was tested
CI

No

### DIFF
--- a/graft/coreth/Taskfile.yml
+++ b/graft/coreth/Taskfile.yml
@@ -119,6 +119,7 @@ tasks:
       E2E_TARGET: ./graft/coreth/tests/warp
     cmds:
       - cmd: ./scripts/run_task.sh build
+      - cmd: cd graft/subnet-evm && ./scripts/build.sh
       - cmd: bash -x ./scripts/tests.e2e.sh {{.CLI_ARGS}}
 
   test-e2e-warp-ci: # consolidated test-e2e-warp
@@ -129,4 +130,5 @@ tasks:
       E2E_SERIAL: 1
     cmds:
       - cmd: ./scripts/run_task.sh build-race
+      - cmd: cd graft/subnet-evm && ./scripts/build.sh
       - cmd: bash -x ./scripts/tests.e2e.sh {{.CLI_ARGS}}


graft/coreth/tests/utils/tmpnet.go
```diff
@@ -24,5 +29,42 @@ func NewTmpnetNetwork(owner string, nodes []*tmpnet.Node, flags tmpnet.FlagsMap)
 		Owner:        owner,
 		DefaultFlags: defaultFlags,
 		Nodes:        nodes,
+		Subnets:      subnets,
+	}
+}
+
+// Create the configuration that will enable creation and access to a
+// subnet created on a temporary network.
+func NewTmpnetSubnet(name string, genesisPath string, chainConfig map[string]any, nodes ...*tmpnet.Node) *tmpnet.Subnet {
```
JonathanOppenheimer: This function is copied directly from https://github.com/ava-labs/avalanchego/blob/f917492d193e1bed6c3f4b25d3cd40df169cbc6e/graft/subnet-evm/tests/utils/tmpnet.go#L48

graft/coreth/tests/warp/genesis.json
```diff
@@ -0,0 +1,45 @@
+{
```
JonathanOppenheimer, line 1: this is temporary, this PR needs to wait until a prior PR gets merged so I can change this.

JonathanOppenheimer: now we can embed!

graft/coreth/tests/utils/tmpnet.go
```diff
@@ -14,7 +18,17 @@ var DefaultChainConfig = map[string]any{
 	"local-txs-enabled": true,
 }
 
-func NewTmpnetNetwork(owner string, nodes []*tmpnet.Node, flags tmpnet.FlagsMap) *tmpnet.Network {
+func NewTmpnetNodes(count int) []*tmpnet.Node {
```
maru-ava: This would appear to duplicate [tmpnet.NewNodesOrPanic](https://github.com/ava-labs/avalanchego/blob/master/tests/fixture/tmpnet/node.go#L129)?

graft/coreth/tests/utils/tmpnet.go
```diff
@@ -14,7 +18,17 @@ var DefaultChainConfig = map[string]any{
...
+	nodes := make([]*tmpnet.Node, count)
+	for i := range nodes {
+		node := tmpnet.NewNode()
+		_ = node.EnsureKeys() // guaranteed to be nil for new node
```
maru-ava: What justifies this assertion?

graft/coreth/tests/utils/tmpnet.go
JonathanOppenheimer: I was matching subnet-evm -- this is a much better clean up though, to use the avalanchego function.

graft/coreth/tests/utils/tmpnet.go
JonathanOppenheimer: Now that the function is removed, the assertion is gone.

graft/coreth/tests/utils/tmpnet.go
```diff
@@ -24,5 +28,37 @@ func NewTmpnetNetwork(owner string, nodes []*tmpnet.Node, flags tmpnet.FlagsMap)
 		Owner:        owner,
 		DefaultFlags: defaultFlags,
 		Nodes:        nodes,
+		Subnets:      subnets,
+	}
+}
+
+// Create the configuration that will enable creation and access to a
+// subnet created on a temporary network.
+func NewTmpnetSubnet(name string, genesis []byte, chainConfig map[string]any, nodes ...*tmpnet.Node) *tmpnet.Subnet {
```
JonathanOppenheimer, line 37: This function is copied from subnet-evm to match.

graft/coreth/tests/warp/genesis.json
alarso16, line 1: Where did this come from?

JonathanOppenheimer: `graft/subnet-evm/tests/warp/genesis.json` 

We can eventually merge these two I hope in `graft/evm`.

graft/coreth/tests/warp/warp_test.go
```diff
@@ -43,10 +45,17 @@ import (
 	ginkgo "github.com/onsi/ginkgo/v2"
 )
 
+const (
+	subnetAName = "warp-subnet-a"
+)
```
powerslider:
```suggestion
const subnetAName = "warp-subnet-a"
```

graft/coreth/tests/utils/tmpnet.go
powerslider: I understand that this code is copied, but using a panic is definitely not ideal. Here is a better, cleaner and way more suitable test approach that involves returning an actual error with a meaningful error message that will be displayed when tests fail:

```
var ErrNoSubnetValidators = errors.New("a subnet must be validated by at least one node")

// Create the configuration that will enable creation and access to a
// subnet created on a temporary network.
func NewTmpnetSubnet(name string, genesis []byte, chainConfig map[string]any, nodes ...*tmpnet.Node) (*tmpnet.Subnet, error) {
	if len(nodes) == 0 {
		return nil, ErrNoSubnetValidators
	}

	validatorIDs := make([]ids.NodeID, len(nodes))
	for i, node := range nodes {
		validatorIDs[i] = node.NodeID
	}

	chainConfigBytes, err := json.Marshal(chainConfig)
	if err != nil {
		return nil, fmt.Errorf("marshal chain config: %w", err)
	}

	return &tmpnet.Subnet{
		Name: name,
		Chains: []*tmpnet.Chain{
			{
				VMID:         constants.EVMID,
				Genesis:      genesis,
				Config:       string(chainConfigBytes),
				PreFundedKey: tmpnet.HardhatKey,
			},
		},
		ValidatorIDs: validatorIDs,
	}, nil
}
```

graft/coreth/tests/warp/warp_test.go
```diff
@@ -85,6 +93,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 			"coreth-warp-e2e",
 			nodes,
 			tmpnet.FlagsMap{},
+			utils.NewTmpnetSubnet(subnetAName, genesis, utils.DefaultChainConfig, nodes...),
 		),
 	)
```
powerslider, line 96: Then I would just assert the error here via the ginkgo runner:
```
	require := require.New(ginkgo.GinkgoT())
	subnetA, err := utils.NewTmpnetSubnet(subnetAName, genesis, utils.DefaultChainConfig, nodes...)
	require.NoError(err)

	env := e2e.NewTestEnvironment(
		tc,
		flagVars,
		utils.NewTmpnetNetwork(
			"coreth-warp-e2e",
			nodes,
			tmpnet.FlagsMap{},
			subnetA,
		),
	)
```

powerslider: and of course make sure to handle the error with `require.NoError` everywhere where `utils.NewTmpnetSubnet` is called. These fixes apply to both `coreth` and `subnet-evm` so make sure to equalize them.

graft/coreth/tests/utils/tmpnet.go
maru-ava: This is test code, a panic really doesn't matter. Failing the test with a panic or an error in an e2e suite is effectively the same thing.

Edit: and fwiw test assertion failures _are_ panics. They just have built-in recovery.

graft/coreth/tests/warp/warp_test.go
alarso16: I'm surprised this didn't fail lint...

graft/coreth/tests/utils/tmpnet.go
JonathanOppenheimer: Yeah this does seem slightly overkill given it is test code, but I don't mind adding this.

graft/coreth/tests/warp/warp_test.go
JonathanOppenheimer: Applied

graft/coreth/tests/utils/tmpnet.go
JonathanOppenheimer: Maru if you strongly disagree, I will revert the last commit.

maru-ava: It's evm code, so pick your poison, but I think consistency with existing avalanchego patterns would be preferable.

My rational for not returning an error for these types of setup functions is so that they can be used inline in constructors without having to bother with error checking. That's why  `tmpnet.NewNodesOrPanic` panics instead of returning an error. It was a deliberate choice taking into account that the intended context of use is compatible with panics. 

In general, I don't think it always makes sense to apply the same rational for code for testing as we'd normally apply to non-test code. That doesn't mean being lazy or careless, just recognizing that the contexts are different and that can enable some nice quality-of-life choices in test code.

JonathanOppenheimer: I will revert the PR then -- it does add a lot of extra lines of code, that the panic avoids - it's easier to read without.

JonathanOppenheimer:
I reverted the last two commits of Tsvetan after Maru's comment.

graft/coreth/tests/utils/tmpnet.go
```diff
@@ -24,5 +28,37 @@ func NewTmpnetNetwork(owner string, nodes []*tmpnet.Node, flags tmpnet.FlagsMap)
...
+	}
+
+	return &tmpnet.Subnet{
+		Name: name,
+		Chains: []*tmpnet.Chain{
+			{
+				VMID:         constants.EVMID,
+				Genesis:      genesis,
+				Config:       string(chainConfigBytes),
+				PreFundedKey: tmpnet.HardhatKey,
```
ceyonur, line 59: I wonder why this is named HardhatKey (pretty much irrelevant for this PR)

graft/coreth/tests/utils/tmpnet.go
```diff
@@ -24,5 +28,37 @@ func NewTmpnetNetwork(owner string, nodes []*tmpnet.Node, flags tmpnet.FlagsMap)
...
+	chainConfigBytes, err := json.Marshal(chainConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	return &tmpnet.Subnet{
+		Name: name,
+		Chains: []*tmpnet.Chain{
+			{
+				VMID:         constants.EVMID,
```
ceyonur: should this be SubnetEVMID?

graft/coreth/tests/utils/tmpnet.go
maru-ava: The original definition for tmpnet was sourced from the testing that used hardhat. Feel free to change the name, or even better generate the genesis and leverage tmpnet's ability to prefund keys dynamicaly.

JonathanOppenheimer: I'm not sure if I quite know what you mean by 

> generate the genesis and leverage tmpnet's ability to prefund keys dynamicaly

Is 1d84d0bf37fa54572de8cf3c3157665cd1f82d6d what you had in mind?

maru-ava: I'm afraid it's going to take more than that. As per the failing e2e jobs, sourcing prefunded keys this way depends on generating the genesis at runtime rather than loading it from disk. I'm guessing the current tmpnet genesis generator would need to be updated to enable the customization required by this testing.

JonathanOppenheimer: In that case, I think those changes would be out of scope for this PR. I will create an issue to address this item of work, and address it seperately.

JonathanOppenheimer: https://github.com/ava-labs/avalanchego/issues/4845

JonathanOppenheimer: I am going to resolve this conversation, please re-open if you think differently.

graft/coreth/tests/utils/tmpnet.go
ceyonur: I don't think this is resolved? @JonathanOppenheimer

JonathanOppenheimer: Sorry this was lost somehow - it's fixed now!

PR #4801, opened 2026-01-02
chore: update license header to 2026

JonathanOppenheimer:
## Why this should be merged

Happy New Year! 

## How this works

4 files were manually updated -- 

- `header.yml`
- `header_upstream.yml`
- `LICENSE`
- `CONTRIBUTING.md`

I then used a bulk replace to replace  "2019-2025" with "2019-2026". This PR is best reviewed in your IDE with a find and replace tool. 

## How this was tested
CI

No

### DIFF
--- a/header_upstream.yml
+++ b/header_upstream.yml
@@ -1,5 +1,5 @@
 header: |
-  // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+  // Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
   // See the file LICENSE for licensing terms.
   //
   // This file is a derived work, based on the go-ethereum library whose original


header.yml
```diff
@@ -1,4 +1,3 @@
 header: |
-  // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+  // Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
   // See the file LICENSE for licensing terms.
```
JonathanOppenheimer, line 3: This file had an extra (2) trailing new lines. I deleted the extra and that is why the difference on this PR is not even.

alarso16: I don't think your comment is quite correct. It had two trailing spaces after the newline, not an extra newline. However, I think you are correct in that these two spaces are not needed - I checked locally that this ensures that the license header is followed by an empty newline

alarso16 (approved):

PR #1315, opened 2023-04-10
Add available memory health check

StephenButtolph:
## Why this should be merged

Closes #1314 

Nodes OOMing can cause extended downtime due to re-generating any non-committed state. This will more aggressively warn a node operator that the node is running out of memory.

## How this works

Adds configs:

1. This adds memory tracking into the process memory tracker.
2. Registers a health check into the node to report unhealthy depending on the tracked memory profile.
3. Adds configs to parameterize the node behavior.
  - `--system-tracker-memory-required-available` (defaults to 500 MiB)
  - `--system-tracker-memory-warning-threshold-available` (defaults to 1 GiB)

## How this was tested

- Locally (on Ubuntu 22.04) I intentionally introduced a memory leak and monitored the behavior of the node.
- Locally (on MacOS)
- Locally (on Windows) Windows handles swap a bit differently than mac / ubuntu... So I had some issues figuring out what the expected values were... but it seemed to be reasonable

PR #4802, opened 2026-01-03
feat: add memory usage health check

### LINKED ISSUE #1314: Add memory usage health check
If the available memory remaining on a host goes below a specified amount, the node should become unhealthy and display warnings. If the node is critically low on memory it may make sense to actually shut down the node in an attempt to gracefully terminate.

This would be similar to how we handle low disk space.

JonathanOppenheimer:
## Why this should be merged

This closes a long standing issue https://github.com/ava-labs/avalanchego/issues/1314. For context, @alarso16 recently made a PR to convert the disk health check to percentage based: https://github.com/ava-labs/avalanchego/pull/4770, which reminded me of this old issue. 

His PR provided an excellent template to implement the memory percentage check -- not much creativity went into this PR -- most of the relevant code blocks are copy and pasted, with configs changed. I considered trying to generalize the code to make it less repetitive, but didn't think it was worth the effort. If other's disagree, I can take a pass at it. 

## How this works

This PR adds the memory usage health check that monitors system memory availability and takes action when memory becomes critically low, following the _exact_ same pattern as the existing disk space health check. 

- Shutdown threshold (default: 3%): When available memory drops below this percentage, the node logs a fatal error and gracefully shuts down 
- Warning threshold (default: 10%): When available memory drops below this percentage (but above the shutdown threshold), the node reports as unhealthy but continues operating
 - The health status is queryable via the `/health` API endpoint, which returns the current memory statistics

Two new flags control the thresholds:
 - `--system-tracker-memory-required-available-percentage` (default: 3) - Minimum available memory percentage before shutdown
 - `--system-tracker-memory-warning-available-percentage` (default: 10) - Warning threshold percentage

 Both thresholds must be between 0-50% (again, same as the disk) 

## How this was tested
CI, and a new test, `TestGetMemoryConfig`

Yes


config/config.go
```diff
@@ -55,7 +55,8 @@ const (
 	chainUpgradeFileName = "upgrade"
 	subnetConfigFileExt  = ".json"
 
-	maxDiskSpaceThreshold = 50
+	maxDiskSpaceThreshold   = 50
+	maxMemorySpaceThreshold = 50
```
JonathanOppenheimer, line 59: We can change this if we'd like. I just copied all the values for memory, but I understand memory is not disk!

config/flags.go
```diff
@@ -361,6 +361,8 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.Uint64(SystemTrackerRequiredAvailableDiskSpacePercentageKey, 3, "Minimum percentage (between 0 and 50) of available disk space, under which the node will shutdown.")
 	fs.Uint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey, 0, fmt.Sprintf("DEPRECATED: Warning threshold for the number of available bytes on disk, under which the node will be considered unhealthy.  Must be >= [%s]", SystemTrackerRequiredAvailableDiskSpaceKey))
 	fs.Uint64(SystemTrackerWarningAvailableDiskSpacePercentageKey, 10, fmt.Sprintf("Warning threshold for the percentage (between 0 and 50) of available disk space, under which the node will be considered unhealthy. Must be >= [%s]", SystemTrackerRequiredAvailableDiskSpacePercentageKey))
+	fs.Uint64(SystemTrackerRequiredAvailableMemoryPercentageKey, 3, "Minimum percentage (between 0 and 50) of available system memory, under which the node will shutdown.")
+	fs.Uint64(SystemTrackerWarningAvailableMemoryPercentageKey, 10, fmt.Sprintf("Warning threshold for the percentage (between 0 and 50) of available system memory, under which the node will be considered unhealthy. Must be >= [%s]", SystemTrackerRequiredAvailableMemoryPercentageKey))
```
JonathanOppenheimer: Same here

utils/resource/usage.go
```diff
@@ -205,6 +242,11 @@ func (m *manager) update(diskPath string, frequency, cpuHalflife, diskHalflife t
 			m.availableDiskPercent = availablePercentage
 		}
 
+		if getMemErr == nil {
+			m.availableMemoryBytes = virtualMem.Available
+			m.availableMemoryPercent = uint64(100 - virtualMem.UsedPercent)
```
DracoLi: I’m not sure we should use `UsedPercent` here. `UsedPercent` represents memory that is actively in use and may exclude buffers/cache, while `Available` may include reclaimable memory (cache, buffers) that the kernel can free if needed.
I think for an accurate `availableMemoryPercent`, we need to calculate percentage using Available / Total.

config/flags.go
```diff
@@ -361,6 +361,8 @@ func addNodeFlags(fs *pflag.FlagSet) {
 	fs.Uint64(SystemTrackerRequiredAvailableDiskSpacePercentageKey, 3, "Minimum percentage (between 0 and 50) of available disk space, under which the node will shutdown.")
 	fs.Uint64(SystemTrackerWarningThresholdAvailableDiskSpaceKey, 0, fmt.Sprintf("DEPRECATED: Warning threshold for the number of available bytes on disk, under which the node will be considered unhealthy.  Must be >= [%s]", SystemTrackerRequiredAvailableDiskSpaceKey))
 	fs.Uint64(SystemTrackerWarningAvailableDiskSpacePercentageKey, 10, fmt.Sprintf("Warning threshold for the percentage (between 0 and 50) of available disk space, under which the node will be considered unhealthy. Must be >= [%s]", SystemTrackerRequiredAvailableDiskSpacePercentageKey))
+	fs.Uint64(SystemTrackerRequiredAvailableMemoryPercentageKey, 3, "Minimum percentage (between 0 and 50) of available system memory, under which the node will shutdown.")
```
DracoLi: I'm not sure if we should use the same disk space defaults for memory. Memory might spike ~3% quickly so if that happens we might not be able to gracefully shutdown before it gets to 100%. Maybe a 5% default would be better here?

JonathanOppenheimer: Yeah, as I noted above, I'm not tied to these percentages, as I had just copied them from disk. I will change to 5%!

utils/resource/usage.go
JonathanOppenheimer: I am fairly certainly the two methods are equivalent from taking a look at https://github.com/shirou/gopsutil/blob/master/mem/mem_linux.go .

The calculation is 

```
ret.UsedPercent = float64(ret.Used) / float64(ret.Total) * 100.0
```

which is the same as what we do here. `WillBeDeletedOptOutMemAvailableCalc` isn't enabled and will be removed in March 2026 (to your concern of buffers and cache). 

I can however still make this change as it aligns better with the disk calculation.

DracoLi: 👍 thx for looking into this. In this case I think either a comment here or updating it to Used / Total directly would all be great solutions.

DracoLi (approved):

utils/resource/usage.go
JonathanOppenheimer: Yup I updated it to `Used  / Total` directly as you suggested!

snow/networking/tracker/resource_tracker.go
```diff
@@ -331,6 +367,14 @@ func newCPUTrackerMetrics(reg prometheus.Registerer) (*trackerMetrics, error) {
 			Name: "disk_available_percentage",
 			Help: "Percentage of database volume available",
 		}),
+		memoryAvailable: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "memory_available_space",
```
maru-ava: Maybe just `memory_available`?

snow/networking/tracker/resource_tracker.go
```diff
@@ -331,6 +367,14 @@ func newCPUTrackerMetrics(reg prometheus.Registerer) (*trackerMetrics, error) {
...
+			Help: "Available memory remaining (bytes) on the system",
+		}),
+		memoryPercentageAvailable: prometheus.NewGauge(prometheus.GaugeOpts{
```
maru-ava, line 374: (No action required) Why `memoryPercentableAvailable` / `memory_available_percentage` instead of the same for both? Not requiring a change because at least it's consistent with the disk version.

maru-ava (commented):
Is there prior art suggesting this is a good idea? I usually think of OOM conditions as something that something at the OS level would be handling, even more so if running in e.g. kubernetes.

JonathanOppenheimer:
> Is there prior art suggesting this is a good idea? I usually think of OOM conditions as something that something at the OS level would be handling, even more so if running in e.g. kubernetes.

I am not familiar with Kubernetes (at all), but it doesn't seem to me to be a bad idea that nodes should be at least aware of available memory, so we can react, and gracefully shutdown (it seems like it'd be the same rationale for disk space, but maybe this is managed differently in Kubernetes and I am not aware). It's not like we're managing memory (reduce our usage) or anything here, it's all passive. Perhaps @StephenButtolph has additional thoughts? Am I correct to read your comment as thinking this is unnecessary?

snow/networking/tracker/resource_tracker.go
JonathanOppenheimer: Yeah I just matched it with the disk. I don't think it's unreasonable to change both, which I will do. Seems odd they both don't match.

maru-ava:
> Am I correct to read your comment as thinking this is unnecessary?

Yeah, I'm not clear on the value of this.

JonathanOppenheimer:
> > Am I correct to read your comment as thinking this is unnecessary?
> 
> Yeah, I'm not clear on the value of this.

In a world where system memory gets extremely / dangerously low on a node to the point that a node can no longer operate, what would currently happen? I assumed it would just crash as we have no way to monitor memory, but I'm inferring I'm wrong?

maru-ava:
> > > Am I correct to read your comment as thinking this is unnecessary?
> > 
> > 
> > Yeah, I'm not clear on the value of this.
> 
> In a world where system memory gets extremely / dangerously low on a node to the point that a node can no longer operate, what would currently happen? I assumed it would just crash as we have no way to monitor memory, but I'm inferring I'm wrong?

I think it really depends on the runtime environment of the node. Unlike disk, memory utilization can change quite dynamically. 

On a bare metal node, a momentary condition might cause the proposed health check to shutdown the node before the kernel VMM (virtual memory manager) has a chance to OOM kill a misbehaving process and free up memory. I'm not sure why we would want to do that. 

For kube, I don't think we'd want to this at all. Kube already has ways of declaring memory requests/limits and orchestrating scheduled workloads accordingly. Adding another mechanism just adds potential for conflict. That suggests to me that if we do want to add this feature that it should be disabled by default since it should really only be enabled for bare metal nodes.

I think there might be more value in warning about a low memory condition rather than responding to it with node shutdown.

What am I missing?

JonathanOppenheimer:
> > > > Am I correct to read your comment as thinking this is unnecessary?
> > > 
> > > 
> > > Yeah, I'm not clear on the value of this.
> > 
> > 
> > In a world where system memory gets extremely / dangerously low on a node to the point that a node can no longer operate, what would currently happen? I assumed it would just crash as we have no way to monitor memory, but I'm inferring I'm wrong?
> 
> I think it really depends on the runtime environment of the node. Unlike disk, memory utilization can change quite dynamically.
> 
> On a bare metal node, a momentary condition might cause the proposed health check to shutdown the node before the kernel VMM (virtual memory manager) has a chance to OOM kill a misbehaving process and free up memory. I'm not sure why we would want to do that.
> 
> For kube, I don't think we'd want to this at all. Kube already has ways of declaring memory requests/limits and orchestrating scheduled workloads accordingly. Adding another mechanism just adds potential for conflict. That suggests to me that if we do want to add this feature that it should be disabled by default since it should really only be enabled for bare metal nodes.
> 
> I think there might be more value in warning about a low memory condition rather than responding to it with node shutdown.
> 
> What am I missing?

So I am incorrect! That makes a lot of sense. Prematurely killing a node would certainly be a bad idea. You're not missing anything, I was missing how systems properly deal with low memory conditions. Rather than close than PR right now, @StephenButtolph as the one who created the issue do you disagree with any of the above? If not, I will close both this PR, and the original issue as not planned.

StephenButtolph:
The original PR to introduce this logic was here: #1315. It was closed because of concern around K8s support.

I agree that we probably never want to FATAL the node due to these memory readings... I do think it could be possible for us to someday report unhealthy...

Regardless, I don't think this PR will be able to be merged as is. Whether we should close the issue as a whole or not... 🤷

JonathanOppenheimer:
> The original PR to introduce this logic was here: #1315. It was closed because of concern around K8s support.
> 
> I agree that we probably never want to FATAL the node due to these memory readings... I do think it could be possible for us to someday report unhealthy...
> 
> Regardless, I don't think this PR will be able to be merged as is. Whether we should close the issue as a whole or not... 🤷

Thanks for the context in the additional issue! I have modified this PR to

- Only report unhealthy, never shutdown
   - Single warning threshold 
   - Disabled by default (default is 0, user would have to manually configure a warning threshold, rather than defaulting to say 3%) 
- Correctly get memory percentage on v1 (`core/v1`) and v2 (`core/v2`) k8s. There's actually a good amount of discussion on the internet about this issue, including an issue on `gopsutil` repo in which a guy shared how to correctly get these figures: https://github.com/shirou/gopsutil/issues/1758.

I'm laughing at the "someday." If this PR is still miles away from being mergeable, I would suggest closing this PR and the issue, or at least edit the issue to be far more accurate (as it is not currently). 

cc @maru-ava

joshua-kim:
Closing PR due to staleness

PR #4804, opened 2026-01-05
feat(reexecution): collect leveldb and meterdb metrics

DracoLi:
## Why this should be merged

Enables reexecution benchmarks to collect leveldb and meterdb metrics. This allows existing leveldb or meterdb related Grafana dashboards to work with metrics from benchmark runs.

With this change, metrics like `avalanche_db_non_level_0_comps` and `avalanche_meterdb_calls{method="get"}` are available during reexecution tests as well, allowing us to monitor db performance.

## How this works

- Registers leveldb metrics under `avalanche_db` namespace
- Wraps the database with meterdb and registers metrics under `avalanche_meterdb` namespace

## How this was tested

I ran a reexecution benchmark and confirmed that the leveldb and meterdb metrics are visible.

No

### DIFF
--- a/tests/reexecute/c/vm_reexecute.go
+++ b/tests/reexecute/c/vm_reexecute.go
@@ -227,19 +227,19 @@ func benchmarkReexecuteRange(
 	r.NoError(err)
 
 	dbLogger := tests.NewDefaultLogger("db")
-	dbReg, err := metrics.MakeAndRegister(
+	dbRegistry, err := metrics.MakeAndRegister(
 		prefixGatherer,
 		metric.AppendNamespace(constants.PlatformName, "db"),
 	)
 	r.NoError(err)
-	db, err := leveldb.New(vmDBDir, nil, dbLogger, dbReg)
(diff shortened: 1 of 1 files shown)


tests/reexecute/c/vm_reexecute.go
```diff
@@ -224,8 +227,19 @@ func benchmarkReexecuteRange(
 	r.NoError(err)
 
 	dbLogger := tests.NewDefaultLogger("db")
-
-	db, err := leveldb.New(vmDBDir, nil, dbLogger, prometheus.NewRegistry())
+	dbReg, err := metrics.MakeAndRegister(
```
RodrigoVillar:
```suggestion
	dbRegistry, err := metrics.MakeAndRegister(
```

(nit/feel-free-to-ignore): Could we follow the existing naming convention for registries in this function (e.g. `meterVMRegistry`, `consensusRegistry`)?

RodrigoVillar: Ah I didn't know `metrics.MakeAndRegister()` existed - awesome!

PR #4809, opened 2026-01-05
test: query historical C-Chain state with Firewood

RodrigoVillar:
## Why this should be merged

As part of testing Firewood's archival support, we need to test that C-Chain queries against historical states are supported.

Closes https://github.com/ava-labs/firewood/issues/1499.

## How this works

Introduces a state access test that, given a start ($X$) and end block ($Y$), queries an instance of Coreth by asking for the nonce of the zero address for all blocks $\alpha \in [X, Y]$. This is sufficient for testing historical queries as this succeeds only if Coreth has the matching trie for each $\alpha$.

## How this was tested

CI

No

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -241,7 +241,7 @@ tasks:
 
   test-firewood-state-access:
     desc: Runs Firewood state access test
-    cmd: ./scripts/tests.firewood_state_access.sh {{.CLI_ARGS}}    
+    cmd: ./scripts/tests.firewood_state_access.sh {{.CLI_ARGS}}
 
   # To use a different fuzz time, run `task test-fuzz FUZZTIME=[value in seconds]`.
   # A value of `-1` will run until it encounters a failing output.
--- a/tests/reexecute/stateaccess/main.go
+++ b/tests/reexecute/stateaccess/main.go
@@ -84,7 +84,7 @@ func main() {
 	client, err := ethclient.Dial(server.URL)
 	r.NoError(err)
 
-	for i := startBlockArg; i <= endBlockArg; i += 1 {
+	for i := startBlockArg; i <= endBlockArg; i++ {
 		nonce, err := client.NonceAt(ctx, common.Address{}, big.NewInt(int64(i)))
 		r.NoErrorf(err, "failed to get nonce at block %d", i)
 		r.Zero(nonce)


tests/reexecute/stateaccess/main.go
```diff
@@ -0,0 +1,92 @@
...
+	flag.Parse()
+}
+
+// main verifies that historical state is accessible for archival node by
+// iterating over a range of blocks and querying the nonce of the zero address
+// at each block height. This confirms the node can serve state queries for
+// arbitrary historical blocks.
+func main() {
+	tc := tests.NewTestContext(tests.NewDefaultLogger(""))
+	tc.RecoverAndExit()
```
RodrigoVillar: Need to defer this.

.github/workflows/firewood-state-access-test.yml
```diff
@@ -0,0 +1,88 @@
...
+      runner:
+        description: 'Runner to execute the state access test. Input to the runs-on field of the job.'
+        required: true
+      timeout-minutes:
+        description: 'Timeout in minutes for the job.'
+        default: '60'
+  schedule:
+    - cron: '0 9 * * *' # Runs every day at 09:00 UTC (04:00 EST)
+  # XXX: REMOVE BEFORE MERGING
+  pull_request:
```
RodrigoVillar, line 28: Will remove before merging.

scripts/tests.firewood_state_access.sh
RodrigoVillar, line 1: I've went with creating a new script rather than extending `./scripts/benchmark_cchain_range.sh` since the state access tests simply reads over the current state, while both the reexecution and chaos tests start with the current state and write to it. This causes common terms such as `START_BLOCK` and `END_BLOCK` to take on different meanings depending on the context they're used in (state access test vs reexecution/chaos tests) and so I went with creating a new script to make things clearer.

maru-ava:
This seems awfully heavyweight to validate that historical state will be accessible via firewood. Are there faster/lighter tests already that validate accessibility in a way that doesn't require access to s3 and processing 250k+ blocks?

Taskfile.yml
```diff
@@ -239,6 +239,10 @@ tasks:
     cmds:
       - cmd: bash -x ./scripts/tests.e2e.kube.sh --ginkgo.focus-file=xsvm.go {{.CLI_ARGS}}
 
+  test-firewood-state-access:
+    desc: Runs Firewood state access test
+    cmd: ./scripts/tests.firewood_state_access.sh {{.CLI_ARGS}}
```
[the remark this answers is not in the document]
RodrigoVillar: Done: 3e8dfc5

tests/reexecute/stateaccess/main.go
```diff
@@ -0,0 +1,92 @@
...
+	r.NoError(err)
+
+	ethRPCEndpoint := "/rpc"
+	server := httptest.NewServer(handlers[ethRPCEndpoint])
+	tc.DeferCleanup(server.Close)
+
+	client, err := ethclient.Dial(server.URL)
+	r.NoError(err)
+
+	for i := startBlockArg; i <= endBlockArg; i += 1 {
```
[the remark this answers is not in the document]
RodrigoVillar: Done: 3e8dfc5

RodrigoVillar:
> This seems awfully heavyweight to validate that historical state will be accessible via firewood. Are there faster/lighter tests already that validate accessibility in a way that doesn't require access to s3 and processing 250k+ blocks?

I could reduce the block range (e.g. only query blocks 100k-250k) but to the point of this being heavyweight, could you elaborate more?

alarso16:
> > This seems awfully heavyweight to validate that historical state will be accessible via firewood. Are there faster/lighter tests already that validate accessibility in a way that doesn't require access to s3 and processing 250k+ blocks?
> 
> I could reduce the block range (e.g. only query blocks 100k-250k) but to the point of this being heavyweight, could you elaborate more?

You can explicitly set `state-history` in the config, which corresponds to `RevisionsInMemory`. With this, you could set it to like 5, and then just reading 10 blocks is sufficient to test that historical state is available

maru-ava:
> I could reduce the block range (e.g. only query blocks 100k-250k) but to the point of this being heavyweight, could you elaborate more?

Why do we even need historical state for this? Unlike reexecution, where block processing involves a lot more than just firewood, this seems like something that could be trivially verified in a unit test.

alarso16:
> Why do we even need historical state for this? Unlike reexecution, where block processing involves a lot more than just firewood, this seems like something that could be trivially verified in a unit test.

It does seem unnecessary to be a re-execution test, you shouldn't need real state to verify this. You could just generate a few random blocks and then query to ensure that the state didn't disappear

RodrigoVillar:
Closing this PR in favor of @alarso16's suggestions.

PR #4810, opened 2026-01-06
chore: remove end year from license headers

JonathanOppenheimer:
## Why this should be merged
Updating the year every year is slightly annoying and unnecessary. Our legal department has confirmed that updating the end year doesn't actually do anything to update our copyright, and is thus unnecessary. 

## How this was tested
CI

No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,7 @@ Mocks are auto-generated using [mockgen](https://pkg.go.dev/go.uber.org/mock/moc
   - if the file `mocks_generate_test.go` does not exist in the package where the interface is located, create it with content (adapt as needed):
 
     ```go
-    // Copyright (C) 2019-2026, Ava Labs, Inc. All rights reserved.
+    // Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
     // See the file LICENSE for licensing terms.
 
     package mypackage
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (C) 2019-2026, Ava Labs, Inc.
+Copyright (C) 2019, Ava Labs, Inc.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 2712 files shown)


joshua-kim (changes requested):
My understanding is that the updating the end year doesn't actually do anything to update our copyright. We should probably be just specifying the start year IMO - I pinged legal to get confirmation on this.

JonathanOppenheimer:
> My understanding is that the updating the end year doesn't actually do anything to update our copyright. We should probably be just specifying the start year IMO - I pinged legal to get confirmation on this.

Oh should we change all of the headers in the repo to just copyright 2019? This would prevent having to update all of them every year. I assumed because the others were, this should be as well.

joshua-kim:
Yeah that is what I am suggesting. I got back from legal and got confirmation - the end year does not need updating. Can we rescope this PR to remove the end year?

JonathanOppenheimer:
> Yeah that is what I am suggesting. I got back from legal and got confirmation - the end year does not need updating. Can we rescope this PR to remove the end year?

Yes, and I will expand it to completely remove it from everywhere. Out of pure curiosity did they just say yes, or is there a reason it's unnecessary?

joshua-kim:
> Yes, and I will expand it to completely remove it from everywhere. Out of pure curiosity did they just say yes, or is there a reason it's unnecessary?

Copyright protections aren't magically updated because we update a license file every year - and conversely we don't suddenly lose our intellectual property if we forget to update the license file. The BSD license doesn't even include an end year - you can check the git history on the license file but someone added it after-the-fact. My guess is that they thought that the year was only valid for the current year and that updating the end year each year somehow extended our copyright, which led to future engineers experiencing confusion when seeing a date range which required an update annually.

JonathanOppenheimer:
> > Yes, and I will expand it to completely remove it from everywhere. Out of pure curiosity did they just say yes, or is there a reason it's unnecessary?
> 
> Copyright protections aren't magically updated because we update a license file every year - and conversely we don't suddenly lose our intellectual property if we forget to update the license file. The BSD license doesn't even include an end year - you can check the git history on the license file but someone added it after-the-fact. My guess is that they thought that the year was only valid for the current year and that updating the end year each year somehow extended our copyright, which led to future engineers experiencing confusion when seeing a date range which required an update annually.

Ah interesting! Well this is a good change then!

JonathanOppenheimer:
Re-opened here with a better branch name: https://github.com/ava-labs/avalanchego/pull/4812

PR #4811, opened 2026-01-06
refactor: Firewood Config helper

alarso16:
## Why this should be merged

This is a small diff pulled out of #4697 to support state sync test for Firewood.

## How this works

Renames some fields for clarity, provides a simpler construction method, and exports the underlying database

## How this was tested

Existing UT

No

### DIFF
--- a/graft/evm/firewood/triedb.go
+++ b/graft/evm/firewood/triedb.go
@@ -63,25 +63,24 @@ type ProposalContext struct {
 	Children []*ProposalContext
 }
 
-// Config provides necessary parameters for creating a Firewood database.
-type Config struct {
-	DatabaseDirPath      string
+type TrieDBConfig struct {
+	DatabaseDir          string
 	CacheSizeBytes       uint
 	FreeListCacheEntries uint
 	RevisionsInMemory    uint // must be >= 2
 	CacheStrategy        ffi.CacheStrategy
-	Archive              bool // whether to write keep all historical revisions on disk
+	Archive              bool
 }
 
-// DefaultConfig returns a default Config with the given directory.
+// DefaultConfig returns a sensible TrieDBConfig with the given directory.
 // The default config is:
 //   - CacheSizeBytes: 1MB
 //   - FreeListCacheEntries: 40,000
 //   - RevisionsInMemory: 100
 //   - CacheStrategy: [ffi.CacheAllReads]
-func DefaultConfig(dir string) Config {
-	return Config{
-		DatabaseDirPath:      dir,
+func DefaultConfig(dir string) TrieDBConfig {
+	return TrieDBConfig{
+		DatabaseDir:          dir,
 		CacheSizeBytes:       1024 * 1024, // 1MB
 		FreeListCacheEntries: 40_000,
 		RevisionsInMemory:    100,
@@ -89,7 +88,7 @@ func DefaultConfig(dir string) Config {
 	}
 }
 
-func (c Config) BackendConstructor(ethdb.Database) triedb.DBOverride {
+func (c TrieDBConfig) BackendConstructor(ethdb.Database) triedb.DBOverride {
 	db, err := New(c)
 	if err != nil {
 		log.Crit("firewood: error creating database", "error", err)
@@ -116,8 +115,8 @@ type TrieDB struct {
 
 // New creates a new Firewood database with the given disk database and configuration.
 // Any error during creation will cause the program to exit.
-func New(config Config) (*TrieDB, error) {
-	path := filepath.Join(config.DatabaseDirPath, firewoodDir)
+func New(config TrieDBConfig) (*TrieDB, error) {
+	path := filepath.Join(config.DatabaseDir, firewoodDir)
 	if err := validatePath(path); err != nil {
 		return nil, err
 	}
--- a/graft/coreth/core/blockchain.go
+++ b/graft/coreth/core/blockchain.go
@@ -230,8 +230,8 @@ func (c *CacheConfig) triedbConfig() *triedb.Config {
 			log.Crit("Chain data directory must be specified for Firewood")
 		}
 
-		config.DBOverride = firewood.Config{
-			DatabaseDirPath:      c.ChainDataDir,
+		config.DBOverride = firewood.TrieDBConfig{
+			DatabaseDir:          c.ChainDataDir,
 			CacheSizeBytes:       uint(c.TrieCleanLimit * 1024 * 1024),
 			FreeListCacheEntries: 40_000,               // Firewood default
 			RevisionsInMemory:    uint(c.StateHistory), // must be at least 2
--- a/graft/subnet-evm/core/blockchain.go
+++ b/graft/subnet-evm/core/blockchain.go
@@ -242,8 +242,8 @@ func (c *CacheConfig) triedbConfig() *triedb.Config {
 			log.Crit("Chain data directory must be specified for Firewood")
 		}
 
-		config.DBOverride = firewood.Config{
-			DatabaseDirPath:      c.ChainDataDir,
+		config.DBOverride = firewood.TrieDBConfig{
+			DatabaseDir:          c.ChainDataDir,
 			CacheSizeBytes:       uint(c.TrieCleanLimit * 1024 * 1024),
 			FreeListCacheEntries: 40_000,               // Firewood default
 			RevisionsInMemory:    uint(c.StateHistory), // must be at least 2


graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
+// Config provides necessary parameters for creating a Firewood database.
 type Config struct {
-	ChainDataDir         string
-	CleanCacheSize       int  // Size of the clean cache in bytes
-	FreeListCacheEntries uint // Number of free list entries to cache
-	Revisions            uint // Number of revisions to keep in memory (must be >= 2)
-	ReadCacheStrategy    ffi.CacheStrategy
-	ArchiveMode          bool
+	DatabasePath         string // directory where the database files will be stored
+	CacheSizeBytes       uint
```
JonathanOppenheimer, line 68: woah... self documenting code...

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
-var Defaults = Config{
-	CleanCacheSize:       1024 * 1024, // 1MB
-	FreeListCacheEntries: 40_000,
-	Revisions:            100,
-	ReadCacheStrategy:    ffi.CacheAllReads,
+// DefaultConfig returns a default Config with the given directory.
+// The default config is:
+//   - CacheSizeBytes: 1MB
+//   - FreeListCacheEntries: 40,000
+//   - MaxRevisions: 100
```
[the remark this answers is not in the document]
JonathanOppenheimer: this makes sense, but also I'm not ure how useful the comment is as you can just read the default config directly. I might remove the comment.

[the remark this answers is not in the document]
alarso16: It's for the godoc, but I should definitely match the name...

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
-	FreeListCacheEntries: 40_000,
-	Revisions:            100,
-	ReadCacheStrategy:    ffi.CacheAllReads,
+// DefaultConfig returns a default Config with the given directory.
+// The default config is:
+//   - CacheSizeBytes: 1MB
+//   - FreeListCacheEntries: 40,000
+//   - RevisionsInMemory: 100
+//   - CacheStrategy: [ffi.CacheAllReads]
+func DefaultConfig(dir string) Config {
```
RodrigoVillar:
```suggestion
func NewConfig(dir string) Config {
```

(nit/feel-free-to-ignore): I would rename this as it's evident that this function will populate all fields (as otherwise, you would just use `Config{}`). If you choose to rename, I would also document in the godoc comment that this function populates default values.

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
 
+// Config provides necessary parameters for creating a Firewood database.
 type Config struct {
-	ChainDataDir         string
-	CleanCacheSize       int  // Size of the clean cache in bytes
-	FreeListCacheEntries uint // Number of free list entries to cache
-	Revisions            uint // Number of revisions to keep in memory (must be >= 2)
-	ReadCacheStrategy    ffi.CacheStrategy
-	ArchiveMode          bool
+	DatabasePath         string // directory where the database files will be stored
```
RodrigoVillar: I think this comment is repetitive. Alternatively, I would rename this to `DatabaseDirPath` if you want to make it clear that it's a directory.

graft/evm/firewood/triedb.go
```diff
@@ -89,7 +98,12 @@ func (c Config) BackendConstructor(ethdb.Database) triedb.DBOverride {
 }
 
 type TrieDB struct {
-	fwDisk       *ffi.Database // The underlying Firewood database, used for storing proposals and revisions.
+	// The underlying Firewood database, used for storing proposals and revisions.
+	// This is exported with the knowledge that consumer will not close it and the latest state can be modified
+	// at any time via block execution. The consumer should only use for read operations,
+	// or ensure that writes occur outside of block execution.
+	Firewood *ffi.Database
```
RodrigoVillar, line 104: q: who is the expected consumer of this exported field (a code snippet would be awesome here)?

alarso16: I need access to create the x/syncer:
```go
import (
     "github.com/ava-labs/avalanchego/firewood/syncer"
     "github.com/ava-labs/avalanchego/graft/coreth/sync"
     "github.com/ava-labs/avalanchego/graft/evm/firewood"
     "github.com/ava-labs/avalanchego/network/p2p"
     
     "github.com/ava-labs/libevm/common"
     "github.com/ava-labs/libevm/triedb"
 )
 
func createFirewoodSyncer(db *triedb.Database, root common.Hash, rpClient, cpClient *p2p.Client) sync.Syncer {
     fwTrieDB, ok := db.Backend().(*firewood.TrieDB)
     if !ok {
         return nil, errors.New("not a firewood database")
     }
     s, err := syncer.New(syncer.Config{}, fwTrieDB.Firewood, root, rpClient, cpClient)
     if err != nil {
          return err
     }
     return s, nil
 }
 ```
 
I couldn't come up with a better way to provide it. I added all the imports if you wanted to check out the code I'm referring to

File: graft/evm/firewood/triedb.go (hunk 3 above)
Comment (alarso16) [reply graft/evm/firewood/triedb.go]: I think I'd rather keep it as `DefaultConfig`, simply because `NewConfig` would usually imply that this is the only safe method of instantiating a `Config`

File: graft/evm/firewood/triedb.go (hunk 5 above)
Comment (RodrigoVillar) [reply graft/evm/firewood/triedb.go:104]: Gotcha - makes sense

Comment (alarso16) [reply graft/evm/firewood/triedb.go:104]: I could add a method and make `Firewood` unexported, but I don't know if it makes any difference

Comment (RodrigoVillar) [reply graft/evm/firewood/triedb.go:104]: I think leaving it exported for now makes sense - adding a getter wouldn't fix the problem + introduces indirection.

File: graft/evm/firewood/triedb.go [hunk 6]
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
 }
 
-// Note that `FilePath` is not specified, and must always be set by the user.
-var Defaults = Config{
-	CleanCacheSize:       1024 * 1024, // 1MB
-	FreeListCacheEntries: 40_000,
-	Revisions:            100,
-	ReadCacheStrategy:    ffi.CacheAllReads,
+// DefaultConfig returns a default Config with the given directory.
+// The default config is:
```
maru-ava, line 76: Why is it desirable to require that these defaults be maintained twice?

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
-	FreeListCacheEntries uint // Number of free list entries to cache
-	Revisions            uint // Number of revisions to keep in memory (must be >= 2)
-	ReadCacheStrategy    ffi.CacheStrategy
-	ArchiveMode          bool
+	DatabaseDirPath      string
+	CacheSizeBytes       uint
+	FreeListCacheEntries uint
+	RevisionsInMemory    uint // must be >= 2
+	CacheStrategy        ffi.CacheStrategy
+	Archive              bool // whether to write keep all historical revisions on disk
```
maru-ava: (No action required) Maybe use a name that is self-documenting? This comment would seem to suggest that 'Archive' by itself is not clear.

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
...
 
+// Config provides necessary parameters for creating a Firewood database.
 type Config struct {
-	ChainDataDir         string
-	CleanCacheSize       int  // Size of the clean cache in bytes
-	FreeListCacheEntries uint // Number of free list entries to cache
-	Revisions            uint // Number of revisions to keep in memory (must be >= 2)
-	ReadCacheStrategy    ffi.CacheStrategy
-	ArchiveMode          bool
+	DatabaseDirPath      string
```
maru-ava: (No action required) Why not just `DatabaseDir` or `DatabasePath`? DirPath seems redundant.

graft/evm/firewood/triedb.go
```diff
@@ -63,21 +63,30 @@ type ProposalContext struct {
 	Children []*ProposalContext
 }
 
+// Config provides necessary parameters for creating a Firewood database.
```
maru-ava: (No action required) Maybe rename to avoid requiring this comment e.g. `FirewoodConfig` or `FirewoodDBConfig`?

alarso16: The package is called `firewood`, so any usage would be `firewood.Config`. Maybe `firewood.TrieDBConfig`?

graft/evm/firewood/triedb.go
alarso16: I did just have a default struct, but the usage was rather cumbersome (if you look in the other files of the diff), since one of the config fields must be populated (it shouldn't be in the config, but I have to maintain the interface). To avoid this, the constructor does help IMO

graft/evm/firewood/triedb.go
alarso16: This is an interesting question! This just what Firewood calls it, but I think that this isn't explicitly for "archiving", but rather storing all the historical revisions on disk.

maru-ava: StoreHistoricalRevisions?

graft/evm/firewood/triedb.go
maru-ava: Apologies for not being clear, I was questioning why it is necessary to duplicate the default values in the doc string so that they have to be maintained twice.

graft/evm/firewood/triedb.go
maru-ava: I guess I was more wondering if the name being generic was why the docstring appeared to be redundant. But not critical.

graft/evm/firewood/triedb.go
alarso16: I think the meaning is likely understood well, so I'll just remove the typo-ridden comment

PR #4814, opened 2026-01-06
refactor: Use `state.Database` to generate tries in sync test helpers

alarso16:
## Why this should be merged

All the syncing helpers rely on the `TrieDB` being backed by a key value store, so the key value store is passed around and a interface around it is constructed from this. However, this will not work for firewood, as the state is detached from a key value store. To enable reuse of most of these helpers, they're refactored slightly to allow it.

## How this works

Refactors sync trie helpers, sharing between coreth and subnet-evm

## How this was tested

UT

No


graft/subnet-evm/sync/statesync/sync_test.go
```diff
@@ -583,37 +569,35 @@ func assertDBConsistency(t testing.TB, root common.Hash, clientDB ethdb.Database
 	require.Equal(t, trieAccountLeaves, numSnapshotAccounts)
 }
```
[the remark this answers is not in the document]
alarso16: I think this is unnecessary

graft/coreth/sync/statesync/sync_test.go
```diff
@@ -51,26 +51,28 @@ func testSync(t *testing.T, test syncTest) {
 		ctx = test.ctx
 	}
 	r := rand.New(rand.NewSource(1))
-	clientDB, serverDB, serverTrieDB, root := test.prepareForTest(t, r)
-	leafsRequestHandler := handlers.NewLeafsRequestHandler(serverTrieDB, message.StateTrieKeyLength, nil, message.Codec, handlerstats.NewNoopHandlerStats())
-	codeRequestHandler := handlers.NewCodeRequestHandler(serverDB, message.Codec, handlerstats.NewNoopHandlerStats())
+	clientDB, serverDB, root := test.prepareForTest(t, r)
+	clientEthDB := clientDB.DiskDB().(ethdb.Database)
+	serverEthDB := serverDB.DiskDB().(ethdb.Database)
```
JonathanOppenheimer: We pass a `state.database `but then immediately unwrap it to a `ethdb.Database`. Should `NewCodeQueue`, `NewCodeSyncer`, and all the other handlers accept `state.Database instead`? There might be a reason for this, but to me it appears like these were missed (in the conversion that is)

graft/coreth/sync/statesync/sync_test.go
```diff
@@ -51,26 +51,28 @@ func testSync(t *testing.T, test syncTest) {
 		ctx = test.ctx
 	}
 	r := rand.New(rand.NewSource(1))
-	clientDB, serverDB, serverTrieDB, root := test.prepareForTest(t, r)
-	leafsRequestHandler := handlers.NewLeafsRequestHandler(serverTrieDB, message.StateTrieKeyLength, nil, message.Codec, handlerstats.NewNoopHandlerStats())
-	codeRequestHandler := handlers.NewCodeRequestHandler(serverDB, message.Codec, handlerstats.NewNoopHandlerStats())
+	clientDB, serverDB, root := test.prepareForTest(t, r)
+	clientEthDB := clientDB.DiskDB().(ethdb.Database)
```
JonathanOppenheimer:
```suggestion
	clientEthDB, ok := clientDB.DiskDB().(ethdb.Database)
	require.True(t, ok, "expected ethdb.Database from DiskDB()")
```

We should check these here, and in all the other locations you do similar.

graft/evm/sync/synctest/test_trie.go
```diff
@@ -0,0 +1,235 @@
...
+		return account
+	})
+}
+
+// GenerateIndependentTrie creates a trie with [numKeys] random key-value pairs inside of [trieDB].
+// Returns the root of the generated trie, the slice of keys inserted into the trie in lexicographical
+// order, and the slice of corresponding values.
+//
+// This is safe for use with HashDB.
+func GenerateIndependentTrie(t *testing.T, r *rand.Rand, trieDB *triedb.Database, numKeys int, keySize int) (common.Hash, [][]byte, [][]byte) {
```
JonathanOppenheimer, line 63: What are these functions independent from? probably missing context.

graft/evm/sync/synctest/test_trie.go
```diff
@@ -0,0 +1,235 @@
...
+			CodeHash: types.EmptyCodeHash[:],
+			Root:     types.EmptyRootHash,
+		}
+		if onAccount != nil {
+			storageTr, err := s.OpenStorageTrie(root, key.Address, types.EmptyRootHash, tr)
+			require.NoError(t, err)
+			acc = onAccount(t, i, key.Address, acc, storageTr)
+			root, nodes, err := storageTr.Commit(false)
+			require.NoError(t, err)
+			if root != types.EmptyRootHash {
```
JonathanOppenheimer, line 190: add check to make sure nodes is nil? or this works?

graft/coreth/sync/statesync/sync_test.go
```diff
@@ -587,19 +574,18 @@ func assertDBConsistency(t testing.TB, root common.Hash, clientDB ethdb.Database
 	require.Equal(t, trieAccountLeaves, numSnapshotAccounts)
 }
 
-func fillAccountsWithStorage(t *testing.T, r *rand.Rand, serverDB ethdb.Database, serverTrieDB *triedb.Database, root common.Hash, numAccounts int) common.Hash {
-	newRoot, _ := statesynctest.FillAccounts(t, r, serverTrieDB, root, numAccounts, func(_ *testing.T, _ int, account types.StateAccount) types.StateAccount {
+func fillAccountsWithStorage(t *testing.T, r *rand.Rand, serverDB state.Database, numAccounts int) common.Hash {
```
JonathanOppenheimer: Can we move this function to the shared package? -- it's duplicated

graft/evm/sync/synctest/test_trie.go
```diff
@@ -0,0 +1,235 @@
...
+	keys := make([][]byte, 0, numKeys)
+	values := make([][]byte, 0, numKeys)
+
+	// Generate key-value pairs
+	for i := start; i < numKeys; i++ {
+		key := make([]byte, keySize)
+		binary.BigEndian.PutUint64(key[:wrappers.LongLen], uint64(i+1))
+		_, err := r.Read(key[wrappers.LongLen:])
+		require.NoError(t, err)
+		value := make([]byte, r.Intn(128)+128) // min 128 bytes, max 256 bytes
```
JonathanOppenheimer:
```suggestion
		value := make([]byte, r.Intn(128)+128) // min 128 bytes, max 255 bytes
```

https://pkg.go.dev/math/rand#Intn

graft/subnet-evm/sync/statesync/sync_test.go
```diff
@@ -15,19 +15,19 @@ import (
```
JonathanOppenheimer, line 15: same comments from subnet-evm

graft/coreth/sync/statesync/sync_test.go
JonathanOppenheimer: (if you choose to keep this as an ethdb that is -- the above comment is more important)

graft/coreth/sync/statesync/sync_test.go
alarso16: `ethdb.Database` represents a key value store. `state.Database` represents a key value store and the TrieDB (which may or may not use the key value store). Specifically, the code queue and and code syncer do not need the TrieDB.

graft/evm/sync/synctest/test_trie.go
alarso16: I don't love the name, but it's not a regular geth trie. It's just "a trie with geth hashing", separated from accounts and storage structure. Specifically, it should be used for the atomic trie, but can be used to create storage tries.

graft/evm/sync/synctest/test_trie.go
alarso16: Ha I didn't write this, but sure

graft/evm/sync/synctest/test_trie.go
alarso16: This is all you need to do. There's some conventions about how these should be use (look at `statedb.Commit`). This whole interface is strange to me

graft/evm/sync/synctest/test_trie.go
JonathanOppenheimer: do you want to add that to the comment? independent of the accounts and storage structure

graft/coreth/sync/statesync/sync_test.go
alarso16: I can do this, but I'm not convinced it's necessary. We understand it to be true because we just created the database, but also, so what if we panic in a test? Is the more graceful handling worth the increase in complexity?

Seems pretty inconsequential though, so I'll just add add it. Just curious on opinions here

graft/evm/sync/synctest/test_trie.go
JonathanOppenheimer: lol sorry, you can hover over it in cursor!

graft/coreth/sync/statesync/sync_test.go
JonathanOppenheimer: If you think it's overkill no need. I honestly thought it was a linter to not ignore the return values.

graft/subnet-evm/sync/statesync/sync_test.go
alarso16: I believe I addressed them for both

JonathanOppenheimer:
LGTM at a micro level, someone should evaluate at a macro level.

powerslider:
LGTM

graft/subnet-evm/sync/statesync/sync_test.go
```diff
@@ -52,9 +52,12 @@ func testSync(t *testing.T, test syncTest) {
 		ctx = test.ctx
 	}
 	r := rand.New(rand.NewSource(1))
-	clientDB, serverDB, serverTrieDB, root := test.prepareForTest(t, r)
-	leafsRequestHandler := handlers.NewLeafsRequestHandler(serverTrieDB, message.StateTrieKeyLength, nil, message.Codec, handlerstats.NewNoopHandlerStats())
-	codeRequestHandler := handlers.NewCodeRequestHandler(serverDB, message.Codec, handlerstats.NewNoopHandlerStats())
+	clientDB, serverDB, root := test.prepareForTest(t, r)
+	clientEthDB, ok := clientDB.DiskDB().(ethdb.Database)
```
ceyonur, line 56: should we also do the same for serverEthDb as in coreth?

alarso16: Oh it's actually the opposite. It was unnecessary in coreth, so I removed it there

PR #4816, opened 2026-01-06
Snapshot generation shutdown race

JonathanOppenheimer:
## Why this should be merged
See this PR for a full explanation of the issue: https://github.com/ethereum/go-ethereum/pull/33540. 

We've been getting pages about this for a specific node, and after investigation, I determined that the issue was upstream in `geth`. As an addendum, our copy of geth is really weird. We have variations of changes they made, and some stuff is synced, and some isn't. Even libevm is off in this case. 

## How this works
See above link PR

## How this was tested
CI, and a new test, `testReleaseStopsGeneration` 

No


JonathanOppenheimer:
If we'd prefer not to merge this and try to cherry-pick the upstream PR, we could also do that.

JonathanOppenheimer:
> Your code works I believe, but to check that the goroutine actually exits, you can add this to the testmain:
> 
> ```go
> goleak.VerifyTestMain(m, goleak.IgnoreCurrent())
> ```
> 
> Additionally, there's a line in the TestMain for `core` that explicitly excludes the generate goroutine, and that can be deleted I believe, to ensure that the generator actually does close.
> 
> I think that we've deviated in this package sufficiently that a cherry-pick wouldn't actually be productive, but I'm good either way
> 
> Finally, we do also have this code in subnet-evm. Should we update this at the same time?

There's actually a very subtle deadlock bug in this implementation where even `dl.genAbort` is not sufficient to determine whether generation has been ordered to cease. I'm glad I made the geth PR, because our test coverage was not sufficient to catch this. Another reason we should work towards getting everything in libevm with the full geth test suite! 

Re: subnet-evm -- yes, made the changes as well. 

As for the main test, great point. I've removed the exception. 

I also move the function, and naming to exactly align it with geth (or at least the version of geth in the PR I have open to it).

graft/coreth/core/main_test.go
```diff
@@ -29,7 +29,6 @@ func TestMain(m *testing.M) {
 
 	opts := []goleak.Option{
 		// No good way to shut down these goroutines:
-		goleak.IgnoreTopFunction("github.com/ava-labs/avalanchego/graft/coreth/core/state/snapshot.(*diskLayer).generate"),
```
JonathanOppenheimer: #1 victory royale

graft/evm/core/state/snapshot/snapshot.go
```diff
@@ -594,33 +594,7 @@ func (t *Tree) AbortGeneration() {
 	defer t.lock.Unlock()
 
 	dl := t.disklayer()
-	dl.abortGeneration()
-}
-
-// abortGeneration sends an abort message to the generate goroutine and waits
-// for it to shutdown before returning (if it is running). This call should not
-// be made concurrently.
-func (dl *diskLayer) abortGeneration() bool {
```
JonathanOppenheimer, line 603: For some reason, we renamed and moved the actual version of this function, `func (dl *diskLayer) stopGeneration()`, here and renamed it from stop to abort instead of just modifying the original version of the function. 

It appears Patrick did this 4 years ago, and it seems like a needless divergence from geth.

graft/evm/core/state/snapshot/snapshot.go
```diff
@@ -594,33 +594,7 @@ func (t *Tree) AbortGeneration() {
...
-	// Store ideal time for abort to get better estimate of load
-	//
-	// Note that we set this time regardless if abortion was skipped otherwise we
-	// will never restart generation (age will always be negative).
-	if dl.abortStarted.IsZero() {
```
JonathanOppenheimer, line 608: This optimization is the actual divergence from geth (I maintained this post alignment if you check stopGeneration)

graft/coreth/core/state/snapshot/generate_test.go
alarso16: Put `	goleak.VerifyTestMain(m, goleak.IgnoreCurrent())` here too

graft/coreth/core/state/snapshot/disklayer.go
```diff
@@ -199,3 +206,45 @@ func (dl *diskLayer) Storage(accountHash, storageHash common.Hash) ([]byte, erro
...
 	return newDiffLayer(dl, blockHash, blockRoot, destructs, accounts, storage)
 }
+
+// stopGeneration aborts the state snapshot generation if it is currently running.
+func (dl *diskLayer) stopGeneration() {
+	// Check if generation goroutine is actually running.
+	//
+	// Note: genMarker can be nil even when the generator is still running (waiting
+	// for abort signal after completing generation), so we can't rely on genMarker.
+	if !dl.genRunning.Load() {
```
alarso16: What's the point of this atomic bool? can't you just check the to see if you set the channel to nil?

JonathanOppenheimer: No... there's no so many weird edge cases with all the channels and concurrent accesses, deferments etc. that determining whether the generator is running by checking the channel doesn't work. If you check the geth PR, you can see a few iterations I tried, and the test failures they induced.

JonathanOppenheimer: the atomic bool is just like the best I could think of to be like "is this ACTUALLY running RIGHT now". An alternative (and probably superior) solution would be to do this properly with contexts, but that would have created a lot more changes in geth.

graft/coreth/core/state/snapshot/generate_test.go
JonathanOppenheimer: Is this not already covered by the overarching test main go leak checks in `main_test.go`. If it doesn't add anything, I'd prefer not to introduce an unnecessary divergence.

alarso16: That's in a different package, so no. You need both

JonathanOppenheimer: the testmain doesn't even exist in subnet-evm. I will add both and hope it doesn't fail...

graft/coreth/core/blockchain.go
```diff
@@ -998,10 +1003,6 @@ func (bc *BlockChain) stopWithoutSaving() {
 func (bc *BlockChain) Stop() {
 	bc.stopWithoutSaving()
 
-	// Ensure that the entirety of the state snapshot is journaled to disk.
```
JonathanOppenheimer, line 1000: This comment is SO misleading `bc.snaps.Release()` doesn't do any journaling. There used to be a whole big code chunk here that did journal the snapshot to disk, but it was all removed from our repo's at some point, and the comment was left...

powerslider:
@JonathanOppenheimer I hope you don't mind posting a fix on top of your current branch (https://github.com/ava-labs/avalanchego/pull/4823) to demonstrate a more robust approach that does not rely on the brittleness of `genAbort`. Feel free to review it and we can discuss it together. I decided to do another branch, because it required changes in many places and would be easier to communicate this way.

graft/evm/core/state/snapshot/disklayer.go
```diff
@@ -199,3 +207,25 @@ func (dl *diskLayer) Storage(accountHash, storageHash common.Hash) ([]byte, erro
 func (dl *diskLayer) Update(blockHash, blockRoot common.Hash, destructs map[common.Hash]struct{}, accounts map[common.Hash][]byte, storage map[common.Hash]map[common.Hash][]byte) *diffLayer {
 	return newDiffLayer(dl, blockHash, blockRoot, destructs, accounts, storage)
 }
+
+// stopGeneration aborts the state snapshot generation if it is currently running.
+func (dl *diskLayer) stopGeneration() {
```
alarso16, line 223: The concurrency guarantees are not clear to me. Will you document what is guaranteed and when it is safe to be called?

JonathanOppenheimer: I lifted the comment largely from Tsvetan's PR - is it acceptable to you now?

alarso16: Are there writers to `dl.cancel` or `dl.done`? If not, I'm happy with this

graft/evm/core/state/snapshot/disklayer.go
ceyonur, line 1: I haven't really reviewed changes yet but if this is a bug in upstream I think we should wait that PR to be merged first. 

Changes looked a bit invasive to me so we should make sure this won't break any invariant for the upstream. (although it looks like they're not %100 identical)

JonathanOppenheimer: Yeah do you know how I can prompt someone in geth to review the PR?

JonathanOppenheimer: They are not 100% identical because we have this optimization added years ago that skips generation if it'd be stale soon, that doesn't exist in geth.

graft/evm/core/state/snapshot/generate_test.go
alarso16, line 1: Do you want to add goleak to this package's `TestMain`?

graft/evm/core/state/snapshot/generate_test.go
```diff
@@ -969,8 +969,89 @@ func testGenerateBrokenSnapshotWithDanglingStorage(t *testing.T, scheme string)
...
+func TestReleaseStopsGeneration(t *testing.T) {
+	for _, scheme := range []string{rawdb.HashScheme, rawdb.PathScheme} {
+		t.Run(scheme+"/after_completion", func(t *testing.T) {
+			testReleaseStopsCompletedGeneration(t, scheme)
+		})
+		t.Run(scheme+"/during_generation", func(t *testing.T) {
+			testReleaseStopsMidGeneration(t, scheme)
+		})
+	}
+}
```
alarso16: These are really two separate tests. I would make two test functions

graft/evm/core/state/snapshot/generate_test.go
```diff
@@ -969,8 +969,89 @@ func testGenerateBrokenSnapshotWithDanglingStorage(t *testing.T, scheme string)
...
+	helper.makeStorageTrie(hashData([]byte("acc-1")), []string{"key-1", "key-2", "key-3"}, []string{"val-1", "val-2", "val-3"}, true)
+	helper.makeStorageTrie(hashData([]byte("acc-3")), []string{"key-1", "key-2", "key-3"}, []string{"val-1", "val-2", "val-3"}, true)
+
+	_, snap := helper.CommitAndGenerate()
+
+	// Call Release immediately without waiting for generation to finish.
+	// This exercises the cancel signal through checkAndFlush.
+	done := make(chan struct{})
+	go func() {
+		snap.Release()
```
alarso16: Doesn't this return an error?

graft/evm/core/state/snapshot/generate_test.go
```diff
@@ -969,8 +969,89 @@ func testGenerateBrokenSnapshotWithDanglingStorage(t *testing.T, scheme string)
...
+		close(done)
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
```
alarso16: I don't think having this in a goroutine is helpful. `go test` already has a timeout (albeit quite long)

graft/evm/core/state/snapshot/generate_test.go
JonathanOppenheimer: I've split these into two tests.

graft/evm/core/state/snapshot/generate_test.go
JonathanOppenheimer: I added a `TestMain` for this package -- one did not exist prior.

graft/evm/core/state/snapshot/snapshot.go
ceyonur: I came to say keep that here for minimal diff but then saw this comment. Good catch!

graft/evm/core/state/snapshot/disklayer.go
```diff
@@ -51,9 +51,15 @@ type diskLayer struct {
...
 	stale     bool        // Signals that the layer became stale (state progressed)
 
-	genMarker  []byte             // Marker for the state that's indexed during initial layer generation
-	genPending chan struct{}      // Notification channel when generation is done (test synchronicity)
-	genAbort   chan chan struct{} // Notification channel to abort generating the snapshot in this layer
+	genMarker  []byte        // Marker for the state that's indexed during initial layer generation
+	genPending chan struct{} // Notification channel when generation is done (test synchronicity)
+
+	// Generator lifecycle management:
+	// - [cancel] is closed to request termination (broadcast).
```
ceyonur, line 58:
```suggestion
	// - [cancel] is closed (via [cancelOnce]) to request termination (broadcast).
```

graft/evm/core/state/snapshot/disklayer.go
```diff
@@ -199,3 +207,41 @@ func (dl *diskLayer) Storage(accountHash, storageHash common.Hash) ([]byte, erro
...
+// After return, it is **guaranteed** that:
+//   - The generator goroutine has terminated
+//   - It is safe to proceed with cleanup operations (e.g. closing databases)
+func (dl *diskLayer) stopGeneration() {
+	cancel := dl.cancel
+	done := dl.done
+	if cancel == nil || done == nil {
+		// Generation was skipped for this layer so there is nothing to stop.
+		// Note: abortStarted remains zero, which diffToDisk accounts for
+		// by falling back to time.Now() when computing disk layer age.
```
ceyonur: would it make sense to set `base.abortStarted` before the early exit so we can remove this invariant and `IsZero` check in `diffToDisk`?

graft/evm/core/state/snapshot/disklayer.go
ceyonur: We still need to reflect these changes to libevm as strevm used libevm's snapshot package. if this is an immediate issue for coreth+subnet-evm we can merge this PR but we still need to track upstream PR or at least have another PR for libevm.

graft/evm/core/state/snapshot/disklayer.go
alarso16: I think this godoc still won't work. You can just see in your editor that this links correctly

JonathanOppenheimer: Yeah this doesn't work.

graft/evm/core/state/snapshot/disklayer.go
JonathanOppenheimer: I don't know if this is an immediate issue - we've lived with for years - but is causing pages/incidents which we can resolve. Do you have any better idea of trying to get this reviewed in geth?

graft/evm/core/state/snapshot/disklayer.go
JonathanOppenheimer: ah this is a very good suggestion!

go.work.sum
joshua-kim, line 1: What is the diff in this file for?

JonathanOppenheimer: I think this is from the bazel task?

JonathanOppenheimer: wait nah this is def pollution sorry

joshua-kim: go.work.sum is dependent on all modules in the workspace's go.mods -- so if there isn't a diff in a go.mod it's unexpected to have a diff here.

JonathanOppenheimer: Yeah I think this carried over from when I was reviewing a firewood PR somehow -- I didn't change branches properly or smth?

PR #4818, opened 2026-01-06
fix: warp e2e flake

JonathanOppenheimer:
## Why this should be merged

The EVM warp E2E tests have been flaky since the monorepo because of time outs -- sometimes transactions take more than 30 seconds to get included, and it fails. See this for a recent flake: https://github.com/ava-labs/avalanchego/actions/runs/20761939951/job/59618890988?pr=4816

## How this works
I extended the timeout to 60 seconds, and immediately canceled, rather than lettering them accumulate via defer in the loop.  From reading online, it's generally not recommended to defer within loops, unless under closure, as it adds additional overhead. (and we're only 0.009s over the threshold in the example linked above). 

I also aligned subnet-evm and coreth's implementation of this file. 

## How this was tested
CI

No

### DIFF
--- a/graft/coreth/tests/utils/proposervm.go
+++ b/graft/coreth/tests/utils/proposervm.go
@@ -21,7 +21,9 @@ import (
 	ethparams "github.com/ava-labs/libevm/params"
 )
 
-const numTriggerTxs = 2 // Number of txs needed to activate the proposer VM fork
+// expectedBlockHeight is the block height that activates the proposerVM fork.
+// We issue 2 txs (one per block) to reach block height 2.
+const expectedBlockHeight = 2
 
 // IssueTxsToActivateProposerVMFork issues transactions at the current
 // timestamp, which should be after the ProposerVM activation time (aka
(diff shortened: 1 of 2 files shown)


graft/coreth/tests/utils/proposervm.go
```diff
@@ -21,7 +21,7 @@ import (
 	ethparams "github.com/ava-labs/libevm/params"
 )
 
-const expectedBlockHeight = 2
+const numTriggerTxs = 2 // Number of txs needed to activate the proposer VM fork
```
ceyonur: nit: I think this guides the reader into thinking that txs activate the proposervm, but it really is the block height no 2 that activates it. We should explicitly comment why it's 2 txs (or revert the name change)

graft/coreth/tests/utils/proposervm.go
```diff
@@ -39,11 +39,11 @@ func IssueTxsToActivateProposerVMFork(
 		return err
 	}
 
-	gasPrice := big.NewInt(ap1.MinGasPrice) // should be pretty generous for c-chain and subnets
+	gasPrice := big.NewInt(ap1.MinGasPrice)
```
ceyonur: I think that comment actually sort-of useful, otherwise it sounds like we're using the ap1 gas price specifically. if we think using `ap1.MinGasPrice` is itself misleading we can put some bigger number as a const and use it (with commented that it should be enough to cover)

graft/coreth/tests/utils/proposervm.go
```diff
@@ -55,17 +55,18 @@ func IssueTxsToActivateProposerVMFork(
...
 			return err
 		}
+		cancel()
 		nonce++
 	}
 
 	log.Info(
 		"Built sufficient blocks to activate proposerVM fork",
-		"blockCount", expectedBlockHeight,
+		"txCount", numTriggerTxs,
```
ceyonur: IMO this is weird to log (as stated above)

graft/coreth/tests/utils/proposervm.go
JonathanOppenheimer: That's fair -- I reverted the name changed and expanded the name change!

graft/coreth/tests/utils/proposervm.go
JonathanOppenheimer: I reverted this and restored the comment (this is a divergence between the two files now so a comment does make sense)

graft/coreth/tests/utils/proposervm.go
JonathanOppenheimer: Reverted!

alarso16 (approved):

PR #4819, opened 2026-01-07
Introduce gossip system helper

StephenButtolph:
## Why this should be merged

This does not really introduce new functionality. However, there is quite a lot of boilerplate code that can be reduced by using this helper throughout the codebase.

One thing this _does_ do is fix metric names. Some metrics were missing the "gossip" prefix. Others had a duplicate"gossip_gossip" prefix.
```
# HELP avalanche_avm_tx_bloomfilter_hit_rate Hit rate (%) of the bloom filter sent by pull gossip
# TYPE avalanche_avm_tx_bloomfilter_hit_rate histogram
avalanche_avm_tx_bloomfilter_hit_rate_bucket{chain="X",le="0"} 0
avalanche_avm_tx_bloomfilter_hit_rate_bucket{chain="X",le="25"} 0
avalanche_avm_tx_bloomfilter_hit_rate_bucket{chain="X",le="50"} 0
avalanche_avm_tx_bloomfilter_hit_rate_bucket{chain="X",le="75"} 0
avalanche_avm_tx_bloomfilter_hit_rate_bucket{chain="X",le="+Inf"} 0
avalanche_avm_tx_bloomfilter_hit_rate_sum{chain="X"} 0
avalanche_avm_tx_bloomfilter_hit_rate_count{chain="X"} 0
# HELP avalanche_platformvm_tx_bloomfilter_hit_rate Hit rate (%) of the bloom filter sent by pull gossip
# TYPE avalanche_platformvm_tx_bloomfilter_hit_rate histogram
avalanche_platformvm_tx_bloomfilter_hit_rate_bucket{chain="P",le="0"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_bucket{chain="P",le="25"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_bucket{chain="P",le="50"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_bucket{chain="P",le="75"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_bucket{chain="P",le="+Inf"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_sum{chain="P"} 0
avalanche_platformvm_tx_bloomfilter_hit_rate_count{chain="P"} 0

# HELP avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking number of gossipables being tracked
# TYPE avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking gauge
avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking{chain="C",type="sent"} 0
avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking{chain="C",type="unsent"} 0
# HELP avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking_lifetime_average average duration a gossipable has been tracked (ns)
# TYPE avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking_lifetime_average gauge
avalanche_evm_sdk_atomic_tx_gossip_gossip_tracking_lifetime_average{chain="C"} 0
# HELP avalanche_evm_sdk_eth_tx_gossip_gossip_tracking number of gossipables being tracked
# TYPE avalanche_evm_sdk_eth_tx_gossip_gossip_tracking gauge
avalanche_evm_sdk_eth_tx_gossip_gossip_tracking{chain="C",type="sent"} 0
avalanche_evm_sdk_eth_tx_gossip_gossip_tracking{chain="C",type="unsent"} 0
# HELP avalanche_evm_sdk_eth_tx_gossip_gossip_tracking_lifetime_average average duration a gossipable has been tracked (ns)
# TYPE avalanche_evm_sdk_eth_tx_gossip_gossip_tracking_lifetime_average gauge
avalanche_evm_sdk_eth_tx_gossip_gossip_tracking_lifetime_average{chain="C"} 0
```

These now report:
```
# HELP avalanche_avm_tx_gossip_bloomfilter_hit_rate Hit rate (%) of the bloom filter sent by pull gossip
# TYPE avalanche_avm_tx_gossip_bloomfilter_hit_rate histogram
avalanche_avm_tx_gossip_bloomfilter_hit_rate_bucket{chain="X",le="0"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_bucket{chain="X",le="25"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_bucket{chain="X",le="50"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_bucket{chain="X",le="75"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_bucket{chain="X",le="+Inf"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_sum{chain="X"} 0
avalanche_avm_tx_gossip_bloomfilter_hit_rate_count{chain="X"} 0
# HELP avalanche_platformvm_tx_gossip_bloomfilter_hit_rate Hit rate (%) of the bloom filter sent by pull gossip
# TYPE avalanche_platformvm_tx_gossip_bloomfilter_hit_rate histogram
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_bucket{chain="P",le="0"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_bucket{chain="P",le="25"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_bucket{chain="P",le="50"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_bucket{chain="P",le="75"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_bucket{chain="P",le="+Inf"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_sum{chain="P"} 0
avalanche_platformvm_tx_gossip_bloomfilter_hit_rate_count{chain="P"} 0

# HELP avalanche_evm_sdk_atomic_tx_gossip_tracking number of gossipables being tracked
# TYPE avalanche_evm_sdk_atomic_tx_gossip_tracking gauge
avalanche_evm_sdk_atomic_tx_gossip_tracking{chain="C",type="sent"} 0
avalanche_evm_sdk_atomic_tx_gossip_tracking{chain="C",type="unsent"} 0
# HELP avalanche_evm_sdk_atomic_tx_gossip_tracking_lifetime_average average duration a gossipable has been tracked (ns)
# TYPE avalanche_evm_sdk_atomic_tx_gossip_tracking_lifetime_average gauge
avalanche_evm_sdk_atomic_tx_gossip_tracking_lifetime_average{chain="C"} 0
# HELP avalanche_evm_sdk_eth_tx_gossip_tracking number of gossipables being tracked
# TYPE avalanche_evm_sdk_eth_tx_gossip_tracking gauge
avalanche_evm_sdk_eth_tx_gossip_tracking{chain="C",type="sent"} 0
avalanche_evm_sdk_eth_tx_gossip_tracking{chain="C",type="unsent"} 0
# HELP avalanche_evm_sdk_eth_tx_gossip_tracking_lifetime_average average duration a gossipable has been tracked (ns)
# TYPE avalanche_evm_sdk_eth_tx_gossip_tracking_lifetime_average gauge
avalanche_evm_sdk_eth_tx_gossip_tracking_lifetime_average{chain="C"} 0
```

## How this works

Adds a new helper and uses it on all the VMs (AVM + PlatformVM + Coreth + Subnet-EVM).

## How this was tested

Existing CI for tx gossip.

No.

### DIFF
--- a/graft/coreth/plugin/evm/atomic/vm/vm.go
+++ b/graft/coreth/plugin/evm/atomic/vm/vm.go
@@ -10,7 +10,6 @@ import (
 	"math/big"
 	"net/http"
 	"sync"
-	"time"
 
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/state"
@@ -88,7 +87,6 @@ type VM struct {
 	AtomicMempool        *txpool.Mempool
 	gossipHandler        p2p.Handler
 	pullGossiper         *avalanchegossip.ValidatorGossiper
-	pullGossipPeriod     time.Duration
 	AtomicTxPushGossiper *avalanchegossip.PushGossiper[*atomic.Tx]
 
 	// AtomicTxRepository maintains two indexes on accepted atomic txs.
@@ -185,32 +183,29 @@ func (vm *VM) Initialize(
 	}
 	vm.AtomicMempool = atomicMempool
 
-	systemConfig := avalanchegossip.SystemConfig{
-		Log:           chainCtx.Log,
-		Registry:      vm.InnerVM.MetricRegistry(),
-		Namespace:     "atomic_tx_gossip",
-		HandlerID:     p2p.AtomicTxGossipHandlerID,
-		RequestPeriod: vm.InnerVM.Config().PullGossipFrequency.Duration,
-		PushGossipParams: avalanchegossip.BranchingFactor{
-			StakePercentage: vm.InnerVM.Config().PushGossipPercentStake,
-			Validators:      vm.InnerVM.Config().PushGossipNumValidators,
-			Peers:           vm.InnerVM.Config().PushGossipNumPeers,
-		},
-		PushRegossipParams: avalanchegossip.BranchingFactor{
-			Validators: vm.InnerVM.Config().PushRegossipNumValidators,
-			Peers:      vm.InnerVM.Config().PushRegossipNumPeers,
-		},
-		RegossipPeriod: vm.InnerVM.Config().RegossipFrequency.Duration,
-	}
-	systemConfig.SetDefaults()
-	vm.pullGossipPeriod = systemConfig.RequestPeriod
 	vm.gossipHandler, vm.pullGossiper, vm.AtomicTxPushGossiper, err = avalanchegossip.NewSystem(
 		vm.Ctx.NodeID,
 		vm.InnerVM.P2PNetwork(),
 		vm.InnerVM.P2PValidators(),
 		atomicMempool,
 		&atomic.TxMarshaller{},
-		systemConfig,
+		avalanchegossip.SystemConfig{
+			Log:           chainCtx.Log,
+			Registry:      vm.InnerVM.MetricRegistry(),
+			Namespace:     "atomic_tx_gossip",
+			HandlerID:     p2p.AtomicTxGossipHandlerID,
+			RequestPeriod: vm.InnerVM.Config().PullGossipFrequency.Duration,
+			PushGossipParams: avalanchegossip.BranchingFactor{
+				StakePercentage: vm.InnerVM.Config().PushGossipPercentStake,
+				Validators:      vm.InnerVM.Config().PushGossipNumValidators,
+				Peers:           vm.InnerVM.Config().PushGossipNumPeers,
+			},
+			PushRegossipParams: avalanchegossip.BranchingFactor{
+				Validators: vm.InnerVM.Config().PushRegossipNumValidators,
+				Peers:      vm.InnerVM.Config().PushRegossipNumPeers,
+			},
+			RegossipPeriod: vm.InnerVM.Config().RegossipFrequency.Duration,
+		},
 	)
 	if err != nil {
 		return fmt.Errorf("failed to initialize atomic gossip system: %w", err)
@@ -307,7 +302,7 @@ func (vm *VM) onNormalOperationsStarted() error {
 
 	vm.shutdownWg.Add(1)
 	go func() {
-		avalanchegossip.Every(ctx, vm.Ctx.Log, vm.pullGossiper, vm.pullGossipPeriod)
+		avalanchegossip.Every(ctx, vm.Ctx.Log, vm.pullGossiper, vm.InnerVM.Config().PullGossipFrequency.Duration)
 		vm.shutdownWg.Done()
 	}()
 
--- a/graft/subnet-evm/plugin/evm/vm.go
+++ b/graft/subnet-evm/plugin/evm/vm.go
@@ -842,32 +842,28 @@ func (vm *VM) onNormalOperationsStarted() error {
 		vm.shutdownWg.Done()
 	}()
 
-	systemConfig := avalanchegossip.SystemConfig{
-		Log:           vm.ctx.Log,
-		Registry:      vm.sdkMetrics,
-		Namespace:     "eth_tx_gossip",
-		RequestPeriod: vm.config.PullGossipFrequency.Duration,
-		PushGossipParams: avalanchegossip.BranchingFactor{
-			StakePercentage: vm.config.PushGossipPercentStake,
-			Validators:      vm.config.PushGossipNumValidators,
-			Peers:           vm.config.PushGossipNumPeers,
-		},
-		PushRegossipParams: avalanchegossip.BranchingFactor{
-			Validators: vm.config.PushRegossipNumValidators,
-			Peers:      vm.config.PushRegossipNumPeers,
-		},
-		RegossipPeriod: vm.config.RegossipFrequency.Duration,
-	}
-	// Set the defaults so that we can use `config.PullGossipFrequency` after
-	// the default has been set.
-	systemConfig.SetDefaults()
 	handler, pullGossiper, pushGossiper, err := avalanchegossip.NewSystem(
 		vm.ctx.NodeID,
 		vm.P2PNetwork(),
 		vm.P2PValidators(),
 		ethTxPool,
 		GossipEthTxMarshaller{},
-		systemConfig,
+		avalanchegossip.SystemConfig{
+			Log:           vm.ctx.Log,
+			Registry:      vm.sdkMetrics,
+			Namespace:     "eth_tx_gossip",
+			RequestPeriod: vm.config.PullGossipFrequency.Duration,
+			PushGossipParams: avalanchegossip.BranchingFactor{
+				StakePercentage: vm.config.PushGossipPercentStake,
+				Validators:      vm.config.PushGossipNumValidators,
+				Peers:           vm.config.PushGossipNumPeers,
+			},
+			PushRegossipParams: avalanchegossip.BranchingFactor{
+				Validators: vm.config.PushRegossipNumValidators,
+				Peers:      vm.config.PushRegossipNumPeers,
+			},
+			RegossipPeriod: vm.config.RegossipFrequency.Duration,
+		},
 	)
 	if err != nil {
 		return fmt.Errorf("failed to initialize gossip system: %w", err)
@@ -902,7 +898,7 @@ func (vm *VM) onNormalOperationsStarted() error {
 	}()
 	vm.shutdownWg.Add(1)
 	go func() {
-		avalanchegossip.Every(ctx, vm.ctx.Log, vm.ethTxPullGossiper, systemConfig.RequestPeriod)
+		avalanchegossip.Every(ctx, vm.ctx.Log, vm.ethTxPullGossiper, vm.config.PullGossipFrequency.Duration)
 		vm.shutdownWg.Done()
 	}()
 
--- a/network/p2p/gossip/gossip.go
+++ b/network/p2p/gossip/gossip.go
@@ -611,9 +611,13 @@ func (p *PushGossiper[_]) updateMetrics(nowUnixNano float64) {
 	p.metrics.trackingLifetimeAverage.Set(averageLifetime)
 }
 
-// Every calls [Gossip] every [frequency] amount of time.
-func Every(ctx context.Context, log logging.Logger, gossiper Gossiper, frequency time.Duration) {
-	ticker := time.NewTicker(frequency)
+// Every calls [Gossip] every [period] amount of time.
+func Every(ctx context.Context, log logging.Logger, gossiper Gossiper, period time.Duration) {
+	if period <= 0 {
+		period = defaultRequestPeriod
+	}
+
+	ticker := time.NewTicker(period)
 	defer ticker.Stop()
 
 	for {
--- a/network/p2p/gossip/system.go
+++ b/network/p2p/gossip/system.go
@@ -16,6 +16,8 @@ import (
 	"github.com/ava-labs/avalanchego/utils/units"
 )
 
+const defaultRequestPeriod = time.Second
+
 // SystemConfig provides all the configurations needed to create a gossip
 // system.
 //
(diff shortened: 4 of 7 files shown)


network/p2p/gossip/gossip.go
StephenButtolph, line 1: The metric name changes are the only "real" changs in this PR.

A number of metrics are currently missing the "gossip" prefix, or have a duplicated "gossip_gossip" prefix. This PR fixes this, so now all metrics are correctly prefixed with exactly one "gossip"

network/p2p/gossip/system.go
```diff
@@ -0,0 +1,182 @@
...
+	gossipHandler := struct {
+		appRequester
+		appGossiper
+	}{
+		appRequester: validatorOnlyHandler,
+		appGossiper:  handler,
+	}
+
+	client := network.NewClient(c.HandlerID, validatorPeers)
+	const pollSize = 1
```
StephenButtolph, line 154: We no longer make this parametrizable. We should really never increase the poll size. If we want to increase the poll size, we should increase the frequency of polling.

At least until we hit a point where message marshaling becomes significant.

graft/coreth/plugin/evm/atomic/vm/vm.go
```diff
@@ -72,22 +71,25 @@ const (
...
-	Fx                    secp256k1fx.Fx
-	baseCodec             codec.Registry
-	AtomicMempool         *txpool.Mempool
-	AtomicTxGossipMetrics avalanchegossip.Metrics
-	AtomicTxPushGossiper  *avalanchegossip.PushGossiper[*atomic.Tx]
+	SecpCache *secp256k1.RecoverCache
+	Fx        secp256k1fx.Fx
+	baseCodec codec.Registry
+
+	// TODO: Remove Atomic prefix and unexport these fields
```
StephenButtolph, line 86: Leaving this as a TODO to minimize the diff of this PR.

network/p2p/gossip/system.go
```diff
@@ -0,0 +1,183 @@
...
+// SystemConfig provides all the configurations needed to create a gossip
+// system.
+//
+// All fields are optional.
+type SystemConfig struct {
+	Log       logging.Logger
+	Registry  prometheus.Registerer
+	Namespace string
+
+	HandlerID uint64 // Defaults to [p2p.TxGossipHandlerID]
```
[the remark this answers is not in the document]
StephenButtolph: The default value (0) _is_ `p2p.TxGossipHandlerID`.

graft/coreth/plugin/evm/config/constants.go
```diff
@@ -3,21 +3,9 @@
...
-
-	"github.com/ava-labs/avalanchego/utils/units"
-)
-
 const (
 	TxGossipBloomMinTargetElements       = 8 * 1024
 	TxGossipBloomTargetFalsePositiveRate = 0.01
 	TxGossipBloomResetFalsePositiveRate  = 0.05
 	TxGossipBloomChurnMultiplier         = 3
-	PushGossipDiscardedElements          = 16_384
```
alarso16: Are all of these just approximately configured into the other gossip constructor?

graft/subnet-evm/plugin/evm/vm.go
```diff
@@ -853,33 +842,40 @@ func (vm *VM) onNormalOperationsStarted() error {
...
-		Validators:      vm.config.PushGossipNumValidators,
-		Peers:           vm.config.PushGossipNumPeers,
-	}
-	pushRegossipParams := avalanchegossip.BranchingFactor{
-		Validators: vm.config.PushRegossipNumValidators,
-		Peers:      vm.config.PushRegossipNumPeers,
+	systemConfig := avalanchegossip.SystemConfig{
+		Log:           vm.ctx.Log,
+		Registry:      vm.sdkMetrics,
+		Namespace:     "eth_tx_gossip",
```
alarso16: nit: why not keep the constant for this? If there's a reason not to, that's ok too

network/p2p/gossip/system.go
```diff
@@ -0,0 +1,183 @@
...
+		throttledHandler,
+		validatorPeers,
+		c.Log,
+	)
+
+	// Pull requests are filtered by validators and are throttled to prevent
+	// spamming. Push messages are not filtered.
+	type (
+		appRequester interface {
+			AppRequest(context.Context, ids.NodeID, time.Time, []byte) ([]byte, *common.AppError)
```
alarso16, line 139: Weird interface... Is this not defined anywhere else?

network/p2p/gossip/system.go
```diff
@@ -0,0 +1,183 @@
...
+	if c.DiscardedPushCacheSize <= 0 {
+		c.DiscardedPushCacheSize = 16_384
+	}
+	if c.RegossipPeriod <= 0 {
+		c.RegossipPeriod = 30 * time.Second
+	}
+}
+
+// SystemSet is the backend interface required to construct a gossip system.
+type SystemSet[T Gossipable] interface {
```
alarso16, line 80: There's quite a few places (e.g. `GossipEthTxPool`) where you could use this new interface for assertions instead. Up to you whether it's worth it

graft/subnet-evm/plugin/evm/vm.go
StephenButtolph: It's now only used in one place... I didn't see the point in making it a const (as the Namespace made it clear).

network/p2p/gossip/system.go
StephenButtolph: Not that I know of

graft/subnet-evm/network/network.go
```diff
@@ -496,12 +492,9 @@ func (n *network) SendSyncedAppRequest(ctx context.Context, nodeID ids.NodeID, r
 	return waitingHandler.WaitForResult(ctx)
 }
 
-func (n *network) NewClient(protocol uint64) *p2p.Client {
-	return n.sdkNetwork.NewClient(protocol, n.p2pValidators) // p2pValidators implements NodeSampler
-}
-
-func (n *network) AddHandler(protocol uint64, handler p2p.Handler) error {
-	return n.sdkNetwork.AddHandler(protocol, handler)
+// P2pNetwork returns the p2p network
```
JonathanOppenheimer:
```suggestion
// P2PNetwork returns the p2p network
```

doesn't match the coreth comment (was just checking if the functions all matched)

graft/coreth/plugin/evm/atomic/vm/vm.go
JonathanOppenheimer: I feel like given that these lines are already a diff in this PR, you could just do that here

JonathanOppenheimer: No strong feelings though

vms/platformvm/config/network.go
```diff
@@ -65,9 +64,6 @@ type Network struct {
 	// PushGossipFrequency is how frequently rounds of push gossip are
 	// performed.
 	PushGossipFrequency time.Duration `json:"push-gossip-frequency"`
-	// PullGossipPollSize is the number of validators to sample when performing
-	// a round of pull gossip.
-	PullGossipPollSize int `json:"pull-gossip-poll-size"`
```
JonathanOppenheimer, line 70: Should this be documented?

StephenButtolph: Yeah probably. I'll update after dinner

graft/subnet-evm/plugin/evm/vm.go
joshua-kim: I think it's fine to inline the string here - package-level vars/consts that are only used in one place add clutter to the package scope and making them vars/consts does not add a concrete benefit without more consumers of it.

network/p2p/gossip/system.go
```diff
@@ -0,0 +1,183 @@
...
+	RequestPeriod    time.Duration // Defaults to one request per second
+
+	PushGossipParams   BranchingFactor // Defaults to 100 validators and top 90% of stake
+	PushRegossipParams BranchingFactor // Defaults to 10 validators
+
+	DiscardedPushCacheSize int           // Defaults to 16,384
+	RegossipPeriod         time.Duration // Defaults to 30 seconds
+}
+
+func (c *SystemConfig) SetDefaults() {
```
joshua-kim: If we want to have a config with defaults the caller can provide a config and `NewSystem` can internally set default values without this having to leaking into the exported API

It seems like we need this because `gossip.Every` doesn't know about what a config is... but if we're introducing a type to configure how the entire gossip "system" works, then I feel that it is reasonable for `gossip.Every` to accept it as a parameter. If we wanted to avoid exposing `gossip.Every` information that it doesn't need we could even wrap this under a requester config type.

graft/coreth/plugin/evm/atomic/vm/vm.go
```diff
@@ -291,43 +292,12 @@ func (vm *VM) onNormalOperationsStarted() error {
 		return err
 	}
 
-	ctx, cancel := context.WithCancel(context.TODO())
-	vm.cancel = cancel
-
-	atomicTxGossipHandler, err := gossip.NewTxGossipHandler[*atomic.Tx](
```
joshua-kim, line 297: I think we should be able to delete this package now?

network/p2p/gossip/system.go
joshua-kim: We should be more specific than "weird" (pushback based on je ne sais quoi makes this seem like a preference more than a technical concern and sidesteps understanding why we implemented it this way). This pattern is less common but I have grown to be appreciative of defining types in a function scope because it localizes the abstraction and avoids cluttering the package scope.

graft/coreth/plugin/evm/atomic/vm/vm.go
joshua-kim: I think we're trying to scope this to adding the helper to minimize the initialization boilerplate - we should avoid scope creep with unrelated changes

network/p2p/gossip/system.go
alarso16: Ha, by "weird" I mean this isn't the `AppRequest` and `AppResponse` functions used elsewhere in the codebase (specifically missing the requestID), so I was just confused where this came from

joshua-kim: It's missing the request id because `p2p` (used for VM application p2p protocols) abstracts away the request id for us - the request id is needed so that VMs can map a response to a request, but p2p manages unique request ids + the mapping of response -> request for us which is why it's not in this interface. This is in the vein of how to use interfaces where we define the shape of the abstraction that we want to use.

graft/coreth/plugin/evm/atomic/vm/vm.go
StephenButtolph: 🦅 👁️. Removed

network/p2p/gossip/system.go
StephenButtolph: > If we want to have a config with defaults the caller can provide a config and NewSystem can internally set default values without this having to leaking into the exported API

fwiw, `NewSystem` does internally set the default values.

> It seems like we need this because gossip.Every doesn't know about what a config is... but if we're introducing a type to configure how the entire gossip "system" works, then I feel that it is reasonable for gossip.Every to accept it as a parameter. If we wanted to avoid exposing gossip.Every information that it doesn't need we could even wrap this under a requester config type.

I'm going to update `gossip.Every` to override the period with the same default value.

graft/coreth/plugin/evm/atomic/vm/vm.go
JonathanOppenheimer: valid!

PR #4820, opened 2026-01-07
test: add Coreth + Firewood archive UT

RodrigoVillar:
## Why this should be merged

As part of testing Firewood's archival support, we need to test that C-Chain queries against historical states are supported.

Closes https://github.com/ava-labs/firewood/issues/1499.

## How this works

Introduces a state access UT that, given a start ($X$) and end block ($Y$), queries an instance of Coreth by asking for the nonce of the zero address for all blocks $\alpha \in [X, Y]$. This is sufficient for testing historical queries as this succeeds only if Coreth has the matching trie for each $\alpha$.

## How this was tested

CI

No

### DIFF
--- a/graft/coreth/plugin/evm/vm_test.go
+++ b/graft/coreth/plugin/evm/vm_test.go
@@ -2218,36 +2218,25 @@ func TestInspectDatabases(t *testing.T) {
 // Querying for the nonce of the zero address at various heights is sufficient
 // as this succeeds only if the EVM has the matching trie at each height.
 func TestArchivalQueries(t *testing.T) {
-	// Setting the state history to 5 means that we keep around only the 5 latest
-	// tries in memory. By creating numBlocks (10), we'll have:
-	//	- Tries 0-5: on-disk
-	// 	- Tries 6-10: in-memory
-	tests := []struct {
-		name   string
-		scheme string
-	}{
-		{
-			name:   "firewood",
-			scheme: customrawdb.FirewoodScheme,
-		},
-		{
-			name:   "hashdb",
-			scheme: rawdb.HashScheme,
-		},
-	}
+	schemes := []string{customrawdb.FirewoodScheme, rawdb.HashScheme}
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
+	for _, scheme := range schemes {
+		t.Run(scheme, func(t *testing.T) {
 			require := require.New(t)
 			ctx := t.Context()
 
 			vm := newDefaultTestVM()
+
+			// Setting the state history to 5 means that we keep around only the
+			// 5 latest tries in memory. By creating numBlocks (10), we'll have:
+			//	- Tries 0-5: on-disk
+			// 	- Tries 6-10: in-memory
 			vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{
 				ConfigJSON: `{
 					"pruning-enabled": false,
 					"state-history": 5
 				}`,
-				Scheme: tt.scheme,
+				Scheme: scheme,
 			})
 
 			numBlocks := 10


graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,62 @@ func TestInspectDatabases(t *testing.T) {
...
+	handlers, err := vm.CreateHandlers(ctx)
+	require.NoError(err)
+
+	server := httptest.NewServer(handlers[ethRPCEndpoint])
+	t.Cleanup(server.Close)
+
+	client, err := ethclient.Dial(server.URL)
+	require.NoError(err)
+
+	for i := 1; i <= numBlocks+1; i++ {
```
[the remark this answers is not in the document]
RodrigoVillar: Remnant from testing - fixed: 6e36082

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,62 @@ func TestInspectDatabases(t *testing.T) {
...
+// Tests that querying states no longer in memory is still possible when using
+// Firewood in archive mode.
+//
+// Querying for the nonce of the zero address at various heights is sufficient
+// as this succeeds only if the EVM has the matching trie at each height.
+func TestFirewoodArchivalNode(t *testing.T) {
+	require := require.New(t)
+	ctx := t.Context()
+
+	firewoodArchiveConfig := `{
```
JonathanOppenheimer: for the uninformed, why are we creating 10 blocks, when the state history is 5?

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,62 @@ func TestInspectDatabases(t *testing.T) {
...
+	handlers, err := vm.CreateHandlers(ctx)
+	require.NoError(err)
+
+	server := httptest.NewServer(handlers[ethRPCEndpoint])
+	t.Cleanup(server.Close)
+
+	client, err := ethclient.Dial(server.URL)
+	require.NoError(err)
+
+	for i := 1; i <= numBlocks; i++ {
```
JonathanOppenheimer: why does this start at 1 instead of 0?

JonathanOppenheimer: should the genesis block not be tested

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,62 @@ func TestInspectDatabases(t *testing.T) {
 	vm.initializeDBs(db)
 	require.NoError(t, vm.inspectDatabases())
 }
+
+// Tests that querying states no longer in memory is still possible when using
+// Firewood in archive mode.
+//
+// Querying for the nonce of the zero address at various heights is sufficient
+// as this succeeds only if the EVM has the matching trie at each height.
+func TestFirewoodArchivalNode(t *testing.T) {
```
JonathanOppenheimer: I think this could be transitioned to a test table with some more cases beyond the single config you have. 

Some ideas: 

- genesis block (see other comment)
- block number past tip 
- negative block number 
- non-existent addresses
- etc. idk

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: Setting the state history to 5 means that we keep around only the 5 latest tries in memory. By creating 10 blocks, we'll have:

- Tries 1-5: on-disk
- Tries 6-10: in-memory

This is what we want for `TestFirewoodArchivalNode()` as we want to test that querying at block heights 1-5 is possible when the tries are on disk.

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: I chose not to test the genesis block since this isn't "client-generated" (i.e. a block doesn't produce the genesis state). However, I don't see any cons to adding the genesis block since Firewood will have that as well.

RodrigoVillar: Done: 0de9d3e

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: While I'm always open to adding more test coverage, the goal of this PR is to verify that archival queries work as expected when using Firewood. The tests you're suggesting are cases which aren't related to the desired verification (e.g. adding a test that queries with an invalid block number doesn't test Firewood but rather, tests the error handling of the ETH RPC client/server).

graft/coreth/plugin/evm/vm_test.go
JonathanOppenheimer: gotcha makes sense. for those of us less informed (like you and austin haha), could this be commented?

JonathanOppenheimer: or if you think this is so basic, just resolve this comment.

graft/coreth/plugin/evm/vm_test.go
JonathanOppenheimer: totally fair!

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: No worries - added an in-code comment to explain the rationale behind setting `state-history`: f36962b

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,65 @@ func TestInspectDatabases(t *testing.T) {
...
```
alarso16: This test isn't necessarily unique to Firewood. Want to run it with HashDB too?

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,65 @@ func TestInspectDatabases(t *testing.T) {
...
+	firewoodArchiveConfig := `{
+		"state-scheme": "firewood",
+		"snapshot-cache": 0,
+		"pruning-enabled": false,
+		"state-sync-enabled": false,
+		"state-history": 5
+	}`
+
+	vm := newDefaultTestVM()
+	vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{
```
alarso16: What you should be able to do is the following config:
```
func testArchive(t *testing.T, scheme string {
....

	archiveConfig := `{
		"pruning-enabled": false,
		"state-history": 5
	}`
	vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{
		ConfigJSON: archiveConfig,
		Scheme: scheme
	})
```
But you have to remove the `pruning-enabled` field from the config map in `plugin/evm/vmtest/test_vm.OverrideSchemeConfig` (and while you're there, will you remove `metrics-expensive-enabled` too?)

graft/coreth/plugin/evm/vm_test.go
alarso16, line 1: Good test, but I do think you should run this in `subnet-evm` too. There's totally different abstractions to create a test VM, but it's relatively simple still. Let me know if you need help

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: Yup - done: 34ddfec

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: Done: 34ddfec

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: Done: 2fd35da

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2209,3 +2211,79 @@ func TestInspectDatabases(t *testing.T) {
...
+	// tries in memory. By creating numBlocks (10), we'll have:
+	//	- Tries 0-5: on-disk
+	// 	- Tries 6-10: in-memory
+	tests := []struct {
+		name   string
+		scheme string
+	}{
+		{
+			name:   "firewood",
+			scheme: customrawdb.FirewoodScheme,
```
alarso16: The `name` and `scheme` fields are essentially identical. I know that I didn't follow the typical table-driven approach in the rest of the file, but this seems reasonable

alarso16 (approved):

graft/coreth/plugin/evm/vm_test.go
RodrigoVillar: Ah great point - went ahead and unified the two fields: 0dce317

PR #4823, opened 2026-01-08
fix(snapshot): cancel+join snapshot generation to avoid shutdown hangs

powerslider:
## Why this should be merged

This change replaces the snapshot generator **abort handshake** (`genAbort chan chan struct{}`) with a standard **cancel + join** lifecycle (`cancel` + `done`).

It’s needed because the current model has two fundamental shutdown problems:

- Shutdown can hang thus the generator goroutine can end up waiting forever for an abort handshake even after it has finished generating (the "wait for someone to abort me" pattern). This is why we previously had to ignore `(*diskLayer).generate` in `goleak`.
- Shutdown can race thus the generator can continue running during shutdown and still touch the DB while other components are closing it. This matches the class of shutdown-ordering hazards fixed upstream in geth.

### What this improves

- Ensures deterministic shutdown - `Release()`/`AbortGeneration()` now guarantee the generator is stopped before proceeding by doing:
  - **broadcast cancel**: `close(cancel)`
  - **join**: `<-done`
- Avoids goroutine leaks since the generator no longer parks waiting for a follow-up abort signal, so we can remove the `goleak.IgnoreTopFunction(...diskLayer).generate` workaround and actually enforce leak-free shutdown in tests.
- We also avoid channel-handshake edge cases (multiple abort callers, timing windows, nil-channel receives) by using a close-only cancel signal and a single `done` join point.
- `genRunning` is no longer needed for correctness with cancel+join, so it’s removed entirely (fewer moving parts, no extra atomics).

### Why it’s better than the prior approach

The prior approach (mutating `genAbort`, using timeouts) is brittle because nil channels and timing windows can deadlock or allow the generator to keep running during shutdown. `cancel`+`done` is simpler and deterministic:

- **idempotent cancel** (`sync.Once`)
- **explicit join** (`<-done`)
- **no timeout/race dependence** (and `cancel`/`done` only exist when a generator goroutine is actually started

Overall, this makes snapshot generation behave like a well-scoped background worker: it can be cancelled, it will exit, and shutdown can reliably wait for it - matching the intent of the upstream fix and eliminating the leak/hang class of issues we’ve been paging on. You can learn more on [this pattern](https://go.dev/blog/pipelines) on the official Go blog.

## How this works

- Replace `genAbort` handshake with `cancel/done` lifecycle in `coreth` and `subnet-evm`.
- Make `stopGeneration()` idempotent and join the generator on shutdown.
- Avoid dead joins when generation is skipped (skipGenThreshold path).
- Fix `goleak TestMain` usage and update tests to use `stopGeneration()`.
- Remove unused `genRunning/atomic` state.

## How this was tested

patched exisiting UT

no

### DIFF
--- a/graft/coreth/core/state/snapshot/generate_test.go
+++ b/graft/coreth/core/state/snapshot/generate_test.go
@@ -101,7 +101,9 @@ func testGeneration(t *testing.T, scheme string) {
 	checkSnapRoot(t, snap, root)
 
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation with existent flat state.
@@ -140,7 +142,9 @@ func testGenerateExistentState(t *testing.T, scheme string) {
 	checkSnapRoot(t, snap, root)
 
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 func checkSnapRoot(t *testing.T, snap *diskLayer, trieRoot common.Hash) {
@@ -353,7 +357,9 @@ func testGenerateExistentStateWithWrongStorage(t *testing.T, scheme string) {
 	}
 	checkSnapRoot(t, snap, root)
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation with existent flat state, where the flat state
@@ -414,7 +420,9 @@ func testGenerateExistentStateWithWrongAccounts(t *testing.T, scheme string) {
 	checkSnapRoot(t, snap, root)
 
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation errors out correctly in case of a missing trie
@@ -452,7 +460,9 @@ func testGenerateCorruptAccountTrie(t *testing.T, scheme string) {
 		// Not generated fast enough, hopefully blocked inside on missing trie node fail
 	}
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation errors out correctly in case of a missing root
@@ -494,7 +504,9 @@ func testGenerateMissingStorageTrie(t *testing.T, scheme string) {
 		// Not generated fast enough, hopefully blocked inside on missing trie node fail
 	}
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation errors out correctly in case of a missing trie
@@ -534,7 +546,9 @@ func testGenerateCorruptStorageTrie(t *testing.T, scheme string) {
 		// Not generated fast enough, hopefully blocked inside on missing trie node fail
 	}
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests that snapshot generation when an extra account with storage exists in the snap state.
@@ -597,7 +611,9 @@ func testGenerateWithExtraAccounts(t *testing.T, scheme string) {
 	checkSnapRoot(t, snap, root)
 
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 	// If we now inspect the snap db, there should exist no extraneous storage items
 	if data := rawdb.ReadStorageSnapshot(helper.diskdb, hashData([]byte("acc-2")), hashData([]byte("b-key-1"))); data != nil {
 		t.Fatalf("expected slot to be removed, got %v", string(data))
@@ -656,7 +672,9 @@ func testGenerateWithManyExtraAccounts(t *testing.T, scheme string) {
 	}
 	checkSnapRoot(t, snap, root)
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // Tests this case
@@ -702,7 +720,9 @@ func testGenerateWithExtraBeforeAndAfter(t *testing.T, scheme string) {
 	}
 	checkSnapRoot(t, snap, root)
 	// Stop the generator (if still running) and wait for it to exit.
-	snap.stopGeneration()
+	if err := snap.Release(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 // TestGenerateWithMalformedSnapdata tests what happes if we have some junk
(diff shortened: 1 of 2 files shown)


graft/coreth/core/state/snapshot/generate.go
```diff
@@ -78,11 +78,12 @@ func generateSnapshot(diskdb ethdb.KeyValueStore, triedb *triedb.Database, cache
...
 		genMarker:  genMarker,
 		genPending: make(chan struct{}),
-		genAbort:   make(chan chan struct{}),
+		cancel:     make(chan struct{}),
+		done:       make(chan struct{}),
 		created:    time.Now(),
 	}
 	go base.generate(stats)
-	log.Info("[LEAK-DEBUG] Starting snapshot generation", "root", root, "blockHash", blockHash)
+	log.Debug("Start snapshot generation", "root", root)
```
alarso16, line 86: Note: reduces diff from master

graft/coreth/core/state/snapshot/generate.go
alarso16, line 1: This is great, so much cleaner

graft/coreth/core/state/snapshot/generate_test.go
```diff
@@ -977,10 +948,8 @@ func testGenerateBrokenSnapshotWithDanglingStorage(t *testing.T, scheme string)
 	}
 	checkSnapRoot(t, snap, root)
 
-	// Signal abortion to the generator and wait for it to tear down
-	stop := make(chan struct{})
-	snap.genAbort <- stop
-	<-stop
+	// Stop the generator (if still running) and wait for it to exit.
+	snap.stopGeneration()
```
alarso16: Since you're fixing this... any reason not to use the exported function?

graft/coreth/core/state/snapshot/disklayer.go
```diff
@@ -211,11 +210,9 @@ func (dl *diskLayer) Update(blockHash, blockRoot common.Hash, destructs map[comm
 
 // stopGeneration aborts the state snapshot generation if it is currently running.
 func (dl *diskLayer) stopGeneration() {
-	// Check if generation goroutine is actually running.
-	//
-	// Note: genMarker can be nil even when the generator is still running (waiting
-	// for abort signal after completing generation), so we can't rely on genMarker.
-	if !dl.genRunning.Load() {
+	cancel := dl.cancel
```
alarso16, line 213: Because you're only reading concurrently (no goroutine will write to these outside of the `Once`), and you use a `sync.Once`, this is concurrent safe

graft/coreth/core/state/snapshot/disklayer.go
```diff
@@ -211,11 +210,9 @@ func (dl *diskLayer) Update(blockHash, blockRoot common.Hash, destructs map[comm
...
+	done := dl.done
+	if cancel == nil || done == nil {
```
alarso16, line 215: When is this case hit?

graft/coreth/core/state/snapshot/generate.go
JonathanOppenheimer: I just did this to try to find the leak in CI. Tsvetan is too goated though.

graft/coreth/core/state/snapshot/disklayer.go
JonathanOppenheimer: Can you add a comment for this? I was about to comment and ask why we didn't need the lock.

graft/coreth/core/state/snapshot/snapshot.go
```diff
@@ -729,11 +729,10 @@ func diffToDisk(bottom *diffLayer) (*diskLayer, bool, error) {
 	// If snapshot generation hasn't finished yet, port over all the starts and
 	// continue where the previous round left off.
 	//
-	// Note, the `base.genAbort` comparison is not used normally, it's checked
+	// Note, the `base.genPending` comparison is not used normally, it's checked
 	// to allow the tests to play with the marker without triggering this path.
-	if base.genMarker != nil && base.genAbort != nil {
+	if base.genMarker != nil && base.genPending != nil {
```
JonathanOppenheimer, line 734: Should we check the cancellation instead or is gen.Pending still good -- hard to keep these apart for me

graft/subnet-evm/core/state/snapshot/generate.go
```diff
@@ -226,9 +212,7 @@ func (dl *diskLayer) generate(stats *generatorStats) {
 	nodeIt, err := accTrie.NodeIterator(accMarker)
 	if err != nil {
 		log.Error("Generator failed to create account iterator", "root", dl)
-		abort := <-dl.genAbort
 		dl.genStats = stats
-		close(abort)
 		return
 	}
 	accIt := trie.NewIterator(nodeIt)
```
JonathanOppenheimer, line 218: I can't comment as far down, but in the loop, how would it know if it gets canceled? Or is it intended to just keep running what it's doing, (account stuff), and then cancel?

graft/coreth/core/state/snapshot/generate.go
```diff
@@ -212,9 +201,7 @@ func (dl *diskLayer) generate(stats *generatorStats) {
 	if err != nil {
 		// The account trie is missing (GC), surf the chain until one becomes available
 		stats.Info("Trie missing, state snapshotting paused", dl.root, dl.genMarker)
```
JonathanOppenheimer, line 203: Same question in places like this -- where we return. Do we have to stop generation explicitly or is it unnecessary?

graft/coreth/core/state/snapshot/generate_test.go
powerslider: Yeah, you have a point. I just did not notice it immediately. Will fix this.

graft/coreth/core/state/snapshot/disklayer.go
powerslider: Not exactly true. `sync.Once` only makes the `close(cancel)` operation idempotent/concurrency-safe. It does not by itself make unsynchronized reads of `dl.cancel/dl.done` safe if those fields could be written concurrently.

graft/coreth/core/state/snapshot/disklayer.go
powerslider: This is hit when there’s no generator goroutine to cancel/join, i.e. the disk layer was created with `cancel/done` unset (snapshot already complete, generation disabled or skipped).

graft/coreth/core/state/snapshot/snapshot.go
powerslider: `genPending` is still the right guard because it tells us generation was enabled/started for this layer. `cancel/done` only indicate a running generator lifecycle, and we already use `base.genMarker != nil` to mean "still generating".

graft/subnet-evm/core/state/snapshot/generate.go
powerslider: The inner account/storage loops call `dl.checkAndFlush(...)` periodically, and `checkAndFlush` checks `dl.cancel`. When it’s closed it flushes progress and returns true, causing the generator to return. So yes, it may keep working until the next `checkAndFlush` call, rather than aborting immediately mid-iteration. If we want faster responsiveness, we can add a cheap `select { case <-dl.cancel: ...; default: }` inside the loops, but I am a bit torn on that if we should do it.

powerslider: I am not so sure what the intended behaviour should exactly be in this case and given that we did not observe any stopping in the original code maybe its better to not add it here as well.

graft/coreth/core/state/snapshot/generate.go
powerslider: No explicit stop is needed there. If we hit an error path like this and return, the generator goroutine is exiting anyway, and because we `defer close(dl.done)` at the top, `stopGeneration()` will still unblock (or is a no-op if `cancel/done` weren’t set). Cancellation is only needed to request exit and here we’re already exiting due to an error.

graft/coreth/core/state/snapshot/generate_test.go
powerslider: check https://github.com/ava-labs/avalanchego/pull/4823/changes/f69082887bbd4c58a5b87d4a0398afd15e23e971

graft/coreth/core/state/snapshot/disklayer.go
JonathanOppenheimer: If they're not safe, should they be done under a lock?

graft/subnet-evm/core/state/snapshot/generate.go
JonathanOppenheimer: I'll leave it to your discretion then!

graft/coreth/core/state/snapshot/disklayer.go
powerslider: No, `dl.cancel/dl.done` are initialized before the layer is published and never reassigned, so reads here are race-free (`sync.Once` only guards `close(cancel)`).

PR #4824, opened 2026-01-08
Remove `ProposerMinBlockDelay` from subnet configs

StephenButtolph:
## Why this should be merged

Since the activation of Granite, all EVM chains have upgraded to reliably enforce their own minimum block times.

This PR removes the proposervm delay on all subnet/L1 chains and the C-chain. The X-chain and the P-chain still have the same default minimum delay, which is able to be configured by the `--proposervm-min-block-delay` flag.

## How this works

If the X-chain or the P-chain is being created, we provide the flag's value for the minimum delay. Otherwise, the minimum delay is set to 0.

## How this was tested

- [x] Locally ensured the P-chain still enforces 1s proposervm delays by default
- [x] Locally ensured the X-chain still enforces 1s proposervm delays by default
- [x] Locally ensured the C-chain no longer enforces proposervm delays

Yes

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -5,6 +5,7 @@
 ### Config
 
 - Removed `pull-gossip-poll-size` from the X-chain and P-chain configs.
+- Removed `proposerMinBlockDelay` from subnet configs.
 
 ### Fixes
 
--- a/subnets/config.md
+++ b/subnets/config.md
@@ -75,15 +75,6 @@ same values used for the Primary Network, which are given [CLI Snow Parameters](
 | --snow-avalanche-batch-size      | `batchSize`           |
 | --snow-avalanche-num-parents     | `parentSize`          |
 
-#### `proposerMinBlockDelay` (duration)
-
-The minimum delay performed when building snowman++ blocks. Default is set to 1 second.
-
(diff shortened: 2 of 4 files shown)


config/node/config.go
```diff
@@ -177,7 +177,14 @@ type Config struct {
...
 	TrackedSubnets set.Set[ids.ID] `json:"trackedSubnets"`
 
-	SubnetConfigs map[ids.ID]subnets.Config `json:"subnetConfigs"`
+	// ProposerMinBlockDelay is the minimum delay this node will enforce when
+	// building a snowman++ block on the P-chain and the X-chain. All other
+	// chains are expected to perform their own block production throttling.
+	//
+	// TODO: Remove this flag once the P-chain and X-chain throttle their own
+	// block production.
+	ProposerMinBlockDelay time.Duration             `json:"proposerMinBlockDelay"`
```
michaelkaplan13, line 186: [No action needed] Part of me wonders if we should simplify this by always using the default 1 second minimum builder delay for the P- and X-Chains, but keeping it configurable is the minimal required change for the PR so I'm good with this.

michaelkaplan13 (approved):

PR #4825, opened 2026-01-08
test: fix `TestAtomicSyncerVM/StateSyncFromScratchTest` flake

### LINKED ISSUE #4554: Fix `TestAtomicSyncerVM/StateSyncFromScratchTest` flake
See https://github.com/ava-labs/coreth/actions/runs/18850271144/job/53784830890

JonathanOppenheimer:
## Why this should be merged

This closes #4554. 

## How this works 

After `FillAccountsWithOverlappingStorage` creates tries, they were not explicitly committed to the database before the syncer tried to sync from the database, which could cause a race condition. The addition of this line matches 

https://github.com/ava-labs/avalanchego/blob/ec05f574dd61f8fbfbd53064d81f8a3a46b866ec/graft/coreth/plugin/evm/atomic/vm/syncervm_test.go#L95-L100

I assume that manually setting the commit interval was supposed to avoid an issue like this, but the call to `SetLastAcceptedBlock` is not persisted without this? 

## How this was tested
Ran the test repeatedly 

No


alarso16 (commented):
I could never get this to reproduce locally, could you?

Also, the tries created don't have to be committed to the database. They are cached in memory for use, and this is not the correct command to commit them to the database anyway

JonathanOppenheimer:
> I could never get this to reproduce locally, could you?
> 
> Also, the tries created don't have to be committed to the database. They are cached in memory for use, and this is not the correct command to commit them to the database anyway

Nah I never could -- I ran it a bunch and it passed lol. What is the correct command to commit them to the database? And if so, does that mean the line is unnecessary in the other test?

graft/coreth/plugin/evm/vmtest/test_syncervm.go
```diff
@@ -339,6 +339,7 @@ func initSyncServerAndClientVMs(t *testing.T, test SyncTestParams, numBlocks int
 	internalBlock, ok := internalWrappedBlock.(*chain.BlockWrapper)
 	require.True(ok)
 	require.NoError(serverVM.SetLastAcceptedBlock(internalBlock.Block))
+	require.NoError(serverVM.VersionDB().Commit())
```
ceyonur, line 343: this looks like it's not directly related with the bug since we don't do any database operation  in`SetLastAcceptedBlock` (it's all cache and memory) that requires to be committed.

So I wonder if this correctly resolves the flake. (and if it effectively resolved then I think it's because we should call it right after `patchBlock`

JonathanOppenheimer: As I can't reproduce the flake locally, I can't have complete confidence it does. Do you think this is reasonable, or not even close? I can move the line to after patch block.

ceyonur: It's not completely unreasonable as we do bunch of other db stuff in `patchBlock`. However it's not related with `SetLastAcceptedBlock`. I think we should move the `Commit` line to either inside of `patchBlock` or right after it and see if it works.

JonathanOppenheimer: Actually, that doesn't do anything. `patchBlock` uses `rawdb.WriteBlock` and `rawdb.WriteCanonicalHash` which write to `chaindb`. These writes bypass the` versiondb` entirely and go directly to the underlying database. These changes should be fine anwyay. I still think the actual flake is that last call is still not persisted properly -- here is how I think you do it properly: 

1. `SetLastAcceptedBlock` updates the in-memory cache
2. `PutLastAcceptedID` writes the block ID
3. `VersionDB().Commit()` persists the change to the underlying database

JonathanOppenheimer: thoughts?

JonathanOppenheimer: to be more explicit, when patching the server's last accepted block in only the in-memory state was updated (`SetLastAcceptedBlock`), but the database wasn't. The `acceptedBlockDB` could still contain the OLD block's ID

graft/coreth/plugin/evm/extension/config.go
```diff
@@ -48,6 +48,8 @@ type ExtensibleVM interface {
 	AddHandler(protocol uint64, handler p2p.Handler) error
 	// SetLastAcceptedBlock sets the last accepted block
 	SetLastAcceptedBlock(lastAcceptedBlock snowman.Block) error
+	// PutLastAcceptedID persists the last accepted block ID to the database
```
JonathanOppenheimer, line 47: If this isn't acceptable, I vote we just close this PR, and the issue and see if it ever happens again.

graft/coreth/plugin/evm/vmtest/test_syncervm.go
ceyonur: it makes more sense now with `PutLastAcceptedID`, but maybe again we need a broader call like `Accept` or `AcceptSync` (where `PutLastAcceptedID` is called).

JonathanOppenheimer: By a broader call, do you mean a combination function like this?

ceyonur: No I meant these [Accept](https://github.com/ava-labs/avalanchego/blob/ceyonur/relayer-mode/graft/coreth/plugin/evm/wrapped_block.go#L94), [AcceptSync](https://github.com/ava-labs/avalanchego/blob/ceyonur/relayer-mode/graft/coreth/plugin/evm/vmsync/client.go#L272)

I think AcceptSync is a better fit in this context

JonathanOppenheimer: We can not use either of those functions here:

- accept we can't use bc we're using a fake root
- acceptsync is internal to the sync client and takes a message.Syncable, not a block/ID directly

ceyonur (approved):

PR #4826, opened 2026-01-08
test: fix `TestUptimeSignatures` flake

### LINKED ISSUE #4706: Fix `TestUptimeSignatures` flake
[Failing run](https://github.com/ava-labs/subnet-evm/actions/runs/19434219922/job/55600699033?pr=1832)

Whoever investigates this should also ensure it doesn't occur in coreth

JonathanOppenheimer:
## Why this should be merged

This closes #4706. 

## How this works 

The test flakes were off by a single second: 

```
expected: "2: current uptime 0 is less than queried uptime 80 for validationID 2MiEo2BzNcA4muUMhYeEjsbzbo16eHF5zFknxQbzJqf1LurbD4"
actual  : "2: current uptime 1 is less than queried uptime 80 for validationID 2MiEo2BzNcA4muUMhYeEjsbzbo16eHF5zFknxQbzJqf1LurbD4"
```

The root cause of this issue is when `startTime` was captured using `time.Now().Unix()` for the `validatorState`, work is done and time elapses between the creation of the mockable clock used in tests. That clock can then be ever so slightly ahead of `startTime`, depending on the execution time, as it defaults to `time.Now`. This can lead to test failures if things are exceptionally slow (on the order of one second) between the test really starting. To fix this, I set the mockable clock the same originally captured start time. 

## How this was tested
Ran the test repeatedly 

No

### DIFF
--- a/graft/subnet-evm/warp/verifier_backend_test.go
+++ b/graft/subnet-evm/warp/verifier_backend_test.go
@@ -259,7 +259,8 @@ func TestUptimeSignatures(t *testing.T) {
 
 	validationID := ids.GenerateTestID()
 	nodeID := ids.GenerateTestNodeID()
-	startTime := uint64(time.Now().Unix())
+	start := time.Now()
+	startTime := uint64(start.Unix())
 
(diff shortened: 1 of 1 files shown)


alarso16 (approved):
This seems like a reasonable fix

powerslider (approved):

PR #4832, opened 2026-01-09
feat: Add interface between x/sync and EVM syncers

alarso16:
## Why this should be merged

Satisfies the syncer interface to be used in the EVM, providing a hook point for the code queue.

## How this works

Implements a new Firewood database type for x/sync, and wraps this with a spot to finalize the code queue.

## How this was tested

Fills out the meat of the unit test, but can't test the code queue functionality until ava-labs/firewood#1597 is merged in and released.

No

### DIFF
--- a/graft/evm/sync/synctest/test_trie.go
+++ b/graft/evm/sync/synctest/test_trie.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/crypto"
 	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/libevm/stateconf"
 	"github.com/ava-labs/libevm/trie"
 	"github.com/ava-labs/libevm/trie/trienode"
 	"github.com/ava-labs/libevm/triedb"
@@ -201,7 +202,8 @@ func FillAccounts(
 	newRoot, nodes, err := tr.Commit(true)
 	require.NoError(t, err)
 	require.NoError(t, mergedSet.Merge(nodes))
-	require.NoError(t, s.TrieDB().Update(newRoot, root, 0, mergedSet, nil))
+	updateOpts := stateconf.WithTrieDBUpdatePayload(common.Hash{}, common.Hash{}) // block hashes required for Firewood
+	require.NoError(t, s.TrieDB().Update(newRoot, root, 0, mergedSet, nil, updateOpts))
 	require.NoError(t, s.TrieDB().Commit(newRoot, false))
 	return newRoot, accounts
 }


graft/coreth/sync/statesync/firewood_syncer.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (f *firewoodSyncer) Sync(ctx context.Context) error {
+	if err := f.s.Start(ctx); err != nil {
+		return fmt.Errorf("starting syncer: %w", err)
+	}
+
+	if err := f.s.Wait(ctx); err != nil {
+		return fmt.Errorf("waiting for syncer: %w", err)
+	}
+	return f.codeQueue.Finalize()
```
powerslider: I would handle the error here for better tracablity.
```suggestion
	if err := f.codeQueue.Finalize(); err != nil {
		return fmt.Errorf("finalizing code queue: %w", err)
	}
	return nil
```

graft/coreth/sync/statesync/firewood_syncer.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (*firewoodSyncer) ID() string {
+	return "firewood_syncer"
```
powerslider: We use a key based convention here. The value should be `state_firewood_sync`.

graft/coreth/sync/statesync/firewood_syncer.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (*firewoodSyncer) Name() string {
+	return "Firewood Syncer"
```
powerslider: Also to be consistent with the other syncers - this should be `Firewood EVM State Syncer`

graft/coreth/sync/evmstate/firewood_syncer_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	eg, egCtx := errgroup.WithContext(t.Context())
+	eg.Go(func() error { return codeSyncer.Sync(egCtx) })
+	eg.Go(func() error { return firewoodSyncer.Sync(egCtx) })
+
+	require.NoError(t, eg.Wait(), "failure during sync")
+}
+
+func createDB(t *testing.T) state.Database {
+	diskdb := rawdb.NewMemoryDatabase()
```
powerslider, line 181: Add `t.Helper()` here.

graft/coreth/sync/evmstate/firewood_syncer_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func testFirewoodSync(t *testing.T, clientState, serverState state.Database, root common.Hash) {
+	// Create the mock P2P client that serves range proofs and change proofs from the server DB.
+	var (
+		codeRequestHandler = handlers.NewCodeRequestHandler(serverState.DiskDB(), message.Codec, handlerstats.NewNoopHandlerStats())
+		mockClient         = statesyncclient.NewTestClient(message.Codec, nil, codeRequestHandler, nil)
+		serverDB           = dbFromState(t, serverState)
+		rHandler           = p2ptest.NewSelfClient(t, t.Context(), ids.EmptyNodeID, syncer.NewGetRangeProofHandler(serverDB))
+		cHandler           = p2ptest.NewSelfClient(t, t.Context(), ids.EmptyNodeID, syncer.NewGetChangeProofHandler(serverDB))
```
powerslider, line 148: You shouldn't use a `var` block when you are just assigning variables with non-zero values. Bad practice - check https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations
```suggestion
	codeRequestHandler := handlers.NewCodeRequestHandler(serverState.DiskDB(), message.Codec, handlerstats.NewNoopHandlerStats())
	mockClient := statesyncclient.NewTestClient(message.Codec, nil, codeRequestHandler, nil)
	serverDB := dbFromState(t, serverState)
	rHandler := p2ptest.NewSelfClient(t, t.Context(), ids.EmptyNodeID, syncer.NewGetRangeProofHandler(serverDB))
	cHandler := p2ptest.NewSelfClient(t, t.Context(), ids.EmptyNodeID, syncer.NewGetChangeProofHandler(serverDB))
```

graft/coreth/sync/evmstate/firewood_syncer_test.go
```diff
@@ -0,0 +1,126 @@
...
+		rHandler,
+		cHandler,
+	)
+	require.NoError(t, err, "NewFirewoodSyncer()")
+
+	eg, egCtx := errgroup.WithContext(t.Context())
+	eg.Go(func() error { return codeSyncer.Sync(egCtx) })
+	eg.Go(func() error { return firewoodSyncer.Sync(egCtx) })
+
+	require.NoError(t, eg.Wait(), "failure during sync")
```
powerslider: I would add the following important checks to additionally assert:
- Checking if the firewood root matches the target root.
- Sanity check for the code queue markers.
```
	// Assert Firewood root matches target root.
	clientFW := dbFromState(t, clientState)
	gotRoot, err := clientFW.Root()
	require.NoError(t, err)
	require.Equal(t, ffi.Hash(root), gotRoot)

	// Assert code queue markers are drained (code syncer made progress).
	it := customrawdb.NewCodeToFetchIterator(clientState.DiskDB().(ethdb.Database))
	defer it.Release()
	require.False(t, it.Next(), "expected no remaining code-to-fetch markers after successful sync")
	require.NoError(t, it.Error())
```

firewood/syncer/evm_syncer.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ava-labs/libevm/common"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/utils/maybe"
+
+	xsync "github.com/ava-labs/avalanchego/x/sync"
+)
+
+type CodeQueue interface {
```
powerslider, line 22: Nice usage of port interfaces here ;). We will definitely need this abstraction here, because the code queue for the dynamic sync version will be enhanced and might be a different implementation than the one we currently use so we shouldn't couple it.

firewood/syncer/evm_syncer.go
```diff
@@ -0,0 +1,64 @@
...
+	AddCode(context.Context, []common.Hash) error
+}
+
+type evmDB struct {
+	*database
```
powerslider: I would prefer composition here:
```suggestion
    db *database
```

graft/coreth/sync/evmstate/firewood_syncer_test.go
maru-ava: I respectfully disagree. avalanchego uses this pattern of declaration pervasively and I don't think there's a good reason to violate that consistency.

firewood/syncer/evm_syncer.go
alarso16: Then I would have to re-implement all the database methods as:
```go
func (db *evmDB) FuncName(...) {
     return db.db.FuncName(...)
 }
 ```
 
 Is this actually an improvement? I know that normally embedding structs is discouraged, but it seems to make a lot of sense here

File: firewood/syncer/evm_syncer.go (hunk 7 above)
Comment (alarso16) [reply firewood/syncer/evm_syncer.go:22]: Thanks! Should I assert that our codequeue implements this interface?

File: graft/coreth/sync/evmstate/firewood_syncer_test.go (hunk 5 above)
Comment (alarso16) [reply graft/coreth/sync/evmstate/firewood_syncer_test.go:148]: I think it's cleaner with the `var`, and with the disagreement, I'll leave it.

File: firewood/syncer/evm_syncer.go (hunk 8 above)
Comment (alarso16) [reply firewood/syncer/evm_syncer.go]: I added as the most recent commit this refactor. If it's not desirable, it can be dropped

File: graft/coreth/sync/evmstate/firewood_syncer_test.go (hunk 5 above)
Comment (powerslider) [reply graft/coreth/sync/evmstate/firewood_syncer_test.go:148]: If it has been already adopted on a team level I can agree (did not know that it was the case).

File: firewood/syncer/evm_syncer.go (hunk 7 above)
Comment (powerslider) [reply firewood/syncer/evm_syncer.go:22]: no, because the idea is for this interface to be local and to describe the contract that is accepted in the current implementation. If you want to reuse it due to usage in many places, then we have to do something like the `sync/types.go` file (which is usually reserved only for those cases of having many implementations). In Go locality is preferred.

File: graft/evm/sync/synctest/test_trie.go [hunk 9]
```diff
@@ -187,7 +187,7 @@ func FillAccounts(
 			acc = onAccount(t, i, key.Address, acc, storageTr)
 			root, nodes, err := storageTr.Commit(false)
 			require.NoError(t, err)
-			if root != types.EmptyRootHash {
+			if root != types.EmptyRootHash && nodes != nil {
```
maru-ava: (No action required) What scenario causes `nodes` to be `nil` when `root != types.EmptyRootHash`? Maybe add a comment to clarify this condition for future maintainers?

firewood/syncer/evm_syncer.go
```diff
@@ -0,0 +1,95 @@
...
+		targetRoot,
+		rangeProofClient,
+		changeProofClient,
+	)
+}
+
+func (e *evmDB) CommitRangeProof(ctx context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
+	nextKey, err := e.db.CommitRangeProof(ctx, start, end, proof)
+	if err != nil {
+		return nextKey, err
```
maru-ava: (No action required) Would it make sense to return `maybe.Nothing[[]byte]()` for consistency?

graft/evm/sync/synctest/test_trie.go
alarso16: I dug into this code a little more (the `trie` type is from go-ethereum), and we should just verify that `nodes` isn't a nil, since it may delete everything. Unlikely in this case, but could cause a bug

firewood/syncer/evm_syncer.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			return maybe.Nothing[[]byte](), err
+		}
+		codeHashes = append(codeHashes, common.Hash(h))
+	}
+	if err := e.codeQueue.AddCode(ctx, codeHashes); err != nil {
+		return maybe.Nothing[[]byte](), err
+	}
+	return nextKey, nil
+}
```
powerslider, line 70: Move `AddCode` ahead of the state commit so a queue failure doesn’t occur after state is already persisted. This makes state+code progression more robust without changing semantics for successful paths. Something like this (just swapping places):
```suggestion
func (e *evmDB) CommitRangeProof(ctx context.Context, start, end maybe.Maybe[[]byte], proof *RangeProof) (maybe.Maybe[[]byte], error) {
	var codeHashes []common.Hash //nolint:prealloc // we don't know how many there will be
	for h, err := range proof.rp.CodeHashes() {
		if err != nil {
			return maybe.Nothing[[]byte](), err
		}
		codeHashes = append(codeHashes, common.Hash(h))
	}
	if err := e.codeQueue.AddCode(ctx, codeHashes); err != nil {
		return maybe.Nothing[[]byte](), err
	}
	nextKey, err := e.db.CommitRangeProof(ctx, start, end, proof)
	if err != nil {
		return maybe.Nothing[[]byte](), err
	}
	return nextKey, nil
}
```

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/ava-labs/avalanchego/graft/coreth/sync/code"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+
+	syncpkg "github.com/ava-labs/avalanchego/graft/coreth/sync/types"
+	xsync "github.com/ava-labs/avalanchego/x/sync"
+)
+
+var _ syncpkg.Syncer = (*firewoodSyncer)(nil)
+
```
powerslider, line 27: You have to implement `types.Finalizer` so registry-driven cleanup closes the code queue on cancellation/errors, matching how other syncers behave in the registry.

```
_ syncpkg.Finalizer = (*firewoodSyncer)(nil)
```

Also you need to make sure finalization is idempotent to avoid a double close situation.

powerslider: Here is a cool trick I learned recently. You can use [sync.OnceValue](https://pkg.go.dev/sync#example-OnceValue). The same as `sync.Once` but returns a value and if the value is the same it does not re-trigger:

- Add a callback called `finalizeOnce` in the struct: https://github.com/ava-labs/avalanchego/pull/4832/changes#r2712856029
- Init on construction: https://github.com/ava-labs/avalanchego/pull/4832/changes#r2712859887
- Return the `Finalize` method on `Sync`: https://github.com/ava-labs/avalanchego/pull/4832/changes#r2712862306

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -0,0 +1,69 @@
...
+type firewoodSyncer struct {
+	s         *xsync.Syncer[*syncer.RangeProof, struct{}]
+	codeQueue *code.Queue
+}
```
powerslider, line 33:
```suggestion 
    // finalizeOnce is initialized in the constructor to make Finalize idempotent.
	finalizeOnce func() error
}
```

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -0,0 +1,69 @@
...
+		rangeProofClient,
+		changeProofClient,
+	)
+	if err != nil {
+		return nil, err
+	}
+	return &firewoodSyncer{
+		s:         s,
+		codeQueue: codeQueue,
+	}, nil
```
powerslider:
```suggestion
	}
	fw.finalizeOnce = sync.OnceValue(func() error {
		// Ensure the syncer stops work and the code queue closes on exit.
		fw.s.Close()
		if err := fw.codeQueue.Finalize(); err != nil {
			return fmt.Errorf("finalizing code queue: %w", err)
		}
		return nil
	})
	return fw, nil
}
```

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if err := f.s.Wait(ctx); err != nil {
+		return fmt.Errorf("waiting for syncer: %w", err)
+	}
+
+	if err := f.codeQueue.Finalize(); err != nil {
+		return fmt.Errorf("finalizing code queue: %w", err)
+	}
+
+	return nil
```
powerslider, line 79:
```suggestion
    return f.Finalize()
```

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (*firewoodSyncer) ID() string {
+	return "state_firewood_sync"
+}
+
+func (*firewoodSyncer) Name() string {
+	return "Firewood EVM State Syncer"
+}
```
powerslider, line 88:
```suggestion
func (f *firewoodSyncer) Finalize() error {
	return f.finalizeOnce()
}
```

graft/coreth/sync/evmstate/firewood_syncer.go
powerslider: `fw` is a var that represents the `firewoodSyncer` instance:
```
fw := &firewoodSyncer{
		s:         s,
		codeQueue: codeQueue,
}
```

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -53,10 +61,19 @@ func (f *firewoodSyncer) Sync(ctx context.Context) error {
...
 
+	return nil
+}
+
+// Finalize implements types.Finalizer.
+func (f *firewoodSyncer) Finalize() error {
+	return f.finalizeOnce()
+}
+
+func (f *firewoodSyncer) finalize() error {
```
powerslider: I would still inline this in the constructor function, because having `finalize` and `Finalize` is a bit confusing, or change the name to not be `finalize`.

alarso16: I renamed it, but I think inlining it adds unnecessary clutter to the constructor

graft/evm/sync/synctest/test_trie.go
alarso16: Fixed now

graft/coreth/sync/evmstate/firewood_syncer.go
alarso16: Discussed offline, but the code syncer will not finish if finalization of the code queue is delated. To be resolved later

powerslider:
Approving again after further feedback and refinements.

maru-ava:
The coreth unit test failures are consistent across the 3 jobs, I'm guessing this PR will need to be rebased before merge will be possible?

alarso16:
> The coreth unit test failures are consistent across the 3 jobs, I'm guessing this PR will need to be rebased before merge will be possible?

Yep! I (correctly) changed an error log in `firewood/triedb.go` to returning an error, so rather than swallowing a potential mistake, it's addressed directly. Easy workaround without having to make block hashes though!

PR #4834, opened 2026-01-09
Return range proofs for change proof requests during Firewood sync

alarso16:
## Why this should be merged

Enables Firewood to be used as a dynamic syncer without change proofs

## How this works

Returns the sentinel error to return a range proof instead of drop the request.

## How this was tested

New UTs

No


firewood/syncer/syncer_test.go
```diff
@@ -99,21 +99,111 @@ func testSync(t *testing.T, seed int64, clientKeys int, serverKeys int) {
...
+		{
+			name:                    "partial sync with replace then update",
+			clientSize:              5_000,
+			serverSize:              10_000, // multiple requests
+			numRequestsBeforeUpdate: 1,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			testSyncWithUpdate(t, 0, tt.clientSize, tt.serverSize, tt.numRequestsBeforeUpdate)
```
rrazvan1: why are we not using the seed parameter?

firewood/syncer/syncer_test.go
```diff
@@ -136,10 +226,12 @@ func generateDB(t *testing.T, numKeys int, seed int64) *ffi.Database {
 		vals = append(vals, val)
 	}
 
-	_, err = db.Update(keys, vals)
+	_, err := db.Update(keys, vals)
 	require.NoError(t, err)
 
-	return db
+	root, err := db.Root()
+	require.NoError(t, err)
```
rrazvan1: instead of getting the root by calling `db.Root()` we can use the first returned value from `db.Update(keys, vals)`

x/merkledb/sync_test.go
```diff
@@ -522,15 +508,10 @@ func Test_Sync_UpdateSyncTarget(t *testing.T) {
 		prometheus.NewRegistry(),
 	)
 	require.NoError(err)
+	synctest.AddFuncOnIntercept(actionHandler, rangeProofHandler, func() {
+		require.NoError(m.UpdateSyncTarget(root2))
```
rrazvan1: If there is an error here, the test is not stopping.

alarso16: Interesting point! Let me fix that

alarso16: I believe that canceling the context in a `defer` is sufficient, because it will run as soon as the test fails

alarso16: I actually don't think so, because we have to wait for the thread to be intercepted. Instead, I explicitly cancel the context on error, report the underlying error, and will eventually fatal error for context canceled

firewood/syncer/syncer_test.go
```diff
@@ -62,58 +65,150 @@ func Test_Firewood_Sync(t *testing.T) {
...
+		root,
+		p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, intercept),
 		p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, NewGetChangeProofHandler(serverDB)),
 	)
 	require.NoError(t, err)
 	require.NotNil(t, syncer)
 
+	synctest.AddFuncOnIntercept(intercept, NewGetRangeProofHandler(serverDB), func() {
+		// Called in separate goroutine, allow graceful cancellation
+		if !assert.NoError(t, syncer.UpdateSyncTarget(newRoot)) {
```
alarso16: This is causing a lint error because "`assert` should not be used with `NoError`", but I think that is acceptable here, since the context will be canceled because of it. However, this is also some repeated logic with merkledb, so I could also find a clear way of handling this, since it is certainly not the most readable. Any opinions?

rrazvan1: What do you think about this approach?

In case there is an error in interceptor, store it and cancel context.
After `syncer.Wait` stops bcs of context cancellation, we check for error `require.NoError(t, interceptErr)`.

I think its cleaner, we get rid of the //nolint, and also with the old approach we would get 2 errors: interceptor error (from assert) + context cancellation error (from syncer.Wait erro).

```
	var interceptErr error
	synctest.AddFuncOnIntercept(intercept, NewGetRangeProofHandler(serverDB), func() {
		interceptErr = syncer.UpdateSyncTarget(newRoot)
		if interceptErr != nil {
			cancel()
		}
	}, numRequestsBeforeUpdate)

	require.NoError(t, syncer.Start(ctx))

	err = syncer.Wait(ctx)
	require.NoError(t, interceptErr)
```

alarso16: That does seem better! I'll go with that

firewood/syncer/syncer_test.go
```diff
@@ -62,58 +65,153 @@ func Test_Firewood_Sync(t *testing.T) {
...
-		return db
+		root, err := db.Root()
+		require.NoError(t, err)
+		return ids.ID(root)
 	}
 
 	var (
-		r         = rand.New(rand.NewSource(seed)) // #nosec G404
 		keys      = make([][]byte, numKeys)
 		vals      = make([][]byte, numKeys)
```
powerslider: This is so so we don’t accidentally send nil entries to `db.Update(...)`
```suggestion
        keys      = make([][]byte, 0, numKeys)
	    vals      = make([][]byte, 0, numKeys)
```

x/merkledb/sync_test.go
```diff
@@ -523,16 +518,18 @@ func Test_Sync_UpdateSyncTarget(t *testing.T) {
...
 
-	// Update sync target on first request
-	once := &sync.Once{}
-	actionHandler.AppRequestF = func(ctx context.Context, nodeID ids.NodeID, deadline time.Time, requestBytes []byte) ([]byte, *common.AppError) {
-		once.Do(func() {
-			require.NoError(m.UpdateSyncTarget(root2))
-		})
-		return rangeProofHandler.AppRequest(ctx, nodeID, deadline, requestBytes)
-	}
+	var interceptErr error
```
powerslider:
```suggestion
    var err error
```

x/merkledb/sync_test.go
```diff
@@ -455,34 +455,29 @@ func Test_Sync_Result_Correct_Root_Update_Root_During(t *testing.T) {
...
 	require.NoError(syncer.Error())
 
-	newRoot, err := db.GetMerkleRoot(t.Context())
+	newRoot, err := db.GetMerkleRoot(ctx)
 	require.NoError(err)
 	require.Equal(secondSyncRoot, newRoot)
 }
 
 func Test_Sync_UpdateSyncTarget(t *testing.T) {
 	require := require.New(t)
```
powerslider, line 475: nit: I know that you have not touched this line and I see that it is used in other tests as well, but just noting it here that we need to probably do another PR to cleanup using `require` in this way. I think this makes sense if we have a custom runner like `require := require.New(ginkgo.GinkgoT())`.

firewood/syncer/syncer_test.go
```diff
@@ -62,58 +65,153 @@ func Test_Firewood_Sync(t *testing.T) {
...
+	serverDB, root := generateDB(t, r, serverKeys)
+	clientDB, _ := generateDB(t, r, clientKeys)
 	defer func() {
 		require.NoError(t, serverDB.Close(ctx))
 		require.NoError(t, clientDB.Close(ctx))
 	}()
 
-	root, err := serverDB.Root()
+	syncer, err := New(
+		Config{},
```
powerslider, line 84: Do we have a constructor function for this `Config`?

x/merkledb/sync_test.go
```diff
@@ -455,34 +455,29 @@ func Test_Sync_Result_Correct_Root_Update_Root_During(t *testing.T) {
...
-		select {
-		case <-updatedRootChan:
-			// do nothing, allow 1 request to go through
-		default:
-			once.Do(func() {
-				require.NoError(syncer.UpdateSyncTarget(secondSyncRoot))
-			})
+	var interceptErr error
+	synctest.AddFuncOnIntercept(actionHandler, xsync.NewGetRangeProofHandler(dbToSync, rangeProofMarshaler), func() {
+		interceptErr = syncer.UpdateSyncTarget(secondSyncRoot)
```
powerslider: Again, rename to `err`, because this is not very conventional.

x/merkledb/sync_test.go
```diff
@@ -437,7 +436,8 @@ func Test_Sync_Result_Correct_Root_Update_Root_During(t *testing.T) {
 
 	actionHandler := &p2p.TestHandler{}
 
-	ctx := t.Context()
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
```
powerslider: Switching from `defer cancel()` to `t.Cleanup(cancel)` makes the intent clearer that this cancellation is part of the test’s lifecycle management. It also behaves better with subtests/early failures because the cleanup is registered on `t` and will always run when the test finishes, which is especially helpful here since sync work and the update trigger can run in other goroutines and we want to reliably unblock `Wait(ctx)` on exit.

```suggestion
    t.Cleanup(cancel)
```

x/merkledb/sync_test.go
alarso16: Maybe create an issue if you think it should be addressed? This is used all over the place

firewood/syncer/syncer_test.go
alarso16: No, an empty config is perfectly acceptable so since we have very little validation to do, this is simpler

x/merkledb/sync_test.go
powerslider: UPDATE: we agreed on changing the name to `errIntercept` and still have it due to the data race situation with `syncer.Wait(ctx)` a bit below.

alarso16: Discussed offline:

>  There are 2 options:
> keep the current name interceptErr , although probably I would personally prefer errIntercept
> or use a buffered error channel (so callback never blocks if the test goroutine is busy). Roughly something like this:
> 
> ```
> interceptErrCh := make(chan error, 1)
> 
> synctest.AddFuncOnIntercept(intercept, NewGetRangeProofHandler(serverDB), func() {
>     e := syncer.UpdateSyncTarget(newRoot)
>     if e != nil {
>         interceptErrCh <- e
>         cancel()
>     } else {
>         interceptErrCh <- nil
>     }
> }, numRequestsBeforeUpdate)
> 
> require.NoError(t, syncer.Start(ctx))
> err := syncer.Wait(ctx)
> 
> require.NoError(t, <-interceptErrCh)
> require.NoError(t, err)
> ```
> the advantage of the channel approach is full guarantee to be race free, because it is a real synchronization point.

powerslider (approved):

firewood/syncer/syncer_test.go
```diff
@@ -62,60 +65,155 @@ func Test_Firewood_Sync(t *testing.T) {
...
+	tests := []struct {
+		name                    string
+		clientSize              int
+		serverSize              int
+		numRequestsBeforeUpdate int
+	}{
+		{
+			name:                    "finish with stale root",
+			clientSize:              0,
+			serverSize:              1000, // one request
```
joshua-kim: I don't understand these comments why is 1k one request and 10k is multiple?

firewood/syncer/syncer_test.go
```diff
@@ -62,60 +65,155 @@ func Test_Firewood_Sync(t *testing.T) {
...
+	require.NoError(t, syncer.Start(ctx))
+	err = syncer.Wait(ctx)
+	if errors.Is(err, sync.ErrFinishedWithUnexpectedRoot) {
+		t.Log("syncer reported root mismatch; logging diff between DBs")
+		logDiff(t, serverDB, clientDB)
+	}
+	require.NoError(t, err)
+}
+
+func Test_Firewood_Sync_WithUpdate(t *testing.T) {
```
joshua-kim: nit: use UpdateSyncTarget naming (avoid calling the same concept different things)

x/merkledb/sync_test.go
```diff
@@ -437,7 +436,8 @@ func Test_Sync_Result_Correct_Root_Update_Root_During(t *testing.T) {
 
 	actionHandler := &p2p.TestHandler{}
 
-	ctx := t.Context()
+	ctx, cancel := context.WithCancel(t.Context())
+	t.Cleanup(cancel)
```
joshua-kim: I think we can just use `defer` here - this context doesn't outlive the test/is in the same goroutine so it's guaranteed to be  called

x/sync/synctest/handlers.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+	"sync/atomic"
+	"time"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
+)
+
+func AddFuncOnIntercept(interceptor *p2p.TestHandler, innerHandler p2p.Handler, fn func(), numToAllow int) {
```
joshua-kim: 1. Instead of modifying the caller handler could we wrap it? This can avoid misuse of the api where callers might chain adds together (general guideline is that mutable apis are prone to misuse).
2. Since we added this to support test-defined behavior on the n'th request, maybe the function could expose the counter to the test to give a more expressive api. Naming could also be improved... something like `NewCounterHandler` if we went w/ this approach?
3. Wondering if this needs a lock or if we expect the implementation of `fn` to handle it - `fn` has potential to be racey here.

firewood/syncer/syncer_test.go
```diff
@@ -62,60 +65,155 @@ func Test_Firewood_Sync(t *testing.T) {
...
+		errIntercept = syncer.UpdateSyncTarget(newRoot)
+		if errIntercept != nil {
+			cancel()
+		}
+	}, numRequestsBeforeUpdate)
+
 	require.NoError(t, syncer.Start(ctx))
+
 	err = syncer.Wait(ctx)
+	require.NoError(t, errIntercept)
```
joshua-kim: Maybe you're looking for something like `context.WithCancelCause` here? We can cancel w/ an error as the reason for cancellation and it should propagate as the error returned from `Wait` - so we can avoid leaking `errIntercept` into the test body.

firewood/syncer/syncer_test.go
```diff
@@ -62,60 +65,155 @@ func Test_Firewood_Sync(t *testing.T) {
...
 	if errors.Is(err, sync.ErrFinishedWithUnexpectedRoot) {
 		t.Log("syncer reported root mismatch; logging diff between DBs")
 		logDiff(t, serverDB, clientDB)
 	}
 	require.NoError(t, err)
+
+	finalRoot, rootErr := clientDB.Root()
+	require.NoError(t, rootErr)
+	require.Equal(t, newRoot, ids.ID(finalRoot))
```
joshua-kim: nit: why not `gotRoot` and `err`?

firewood/syncer/syncer_test.go
```diff
@@ -62,60 +65,155 @@ func Test_Firewood_Sync(t *testing.T) {
...
+func testSyncWithUpdate(t *testing.T, clientKeys int, serverKeys int, numRequestsBeforeUpdate int) {
+	r := rand.New(rand.NewSource(1))
 
+	serverDB, root := generateDB(t, r, serverKeys)
+	clientDB, _ := generateDB(t, r, clientKeys)
+	defer func() {
+		require.NoError(t, serverDB.Close(t.Context()))
+		require.NoError(t, clientDB.Close(t.Context()))
+	}()
+	newRoot := fillDB(t, r, serverDB, serverKeys)
```
joshua-kim: nit: `wantRoot`?

firewood/syncer/syncer_test.go
alarso16: The xsync default request size is 2048. How could I make this clearer? Should I reference this constant directly?

x/merkledb/sync_test.go
alarso16: I originally had `defer`, but a previous comment suggested I change it. Both clearly work just fine, but we should find some team-wide convention

firewood/syncer/syncer_test.go
alarso16: Neat! hadn't seen this, but it feels cleaner

x/sync/synctest/handlers.go
alarso16: 1. Done, that does seem better
2. Done, more go-ey for sure
3. I don't think so. Without more overhead, I don't think it would be reasonable to guarantee that `fn` is thread-safe. In future changes (e.g. `index int` becomes `indices []int`), that might be necessary, but not yet

firewood/syncer/syncer_test.go
```diff
@@ -62,29 +64,106 @@ func Test_Firewood_Sync(t *testing.T) {
...
-	root, err := serverDB.Root()
+	syncer, err := New(
+		Config{},
+		clientDB,
+		root,
+		p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, NewGetRangeProofHandler(serverDB)),
+		p2ptest.NewSelfClient(t, ctx, ids.EmptyNodeID, NewGetChangeProofHandler(serverDB)),
+	)
 	require.NoError(t, err)
+	require.NotNil(t, syncer)
```
joshua-kim: nit: these assertions don't add any meaningful checks into the code (nil will panic anyways).

firewood/syncer/syncer_test.go
joshua-kim: Yeah - I would just reference something like `MaxKeyValuesLimit` or `2*MaxKeyValuesLimit`. One nice side-effect is that we maintain coverage even if we change the limits.

alarso16: I used the default request size (since even if the max changes, without adjusting the test, we will still be using the default)

PR #4839, opened 2026-01-11
feat: add Sync to database.HeightIndex

DracoLi:
## Why this should be merged

Allows database users to `Sync` after a batch of HeightIndex db writes instead of on every height write.

## How this works

Add a Sync method to `database.HeightIndex` so callers can manually Sync a range of heights if needed.

```golang
// Sync flushes underlying writes from the OS buffer cache to disk for
// data in the range [start, end].
Sync(start, end uint64) error
```

A start and end height is needed since the data can be stored across multiple files.

## How this was tested

Unit tests

No

### DIFF
--- a/database/heightindexdb/dbtest/dbtest.go
+++ b/database/heightindexdb/dbtest/dbtest.go
@@ -243,17 +243,17 @@ func TestSync(t *testing.T, newDB func() database.HeightIndex) {
 		{
 			name:    "range with gaps",
 			heights: []uint64{1, 3, 5},
-			start:   0,
-			end:     10,
+			start:   1,
+			end:     5,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			db := newDB()
-			defer func() {
+			t.Cleanup(func() {
 				require.NoError(t, db.Close())
-			}()
+			})
 
 			for _, h := range tt.heights {
 				require.NoError(t, db.Put(h, []byte("data")))
--- a/x/blockdb/database_test.go
+++ b/x/blockdb/database_test.go
@@ -37,6 +37,38 @@ func TestInterface(t *testing.T) {
 	}
 }
 
+func TestSyncPersistence(t *testing.T) {
+	config := DefaultConfig().WithMaxDataFileSize(250).WithSyncToDisk(false)
+	db := newDatabase(t, config)
+	db.compressor = compression.NewNoCompressor()
+	tempDir := db.config.DataDir
+
+	// Write 4 blocks spanning 2 data files
+	blocks := make(map[uint64][]byte)
+	for h := range uint64(4) {
+		blocks[h] = fixedSizeBlock(t, 100, h)
+		require.NoError(t, db.Put(h, blocks[h]))
+	}
+
+	require.NoError(t, db.Sync(0, 3))
+	require.NoError(t, db.Close())
+
+	// Reopen and verify all data persisted across files
+	hdb, err := New(config.WithDir(tempDir).WithBlockCacheSize(0), logging.NoLog{})
+	require.NoError(t, err)
+	reopenedDB := hdb.(*Database)
+	reopenedDB.compressor = compression.NewNoCompressor()
+	t.Cleanup(func() {
+		require.NoError(t, reopenedDB.Close())
+	})
+
+	for h, expected := range blocks {
+		data, err := reopenedDB.Get(h)
+		require.NoError(t, err)
+		require.Equal(t, expected, data)
+	}
+}
+
 func TestNew_Params(t *testing.T) {
 	tempDir := t.TempDir()
 	tests := []struct {


x/blockdb/database.go
```diff
@@ -518,6 +519,52 @@ func (s *Database) hasWithoutLock(height BlockHeight) (bool, error) {
...
+
+	if s.closed {
+		s.log.Error("Failed Sync: database closed",
+			zap.Uint64("start", start),
+			zap.Uint64("end", end),
+		)
+		return database.ErrClosed
+	}
+
+	if err := s.indexFile.Sync(); err != nil {
```
yacovm: why do we need to sync the index too? If we only sync the block and we crash, when we recover we will fix the index won't we?

x/blockdb/database.go
```diff
@@ -518,6 +519,52 @@ func (s *Database) hasWithoutLock(height BlockHeight) (bool, error) {
...
+		return fmt.Errorf("failed to sync index file: %w", err)
+	}
+
+	var synced set.Set[int]
+	for h := start; h <= end; h++ {
```
yacovm: This loop works, but it seems to be very exhaustive isn't it? 

We'll Sync() once in every 1000 blocks or so. Is it possible to do something more efficient here?

When we write blocks during normal operations, we never write them out of order and always write them in-order. 

So can we perhaps make an optimization where we check the first and the last index and if they're the same file then we just sync that file, and assume all files are written also in that file?

yacovm: I know that it's a very Simplex-specific assumption though. But currently it's the only expected consumer of this API 🤷

DracoLi: yeah it feels a bit weird since blockdb does not enforce write order right now but considering that this will be only used for simplex, I think we can make this assumption as long as we document it here. I'm good with adding this optimization here.

x/blockdb/database.go
DracoLi: yep, we don't need the index file sync here. recover will fix index file if we have more data. I'll remove this.

x/blockdb/database.go
DracoLi: if we are assuming writes are in order here, then I think we can just remove the loop? 
We can calculate the data file indexes in between first and last and only sync those. Thoughts?

yacovm: yep

DracoLi: updated to assume data is written in order. https://github.com/ava-labs/avalanchego/pull/4839/commits/4f0016862283d5b626e9ac453c464bc5a5701afb

x/blockdb/database.go
```diff
@@ -533,33 +533,29 @@ func (s *Database) Sync(start, end uint64) error {
...
+	_, firstIdx, err := s.getDataFileForHeight(start)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			return nil
+		}
+		return err
 	}
-
-	var synced set.Set[int]
-	for h := start; h <= end; h++ {
```
yacovm: actually can we just wire in some kind of `Option` enum that defines the assumption of out of order and/or in-order and act accordingly? Then we don't need to have this assumption.

x/blockdb/database.go
```diff
@@ -533,33 +533,29 @@ func (s *Database) Sync(start, end uint64) error {
...
-		f, _, idx, err := s.getDataFileAndOffset(entry.Offset)
+	for idx := firstIdx; idx <= lastIdx; idx++ {
+		f, err := s.getOrOpenDataFile(idx)
 		if err != nil {
-			return fmt.Errorf("failed to get data file for height %d: %w", h, err)
-		}
-
-		if synced.Contains(idx) {
-			continue
+			return fmt.Errorf("failed to open data file %d: %w", idx, err)
```
yacovm, line 566: don't we need to do this synced set regardless?

x/blockdb/database.go
DracoLi: we can, but it feels weird to include that in the database.HeightIndex interface for Sync. I think a better solution would be an option on the db config to enforce in-order writes, then we can assume that as well on Sync. But I think that can be a separate feature request.

x/blockdb/database.go
DracoLi: why is it needed? we are looping through all data files in the height ranges. we don't need to keep track of the synced data idx anymore since we cannot revisit a data file.

x/blockdb/database.go
yacovm: but we can have out of order writes and then we cannot have them after we are finished bootstrapping.
Do you suggest to open the DB at startup and then close it and re-open it again? Seems unnecessary.

DracoLi: i was suggesting the in order option can enforce all blocks to be saved in order, including bootstrapped blocks. If this is not easily possible to change on the bootstrapping side, then maybe we can pass in the height at which we enforce in order writes in db config? In any case, I think this would be a future feature that we can add.

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -201,3 +210,62 @@ func TestClose(t *testing.T, newDB func() database.HeightIndex) {
...
+			end:     3,
+		},
+		{
+			name:    "partial range",
+			heights: []uint64{0, 1, 2, 3, 4, 5},
+			start:   2,
+			end:     4,
+		},
+		{
+			name:    "range with gaps",
```
maru-ava, line 244: (No action required) Is it intentional that this test doesn't invoke the sync loop?

```
  Sync(0, 10)
    - getDataFileIndexForHeight(0)
        - readBlockIndex(0)
            - readIndexEntry(0)
                - entry.IsEmpty() = true (height 0 never written)
                - returns ErrNotFound
            - returns ErrNotFound
        - returns ErrNotFound
    - errors.Is(err, ErrNotFound) = true
    - return nil  <-- exits without syncing
```

database/heightindexdb/dbtest/dbtest.go
```diff
@@ -201,3 +210,62 @@ func TestClose(t *testing.T, newDB func() database.HeightIndex) {
...
+				require.NoError(t, db.Close())
+			}()
+
+			for _, h := range tt.heights {
+				require.NoError(t, db.Put(h, []byte("data")))
+			}
+
+			require.NoError(t, db.Sync(tt.start, tt.end))
+
+			for _, h := range tt.heights {
```
maru-ava, line 264: (No action required) Maybe close and reopen the db before validating the contents to check both that sync ran without error and produced the expected result?

x/blockdb/database.go
```diff
@@ -518,6 +518,61 @@ func (s *Database) hasWithoutLock(height BlockHeight) (bool, error) {
...
+		return 0, err
+	}
+	_, _, idx, err := s.getDataFileAndOffset(entry.Offset)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get data file index for height %d: %w", height, err)
+	}
+	return idx, nil
+}
+
+// Sync calls sync on all data files in the range [start, end],
```
maru-ava, line 533: (No action required) What's the recovery mechanism for index files?

x/blockdb/database.go
```diff
@@ -518,6 +518,61 @@ func (s *Database) hasWithoutLock(height BlockHeight) (bool, error) {
...
+// assuming data are written in-order. If no data exists at start or end,
```
maru-ava, line 534: (No action required) What's the rational for not syncing if data doesn't exist at the start or the end? e.g. if a caller writes heights 50-100 and calls Sync(0, 100) using 0 as a safe lower bound, nothing is synced because height 0 does not exist. Would it make sense to instead find the first existing height >= start?

maru-ava (commented):
Just questions

x/blockdb/database.go
yacovm: So, if we call `Sync(i,j)` where `j>i` then it means `i` also exists. 
We will never use `0` because it's the genesis block, there is no reason to sync it.

x/blockdb/database.go
yacovm: We scan the block data files and compare them with the index files. If a block data entry isn't indexed, we just add it to the index.

x/blockdb/database.go
maru-ava: Fair enough that there's no reason to sync the genesis block, but why is it desirable to skip sync if `start` is less than the minimum height?

yacovm: I guess it's just to document that if blockDB doesn't contain a block (it was written by an earlier version of avalanchego) then the operation will be a no-op. 

I'm personally fine with that...

DracoLi: Adding some context here. Sync is added for for simplex's specific usage, in which we should know the range of blocks to sync, and is called after the blocks in the range are written.
We can make this more general where we handle when start or end is missing or when end is larger than max height, but this adds complexity that is not really needed right now IMO.

database/heightindexdb/dbtest/dbtest.go
DracoLi: this is not intentional after the logic changes (before we loop through every height in blockdb). thx for the comment, I've updated this to start at 1 instead.

database/heightindexdb/dbtest/dbtest.go
DracoLi: unfortunately we cannot close and reopen db here since `heightindexdb.memdb` implements the database and does not persist data.
As an alternative, I've added a `TestSyncPersistence` test to x/blockdb that tests specifically for this so we have it covered. https://github.com/ava-labs/avalanchego/pull/4839/commits/b233a4caf8220311ecd24b7e2f29b74cc4993d57

Issue #4840, opened 2026-01-12
Support genesis runs in C-Chain re-execution benchmark


Elvis339:
**Describe the bug**
Can't run C-Chain re-execution from genesis (block 1) via CI workflow. The script requires both BLOCK_DIR_SRC and CURRENT_STATE_DIR_SRC to be set, but for genesis there's no state to import just need an empty dir.

**To Reproduce**
```
gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --repo ava-labs/avalanchego \
  -f runner="avalanche-avalanchego-runner-2ti" \
  -f block-dir-src="cchain-mainnet-blocks-1m-ldb" \
  -f current-state-dir-src="" \ # or remove this arg, same effect
  -f start-block="1" \
  -f end-block="1000000" \
  -f config="firewood"
```

**Expected behavior**
Should start from genesis with empty state directory, same as running locally with:
```
 BLOCK_DIR=/tmp/c-chain-reexecution-firewood-1m-20260112-115737/blocks \
CURRENT_STATE_DIR=/tmp/c-chain-reexecution-firewood-1m-20260112-115737/current-state \
START_BLOCK=1 \               
END_BLOCK=1000000 \                             
CONFIG=firewood \                                   
./scripts/benchmark_cchain_range.sh
```

Where `CURRENT_STATE_DIR` is empty directory.

**Screenshots**

**Logs**
```
Error: Both BLOCK_DIR_SRC and CURRENT_STATE_DIR_SRC must be provided together
Error: Process completed with exit code 201.
```
Run: https://github.com/ava-labs/avalanchego/actions/runs/20920754409/job/60105390794
Args: 
```
Run ./scripts/run_task.sh test-cchain-reexecution -- ""
  ./scripts/run_task.sh test-cchain-reexecution -- ""
  shell: nix develop --impure --command bash -x {0}
  env:
    BENCHMARK_OUTPUT_FILE: /__w/avalanchego/avalanchego/benchmark-output.json
    BLOCK_DIR_SRC: cchain-mainnet-blocks-1m-ldb
    CURRENT_STATE_DIR_SRC: 
    START_BLOCK: 1
    END_BLOCK: 1000000
    CONFIG: firewood
    PUSH_POST_STATE: 
    RUNNER_TYPE: avalanche-avalanchego-runner-2ti
    METRICS_COLLECTOR_ENABLED: true
```

PR #4841, opened 2026-01-12
fix(c-chain-reexecution): support genesis execution

Elvis339:
## Why this should be merged
This patch is closing: https://github.com/ava-labs/avalanchego/issues/4840

Fixes CI workflow failure when running C-Chain re-execution from genesis (block 1). Previously required both BLOCK_DIR_SRC and CURRENT_STATE_DIR_SRC, but genesis has no state to import just needs an empty directory.

## How this works
- import_cchain_data.sh: CURRENT_STATE_DIR_SRC is now optional. If unset, creates an empty current-state directory instead of downloading from S3.
- benchmark_cchain_range.sh: Single S3 import branch that delegates to the import script. Removes duplicated logic and hardcoded S3 bucket.

## How this was tested
```
avalanchego git:(es/reexecution-genesis-runs) ✗ gh workflow run "C-Chain Re-Execution Benchmark w/ Container" \
  --repo ava-labs/avalanchego \
  --ref es/reexecution-genesis-runs \   
  -f block-dir-src="cchain-mainnet-blocks-1m-ldb" \
  -f start-block="1" \
  -f end-block="1000000" \
  -f timeout-minutes="1440" \
  -f config="firewood"
  ```

https://github.com/ava-labs/avalanchego/actions/runs/20921571440/job/60108217972

No

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -21,7 +21,7 @@ set -euo pipefail
 #   Data sources (provide S3 sources OR local paths):
 #     BLOCK_DIR_SRC: S3 object key for blocks (triggers S3 import).
 #     CURRENT_STATE_DIR_SRC: S3 object key for state. Optional—if unset, creates
-#                            empty state directory for genesis execution (block 1).
+#                            empty state directory for genesis execution.
 #     BLOCK_DIR: Path to local block directory.
 #     CURRENT_STATE_DIR: Path to local current state directory.
 #
--- a/scripts/setup_cchain_data.sh
+++ b/scripts/setup_cchain_data.sh
@@ -4,8 +4,8 @@ set -euo pipefail
 
 # Set up C-Chain test data
 #
-# Downloads blocks from S3 and sets up state directory. For genesis execution
-# (block 1), omit CURRENT_STATE_DIR_SRC to create an empty state directory.
+# Downloads blocks from S3 and sets up state directory. For genesis execution,
+# omit CURRENT_STATE_DIR_SRC to create an empty state directory.
 #
 # Required env vars:
 #   BLOCK_DIR_SRC         - S3 object key for blocks (e.g., cchain-mainnet-blocks-200-ldb)

### CODE REVIEW

File: scripts/setup_cchain_data.sh
Comment (RodrigoVillar) [inline scripts/setup_cchain_data.sh:1]: Instead of having `import_cchain_data.sh` create the empty directory, could this be delegated to `benchmark_cchain_range.sh`?

Comment (Elvis339) [reply scripts/setup_cchain_data.sh:1]: IMO the script's contract is "set up execution data directory" callers get blocks/ and current-state/ ready to use. 
Whether state comes from S3 or is empty is an implementation detail they shouldn't leak into their own code.

Comment (RodrigoVillar) [reply scripts/setup_cchain_data.sh:1]: I don't think this is true. The description of `import_cchain_data.sh` is: 

```
Import C-Chain blocks and state from S3
```

which is different than setting up the execution data directory (which I would argue `benchmark_cchain_range.sh` is already doing by setting the associated environment variables).

Comment (Elvis339) [reply scripts/setup_cchain_data.sh:1]: I use import standalone for custom builds with different CGO flags and different current state directories. That's my workflow and why I added genesis support there.

I think of it as three layers, each independent and composable:
- benchmark_cchain_range.sh is the orchestrator. It sets variables like EXECUTION_DATA_DIR, picks test modes, passes config. It does not touch the filesystem and neither it is setting up test data. Setting EXECUTION_DATA_DIR=/tmp/foo is just a string assignment.
- import_cchain_data.sh does the actual data setup. It calls copy_dir.sh which creates blocks/ and current-state/ on disk. This is where filesystem operations happen. When it finishes, the directories exist and are ready.
- vm_reexecute.go is the test. It takes paths as arguments and runs. Doesn't care how the data got there.
Each layer works standalone. I can call import directly, then compile my own binary and run it:

```
BLOCK_DIR_SRC=cchain-mainnet-blocks-1m-ldb \
EXECUTION_DATA_DIR=/tmp/custom-test \
./scripts/import_cchain_data.sh

CGO_CFLAGS="..." go build ./tests/reexecute/c
./c --block-dir=/tmp/custom-test/blocks --current-state-dir=/tmp/custom-test/current-state-prev ...
./c --block-dir=/tmp/custom-test/blocks --current-state-dir=/tmp/custom-test/current-state ...
```

With genesis in import, this just works. Without it, I'm adding mkdir between steps every time.
Benchmark wrapper is optional. If genesis logic moves to benchmark, anyone bypassing benchmark reimplements it. Genesis mkdir is a filesystem operation. Filesystem operations belong in the layer that does filesystem operations.
If the description "Import C-Chain blocks and state from S3" is the sticking point, I can update it to "Set up C-Chain test data from S3" or rename to prepare_cchain_data.sh. 

This is a minor implementation detail. Both approaches work and not worth bikesheding too long.

Comment (Elvis339) [reply scripts/setup_cchain_data.sh:1]: That said, your point about naming is valid. The original script assumed CURRENT_STATE_DIR_SRC would always be provided. Genesis mode wasn't a use case when it was written. Now that we're supporting empty state, the script does more than import. The name should reflect that. I'll rename import_cchain_data.sh to setup_cchain_data.sh and update the description to match.

Comment (RodrigoVillar) [reply scripts/setup_cchain_data.sh:1]: Thanks for the context. Renaming the script to `setup_cchain_data.sh` and updating it's use case sounds good to me.

Comment (Elvis339) [reply scripts/setup_cchain_data.sh:1]: No problem, always. Here's the change: https://github.com/ava-labs/avalanchego/pull/4841/commits/e58848d92d4d018f59ca92fe2392cc84e91d174d

File: Taskfile.yml [hunk 1]
```diff
@@ -140,9 +140,9 @@ tasks:
     desc: Runs ginkgo against the current working directory
     cmd: ./bin/ginkgo build {{.USER_WORKING_DIR}}
 
-  import-chain-data:
-    desc: Imports the C-Chain block and state data to re-execute.
-    cmd: ./scripts/import_cchain_data.sh
+  setup-chain-data:
+    desc: Sets up C-Chain test data (blocks from S3, state from S3 or empty for genesis).
+    cmd: ./scripts/setup_cchain_data.sh
```
RodrigoVillar: nit: this should be the last task in the file (tasks are alphabetically ordered)

scripts/benchmark_cchain_range.sh
```diff
@@ -20,7 +20,8 @@ set -euo pipefail
 # Environment variables:
 #   Data sources (provide S3 sources OR local paths):
 #     BLOCK_DIR_SRC: S3 object key for blocks (triggers S3 import).
-#     CURRENT_STATE_DIR_SRC: S3 object key for state (triggers S3 import).
+#     CURRENT_STATE_DIR_SRC: S3 object key for state. Optional—if unset, creates
+#                            empty state directory for genesis execution (block 1).
```
RodrigoVillar: (nit/feel-free-to-ignore): Stating `block 1` after genesis execution is repetitive IMO (similar comment for `scripts/setup_cchain_data.sh`)

Taskfile.yml
Elvis339: https://github.com/ava-labs/avalanchego/pull/4841/commits/cda628ca09eafd299b2a8744b86823da26c47431

scripts/benchmark_cchain_range.sh
Elvis339: https://github.com/ava-labs/avalanchego/pull/4841/commits/6c5ade126a88e6f7abe9d92767bb854cd8b6eebc

Taskfile.yml
```diff
@@ -22,17 +22,17 @@ tasks:
       TEST_SETUP: avalanchego
     cmd: bash -x ./scripts/build_antithesis_images.sh
 
+  build-antithesis-images-subnet-evm:
```
maru-ava, line 25: Appreciate the ordering cleanup, but for future please propose in a separate PR rather than mixing functional and non-functional changes.

maru-ava (approved):

PR #4842, opened 2026-01-12
add relayer mode

### LINKED ISSUE #4808: Primary Network/P-Chain Relayer
**Context and scope**
We are supporting an external team that is planning to deploy an Avalanche L1 with strict network isolation requirements. All L1 validator and node infrastructure is expected to run inside closed systems, where communication is only permitted via explicit firewall rules (fixed IPs and ports) between approved institutions. 

The team has proposed the following design:

* Operate designated single/multi P-Chain validators within their infrastructure that is allowed to connect to the public Avalanche mainnet.
* This validator tracks the P-Chain normally.
* All other L1 nodes would:
  * Remain fully isolated from public networks
  * Follow P-Chain updates exclusively via this internal validator
  * This validator would act as a relayer / proxy to the P-Chain for the L1 network.

Image

**Current Limitation:**
Under the existing implementation, this type of relayer-based topology is not supported by the consensus engine. The engine expects nodes to be connected to approximately 80% of the P-Chain stake in order to be healthy and operate correctly. Architectures that rely on a single upstream P-Chain validator—or a very small, fixed set of relayers—do not satisfy this assumption.

Even if we modify the staking parameters in each client node, then there is a risk of chit/preference messages from the relayer can be seemed as final from the client nodes. In this case even if the preference changes on the relayer node, client nodes won't be able to change as they'd quickly finalize it in the consensus.

**Discussion and alternatives**
* Gather designated relayer node IDs with a config flag (can be multiple), should be toggled by the client L1 side
* Then we can override the validator set with those designated relayer node IDs
* Modify the consensus code for receiving chits and just only handle the `acceptedIDs` from the chits on the client side (roughly this [function](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/snowman/engine.go#L359))
  * This change should only impact the client who uses the flag, *it should not impact the server/relayer validator side*
 
**Open questions**
1: Would overriding validator set breaks the warp messaging (from primary network)? 
2: Do we still need to modify the staking paramaters?

ceyonur:
## Why this should be merged

This PR introduces relayer mode for subnets, allowing isolated networks to rely on a designated set of trusted relayer nodes for consensus rather than the full validator set. This supports both L1s and Primary Network.

Closes: #4808 

### `config` package

* Adds a new `--primary-network-relayers` flag with `NodeID=IP:Port` formatting.
* Primary network relayers are used as bootstrappers 
* If this flag is used consensus parameters (K,alpha) are adjusted

### `subnet` package 

* Adds a new field to subnet config jsons for `relayerIDs`

### `chains` package

* Chain manager creates a dedicated validator manager for relayer mode that only contains relayers to be used in consensus engine
* Each relayer gets equal weight (1) for uniform voting power

### consensus changes

When relayer mode is enabled, the consensus engine:
* Only samples from designated relayer nodes
* Follows acceptedID from chits (not preferredID) to ensure finality. This is to make sure client nodes (nodes that activate relayer mode) won't end-up with a early finalization, but rather finalize once blocks are accepted in relayer nodes.

## How this was tested

UTs

Yes


subnets/config.go
```diff
@@ -49,6 +49,38 @@ type Config struct {
...
+
+// AdjustForRelayerMode modifies consensus parameters if relayer mode is enabled.
+// If relayer mode is not enabled (no RelayerIDs), this is a no-op.
+//
+// In relayer mode:
+//   - K is set to the number of relayers (sample all relayers)
+//   - AlphaPreference and AlphaConfidence are set to (numRelayers/2)+1 (simple majority)
+//   - Beta remains unchanged to maintain strong finalization guarantees
+func (c *Config) AdjustForRelayerMode() {
+	if !c.IsRelayerMode() {
```
ceyonur, line 76: I wonder if this should just return an error. I did not want to complicate call-sites by constantly checking if it's relayer mode.

Issue #4702, opened 2025-12-04
Fix `TestStateSyncToggleEnabledToDisabled` flake


JonathanOppenheimer:
See https://github.com/ava-labs/subnet-evm/actions/runs/19942508463/job/57183692605?pr=1896

JonathanOppenheimer:
Linked to wrong issue

PR #4850, opened 2026-01-13
test: Fix `TestResyncNewRootAfterDeletes` flake

JonathanOppenheimer:
## Why this should be merged

Closes https://github.com/ava-labs/avalanchego/issues/4702

## How this works

For context understanding this test: 

The `TestResyncNewRootAfterDeletes` test simulates a scenario where:

1. A client syncs to `root1`, fetching and storing code
2. Code is deleted from the database (simulating corruption)
3. The client then syncs to `root2`

In the flake case, when the code is deleted in step 2 for `root1` it doesn't clean up it's fetch markers, so then when iterating over all `CodeToFetch` markers and checks if the code exists. For markers where rawdb.HasCode() returns false, it re-enqueues the hash for fetching, but since we deleted the code but NOT the markers, stale code hashes from `root1` are added to the fetch queue. The code syncer then tries to fetch these old code hashes, which may cause a hang waiting for code that will never arrive. 

## How this was tested
CI

No

### DIFF
--- a/graft/subnet-evm/sync/statesync/sync_test.go
+++ b/graft/subnet-evm/sync/statesync/sync_test.go
@@ -424,7 +424,7 @@ func TestResyncNewRootAfterDeletes(t *testing.T) {
 				defer codeToFetchIt.Release()
 				for codeToFetchIt.Next() {
 					codeHash := common.BytesToHash(codeToFetchIt.Key()[len(customrawdb.CodeToFetchPrefix):])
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer:
@powerslider requesting you as a reviewer as this is your test I believe?

powerslider (approved):

maru-ava (approved):

PR #4852, opened 2026-01-13
Fix race in tests on VM Shutdown

alarso16:
## Why this should be merged

Data race: see https://github.com/ava-labs/avalanchego/actions/runs/20971917102/job/60277170521?pr=4851

## How this works

The VM was never shutdown, so it didn't clean up any of the threads it started up (including the txpool, which is the race seen)

## How this was tested

Look at the thread numbers in the test run above. The reading thread is less than the writing thread, prior to the txpool being created. This means that it's from the previous test.

No

### DIFF
--- a/graft/coreth/plugin/evm/vm_test.go
+++ b/graft/coreth/plugin/evm/vm_test.go
@@ -2218,7 +2218,9 @@ func TestInspectDatabases(t *testing.T) {
 // Querying for the nonce of the zero address at various heights is sufficient
 // as this succeeds only if the EVM has the matching trie at each height.
 func TestArchivalQueries(t *testing.T) {
-	for _, scheme := range vmtest.Schemes {
+	schemes := []string{customrawdb.FirewoodScheme, rawdb.HashScheme}
+
+	for _, scheme := range schemes {
 		t.Run(scheme, func(t *testing.T) {
 			require := require.New(t)
 			ctx := t.Context()
@@ -2236,9 +2238,9 @@ func TestArchivalQueries(t *testing.T) {
 				}`,
 				Scheme: scheme,
 			})
-			defer func() {
+			t.Cleanup(func() {
 				require.NoError(vm.Shutdown(ctx))
-			}()
+			})
 
 			numBlocks := 10
 			for range numBlocks {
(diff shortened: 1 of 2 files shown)


graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2218,9 +2218,7 @@ func TestInspectDatabases(t *testing.T) {
 // Querying for the nonce of the zero address at various heights is sufficient
 // as this succeeds only if the EVM has the matching trie at each height.
 func TestArchivalQueries(t *testing.T) {
-	schemes := []string{customrawdb.FirewoodScheme, rawdb.HashScheme}
-
-	for _, scheme := range schemes {
+	for _, scheme := range vmtest.Schemes {
```
joshua-kim: This is a good change but we should avoid coupling unrelated changes

graft/subnet-evm/plugin/evm/vm_test.go
```diff
@@ -3655,11 +3655,11 @@ func TestArchivalQueries(t *testing.T) {
 	//	- Tries 0-5: on-disk
 	// 	- Tries 6-10: in-memory
 	tests := []struct {
-		name   string
+		scheme string
```
joshua-kim: Test tables typically use `name` naming - but this is similarly an unrelated change

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -2238,6 +2236,9 @@ func TestArchivalQueries(t *testing.T) {
 				}`,
 				Scheme: scheme,
 			})
+			defer func() {
```
joshua-kim: I think we want to use `t.Cleanup` here (or rather in `newDefaultTestVM`?) - `defer` isn't guaranteed to be run if a test completes because `t.FailNow()` will only respect defers in the goroutine it's called from ([ref](https://pkg.go.dev/testing#T.FailNow))

alarso16: Interesting, I didn't know that! Although it makes sense. I'll use t.Cleanup

graft/coreth/plugin/evm/vm_test.go
alarso16: I'll revert it then

graft/coreth/plugin/evm/vm_test.go
alarso16: Discussed offline, the test should own its own cleanup

PR #4853, opened 2026-01-13
test(reexecution): add additional Firewood archival tests

RodrigoVillar:
## Why this should be merged

As part of https://github.com/ava-labs/firewood/issues/1386, we need to compare the performance of running Firewood in archival mode vs running it w/o archival mode enabled. Ideally, this can be done by running the reexecution tests in parallel using our ARC runners.

## How this works

Adds additional predefined tests in `benchmark_cchain_range.sh`  for testing Firewood in archival mode. As a result of this PR, every predefined regular Firewood test has an archival mode counterpart.

## How this was tested

https://github.com/ava-labs/avalanchego/actions/runs/20973131554/job/60281357152

No

### DIFF
--- a/scripts/benchmark_cchain_range.sh
+++ b/scripts/benchmark_cchain_range.sh
@@ -61,16 +61,18 @@ show_usage() {
 Usage: $0 [test-name]
 
 Available tests:
-  help                         - Show this help message
+  help                              - Show this help message
 
-  default                      - Quick test run (blocks 101-200, hashdb)
-  hashdb-101-250k              - Blocks 101-250k with hashdb
-  hashdb-archive-101-250k      - Blocks 101-250k with hashdb archive
-  hashdb-33m-33m500k           - Blocks 33m-33.5m with hashdb
-  firewood-101-250k            - Blocks 101-250k with firewood
-  firewood-archive-101-250k    - Blocks 101-250k with firewood archive
-  firewood-33m-33m500k         - Blocks 33m-33.5m with firewood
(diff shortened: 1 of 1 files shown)


scripts/benchmark_cchain_range.sh
```diff
@@ -129,13 +131,27 @@ if [[ -n "$TEST_NAME" ]]; then
             END_BLOCK="${END_BLOCK:-33500000}"
             CONFIG="${CONFIG:-firewood}"
             ;;
+        firewood-archive-33m-33m500k)
```
maru-ava, line 134: (No action required) If config can vary by any given firewood test, maybe separate config so that it can be configured separately?

RodrigoVillar: Hmm I'm not sure what you mean - could you elaborate?

maru-ava: Vary config via env var and target a test that defaults to non-archive: `CONFIG=archive scripts/benchmark_cchain_range.sh firewood-33m-33m500k `

RodrigoVillar: I'm open to this but wouldn't this be a deviation from our strategy of having tests that are easily invocable both locally and in CI? (i.e. both CI and the local user would need to set the env variable prior to calling the test).

maru-ava: Fair point, just trying to avoid duplicating test configuration but its not that bad. If it becomes a habit we could rename the test so that the config was a suffix on the test name and set common parameters on a prefix match.

PR #4857, opened 2026-01-14
refactor(sync): reorganize state sync packages

### LINKED ISSUE #4856: Rename and re-structure state sync code packages

powerslider:
## How this works

- Move block sync, code sync, evm state sync into new subpackages.
- Move `vmsync` core into `sync/engine`.
- Move sync types to `sync/types`.

## How this was tested

existing UT

no 

resolves #4856

### DIFF
--- a/graft/coreth/plugin/evm/vm_test.go
+++ b/graft/coreth/plugin/evm/vm_test.go
@@ -1544,7 +1544,6 @@ func TestWaitForEvent(t *testing.T) {
 		err error
 	}
 
-	fortunaFork := upgradetest.Fortuna
 	for _, testCase := range []struct {
 		name     string
 		Fork     *upgradetest.Fork
@@ -1709,48 +1708,6 @@ func TestWaitForEvent(t *testing.T) {
 				res := <-results
 				require.NoError(t, res.err)
 				require.Equal(t, commonEng.PendingTxs, res.msg)
-			},
-		},
-		// TODO (ceyonur): remove this test after Granite is activated. (See https://github.com/ava-labs/avalanchego/graft/coreth/issues/1318)
-		{
-			name: "WaitForEvent does not wait for new block to be built in fortuna",
-			Fork: &fortunaFork,
-			testCase: func(t *testing.T, vm *VM) {
-				t.Parallel()
-				signedTx := newSignedLegacyTx(t, vm.chainConfig, vmtest.TestKeys[0].ToECDSA(), 0, &vmtest.TestEthAddrs[1], big.NewInt(1), 21000, vmtest.InitialBaseFee, nil)
-				blk, err := vmtest.IssueTxsAndSetPreference([]*types.Transaction{signedTx}, vm)
-				require.NoError(t, err)
-				require.NoError(t, blk.Accept(t.Context()))
-				signedTx = newSignedLegacyTx(t, vm.chainConfig, vmtest.TestKeys[0].ToECDSA(), 1, &vmtest.TestEthAddrs[1], big.NewInt(1), 21000, vmtest.InitialBaseFee, nil)
-
-				for _, err := range vm.txPool.AddRemotesSync([]*types.Transaction{signedTx}) {
-					require.NoError(t, err)
-				}
-
-				msg, err := vm.WaitForEvent(t.Context())
-				require.NoError(t, err)
-				require.Equal(t, commonEng.PendingTxs, msg)
-			},
-		},
-		// TODO (ceyonur): remove this test after Granite is activated. (See https://github.com/ava-labs/avalanchego/graft/coreth/issues/1318)
-		{
-			name: "WaitForEvent waits for a delay with a retry in fortuna",
-			Fork: &fortunaFork,
-			testCase: func(t *testing.T, vm *VM) {
-				t.Parallel()
-				lastBuildBlockTime := time.Now()
-				_, err := vm.BuildBlock(t.Context())
-				require.NoError(t, err)
-				// we haven't accepted the previous built block, so this should be a retry
-				signedTx := newSignedLegacyTx(t, vm.chainConfig, vmtest.TestKeys[0].ToECDSA(), 0, &vmtest.TestEthAddrs[1], big.NewInt(1), 21000, vmtest.InitialBaseFee, nil)
-				for _, err := range vm.txPool.AddRemotesSync([]*types.Transaction{signedTx}) {
-					require.NoError(t, err)
-				}
-
-				msg, err := vm.WaitForEvent(t.Context())
-				require.NoError(t, err)
-				require.Equal(t, commonEng.PendingTxs, msg)
-				require.GreaterOrEqual(t, time.Since(lastBuildBlockTime), RetryDelay)
 			},
 		},
 	} {
--- a/graft/subnet-evm/plugin/evm/vm_test.go
+++ b/graft/subnet-evm/plugin/evm/vm_test.go
@@ -2912,7 +2912,6 @@ func TestWaitForEvent(t *testing.T) {
 		err error
 	}
 
-	fortunaFork := upgradetest.Fortuna
 	for _, testCase := range []struct {
 		name     string
 		Fork     *upgradetest.Fork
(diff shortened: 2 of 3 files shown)


graft/coreth/plugin/evm/atomic/sync/summary_provider.go
```diff
@@ -7,14 +7,15 @@ import (
...
 
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/types"
 
 	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/atomic/state"
-	"github.com/ava-labs/avalanchego/graft/coreth/sync"
+	"github.com/ava-labs/avalanchego/graft/coreth/sync/types"
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+
+	evmtypes "github.com/ava-labs/libevm/core/types"
```
alarso16:
```suggestion
	ethtypes "github.com/ava-labs/libevm/core/types"
```

I think the scope of this is a lot clearer, since `evmtypes` might lead me to believe this is the `plugin/evm` block, not the `libevm` block

graft/coreth/sync/engine/client.go
alarso16, line 1: I don't love `core` as the package name, since it doesn't actually give us any sense of scope. I'm having a tough time coming up with something more clear, but I think these are options:
* `vm`
* move up into `sync`

graft/coreth/sync/engine/registry.go
alarso16, line 1: Do all the types in this file still need to be exported? At minimum, I think `SyncerRegistry` and `SyncerTask` are redundant names

graft/coreth/sync/evmstate/state_syncer.go
alarso16, line 1: This package name is a little weird to me (albeit an improvement). Specifically, Isn't this stateSync object used in the atomic syncer (which is not evmstate)? How about `hashdb` to separate from the incoming `firewood` syncer?

graft/coreth/sync/leaf/syncer.go
alarso16, line 1: This seems to be specific to the `evmstate` stuff, why does this belong to the client?

graft/coreth/sync/client/client.go
alarso16, line 1: Why not make the path `sync/client/client.go`? Sounds less redundant

graft/coreth/sync/syncclient/client.go
alarso16:
```suggestion
type Config struct {
```

alarso16:
```suggestion
func New(config *ClientConfig) *client {
```

graft/coreth/plugin/evm/atomic/sync/extender.go
```diff
@@ -29,7 +28,7 @@ func (a *Extender) Initialize(backend *state.AtomicBackend, trie *state.AtomicTr
 }
 
 // CreateSyncer creates the atomic syncer with the given client and verDB.
-func (a *Extender) CreateSyncer(client syncclient.LeafClient, verDB *versiondb.Database, summary message.Syncable) (sync.Syncer, error) {
+func (a *Extender) CreateSyncer(client syncclient.LeafClient, verDB *versiondb.Database, summary message.Syncable) (types.Syncer, error) {
```
alarso16: slightly out of scope
```suggestion
func (e *Extender) CreateSyncer(client syncclient.LeafClient, verDB *versiondb.Database, summary message.Syncable) (types.Syncer, error) {
```

graft/coreth/sync/client/test_network.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package statesyncclient
```
JonathanOppenheimer, line 4: could this just be `client`? same logic as `block`

graft/coreth/sync/client/client.go
JonathanOppenheimer: oh austin beat me to this -- but yeah, the package could be renamed too?

graft/coreth/plugin/evm/atomic/sync/summary_provider.go
JonathanOppenheimer: If you change this here, there are log debug statements you can change too

graft/coreth/sync/engine/client.go
powerslider: That's exactly the same issue I was facing as well. We have two options - either use `core` or further subdivide into more packages which I thought was a bit excessive for now. I'd rather keep `core` for now and later we can decide something better. Wdyt?

Also moving up into `sync` would result in mega usage of import aliases everywhere, due to conflicting with Go's `sync` package and not only.

graft/coreth/sync/evmstate/state_syncer.go
powerslider: I think it makes lots of sense to change it to `hashdb`, but I was a bit hesitant if that is going to mess with the semantics (because it's a `hashdb` based syncer and not anything directly storage related), but I am perfectly ok with naming it `hashdb`.

The only possible downside is that we are kinda switching the semantics from "what to sync? - like `block` and `code` and `evmstate`" to "where to sync? - like `hashdb` and `firewood`". This is a bit inconsistent and we probably need to chose only one of the questions.

At some point we will deprecate `hashdb` and `firewood` will be the only option and at that point I think it will matter less "where to sync?" and more again "what to sync? - `evmstate`". There is no right or wrong answer in the grand scheme of things, but just pushing those arguments as food for thought.

graft/coreth/sync/leaf/syncer.go
powerslider: Great question, mostly legacy reasons. I'll think about if there is a better place to move it.

graft/coreth/sync/client/client.go
powerslider: The reason I did not go for `client` was because it shadowed pretty frequently with client named variables and other packages as well. And since we had packages like `synctest` and `syncutils` I thought that it would be acceptable to have `syncclient`, but it's a mixed bag here.

graft/coreth/sync/syncclient/client.go
powerslider: yeah this is a good opportunity to rename it as well.

graft/coreth/sync/evmstate/state_syncer.go
alarso16: Yeah I understand your point, but the state syncer object already is a "where to sync?" object. I think we should rename it, since I think it will be clearer and we will hopefully delete it all soon ;)

graft/coreth/sync/client/client.go
alarso16: Ah yeah that does make sense too. I don't love that there's multiple kinds of sync clients, so that seems like something we'd want fix eventually. Don't have any ideas for better names though

graft/coreth/plugin/evm/vmtest/test_syncervm.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/ava-labs/avalanchego/graft/coreth/params/paramstest"
 	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/customtypes"
 	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/extension"
-	"github.com/ava-labs/avalanchego/graft/coreth/plugin/evm/vmsync"
+	"github.com/ava-labs/avalanchego/graft/coreth/sync/syncclient"
```
alarso16: Where is this? I though you renamed it

graft/coreth/sync/evmstate/state_syncer.go
alarso16: On second review, the atomic syncer does not import this, so you might be right. What we could do is just have a single entry-point, and then return different syncer implementations based on whether it's a firewood or hashdb database

graft/coreth/sync/leaf/syncer.go
alarso16: I like this!

graft/coreth/sync/engine/registry.go
powerslider: I think they should remain the way they are because there is no stutter in the names, e.g. `core.SyncerRegistry`. I did unexport `syncerTask` as it has only private usage and that is valid.

graft/coreth/sync/client/client.go
powerslider: I actually did rename to `client` and kinda made it work well when I extracted `leaf_syncer.go` into its own `leaf` package. That automatically reduced the imports of the client package and actually was acceptable to rename it `client`.

graft/coreth/plugin/evm/atomic/sync/syncer.go
```diff
@@ -31,9 +31,9 @@ const (
 )
 
 var (
-	_ sync.Syncer             = (*Syncer)(nil)
-	_ syncclient.LeafSyncTask = (*syncerLeafTask)(nil)
-	_ sync.Finalizer          = (*Syncer)(nil)
+	_ types.Syncer    = (*Syncer)(nil)
+	_ leaf.SyncTask   = (*syncerLeafTask)(nil)
```
alarso16: nit: will you group these together by type

graft/coreth/sync/engine/registry.go
```diff
@@ -12,21 +12,20 @@ import (
...
-	syncpkg "github.com/ava-labs/avalanchego/graft/coreth/sync"
+	"github.com/ava-labs/avalanchego/graft/coreth/sync/types"
 )
 
 var errSyncerAlreadyRegistered = errors.New("syncer already registered")
 
-// SyncerTask represents a single syncer with its name for identification.
-type SyncerTask struct {
+// syncerTask represents a single syncer with its name for identification.
+type syncerTask struct {
```
alarso16, line 21: Not sure these types need to be prefaced with "syncer"

powerslider: I still think they should because for now we have `core.syncerTask` and `core.SyncerRegistry` - there is no stutter in the naming and I still think it adds meaning by implying that it manages syncers.

graft/coreth/sync/core/client.go
```diff
@@ -1,15 +1,14 @@
 // Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package vmsync
+package core
```
ceyonur: can we avoid using `core` as we have a massive `core` package already?

graft/coreth/sync/engine/server.go
```diff
@@ -31,7 +30,7 @@ type Server interface {
 	GetStateSummary(context.Context, uint64) (block.StateSummary, error)
 }
 
-func NewServer(chain *core.BlockChain, provider synccommon.SummaryProvider, syncableInterval uint64) Server {
+func NewServer(chain *core.BlockChain, provider types.SummaryProvider, syncableInterval uint64) Server {
```
ceyonur, line 33: nit: I think the suggested behaviour is to accept interfaces and return concrete types.

ceyonur: might not be directly related to this PR though

powerslider: yes, this is a general good practice rule, although there are cases where it is ok to return an interface. I have not touched anything in this file outside of the package name change. I prefer not to make changes here since it is outside of the non-functional changes scope of this PR.

graft/coreth/sync/core/client.go
powerslider: do you have a better name suggestion?

ceyonur: I think the previous `statesync` describes this at best or if we want to be more generic maybe `engine` works?

powerslider: if we are ok with the term `engine` for sync related stuff, then I am fine with it.

PR #4860, opened 2026-01-14
refactor: granite cleanup - remove pre-granite test-cases

JonathanOppenheimer:
## Why this should be merged
Granite has been activated for several months now and we can clean up pre-Granite test code. This PR removes test cases that were specifically testing pre-Granite behavior, which is no longer relevant.

## How this works
- Remove pre-Granite test cases from `vm_test.go` in both coreth and subnet-evm
   - With Granite always active, these test cases are no longer meaningful

## How this was tested
CI

No

### DIFF
--- a/graft/coreth/plugin/evm/vm_test.go
+++ b/graft/coreth/plugin/evm/vm_test.go
@@ -1544,7 +1544,6 @@ func TestWaitForEvent(t *testing.T) {
 		err error
 	}
 
-	fortunaFork := upgradetest.Fortuna
 	for _, testCase := range []struct {
 		name     string
 		Fork     *upgradetest.Fork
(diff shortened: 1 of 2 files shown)


graft/coreth/plugin/evm/vm_test.go
```diff
@@ -1711,48 +1710,6 @@ func TestWaitForEvent(t *testing.T) {
 				require.Equal(t, commonEng.PendingTxs, res.msg)
 			},
 		},
-		// TODO (ceyonur): remove this test after Granite is activated. (See https://github.com/ava-labs/avalanchego/graft/coreth/issues/1318)
-		{
-			name: "WaitForEvent does not wait for new block to be built in fortuna",
-			Fork: &fortunaFork,
```
JonathanOppenheimer, line 1717: Here is the operative TODO!

PR #4863, opened 2026-01-14
refactor: Share core/state

alarso16:
## Why this should be merged

This code is literally duplicated, besides a couple hash changes in tests (which are unnecessary in the first place).

## How this works

Moves to `graft/evm`

## How this was tested

CI

No


graft/coreth/core/state/snapshot/generate_test.go
```diff
@@ -1,982 +0,0 @@
...
-	"github.com/ava-labs/libevm/rlp"
-	"github.com/ava-labs/libevm/trie"
-	"github.com/ava-labs/libevm/trie/trienode"
-	"github.com/ava-labs/libevm/triedb"
-	"github.com/holiman/uint256"
-	"golang.org/x/crypto/sha3"
-)
-
-func TestMain(m *testing.M) {
-	customtypes.Register()
```
alarso16, line 53: This type registration is unnecessary. The test helpers create accounts, rlp encode them, and insert them using a test API to the trie. The only reason the hashes are different is due to IsMulticoin. The differences in these tests do not actually test any snapshot behavior

graft/subnet-evm/core/state/trie_prefetcher_extra_test.go
alarso16, line 1: This is long since deleted in coreth, since the functionality it's checking has been moved (can't find the exact commit though)

PR #4865, opened 2026-01-15
refactor: Use libevm accounts/abi

alarso16:
## Why this should be merged

We had almost all the functionality necessary in libevm to use `accounts/abi`, so a few small refactor PRs to make this work

***DO NOT MERGE WITHOUT SENIOR EVM APPROVAL***

## How this works

Updates libevm, uses those imports

## How this was tested

CI.

No.

### DIFF
--- a/graft/coreth/core/blockchain_test.go
+++ b/graft/coreth/core/blockchain_test.go
@@ -376,6 +376,7 @@ func TestCorruptSnapshots(t *testing.T) {
 }
 
 func TestBlockChainOfflinePruningUngracefulShutdown(t *testing.T) {
+	t.Skip("Flaky test from go-ethereum")
 	create := func(db ethdb.Database, gspec *Genesis, lastAcceptedHash common.Hash, _ string) (*BlockChain, error) {
 		// Import the chain. This runs all block validation rules.
 		blockchain, err := createBlockChain(db, pruningConfig, gspec, lastAcceptedHash)
@@ -417,7 +418,6 @@ func TestBlockChainOfflinePruningUngracefulShutdown(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.Name, func(t *testing.T) {
-			t.Parallel()
 			tt.testFunc(t, create)
 		})
 	}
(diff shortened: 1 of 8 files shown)


graft/coreth/core/evm.go
```diff
@@ -69,6 +69,11 @@ func WithTempRegisteredExtras(lock libevm.ExtrasLock, fn func() error) error {
 
 type hooks struct{}
 
+// PreprocessingGasCharge is not necessary.
```
JonathanOppenheimer, line 72: Can you explain this? I must have missed prior context -- if it's not necessary why did we add it? Does this satisfy an interface or something

graft/subnet-evm/precompile/contracts/feemanager/feemanagertest/bindings/gen_feemanagertest_binding.go
```diff
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	ethereum "github.com/ava-labs/libevm"
-	"github.com/ava-labs/avalanchego/graft/subnet-evm/accounts/abi"
+	"github.com/ava-labs/libevm/accounts/abi"
```
JonathanOppenheimer, line 12: W

graft/coreth/core/evm.go
alarso16: It's for strevm, but need it to fulfill an interface

maru-ava: (No action required) Maybe capture this kind of context in the docstring so it is available to future maintainers/reviewers? 

When I'm reviewing or being reviewed, questions like this often indicate an opportunity to bolster context in the code so that someone browsing it directly can more easily infer intent without having to luck into reading a comment thread like this.

graft/evm/go.mod
```diff
@@ -4,9 +4,9 @@ go 1.24.12
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.14.1-0.20251120155522-df4a8e531761
+	github.com/ava-labs/avalanchego v0.0.0-00010101000000-000000000000
```
ceyonur: this looks weird

graft/coreth/core/evm.go
alarso16: Seems reasonable. I extended the comment to say exactly this.

graft/evm/go.mod
alarso16: Fixed (since it's ignored, super easy)

PR #4871, opened 2026-01-16
build: add ACP-224 interface and build ABI +  bindings 

JonathanOppenheimer:
## Why this should be merged

This is the first stage of implementing [ACP 224](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md). The solidity interface is added, as specified by the ACP, and automatic compilation is added, tying the `.abi` file to the actual Solidity source, establishing a source of truth that is verified automatically by CI.

## How this works

This PR employs the standard `compile.go`.

## How this was tested
CI, specifically `check-generate-bindings`

No


graft/subnet-evm/precompile/contracts/acp224feemanager/IACP224FeeManager.sol
```diff
@@ -0,0 +1,43 @@
+//SPDX-License-Identifier: MIT
```
JonathanOppenheimer, line 1: See [here](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md#dynamic-gas-target--fee-configuration-via-acp224feemanagerprecompile) for the source of this file.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package bindings
+
+// Step 1: Compile interface to generate ABI at top level
+//go:generate sh -c "solc-v0.8.30 -o ../.. --overwrite --abi --base-path ../../../../.. --pretty-json --evm-version cancun ../../IACP224FeeManager.sol"
```
maru-ava: Maybe file an issue to replace use of nix-provided solc?

graft/subnet-evm/precompile/contracts/acp224feemanager/IACP224FeeManager.sol
```diff
@@ -0,0 +1,43 @@
...
+/// @title ACP-224 Fee Manager Interface
+/// @notice Interface for managing dynamic gas limit and fee parameters according to ACP-224
+/// @dev Inherits from IAllowList for access control
+interface IACP224FeeManager is IAllowList {
+    /// @notice Configuration parameters for the ACP-224 dynamic fee mechanism
+    struct FeeConfig {
+        uint256 targetGas; // Target gas consumption per second
+        uint256 minGasPrice; // Minimum gas price in wei
+        uint256 maxCapacityFactor; // Maximum capacity factor (C = factor * T)
+        uint256 timeToDouble; // Time for gas price to double at max capacity
```
maru-ava: (No action required) Maybe reflect that the time is in seconds as per the original?

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
JonathanOppenheimer: Yeah we should definitely do this -- I can take a look at doing so next week. Here's the issue: https://github.com/ava-labs/avalanchego/issues/4876

graft/subnet-evm/precompile/contracts/acp224feemanager/IACP224FeeManager.sol
```diff
@@ -0,0 +1,43 @@
...
+
+/// @title ACP-224 Fee Manager Interface
+/// @notice Interface for managing dynamic gas limit and fee parameters
+/// @dev Inherits from IAllowList for access control
+interface IACP224FeeManager is IAllowList {
+    /// @notice Configuration parameters for the dynamic fee mechanism
+    struct FeeConfig {
+        uint256 targetGas; // Target gas consumption per second
+        uint256 minGasPrice; // Minimum gas price in wei
+        uint256 maxCapacityFactor; // Maximum capacity factor (C = factor * T)
```
ceyonur: this has been removed from ACP

JonathanOppenheimer: I removed this field.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
ARR4N: Why is `-evm-version` being specified for ABI? Same as below.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
```diff
@@ -0,0 +1,14 @@
...
+// Step 2: Compile interface to generate bin file for bindings
+//go:generate solc-v0.8.30 -o artifacts --overwrite --abi --bin --base-path ../../../../.. --metadata-hash none --evm-version cancun ../../IACP224FeeManager.sol
```
ARR4N: Why is the first step using `sh -c` while this is direct?

ARR4N: There is no `--bin` for an interface.

graft/subnet-evm/precompile/contracts/acp224feemanager/acp224feemanagertest/bindings/compile.go
```diff
@@ -0,0 +1,14 @@
...
+
+// Step 1: Compile interface to generate ABI at top level
+//go:generate sh -c "solc-v0.8.30 -o ../.. --overwrite --abi --base-path ../../../../.. --pretty-json --evm-version cancun ../../IACP224FeeManager.sol"
+// Step 2: Compile interface to generate bin file for bindings
+//go:generate solc-v0.8.30 -o artifacts --overwrite --abi --bin --base-path ../../../../.. --metadata-hash none --evm-version cancun ../../IACP224FeeManager.sol
+// Step 3: Generate Go bindings from the compiled artifacts
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type IACP224FeeManager --abi ../../IACP224FeeManager.abi --bin artifacts/IACP224FeeManager.bin --out gen_iacp224feemanager_binding.go
+// Step 4: Replace import paths in generated binding to use subnet-evm instead of libevm
+// This is necessary because the libevm bindings package is not compatible with the subnet-evm simulated backend, which is used for testing.
+//go:generate sh -c "sed -i.bak -e 's|github.com/ava-labs/libevm/accounts/abi/bind|github.com/ava-labs/avalanchego/graft/subnet-evm/accounts/abi/bind|g' gen_iacp224feemanager_binding.go && rm -f gen_iacp224feemanager_binding.go.bak"
```
ARR4N, line 12: Why is it being used for testing if it's not compatible? We shouldn't be introducing new dependencies on deprecated code.

JonathanOppenheimer: We have decided to work on the new precompilegen post SAE.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
JonathanOppenheimer: This is an artifact of using your solc github action. after we switched to nix it was no longer necessary, but I haven't touched this PR in awhile. Fixed

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
JonathanOppenheimer: This was a bad copy and paste from the other files - I've fixed it everywhere.

JonathanOppenheimer: `sh -c` was used elsewhere when necessary for commands with `&&` but it looks like it got copied into a lot of unnecessary steps. Removed all the unnecessary usages.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
ARR4N: Using Nix vs the GitHub action have no bearing on whether or not the EVM version needs to be specified.

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
ARR4N: From which other files? Why was it in those? Just because it's not necessary here doesn't mean that it's not necessary in the source from which you copied it. This is [cargo-culting](https://en.wikipedia.org/wiki/Cargo_cult) and it's bad practice to blindly copy-paste without delving into the rationale.

JonathanOppenheimer: Check the diff -- it was present for all the interfaces, and was necessary nowhere for the reasons you said -- a good catch!!

graft/subnet-evm/precompile/contracts/acp224feemanager/acpfeemanagertest/bindings/compile.go
JonathanOppenheimer: IIRC there were issues with the newer version of solc and the cancun Shanghai op code optimizations so we failed some unit tests, so I had specified the version explicitly as Shanghai or cancun for a few files - I don't remember exactly...

Issue #4679, opened 2025-12-11
Refactor subnet-evm antithesis test setup into `./scripts/build_antithesis_images.sh` 


JonathanOppenheimer:
Maybe file an issue to refactor the subnet-evm antithesis test setup from its own script to a new subnet-evm TEST_SETUP in ./scripts/build_antithesis_images.sh?

_Originally posted by @maru-ava in https://github.com/ava-labs/avalanchego/pull/4671#discussion_r2611018896_

PlatCore:
The #4836 has been merged. 

Closing.

PR #4836, opened 2026-01-10
Migrate subnet-evm antithesis builder to the root

PlatCore:
## Why this should be merged

Consolidates subnet-evm antithesis test infrastructure to the root level, matching the pattern used by avalanchego and xsvm. Closes #4677,  #4679

## How this works

Uses a hybrid structure:
- Dockerfiles moved to `tests/antithesis/subnet-evm/` (consolidated with other test setups)
- Go source stays in `graft/subnet-evm/tests/antithesis/` (needs to import from graft packages)

This avoids import violations while consolidating the build infrastructure. The root build script now runs the subnet-evm antithesis tests with `TEST_SETUP=subnet-evm`.

## How this was tested

CI passes, but you can verify that: 
- `go test ./vms/evm -run TestImportViolations` passes
- Built workload and gencomposeconfig binaries successfully

## Need to be documented in RELEASES.md?

No

PR #4877, opened 2026-01-17
test: share antithesis test setup for subnet-evm

### LINKED ISSUE #4677: Migrate `graft/subnet-evm/tests/antithesis` to `./tests/antithesis/subnet-evm`
Maybe file an issue to migrate the contents of this package to ./tests/antithesis/submet-evm so that all our antithesis test setups are in one place?

_Originally posted by @maru-ava in https://github.com/ava-labs/avalanchego/pull/4671#discussion_r2611080190_

JonathanOppenheimer:
## Why this should be merged

Consolidates subnet-evm antithesis test infrastructure to the root level, matching the pattern used by avalanchego and xsvm. Closes #4677,  #4679

## How this works

Uses a hybrid structure:
- Dockerfiles moved to `tests/antithesis/subnet-evm/` (consolidated with other test setups)
- Go source stays in `graft/subnet-evm/tests/antithesis/` (needs to import from graft packages)

This avoids import violations while consolidating the build infrastructure. The root build script now runs the subnet-evm antithesis tests with `TEST_SETUP=subnet-evm`.

## How this was tested

CI passes, but you can verify that: 
- `go test ./vms/evm -run TestImportViolations` passes
- Built workload and gencomposeconfig binaries successfully

No


Taskfile.yml
```diff
@@ -24,7 +24,8 @@ tasks:
 
   build-antithesis-images-subnet-evm:
     desc: Builds docker images for antithesis for the subnet-evm test setup
-    dir: graft/subnet-evm
+    env:
+      TEST_SETUP: subnet-evm
```
JonathanOppenheimer, line 28: This specifies which one to build for. A reasonable approach?

JonathanOppenheimer:
Note that the "Build Antithesis images" job is a required check, but no longer exists, as it's been split into xsvm and subnet-evm jobs, which both complete successfully, but still displays as waiting for status to be reported. I believe the required status checks need to be updated.

scripts/build_antithesis_images.sh
```diff
@@ -70,22 +70,55 @@ if [[ "${TEST_SETUP}" == "avalanchego" ]]; then
...
+    # subnet-evm gencomposeconfig is in a different module, so cd there first
+    (cd "${AVALANCHE_PATH}/graft/subnet-evm" && \
+     env TARGET_PATH="${AVALANCHE_PATH}/build/antithesis/${TEST_SETUP}" \
+         IMAGE_TAG="${IMAGE_TAG}" \
+         AVALANCHEGO_PATH="${AVALANCHE_PATH}/build/avalanchego" \
+         AVAGO_PLUGIN_DIR="${AVALANCHE_PATH}/build/plugins" \
+         go run ./tests/antithesis/gencomposeconfig)
+  else
+    # xsvm and others are in the root module
+    if [[ "${TEST_SETUP}" == "xsvm" ]]; then
```
alarso16: This is the only case here, doesn't really make sense to include this part in an if statement

JonathanOppenheimer: I was thinking for other TEST_SETUP's in the future

JonathanOppenheimer: although I guess YAGNI

Taskfile.yml
PlatCore: As long as others follow this convention, it seems so.

PlatCore (commented):
@JonathanOppenheimer, instead of walking "the migration" path, you've chosen a more conservative strategy: consolidating existing parts (and removing them from `graft`), and reassigning them to new entry points in `scripts/build_antithesis_images.sh` and the root `Taskfile` (leveraging the TEST_SETUP flag as a dispatcher).
Most importantly, this approach preserves dependencies by avoiding loops.

If my understanding is correct, your approach looks cleaner than the "migration" one I used in #4836. Should I abandon it?

JonathanOppenheimer:
> @JonathanOppenheimer, instead of walking "the migration" path, you've chosen a more conservative strategy: consolidating existing parts (and removing them from `graft`), and reassigning them to new entry points in `scripts/build_antithesis_images.sh` and the root `Taskfile` (leveraging the TEST_SETUP flag as a dispatcher).
> Most importantly, this approach preserves dependencies by avoiding loops.
> 
> If my understanding is correct, your approach looks cleaner than the "migration" one I used in #4836. Should I abandon it?

Yup that's a good summary -- this is just what I had in mind, given the restrictions of the repository! 

One thing I will note, is that in my opinion, it doesn't make much sense to "duplicate" the antithesis testing, by leaving it in subnet-evm and adding it to the root directory. That approach will just run everything twice in CI, for a gain of..? (if the consolidation PR adds more lines of code than it removes, it's not much of a consolidation).

Ultimately I will defer to Maru, as he originally proposed the issue and knows this area a lot better than I do -- and feel free to close this PR if you're not a fan of my approach.

PlatCore:
> One thing I will note, is that in my opinion, it doesn't make much sense to "duplicate" the antithesis testing,

@JonathanOppenheimer there was / is an additional commit on #4836 that cleaned up the "duplicates." However, after meeting with Maru, receiving his feedback, I set out on refactoring this PR and started from fixing CI. I reset the branch head to the "duplicating" commit. This allowed me to test the "migrated" `task` and CI calls against the original ones side-by-side. As a result, the clean-up commit/stage is no longer present / visible.

Anyways, @maru-ava pplease review this PR and let me know if any antithesis migration scope remains unaddressed (IIUIC, it's been handled).

JonathanOppenheimer:
> > One thing I will note, is that in my opinion, it doesn't make much sense to "duplicate" the antithesis testing,
> 
> @JonathanOppenheimer there was / is an additional commit on #4836 that cleaned up the "duplicates." However, after meeting with Maru, receiving his feedback, I set out on refactoring this PR and started from fixing CI. I reset the branch head to the "duplicating" commit. This allowed me to test the "migrated" `task` and CI calls against the original ones side-by-side. As a result, the clean-up commit/stage is no longer present / visible.
> 
> Anyways, @maru-ava pplease review this PR and let me know if any antithesis migration scope remains unaddressed (IIUIC, it's been handled).

@PlatCore if you'd prefer to defer to this PR, I would appreciate your review as well!

PR #4879, opened 2026-01-19
Add support for mutable validator state

rrazvan1:
## Why this should be merged
This PR adds support for modifying validators, which is required for ACP-236.

## How this works
It introduces a new `modified` status for validators and implements `UpdateCurrentValidator` across the state interfaces.

## How this was tested
Unit tests

No

### DIFF
--- a/vms/platformvm/state/diff_test.go
+++ b/vms/platformvm/state/diff_test.go
@@ -1045,34 +1045,34 @@ func TestDiffStacking(t *testing.T) {
 func TestDiffUpdateValidatorErrors(t *testing.T) {
 	tests := []struct {
 		name            string
-		updateValidator func(*Staker)
+		mutateValidator func(*Staker)
 		updateState     func(*require.Assertions, Diff)
 		expectedErr     error
 	}{
 		{
 			name: "invalid mutation",
-			updateValidator: func(validator *Staker) {
+			mutateValidator: func(validator *Staker) {
 				validator.Weight = 5
 			},
 			expectedErr: ErrInvalidStakerMutation,
 		},
 		{
 			name: "invalid staker mutation",
-			updateValidator: func(staker *Staker) {
+			mutateValidator: func(staker *Staker) {
 				staker.Weight -= 1
 			},
 			expectedErr: ErrInvalidStakerMutation,
 		},
 		{
 			name: "staker not found",
-			updateValidator: func(staker *Staker) {
+			mutateValidator: func(staker *Staker) {
 				staker.NodeID = ids.GenerateTestNodeID()
 			},
 			expectedErr: database.ErrNotFound,
 		},
 		{
 			name: "missing validator",
-			updateValidator: func(validator *Staker) {
+			mutateValidator: func(validator *Staker) {
 				validator.NodeID = ids.GenerateTestNodeID()
 			},
 			expectedErr: database.ErrNotFound,
@@ -1093,7 +1093,7 @@ func TestDiffUpdateValidatorErrors(t *testing.T) {
 		},
 		{
 			name: "valid mutation",
-			updateValidator: func(validator *Staker) {
+			mutateValidator: func(validator *Staker) {
 				validator.Weight = 15
 			},
 		},
@@ -1129,8 +1129,8 @@ func TestDiffUpdateValidatorErrors(t *testing.T) {
 			}
 
 			validator := *currentValidator
-			if test.updateValidator != nil {
-				test.updateValidator(&validator)
+			if test.mutateValidator != nil {
+				test.mutateValidator(&validator)
 			}
 
 			err = d.UpdateCurrentValidator(&validator)


vms/platformvm/state/stakers.go
```diff
@@ -269,17 +306,31 @@ type diffValidator struct {
...
-		Decrease: d.validatorStatus == deleted,
-	}
-	if d.validatorStatus != unmodified {
+	var weightDiff ValidatorWeightDiff
+	switch d.validatorStatus {
+	case added:
 		weightDiff.Amount = d.validator.Weight
+	case modified:
+		var err error
+		weightDiff.Amount, err = math.Sub(d.validator.Weight, d.oldWeight)
```
yacovm, line 322: But we can decrease the weight, can't we? 

If the previous weight is 100 and the new weight is 50, the delta is negative and we have decreased the weight. 

So shouldn't it be something like:

```
		if d.validator.Weight >= d.oldWeight {
			weightDiff.Amount, err = math.Sub(d.validator.Weight, d.oldWeight)
		} else {
			weightDiff.Amount, err = math.Sub(d.oldWeight, d.validator.Weight)
			weightDiff.Decrease = true
		}
```

?

vms/platformvm/state/staker.go
```diff
@@ -111,3 +118,27 @@ func NewPendingStaker(txID ids.ID, staker txs.ScheduledStaker) (*Staker, error)
...
+
+	if s.Weight > ms.Weight {
+		// Weight can only increase (by accruing rewards from continuous staking).
+		return errDecreasedWeight
+	}
+
+	return nil
+}
+
+func (s *Staker) immutableFieldsAreUnmodified(ms *Staker) bool {
```
yacovm, line 135: just because I'm paranoid, can we do it like this?

```
func (s *Staker) immutableFieldsAreUnmodified(ms *Staker) bool {
	bothPKNil := s.PublicKey == nil && ms.PublicKey == nil
	bothPKNotNil := s.PublicKey != nil && ms.PublicKey != nil
	publicKeysEqual := bothPKNil || bothPKNotNil && s.PublicKey.Equals(ms.PublicKey)
	
	return s.TxID == ms.TxID &&
		s.NodeID == ms.NodeID &&
		publicKeysEqual &&
		s.SubnetID == ms.SubnetID &&
		s.Priority == ms.Priority
}
```

vms/platformvm/state/staker.go
```diff
@@ -111,3 +118,27 @@ func NewPendingStaker(txID ids.ID, staker txs.ScheduledStaker) (*Staker, error)
 		Priority:  staker.PendingPriority(),
 	}, nil
 }
+
+func (s *Staker) ValidateMutation(ms *Staker) error {
```
yacovm, line 122: why is this an exported function? It should be un-exported, since it's only used in the same package

rrazvan1: Platformvm's packages are more than they should be imo, so I am using exported/unexported more than their normal definitions, like if they are used internally or externally.

The real solution here would be to use interfaces, but that requires changing too much.

yacovm: I think we have enough interfaces in the project 😉

vms/platformvm/state/diff_test.go
```diff
@@ -1040,3 +1041,168 @@ func TestDiffStacking(t *testing.T) {
 	require.NoError(err)
 	require.Equal(owner3, owner)
 }
+
+func TestDiffUpdateValidatorErrors(t *testing.T) {
+	tests := []struct {
+		name            string
+		updateValidator func(*Staker)
```
yacovm: should probably be called "mutateValidator" to not conflict with the function we're testing

vms/platformvm/state/staker_test.go
```diff
@@ -200,6 +200,105 @@ func TestNewPendingStaker(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			require.ErrorIs(
+				test.expectedErr,
+				staker.ValidateMutation(
+					test.mutateFn(*staker),
+				),
```
DracoLi, line 296: argument order for require.ErrorIs is (actual, target) so I think we need to switch this

vms/platformvm/state/stakers.go
```diff
@@ -269,17 +306,31 @@ type diffValidator struct {
 	// mean that diffValidator hasn't change, since delegators may have changed.
 	validatorStatus diffValidatorStatus
```
DracoLi, line 307: this status enum current has two concerns "did the validator join/leave the set?" and "was the validator updated in-place?". This seemed a bit weird and hard to understand when reviewing the code.
Should we consider splitting this out to something like the following?
```golang
type diffValidator struct {
    lifecycle  validatorLifecycle // unchanged | added | removed
    updated bool                   // was the staker config updated
    oldWeight  uint64            // pre-update weight
    // ...
}
```

not blocking but worth considering.

vms/platformvm/state/stakers.go
rrazvan1: There is an invariant: existing validators can only increase their weights.

rrazvan1:
Because this has been implemented in #5017.

PR #4881, opened 2026-01-19
Add support for continuous stakers

rrazvan1:

### DIFF
--- a/vms/platformvm/state/diff_test.go
+++ b/vms/platformvm/state/diff_test.go
@@ -1180,8 +1180,7 @@ func TestDiffUpdateValidator(t *testing.T) {
 	mutatedValidator = *currentValidator
 	mutatedValidator.Weight = 45
 
-	err = d.UpdateCurrentValidator(&mutatedValidator)
-	require.NoError(err)
+	require.NoError(d.UpdateCurrentValidator(&mutatedValidator))
 
 	currentValidator, err = d.GetCurrentValidator(subnetID, validator.NodeID)
 	require.NoError(err)


vms/platformvm/block/executor/proposal_block_test.go
```diff
@@ -107,7 +107,6 @@ func TestApricotProposalBlockTimeVerification(t *testing.T) {
 	)
 	onParentAccept.EXPECT().GetTx(addValTx.ID()).Return(addValTx, status.Committed, nil)
 	onParentAccept.EXPECT().GetCurrentSupply(constants.PrimaryNetworkID).Return(uint64(1000), nil).AnyTimes()
-	onParentAccept.EXPECT().GetDelegateeReward(constants.PrimaryNetworkID, utx.NodeID()).Return(uint64(0), nil).AnyTimes()
```
yacovm, line 110: why is this removed?

vms/platformvm/block/executor/proposal_block_test.go
```diff
@@ -221,8 +220,6 @@ func TestBanffProposalBlockTimeVerification(t *testing.T) {
 	onParentAccept.EXPECT().GetActiveL1ValidatorsIterator().Return(iterator.Empty[state.L1Validator]{}, nil).AnyTimes()
 	onParentAccept.EXPECT().GetExpiryIterator().Return(iterator.Empty[state.ExpiryEntry]{}, nil).AnyTimes()
 
-	onParentAccept.EXPECT().GetDelegateeReward(constants.PrimaryNetworkID, unsignedNextStakerTx.NodeID()).Return(uint64(0), nil).AnyTimes()
```
yacovm, line 224: ditto

vms/platformvm/state/staker.go
```diff
@@ -74,24 +103,93 @@ func (s *Staker) Less(than *Staker) bool {
 	return bytes.Compare(s.TxID[:], than.TxID[:]) == -1
 }
 
-func NewCurrentStaker(
+func NewCurrentValidator(
```
yacovm, line 106: why the rename?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Fee this validator charges delegators as a percentage, times 10,000
+	// For example, if this validator has DelegationShares=300,000 then they
+	// take 30% of rewards from delegators
+	DelegationShares uint32 `serialize:"true" json:"shares"`
+
+	// Weight of this validator used when sampling
+	Wght uint64 `serialize:"true" json:"weight"`
+
+	// Percentage of rewards to auto-restake at the end of each cycle, expressed in millionths (percentage * 10,000).
```
yacovm, line 66: For my education - Why do we need such a fine grained accuracy?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+		return fmt.Errorf("%w: weight %d != stake %d", errValidatorWeightMismatch, tx.Wght, totalStakeWeight)
+	}
+
+	// cache that this is valid
+	tx.SyntacticallyVerified = true
+	return nil
+}
+
+func (*AddContinuousValidatorTx) Visit(_ Visitor) error {
+	// TODO ACP-236: implement properly
```
yacovm, line 210: In which PR is this implemented?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+	tx.ValidatorRewardsOwner.InitCtx(ctx)
+	tx.DelegatorRewardsOwner.InitCtx(ctx)
+}
+
+func (*AddContinuousValidatorTx) SubnetID() ids.ID {
+	return constants.PrimaryNetworkID
+}
+
+func (tx *AddContinuousValidatorTx) PublicKey() (*bls.PublicKey, bool, error) {
+	if err := tx.Signer.Verify(); err != nil {
```
yacovm, line 99: This object seems immutable to me. 

Why do we need to verify this each time? 

Can't we verify it before we create the object or when we create it?

vms/platformvm/state/staker.go
```diff
@@ -120,6 +218,10 @@ func NewPendingStaker(txID ids.ID, staker txs.ScheduledStaker) (*Staker, error)
 }
 
 func (s *Staker) ValidateMutation(ms *Staker) error {
```
yacovm, line 220: The most secure way to validate any mutation is to:

- Compute the expected result
- Check if the expected result matches the proposed value.

Can we therefore take `s`, copy it into `s'`, and then apply mutations of `s'` and then compare `s'` to `ms`?

rrazvan1:
replaced by #4931

PR #4882, opened 2026-01-19
refactor(graft/evm): extract unified `message` package from `coreth` and `subnet-evm`

powerslider:
## Why this should be merged

Check #4880 

## How this works

Move the message package from `graft/coreth/plugin/evm/message` and `graft/subnet-evm/plugin/evm/message` to a shared location at `graft/evm/message` while maintaining backward compatibility for both
networks.

- Create dual codec support: `CorethCodec` (`SkipRegistrations=3`) and `SubnetEVMCodec` (`SkipRegistrations=2`) to preserve wire format compatibility.
- Unified `LeafsRequest` with `NodeType` field that is NOT serialized (`subnet-evm` style) to maintain backward compatibility.
- Remove duplicate message packages from both `coreth` and `subnet-evm`.

This consolidation reduces code duplication while preserving the network-specific codec configurations required for state sync protocol.

## How this was tested

existing UT with support for running for both `coreth` codec and `subnet-evm` codec

no 

resolves https://github.com/ava-labs/avalanchego/issues/4880

### DIFF
--- a/graft/evm/message/codec.go
+++ b/graft/evm/message/codec.go
@@ -13,10 +13,6 @@ import (
 const (
 	Version        = uint16(0)
 	maxMessageSize = 2*units.MiB - 64*units.KiB // Subtract 64 KiB from p2p network cap to leave room for encoding overhead from AvalancheGo
-
-	// warpSkipRegistrations is the number of deprecated Warp request/response types to skip
-	// See https://github.com/ava-labs/coreth/pull/999
-	warpSkipRegistrations = 3
 )
 
 // CodecType identifies the codec variant to use for message serialization.
@@ -78,7 +74,7 @@ func newCodec(codecType CodecType) codec.Manager {
 	// Skip 3 registrations for deprecated Warp types:
 	// MessageSignatureRequest, BlockSignatureRequest, SignatureResponse
 	// See https://github.com/ava-labs/coreth/pull/999
-	c.SkipRegistrations(warpSkipRegistrations)
+	c.SkipRegistrations(3)
 
 	errs.Add(mgr.RegisterCodec(Version, c))
 


graft/coreth/sync/handlers/block_request_test.go
```diff
@@ -92,7 +94,7 @@ func executeBlockRequestTest(t testing.TB, test blockRequestTest, blocks []*type
 	require.NotEmpty(t, responseBytes)
 
 	var response message.BlockResponse
-	_, err = message.Codec.Unmarshal(responseBytes, &response)
```
ceyonur, line 95: why don't we just use `message.CorethCodec`, are we planning to merge these as well?

graft/coreth/sync/block/syncer_test.go
```diff
@@ -107,38 +108,43 @@ func TestBlockSyncer_ParameterizedTests(t *testing.T) {
 		},
 	}
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			t.Parallel()
-			env := newTestEnvironment(t, tt.numBlocks)
-			require.NoError(t, env.prePopulateBlocks(tt.prePopulateBlocks))
+	utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
```
ceyonur: Why are we running these with two codecs here?

graft/coreth/sync/block/syncer_test.go
```diff
@@ -107,38 +108,43 @@ func TestBlockSyncer_ParameterizedTests(t *testing.T) {
...
-	env := newTestEnvironment(t, 10)
-	syncer, err := env.createSyncer(5, 3)
-	require.NoError(t, err)
 
-	// Immediately cancel the context to simulate cancellation.
-	ctx, cancel := context.WithCancel(t.Context())
-	cancel()
-	err = syncer.Sync(ctx)
-	require.ErrorIs(t, err, context.Canceled)
+	utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
```
ceyonur: I think it's probably best to keep using `message.CorethCodec` and `message.SubnetEVMCodec` in respective packages until we merge these, or really in a separate PR.

graft/coreth/sync/handlers/leafs_request.go
```diff
@@ -97,14 +102,14 @@ func (lrh *leafsRequestHandler) OnLeafsRequest(ctx context.Context, nodeID ids.N
 	// as this information will be necessary to access storage tries when
 	// the trie is path based.
 	// stateRoot := common.Hash{}
-	t, err := trie.New(trie.TrieID(leafsRequest.Root), lrh.trieDB)
+	t, err := trie.New(trie.TrieID(root), lrh.trieDB)
 	if err != nil {
-		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", leafsRequest.Root, "err", err)
+		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", leafsRequest.RootHash(), "err", err)
```
ceyonur:
```suggestion
		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", root, "err", err)
```

graft/coreth/sync/handlers/leafs_request.go
```diff
@@ -155,7 +160,7 @@ func (lrh *leafsRequestHandler) OnLeafsRequest(ctx context.Context, nodeID ids.N
 }
 
 type responseBuilder struct {
-	request   *message.LeafsRequest
```
ceyonur, line 158: why have we changed this?

graft/evm/message/codec.go
```diff
@@ -0,0 +1,74 @@
...
+		c.RegisterType(BlockRequest{}),
+		c.RegisterType(BlockResponse{}),
+		c.RegisterType(leafsRequest),
+		c.RegisterType(LeafsResponse{}),
+		c.RegisterType(CodeRequest{}),
+		c.RegisterType(CodeResponse{}),
+	)
+
+	// Deprecated Warp request/response types are skipped
+	c.SkipRegistrations(warpSkipRegistrations)
```
ceyonur: do we need this to be a variable? I think in both codec we skip by 3

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+
+// LeafsRequest defines the interface for leaf sync requests.
+type LeafsRequest interface {
+	Request
+	RootHash() common.Hash
+	AccountHash() common.Hash
+	StartKey() []byte
+	EndKey() []byte
+	LimitValue() uint16
+	NodeTypeValue() NodeType
```
ceyonur: nit
```suggestion
	NodeType() NodeType
```

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// LeafsRequestType selects which wire format to use when building leafs requests.
+type LeafsRequestType string
```
ceyonur: why does it have to be a string?

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+// SubnetEVMLeafsRequest preserves the original subnet-evm wire format where NodeType is not serialized.
+// NOTE: NodeType is not serialized to maintain backward compatibility with subnet-evm nodes.
+type SubnetEVMLeafsRequest struct {
+	Root     common.Hash `serialize:"true"`
+	Account  common.Hash `serialize:"true"`
+	Start    []byte      `serialize:"true"`
+	End      []byte      `serialize:"true"`
+	Limit    uint16      `serialize:"true"`
```
ceyonur: I wonder if we can use embedded type for these fields

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+	case SubnetEVMLeafsRequestType:
+		return SubnetEVMLeafsRequest{
+			Root:     root,
+			Account:  account,
+			Start:    start,
+			End:      end,
+			Limit:    limit,
+			NodeType: nodeType,
+		}
+	default:
```
ceyonur: Can we instead return error as it's not expected to have any other input? This would hinder any bug/weird path if we provide other than these two inputs.

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+			Account:  account,
+			Start:    start,
+			End:      end,
+			Limit:    limit,
+			NodeType: nodeType,
+		}
+	}
+}
+
+func NewEmptyLeafsRequest(leafReqType LeafsRequestType) LeafsRequest {
```
ceyonur: We should probably not export this.

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+	switch leafReqType {
+	case SubnetEVMLeafsRequestType:
+		return SubnetEVMLeafsRequest{}
+	default:
+		return CorethLeafsRequest{}
+	}
+}
+
+func LeafsRequestTypeForCodec(c codec.Manager) LeafsRequestType {
```
ceyonur: This is weird as an exported function, as whoever calling this should already be aware of their codec.

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+	return formatLeafsRequest("LeafsRequest", l.Root, l.Account, l.Start, l.End, l.Limit, l.NodeType)
+}
+
+func (l SubnetEVMLeafsRequest) Handle(ctx context.Context, nodeID ids.NodeID, requestID uint32, handler RequestHandler) ([]byte, error) {
+	return handler.HandleLeafsRequest(ctx, nodeID, requestID, l)
+}
+
+func (l CorethLeafsRequest) Handle(ctx context.Context, nodeID ids.NodeID, requestID uint32, handler RequestHandler) ([]byte, error) {
+	return handler.HandleLeafsRequest(ctx, nodeID, requestID, l)
+}
```
ceyonur: If these duplication are unavoidable can we separate those into two files?

Otherwise an embeded common type for common fields would probably remove the duplication here. But we should definetly check if the codec changes the behaviour for embedded types.

graft/evm/message/block_sync_summary_parser.go
```diff
@@ -8,18 +8,21 @@ import (
...
 
-type BlockSyncSummaryParser struct{}
+type BlockSyncSummaryParser struct {
+	codec codec.Manager
+}
 
-func NewBlockSyncSummaryParser() *BlockSyncSummaryParser {
-	return &BlockSyncSummaryParser{}
+func NewBlockSyncSummaryParser(c codec.Manager) *BlockSyncSummaryParser {
+	return &BlockSyncSummaryParser{codec: c}
```
ceyonur: I wonder if we should just merge this type with the `BlockSyncSummaryProvider`

graft/coreth/sync/leaf/syncer.go
```diff
@@ -98,13 +100,15 @@ func (c *CallbackSyncer) syncTask(ctx context.Context, task SyncTask) error {
...
-			Start:    start,
-			Limit:    c.config.RequestSize,
-			NodeType: task.NodeType(),
-		})
+		leafsResponse, err := c.client.GetLeafs(ctx, message.NewLeafsRequest(
+			c.config.LeafsRequestType,
+			root,
+			task.Account(),
+			start,
+			end,
```
ceyonur: This looks sus, we have not used any `End` field here previously.

graft/coreth/plugin/evm/atomic/sync/summary.go
```diff
@@ -28,7 +29,7 @@ type Summary struct {
 	acceptImpl message.AcceptImplFn
 }
 
-func NewSummary(blockHash common.Hash, blockNumber uint64, blockRoot common.Hash, atomicRoot common.Hash) (*Summary, error) {
+func NewSummary(c codec.Manager, blockHash common.Hash, blockNumber uint64, blockRoot common.Hash, atomicRoot common.Hash) (*Summary, error) {
```
ceyonur: I think it's a bit weird to take the codec as a param here because:

1- This is really a not shared package
2- We sometimes take codec as a param but sometimes use `CorethCodec` or `SubnetEVMCodec`. We should align how to use them.
3- It increases the diff and needs an extra tracking for where we actually pass the codec.

JonathanOppenheimer:
@powerslider Before reviewing this fully, @ceyonur I discussed serializing `LeafsRequest` in subnet-evm, rather than this sort of "workaround" I'll call it. This would need to be during a network upgrade (ecause it is a breaking change and nodes would need to know how to parse the new format. 

I think this would be fairly easy to implement -- we'd create a new, single codec, bump the codec version to v2, serialize all the appropriate fields, and tag serialization with v2. This would allow us to then delete a LOT of code, and ensure that coreth and subnet-evm are passing the same messages (which I think is valuable).

What are your thoughts on this idea? This may be a good fix in in the short term, but longer term, I'm not a huge fan of all this infrastructure to do things with two codecs.

powerslider:
@JonathanOppenheimer this is a great idea and definitely the best way to move forward, but I need this to be merged now so that is why I made it in a way to ensure 100% compatibility. If we want to skip running tests for both codecs, I can just test the `CorethCodec`, but we still need to have a `SubnetEVMCodec` as well to not cause any issues. This PR is a crucial step towards rolling out state sync changes to both `coreth` and `subnet-evm`, so that is why I need it now.

graft/coreth/sync/leaf/syncer.go
JonathanOppenheimer: Yeah something is up here -- previously: neither coreth nor subnet-evm populated the End field in LeafsRequest (even though it existed in the struct and was serialized), and now it's sent. The comment was:

```go
// Note: We truncate the response here as opposed to sending End
// in the request, as [VerifyRangeProof] does not handle empty
// responses correctly with a non-empty end key for the range.
```

JonathanOppenheimer: I think either you revert the behavior or determine if `VerifyRangeProof` has been fixed/fix it here.

graft/coreth/plugin/evm/atomic/sync/summary_parser.go
```diff
@@ -22,7 +22,7 @@ func NewSummaryParser() *summaryParser {
 
 func (*summaryParser) Parse(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
 	summary := Summary{}
-	if _, err := message.Codec.Unmarshal(summaryBytes, &summary); err != nil {
+	if _, err := message.CorethCodec.Unmarshal(summaryBytes, &summary); err != nil {
```
JonathanOppenheimer: As a random comment (and should be applied throughout this PR). I've been trying to use c-chain vs l1, as the distinguishers in my consolidation PRs, instead of coreth vs subnet-evm (as those repos don't exist anymore, and we want to eventually get rid of all the graft). what are your thoughts?

graft/evm/message/leafs_request.go
JonathanOppenheimer: Assuming the serialization isn't changed, something like 

```go
type leafsRequestCommon struct {
    Root     common.Hash `serialize:"true"`
    Account  common.Hash `serialize:"true"`
    Start    []byte      `serialize:"true"`
    End      []byte      `serialize:"true"`
    Limit    uint16      `serialize:"true"`
    NodeType NodeType
}

type SubnetEVMLeafsRequest struct {
    leafsRequestCommon
}

type CorethLeafsRequest struct {
    leafsRequestCommon
    NodeType NodeType `serialize:"true"` // Override to serialize
}
```

could be solid

JonathanOppenheimer: a lot cleaner!

graft/evm/message/leafs_request.go
JonathanOppenheimer: I think an embedded type to reduce all this duplication would be a lot better - this just seems wrong, and I think even in separate files, it would still feel wrong.

graft/evm/message/leafs_request_test.go
```diff
@@ -0,0 +1,217 @@
...
+	// set random seed for deterministic random
+	r := newTestRand()
+	startBytes := randomBytes(t, r, common.HashLength)
+	endBytes := randomBytes(t, r, common.HashLength)
+	expectedByType := map[message.LeafsRequestType]string{
+		message.CorethLeafsRequestType:    corethLeafsRequestB64,
+		message.SubnetEVMLeafsRequestType: subnetEVMLeafsRequestB64,
+	}
+
+	forEachMessageFormat(t, func(t *testing.T, _ string, format messageFormat) {
```
JonathanOppenheimer: You use this function and `utilstest.ForEachCodec` inconsistently, but I think Cey already raised a larger issue with the for each codec strategy.

graft/evm/message/leafs_request.go
JonathanOppenheimer: I feel like this should be `iota` or something.

graft/evm/message/leafs_request.go
JonathanOppenheimer: Yeah this is strange behavior for 

```go
req := message.NewLeafsRequest("invalid-type", ...)
```

Whatever strategy is settled upon, it should be tested, which I do not think is the case right now.

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+// LeafsRequest defines the interface for leaf sync requests.
+type LeafsRequest interface {
+	Request
+	RootHash() common.Hash
+	AccountHash() common.Hash
+	StartKey() []byte
+	EndKey() []byte
+	LimitValue() uint16
```
JonathanOppenheimer:
```suggestion
	Limit() uint16
```

JonathanOppenheimer: this one too imo

graft/evm/message/leafs_request.go
```diff
@@ -0,0 +1,177 @@
...
+	default:
+		return CorethLeafsRequest{}
+	}
+}
+
+func LeafsRequestTypeForCodec(c codec.Manager) LeafsRequestType {
+	switch c {
+	case SubnetEVMCodec:
+		return SubnetEVMLeafsRequestType
+	default:
```
JonathanOppenheimer: Same very strange behavior here.

JonathanOppenheimer: and yeah honestly, I think this entire function should be removed entirely -- it's not necessary.

graft/evm/message/codec.go
```diff
@@ -0,0 +1,74 @@
+// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package message
```
JonathanOppenheimer, line 4: This could probably use a godoc

graft/evm/message/leafs_request.go
powerslider: if we make it private, then yes

graft/evm/message/leafs_request.go
powerslider: We still need the `Value` suffix as-is unless we rename the underlying fields. Go doesn’t allow a method and field with the same name on a struct (`Limit/NodeType`), so dropping the suffix causes compile errors.

graft/evm/message/leafs_request.go
powerslider: We still need the `Value` suffix as-is unless we rename the underlying fields. Go doesn’t allow a method and field with the same name on a struct (`Limit/NodeType`), so dropping the suffix causes compile errors.

graft/evm/message/leafs_request.go
powerslider: @JonathanOppenheimer avoid words like `common`, `util`, etc., because they communicate nothing and are considered a bad practice in Go especially.

graft/evm/message/leafs_request.go
JonathanOppenheimer: gotcha my apologies!

graft/evm/message/leafs_request.go
JonathanOppenheimer: something like `base` or `shared` then?

graft/coreth/sync/handlers/leafs_request.go
powerslider: `LeafsRequest` is now an interface (to support coreth and subnet-evm wire formats). Pointers to interfaces are unnecessary in Go since interface values already contain a pointer internally.

graft/evm/message/leafs_request.go
powerslider: No strong reason. `iota` works equally well here since the values are only used as type discriminators in switch statements, not for serialization or display. Happy to change it. I usually prefer string constants for better clarity, but `iota` would fit well here.

graft/evm/message/leafs_request.go
powerslider: I used `base`

graft/coreth/sync/leaf/syncer.go
powerslider: Good catch. The old code intentionally omitted `End` from the request to work around the `VerifyRangeProof` bug mentioned in the comment. I'll revert to passing `nil` for `end` to preserve the existing behavior - the client-side truncation logic handles this correctly.

graft/evm/message/codec.go
powerslider: it's a constant and I think it serves as better documentation to know what the value actually means.

graft/evm/message/leafs_request_test.go
powerslider: `ForEachCodec` only provides the codec, but these tests need both the codec and the corresponding `LeafsRequestType` to call `NewLeafsRequest()` and validate type-specific serialization behavior. Changing `ForEachCodec` to also pass `LeafsRequestType` would break all existing call sites that only need the codec. Keeping `forEachMessageFormat` as a local helper bundles the related codec and request type together without affecting other code.

What Cey mentioned was related to the state sync code. The message package will run all its tests for both codecs even in this PR.

UPDATE: it is no longer needed due to other changes. So it is gone at this point.

graft/coreth/plugin/evm/atomic/sync/summary_provider.go
```diff
@@ -45,3 +45,16 @@ func (a *SummaryProvider) StateSummaryAtBlock(blk *ethtypes.Block) (block.StateS
 	}
 	return summary, nil
 }
+
+func (*summaryParser) Parse(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
```
alarso16: Why did this method move? Seems more intuitive to keep it

graft/coreth/sync/evmstate/sync_test.go
```diff
@@ -240,7 +239,6 @@ func (i *interruptLeafsIntercept) getLeafsIntercept(request message.LeafsRequest
 }
 
 func TestResumeSyncAccountsTrieInterrupted(t *testing.T) {
```
alarso16, line 242: Did you remove the parallelism because the new codec abstraction caused a race? It's fine either way, just checking

graft/evm/message/codec.go
```diff
@@ -0,0 +1,83 @@
...
+	// SubnetEVMCodec is used for Subnet-EVM state sync messages
+	SubnetEVMCodec codec.Manager
+)
+
+func init() {
+	CorethCodec = newCodec(corethSkipRegistrations, CorethLeafsRequestType)
+	SubnetEVMCodec = newCodec(subnetEVMSkipRegistrations, SubnetEVMLeafsRequestType)
+}
+
+func newCodec(skipCount int, leafsReqType LeafsRequestType) codec.Manager {
```
alarso16: Why not just take in one field as an enum? I wouldn't ever do:
```
someWeirdCodec := newCodec(corethSkipRegistrations, SubnetEVMLeafsRequestType)
```

graft/coreth/plugin/evm/vm_test.go
```diff
@@ -88,9 +88,10 @@ var (
 )
 
 func defaultExtensions() *extension.Config {
+	syncSummaryProvider := message.NewBlockSyncSummaryProvider(message.CorethCodec)
 	return &extension.Config{
-		SyncSummaryProvider: &message.BlockSyncSummaryProvider{},
-		SyncableParser:      &message.BlockSyncSummaryParser{},
+		SyncSummaryProvider: syncSummaryProvider,
```
alarso16: This config is only used by coreth, so I think it would be reasonable to change and make as a single `SyncSummaryProvider` with the interfaces joined. Is it necessary for testing or something?

graft/coreth/sync/client/client_test.go
```diff
@@ -780,3 +816,20 @@ func TestStateSyncNodes(t *testing.T) {
 	require.Contains(t, testNetClient.nodesRequested, stateSyncNodes[2])
 	require.Contains(t, testNetClient.nodesRequested, stateSyncNodes[3])
 }
+
+func newLeafsRequest(
```
alarso16, line 822: Will you write a short doc for what this does? Specifically that setting `account = common.Hash{}` equates to an account trie request

graft/coreth/sync/handlers/code_request_test.go
```diff
@@ -85,28 +85,32 @@ func TestCodeRequestHandler(t *testing.T) {
 		},
 	}
 
-	for name, test := range tests {
-		// Reset stats before each test
-		testHandlerStats.Reset()
+	utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
```
alarso16: Are we running with both codecs to make the transition to a single sync package easier?

graft/coreth/sync/handlers/leafs_request.go
```diff
@@ -97,14 +102,14 @@ func (lrh *leafsRequestHandler) OnLeafsRequest(ctx context.Context, nodeID ids.N
...
+	t, err := trie.New(trie.TrieID(root), lrh.trieDB)
 	if err != nil {
-		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", leafsRequest.Root, "err", err)
+		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", root, "err", err)
 		lrh.stats.IncMissingRoot()
 		return nil, nil
 	}
 	// override limit if it is greater than the configured maxLeavesLimit
-	limit := leafsRequest.Limit
+	limit := limitValue
```
alarso16: Why not just name it `limit` to start with?

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -709,3 +760,17 @@ func requireRangeProofIsValid(t *testing.T, request *message.LeafsRequest, respo
 	require.NoError(t, err)
 	require.Equal(t, expectMore, more)
 }
+
+func newLeafsRequest(
```
alarso16, line 750: I'm not sure wrapping the ctx in this call is productive, since it just passes through to the return value

graft/evm/message/code_request_test.go
```diff
@@ -42,17 +48,19 @@ func TestMarshalCodeResponse(t *testing.T) {
...
-	codeResponse := CodeResponse{
+	codeResponse := message.CodeResponse{
 		Data: [][]byte{codeData},
 	}
 
 	base64CodeResponse := "AAAAAAABAAAAMlL9/AchgmVPFj9fD5piHXKVZsdNEAN8TXu7BAfR4sZJgYVa2GgdDYbR6R4AFnk5y2aU"
-	codeResponseBytes, err := Codec.Marshal(Version, codeResponse)
-	require.NoError(t, err)
-	require.Equal(t, base64CodeResponse, base64.StdEncoding.EncodeToString(codeResponseBytes))
+	utilstest.ForEachCodec(t, func(_ string, cdc codec.Manager) {
```
alarso16: I think `cdc` is a bad name. Just rename `c` below to `resp` or something

graft/evm/message/leafs_request.go
alarso16: Does renaming the field cause changes in serialization? If not, they shouldn't be exported anyway. Otherwise, what if you call it `KeyLimit()` and `LeafType()`

graft/coreth/sync/evmstate/sync_test.go
powerslider: I'll restore it, because I reverted the `utilstest.ForEachCodec` which could not have the `t.Parallel` at that place, but I'll roll that out in a follow up PR.

graft/coreth/sync/handlers/code_request_test.go
powerslider: I missed to revert this, will introduce the dual codec support in a follow up PR.

graft/coreth/sync/handlers/leafs_request_test.go
powerslider: This is the cleaner solution, because of the way we parameterize the test cases. This results in the smallest diff and also encapsulates returning an error for an unknown leafs request type.

graft/coreth/plugin/evm/atomic/sync/summary_provider.go
powerslider: mistake from my side, will fix

powerslider:
@alarso16  responding to https://github.com/ava-labs/avalanchego/pull/4882#discussion_r2738521076. 

In Go, when you embed a struct, its fields are promoted to the outer struct. So `CorethLeafsRequest` would have a field named `Limit`. But you can't have both a field named `Limit` and a method named `Limit()` on the same type - the method would shadow the field or cause a compilation error.

```
// The embedded base struct had fields:
type LeafsRequestBase struct {
    Limit    uint16
    NodeType NodeType
    ...
}

// The interface had methods:
type LeafsRequest interface {
    Limit() uint16      // COLLISION - can't have method named "Limit" 
    NodeType() NodeType // COLLISION - can't have method named "NodeType"
    ...
}

```

So I had to use `LimitValue()` and `NodeTypeValue()` to avoid colliding with the field names `Limit` and `NodeType`.

Now with duplicated structs (no embedding), we still have the same field names, but we could rename the methods to `KeyLimit()` and `LeafType()` because:
- `KeyLimit()` doesn't collide with field `Limit`
- `LeafType()` doesn't collide with field `NodeType`

So now it makes sense to do it. Check here https://github.com/ava-labs/avalanchego/pull/4882/changes/17f848fe6362425d15ce11e61904aa28a68febcb#diff-6d22b8aa4b0f9895347bcab8a65a86035e47cfc02c443a8e7f0edafca8cb2a07L36

alarso16:
> In Go, when you embed a struct, its fields are promoted to the outer struct.

This I understand. However, why not just change the name of the field from `Limit` to `limit`?

JonathanOppenheimer:
> @JonathanOppenheimer this is a great idea and definitely the best way to move forward, but I need this to be merged now so that is why I made it in a way to ensure 100% compatibility. If we want to skip running tests for both codecs, I can just test the `CorethCodec`, but we still need to have a `SubnetEVMCodec` as well to not cause any issues. This PR is a crucial step towards rolling out state sync changes to both `coreth` and `subnet-evm`, so that is why I need it now.

Could we put some sort of longer TODO comment explaining how to do this far cleaner / "properly" with breaking changes somewhere in the code so the context isn't lost? You know the approach extremely well at this point.

graft/evm/message/codec.go
```diff
@@ -0,0 +1,97 @@
...
+	// and uses SubnetEVMLeafsRequest wire format.
+	SubnetEVMCodecType
+)
+
+var (
+	// CorethCodec is used for C-Chain state sync messages
+	CorethCodec codec.Manager
+	// SubnetEVMCodec is used for Subnet-EVM state sync messages
+	SubnetEVMCodec codec.Manager
+)
```
JonathanOppenheimer, line 34: This is a higher level comment -- I received feedback from @joshua-kim when I did a very similar thing (combining configs into a single file in graft/evm) -- this abstraction does not really make sense for both the subnet and coreth codec -- the package division should not be along what seems similar to each other, but what should actually be seperate and I think the feedback does ring true here. Why does the subnet and coreth codecs need to know about each other? 

The 'correct' approach would be to separate out the coreth and subnet-evm codec that both use the shared codec items (`newCodec` or something) separately in their own package 

I think when the goal is try to forcibly combine items, it's not avoidable, but that pattern is in my head. Not asking for changes, just your thoughts on this

JonathanOppenheimer: You even sort of do this with the leafs request items already.

graft/evm/message/coreth_leafs_request.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c CorethLeafsRequest) Handle(ctx context.Context, nodeID ids.NodeID, requestID uint32, handler RequestHandler) ([]byte, error) {
+	return handler.HandleLeafsRequest(ctx, nodeID, requestID, c)
+}
+
+func (c CorethLeafsRequest) String() string {
+	return fmt.Sprintf(
+		"LeafsRequest(Root=%s, Account=%s, Start=%s, End=%s, Limit=%d, NodeType=%d)",
+		c.Root, c.Account, common.Bytes2Hex(c.Start), common.Bytes2Hex(c.End), c.Limit, c.NodeType,
```
JonathanOppenheimer, line 42: Why do we populate `End` (here and in subnet-evm), but it's never used as it's always passed as nil in `NewLeafsRequest`

JonathanOppenheimer:
> > In Go, when you embed a struct, its fields are promoted to the outer struct.
> 
> This I understand. However, why not just change the name of the field from `Limit` to `limit`?

The field has to be serialized. Can't un-export it.

graft/evm/message/coreth_leafs_request.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+// CorethLeafsRequest preserves the original coreth wire format where NodeType is serialized.
+//
+// Note: The fields Root, Account, Start, End, and Limit are duplicated in SubnetEVMLeafsRequest.
+// This duplication is intentional because the avalanchego codec requires all serialized fields
+// to be exported (start with uppercase). Using an embedded struct would require exporting it,
+// which would unnecessarily expose internal types in the public API. By duplicating fields,
+// we keep the implementation details private while ensuring correct serialization.
```
JonathanOppenheimer, line 21: This comment seems incorrect to me. I am confident we can do: 

```go
  type LeafsRequestBase struct {                                                                                                                                           
      Root     common.Hash `serialize:"true"`                                                                                                                                
      Account  common.Hash `serialize:"true"`                                                                                                                                
      Start    []byte      `serialize:"true"`                                                                                                                                
      End      []byte      `serialize:"true"`                                                                                                                                
      Limit    uint16      `serialize:"true"`                                                                                                                                
      NodeType NodeType    // NOT serialized                                                                                                                                 
  }                                                                                                                                                                          
                                                                                                                                                                             
  type CorethLeafsRequest struct {                                                                                                                                           
      LeafsRequestBase     `serialize:"true"`                                                                                                                                                                            
      NodeType NodeType `serialize:"true"` // Override to serialize                                                                                                          
  }                                                                                                                                                                          
                                                                                                                                                                             
  type SubnetEVMLeafsRequest struct {                                                                                                                                        
      LeafsRequestBase // NodeType not serialized                                                                                                                          
  }
```

I think a better counterargument is that we should split these into separate packages (as I commented in another comment). I then don't see a huge problem with exposing the base struct - as you would not "needlessly" be exporting them, but also if you wanted to just keep them duplicated that would be fine. I think the middle path (this) is probably worse than either of the two above approaches.

graft/evm/message/leafs_request.go
```diff
@@ -24,28 +26,49 @@ const (
...
+	case CorethLeafsRequestType:
+		return CorethLeafsRequest{
+			Root:     root,
+			Account:  account,
+			Start:    start,
+			End:      end,
+			Limit:    limit,
+			NodeType: nodeType,
+		}, nil
+	default:
```
JonathanOppenheimer, line 69: Could we still test this with invalid input? The function silently returns nil interface + error, which could cause nil pointer panics if callers don't check

graft/evm/message/coreth_leafs_request.go
powerslider: Because this is a `String` method and should print all fields of the struct. The value depends on what you pass.

graft/coreth/plugin/evm/atomic/sync/summary_provider.go
```diff
@@ -45,3 +45,17 @@ func (a *SummaryProvider) StateSummaryAtBlock(blk *ethtypes.Block) (block.StateS
 	}
 	return summary, nil
 }
+
+// Parse parses the summary bytes into a Syncable summary.
+func (*SummaryProvider) Parse(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
+	summary := Summary{}
```
alarso16: nit:
```suggestion
	var summary Summary
```
Something in my gut is telling me you'll disagree, but will you explain why?

graft/coreth/sync/client/client.go
```diff
@@ -136,8 +136,9 @@ func parseLeafsResponse(codec codec.Manager, reqIntf message.Request, data []byt
 	leafsRequest := reqIntf.(message.LeafsRequest)
 
 	// Ensure the response does not contain more than the maximum requested number of leaves.
-	if len(leafsResponse.Keys) > int(leafsRequest.Limit) || len(leafsResponse.Vals) > int(leafsRequest.Limit) {
-		return nil, 0, fmt.Errorf("%w: (%d) > %d)", errTooManyLeaves, len(leafsResponse.Keys), leafsRequest.Limit)
+	limitValue := leafsRequest.KeyLimit()
```
alarso16:
```suggestion
	limit:= leafsRequest.KeyLimit()
```
Is there another limit in scope?

graft/coreth/sync/handlers/leafs_request_test.go
alarso16: I disagree, but not strongly enough to block this from being merged

graft/evm/message/block_sync_summary_provider.go
```diff
@@ -0,0 +1,39 @@
...
+package message
+
+import (
+	"github.com/ava-labs/libevm/core/types"
+
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+)
+
+var _ SyncSummaryProvider = (*BlockSyncSummaryProvider)(nil)
```
alarso16, line 13: I'm not sure we should still be calling these `SummaryProvider`s if they now also parse. I don't have a better name, but maybe you do

graft/evm/message/leafs_request.go
alarso16: If the callers don't check the error, that's on them

graft/evm/utils/utilstest/codec.go
```diff
@@ -0,0 +1,25 @@
...
+
+// ForEachCodec runs fn as a subtest for each supported codec (coreth and subnet-evm).
+func ForEachCodec(t *testing.T, fn func(name string, c codec.Manager)) {
+	t.Helper()
+	codecs := map[string]codec.Manager{
+		"coreth":     message.CorethCodec,
+		"subnet-evm": message.SubnetEVMCodec,
+	}
+	for name, c := range codecs {
+		t.Run(name, func(_ *testing.T) {
```
alarso16: Won't the tracing be funky if `fn` doesn't take `t`?

If you disagree, then you should remove the underscore name (team convention)
```suggestion
		t.Run(name, func(*testing.T) {
```

graft/evm/utils/utilstest/codec.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ava-labs/avalanchego/graft/evm/message"
+)
+
+// ForEachCodec runs fn as a subtest for each supported codec (coreth and subnet-evm).
+func ForEachCodec(t *testing.T, fn func(name string, c codec.Manager)) {
```
alarso16:
```suggestion
func ForEachCodec(t *testing.T, fn func(string, codec.Manager)) {
```

I don't think a testing name needs to be passed through, but I think we need `t` (see the other comment)

graft/subnet-evm/sync/handlers/leafs_request.go
```diff
@@ -79,32 +79,39 @@ func (lrh *leafsRequestHandler) OnLeafsRequest(ctx context.Context, nodeID ids.N
...
 
-	if (len(leafsRequest.End) > 0 && bytes.Compare(leafsRequest.Start, leafsRequest.End) > 0) ||
-		leafsRequest.Root == (common.Hash{}) ||
-		leafsRequest.Root == types.EmptyRootHash ||
-		leafsRequest.Limit == 0 {
+	var (
+		startKey   = leafsRequest.StartKey()
+		endKey     = leafsRequest.EndKey()
+		root       = leafsRequest.RootHash()
+		limitValue = leafsRequest.KeyLimit()
```
alarso16:
```suggestion
		limit = leafsRequest.KeyLimit()
```

Ha I see you're doing a var block for declarations!

graft/subnet-evm/sync/handlers/leafs_request.go
```diff
@@ -79,32 +79,39 @@ func (lrh *leafsRequestHandler) OnLeafsRequest(ctx context.Context, nodeID ids.N
...
+	t, err := trie.New(trie.TrieID(root), lrh.trieDB)
 	if err != nil {
-		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", leafsRequest.Root, "err", err)
+		log.Debug("error opening trie when processing request, dropping request", "nodeID", nodeID, "requestID", requestID, "root", root, "err", err)
 		lrh.stats.IncMissingRoot()
 		return nil, nil
 	}
 	// override limit if it is greater than the configured maxLeavesLimit
-	limit := leafsRequest.Limit
+	limit := limitValue
```
alarso16:
```suggestion
```

graft/subnet-evm/sync/handlers/leafs_request_test.go
```diff
@@ -85,226 +85,252 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
+				return newLeafsRequest(t,
+					t.Context(),
+					common.Hash{},
+					common.Hash{},
+					bytes.Repeat([]byte{0x00}, common.HashLength+2),
+					bytes.Repeat([]byte{0xff}, common.HashLength),
+					maxLeavesLimit,
+					message.StateTrieNode,
+				)
+			},
```
alarso16:
```suggestion
				return t.Context(), newLeafsRequest(t,
					common.Hash{},
					common.Hash{},
					bytes.Repeat([]byte{0x00}, common.HashLength+2),
					bytes.Repeat([]byte{0xff}, common.HashLength),
					maxLeavesLimit,
					message.StateTrieNode,
				)
			},
```
I don't think this is any worse

alarso16: But it's fine, I don't care that much

graft/evm/message/coreth_leafs_request.go
alarso16: I believe the comment is correct. We can always adjust it more in the future (if you disagree with the comment, you can make a PR after this is merged), but this is certainly a step in the right direction

graft/evm/message/codec.go
alarso16: It's certainly not ideal, I agree. However, the codec must be global to coreth/subnet-evm, so you can't just "create it" when you need it. I think this is by far the simplest solution to unblock further de-duplication, and as we transition to the `p2p` package, a lot of this will become outdated

graft/coreth/plugin/evm/atomic/sync/summary_provider.go
powerslider: I do agree - var should be used for zero values ;)

graft/coreth/sync/handlers/leafs_request_test.go
powerslider: I changed it at the end, exactly how you wanted. Actually it did not impact the diff that much.

graft/subnet-evm/sync/handlers/leafs_request.go
powerslider: Following team conventions ;)

graft/evm/message/block_sync_summary_provider.go
powerslider: Honestly, I think the current name `SyncSummaryProvider` is acceptable. The interface provides everything needed to work with sync summaries - both creating them and parsing them. The term "provider" doesn't strictly mean "only creates" - it can mean "provides the capability to work with."

graft/evm/message/messagetest/codec.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ava-labs/avalanchego/graft/evm/message"
+)
+
+// ForEachCodec runs fn as a subtest for each supported codec (coreth and subnet-evm).
+func ForEachCodec(t *testing.T, fn func(string, codec.Manager)) {
```
alarso16, line 14:
```go
// ForEachCodec runs fn as a subtest for each supported codec (coreth and subnet-evm).
func ForEachCodec(t *testing.T, fn func(*testing.T, codec.Manager)) {
	t.Helper()
	codecs := map[string]codec.Manager{
		"coreth":     message.CorethCodec,
		"subnet-evm": message.SubnetEVMCodec,
	}
	for name, c := range codecs {
		t.Run(name, func(t *testing.T) {
			fn(t, c)
		})
	}
}
```

Does the test need to know which codec is running? I don't think it needs to. Also, aren't you supposed to pass in `t` to subtests?

graft/coreth/sync/client/client_test.go
```diff
@@ -151,7 +151,8 @@ func TestGetBlocks(t *testing.T) {
 	blocks, _, err := core.GenerateChain(params.TestChainConfig, genesis, engine, memdb, numBlocks, 0, func(_ int, _ *core.BlockGen) {})
 	require.NoError(t, err)
 	require.Len(t, blocks, numBlocks)
-	blocksRequestHandler := handlers.NewBlockRequestHandler(buildGetter(blocks), message.Codec, handlerstats.NewNoopHandlerStats())
+
```
ceyonur, line 154: nit: extra whitespace diff

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -75,236 +75,230 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
 	snapshotProvider := &TestSnapshotProvider{}
-	leafsHandler := NewLeafsRequestHandler(trieDB, message.StateTrieKeyLength, snapshotProvider, message.Codec, testHandlerStats)
 	snapConfig := snapshot.Config{
 		CacheSize:  64,
 		AsyncBuild: false,
 		NoBuild:    false,
 		SkipVerify: true,
 	}
 
+	leafsHandler := NewLeafsRequestHandler(trieDB, message.StateTrieKeyLength, snapshotProvider, message.CorethCodec, testHandlerStats)
```
ceyonur, line 87: nit: why have we moved this here?

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -75,236 +75,230 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
-					NodeType: message.StateTrieNode,
-				}
-			},
-			requireResponseFn: func(t *testing.T, _ message.LeafsRequest, response []byte, err error) {
-				require.Nil(t, response)
-				require.NoError(t, err)
-				require.Equal(t, uint32(1), testHandlerStats.InvalidLeafsRequestCount)
-			},
-		},
-		"empty storage root dropped": {
```
ceyonur, line 155: why have we removed this test case?

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -500,16 +505,19 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
-				return t.Context(), message.LeafsRequest{
-					Root:     accountTrieRoot,
-					Limit:    maxLeavesLimit,
-					NodeType: message.StateTrieNode,
-				}
+				return t.Context(), newLeafsRequest(t,
+					accountTrieRoot,
+					common.Hash{},
+					nil,
+					bytes.Repeat([]byte{0xff}, common.HashLength),
```
ceyonur: why we need an end now?

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -526,21 +534,24 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
-					Account:  largeStorageAccount,
-					Limit:    maxLeavesLimit,
-					NodeType: message.StateTrieNode,
-				}
+
+				return t.Context(), newLeafsRequest(t,
+					largeTrieRoot,
+					largeStorageAccount,
+					nil,
+					bytes.Repeat([]byte{0xff}, common.HashLength),
```
ceyonur: similarly why we need end?

graft/coreth/sync/handlers/leafs_request_test.go
```diff
@@ -647,18 +662,19 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
-					Account:  smallStorageAccount,
-					Start:    lastKey[:],
-					Limit:    maxLeavesLimit,
-					NodeType: message.StateTrieNode,
-				}
+				return t.Context(), newLeafsRequest(t,
+					smallTrieRoot,
+					smallStorageAccount,
+					lastKey[:],
+					bytes.Repeat([]byte{0xff}, common.HashLength),
```
ceyonur: again why changed end?

graft/evm/message/codec.go
ceyonur: These skips are ordered and we instead should comment why/what we're skipping in-line here.

graft/evm/message/codec.go
```diff
@@ -0,0 +1,97 @@
...
+		c.RegisterType(BlockRequest{}),
+		c.RegisterType(BlockResponse{}),
+	)
+
+	// Register the concrete leafs request type for the wire format.
+	// Must register concrete type, not interface, for codec to serialize correctly.
+	switch codecType {
+	case SubnetEVMCodecType:
+		errs.Add(c.RegisterType(SubnetEVMLeafsRequest{}))
+	default:
```
ceyonur: nit: "default" is kinda misleading here and it should be hardcoded to CorethCodecType. Although we don't have any other codecs, so this is semantically correct it's weird to see we're registering coreth. (we should probably panic/return err if codecType is different)

graft/evm/message/codec.go
```diff
@@ -0,0 +1,97 @@
...
+
+	errs := wrappers.Errs{}
+
+	// Skip registration to keep registeredTypes unchanged after legacy gossip deprecation
+	// Gossip types and sync summary type removed from codec
+	switch codecType {
+	case SubnetEVMCodecType:
+		// Skip 2 registrations for deprecated gossip types
+		c.SkipRegistrations(2)
+	default:
```
ceyonur, line 54: simirlay as below, this probably should just be check for Coreth type

graft/evm/message/codec.go
ceyonur: If it's much desired, I think we can still move global codecs to their respective vm package by exporting the `NewCodec` function. Not saying we should do it, but pointing it should be possible.

ceyonur: but if we prefer to go down that path, I'd suggest duplicating a bit of code (not more than what's duplicated here already) by moving respective leaf requests structs to vm packages

graft/evm/message/messagetest/codec.go
ceyonur: shouldn't this be in `message/messagetest`?

graft/evm/message/codec.go
JonathanOppenheimer: Do you have a preference Cey?

powerslider: Here is my take on this: 
- Yes, it would be better to be able to initialize them in `coreth` and `subnet-evm` respectably, but bare in mind that they are referred still in `graft/evm` in the state sync code. We have a strict policy to not import anything from `graft/coreth` and `graft/subnet-evm` into `graft/evm`. The dependency direction is allowed only in the opposite way.
- In order to do this properly we have to apply a ports and adapters approach (hexagonal architecture) - the same way I did the decoupling of `Blockchain`, `BlockProvider`, `SnapshotProvider` and `ChainContext` via interfaces in `graft/evm` and adapter implementations in `coreth` and `subnet-evm`. 
- That way I can call `CorethCodec()` and `SubnetEVMCodec()` as methods of that interface instead of global variables. 
- Having said that in the next network upgrade we will unify those codec wire formats and at that point the code will then have to be only in `graft/evm` and shared across `coreth` and `subnet-evm`.

My verdict would be to leave it as is for now (because it was the existing approach we already had) and potentially we can either refactor according to the way I described it a follow up PR or just delete one of them once we unify the wire format.

graft/coreth/sync/handlers/leafs_request_test.go
powerslider: By accident -> https://github.com/ava-labs/avalanchego/pull/4882/changes/3ccea0b5ea8a040abeef5d5f92bc50464451274e. I've restored it.

graft/coreth/sync/handlers/leafs_request_test.go
powerslider: This is intentional. The `leafsHandler` was moved out of the `ForEachCodec` closure because the test no longer needs to run for each codec - coreth tests now use `CorethCodec` directly.

graft/coreth/sync/handlers/leafs_request_test.go
powerslider: Mistake during refactoring. Before we inlined the structs and only initialized what we needed, now we use a test helper function to do that and I fixed it by passing nil. You can check here: https://github.com/ava-labs/avalanchego/pull/4882/changes/54728921d59b94619387316d945bc2775f2f1110

graft/evm/message/codec.go
powerslider: fix is here https://github.com/ava-labs/avalanchego/pull/4882/changes/a4231f12d7a4ad534775ad15f47c8e81977a5dd2

graft/evm/message/codec.go
ceyonur: >  but bare in mind that they are referred still in graft/evm in the state sync code

where? I think we're only using interfaces rather than direct references. 

As I said we don't need to do this if we say it's temporary.

powerslider: I've created an issue for unifying the message format in the following `Helicon` network upgrade: https://github.com/ava-labs/avalanchego/issues/4932. Feel free to express feedback on the description.

graft/evm/message/codec.go
JonathanOppenheimer: This is resolved now

PR #4884, opened 2026-01-20
refactor: remove strict mode in ABI

alarso16:
## Why this should be merged

This reduces the coupling between the `accounts/abi` package and the actual precompile implementation

## How this works

Perform length verification at call sites as necessary

## How this was tested

New UT

 No


graft/subnet-evm/accounts/abi/abi_extra_test.go
```diff
@@ -38,35 +38,27 @@ func TestUnpackInputIntoInterface(t *testing.T) {
...
 		{
-			name:                   "Valid input data with extra padding indivisible by 32",
-			extraPaddingBytes:      33,
-			strictMode:             true,
-			expectedErrorSubstring: "abi: improperly formatted input:",
+			name:              "Valid input data with extra padding indivisible by 32",
+			extraPaddingBytes: 33,
 		},
 		{
 			name:              "Valid input data with extra padding indivisible by 32, no strict mode",
```
JonathanOppenheimer: strict mode no longer exists

graft/subnet-evm/precompile/contracts/rewardmanager/contract.go
```diff
@@ -184,7 +185,15 @@ func PackSetRewardAddress(addr common.Address) ([]byte, error) {
...
-	res, err := RewardManagerABI.UnpackInput("setRewardAddress", input, useStrictMode)
+	// Solidity does not always pack the input to the correct length, and allows
+	// for extra padding bytes to be added to the end of the input. Therefore, we have removed
+	// this check with Durango. We still need to keep this check for backwards compatibility.
+	// However, as opposed to other precompiles, we only check that the length is divisible by 32,
+	// since historical execution didn't enforce any particular length.
+	if useStrictMode && len(input)%32 != 0 {
+		return common.Address{}, fmt.Errorf("%w: %d", ErrInvalidLen, len(input))
+	}
+	res, err := RewardManagerABI.UnpackInput("setRewardAddress", input)
```
JonathanOppenheimer, line 196: I assume the key differentiator here is that the `RewardManager` precompile was deployed pre-Durango while the others were not? Should this be noted?

graft/subnet-evm/accounts/abi/abi_extra_test.go
JonathanOppenheimer: in fact this test case is just duplicated now -- with the removal of strict mode the cases are the same.

JonathanOppenheimer: This is also the case in `graft/coreth/accounts/abi/abi_extra_test.go`

JonathanOppenheimer (changes requested):
A couple things

graft/subnet-evm/precompile/contracts/rewardmanager/contract.go
alarso16: I noted the reason this one is treated differently. I updated the comments for all locations. This one was deployed pre-Durango, but relied on the internal `useStrictMode` feature to verify length. Is this unclear from the comments?

graft/subnet-evm/accounts/abi/abi_extra_test.go
alarso16: Ah this one is incorrect (now fixed). However, I believe that all test cases in coreth are unique

graft/subnet-evm/precompile/contracts/rewardmanager/contract.go
JonathanOppenheimer: No I assumed it was deployed pre-dunrago -- I just wanted to make sure. It would be nice if there was some source of dates when each precompile was deployed (out of scope of this PR)

JonathanOppenheimer: Can I create an issue for this, or does this exist somewhere I am ignorant of

alarso16: I don't know of it, but it's definitely available in release notes

PR #4885, opened 2026-01-20
Simplex nits & bug fixes

samliok:
## Why this should be merged

A couple of nits and a bug fixes in the simplex package.


simplex/messages.go
```diff
@@ -224,6 +223,10 @@ func quorumRoundToP2P(qr *simplex.VerifiedQuorumRound) (*p2p.QuorumRound, error)
 		}
 	}
 	if qr.Finalization != nil {
+		// This can only happen if the finalization of the genesis block is being sent
+		if qr.Finalization.QC == nil {
+			return nil, nil
+		}
```
JonathanOppenheimer, line 235: Should this have an explicit test case?

simplex/messages.go
```diff
@@ -224,6 +223,10 @@ func quorumRoundToP2P(qr *simplex.VerifiedQuorumRound) (*p2p.QuorumRound, error)
 		}
 	}
 	if qr.Finalization != nil {
+		// This can only happen if the finalization of the genesis block is being sent
+		if qr.Finalization.QC == nil {
```
yacovm, line 233: Can we instead skip the QR and not include it in the replication response in case it's a `ReplicationResponse.Data`, and if it's the latest round we can just not respond?

samliok: ok, added a check for nil in `Send`

simplex/messages.go
samliok: asked claude to write some tests for me. I don't feel too strongly about keeping what claude wrote(which I reviewed & edited) especially since it adds 650 lines for some trivial functions 🤷.

samliok: i removed them, maybe will add in a separate pr.

JonathanOppenheimer: sgsg

PR #4886, opened 2026-01-20
refactor: share log package

JonathanOppenheimer:
## Why this should be merged

These two packages are identical and should be consolidated post-monorepo. 

## How this works

Shares the duplicated files in `graft/evm` 

## How this was tested
CI

No

### DIFF
--- a/graft/evm/log/format.go
+++ b/graft/evm/log/format.go
@@ -44,7 +44,7 @@ type TerminalStringer interface {
 
 func (h *TerminalHandler) format(buf []byte, r slog.Record, usecolor bool) []byte {
 	msg := escapeMessage(r.Message)
-	color := ""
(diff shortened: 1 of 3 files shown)


JonathanOppenheimer:
Note: I am sort of confused as to whether this is geth code or not -- the top of the files have our leading license, but don't have another.

alarso16 (approved):
This is fine, but there is only a couple differences between this implementation and geth's, so it would be relatively easy to libevmify

JonathanOppenheimer:
> This is fine, but there is only a couple differences between this implementation and geth's, so it would be relatively easy to libevmify

I thought we might as well consolidate first -- I am not really familiar on how to libevmify a package, but perhaps this would be a good learning experience to do so.

PR #4887, opened 2026-01-20
refactor: in-line network stats package

JonathanOppenheimer:
## Why this should be merged

~~These two packages are identical and should be consolidated post-monorepo.~~.

See https://github.com/ava-labs/avalanchego/pull/4887#pullrequestreview-3684344872

## How this works

In lines the function call. 

## How this was tested
CI

No


alarso16 (commented):
Nothing wrong with this, but it is unclear why this interface has one implementation and one use in a package separate from where it is used. I think this isn't really helpful, since it should really be a private struct in `network`, or just be fields in the `network` struct in the first place. I think it would be better to just delete the struct

JonathanOppenheimer:
> Nothing wrong with this, but it is unclear why this interface has one implementation and one use in a package separate from where it is used. I think this isn't really helpful, since it should really be a private struct in `network`, or just be fields in the `network` struct in the first place. I think it would be better to just delete the struct



> Nothing wrong with this, but it is unclear why this interface has one implementation and one use in a package separate from where it is used. I think this isn't really helpful, since it should really be a private struct in `network`, or just be fields in the `network` struct in the first place. I think it would be better to just delete the struct

Good suggestion! I've done so.

graft/coreth/network/network.go
```diff
@@ -352,10 +354,10 @@ func (n *network) AppRequestFailed(ctx context.Context, nodeID ids.NodeID, reque
 // calculateTimeUntilDeadline calculates the time until deadline and drops it if we missed he deadline to response.
 // This function updates metrics for app requests.
 // This is called by [AppRequest].
-func calculateTimeUntilDeadline(deadline time.Time, stats stats.RequestHandlerStats) (time.Time, error) {
+func calculateTimeUntilDeadline(deadline time.Time, n *network) (time.Time, error) {
```
alarso16: Why not make this a method on `network`?

alarso16: And if you're gonna edit this...
```suggestion
func (n *network) timeUntil(deadline time.Time) (time.Time, error) {
```

graft/coreth/network/network.go
```diff
@@ -103,7 +103,8 @@ type network struct {
 	codec                      codec.Manager                      // Codec used for parsing messages
 	appRequestHandler          message.RequestHandler             // maps request type => handler
 	peers                      *peerTracker                       // tracking of peers & bandwidth
-	appStats                   stats.RequestHandlerStats          // Provide request handler metrics
+	timeUntilDeadline          metrics.Timer                      // tracks time until deadline for incoming requests
```
alarso16: Actually, I think these can be global. I would check with someone else too, but it looks like `metrics.GetOrRegisterTimer` actually only "creates" a timer the first time, and then afterward, gets the same timer. So even if multiple `network`s are instantiated, then they still share the same timer/counter

graft/coreth/network/network.go
```diff
@@ -349,13 +351,13 @@ func (n *network) AppRequestFailed(ctx context.Context, nodeID ids.NodeID, reque
 	return handler.OnFailure()
 }
 
-// calculateTimeUntilDeadline calculates the time until deadline and drops it if we missed he deadline to response.
+// timeUntil calculates the time until deadline and drops it if we missed he deadline to response.
 // This function updates metrics for app requests.
 // This is called by [AppRequest].
-func calculateTimeUntilDeadline(deadline time.Time, stats stats.RequestHandlerStats) (time.Time, error) {
+func (n *network) timeUntil(deadline time.Time) (time.Time, error) {
```
alarso16: Slightly misleading comment. it doesn't "drop" the request if the deadline was missed, it returns an error. That should definitely be clarified

alarso16: If you can make the metrics global, then you can also remove the `network` reference
```suggestion
func timeUntil(deadline time.Time) (time.Time, error) {
```

graft/coreth/network/network.go
JonathanOppenheimer: It does appear that way -- for `GetOrRegister`: 
> // Gets an existing metric or registers the given one.
	// The interface can be the metric to register if not found in registry,
	// or a function returning the metric for lazy instantiation.
	
I've made these changes!

graft/coreth/network/network.go
```diff
@@ -364,7 +366,7 @@ func calculateTimeUntilDeadline(deadline time.Time, stats stats.RequestHandlerSt
 	// check if we have enough time to handle this request
 	if time.Until(bufferedDeadline) < minRequestHandlingDuration {
 		// Drop the request if we already missed the deadline to respond.
-		stats.IncDeadlineDroppedRequest()
+		droppedRequests.Inc(1)
```
alarso16: I think you should move this to where you actually drop the request on line 286

JonathanOppenheimer:
> I think this is an actual improvement, thanks going through several iterations

No thank you for the thorough reviews!

PR #4787, opened 2025-12-22
Switch to using solc provided by nix

maru-ava:
Updates #4782 to replace the use of the setup-solc (v0.8.30) with nix-supplied solc (v0.8.28). This ensures the same version of solidity can be used everywhere via the nix shell - CI/local, macos/linux - without requiring manual installation.

PR #4888, opened 2026-01-20
test: use nix provided solc and shellcheck

### LINKED ISSUE #4876: Replace all precompile `compile.go` solc with nix-provided solc
> Maybe file an issue to replace use of nix-provided solc?

_Originally posted by @maru-ava in https://github.com/ava-labs/avalanchego/pull/4871#discussion_r2700312237_

This should be done everywhere we use Arran's solc is currently used.

JonathanOppenheimer:
## Why this should be merged 

Closes #4876 

This PR replaces all usage of the `ARR4N/setup-solc` GitHub Action with nix-provided solc, and shellcheck. 

## How this works 

The `flake.nix` file already includes `solc` in the list of development dependencies, so we can just use `nix develop` 

## How this was tested
CI

No


graft/coreth/scripts/shellcheck.sh
```diff
@@ -2,38 +2,6 @@
```
maru-ava: Possible to use the avalanchego script and remove this one?

graft/subnet-evm/scripts/shellcheck.sh
```diff
@@ -2,8 +2,6 @@
```
maru-ava: ditto

JonathanOppenheimer:
> What appears here looks fine, but I think you'll also want to check the version of solc provided by nix. iirc it's less than v0.8.30 and it may not be necessary to set the emVersion to cancun because that is still the default of the nix version. The README.md for both subnet-evm and coreth both mention this issue.

Yeah it looks like nix provides 0.8.28 by default.

scripts/shellcheck.sh
```diff
@@ -16,9 +16,6 @@ fi
 # leading `.` in filenames it emits. A leading `.` will prevent the
 # use of `git apply` to fix reported shellcheck issues. This is
 # compatible with both macos and linux (unlike the use of -printf).
-# We exclude the graft/coreth and graft/subnet-evm directories to
```
maru-ava, line 19: Is this now safe to do? I forgot that the reason we were using separate shellchecks had to do with false positives.

flake.lock
```diff
@@ -8,23 +8,22 @@
       },
       "locked": {
         "lastModified": 1,
-        "narHash": "sha256-PPX4of5NK7HA4g4MMT1xYHjgpHYotB+uAA4Oh9sLhpI=",
+        "narHash": "sha256-B8igvkHnrQky5Zhtd/IkkJMBCDfDWwJKdIVBGl7fcR4=",
```
maru-ava: Why is this getting updated?

scripts/shellcheck.sh
JonathanOppenheimer: The false positives are just the ignored files in the subnet-evm shellcheck I believe:

```
IGNORED_FILES="
  metrics/validate.sh
  avalanchego/*
  contracts/lib/*
  contracts/node_modules/*
"
```

However, none of these files exist anymore.

flake.lock
JonathanOppenheimer: I ran nix update -- as you can see CI exploded after doing so -- I just reverted it.

maru-ava: Yeah, nix update is not something we want to be doing without a good reason.

maru-ava:
> > What appears here looks fine, but I think you'll also want to check the version of solc provided by nix. iirc it's less than v0.8.30 and it may not be necessary to set the emVersion to cancun because that is still the default of the nix version. The README.md for both subnet-evm and coreth both mention this issue.
> 
> Yeah it looks like nix provides 0.8.28 by default.

So, README updates?

JonathanOppenheimer:
> > > What appears here looks fine, but I think you'll also want to check the version of solc provided by nix. iirc it's less than v0.8.30 and it may not be necessary to set the emVersion to cancun because that is still the default of the nix version. The README.md for both subnet-evm and coreth both mention this issue.
> > 
> > 
> > Yeah it looks like nix provides 0.8.28 by default.
> 
> So, README updates?

yup, done!

.github/workflows/coreth-ci.yml
```diff
@@ -16,14 +16,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup-go-for-project
-      - name: Set up solc
-        uses: ARR4N/setup-solc@v0.2.0
-        with:
-          versions: "0.8.30"
+      - uses: ./.github/actions/install-nix
       - name: Run all lint checks
+        shell: nix develop --command bash -x {0}
```
alarso16, line 21: Is `bash -x` too verbose?

graft/coreth/plugin/evm/ExampleWarp.bin
alarso16, line 1: Why did the binary change?

graft/coreth/plugin/evm/compile.go
```diff
@@ -3,5 +3,8 @@
 
 package evm
 
+// Using shanghai EVM version is necessary in this case, as more recent solc optimizations change the binary output.
```
alarso16, line 6: the readme says to use `--evm-version cancun`

graft/coreth/Taskfile.yml
```diff
@@ -108,9 +106,7 @@ tasks:
     desc: Run golangci-lint with auto-fix where possible
     cmd: ../evm/scripts/lint_fix.sh
 
-  shellcheck:
-    desc: Run shellcheck static analysis on all shell scripts with version management
-    cmd: ./scripts/shellcheck.sh 
+
```
alarso16: Delete whitespace

graft/subnet-evm/core/compile.go
```diff
@@ -4,4 +4,4 @@
 package core
 
 // Step 1: Compile TrieStressTest contract to generate ABI and bin files
-//go:generate solc-v0.8.30 -o . --overwrite --abi --bin --pretty-json --evm-version cancun TrieStressTest.sol
+//go:generate solc -o . --overwrite --abi --bin --pretty-json TrieStressTest.sol
```
alarso16, line 7: You don't have to define an evm version?

graft/coreth/plugin/evm/compile.go
JonathanOppenheimer: Yes, but that causes this test to fail, which is why we need an exception to use the shanghai compiler

graft/subnet-evm/core/compile.go
JonathanOppenheimer: Nope, it's defined by nix.

.github/workflows/coreth-ci.yml
JonathanOppenheimer: `-x` seemed to be an existing pattern, and I think the additional output would never hurt. Do you think it's too verbose?

graft/coreth/plugin/evm/ExampleWarp.bin
JonathanOppenheimer: We compile with 0.28 instead of 0.30 on Shanghai.

alarso16 (commented):
The version changes and hard fork definitions feel arbitrary to me - I'm rather unfamiliar with the differences. I don't think I can give an accurate review

JonathanOppenheimer:
> The version changes and hard fork definitions feel arbitrary to me - I'm rather unfamiliar with the differences. I don't think I can give an accurate review

These are just what we use to compile the local test contracts - I don't think minor differences here are impactful in the bindings binary. We don't change the definition of any of the contracts or anything that would be harmful in this PR. I'll seek another review! thanks for taking a look

JonathanOppenheimer:
Closed in favor of https://github.com/ava-labs/avalanchego/pull/4787

PR #4890, opened 2026-01-21
Firewood v0.1.0

alarso16:
## Why this should be merged

Updates Firewood version!

## How this works

There were some breaking interface changes, but nothing else significant

## How this was tested

CI

No

### DIFF
--- a/firewood/syncer/syncer.go
+++ b/firewood/syncer/syncer.go
@@ -138,6 +138,6 @@ func (*database) CommitChangeProof(context.Context, maybe.Maybe[[]byte], struct{
 
 func (db *database) Clear() error {
 	// Prefix delete key of length 0.
-	_, err := db.db.Update([][]byte{{}}, [][]byte{nil})
+	_, err := db.db.Update([]ffi.BatchOp{ffi.PrefixDelete([]byte{})})
 	return err
 }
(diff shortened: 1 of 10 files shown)


graft/evm/firewood/account_trie.go
```diff
@@ -172,9 +170,8 @@ func (a *accountTrie) DeleteAccount(addr common.Address) error {
 	key := crypto.Keccak256Hash(addr.Bytes()).Bytes()
 	// Queue the key for deletion
 	a.dirtyKeys[string(key)] = nil
-	a.updateKeys = append(a.updateKeys, key)
-	a.updateValues = append(a.updateValues, nil) // Must use nil to indicate deletion
-	a.hasChanges = true                          // Mark that there are changes to commit
+	a.updateOps = append(a.updateOps, ffi.PrefixDelete(key)) // Remove all storage
```
RodrigoVillar, line 173: q: why use `ffi.PrefixDelete()` instead of `ffi.Delete()`? My understanding is that the account info for a given account will always live in one specific location in state.

alarso16: `PrefixDelete` will also remove all associated storage. Otherwise, we have to iterate through its children and delete each explicitly

RodrigoVillar: Ah understood.

maru-ava:
Oops, my apologies I neglected to check the base branch.

PR #4749, opened 2025-12-16
docs: rewrite releasing readme

JonathanOppenheimer:
## Why this should be merged
With the monorepo, now that coreth and subnet-evm are added, the release process has changed drastically. Here, I update the releasing doc instructions to align with the new process.

This is best reviewed by looking commit commit at the actual changes: 

- [docs: add explanation of modules](https://github.com/ava-labs/avalanchego/pull/4749/commits/ac42f95232f18b5bf4ec55bd06eb0b45dca7fbd4)
- [docs: add releasing instructions](https://github.com/ava-labs/avalanchego/pull/4749/commits/4b9038d2bf43fc78ea1df7e11b833432b27fa5ba)
- [docs: add explanation of the building process](https://github.com/ava-labs/avalanchego/pull/4749/commits/e7e925f7204181db60bba8a909cc864e95589198)
- [docs: update post release section](https://github.com/ava-labs/avalanchego/pull/4749/commits/0ac1ad2c986a0f77fdf47dc13dcaa834799268db)
- [docs: add appendix of sorts of information](https://github.com/ava-labs/avalanchego/pull/4749/commits/c070f6c46718bfaf4be84cf433e6a7b801218e31)

Closes #4658 

## How this was tested
No code changes

## Need to be documented in RELEASES.md?
No

PR #4891, opened 2026-01-21
refactor: share version across codebase

JonathanOppenheimer:
## Why this should be merged

This unifies the version across AvalancheGo, coreth, and subnet-evm. A precursor to https://github.com/ava-labs/avalanchego/pull/4749. 

## How this works
Derives coreth and subnet-evm version from AvalancheGo version. 

## How this was tested
CI, modified version test

Yes, done so in follow up PR.


ceyonur (approved):

graft/subnet-evm/plugin/evm/version_test.go
```diff
@@ -4,34 +4,22 @@
...
-		"%s has subnet-evm version %s stated as compatible with RPC chain VM protocol version %d but AvalancheGo protocol version is %d",
-		filepath.Base(compatibilityFile), Version, rpcChainVMVersion, version.RPCChainVMProtocol)
+	// Verify that the Subnet-EVM version is properly derived from AvalancheGo version
+	expectedVersion := fmt.Sprintf("v%d.%d.%d", version.Current.Major, version.Current.Minor, version.Current.Patch)
+	require.Equal(t, expectedVersion, Version,
+		"Subnet-EVM version should be automatically derived from version.Current")
+
+	// Verify that the version is compatible with current RPC chain VM protocol
+	// Compatibility is now maintained through version/compatibility.json in the main repo
+	require.NotZero(t, version.RPCChainVMProtocol,
```
alarso16: Isn't this testing avalalnchego? I think you should remove this

JonathanOppenheimer: removed the whole test.

alarso16 (approved):

graft/coreth/plugin/evm/version.go
StephenButtolph, line 1: Why does this file still exist?

graft/subnet-evm/plugin/evm/version.go
StephenButtolph, line 1: Why does this file still exist?

graft/coreth/plugin/evm/version.go
JonathanOppenheimer: We use this for logging across the vm, and the version is modified from the avalanchego version (so it doesn't lead with `avalanchego`. What would you prefer to be done here instead? Use the avalanchego version file and trim out the prefix each time?

StephenButtolph: I think anything that exists in these files should be moved to avalanchego's version package if needed.

JonathanOppenheimer: This is a much bigger difference, but arguably cleaner. Is this what you had in mind?

version/application_test.go
```diff
@@ -51,6 +51,77 @@ func TestApplicationString(t *testing.T) {
 	}
 }
 
+func TestApplicationSemantic(t *testing.T) {
```
JonathanOppenheimer, line 54: These tests seem excessive to me, but I copied the format of the other tests in this package.

graft/subnet-evm/plugin/main.go
```diff
@@ -14,6 +14,6 @@ import (
 func main() {
 	evm.RegisterAllLibEVMExtras()
 
-	versionString := fmt.Sprintf("Subnet-EVM/%s [AvalancheGo=%s, rpcchainvm=%d]", evm.Version, version.Current, version.RPCChainVMProtocol)
+	versionString := fmt.Sprintf("Subnet-EVM/%s [AvalancheGo=%s, rpcchainvm=%d]", version.Current.Semantic(), version.Current, version.RPCChainVMProtocol)
```
StephenButtolph: This format doesn't seem useful anymore. The subnet-EVM version isn't different than the avalanchego version.

graft/subnet-evm/plugin/evm/version.go
```diff
@@ -1,19 +0,0 @@
-// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
-// See the file LICENSE for licensing terms.
-
-package evm
-
-import "fmt"
-
-var (
-	// GitCommit is set by the build script
-	GitCommit string
```
StephenButtolph, line 10: I think the subnet-evm build script needs to set the avalancego `GitCommit` now, rather than this (now deleted) var.

version/application_test.go
StephenButtolph: overkill? probably... But they are already here... so lgtm

graft/subnet-evm/plugin/main.go
JonathanOppenheimer: Yeah they're basically the same -- we can just print the semantic one.

graft/subnet-evm/plugin/evm/version.go
JonathanOppenheimer: ah good catch! I think I did this correctly now.

PR #4893, opened 2026-01-21
Firewood v0.1.0

alarso16:
## Why this should be merged

Updates Firewood to v0.1.0 with associated interface changes

## How this works

Uses new interfaces, replaces #4890.

## How this was tested

CI

No

### DIFF
--- a/graft/evm/firewood/account_trie.go
+++ b/graft/evm/firewood/account_trie.go
@@ -5,6 +5,7 @@ package firewood
 
 import (
 	"errors"
+	"slices"
 
 	"github.com/ava-labs/firewood-go-ethhash/ffi"
 	"github.com/ava-labs/libevm/common"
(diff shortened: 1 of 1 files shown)


Elvis339 (approved):

graft/evm/firewood/account_trie.go
```diff
@@ -269,26 +265,22 @@ func (*accountTrie) Prove([]byte, ethdb.KeyValueWriter) error {
...
 		newTrie.dirtyKeys[k] = append([]byte{}, v...)
 	}
 
-	// Deep copy updateKeys and updateValues slices
-	for i := range a.updateKeys {
-		newTrie.updateKeys[i] = append([]byte{}, a.updateKeys[i]...)
-		newTrie.updateValues[i] = append([]byte{}, a.updateValues[i]...)
-	}
+	// Copy updateOps slice
+	newTrie.updateOps = append([]ffi.BatchOp{}, a.updateOps...)
```
StephenButtolph: Something is weird here with `updateOps`. We are currently allocating the `updateOps` slice twice (once above and once here).

Also, are we expecting to have deep-copied the ops? (not sure if this even makes sense).

graft/evm/firewood/account_trie.go
```diff
@@ -269,26 +265,22 @@ func (*accountTrie) Prove([]byte, ethdb.KeyValueWriter) error {
...
+		root:       a.root,
+		reader:     a.reader, // Share the same reader
+		hasChanges: a.hasChanges,
+		dirtyKeys:  make(map[string][]byte, len(a.dirtyKeys)),
+		updateOps:  make([]ffi.BatchOp, len(a.updateOps)),
 	}
 
 	// Deep copy dirtyKeys map
 	for k, v := range a.dirtyKeys {
 		newTrie.dirtyKeys[k] = append([]byte{}, v...)
```
StephenButtolph: This style of appending to an empty slice should probably be replaced with the more modern `slices.Clone`... Unless there is extremely specific care taken to avoid preserving "nilness" of `v`

alarso16: Nil-ness was lazy, good catch

graft/evm/firewood/account_trie.go
alarso16: Something weird here for sure, I'll just use `slices.Clone`

PR #4894, opened 2026-01-21
feat(sync): finalize firewood syncer and add finalization tests

powerslider:
## Why this should be merged

Enable correct finalization support for `FirewoodSyncer` that is idempotent and disallow hanging issues on code syncing.

## How this works

- Close code queue via Finalize (idempotent) and call from Sync.
- Enqueue code hashes before committing range proofs.
- Add finalize scenarios and registry finalize tests.

## How this was tested

added UT testing finalization logic

no


alarso16:
Looks mostly good, but I'm going to merge into mine and make some changes

PR #4896, opened 2026-01-21
[tools] Allow ./tools path to be used for local tools

maru-ava:
Previously the go module for cli tools at ./tools/go.mod precluded the ./tools path for being used for defining cli tools that are part of the main go module. Moving the module to ./tools/external enables ./tools to be used for both.

## How this works

Move `./tools/go.mod` to `./tools/external/go.mod` and update all references.

## How this was tested

CI

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,7 +183,7 @@ jobs:
       - uses: ./.github/actions/install-nix
       - shell: nix develop --command bash -x {0}
         run: task check-generate-load-contract-bindings
-  go_mod_tidy:
+  check_go_mod_tidy:
     name: Up-to-date go.mod and go.sum
     runs-on: ubuntu-latest
     steps:
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -100,7 +100,7 @@ tasks:
       - cmd: go mod tidy
       - task: check-clean-branch
       - cmd: go mod tidy
-        dir: tools
+        dir: tools/external
       - task: check-clean-branch
 
   create-kind-cluster:


tools/README.md
```diff
@@ -0,0 +1,19 @@
+# Tools
```
JonathanOppenheimer, line 1: (no action needed) could say when to use external tools  vs. when to use tools (like lint is one right)

JonathanOppenheimer (approved):
LGTM

tools/README.md
maru-ava: Suggested changes to what appears below?

JonathanOppenheimer: You answer the question in one of your comments but not in the `README`

```go
// CLI tools intended for invocation with `go tool` should be added to
// tools/external/go.mod to avoid polluting the main module's dependencies. See
// tools/external/go.mod for usage details.
```

something like 

```markdown
## When to Use Each Location

- `./tools/external/go.mod`: CLI tools intended for invocation with `go tool` should be added here to avoid polluting the main module's dependencies
  - e.g. golangci-lint, gencodec, go-license, task
  - See `external/go.mod` for usage details
  - `./scripts/run_tool.sh <tool>`

- `./tools/<command>/`: Repo-level commands that are part of the main avalanchego module
  - `go run ./tools/<command>`
  ```
  
  maybe not necessary though - you cover this with `## `./external/go.mod` and with `## Adding Repo-Local Commands`

JonathanOppenheimer: The only thing of value here would be to be like (`e.g. golangci-lint, gencodec, go-license, task`)

maru-ava: Not disagreeing that docs could be clearer, but 'when to use each location' is basically 'am I defining a dependency on an external tool (hence tools/external/go.mod) or defining a new tool in this repo.

JonathanOppenheimer: sg!

PR #4904, opened 2026-01-22
Update vms/README.md

meaghanfitzgerald:
## Why this should be merged

This information is obsolete

## How this works

markdown

## How this was tested

it wasn't

no


meaghanfitzgerald:
we could also delete this whole page if we want

PR #4908, opened 2026-01-23
test: Ensure code hashes are synced in Firewood

alarso16:
## Why this should be merged

Although Firewood does properly sync all code hashes, it wasn't tested. Now it is!

## How this works

Adds the minimum iterating functionality through the geth interface to check all accounts. I had to refactor quite a bit though

## How this was tested

It is the test!

No

### DIFF
--- a/graft/evm/firewood/account_trie.go
+++ b/graft/evm/firewood/account_trie.go
@@ -248,10 +248,14 @@ func (*accountTrie) GetKey([]byte) []byte {
 }
 
 // NodeIterator implements state.Trie.
-// Firewood does not support iterating over internal nodes.
-// This always returns an error.
-func (*accountTrie) NodeIterator([]byte) (trie.NodeIterator, error) {
-	return nil, errors.New("NodeIterator not implemented for Firewood")
+// It iterates only over account nodes, so all nodes are leaf nodes.
+// The first key yielded will be >= [start].
+func (a *accountTrie) NodeIterator(start []byte) (trie.NodeIterator, error) {
+	r, ok := a.reader.(*reader)
+	if !ok {
+		return nil, errors.New("invalid reader type for account trie")
+	}
+	return newAccountIt(r.revision, start)
 }
 
 // Prove implements state.Trie.
--- a/graft/evm/firewood/account_trie.go
+++ b/graft/evm/firewood/account_trie.go
@@ -255,7 +255,7 @@ func (a *accountTrie) NodeIterator(start []byte) (trie.NodeIterator, error) {
 	if !ok {
 		return nil, errors.New("invalid reader type for account trie")
 	}
-	return newAccountIt(r.revision, start)
+	return newAccountIterator(r.revision, start)
 }
 
 // Prove implements state.Trie.
--- a/graft/evm/firewood/account_trie.go
+++ b/graft/evm/firewood/account_trie.go
@@ -5,6 +5,7 @@ package firewood
 
 import (
 	"errors"
+	"fmt"
 	"slices"
 
 	"github.com/ava-labs/firewood-go-ethhash/ffi"
@@ -247,13 +248,15 @@ func (*accountTrie) GetKey([]byte) []byte {
 	return nil
 }
 
+var errInvalidReader = errors.New("invalid reader type")
+
 // NodeIterator implements state.Trie.
 // It iterates only over account nodes, so all nodes are leaf nodes.
 // The first key yielded will be >= [start].
 func (a *accountTrie) NodeIterator(start []byte) (trie.NodeIterator, error) {
 	r, ok := a.reader.(*reader)
 	if !ok {
-		return nil, errors.New("invalid reader type for account trie")
+		return nil, fmt.Errorf("%w: got %T", errInvalidReader, a.reader)
 	}
 	return newAccountIterator(r.revision, start)
 }
--- a/graft/evm/firewood/iterator.go
+++ b/graft/evm/firewood/iterator.go
@@ -12,18 +12,18 @@ import (
 	"github.com/ava-labs/libevm/trie"
 )
 
-var _ trie.NodeIterator = (*accountIt)(nil)
+var _ trie.NodeIterator = (*accountIterator)(nil)
 
-type accountIt struct {
+type accountIterator struct {
 	it *ffi.Iterator
 }
 
-func newAccountIt(rev *ffi.Revision, start []byte) (*accountIt, error) {
+func newAccountIterator(rev *ffi.Revision, start []byte) (*accountIterator, error) {
 	it, err := rev.Iter(start)
 	if err != nil {
 		return nil, err
 	}
-	a := &accountIt{
+	a := &accountIterator{
 		it: it,
 	}
 
@@ -39,7 +39,7 @@ func newAccountIt(rev *ffi.Revision, start []byte) (*accountIt, error) {
 }
 
 // Next advances the iterator to the next account.
-func (a *accountIt) Next(bool) bool {
+func (a *accountIterator) Next(bool) bool {
 	next := a.it.Next()
 	for ; next; next = a.it.Next() {
 		if len(a.it.Key()) == common.HashLength {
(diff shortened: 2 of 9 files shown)


graft/evm/firewood/storage_trie.go
```diff
@@ -34,6 +37,10 @@ func (*storageTrie) Hash() common.Hash {
 	return common.Hash{}
 }
 
+func (*storageTrie) NodeIterator([]byte) (trie.NodeIterator, error) {
+	return nil, errors.New("storage trie does not support node iteration")
```
powerslider: Extract this error as a sentinel error.

graft/evm/firewood/iterator.go
powerslider, line 1: I don't think I like the name `account_it.go`, because it could be interpreted in many ways. Some people read `it` as integration test which is definitely not the case here. I would probably just name it `account_iterator.go` since it communicates exactly what it is and what it does. Of course the struct has to be renamed as well.

graft/evm/firewood/storage_trie.go
powerslider: Also is this just temporary or it won't ever support it?

What worries me here is that when `AssertTrieConsistency` is used on a storage root (e.g., in the DB consistency checks), a Firewood-backed iterator will return zero leaves and still pass, masking storage mismatches.

graft/evm/firewood/iterator.go
```diff
@@ -0,0 +1,95 @@
...
+	a := &accountIt{
+		it: it,
+	}
+
+	// Because the user manages the lifetime of the iterator, we must guarantee
+	// that the underlying resources are freed when the iterator is garbage
+	// collected.
+	runtime.AddCleanup(a, func(it *ffi.Iterator) {
+		if err := it.Drop(); err != nil {
+			log.Warn("failed to drop iterator: %w, Firewood may leak resources", err)
```
powerslider: The `libevm` logger typically expects structured key/value pairs, not `fmt`-style formatting. This likely logs the literal `%w` and omits the error. To mitigate ⏬ 
```suggestion
            log.Warn("cannot drop iterator - Firewood may leak resources", "error", err)
```

graft/evm/firewood/account_trie.go
```diff
@@ -248,10 +248,14 @@ func (*accountTrie) GetKey([]byte) []byte {
...
-// Firewood does not support iterating over internal nodes.
-// This always returns an error.
-func (*accountTrie) NodeIterator([]byte) (trie.NodeIterator, error) {
-	return nil, errors.New("NodeIterator not implemented for Firewood")
+// It iterates only over account nodes, so all nodes are leaf nodes.
+// The first key yielded will be >= [start].
+func (a *accountTrie) NodeIterator(start []byte) (trie.NodeIterator, error) {
+	r, ok := a.reader.(*reader)
+	if !ok {
+		return nil, errors.New("invalid reader type for account trie")
```
powerslider: Extract sentinel error here.

graft/evm/firewood/iterator.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}, it)
+	return a, nil
+}
+
+// Next advances the iterator to the next account.
+func (a *accountIterator) Next(bool) bool {
+	next := a.it.Next()
+	for ; next; next = a.it.Next() {
+		if len(a.it.Key()) == common.HashLength {
```
powerslider, line 48: I would add a clarifying comment above this check. Something like:
```
// Account keys are 32 bytes, storage keys are 64 bytes (account + storage).
```

graft/evm/firewood/iterator.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}, it)
+	return a, nil
+}
+
+// Next advances the iterator to the next account.
+func (a *accountIterator) Next(bool) bool {
+	next := a.it.Next()
+	for ; next; next = a.it.Next() {
+		// Account keys are 32 bytes
```
JonathanOppenheimer: Wait so there are no keys which are 32 bytes and are not account keys right? So all 32 byte keys are account keys.

graft/evm/firewood/iterator.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// Because the user manages the lifetime of the iterator, we must guarantee
+	// that the underlying resources are freed when the iterator is garbage
+	// collected.
+	runtime.AddCleanup(a, func(it *ffi.Iterator) {
+		if err := it.Drop(); err != nil {
+			// Best-effort GC cleanup: there is no safe way to surface or recover from
+			// Drop errors here, and at this point the iterator is already unreachable.
+			log.Warn("failed to drop iterator, Firewood may leak resources", "error", err)
```
JonathanOppenheimer, line 37: Could we add a `.release()` method for the iterator and call it here? It could just set the iterator to nil, if it's not already nil. (and drop the error?)

graft/evm/firewood/storage_trie.go
```diff
@@ -34,6 +37,12 @@ func (*storageTrie) Hash() common.Hash {
 	return common.Hash{}
 }
 
+var errStorageIteratorNotSupported = errors.New("storage trie does not support node iteration")
+
+func (*storageTrie) NodeIterator([]byte) (trie.NodeIterator, error) {
```
JonathanOppenheimer, line 42: I assume this is not required -- looks like the tests check storage root hashes, but does not iterate storage tries.

graft/subnet-evm/sync/statesync/sync_test.go
```diff
@@ -530,13 +530,12 @@ func assertDBConsistency(t testing.TB, root common.Hash, clientDB state.Database
 	require.NoError(t, accountIt.Error())
 	trieAccountLeaves := 0
 
-	synctest.AssertTrieConsistency(t, root, serverDB.TrieDB(), clientDB.TrieDB(), func(key, val []byte) error {
```
JonathanOppenheimer, line 533: is there a reason why the consistency is only checked client side, rather than server side as well?

JonathanOppenheimer (commented):
I have a few (maybe obvious) questions.

graft/evm/firewood/iterator.go
alarso16: Nope, we have to adhere to the interface definition (which notably does not include any freeing mechanism)

graft/evm/firewood/storage_trie.go
alarso16: Yes. I could have added it, but it's not necessary, so I didn't. The fact that the state root is correct proves that all keys and values match. Specifically, what we are checking is that the code hashes in the accounts are present via `rawdb`, and we don't need to iterate storage for that

graft/subnet-evm/sync/statesync/sync_test.go
alarso16: Nothing can be wrong with the server side, right? It's the source of truth

graft/evm/firewood/iterator.go
alarso16: Yes

graft/evm/firewood/iterator.go
JonathanOppenheimer: gotcha!

graft/evm/firewood/iterator.go
JonathanOppenheimer: Could you modify the comment to say that? There's a difference between the two, that makes this explicitly okay to do (if all rectangles were squares, then this would be okay)

graft/subnet-evm/sync/statesync/sync_test.go
JonathanOppenheimer: Ah I just meant like checking that the code also exists in the server and that it matches? Maybe I'm misunderstanding but `serverDB` isn't even used in the test -- it is the source of truth ofc.

powerslider (approved):

graft/subnet-evm/sync/statesync/sync_test.go
alarso16: The server is constructed to contain the code hashes. If it doesn't have them, then the client certainly won't either. `synctest.AssertTrieConsistency` checks that the server's keys and values match with the client's prior to the callback. Since the keys and values match, we must only ensure that the client's state is valid at that point.

JonathanOppenheimer (approved):

alarso16:
There's a simpler way to implement this, so I'll close this PR

PR #4909, opened 2026-01-26
Add continuous staking TXs

rrazvan1:

### DIFF
--- a/vms/platformvm/block/builder/builder.go
+++ b/vms/platformvm/block/builder/builder.go
@@ -13,7 +13,6 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/snow"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/engine/common"
 	"github.com/ava-labs/avalanchego/utils/set"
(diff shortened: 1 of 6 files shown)


vms/platformvm/block/builder/builder.go
```diff
@@ -323,9 +323,23 @@ func buildBlock(
...
-		rewardValidatorTx, err := NewRewardValidatorTx(builder.txExecutorBackend.Ctx, stakerTxID)
+		var rewardValidatorTx *txs.Tx
+
+		stakerTx, _, err := parentState.GetTx(stakerTxID)
 		if err != nil {
-			return nil, fmt.Errorf("could not build tx to reward staker: %w", err)
+			return nil, err
+		}
+
+		if _, ok := stakerTx.Unsigned.(txs.ContinuousStaker); ok {
```
yacovm: Seems like this is an abstraction leak. We're building the block here, why should we be aware of what a continuous staker is? 

I think we should organize the code in a different way to hide this continuous staker. 

For example, we can compute `rewardValidatorTx` inside `getNextStakerToReward`.

rrazvan1: what do you think about then new approach that I pushed?

having a `reward_tx_factory.go` in the `txs` package which is responsible to create the proper reward tx based on the staker transaction?

vms/platformvm/txs/reward_tx_factory.go
```diff
@@ -0,0 +1,38 @@
...
+	if _, ok := stakerTx.Unsigned.(ContinuousStaker); ok {
+		return NewRewardContinuousValidatorTx(ctx, txID, uint64(timestamp.Unix()))
+	}
+
+	return NewRewardValidatorTx(ctx, txID)
+}
+
+func NewRewardValidatorTx(ctx *snow.Context, txID ids.ID) (*Tx, error) {
+	utx := &RewardValidatorTx{TxID: txID}
+	tx, err := NewSigned(utx, Codec, nil)
```
yacovm, line 24: This might seem like an unimportant rant, but - I don't think we should be using NewSigned in the P-chain production code. 

It's only used in tests, and in my opinion this should be moved to some kind of utility package. 

We're passing `nil` which is supposed to be an array of signers. 

Why not just initialize `Tx` directly? All fields but `bytes` are exported so it's not an issue. 

We're just doing un-necessary serialization and hashing by using this constructor.

vms/platformvm/block/builder/builder.go
yacovm: Yeah makes sense. I am not sure the file name fits though, and I'm not sure we should have a separate file at all, given that we're only using it in the builder.

vms/platformvm/reward/calculator.go
```diff
@@ -81,3 +81,16 @@ func Split(totalAmount uint64, shares uint32) (uint64, uint64) {
...
 	return amountFromShares, remainderAmount
 }
+
+// ProportionalAmount computes (amount * numerator) / denominator with overflow safety.
+// The result is floored (rounded down) due to integer division. When splitting a
+// value into parts, consider whether rounding errors should favor one side:
+//   - To ensure you don't over-allocate: compute the smaller portion with this
+//     function, then subtract from total to get the remainder.
+func ProportionalAmount(amount, numerator, denominator uint64) uint64 {
+	result := amount / denominator * numerator
```
yacovm, line 91: Can we add some division by zero checks?

vms/platformvm/txs/reward_tx_factory.go
rrazvan1: The only extra thing is hashing the unsigned bytes which should be signed, but since there are no signers, it's useless.

I agree with you that this can be improved, but ACP-236 is already touching a lot of components, and if something was already there, I tried using it as it is to avoid changing stuff which isn't necessary.

vms/platformvm/reward/calculator.go
rrazvan1: Makes sense, will do it!

rrazvan1:
Closed in favor of #4944

PR #4911, opened 2026-01-26
Fail HealthCheck if Average Block Acceptance Times are too High

samliok:
## Why this should be merged

During on-call, we’ve occasionally been paged due to high block processing times. If a block takes longer than 50 seconds to process, the node fails a health check. Once the block is verified, the node is marked healthy again only to fail repeatedly if processing times remain high.

This PR adds an additional health check to trigger only when block acceptance times have been elevated for more than 5 minutes to give us additional context during high block processing times. 

## How this works

Added a metric + prometheus gauge to `snowman.metrics`. Also created a wrapper around the `Averager`.

## How this was tested

Ran these metrics on an AvalancheGo node for ~a day. 

image

We can see that for the first 5 minutes the metric is 0(to avoid failing health check before enough time has passed), then it averages around 2 to 3 seconds.

### DIFF
--- a/utils/math/matured_averager.go
+++ b/utils/math/matured_averager.go
@@ -18,8 +18,8 @@ type maturedAverager struct {
 var _ Averager = (*maturedAverager)(nil)
 
 // NewMaturedAverager creates a new matured averager that wraps the provided
-// averager. It returns zero from [averager.Read] until at least `halfLife“ duration has
-// passed since the first Observe call.
+// `averager`. It returns zero from [Averager.Read] until at least `halfLife` duration has
+// passed since the first [Averager.Observe] call.
 func NewMaturedAverager(halfLife time.Duration, averager Averager) Averager {
 	return &maturedAverager{
 		Averager: averager,


snow/consensus/snowman/topological.go
samliok, line 330: should we change this to 

`	if maxTimeProcessing > ts.params.MaxItemProcessingTime  &  avgAcceptanceTime > maxAcceptanceTime {`

snow/consensus/snowman/topological.go
```diff
@@ -331,9 +334,21 @@ func (ts *Topological) HealthCheck(context.Context) (interface{}, error) {
 		errs = append(errs, err)
 	}
 
+	// Check average acceptance time
+	avgAcceptanceTime := ts.metrics.GetAverageAcceptanceTime()
+	if avgAcceptanceTime > maxAcceptanceTime {
```
yacovm, line 341:
```
	a := math.NewUninitializedAverager(5 * time.Minute)
	a.Observe(float64(time.Second*16), time.Now())
	a.Observe(float64(time.Second*17), time.Now().Add(time.Second))
	a.Observe(float64(time.Second*18), time.Now().Add(time.Second*2))
	a.Observe(float64(time.Second*14), time.Now().Add(time.Second*3))
	fmt.Println(time.Duration(a.Read())) // prints 16.248552608s
```

I think the intent was to check that we have a high acceptance time for a sustained period of time. This check will immediately cause us to fail, no?

samliok: Screenshot 2026-01-28 at 2 05 58 PM
just got my node running, looks like it doesn't print a high value on startup

yacovm: I am saying that we want to eliminate possible flakyness by making sure the averager samples high acceptance time for a period of time that is longer than a threshold, and not failing the health check immediately if we sample a single sample for a high acceptance time.

You can test this by adding an iptables rule `sudo iptables -A OUTPUT --destination-port 9651 -j DROP` and seeing when the healthcheck fails - after 5 minutes or after a minute.

samliok: Screenshot 2026-01-29 at 2 27 51 PM
updated so that it we wait `halfLife` time before reporting values.

snow/consensus/snowman/metrics.go
```diff
@@ -138,6 +143,16 @@ func newMetrics(
 			Name: "polls_failed",
 			Help: "number of failed polls",
 		}),
+		avgAcceptanceTime: math.NewMaturedAverager(
+			5*time.Minute,
```
yacovm: Let's extract this 5 minutes to a constant

snow/consensus/snowman/metrics.go
```diff
@@ -138,6 +143,16 @@ func newMetrics(
...
+			math.NewUninitializedAverager(5*time.Minute),
```
yacovm: ditto, let's extract this 5 minutes to a different constant

snow/consensus/snowman/metrics.go
```diff
@@ -202,6 +218,8 @@ func (m *metrics) Accepted(
 
 	builtDuration := now.Sub(timestamp)
 	m.buildLatencyAccepted.Add(float64(builtDuration))
+	m.avgAcceptanceTime.Observe(float64(builtDuration), now)
+	m.avgAcceptanceTimeGauge.Set(m.avgAcceptanceTime.Read())
```
yacovm: why do we need this? Don't we have `m.buildLatencyAccepted`?

snow/consensus/snowman/topological.go
yacovm: No.

Max item processing is an indicator of a queue forming in block processing and consensus. 

The acceptance time is from the block build time, so it also accounts for block dissemination.

Moreover, if you don't have any blocks being accepted due to some bug, then you might have a queue of blocks being processed but the acceptance time will be 0 since nothing is accepted... I think it's better to keep both

snow/consensus/snowman/metrics.go
samliok: fair, i guess i wanted to see the actual graph of the metric but we can just probably just average `builtDuration` on grafana right?

yacovm: but you can average it in Grafana. There is no need to average it using an averager...

snow/consensus/snowman/metrics.go
```diff
@@ -12,10 +12,13 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/linked"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/utils/metric"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
 )
 
+const acceptanceTimeSmoothingFactor = 5 * time.Minute
```
joshua-kim: 1. nit: Can we inline this into `newMetrics`? This clutters the package scope for a variable that's only used in one scope.
2. Avoid calling the same concept different names -- we refer to this as a "smoothing factor" but the rest of the averager code refers to this as a "half-life".

utils/math/matured_averager.go
```diff
@@ -0,0 +1,50 @@
...
+// maturedAverager wraps an Averager and returns zero until the averager has
+// been running for at least the half-life duration.
+type maturedAverager struct {
+	Averager
+	halflife     time.Duration
+	startTime    time.Time
+	lastObserved time.Time
+	hasObserved  bool
+}
+
```
joshua-kim, line 17: Can we add `_ Averager = (*maturedAverager)(nil)` here to indicate that this satisfies that interface? ([example](https://github.com/golang/go/blob/master/src/net/http/server.go#L3879-L3891)).

utils/math/matured_averager_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMaturedAverager(t *testing.T) {
+	require := require.New(t)
```
joshua-kim: Team style is to avoid using concrete instances of `require` since it can cause issues where the wrong instance is passed around in subtests -- it doesn't make a semantic difference in this test but we should still be consistent.

utils/math/matured_averager_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	halflife := 5 * time.Minute
+	startTime := time.Now()
```
joshua-kim: nit: hard-code timestamps to create deterministic tests (e.g `time.Time{}` should work here) -- non-determinism makes debugging more difficult

utils/math/matured_averager.go
```diff
@@ -0,0 +1,50 @@
...
+package math
+
+import "time"
+
+// maturedAverager wraps an Averager and returns zero until the averager has
+// been running for at least the half-life duration.
+type maturedAverager struct {
+	Averager
+	halflife     time.Duration
```
joshua-kim: Go convention is to camelCase multiword expressions -- so half-life would be `halfLife`

utils/math/matured_averager.go
```diff
@@ -0,0 +1,50 @@
...
+	startTime    time.Time
+	lastObserved time.Time
+	hasObserved  bool
+}
+
+// NewMaturedAverager creates a new matured averager that wraps the provided
+// averager. It returns zero from Read() until at least halflife duration has
```
joshua-kim: Godoc won't render this correctly -- for exoprted code we should reference it with brackets as `[Averager.Read]`. You can use tooling like `pkgsite` to confirm godoc rendering. To refer to parameters, since godoc uses plain markdown we'd want to use backticks around `halfLife`.

utils/math/matured_averager.go
```diff
@@ -0,0 +1,50 @@
...
+// passed since the first Observe call.
+func NewMaturedAverager(halflife time.Duration, averager Averager) Averager {
+	return &maturedAverager{
+		Averager: averager,
+		halflife: halflife,
```
joshua-kim: Go convention is to camelCase multiword expressions for variables -- so half-life would be `halfLife`, instead of `halflife`.

snow/consensus/snowman/metrics.go
```diff
@@ -209,7 +219,7 @@ func (m *metrics) Accepted(
 
 	builtDuration := now.Sub(timestamp)
 	m.buildLatencyAccepted.Add(float64(builtDuration))
-
+	m.avgAcceptanceTime.Observe(float64(builtDuration), now)
```
joshua-kim: nit: latency instead of time to match the surrounding metrics?

snow/consensus/snowman/topological.go
```diff
@@ -333,9 +336,21 @@ func (ts *Topological) HealthCheck(context.Context) (interface{}, error) {
...
+			avgAcceptanceTime,
+			maxAcceptanceTime,
+		)
+		errs = append(errs, err)
+	}
+
 	return map[string]interface{}{
 		"processingBlocks":       numProcessingBlks,
 		"longestProcessingBlock": maxTimeProcessing.String(), // .String() is needed here to ensure a human readable format
+		"avgAcceptanceTime":      avgAcceptanceTime.String(),
```
joshua-kim, line 353: I want to flag this since I noticed it in review but I don't want to block on this -- but this health check isn't tested. Ideally we would test this with a unit test on `Topological`, but it looks like we'd have to refactor to do this and I don't want to scope creep right now.

utils/math/matured_averager_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	matured := NewMaturedAverager(halflife, NewAverager(0, halflife, startTime))
+
+	// Before any observations, should return zero
+	require.Equal(0.0, matured.Read())
```
joshua-kim: nit: use `require.Zero` and `require.NotZero` for these assertions?

utils/math/matured_averager.go
```diff
@@ -0,0 +1,50 @@
...
+	halflife     time.Duration
+	startTime    time.Time
+	lastObserved time.Time
+	hasObserved  bool
+}
+
+// NewMaturedAverager creates a new matured averager that wraps the provided
+// averager. It returns zero from Read() until at least halflife duration has
+// passed since the first Observe call.
+func NewMaturedAverager(halflife time.Duration, averager Averager) Averager {
```
joshua-kim: Where does the naming of "matured" come from?

samliok: i use matured to denote the fact that the average waits a certain amount of time before `Observe` does anything. Open to different naming suggestions though

utils/math/matured_averager.go
```diff
@@ -0,0 +1,52 @@
...
+	startTime    time.Time
+	lastObserved time.Time
+	hasObserved  bool
+}
+
+var _ Averager = (*maturedAverager)(nil)
+
+// NewMaturedAverager creates a new matured averager that wraps the provided
+// averager. It returns zero from [averager.Read] until at least `halfLife“ duration has
+// passed since the first Observe call.
```
joshua-kim: Refer to earlier godoc comment https://github.com/ava-labs/avalanchego/pull/4911#discussion_r2765905362
* Square brackets to generate links
* backticks for code references (diff uses a double-quote)

I actually found out that this is broken on go prior to 1.26 (fixed in https://github.com/golang/go/issues/54033)...
```suggestion
// NewMaturedAverager creates a new matured averager that wraps the provided
// `averager`. It returns zero from [Averager.Read] until at least `halfLife` duration has
// passed since the first [Averager.Observe] call.
```

PR #4913, opened 2026-01-27
Add Firewood as a state syncer

alarso16:
## Why this should be merged

This adds all the VM and network layer logic to ensure that state syncing works.

## How this works

Refactors networking for syncing to allow registering the p2p clients, updates config options, and a lot of testing.

## How this was tested

New unit tests.

Not yet, they're not fully supported

### DIFF
--- a/graft/evm/sync/client/test_client.go
+++ b/graft/evm/sync/client/test_client.go
@@ -41,8 +41,6 @@ type TestClient struct {
 	// GetBlocksIntercept is called on every GetBlocks request if set to a non-nil callback.
 	// The returned response will be returned by TestClient to the caller.
 	GetBlocksIntercept func(blockReq message.BlockRequest, blocks types.Blocks) (types.Blocks, error)
-
-	clients map[uint64]*p2p.Client
 }
 
 func NewTestClient(


graft/coreth/sync/client/test_network.go
```diff
@@ -12,9 +12,8 @@ import (
 	"github.com/ava-labs/avalanchego/version"
 )
 
-var _ network.SyncedNetworkClient = (*testNetwork)(nil)
-
 type testNetwork struct {
+	network.Network
```
powerslider: The old code had an explicit compile-time assertion (`var _ network.SyncedNetworkClient = ...`) that was removed. Now `testNetwork` embeds the `Network` interface directly, so any unimplemented method (e.g. `P2PNetwork()`, `P2PValidators()`) compiles fine, but panics at runtime with a nil dereference. 

Consider either adding stub implementations for the new methods or restoring a compile-time assertion with a comment that only the subset of methods used in tests are implemented.

graft/coreth/sync/client/test_client.go
```diff
@@ -47,15 +49,28 @@ func NewTestClient(
...
 		codec:         codec,
 		leafsHandler:  leafHandler,
 		codesHandler:  codesHandler,
 		blocksHandler: blocksHandler,
+		p2pClients:    p2pClients,
 	}
 }
 
+func (ml *TestClient) NewClient(handlerID uint64) *p2p.Client {
+	return ml.p2pClients[handlerID]
```
powerslider: A map lookup on a missing key returns a nil `*p2p.Client`, which will panic somewhere downstream rather than at the point of misconfiguration. Not pretty, but:
```suggestion
	c, ok := ml.p2pClients[handlerID]
	if !ok {
    	panic("NewClient called with unregistered handler ID")
  	}
	return c
```

graft/evm/sync/engine/client.go
```diff
@@ -97,6 +102,8 @@ type ClientConfig struct {
 	Acceptor   BlockAcceptor
 	VerDB      *versiondb.Database
 	MetadataDB database.Database
+	Network    network.Network
```
powerslider: I don't think this is used anywhere so could be removed.

graft/evm/firewood/triedb.go
```diff
@@ -211,6 +211,8 @@ func (t *TrieDB) SetHashAndHeight(blockHash common.Hash, height uint64) {
 	clear(t.tree.blockHashes)
 	t.tree.blockHashes[blockHash] = struct{}{}
 	t.tree.height = height
+	root, _ := t.Firewood.Root()
```
powerslider: This failure is theoretical in practice -`New()` already calls `Root()` and fails construction if it errors, so by the time `SetHashAndHeight` runs the database handle is known-good. There's also no retry path that could call it on a degraded instance.

Suggest we just log instead of silently discarding:

```go
root, err := t.Firewood.Root()
if err != nil {
    log.Error("getting firewood root in SetHashAndHeight", "error", err)
}
```
Does that seem sufficient, or perhaps we should do the full signature change?

graft/subnet-evm/plugin/evm/vm.go
```diff
@@ -643,32 +641,50 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash, ethConfig ethconfig.
...
+		n := vm.Network.P2PNetwork()
+		if err := n.AddHandler(p2p.FirewoodRangeProofHandlerID, syncer.NewGetRangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood range proof handler: %w", err)
+		}
+		if err := n.AddHandler(p2p.FirewoodChangeProofHandlerID, syncer.NewGetChangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood change proof handler: %w", err)
+		}
+	default:
+		// PathDB is not supported.
+	}
```
powerslider, line 687: The `default` case in the state-scheme switch is a silent no-op. If a user configures `path` (PathDB) or any unrecognized scheme, the node will start but silently fail to register any state sync handlers, it won't be able to serve or request state sync data. This is confusing to debug. Please log a warning or return an error.

```suggestion
 	default:
		// PathDB is not supported.
		log.Warn("state sync is not supported for state scheme, no handlers registered", "scheme", scheme)
 	}
```

I might be leaning towards the error option actually, but let's start with a warning first.

graft/evm/firewood/triedb.go
alarso16: ava-labs/firewood#1671
It really doesn't check anything, so logging for now would probably fine

graft/subnet-evm/plugin/evm/vm.go
alarso16: It's not an error though - it just can't state sync. I think a warning is appropriate

graft/coreth/core/chain_makers.go
```diff
@@ -273,6 +273,12 @@ func (b *BlockGen) SetOnBlockGenerated(onBlockGenerated func(*types.Block)) {
 // values. Inserting them into BlockChain requires use of FakePow or
 // a similar non-validating proof of work implementation.
 func GenerateChain(config *params.ChainConfig, parent *types.Block, engine consensus.Engine, db ethdb.Database, n int, gap uint64, gen func(int, *BlockGen)) ([]*types.Block, []types.Receipts, error) {
+	stateCache := state.NewDatabase(db)
+	defer stateCache.TrieDB().Close()
+	return GenerateChainFromStateCache(config, parent, engine, stateCache, n, gap, gen)
+}
+
+func GenerateChainFromStateCache(config *params.ChainConfig, parent *types.Block, engine consensus.Engine, stateCache state.Database, n int, gap uint64, gen func(int, *BlockGen)) ([]*types.Block, []types.Receipts, error) {
```
alarso16: this needed add because the basic `GenerateChain` function relies on using hashdb, where the nodes can be easily read and hashed without altering database state. To avoid this in Firewood, we can provide all necessary state abstractions as a copy.

graft/coreth/plugin/evm/config/config.md
```diff
@@ -221,7 +221,7 @@ Configuration is provided as a JSON object. All fields are optional unless other
 
 | Option | Type | Description | Default |
 |--------|------|-------------|---------|
-| `metrics-expensive-enabled` | bool | Enable expensive debug-level metrics; this includes Firewood metrics | `true` |
+| `metrics-expensive-enabled` | bool | Enable expensive debug-level metrics | `true` |
```
alarso16, line 224: This just wasn't very informative

graft/coreth/plugin/evm/vm.go
```diff
@@ -582,36 +580,51 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
...
+			return fmt.Errorf("expected a %T with %s scheme, got %T", tdb, customrawdb.FirewoodScheme, vm.eth.BlockChain().TrieDB().Backend())
+		}
+		n := vm.Network.P2PNetwork()
+		if err := n.AddHandler(p2p.FirewoodRangeProofHandlerID, syncer.NewGetRangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood range proof handler: %w", err)
+		}
+		if err := n.AddHandler(p2p.FirewoodChangeProofHandlerID, syncer.NewGetChangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood change proof handler: %w", err)
+		}
+	default:
```
JonathanOppenheimer, line 624: Is this intended behavior -- shouldn't this fail harder? If we return early, we don't set `vm.Server` or `vm.Client uninitialized,` but later we'll call on these interfaces (e.g. vm.Client.ClearOngoingSummary()`). 

So if we used "path" (which is supported right), the VM could initialize but then panic because the interfaces are nil.

graft/subnet-evm/plugin/evm/vm.go
```diff
@@ -643,32 +641,51 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash, ethConfig ethconfig.
...
+		}
+		n := vm.Network.P2PNetwork()
+		if err := n.AddHandler(p2p.FirewoodRangeProofHandlerID, syncer.NewGetRangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood range proof handler: %w", err)
+		}
+		if err := n.AddHandler(p2p.FirewoodChangeProofHandlerID, syncer.NewGetChangeProofHandler(tdb.Firewood)); err != nil {
+			return fmt.Errorf("adding firewood change proof handler: %w", err)
+		}
+	default:
+		log.Warn("state sync is not supported for this scheme, no leaf handlers will be registered", "scheme", scheme)
```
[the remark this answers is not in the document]
JonathanOppenheimer: There could also be a test to catch this maybe?

graft/coreth/plugin/evm/config/config.md
```diff
@@ -259,7 +259,6 @@ Configuration is provided as a JSON object. All fields are optional unless other
 > **WARNING**: `firewood` and `path` schemes are untested in production. Using `path` is strongly discouraged. To use `firewood`, you must also set the following config options:
 >
 > - `populate-missing-tries: nil`
-> - `state-sync-enabled: false`
 > - `snapshot-cache: 0`
```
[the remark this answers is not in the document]
alarso16: I actually think we can't advertise it until range proofs are fully implemented.

graft/subnet-evm/plugin/evm/vm.go
JonathanOppenheimer, line 423: Here, we explicitly reject path before state-sync init, so we do not get the same error we do here: https://github.com/ava-labs/avalanchego/pull/4913/changes#r2830523448

graft/coreth/plugin/evm/vm.go
JonathanOppenheimer: Maybe we can add a test to catch this?

alarso16: Path isn't really supported. but there is a test that uses it, so it must not panic for bootstrapping

graft/coreth/plugin/evm/vm.go
```diff
@@ -582,36 +580,51 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
...
+		log.Warn("state sync is not supported for this scheme, no leaf handlers will be registered", "scheme", scheme)
+		return nil
```
powerslider: Make the `default` case return an error in regards to https://github.com/ava-labs/avalanchego/pull/4913/changes#r2864541003

```suggestion
        return fmt.Errorf("state sync initialization is not supported for state scheme %q", scheme)
```

graft/coreth/plugin/evm/vm.go
powerslider, line 409: Use the error here is regards to https://github.com/ava-labs/avalanchego/pull/4913/changes#r2864541003
```suggestion
        log.Error("Path state scheme is not supported. Please use HashDB or Firewood state schemes instead")
		return errPathStateUnsupported
```

graft/coreth/plugin/evm/vm.go
```diff
@@ -142,7 +144,6 @@ var (
 	errShuttingDownVM                             = errors.New("shutting down VM")
 	errFirewoodSnapshotCacheDisabled              = errors.New("snapshot cache must be disabled for Firewood")
 	errFirewoodOfflinePruningUnsupported          = errors.New("offline pruning is not supported for Firewood")
-	errFirewoodStateSyncUnsupported               = errors.New("state sync is not yet supported for Firewood")
 	errFirewoodMissingTrieRepopulationUnsupported = errors.New("missing trie repopulation is not supported for Firewood")
 )
```
powerslider, line 148: `subnet-evm` already rejects `path` with `return errPathStateUnsupported`. `coreth` currently only logs a warning and continues - this needs to match:
```
errPathStateUnsupported                       = errors.New("path state scheme is not supported")
```

graft/evm/firewood/triedb.go
```diff
@@ -211,6 +211,12 @@ func (t *TrieDB) SetHashAndHeight(blockHash common.Hash, height uint64) {
 	clear(t.tree.blockHashes)
 	t.tree.blockHashes[blockHash] = struct{}{}
 	t.tree.height = height
+	root, err := t.Firewood.Root()
+	if err != nil {
+		log.Error("get root in SetHashAndHeight", "error", err)
+		return
+	}
```
powerslider: `SetHashAndHeight` modifies `t.tree.blockHashes` and `t.tree.height` before calling `t.Firewood.Root()`. 

If `Root()` fails, the method returns early with `blockHashes` and `height` updated but `root` still holding its previous (now stale) value. 

This leaves the proposal tree in an inconsistent state - the height says one thing but the root says another.

I would suggest call `Root()` first, before modifying any in-memory state. If it fails, return early without touching anything. Just move this block of code three lines above to prevent this inconsistency. This ensures the method is atomic: either all state is updated (including the root from Firewood), or none of it is.

alarso16: I need to rebase - Root doesn't actually ever return an error...

graft/coreth/plugin/evm/vm.go
alarso16: As much sense as this would make, I added this for performance testing, a request of the testing team.

graft/coreth/plugin/evm/vm.go
alarso16: Can't do that, since this method is required. And we can still serve atomic trie requests

alarso16: I can continue and just no register any other handlers though. If you're using an unofficially supported state scheme, you should be able to start up still

graft/coreth/core/chain_makers.go
```diff
@@ -273,6 +273,14 @@ func (b *BlockGen) SetOnBlockGenerated(onBlockGenerated func(*types.Block)) {
...
 // a similar non-validating proof of work implementation.
 func GenerateChain(config *params.ChainConfig, parent *types.Block, engine consensus.Engine, db ethdb.Database, n int, gap uint64, gen func(int, *BlockGen)) ([]*types.Block, []types.Receipts, error) {
+	stateCache := state.NewDatabase(db)
+	defer stateCache.TrieDB().Close()
+	return GenerateChainFromStateCache(config, parent, engine, stateCache, n, gap, gen, true)
+}
+
+// GenerateChainFromStateCache is exactly like [GenerateChain], except allows other [triedb.Database] implementations.
+// The commit parameter controls whether the state changes are committed to the database.
+func GenerateChainFromStateCache(config *params.ChainConfig, parent *types.Block, engine consensus.Engine, stateCache state.Database, n int, gap uint64, gen func(int, *BlockGen), commit bool) ([]*types.Block, []types.Receipts, error) {
```
alarso16: I apologize for how disgusting this fix is. i don't know how else to easily create blocks that can be wrapped and verified/accepted separately

graft/coreth/plugin/evm/vm.go
alarso16: We should be able to start up a node with PathDB for performance testing, which includes registering a sync client. If you disagree with any of this, we can change it, but it's out of scope for this PR IMO

graft/coreth/plugin/evm/vmtest/test_syncervm.go
```diff
@@ -326,8 +346,24 @@ func initSyncServerAndClientVMs(t *testing.T, test SyncTestParams, numBlocks int
...
-	root, accounts := synctest.FillAccountsWithOverlappingStorage(t, r, serverVM.Ethereum().BlockChain().StateCache(), types.EmptyRootHash, 1000, 16)
+	var (
+		r                           = rand.New(rand.NewSource(1))
+		currentRoot                 = serverVM.Ethereum().BlockChain().LastAcceptedBlock().Root()
+		root                        common.Hash
+		fundedAccounts, allAccounts map[*utilstest.Key]*types.StateAccount
+	)
+	if test.StateScheme == customrawdb.FirewoodScheme {
+		tdb, ok := serverVM.Ethereum().BlockChain().TrieDB().Backend().(*firewood.TrieDB)
+		require.True(ok)
```
JonathanOppenheimer: For these `require.True` you should have a message (as Arran likes)

graft/subnet-evm/network/network.go
```diff
@@ -157,6 +158,31 @@ func NewNetwork(
 	}, nil
 }
 
+// Sample returns a random sample of connected peers.
+// `limit` is ignored, and one peer will be returned.
```
JonathanOppenheimer, line 162: Why did you design it this way? Just curious.

graft/evm/sync/client/client.go
```diff
@@ -82,6 +88,12 @@ type Client interface {
 
 	// GetCode synchronously retrieves code associated with the given hashes
 	GetCode(ctx context.Context, hashes []common.Hash) ([][]byte, error)
+
+	// AddClient creates a separate client on the underlying [p2p.Network].
+	AddClient(handlerID uint64) *p2p.Client
+
+	// StateSyncNodes returns the list of nodes provided via config.
+	StateSyncNodes() []ids.NodeID
```
JonathanOppenheimer, line 96: Channeling Josh here, but is this interface bloat? Both of these are only used by firewood.

graft/evm/sync/engine/client.go
```diff
@@ -394,14 +399,36 @@ func (c *client) newSyncerRegistry(summary message.Syncable) (*SyncerRegistry, e
...
-		return nil, fmt.Errorf("failed to create EVM state syncer: %w", err)
+	var stateSyncer types.Syncer
+	if tdb, ok := c.config.Chain.BlockChain().TrieDB().Backend().(*firewood.TrieDB); ok {
+		registerer, err := metrics.MakeAndRegister(c.config.SnowCtx.Metrics, "sync_firewood")
+		if err != nil {
+			return nil, fmt.Errorf("failed to create firewood syncer metrics registerer: %w", err)
+		}
+		stateSyncer, err = evmstate.NewFirewoodSyncer(
+			syncer.Config{
+				Log:            c.config.SnowCtx.Log,
```
JonathanOppenheimer, line 411: Would it ever be possible for the `SnowCtx` to be `nil`?

graft/evm/sync/evmstate/firewood_syncer.go
```diff
@@ -33,14 +34,14 @@ type FirewoodSyncer struct {
...
+func NewFirewoodSyncer(config syncer.Config, db *ffi.Database, target common.Hash, codeQueue *code.Queue, client client.Client) (*FirewoodSyncer, error) {
 	s, err := syncer.NewEVM(
 		config,
 		db,
 		codeQueue,
 		ids.ID(target),
-		rangeProofClient,
-		changeProofClient,
+		client.AddClient(p2p.FirewoodRangeProofHandlerID),
+		client.AddClient(p2p.FirewoodChangeProofHandlerID),
```
JonathanOppenheimer: Why did you change this? Now we accept the full `client.Client` just to call AddClient twice -- you're accepting a wider interface than needed.

graft/evm/sync/engine/client.go
```diff
@@ -394,14 +399,36 @@ func (c *client) newSyncerRegistry(summary message.Syncable) (*SyncerRegistry, e
...
-	stateSyncer, err := evmstate.NewSyncer(
-		c.config.Client, c.config.ChainDB,
-		summary.GetBlockRoot(),
-		codeQueue, c.config.RequestSize,
-		c.config.LeafsRequestType,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create EVM state syncer: %w", err)
+	var stateSyncer types.Syncer
+	if tdb, ok := c.config.Chain.BlockChain().TrieDB().Backend().(*firewood.TrieDB); ok {
```
JonathanOppenheimer, line 404: branching on the scheme string would be more explicit and probably make more sense here -- rather than having the syncer know what firewood is like.

graft/evm/sync/synctest/trie.go
```diff
@@ -210,13 +210,15 @@ func FillAccounts(
...
 func FillAccountsWithStorageAndCode(t *testing.T, r *rand.Rand, serverDB state.Database, root common.Hash, numAccounts int) (common.Hash, map[*utilstest.Key]*types.StateAccount) {
 	return FillAccounts(t, r, serverDB, root, numAccounts, func(t *testing.T, _ int, addr common.Address, account types.StateAccount, storageTr state.Trie) types.StateAccount {
-		codeBytes := make([]byte, 256)
-		_, err := r.Read(codeBytes)
-		require.NoError(t, err, "error reading random code bytes")
-
-		codeHash := crypto.Keccak256Hash(codeBytes)
-		rawdb.WriteCode(serverDB.DiskDB(), codeHash, codeBytes)
-		account.CodeHash = codeHash[:]
+		if r.Intn(2) == 0 {
```
JonathanOppenheimer, line 215: Why was this changed -- now we randomly fill half the accounts?

graft/subnet-evm/plugin/evm/vm.go
```diff
@@ -643,32 +641,50 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash, ethConfig ethconfig.
...
-			DBOverride: hashdb.Config{
-				CleanCacheSize: vm.config.StateSyncServerTrieCache * units.MiB,
-			}.BackendConstructor,
-		},
-	)
-
-	// register default leaf request handler for state trie
+	leafHandlers := make(LeafHandlers)
+	leafMetricsNames := make(map[message.NodeType]string)
 	syncStats := handlerstats.GetOrRegisterHandlerStats(metrics.Enabled)
```
JonathanOppenheimer, line 646: would it make sense to only allocate declare this where it's used? you could even inline it -- it's only used in the `HashScheme` path.

graft/subnet-evm/network/network.go
alarso16: I avoided adding additional functionality when it wasn't necessary. `limit` does clearly mean that you just can't provide more. Additionally, the syncer only ever asks for one

graft/evm/sync/engine/client.go
alarso16: We have to get the `Firewood` element in the struct eventually anyway, so we need to do this cast eventually

graft/evm/sync/engine/client.go
alarso16: I hope not! the `Config` struct is not exactly a config in the first place, since there are critical things in here. Almost all of the elements in the struct are not configurations, and must be non-nil. 

You're right that this isn't great, but it's definitely out of scope from this PR to cleanup the config abstraction

graft/evm/sync/client/client.go
alarso16: This was the most confusing part of making this PR. I tried very hard to come up with a better abstraction, but couldn't. Any ideas?

graft/evm/sync/engine/client.go
JonathanOppenheimer: something something clean up while you go along, but valid for out of scope - I think that's the theme with most of these comments. feel free to resolve all you're not going to change anything for!

graft/evm/sync/evmstate/firewood_syncer.go
alarso16: I thought it was clearer, but if it's not...

graft/evm/sync/synctest/trie.go
alarso16: This is not at all trivial - took me some experimenting to remember why. The tests create lots of balance transfers, but they can only be initiated by EOAs, not contracts. If the code hash is non-empty, then it's a contract. Probably, it should also not have any storage (so I'll fix that too)

JonathanOppenheimer: If you could leave a comment as well!

alarso16: I added a comment to clarify this.

graft/subnet-evm/plugin/evm/vm.go
alarso16: They're still passed to the network handler

graft/evm/sync/evmstate/firewood_syncer.go
JonathanOppenheimer: I think this is worse tbh, but if others (or you) disagree, then valid

graft/evm/sync/client/client.go
JonathanOppenheimer: maybe you could pass a factory function instead of widening the interface? then you also wouldn't need test stubs on the test client.

alarso16: factory for what?

JonathanOppenheimer: Rather than 
```go
func NewFirewoodSyncer(config syncer.Config, db *ffi.Database, target common.Hash,
  codeQueue *code.Queue, client client.Client) (*FirewoodSyncer, error)
```

you could do something like 
```go
func NewFirewoodSyncer(config syncer.Config, db *ffi.Database, target common.Hash,
  codeQueue *code.Queue, newClient func(handlerID uint64) *p2p.Client) (*FirewoodSyncer, error)
```

which would allow you to do: 

```go
rangeProofClient := newClient(p2p.FirewoodRangeProofHandlerID)
changeProofClient := newClient(p2p.FirewoodChangeProofHandlerID)
```

if you don't like the other comment -- you could also keep that is. The same could be said for `StateSyncNodes` - it's already on `syncer.Config`, seems weird to bring it up to `Client`. This is just an idea to avoid touching the interface -- it might not be better.

graft/evm/sync/evmstate/firewood_syncer.go
alarso16: Done

joshua-kim (approved):

PR #4912, opened 2026-01-27
refactor: de-duplicate docker image build scripts

JonathanOppenheimer:
## Why this should be merged

Closes #4678. Consolidates the separate avalanchego and subnet-evm Docker build pipelines into a single multi-stage Dockerfile and shared build script.

## How this works

- The root `Dockerfile` gains subnet-evm build and final stages alongside the existing avalanchego stages. Both share a common base stage (Go setup, dependency download, cross-compilation config).
- `scripts/build_image.sh` accepts a `TARGET` env var (`avalanchego` or `subnet-evm`, default `avalanchego`) to select which image to build. Target-specific config (image name defaults, build args, base image resolution) is handled in a case block; everything else (buildx setup, push/load logic, tagging) is shared.
- For subnet-evm, the script automatically builds the local avalanchego base image if it doesn't exist, or accepts `AVALANCHEGO_NODE_IMAGE` / `AVALANCHEGO_NODE_IMAGE_REPO` for CI.
- `graft/subnet-evm/Dockerfile` and `graft/subnet-evm/scripts/build_docker_image.sh` are deleted.
- Antithesis build scripts pass a `--target` flag when building uninstrumented node images from the unified Dockerfile on arm64.
- Removes unused `ALLOW_TAG_LATEST` env var from CI workflow and dead constants from subnet-evm.

## How this was tested

CI

## Need to be documented in RELEASES.md?

No

PR #4914, opened 2026-01-27
refactor: share constants between avalanchego and subnet-evm

### LINKED ISSUE #4681: Clean up subnet-evm and coreth `/scripts/constants.sh`
Maybe file an issue to clean this file up further? Most of it should come from ~/scripts/constants.sh and there won't be a need to discover the avalanchego version.

_Originally posted by @maru-ava in https://github.com/ava-labs/avalanchego/pull/4671#discussion_r2611065704_

JonathanOppenheimer:
## Why this should be merged

Closes #4681 

## How this works

Sources constants from AvalancheGo constants rather than re-defining. 

## How this was tested
CI 

No

### DIFF
--- a/snow/consensus/snowman/metrics.go
+++ b/snow/consensus/snowman/metrics.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/linked"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/utils/metric"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
 )
@@ -61,6 +62,9 @@ type metrics struct {
 
 	// numSuccessfulPolls keeps track of the number of polls that succeeded
 	numSuccessfulPolls prometheus.Counter
+
+	// avgAcceptanceLatency tracks the average acceptance time
+	avgAcceptanceLatency math.Averager
 }
 
 func newMetrics(
@@ -69,6 +73,7 @@ func newMetrics(
 	lastAcceptedHeight uint64,
 	lastAcceptedTime time.Time,
 ) (*metrics, error) {
+	acceptanceHalfLife := 5 * time.Minute
 	errs := wrappers.Errs{}
 	m := &metrics{
 		log:                      log,
@@ -144,6 +149,10 @@ func newMetrics(
 			Name: "polls_failed",
 			Help: "number of failed polls",
 		}),
+		avgAcceptanceLatency: math.NewMaturedAverager(
+			acceptanceHalfLife,
+			math.NewUninitializedAverager(acceptanceHalfLife),
+		),
 	}
 
 	// Initially set the metrics for the last accepted block.
@@ -209,7 +218,7 @@ func (m *metrics) Accepted(
 
 	builtDuration := now.Sub(timestamp)
 	m.buildLatencyAccepted.Add(float64(builtDuration))
-
+	m.avgAcceptanceLatency.Observe(float64(builtDuration), now)
 	m.consensusLatencies.Observe(float64(processingDuration))
 }
 
@@ -247,4 +256,8 @@ func (m *metrics) SuccessfulPoll() {
 
 func (m *metrics) FailedPoll() {
 	m.numFailedPolls.Inc()
+}
+
+func (m *metrics) GetAverageAcceptanceTime() time.Duration {
+	return time.Duration(m.avgAcceptanceLatency.Read())
 }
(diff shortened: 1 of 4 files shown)


graft/subnet-evm/scripts/build_docker_image.sh
```diff
@@ -13,106 +13,99 @@ set -euo pipefail
 # Reference: https://docs.docker.com/build/building/multi-platform/
 PLATFORMS="${PLATFORMS:-}"
 
-# If set to non-empty, the image will be published to the registry.
```
maru-ava, line 16: - Why are the non-functional changes to this file suggested?
- Does this script need to continue existing or could it be replaced by a version of the avalanchego image build script refactored for reuse?

JonathanOppenheimer: - I thought the current comments could be cleaned up while updating the constants usage. 
- I shared the docker_build [script in this follow up PR ](https://github.com/ava-labs/avalanchego/pull/4912/changes). I found it would be more complicated to try to condense everything into a single script. However if that approach is preferred, I will take another crack at that draft PR to try to use a single script, after this PR is merged.

maru-ava: While cleanup is appreciated, I encourage you to separate functional from non-functional changes where possible to minimize the effort required of your reviewers. I'm also not sure this script is worth cleaning up given that it largely duplicates avalanchego's image build script and as we've discussed will likely go away in the not-too-distant future.

JonathanOppenheimer: That's totally fair -- I apologize I let creep get to me. I've reverted all the non-functional cleanup in `build_docker_image.sh`, while keeping only the functional changes needed to use the newly shared constants from the root scripts.

I will update the next PR to completely condense the build script into a single one with the advice you gave me.

graft/subnet-evm/scripts/build_docker_image.sh
```diff
@@ -103,18 +103,19 @@ fi
...
   --build-arg GO_VERSION="${GO_VERSION}" \
   --build-arg AVALANCHEGO_NODE_IMAGE="$AVALANCHEGO_NODE_IMAGE" \
-  --build-arg SUBNET_EVM_COMMIT="$SUBNET_EVM_COMMIT" \
-  --build-arg CURRENT_BRANCH="$CURRENT_BRANCH" \
+  --build-arg SUBNET_EVM_COMMIT="$git_commit" \
+  --build-arg CURRENT_BRANCH="$image_tag" \
   --build-arg VM_ID="$VM_ID"
 
-if [[ -n "${PUBLISH}" && $CURRENT_BRANCH == "master" && $ALLOW_TAG_LATEST == true ]]; then
+if [[ -n "${PUBLISH}" && $image_tag == "master" && $ALLOW_TAG_LATEST == true ]]; then
```
maru-ava, line 116: (No action required) Not for this PR, just wanted to point out that subnet-evm is currently tagging latest on any merge to master but as per the example of avalanchego should be tagging latest only for stable releases.

graft/subnet-evm/scripts/constants.sh
```diff
@@ -5,75 +5,30 @@
...
+# Subnet-EVM specific constants
 GOPATH="$(go env GOPATH)"
 DEFAULT_PLUGIN_DIR="${HOME}/.avalanchego/plugins"
 DEFAULT_VM_NAME="subnet-evm"
 DEFAULT_VM_ID="srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy"
 
-# Avalabs docker hub
-# avaplatform/avalanchego - defaults to local as to avoid unintentional pushes
-# You should probably set it - export IMAGE_NAME='avaplatform/subnet-evm_avalanchego'
-IMAGE_NAME=${IMAGE_NAME:-"subnet-evm_avalanchego"}
```
ceyonur, line 17: this could be a breaking change, have discussed this with other teams?

graft/subnet-evm/scripts/constants.sh
```diff
@@ -5,75 +5,30 @@
...
 
-# Set the PATHS
+SUBNET_EVM_PATH=$(
+  cd "$(dirname "${BASH_SOURCE[0]}")"
+  cd .. && pwd
+)
+
+# Source common constants from root
+# shellcheck disable=SC1091
+source "$SUBNET_EVM_PATH"/../../scripts/constants.sh
```
ceyonur: can/should we define a AVALANCHEGO_ROOT? I find it very ugly to do .. everytime to go upper level.

graft/subnet-evm/scripts/constants.sh
JonathanOppenheimer: If someone relies on on the docker images, with CI or something that doesn't set `IMAGE_NAME`, than yeah the name could be different and things could break. I don't think this is a very big concern though, and the fix would be very simple. What other teams rely on these builds by name?

graft/subnet-evm/scripts/constants.sh
JonathanOppenheimer: Yeah this is actually an incredibly good change. Done!

graft/subnet-evm/scripts/build_docker_image.sh
JonathanOppenheimer: I made a separate PR to correct this! https://github.com/ava-labs/avalanchego/pull/4951

PR #4917, opened 2026-01-28
Add proposervm block delay logs

yacovm:
## Why this should be merged

Adds some innocuous logging statements to proposerVM's mechanism of determining minimum delay.

## How this works

Adds logging statements

## How this was tested

CI

### DIFF
--- a/vms/proposervm/proposer/windower.go
+++ b/vms/proposervm/proposer/windower.go
@@ -205,7 +205,7 @@ func (w *windower) MinDelayForProposer(
 		return 0, err
 	}
 
-	w.logger.Debug("Sampled validators for P-chain height",
+	w.logger.Debug("sampled validators for P-chain height",
 		zap.Uint64("pChainHeight", pChainHeight), zap.Int("numValidators", len(validators)))
 
 	if len(validators) == 0 {
--- a/vms/proposervm/vm.go
+++ b/vms/proposervm/vm.go
@@ -577,17 +577,17 @@ func (vm *VM) getPostDurangoSlotTime(
 	// validators can specify. This delay may be an issue for high performance,
 	// custom VMs. Until the P-chain is modified to target a specific block
 	// time, ProposerMinBlockDelay can be configured in the node config.
-	if err != nil {
-		vm.ctx.Log.Debug("Minimum delay", zap.Duration("delay", delay), zap.Error(err))
-	}
 
 	switch {
 	case err == nil:
+		vm.ctx.Log.Debug("slot time", zap.Duration("delay", delay), zap.Duration("min block delay", vm.MinBlkDelay))
 		delay = max(delay, vm.MinBlkDelay)
 		return parentTimestamp.Add(delay), nil
 	case errors.Is(err, proposer.ErrAnyoneCanPropose):
+		vm.ctx.Log.Debug("anyone can propose", zap.Duration("min block delay", vm.MinBlkDelay), zap.Time("parent timestamp", parentTimestamp))
 		return parentTimestamp.Add(vm.MinBlkDelay), nil
 	default:
+		vm.ctx.Log.Debug("failed fetching the expected delay", zap.Error(err))
 		return time.Time{}, err
 	}
 }


vms/proposervm/vm.go
```diff
@@ -570,13 +570,17 @@ func (vm *VM) getPostDurangoSlotTime(
...
+		vm.ctx.Log,
 	)
 	// Note: The P-chain does not currently try to target any block time. It
 	// notifies the consensus engine as soon as a new block may be built. To
 	// avoid fast runs of blocks there is an additional minimum delay that
 	// validators can specify. This delay may be an issue for high performance,
 	// custom VMs. Until the P-chain is modified to target a specific block
 	// time, ProposerMinBlockDelay can be configured in the node config.
+
+	vm.ctx.Log.Debug("proposervm minimum delay", zap.Duration("delay", delay), zap.Error(err))
```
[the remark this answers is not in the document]
yacovm: This is intentional.

maru-ava (approved):

vms/proposervm/proposer/windower.go
```diff
@@ -92,6 +94,7 @@ type Windower interface {
 		pChainHeight uint64,
 		nodeID ids.NodeID,
 		startSlot uint64,
+		logger logging.Logger,
```
joshua-kim: The output of this function calculating the minimum delay is independent of the logger -- instead of parameterizing could we pass it into `New` since the logger will always be the same for the lifetime of the windower?

vms/proposervm/proposer/windower_test.go
```diff
@@ -303,7 +304,6 @@ func TestExpectedProposerChangeBySlot(t *testing.T) {
 
 func TestCoherenceOfExpectedProposerAndMinDelayForProposer(t *testing.T) {
 	require := require.New(t)
-
```
joshua-kim, line 306: Unrelated diff

vms/proposervm/vm.go
```diff
@@ -570,13 +570,17 @@ func (vm *VM) getPostDurangoSlotTime(
...
+		vm.ctx.Log,
 	)
 	// Note: The P-chain does not currently try to target any block time. It
 	// notifies the consensus engine as soon as a new block may be built. To
 	// avoid fast runs of blocks there is an additional minimum delay that
 	// validators can specify. This delay may be an issue for high performance,
 	// custom VMs. Until the P-chain is modified to target a specific block
 	// time, ProposerMinBlockDelay can be configured in the node config.
+
+	vm.ctx.Log.Debug("ProposerVM minimum delay", zap.Duration("delay", delay), zap.Error(err))
```
joshua-kim: * Doesn't this log the error key even when it's nil? We could log differently based on the error value in the switch to make this less noisy.
* Our logs typically follow a lowercase format -- I think `proposervm minimum delay` is what we would expect here

vms/proposervm/proposer/windower.go
yacovm: Yeah you're right, sloppy from my side 😁

vms/proposervm/proposer/windower_test.go
yacovm: 🦅 👁️

vms/proposervm/vm.go
```diff
@@ -577,6 +577,10 @@ func (vm *VM) getPostDurangoSlotTime(
 	// validators can specify. This delay may be an issue for high performance,
 	// custom VMs. Until the P-chain is modified to target a specific block
 	// time, ProposerMinBlockDelay can be configured in the node config.
+	if err != nil {
+		vm.ctx.Log.Debug("Minimum delay", zap.Duration("delay", delay), zap.Error(err))
```
joshua-kim: 1. Refer to comment regarding lowercase log format: https://github.com/ava-labs/avalanchego/pull/4917#discussion_r2739378867
2. Should this log be in the switch and maybe we should be logging the return value? In the case of `proposer.ErrAnyoneCanPropose` the delay is `vm.MinBlkDelay` and not `delay`

vms/proposervm/proposer/windower.go
```diff
@@ -200,6 +204,10 @@ func (w *windower) MinDelayForProposer(
 	if err != nil {
 		return 0, err
 	}
+
+	w.logger.Debug("Sampled validators for P-chain height",
```
joshua-kim: Refer to comment regarding lowercase log format: https://github.com/ava-labs/avalanchego/pull/4917#discussion_r2739378867

joshua-kim (approved):

vms/proposervm/vm.go
```diff
@@ -577,13 +577,17 @@ func (vm *VM) getPostDurangoSlotTime(
 	// validators can specify. This delay may be an issue for high performance,
 	// custom VMs. Until the P-chain is modified to target a specific block
 	// time, ProposerMinBlockDelay can be configured in the node config.
+
 	switch {
 	case err == nil:
+		vm.ctx.Log.Debug("slot time", zap.Duration("delay", delay), zap.Duration("min block delay", vm.MinBlkDelay))
```
joshua-kim, line 583: I wanted to comment on this because I thought it would cause issues but it actually turns out that spaces are permitted in json keys -- so I think this does not cause issues w/ structured logs

PR #4920, opened 2026-01-29
Update consensus logs

StephenButtolph:
## Why this should be merged

I documented my rationale at the related diffs.

## How this works

This should be a strictly logging related change.

## How this was tested

Existing tests.

No

### DIFF
--- a/vms/proposervm/proposer/windower.go
+++ b/vms/proposervm/proposer/windower.go
@@ -9,11 +9,13 @@ import (
 	"math/bits"
 	"time"
 
+	"go.uber.org/zap"
 	"gonum.org/v1/gonum/mathext/prng"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/avalanchego/utils/sampler"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
@@ -101,11 +103,13 @@ type windower struct {
 	state       validators.State
 	subnetID    ids.ID
 	chainSource uint64
+	logger      logging.Logger
 }
 
-func New(state validators.State, subnetID, chainID ids.ID) Windower {
+func New(state validators.State, subnetID, chainID ids.ID, logger logging.Logger) Windower {
 	w := wrappers.Packer{Bytes: chainID[:]}
 	return &windower{
+		logger:      logger,
 		state:       state,
 		subnetID:    subnetID,
 		chainSource: w.UnpackLong(),
@@ -200,6 +204,10 @@ func (w *windower) MinDelayForProposer(
 	if err != nil {
 		return 0, err
 	}
+
+	w.logger.Debug("sampled validators for P-chain height",
+		zap.Uint64("pChainHeight", pChainHeight), zap.Int("numValidators", len(validators)))
+
 	if len(validators) == 0 {
 		return 0, ErrAnyoneCanPropose
 	}
--- a/vms/proposervm/proposer/windower_test.go
+++ b/vms/proposervm/proposer/windower_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/validators"
 	"github.com/ava-labs/avalanchego/snow/validators/validatorstest"
+	"github.com/ava-labs/avalanchego/utils/logging"
 
 	safemath "github.com/ava-labs/avalanchego/utils/math"
 )
@@ -48,6 +49,7 @@ func TestWindowerNoValidators(t *testing.T) {
 				makeValidatorState(t, test.validators),
 				subnetID,
 				randomChainID,
+				&logging.NoLog{},
 			)
 
 			var (
@@ -91,7 +93,7 @@ func TestWindowerRepeatedValidator(t *testing.T) {
 		},
 	}
 
-	w := New(vdrState, subnetID, randomChainID)
+	w := New(vdrState, subnetID, randomChainID, &logging.NoLog{})
 
 	validatorDelay, err := w.Delay(t.Context(), 1, 0, validatorID, MaxVerifyWindows)
 	require.NoError(err)
@@ -106,7 +108,7 @@ func TestDelayChangeByHeight(t *testing.T) {
 	require := require.New(t)
 
 	validatorIDs, vdrState := makeValidators(t, MaxVerifyWindows)
-	w := New(vdrState, subnetID, fixedChainID)
+	w := New(vdrState, subnetID, fixedChainID, &logging.NoLog{})
 
 	expectedDelays1 := []time.Duration{
 		2 * WindowDuration,
@@ -154,8 +156,8 @@ func TestDelayChangeByChain(t *testing.T) {
 	require.NoError(err)
 
 	validatorIDs, vdrState := makeValidators(t, MaxVerifyWindows)
-	w0 := New(vdrState, subnetID, chainID0)
-	w1 := New(vdrState, subnetID, chainID1)
+	w0 := New(vdrState, subnetID, chainID0, &logging.NoLog{})
+	w1 := New(vdrState, subnetID, chainID1, &logging.NoLog{})
 
 	expectedDelays0 := []time.Duration{
 		5 * WindowDuration,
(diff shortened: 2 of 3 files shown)


snow/consensus/snowman/poll/set.go
```diff
@@ -152,7 +152,7 @@ func (s *set) processFinishedPolls() []bag.Bag[ids.ID] {
 			break
 		}
 
-		s.log.Verbo("poll finished",
+		s.log.Debug("poll finished",
```
StephenButtolph, line 155: This was a duplicate of the log in the voter. I deleted the other log and increased the log level here so that the log would include the `requestID`.

snow/consensus/snowman/topological.go
```diff
@@ -144,9 +144,12 @@ func (ts *Topological) NumProcessing() int {
 func (ts *Topological) Add(blk Block) error {
 	blkID := blk.ID()
 	height := blk.Height()
-	ts.ctx.Log.Verbo("adding block",
+	parentID := blk.Parent()
+	ts.ctx.Log.Trace("adding block",
```
StephenButtolph, line 148: I promoted this to Trace (between debug and info) to align with the accept and reject calls.

snow/engine/snowman/syncer/state_syncer.go
```diff
@@ -191,16 +191,11 @@ func (ss *stateSyncer) StateSummaryFrontier(ctx context.Context, nodeID ids.Node
...
-				zap.Binary("summary", summaryBytes),
-				zap.Error(err),
-			)
-		} else {
-			ss.Ctx.Log.Debug("failed to parse summary",
-				zap.Error(err),
-			)
+			fields = append(fields, zap.Binary("summary", summaryBytes))
 		}
+		ss.Ctx.Log.Debug("failed to parse summary", fields...)
```
StephenButtolph, line 198: It was weird to decrease the log level of the log just because the field now included more data. Now only the field is added.

snow/engine/snowman/voter.go
```diff
@@ -56,9 +54,6 @@ func (v *voter) Execute(ctx context.Context, _ []ids.ID, _ []ids.ID) error {
 	}
 
 	for _, result := range results {
-		v.e.Ctx.Log.Debug("finishing poll",
```
StephenButtolph, line 59: Removed to prefer the log with the requestID attached.

snow/networking/handler/handler.go
```diff
@@ -443,17 +443,18 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
...
-		h.ctx.Log.Debug("forwarding sync message to consensus",
+	{
+		fields := []zap.Field{
 			zap.Stringer("nodeID", nodeID),
 			zap.String("messageOp", op),
-		)
+		}
+		if requestID, ok := message.GetRequestID(body); ok {
+			fields = append(fields, zap.Uint32("requestID", requestID))
+		}
```
StephenButtolph, line 453: This was the big add here, logging the requestID for connecting what triggered the event is critical for actually tracing events.

snow/networking/handler/handler.go
```diff
@@ -475,7 +476,7 @@ func (h *handler) handleSyncMsg(ctx context.Context, msg Message) error {
 		h.metrics.messageHandlingTime.With(labels).Add(float64(handlingTime))
 
 		msg.OnFinishedHandling()
-		h.ctx.Log.Debug("finished handling sync message",
+		h.ctx.Log.Verbo("finished handling sync message",
```
StephenButtolph, line 479: This log just spammed everywhere and was totally useless imo

snow/networking/sender/sender.go
```diff
@@ -80,559 +80,392 @@ func New(
...
+				deadline,
+				s.ctx.NodeID,
+			),
 		)
-		s.router.HandleInternal(ctx, inMsg)
 	}
 
-	// Create the outbound message.
-	outMsg, err := s.msgCreator.GetStateSummaryFrontier(
+	log := s.ctx.Log.With(
```
StephenButtolph, line 116: The duplication between the different log lines was driving me crazy.

snow/networking/sender/sender.go
```diff
@@ -1151,359 +850,258 @@ func (s *sender) SendChits(
...
+	}
+	return s.ctx.Log.With(fields...)
+}
+
+func (s *sender) send(log logging.Logger, to common.SendConfig, msg *message.OutboundMessage) set.Set[ids.NodeID] {
+	sent := s.sender.Send(msg, to, s.ctx.SubnetID, s.subnet)
+	log.Debug("sent message",
+		zap.Reflect("to", to),
+		zap.Reflect("sent", sent),
+	)
```
StephenButtolph, line 981: Now we always log who we tried to send to and who we actually sent to. Previously we only logged messages that failed to be sent... Which made tracing events super hard.

snow/networking/sender/sender.go
```diff
@@ -1151,359 +850,258 @@ func (s *sender) SendChits(
...
+	msg, err := s.msgCreator.AppError(
 		s.ctx.ChainID,
 		requestID,
 		errorCode,
 		errorMessage,
 	)
 	if err != nil {
-		s.ctx.Log.Error("failed to build message",
-			zap.Stringer("messageOp", message.AppErrorOp),
-			zap.Stringer("nodeID", nodeID),
```
StephenButtolph, line 1428: I removed the `nodeID` from all these logs, they have nothing to do with the error log (and they are included in the to/sent logs.

snow/networking/sender/sender.go
```diff
@@ -1151,359 +850,258 @@ func (s *sender) SendChits(
...
+				requestID,
+				errorCode,
+				errorMessage,
+			),
 		)
-		s.router.HandleInternal(ctx, inMsg)
 		return nil
 	}
 
-	// Create the outbound message.
```
StephenButtolph, line 1418: Removed useless comments

snow/networking/sender/sender.go
```diff
@@ -1023,120 +746,96 @@ func (s *sender) SendPullQuery(
...
-			zap.Stringer("containerID", containerID),
-			zap.Uint64("requestedHeight", requestedHeight),
+		log.Error("failed to build message",
 			zap.Error(err),
 		)
 	}
 
 	for nodeID := range nodeIDs {
-		if !sentTo.Contains(nodeID) {
-			s.ctx.Log.Debug("failed to send message",
```
StephenButtolph, line 1123: These types of logs where we logged individual lines for each failure were removed and replaced with the single log inside of `send`

snow/networking/sender/sender.go
```diff
@@ -1151,359 +850,258 @@ func (s *sender) SendChits(
...
-				zap.Stringer("messageOp", message.AppGossipOp),
-				zap.Stringer("chainID", s.ctx.ChainID),
-			)
-		}
-	}
+	s.send(log, to, msg)
 	return nil
 }
+
+func (s *sender) logWith(verboField zap.Field, fields ...zap.Field) logging.Logger {
```
joshua-kim: The first field being the verbo-only field is confusing -- would it make sense for us to have the caller check if verbo is enabled to append the verbo field isntead?

StephenButtolph: Hm, this is kind of the whole point of this helper... Would it make sense with a different name? Or different argument orders?

StephenButtolph: I removed it in favor of just logging the length. I don't think we will ever care about the actual bytes.

snow/networking/sender/sender.go
JonathanOppenheimer: From googling online, `zap.Reflect` is very slow. I was reading 

> Zap takes a different approach. It includes a reflection-free, zero-allocation JSON encoder, and the base Logger strives to avoid serialization overhead and allocations wherever possible

From the function comment on reflect 

> It uses an encoding-appropriate, reflection-based function to lazily serialize nearly  any object into the logging context, but it's relatively slow and allocation-heavy. Outside tests, Any is always a better choice.

As this is not a test, should we do something else? perchance 

```go
if log.Enabled(logging.Verbo) {                                                                                               
      log.Verbo("sent message",                                                                                                 
          zap.Int("attempted", to.NodeIDs.Len()),                                                                               
          zap.Int("succeeded", sent.Len()),                                                                                     
      )                                                                                                                         
  }  
```

so the level can be determined. from randomly looking it looks like there's a lot of other stuff you could log

snow/networking/handler/handler.go
JonathanOppenheimer: should this `{ }`  chunk be a function or something

snow/networking/sender/sender.go
JonathanOppenheimer: this function is certainly used a lot.

JonathanOppenheimer: talked in office i learned a lot

snow/networking/handler/handler.go
StephenButtolph: I felt like this was cleanest 🤷

alarso16 (approved):
Looks good. I didn't pay too close attention to what level each log should be, but the functional logic is the same

PR #4921, opened 2026-01-29
refactor(sync): extract coreth/sync to evm/sync and adapt coreth and subnet-evm

powerslider:
## Why this should be merged

Eliminates duplicate code and removes backporting burden.

Before this PR, state sync code existed in both coreth and subnet-evm repositories with nearly identical implementations. Any bug fix or improvement required manual backporting between the two codebases, which was error-prone and time-consuming.

This PR extracts the shared state sync logic into `graft/evm/sync`, creating a single source of truth. Now:
- Bug fixes apply automatically to both `coreth` and `subnet-evm`.
- No backporting required!!! - changes to sync code benefit both VMs immediately.
- Both VMs use identical sync algorithms.
- One codebase to maintain instead of two.

The extraction includes all core sync components: block syncing, code syncing, trie/leaf syncing, network handlers, and the sync engine orchestration.

Check #4848 

## How this works

New package location and structure:

```
graft/evm/sync/          <- Shared sync implementation (NEW)
├── block/               <- Block syncer (fetches parent blocks for BLOCKHASH)
├── code/                <- Contract code syncer with durable recovery
├── client/              <- Validates peer responses, provides trie sync support
├── handlers/            <- Network request handlers (Block, Code, Leafs)
├── engine/              <- High-level orchestration (Client, Server, Registry)
├── evmstate/            <- EVM state syncing (account/storage tries)
├── leaf/                <- Generic trie leaf syncing
└── types/               <- Shared interfaces (Syncer, Extender, etc.)

graft/coreth/            <- Uses evm/sync, keeps only coreth-specific code
└── plugin/evm/atomic/sync/  <- Atomic trie syncing (C-Chain specific)

graft/subnet-evm/        <- Uses evm/sync, minimal sync code remaining
```
- Both coreth and subnet-evm now import from `graft/evm/sync` instead of maintaining their own copies. Coreth retains only its C-Chain-specific atomic trie syncing code in `plugin/evm/atomic/sync/`.
- Thin adapter types (`chainContextAdapter`) bridge each VM's `eth.Ethereum` to the sync engine's interface.
- All sync tests use `utilstest.ForEachCodec` to run with both coreth and subnet-evm codecs, ensuring compatibility across both VMs.

## How this was tested

existing UT

no

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -288,8 +288,8 @@ gioui.org v0.0.0-20210308172011-57750fc8a0a6/go.mod h1:RSH6KIUZ0p2xy5zHDxgAM4zum
 git.sr.ht/~sbinet/gg v0.3.1/go.mod h1:KGYtlADtqsqANL9ueOFkWymvzUvLMQllU5Ixo+8v3pc=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
@@ -359,8 +359,9 @@ github.com/btcsuite/btcd/btcutil v1.1.0/go.mod h1:5OapHB7A2hBBWLm48mmw4MOHNJCcUB
 github.com/btcsuite/btcd/btcutil v1.1.3 h1:xfbtw8lwpp0G6NwSHb+UE67ryTFHJAiNuipusjXSohQ=
 github.com/btcsuite/btcd/btcutil v1.1.3/go.mod h1:UR7dsSJzJUfMmFiiLlIrMq1lS9jh9EdCV7FStZSnpi0=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
-github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 h1:q0rUy8C/TYNBQS1+CGKw68tLOFYSNEs0TFnxxnS9+4U=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0 h1:59Kx4K6lzOW5w6nFlA0v5+lk/6sjybR934QNHSJZPTQ=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f/go.mod h1:TdznJufoqS23FtqVCzL0ZqgP5MqXbb4fg/WgDys70nA=
 github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
 github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd/go.mod h1:HHNXQzUsZCxOoE+CPiyCTO6x34Zs86zZUiwtpXoGdtg=
--- a/graft/coreth/go.sum
+++ b/graft/coreth/go.sum
@@ -5,8 +5,8 @@ connectrpc.com/grpcreflect v1.3.0 h1:Y4V+ACf8/vOb1XOc251Qun7jMB75gCUNw6llvB9csXc
 connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JIovCtDYs=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
 github.com/DataDog/zstd v1.5.2 h1:vUG4lAyuPCXO0TLbXvPv7EB7cNK1QV/luu55UHLrrn8=
@@ -48,8 +48,9 @@ github.com/btcsuite/btcd/btcutil v1.1.0/go.mod h1:5OapHB7A2hBBWLm48mmw4MOHNJCcUB
 github.com/btcsuite/btcd/btcutil v1.1.3 h1:xfbtw8lwpp0G6NwSHb+UE67ryTFHJAiNuipusjXSohQ=
 github.com/btcsuite/btcd/btcutil v1.1.3/go.mod h1:UR7dsSJzJUfMmFiiLlIrMq1lS9jh9EdCV7FStZSnpi0=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
-github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 h1:q0rUy8C/TYNBQS1+CGKw68tLOFYSNEs0TFnxxnS9+4U=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0 h1:59Kx4K6lzOW5w6nFlA0v5+lk/6sjybR934QNHSJZPTQ=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f/go.mod h1:TdznJufoqS23FtqVCzL0ZqgP5MqXbb4fg/WgDys70nA=
 github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
 github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd/go.mod h1:HHNXQzUsZCxOoE+CPiyCTO6x34Zs86zZUiwtpXoGdtg=
--- a/graft/subnet-evm/go.sum
+++ b/graft/subnet-evm/go.sum
@@ -5,8 +5,8 @@ connectrpc.com/grpcreflect v1.3.0 h1:Y4V+ACf8/vOb1XOc251Qun7jMB75gCUNw6llvB9csXc
 connectrpc.com/grpcreflect v1.3.0/go.mod h1:nfloOtCS8VUQOQ1+GTdFzVg2CJo4ZGaat8JIovCtDYs=
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
-github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/CloudyKit/fastprinter v0.0.0-20200109182630-33d98a066a53/go.mod h1:+3IMCy2vIlbG1XG/0ggNQv0SvxCAIpPM5b1nCz56Xno=
 github.com/CloudyKit/jet/v3 v3.0.0/go.mod h1:HKQPgSJmdK8hdoAbKUUWajkHyHo4RaU5rMdUywE7VMo=
 github.com/DataDog/zstd v1.5.2 h1:vUG4lAyuPCXO0TLbXvPv7EB7cNK1QV/luu55UHLrrn8=
@@ -50,8 +50,9 @@ github.com/btcsuite/btcd/btcutil v1.1.0/go.mod h1:5OapHB7A2hBBWLm48mmw4MOHNJCcUB
 github.com/btcsuite/btcd/btcutil v1.1.3 h1:xfbtw8lwpp0G6NwSHb+UE67ryTFHJAiNuipusjXSohQ=
 github.com/btcsuite/btcd/btcutil v1.1.3/go.mod h1:UR7dsSJzJUfMmFiiLlIrMq1lS9jh9EdCV7FStZSnpi0=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
-github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 h1:q0rUy8C/TYNBQS1+CGKw68tLOFYSNEs0TFnxxnS9+4U=
 github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0 h1:59Kx4K6lzOW5w6nFlA0v5+lk/6sjybR934QNHSJZPTQ=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0/go.mod h1:7SFka0XMvUgj3hfZtydOrQY2mwhPclbT2snogU7SQQc=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f/go.mod h1:TdznJufoqS23FtqVCzL0ZqgP5MqXbb4fg/WgDys70nA=
 github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
 github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd/go.mod h1:HHNXQzUsZCxOoE+CPiyCTO6x34Zs86zZUiwtpXoGdtg=


graft/evm/sync/block/syncer_test.go
```diff
@@ -6,35 +6,22 @@ package block
...
+	"github.com/ava-labs/avalanchego/graft/evm/utils/utilstest"
 
-	handlerstats "github.com/ava-labs/avalanchego/graft/coreth/sync/handlers/stats"
-	ethparams "github.com/ava-labs/libevm/params"
+	handlerstats "github.com/ava-labs/avalanchego/graft/evm/sync/handlers/stats"
 )
 
-func TestMain(m *testing.M) {
-	customtypes.Register()
-	params.RegisterExtras()
```
alarso16, line 34: Probably didn't need to register params, but I wonder if we do need to be testing with customtypes. I don't think so, so since we do separate marshaling tests in the customtypes package, but I wanted to throw out there that I am somewhat worried about that

graft/evm/sync/block/syncer_test.go
```diff
@@ -210,8 +172,7 @@ func newTestEnvironment(t *testing.T, numBlocks int) *testEnvironment {
 func (e *testEnvironment) prePopulateBlocks(blockHeights []int) error {
 	batch := e.chainDB.NewBatch()
 	for _, height := range blockHeights {
-		if height <= len(e.blocks) {
-			// blocks[0] is block number 1, blocks[1] is block number 2, etc.
+		if height < len(e.blocks) {
```
alarso16: Good catch! But should we make this change separate? It's hard to review this edge case boundary change at the same time as refactoring

graft/evm/sync/client/client.go
```diff
@@ -52,6 +51,24 @@ var (
 )
 var _ Client = (*client)(nil)
 
+// NetworkClient defines the interface for sending sync requests over the network.
+// This interface is implemented by the network layer in coreth and subnet-evm.
+type NetworkClient interface {
```
alarso16: q: did you move this because the interface should be declared at the place it's used?
```suggestion
type Network interface {
```

graft/evm/sync/engine/client.go
```diff
@@ -97,6 +114,10 @@ type ClientConfig struct {
 	RequestSize        uint16 // number of key/value pairs to ask peers for per request
 	Enabled            bool
 	SkipResume         bool
+
+	// LeafsRequestType specifies the wire format for leafs requests.
+	// Defaults to CorethLeafsRequestType if not set.
+	LeafsRequestType message.LeafsRequestType
```
alarso16, line 120: Should we require it to be set explicitly? I think the scary part is that it seems easy to accidentally make a breaking change in subnet-evm

graft/evm/sync/engine/registry_test.go
alarso16, line 1: Did this file change at all? Would have though that github could tell it moved

graft/evm/sync/evmstate/firewood_syncer_test.go
```diff
@@ -62,23 +62,25 @@ func TestFirewoodSync(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			r := rand.New(rand.NewSource(1))
+			utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
```
alarso16: this is an interesting case, because I'm not sure that testing with multiple codecs is helpful. Firewood uses the `p2p` package for messages, not the `message` package. So using multiple codecs only tests the `code.Syncer` functionality, which is tested separately. Do we need to run with both do you think?

graft/evm/sync/evmstate/firewood_syncer_test.go
```diff
@@ -180,11 +184,15 @@ func createDB(t *testing.T) state.Database {
 	t.Helper()
 	diskdb := rawdb.NewMemoryDatabase()
 	config := firewood.DefaultConfig(t.TempDir())
-	db := extstate.NewDatabaseWithConfig(diskdb, &triedb.Config{DBOverride: config.BackendConstructor})
+	triedbConfig := &triedb.Config{
+		DBOverride: config.BackendConstructor,
+	}
+	internalState := state.NewDatabaseWithConfig(diskdb, triedbConfig)
```
alarso16, line 193: This through me for a loop. Very strange change, but it's to avoid the import cycles. Will you write a brief comment of what you're doing/why you're doing it like this? It wasn't immediately clear to me, so I assume it won't be clear to any future reader why you use the libevm `state` package and then wrap it

graft/evm/sync/evmstate/sync_test.go
alarso16, line 1: throughout this file, the server database can be shared between codec tests I believe. If this is confusing for readers, we can leave it how you edited it. But that seems like unnecessary double-work

graft/evm/sync/handlers/code_request_test.go
```diff
@@ -87,27 +88,30 @@ func TestCodeRequestHandler(t *testing.T) {
 	}
 
 	for name, test := range tests {
-		// Reset stats before each test
-		testHandlerStats.Reset()
+		utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
+			// Reset stats before each test
+			testHandlerStats.Reset()
+			codeRequestHandler := NewCodeRequestHandler(database, c, testHandlerStats)
```
alarso16: It's weird you have to make a new code handler every time. What if you reformatted like this:
```go
utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
    codeRequestHandler := NewCodeRequestHandler(...)
    for name, test := range tests {
        testHandlerStats.Reset()
        t.Run(name, func (t *testing.T) {
         ....
```

graft/evm/sync/handlers/leafs_request.go
```diff
@@ -17,16 +17,26 @@ import (
...
 
 var _ LeafRequestHandler = (*leafsRequestHandler)(nil)
 
+// LeafRequestHandler handles incoming leaf requests from peers.
+type LeafRequestHandler interface {
+	OnLeafsRequest(ctx context.Context, nodeID ids.NodeID, requestID uint32, leafsRequest message.LeafsRequest) ([]byte, error)
+}
+
+// SnapshotProvider provides access to the snapshot tree.
+type SnapshotProvider interface {
```
alarso16, line 36: Why not put this definition closer to the BlockProvider one? Trying to trace through all of the interface definitions in this refactor is a little confusing

graft/evm/sync/client/client.go
powerslider: yes, exactly and I've renamed it `Network`

graft/evm/sync/block/syncer_test.go
powerslider: The `TestMain` was removed because the sync tests don't need custom params registration - they test message serialization using the message codec directly, not the params package. The marshaling tests for custom types are already covered in the `customtypes` package as you mentioned, so there's no need to duplicate that coverage here. I don't think there is anything to worry about - quite the opposite since we are now focusing the tests only for the state sync concern.

graft/evm/sync/engine/client.go
powerslider: The default is intentional for backwards compatibility with coreth (the primary use case). Subnet-evm explicitly sets `SubnetEVMLeafsRequestType` in its VM initialization, so there's no risk there. The comment documents the default behavior, and any new integrations would need to consciously choose their codec type. That said, if you feel strongly about making it required, I can make that change.

graft/evm/sync/engine/registry_test.go
powerslider: The file contains changes mostly from package name updates, import changes, and the `ForEachCodec` refactoring. GitHub's rename detection doesn't trigger when changes exceed ~50% similarity threshold. Git tracks it correctly with `git log --follow`. So it is a Github limitation, not a git issue.

graft/evm/sync/evmstate/sync_test.go
powerslider: Yep, the server database could theoretically be shared since the codec only affects wire format, not the underlying data. However, I'd prefer to keep it as-is for now - the test isolation is clearer and there's less risk of subtle bugs from shared state. We can optimize this in a follow-up if the test runtime becomes a concern.

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,889 @@
...
+	for name, test := range tests {
+		utilstest.ForEachCodec(t, func(_ string, c codec.Manager) {
+			t.Run(name, func(t *testing.T) {
+				t.Parallel()
+				testNetClient := &testNetwork{}
+				stateSyncClient := New(&Config{
+					Network:          testNetClient,
+					Codec:            c,
+					Stats:            clientstats.NewNoOpStats(),
+					StateSyncNodeIDs: nil,
```
alarso16:
```suggestion
```
The reasonable default value is the zero value

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,889 @@
...
+
+		return blockBytes
+	}
+	tests := map[string]struct {
+		request        message.BlockRequest
+		getResponse    func(t *testing.T, c codec.Manager, blocksRequestHandler *handlers.BlockRequestHandler, request message.BlockRequest) []byte
+		assertResponse func(t *testing.T, response []*types.Block)
+		expectedErr    error
+	}{
+		"normal resonse": {
```
alarso16:
```suggestion
		"normal response": {
```

graft/evm/sync/handlers/leafs_request.go
powerslider: `SnapshotProvider` is defined where it's consumed (in `leafs_request.go`), following the Go idiom of defining interfaces at the point of use rather than alongside implementations. `BlockProvider` is in the same file because the leafs handler uses both interfaces.

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,889 @@
...
+	smallTrieRoot, _, _ := synctest.GenerateIndependentTrie(t, r, trieDB, leafsLimit, common.HashLength)
+
+	tests := map[string]struct {
+		getRequest      func(t *testing.T, leafReqType message.LeafsRequestType) message.LeafsRequest
+		getResponse     func(t *testing.T, c codec.Manager, handler handlers.LeafRequestHandler, request message.LeafsRequest, leafReqType message.LeafsRequestType) []byte
+		requireResponse func(t *testing.T, response message.LeafsResponse)
+		expectedErr     error
+	}{
+		"full response for small (single request) trie": {
+			getRequest: func(t *testing.T, leafReqType message.LeafsRequestType) message.LeafsRequest {
```
alarso16:
```suggestion
			request: func(t *testing.T, leafReqType message.LeafsRequestType) message.LeafsRequest {
```
This was a better name, `get` doesn't seem accurate and is also uninformative. I think `make` is an appropriate action though

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,889 @@
...
+		require.Contains(t, testNetClient.nodesRequested, stateSyncNodes[1])
+		require.Contains(t, testNetClient.nodesRequested, stateSyncNodes[2])
+		require.Contains(t, testNetClient.nodesRequested, stateSyncNodes[3])
+	})
+}
+
+// newLeafsRequest creates a new LeafsRequest for testing.
+// When account is common.Hash{} (empty), it creates an account trie request.
+// When account is set to a specific account hash, it creates a storage trie request for that account.
+// Uses CorethLeafsRequestType which is compatible with both codecs.
```
alarso16: Is this true? the code really looks like it sends back a subnet-evm type if you ask for it

graft/evm/sync/engine/registry_test.go
alarso16: Tip for anybody else reviewing this: since git can find it, you can just run:
```bash
git diff -w master:graft/coreth/sync/engine/registry_test.go HEAD:graft/evm/sync/engine/registry_test.go
```
which will show all diff lines, ignoring whitespace

graft/evm/sync/block/syncer_test.go
powerslider: reverted and will push a new PR for that

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -117,15 +122,17 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
 				require.NoError(t, err)
 				require.Equal(t, uint32(1), testHandlerStats.InvalidLeafsRequestCount)
 			},
 		},
 		"bad start len dropped": {
-			prepareTestFn: func() (context.Context, message.CorethLeafsRequest) {
-				return t.Context(), newLeafsRequest(t,
+			prepareTestFn: func(_ codec.Manager, leafReqType message.LeafsRequestType) (context.Context, message.LeafsRequest) {
+				return newLeafsRequest(t,
+					t.Context(),
```
alarso16: :( you can keep it it's fine

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -279,15 +305,20 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
 				require.NoError(t, err)
 				require.Equal(t, uint32(1), testHandlerStats.InvalidLeafsRequestCount)
 			},
 		},
 		"invalid node type dropped": {
-			prepareTestFn: func() (context.Context, message.CorethLeafsRequest) {
-				return t.Context(), newLeafsRequest(t,
+			prepareTestFn: func(_ codec.Manager, leafReqType message.LeafsRequestType) (context.Context, message.LeafsRequest) {
+				ctx, cancel := context.WithCancel(t.Context())
+				defer cancel()
```
alarso16: Why is this one now canceled?

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -259,18 +280,23 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
 				require.Len(t, leafsResponse.Keys, 500)
 				require.Len(t, leafsResponse.Vals, 500)
 			},
 		},
 		"end greater than start dropped": {
-			prepareTestFn: func() (context.Context, message.CorethLeafsRequest) {
-				return t.Context(), newLeafsRequest(t,
+			prepareTestFn: func(_ codec.Manager, leafReqType message.LeafsRequestType) (context.Context, message.LeafsRequest) {
+				ctx, cancel := context.WithCancel(t.Context())
+				defer cancel()
```
alarso16: This one too

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -202,18 +217,20 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
 				require.NoError(t, err)
 				require.Equal(t, uint32(1), testHandlerStats.TrieErrorCount)
 			},
 		},
 		"cancelled context dropped": {
-			prepareTestFn: func() (context.Context, message.CorethLeafsRequest) {
+			prepareTestFn: func(_ codec.Manager, leafReqType message.LeafsRequestType) (context.Context, message.LeafsRequest) {
 				ctx, cancel := context.WithCancel(t.Context())
-				cancel()
+				defer cancel()
```
alarso16: You don't actually need the defer

graft/evm/utils/utilstest/codec.go
alarso16, line 1: Wasn't this added in the previous PR?

graft/subnet-evm/plugin/evm/sync_adapter.go
alarso16, line 1: This adaptor seems unnecessary. Can't we just ensure that `eth.Ethereum` fulfills the interface, and then pass that in?

graft/coreth/plugin/evm/sync_adapter.go
alarso16, line 1: Neither chain adaptor actually does anything

graft/evm/sync/synctest/test_blocks.go
```diff
@@ -0,0 +1,165 @@
...
+	m := make(map[common.Hash]*types.Block, len(blocks))
+	for _, block := range blocks {
+		m[block.Hash()] = block
+	}
+	return m
+}
+
+// BlockProviderFunc returns a function suitable for handlers.TestBlockProvider.GetBlockFn
+// that looks up blocks by height and verifies the hash matches.
+func BlockProviderFunc(blocks []*types.Block) func(common.Hash, uint64) *types.Block {
```
alarso16: Is this logic really repeated that much? We probably don't have to verify that the hash matches

graft/evm/sync/synctest/blocks.go
alarso16, line 1: nit: probably `blocks.go` is sufficient. It's in a testing folder

alarso16: Also, why not use the libevm functions for generating blocks?

graft/evm/sync/client/client_test.go
powerslider: Yeah, this was valid in a previous version of this function. Will remove it now, good catch.

graft/evm/sync/handlers/leafs_request_test.go
powerslider: This regressed, when I was rebasing from the previous base branch to master. Will fix this.

graft/evm/sync/synctest/blocks.go
powerslider: I followed the same way the other files were named, but I agree with your suggestion. And regarding using libevm functions I had not investigated that. I will see if I can use something from there ;)

graft/coreth/plugin/evm/sync_adapter.go
powerslider: ok, let me break down exactly what is happening here and why we are doing it this way:

- The adapter is necessary due to Go's strict interface matching rules. `eth.Ethereum.BlockChain()` returns `*core.BlockChain` (concrete type), but `engine.ChainContext` expects `BlockChain() engine.BlockChain` (interface return type). Even though `*core.BlockChain` implements `engine.BlockChain`, Go doesn't allow covariant return types - the method signatures must match exactly.

- This is the idiomatic Go solution - a thin adapter that converts concrete return types to interface types. The alternative would be modifying `eth.Ethereum` to return interfaces directly, but we want to minimize changes to `geth`-derived code.

- What we gain - the sync engine in `graft/evm` is fully decoupled from `coreth/subnet-evm` specific types. It depends only on the `engine.ChainContext` interface, making it testable and reusable without importing the `eth` package.

- The adapters are used in `coreth/plugin/evm/vm.go` and `subnet-evm/plugin/evm/vm.go` when initializing the sync engine client config.

graft/subnet-evm/plugin/evm/sync_adapter.go
powerslider: Check https://github.com/ava-labs/avalanchego/pull/4921#discussion_r2763974526

graft/evm/sync/handlers/leafs_request_test.go
powerslider: yeah, same rebase issue, will fix this

graft/evm/sync/synctest/blocks.go
powerslider: Investigated `core.GenerateChain` and `core.GenerateChainWithGenesis` from `libevm`. These are full-featured block generators that handle state roots, receipts, EIP compliance, and require a consensus engine + database setup. Our current implementation is intentionally minimal - we only need structurally valid blocks with correct parent-child hash chains for sync testing, not full state execution. Using `libevm`'s generator would add complexity without benefit for this use case.

I'd still prefer having more focused tests, because the goal here is not doing end-to-end testing. We have other tests covering that and I will add scenarios with more bells and whistles once I start doing dynamic state sync end-to-end testing.

graft/coreth/plugin/evm/sync_adapter.go
alarso16: Ah didn't realize go couldn't figure this out. Thanks

graft/evm/sync/handlers/leafs_request.go
alarso16: Did you already break this idiom by providing a `types` package? Why not put them there? I understand this might be conventionaly, but I personally find it confusing. You have to check a bunch of different packages for your interface definitions, just to initialize a state syncer?

graft/evm/sync/synctest/blocks.go
alarso16: Yeah this seems ok, if you need all the engine and database stuff, that seems excessive.

graft/evm/sync/block/syncer_test.go
```diff
@@ -222,7 +194,8 @@ func (e *testEnvironment) prePopulateBlocks(blockHeights []int) error {
 
 // createSyncer creates a block syncer with the given configuration
 func (e *testEnvironment) createSyncer(fromHeight uint64, blocksToFetch uint64) (*BlockSyncer, error) {
-	if fromHeight > uint64(len(e.blocks)) {
+	// Use >= because we access e.blocks[fromHeight] below.
+	if fromHeight >= uint64(len(e.blocks)) {
```
alarso16: I think this semantic change is coupled with the fix you reverted

graft/evm/utils/utilstest/codec.go
powerslider: rebasing residue, fixed now

graft/evm/sync/handlers/leafs_request.go
powerslider: These serve different purposes. The `types` package contains interfaces that form the public API of the sync system - what external code needs to implement to plug in (`Syncer`, `LeafClient`, `Extender`). `SnapshotProvider` and `BlockProvider` are internal ports - they define what the leafs handler needs from its dependencies, not what external code should implement. Moving them to `types` would conflate the public sync API with internal implementation details. 

So I wouldn't say the debate here is what is idiomatic for Go or not, rather what is the scope and use case of these interfaces.

graft/evm/sync/engine/doubles_test.go
JonathanOppenheimer: Now that this is being shared, I think you could use a config for this name. If you try to register multiple, it will fail with `errSyncerAlreadyRegistered`. I see you work around this in tests work with `namedSyncer`, but it seems like a layer of indirection that culd be fixed at the root.

graft/evm/sync/evmstate/state_syncer.go
```diff
@@ -40,15 +40,16 @@ var (
...
-	syncer    *leaf.CallbackSyncer      // performs the sync, looping over each task's range and invoking specified callbacks
-	codeQueue *code.Queue               // queue that manages the asynchronous download and batching of code hashes
-	trieQueue *trieQueue                // manages a persistent list of storage tries we need to sync and any segments that are created for them
+	db               ethdb.Database            // database we are syncing
+	root             common.Hash               // root of the EVM state we are syncing to
+	trieDB           *triedb.Database          // trieDB on top of db we are syncing. used to restore any existing tries.
+	snapshot         snapshot.SnapshotIterable // used to access the database we are syncing as a snapshot.
+	batchSize        uint                      // write batches when they reach this size
+	leafsRequestType message.LeafsRequestType  // type of leafs request to use (coreth or subnet-evm wire format)
+	segments         chan leaf.SyncTask        // channel of tasks to sync
```
JonathanOppenheimer, line 49: I see this can be closed in multiple places -- is the `sync.Once` pattern needed here or are there no risk of races?

graft/evm/sync/evmstate/state_syncer.go
JonathanOppenheimer, line 63: Same for this channel in multiple places

graft/coreth/plugin/evm/atomic/sync/syncer.go
JonathanOppenheimer, line 83: random nit comment but `leafs` vs `leaves` is used throughout all this syncer code.

graft/evm/sync/leaf/syncer.go
JonathanOppenheimer: `onFailure` no longer exists

powerslider:
Awesome news! Dispatch L1 synced without any issues testing that the extraction works correctly and also the fact that this was the first time we introduced the up to date sync code to `subnet-evm`. You can check the attached logs below:
[dispatch_l1_state_sync.log](https://github.com/user-attachments/files/25182139/dispatch_l1_state_sync.log)

powerslider:
C-Chain on Fuji was also successfully state synced. Check full logs below:
[cchain_state_sync_rotated.log](https://github.com/user-attachments/files/25187294/cchain_state_sync_rotated.log)
[cchain_state_sync.log](https://github.com/user-attachments/files/25187296/cchain_state_sync.log)

These manual tests proove that the implementation is correct and we did not introduce any regressions.

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,887 @@
...
+			}
+		}, invalidResponse)
+		_, err = client.GetLeafs(ctx, request)
+		require.ErrorIs(t, err, context.Canceled)
+	})
+}
+
+func TestStateSyncNodes(t *testing.T) {
+	messagetest.ForEachCodec(t, func(codecName string, c codec.Manager) {
+		leafReqType := message.CorethLeafsRequestType
```
ceyonur: Should we pass `leafReqType` here instead of `codecName`?

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,887 @@
...
+	"github.com/ava-labs/avalanchego/graft/evm/sync/handlers"
+	"github.com/ava-labs/avalanchego/graft/evm/sync/synctest"
+	"github.com/ava-labs/avalanchego/ids"
+
+	clientstats "github.com/ava-labs/avalanchego/graft/evm/sync/client/stats"
+	handlerstats "github.com/ava-labs/avalanchego/graft/evm/sync/handlers/stats"
+	ethparams "github.com/ava-labs/libevm/params"
+)
+
+const subnetEVMCodecName = "subnet-evm"
```
ceyonur: we should avoid hardcoding any vm specific thing as much as possible and I think the test util can just do that instead.

graft/evm/sync/engine/client.go
ceyonur: We should avoid defaulting to anything. I'd prefeer to have a panic/error in initialization rather than finding out Subnet-EVM or another VM is defaulting to coreth codec.

graft/evm/sync/evmstate/firewood_syncer_test.go
ceyonur: Aren't we reducing the test coverage by doing so? I think this is a question for @alarso16.

graft/evm/sync/evmstate/state_syncer.go
```diff
@@ -80,19 +81,27 @@ func WithBatchSize(n uint) SyncerOption {
 	})
 }
 
+// WithLeafsRequestType sets the leafs request type (coreth or subnet-evm wire format).
+func WithLeafsRequestType(t message.LeafsRequestType) SyncerOption {
```
ceyonur: I don't think this should be an option.

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -688,45 +714,53 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
-		t.Run(name, func(t *testing.T) {
-			ctx, request := test.prepareTestFn()
-			t.Cleanup(func() {
-				<-snapshot.WipeSnapshot(memdb, true)
-				testHandlerStats.Reset()
-				snapshotProvider.Snapshot = nil // reset the snapshot to nil
-			})
+		messagetest.ForEachCodec(t, func(codecName string, c codec.Manager) {
+			leafReqType := message.CorethLeafsRequestType
+			if codecName == "subnet-evm" {
```
ceyonur: Passing the leaft ype definely would solve this weird string check.

graft/evm/sync/synctest/blocks.go
```diff
@@ -0,0 +1,150 @@
...
+			txDataSize = cfg.TxDataSizeFunc(i)
+		}
+		blocks[i] = newBlock(t, blocks[i-1], i, txDataSize, gasLimit)
+	}
+
+	return blocks
+}
+
+// BlockMap creates a map from block hash to block for quick lookups.
+func BlockMap(blocks []*types.Block) map[common.Hash]*types.Block {
```
ceyonur: I think this is ever used in single test file, we can just define it there.

graft/evm/sync/handlers/leafs_request_test.go
```diff
@@ -739,23 +773,24 @@ func requireRangeProofIsValid(t *testing.T, request *message.CorethLeafsRequest,
...
+	more, err := trie.VerifyRangeProof(request.RootHash(), start, response.Keys, response.Vals, proof)
 	require.NoError(t, err)
 	require.Equal(t, expectMore, more)
 }
 
-// newLeafsRequest creates a new CorethLeafsRequest for testing.
+// newLeafsRequest creates a new LeafsRequest for testing.
 // When account is common.Hash{} (empty), it creates an account trie request.
 // When account is set to a specific account hash, it creates a storage trie request for that account.
 func newLeafsRequest(
```
ceyonur: this probably should be renamed with `mustNewLeafsRequest`

graft/evm/sync/client/client_test.go
```diff
@@ -0,0 +1,887 @@
...
+	blocks := synctest.GenerateTestBlocks(t, numBlocks, nil)
+	require.Len(t, blocks, numBlocks+1) // includes genesis
+
+	// encodeBlockSlice takes a slice of blocks that are ordered in increasing height order
+	// and returns a slice of byte slices with those blocks encoded in reverse order
+	encodeBlockSlice := func(blocks []*types.Block) [][]byte {
+		blockBytes := make([][]byte, 0, len(blocks))
+		for i := len(blocks) - 1; i >= 0; i-- {
+			buf := new(bytes.Buffer)
+			require.NoError(t, blocks[i].EncodeRLP(buf))
```
ceyonur, line 146: Should we have any concern that this EncodeRLP does not actually encode any coreth or subnet-evm types anymore?

graft/subnet-evm/sync/README.md
ceyonur, line 1: Where do we keep the new readme? Coreth README has outdated references, and I think we should move the readme to the new shared package?

graft/evm/sync/engine/registry.go
ceyonur: is this registeredIDs or names?

graft/evm/sync/client/client.go
ceyonur: we should probably rename this

graft/evm/sync/engine/client.go
ceyonur, line 266: is this (and `Error()` thread-safe? (not directly related with this PR)

graft/subnet-evm/sync/README.md
powerslider: Moved and updated the README.md. You can check it here: https://github.com/ava-labs/avalanchego/pull/4921/changes/59014827f547faa27a73dbe1c1e35f17f73acb0c

graft/evm/sync/engine/client.go
powerslider: Not strictly thread-safe on its own, but in practice the write in the goroutine happens before the read: `Error()` is only called after `StateSyncDone` is closed and `Shutdown()` calls `c.wg.Wait()`. Could make it explicit with `atomic.Value` if we want to be defensive, but that's separate from this PR.

graft/evm/sync/client/client_test.go
powerslider: I wouldn't say so. `EncodeRLP` here is the standard `geth` block encoding (header, txs, uncles). It was never encoding `coreth` or `subnet-evm` custom types. Those are handled via `customtypes` registration at the VM layer.

graft/evm/sync/evmstate/state_syncer.go
powerslider: No risk of races because each channel is closed exactly once on a single deterministic code path: `mainTrieDone` after the main trie sync completes, `storageTriesDone` when the last storage trie page arrives, and `segments` on one of two mutually exclusive paths (no storage tries vs. last storage trie finishes). No concurrent close is possible, so `sync.Once` isn't needed here.

graft/evm/sync/evmstate/state_syncer.go
powerslider: Same https://github.com/ava-labs/avalanchego/pull/4921#discussion_r2793499384

graft/coreth/plugin/evm/atomic/sync/syncer.go
powerslider: `leafs` is kinda the convention inherited from `geth` throughout the trie and sync code. I wouldn't bother changing it since it would be a large cross-cutting rename.

graft/evm/sync/client/client_test.go
ceyonur: We were registering `customtypes` previously for this test https://github.com/ava-labs/avalanchego/blob/master/graft/coreth/sync/client/client_test.go#L34-L39

graft/evm/message/messagetest/codec.go
```diff
@@ -11,15 +11,21 @@ import (
...
-		"subnet-evm": message.SubnetEVMCodec,
+	type codecEntry struct {
+		codec       codec.Manager
+		leafReqType message.LeafsRequestType
 	}
-	for name, c := range codecs {
+	codecs := map[string]codecEntry{
+		"coreth":     {message.CorethCodec, message.CorethLeafsRequestType},
+		"subnet-evm": {message.SubnetEVMCodec, message.SubnetEVMLeafsRequestType},
+	}
```
ceyonur, line 25: I don't think we need a map here

ceyonur: In general the less we write `coreth` or `subnet-evm` in the shared package (even as comments/strings) it's better.

powerslider: What do you propose as an alternative to implement this?

The map could be replaced with a slice, but we'd still need human-readable names for `t.Run`. The alternatives are:
- add a `String()` method to `LeafsRequestType` -  but that adds production code just for test naming
- use generic names like `"codec-0"` - but that hurts test output readability

The strings here are only test subtest labels, not logic coupled to a specific VM. Happy to change if you have a preferred approach in mind.

graft/evm/sync/evmstate/state_syncer.go
JonathanOppenheimer: Got it! thanks for the explanation!

graft/coreth/plugin/evm/atomic/sync/syncer.go
JonathanOppenheimer: got it!

graft/evm/sync/client/client_test.go
powerslider: The old test used `core.GenerateChain` with `dummy.NewETHFaker()`, which required `customtypes.Register()` and pulled in coreth-specific dependencies (`consensus`, `core`, `params`). We replaced it with a lightweight `synctest.GenerateTestBlocks` because:
 - Block sync tests only need structurally valid blocks with correct parent hash linkage - they don't exercise consensus rules or custom RLP encoding.
 - `customtypes` registration was only needed as a transitive requirement of `core.GenerateChain`.
 - The shared `evm/sync` package shouldn't depend on VM-specific packages.

graft/evm/message/messagetest/codec.go
ceyonur: ah ok I did not realize we use this in test runs. No worries.

ceyonur (approved):

powerslider:
I am specifying some details for @maru-ava or anyone available from @ava-labs/platform-avalanchego  that is interested to review/approve the dependency changes in `graft/evm` needed by this PR:

New direct dependencies:
- `google/go-cmp` - test assertions
- `go.uber.org/goleak` - goroutine leak detection in tests
- `golang.org/x/sync` - promoted from indirect to direct - used for concurrency primitives (e.g. `errgroup`) in the extracted sync code

New indirect dependency:
- `google/btree` - pulled in transitively via `avalanchego/database/merkle/sync`, imported by `sync/evmstate/firewood_syncer.go`

PR #4922, opened 2026-01-30
Extract P-chain uptime tracking into a separate db and state handler

rrazvan1:
## Why this should be merged
This PR separates uptime tracking from validators' storage, which is required for ACP-236.

More specifically: currently, the only entrypoint of mutating validator's information (uptime and accrued delegatee rewards) is the following: `state.SetUptime(..)` and `state.SetDelegateeRewards(..)`, which keeps track of these values through `validatorsMetadata` structure.

With ACP-236, mutable fields are added to staker struct itself, and mutated through a new method `state.UpdateCurrentValidator(..)`. For a clean and consistent approach, there should be only entrypoint of mutating the validator which will be `state.UpdateCurrentValidator(..)`, so we should get rid of all the other ways of mutating the validator through following steps:
1. extracting uptime to a different db (this PR)
2. moving delegatee rewards to the staker struct (subseqnet PR).

## How this works
Adds a dedicated state manager for primary network's validators uptime.

Migration process: at startup, try to read uptime from the new DB, if not found, fallback to old DB. On next commit, uptime from all validators is written to new db.

## How this was tested
UTs and manual testing

No.

### DIFF
--- a/vms/platformvm/state/metadata_validator.go
+++ b/vms/platformvm/state/metadata_validator.go
@@ -29,14 +29,15 @@ type preDelegateeRewardMetadata struct {
 }
 
 type validatorMetadata struct {
-	UpDuration               time.Duration `v0:"true"`
-	LastUpdated              uint64        `v0:"true"` // Unix time in seconds
-	PotentialReward          uint64        `v0:"true"`
-	PotentialDelegateeReward uint64        `v0:"true"`
-	StakerStartTime          uint64        `          v1:"true"`
+	UpDuration  time.Duration `v0:"true"` // Deprecated: It is now tracked in the uptime state.
+	LastUpdated uint64        `v0:"true"` // Unix time in seconds. Deprecated: It is now tracked in the uptime state.
+
+	PotentialReward          uint64 `v0:"true"`
+	PotentialDelegateeReward uint64 `v0:"true"`
+	StakerStartTime          uint64 `          v1:"true"`
 
 	txID        ids.ID
-	lastUpdated time.Time
+	lastUpdated time.Time // Deprecated: It is now tracked in the uptime state.
 }
 
 // Permissioned validators originally wrote their values as nil.
--- a/vms/platformvm/state/state.go
+++ b/vms/platformvm/state/state.go
@@ -1926,12 +1926,14 @@ func (s *state) loadCurrentValidators() error {
 		}
 
 		metadataBytes := validatorIt.Value()
-		metadata := &validatorMetadata{}
+		metadata := &validatorMetadata{
+			txID: txID,
+		}
 		if scheduledStakerTx, ok := tx.Unsigned.(txs.ScheduledStaker); ok {
 			// Populate [StakerStartTime] using the tx as a default in the event
 			// it was added pre-durango and is not stored in the database.
 			//
-			// Note: We do not populate [LastUpdated] since it is expected to
+			// Note: We do not populate [LastUpdatedUnix] since it is expected to
 			// always be present on disk.
 			metadata.StakerStartTime = uint64(scheduledStakerTx.StartTime().Unix())
 		}
@@ -1984,7 +1986,7 @@ func (s *state) loadCurrentValidators() error {
 			txID: txID,
 		}
 		if scheduledStakerTx, ok := tx.Unsigned.(txs.ScheduledStaker); ok {
-			// Populate [StakerStartTime] and [LastUpdated] using the tx as a
+			// Populate [StakerStartTime] and [LastUpdatedUnix] using the tx as a
 			// default in the event they are not stored in the database.
 			startTime := uint64(scheduledStakerTx.StartTime().Unix())
 			metadata.StakerStartTime = startTime
--- a/vms/platformvm/state/uptime_tracker.go
+++ b/vms/platformvm/state/uptime_tracker.go
@@ -11,8 +11,8 @@ import (
 )
 
 type uptimeData struct {
-	UpDuration  time.Duration `v0:"true"`
-	LastUpdated uint64        `v0:"true"` // Unix time in seconds
+	UpDuration      time.Duration `v0:"true"`
+	LastUpdatedUnix uint64        `v0:"true"` // Unix time in seconds
 
 	lastUpdated time.Time // in-memory convenience form
 }
@@ -79,7 +79,7 @@ func (u *UptimeTrackerState) LoadUptime(nodeID ids.NodeID) error {
 	if _, err := UptimeCodec.Unmarshal(bytes, uptime); err != nil {
 		return err
 	}
-	uptime.lastUpdated = time.Unix(int64(uptime.LastUpdated), 0)
+	uptime.lastUpdated = time.Unix(int64(uptime.LastUpdatedUnix), 0)
 
 	u.validatorsUptime[nodeID] = uptime
 	return nil
@@ -96,7 +96,7 @@ func (u *UptimeTrackerState) WriteUptime(codecVersion uint16) error {
 			continue
 		}
 
-		uptime.LastUpdated = uint64(uptime.lastUpdated.Unix())
+		uptime.LastUpdatedUnix = uint64(uptime.lastUpdated.Unix())
 
 		bytes, err := UptimeCodec.Marshal(codecVersion, uptime)
 		if err != nil {
--- a/vms/platformvm/state/state_test.go
+++ b/vms/platformvm/state/state_test.go
@@ -2404,3 +2404,102 @@ func TestGetCurrentValidators(t *testing.T) {
 		})
 	}
 }
+
+func TestDelegateeRewardCommitAndLoad(t *testing.T) {
+	require := require.New(t)
+
+	db := memdb.New()
+	s := newTestState(t, db)
+
+	const expectedDelegateeReward = uint64(12345)
+	require.NoError(s.SetDelegateeReward(
+		constants.PrimaryNetworkID,
+		defaultValidatorNodeID,
+		expectedDelegateeReward,
+	))
+	require.NoError(s.Commit())
+
+	actualDelegateeReward, err := s.GetDelegateeReward(constants.PrimaryNetworkID, defaultValidatorNodeID)
+	require.NoError(err)
+	require.Equal(expectedDelegateeReward, actualDelegateeReward)
+
+	// Reload state.
+	s = newTestState(t, db)
+	actualDelegateeReward, err = s.GetDelegateeReward(constants.PrimaryNetworkID, defaultValidatorNodeID)
+	require.NoError(err)
+	require.Equal(expectedDelegateeReward, actualDelegateeReward)
+}
+
+func TestUptimeMigrationFromLegacyMetadata(t *testing.T) {
+	require := require.New(t)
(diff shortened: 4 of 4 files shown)


vms/platformvm/state/state.go
```diff
@@ -1912,9 +1926,7 @@ func (s *state) loadCurrentValidators() error {
 		}
 
 		metadataBytes := validatorIt.Value()
-		metadata := &validatorMetadata{
-			txID: txID,
-		}
+		metadata := &validatorMetadata{}
```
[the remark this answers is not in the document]
rrazvan1: We dont need `txID` anymore because the only usage for it was to have it in order to be able to put in db the new uptime for a specific node.

For the other entrypoints we have the staker changed, so we have access to `TxID`.

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+type uptimeData struct {
+	UpDuration  time.Duration `v0:"true"`
+	LastUpdated uint64        `v0:"true"` // Unix time in seconds
+
+	lastUpdated time.Time // in-memory convenience form
+}
```
StephenButtolph, line 18: If we do go forward with the migration, I would prefer us to use canoto rather than continuing the usage of the prior codec. The codec just isn't flexible enough and introduces techdebt.

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (u *UptimeTrackerState) GetUptime(nodeID ids.NodeID) (upDuration time.Duration, lastUpdated time.Time, err error) {
+	if uptime, ok := u.validatorsUptime[nodeID]; ok && uptime != nil {
+		return uptime.UpDuration, uptime.lastUpdated, nil
+	}
+
+	return 0, time.Time{}, database.ErrNotFound
+}
+
+func (u *UptimeTrackerState) Exists(nodeID ids.NodeID) bool {
```
yacovm, line 44: I think this isn't tested in UTs

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	u.validatorsUptime[nodeID] = &uptimeData{
+		UpDuration:  upDuration,
+		lastUpdated: lastUpdated,
+	}
+	u.modified[nodeID] = struct{}{}
+}
+
+func (u *UptimeTrackerState) DeleteUptime(nodeID ids.NodeID) {
```
yacovm, line 67: why not just delete from the maps?

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+	bytes, err := u.db.Get(nodeID[:])
+	if err != nil {
+		return err
+	}
+
+	uptime := &uptimeData{}
+	if _, err := UptimeCodec.Unmarshal(bytes, uptime); err != nil {
+		return err
+	}
+	uptime.lastUpdated = time.Unix(int64(uptime.LastUpdated), 0)
```
yacovm: why do we update the last updated here if we don't change this validator's uptime?

vms/platformvm/state/state.go
```diff
@@ -2247,6 +2273,7 @@ func (s *state) write(updateValidators bool, height uint64) error {
 		s.writeExpiry(),
 		s.updateValidatorManager(updateValidators),
 		s.writeValidatorDiffs(height),
+		s.uptimeTrackerState.WriteUptime(UptimeCodecVersion0),
```
yacovm, line 2269: if we always write using this codec version, shouldn't the codec version be part of the initialization?

vms/platformvm/state/state.go
```diff
@@ -1942,6 +1954,20 @@ func (s *state) loadCurrentValidators() error {
 		s.currentStakers.stakers.ReplaceOrInsert(staker)
 
 		s.validatorState.LoadValidatorMetadata(staker.NodeID, staker.SubnetID, metadata)
+
+		// Load uptime from new DB, fallback to old metadata for migration
```
yacovm: can we extract this block of code to a function?

vms/platformvm/state/uptime_tracker.go
yacovm: oh I see, a nil means it's deleted. 

I think it's better to be explicit and add a boolean "isDeleted".

Or at least:

```
type UptimeTrackerState struct {
	validatorsUptime map[ids.NodeID]*uptimeData // can be nil, if deleted
	modified         map[ids.NodeID]struct{}
	db               database.Database
}
```

vms/platformvm/state/uptime_tracker.go
rrazvan1: `LoadUptime` is used to loads all the validator's latest uptime data (`uptime` and `lastUpdated`) in the memory.

Usage eg.: if I start the node, there are 0 uptime updates, and that validator is exiting and should be rewarded, the node needs to query the uptime, so it should be in the memory.

```
lastUpdated time.Time // in-memory convenience form
```

vms/platformvm/state/state.go
rrazvan1: currently yes but in the future we may change the stored data structure, so based on the upgrades it may use different codec versions.

vms/platformvm/state/uptime_tracker.go
rrazvan1: I agree, I added the comment, bcs I prefer keeping the structure as simple as possible.

vms/platformvm/state/state.go
```diff
@@ -1912,9 +1926,7 @@ func (s *state) loadCurrentValidators() error {
 		}
 
 		metadataBytes := validatorIt.Value()
-		metadata := &validatorMetadata{
-			txID: txID,
```
DracoLi, line 1916: I think we still need this? 
When delegatee reward is set, it will save the metadata, but without txID, it will save to an empty txId.
See this test I wrote to verify this issue. 

```golang
func TestStatePrimaryDelegateeRewardCommitAndLoad(t *testing.T) {
	require := require.New(t)

	db := memdb.New()
	s := newTestState(t, db)

	primaryValidator, err := s.GetCurrentValidator(constants.PrimaryNetworkID, defaultValidatorNodeID)
	require.NoError(err)

	const expectedDelegateeReward uint64 = 12345
	require.NoError(s.SetDelegateeReward(
		constants.PrimaryNetworkID,
		defaultValidatorNodeID,
		expectedDelegateeReward,
	))
	require.NoError(s.Commit())

	// Get delegate reward is correct since it value is updated in memory
	actualDelegateeReward, err := s.GetDelegateeReward(constants.PrimaryNetworkID, defaultValidatorNodeID)
	require.NoError(err)
	require.Equal(expectedDelegateeReward, actualDelegateeReward)

	// The original txID still exists but is stale, and the updated value was written under ids.Empty.
	metadataBytes, err := s.currentValidatorList.Get(primaryValidator.TxID[:])
	require.NoError(err)
	metadata := &validatorMetadata{}
	require.NoError(parseValidatorMetadata(metadataBytes, metadata))
	require.Equal(uint64(0), metadata.PotentialDelegateeReward)
	emptyKeyMetadataBytes, err := s.currentValidatorList.Get(ids.Empty[:])
	require.NoError(err)
	emptyKeyMetadata := &validatorMetadata{}
	require.NoError(parseValidatorMetadata(emptyKeyMetadataBytes, emptyKeyMetadata))
	require.Equal(expectedDelegateeReward, emptyKeyMetadata.PotentialDelegateeReward)

	// Reload state again.
	// This currently fails because loadCurrentValidators reads an entry written
	// under ids.Empty (missing txID), then fails to find that tx in txDB.
	s = newTestState(t, db)
	actualDelegateeReward, err = s.GetDelegateeReward(constants.PrimaryNetworkID, defaultValidatorNodeID)
	require.NoError(err)
	require.Equal(expectedDelegateeReward, actualDelegateeReward)
}
```

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+
+// UptimeTrackerState tracks primary network validators' uptime.
+// Important: it's not thread-safe.
+type UptimeTrackerState struct {
+	validatorsUptime map[ids.NodeID]*uptimeData // can be nil, if deleted
+	modified         map[ids.NodeID]struct{}
+	db               database.Database
+}
+
+func NewUptimeTrackerState(db database.Database) *UptimeTrackerState {
```
DracoLi, line 28: does NewUptimeTrackerState and related methods need to be exported? seems like we are only using in this package.

vms/platformvm/state/uptime_tracker.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"time"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+type uptimeData struct {
+	UpDuration  time.Duration `v0:"true"`
+	LastUpdated uint64        `v0:"true"` // Unix time in seconds
```
DracoLi: nit: `LastUpdated` and `lastUpdated` is a bit hard to identity which is which when skimming, maybe rename this to `LastUpdatedUnix` or `LastUpdatedSeconds` so its clearer?

vms/platformvm/state/metadata_validator.go
DracoLi: should we mark this as deprecated? 
Since we are not removing this, we want to avoid other devs from using this accidentaly.

vms/platformvm/state/state.go
```diff
@@ -3396,11 +3415,35 @@ func (s *state) ReindexBlocks(lock sync.Locker, log logging.Logger) error {
...
+	return nil
+}
+
+// loadUptime loads the uptime data for a validator from the uptime tracker database.
+// If the uptime data is not found in the new database, it falls back to the legacy
+// validator metadata and migrates the data by marking it for write on the next commit.
+func (s *state) loadUptime(nodeID ids.NodeID, metadata *validatorMetadata) error {
+	switch err := s.uptimeTrackerState.LoadUptime(nodeID); {
+	case errors.Is(err, database.ErrNotFound):
+		// Fallback to old metadata, this will mark it for migration.
```
DracoLi, line 3438: worth adding a test case for this fallback?

vms/platformvm/state/state.go
rrazvan1: nice catch! I will fix it and also include a test for it.

vms/platformvm/state/uptime_tracker.go
rrazvan1: https://github.com/ava-labs/avalanchego/pull/4879#discussion_r2798465550
I can probably create another package inside `state/uptime`. 🤔

vms/platformvm/state/state.go
rrazvan1: I added a test for this.

rrazvan1:
Not required anymore, because we will change validator state.

PR #4926, opened 2026-02-02
Remove Start/Wait mechanic in x/sync

alarso16:
## Why this should be merged

Having two functions for a single functionality is redundant. Users are always expected to call `Start` and then `Wait`, so why let them do anything different?

Also, at some point the comments referencing the struct started using the wrong variable. Seemed simple enough to include here, but I can take it out.

## How this works

Self-explanatory refactor, no functionality changed.

## How this was tested

Existing tests

No


x/merkledb/sync_test.go
```diff
@@ -377,9 +379,7 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
 	require.NoError(err)
 	require.NotNil(newSyncer)
 
-	require.NoError(newSyncer.Start(t.Context()))
-	require.NoError(newSyncer.Error())
```
alarso16, line 380: Checking `(*Syncer[_,_]).Error()` is redundant, since it will also be returned by `Sync`

database/merkle/sync/syncer.go
```diff
@@ -623,44 +655,13 @@ func (s *Syncer[R, C]) handleChangeProofResponse(
 	return nil
 }
 
-func (s *Syncer[_, _]) Error() error {
+func (s *Syncer[_, _]) error() error {
```
alarso16, line 648: I don't know why this was exported, the error is returned elsewhere

database/merkle/sync/syncer.go
```diff
@@ -731,13 +732,13 @@ func (s *Syncer[_, _]) setError(err error) {
...
 //
-// Assumes [m.workLock] is not held.
+// Assumes [s.workLock] is not held.
 func (s *Syncer[_, _]) completeWorkItem(
 	work *workItem,
 	largestHandledKey maybe.Maybe[[]byte],
 	rootID ids.ID,
 ) {
-	// nextStartKey being Nothing indicates that the entire range has been completed
+	// largestHandledKey being Nothing indicates that the entire range has been completed
```
alarso16, line 734: This variable doesn't exist, just fixing the comment

x/sync/syncer.go
```diff
@@ -181,20 +181,53 @@ func NewSyncer[R any, C any](
...
+		// This should never happen.
+		return fmt.Errorf("%w: expected %s, got %s", ErrFinishedWithUnexpectedRoot, targetRootID, root)
+	}
+
+	s.config.Log.Info("completed", zap.Stringer("root", root))
+	return nil
+}
+
+// setup initiates the work queue and enables cancellation.
+func (s *Syncer[_, _]) setup(ctx context.Context) error {
```
alarso16: I kept this in a different function to avoid changing the lock semantics. I think it's much clearer being able to defer unlocking, but let me know if you have a better idea to avoid returning a child context

database/merkle/sync/syncer.go
```diff
@@ -181,20 +181,53 @@ func NewSyncer[R any, C any](
...
+// - sync fatally errored.
+// - [ctx] is canceled.
+// If [ctx] is canceled, returns [ctx].Err().
+func (s *Syncer[_, _]) Sync(ctx context.Context) error {
+	if err := s.setup(ctx); err != nil {
+		return err
+	}
+
+	// Blocks until syncing is done or canceled.
+	s.workLoop(ctx)
```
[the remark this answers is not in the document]
alarso16: Ha it's funny that the locking confused Copilot too

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -53,12 +53,8 @@ func NewFirewoodSyncer(config syncer.Config, db *ffi.Database, target common.Has
...
 
 func (f *FirewoodSyncer) Sync(ctx context.Context) error {
-	if err := f.s.Start(ctx); err != nil {
-		return fmt.Errorf("starting syncer: %w", err)
-	}
-
-	if err := f.s.Wait(ctx); err != nil {
-		return fmt.Errorf("waiting for syncer: %w", err)
+	if err := f.s.Sync(ctx); err != nil {
+		return err
```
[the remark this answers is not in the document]
alarso16: I don't know if wrapping with something as simple as `syncing` is helpful

database/merkle/sync/syncer.go
```diff
@@ -181,25 +181,59 @@ func NewSyncer[R any, C any](
...
+	return s, nil
 }
 
-func (s *Syncer[_, _]) Start(ctx context.Context) error {
+// Sync initiates the trie syncing process and blocks until one of the following occurs:
+// - sync is complete.
+// - sync fatally errored.
+// - [ctx] is canceled.
+// If [ctx] is canceled, returns [ctx].Err().
+func (s *Syncer[_, _]) Sync(ctx context.Context) error {
```
joshua-kim, line 203: Since we're refactoring this into `Sync` -- I'm wondering if it's reasonable for us to pull the trigger on refactoring out `UpdateSyncTarget` too. We could change the signature of this to be `Sync(ctx context.Context, root ids.ID) error` to enqueue the latest requested root to be synced.

x/sync/syncer.go
```diff
@@ -181,25 +181,59 @@ func NewSyncer[R any, C any](
...
-	return m, nil
+	return s, nil
 }
 
-func (s *Syncer[_, _]) Start(ctx context.Context) error {
+// Sync initiates the trie syncing process and blocks until one of the following occurs:
+// - sync is complete.
+// - sync fatally errored.
+// - [ctx] is canceled.
+// If [ctx] is canceled, returns [ctx].Err().
```
joshua-kim: I think we're misusing some godoc linking tooling -- you can test this using `pkgsite` (install `golang.org/x/pkgsite/cmd/pkgsite@latest`). 

Generally:
* godoc is only generated for exported code, so you don't need to do special formatting like `[]` if the code isn't exported
* When linking in godoc, you can link to an exported identifier using `[SomethingInMyPackage]` or `[otherpackage.Thing]`. In this example -- we should be using `[context.Context.Err]` instead of `[ctx].Err()`.
* Similarly -- if you want to refer to a non-exported type like a var name -- godoc will render plain markdown, so `ctx` to refer to the function parameter would be appropriate.

In `pkgsite` you can see a few nitty things about godoc here, but one example is that the bulleted list isn't rendered correctly because there isn't a leading indent/tab.

x/merkledb/sync_test.go
```diff
@@ -278,7 +277,10 @@ func Test_Sync_Result_Correct_Root(t *testing.T) {
 			require.NotNil(syncer)
 
 			// Start syncing from the server
-			require.NoError(syncer.Start(ctx))
+			errChan := make(chan error, 1)
+			go func() {
+				errChan <- syncer.Sync(ctx)
+			}()
```
joshua-kim: nit: You can avoid the channel definition if you use an `errgroup.Group`, which is designed to block until a set of goroutines finish

database/merkle/sync/syncer.go
```diff
@@ -181,25 +181,59 @@ func NewSyncer[R any, C any](
...
+
+	root, err := s.db.GetMerkleRoot(ctx)
+	if err != nil {
+		return err
+	}
+
+	if targetRootID := s.getTargetRoot(); targetRootID != root {
+		// This should never happen.
+		return fmt.Errorf("%w: expected %s, got %s", ErrFinishedWithUnexpectedRoot, targetRootID, root)
+	}
```
joshua-kim, line 225: `Sync` guarantees this -- so I don't think we need this check? We can move it to tests to verify the correctness of `Sync`.

database/merkle/sync/syncer.go
```diff
@@ -210,28 +244,28 @@ func (s *Syncer[_, _]) Start(ctx context.Context) error {
 
 	s.syncing = true
 	ctx, s.cancelCtx = context.WithCancel(ctx)
-
-	go s.sync(ctx)
-	return nil
+	return ctx, nil
```
joshua-kim, line 248: I think we can now remove `Close` as well -- since we can now close by canceling the context or waiting for a return value from `Sync`?

database/merkle/sync/syncer.go
```diff
@@ -210,28 +244,28 @@ func (s *Syncer[_, _]) Start(ctx context.Context) error {
 
 	s.syncing = true
 	ctx, s.cancelCtx = context.WithCancel(ctx)
```
joshua-kim, line 247: if we remove `Close` - we can also remove the `nil` checks on `cancelCtx`, since it's guaranteed to be non-nil

x/sync/syncer.go
alarso16: This is neat! appreciate the tool

database/merkle/sync/syncer.go
alarso16: We definitely could, but I'm not sure what the design should be yet. Let's make incremental steps toward a better x/sync

database/merkle/sync/syncer.go
alarso16: You're right that it should never happen, but that's only if:
- You implemented your database correctly (range proofs, next key, etc.)
- x/sync actually works
- You used the API correctly (right now, using UpdateSyncTarget is racy)

I ran across this error several times trying to make Firewood work, so we definitely need it

x/sync/syncer.go
alarso16: Should I fix all the comments in this file? There's a lot of very strange rendering I see

x/sync/syncer.go
```diff
@@ -274,7 +275,7 @@ func (s *Syncer[_, _]) workLoop(ctx context.Context) {
 			if s.processingWorkItems == 0 {
 				// There's no work to do, and there are no work items being processed
 				// which could cause work to be added, so we're done.
-				return // [s.workLock] released by defer.
+				break // [s.workLock] released by defer.
```
joshua-kim: Why is this a break now?

database/merkle/sync/syncer.go
joshua-kim: I did a bit of investigation to see how we can fix `UpdateSyncTarget`'s race-y behavior and I think it's going to be significant scope creep -- I'm okay with not blocking on this

graft/coreth/sync/evmstate/firewood_syncer.go
```diff
@@ -53,6 +54,7 @@ func NewFirewoodSyncer(config syncer.Config, db *ffi.Database, target common.Has
 }
 
 func (f *FirewoodSyncer) Sync(ctx context.Context) error {
+	ctx, f.cancel = context.WithCancel(ctx)
```
joshua-kim, line 58: Some thoughts on the usage of finalize/cancel:
1. Why is `Finalize` exported? 
2. `Finalize` isn't called when we cancel the context -- which is now our closing path. If it's possible to un-export `Finalize` we should be able to simplify a lot of the stuff w/ the `*Once` code.
3. Why are we creating a child context?

x/sync/syncer.go
joshua-kim: We don't have to tackle all comments in this file -- but I think since we're writing a new godoc it would be reasonable to fix the new godoc we added on `Sync`.

database/merkle/sync/syncer.go
joshua-kim: ref: https://github.com/ava-labs/avalanchego/pull/4926#discussion_r2759881462

x/sync/syncer.go
alarso16: Ah oops, didn't revert this

graft/coreth/sync/evmstate/firewood_syncer.go
alarso16: This is to fulfill a pattern in coreth (`types.Finalizer`), since coreth doesn't cancel the context on error, but rather calls this. When we refactor `UpdateSyncTarget`, we might be able to get rid of this.

x/merkledb/sync_test.go
```diff
@@ -353,7 +355,7 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
 	require.NotNil(syncer)
 
 	// Start syncing from the server, will be cancelled by the range proof handler
-	require.NoError(syncer.Start(ctx))
+	require.NoError(syncer.Sync(ctx))
```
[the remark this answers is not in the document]
powerslider: Yes, this is completly valid. Check https://github.com/ava-labs/avalanchego/pull/4926/changes#r2781745360

x/merkledb/sync_test.go
```diff
@@ -352,14 +356,15 @@ func Test_Sync_Result_Correct_Root_With_Sync_Restart(t *testing.T) {
...
 
 	// Wait until we've processed some work before closing
 	require.Eventually(func() bool {
 		return db.NewIterator().Next()
 	}, 5*time.Second, 5*time.Millisecond)
-	syncer.Close()
+	cancel()
```
powerslider: This needs to be updated to account for `Sync` being a blocking call. Since `Sync` now blocks until completion (unlike the old `Start` which returned immediately), the sync finishes entirely on the `require.NoError` line before `cancel()` is ever reached. The test passes, but it no longer exercises the restart-after-partial-sync scenario - it just syncs to completion twice.

```suggestion
    // Start syncing from the server in a separate goroutine so we can
	// cancel it mid-flight to test restart behavior.
	var eg errgroup.Group
	eg.Go(func() error {
		return syncer.Sync(ctx)
	})

	// Wait until we've processed some work before cancelling.
	require.Eventually(func() bool {
		return db.NewIterator().Next()
	}, 5*time.Second, 5*time.Millisecond)
	cancel()

	// Sync should return a context-cancellation error.
	require.Error(eg.Wait())
```

graft/coreth/sync/evmstate/firewood_syncer.go
powerslider, line 52: You have to init the cancel func here, because:
- `cancel` should be initialized to a no-op to avoid a nil panic if `Finalize()` is called without `Sync()`. 
- This is reachable via `SyncerRegistry.RunSyncerTasks`, which unconditionally defers `FinalizeAll` - even if it returns early before any syncer starts. 
- The old code was safe because `f.s.Close()` had a nil guard on `cancelCtx`. This PR replaces that with a direct `f.cancel()` call.

```suggestion
	f := &FirewoodSyncer{
		s: s,
		// Initialize to no-op so Finalize is safe to call even if Sync was never started.
		cancel:    func() {},
		codeQueue: codeQueue,
	}
```

database/merkle/sync/syncer.go
```diff
@@ -240,11 +275,11 @@ func (s *Syncer[_, _]) sync(ctx context.Context) {
...
 				// There's no work to do, and there are no work items being processed
 				// which could cause work to be added, so we're done.
-				return // [m.workLock] released by defer.
+				return // [s.workLock] released by defer.
 			}
 			// There's no work to do.
-			// Note that if [m].Close() is called, or [ctx] is canceled,
-			// Close() will be called, which will broadcast on [m.unprocessedWorkCond],
+			// Note that if [s].Close() is called, or [ctx] is canceled,
+			// Close() will be called, which will broadcast on [s.unprocessedWorkCond],
```
powerslider: nit: this comment needs an update:
```suggestion
			// Note that if [ctx] is canceled, [s.close] will be called,
			// which will signal [s.unprocessedWorkCond], unblocking this goroutine.
```

PR #4931, opened 2026-02-03
Add support for continuous stakers (no transactions)

rrazvan1:
## Why this should be merged
This PR adds support for **continuous stakers**, the foundation for ACP-236.
There is no entrypoint/possible usage for the new added concepts.

## How this works
This PR introduces the state layer and data structures for continuous stakers.

## How this was tested
Unit tests

No

### DIFF
--- a/vms/platformvm/state/diff.go
+++ b/vms/platformvm/state/diff.go
@@ -635,8 +635,15 @@ func (d *diff) Apply(baseState Chain) error {
 		}
 	}
 	// Apply uptime resets for continuous validators that restarted their cycle
-	if s, ok := baseState.(uptime.State); ok {
-		for nodeID, startTime := range d.uptimeResets {
+	for nodeID, startTime := range d.uptimeResets {
+		// If baseState is another diff, propagate the resets to it
+		if baseDiff, ok := baseState.(*diff); ok {
+			if baseDiff.uptimeResets == nil {
+				baseDiff.uptimeResets = make(map[ids.NodeID]time.Time)
+			}
+			baseDiff.uptimeResets[nodeID] = startTime
+		} else if s, ok := baseState.(uptime.State); ok {
+			// If it's the actual state, apply the reset
 			if err := s.SetUptime(nodeID, 0, startTime); err != nil {
 				return err
 			}
--- a/vms/platformvm/state/diff_test.go
+++ b/vms/platformvm/state/diff_test.go
@@ -1359,3 +1359,81 @@ func TestDiffResetContinuousValidatorCycleResetsUptime(t *testing.T) {
 	require.Equal(time.Duration(0), upDuration, "uptime should be reset to 0 after cycle restart")
 	require.Equal(continuousValidator.EndTime, lastUpdated, "lastUpdated should be set to new cycle start time")
 }
+
+func TestDiffUptimeResetPropagatesThroughDiffChain(t *testing.T) {
+	require := require.New(t)
+
+	state := newTestState(t, memdb.New())
+
+	blsKey, err := localsigner.New()
+	require.NoError(err)
+
+	continuousValidator := &Staker{
+		ContinuousValidator: ContinuousValidator{
+			AccruedRewards:          10,
+			AccruedDelegateeRewards: 5,
+			AutoRestakeShares:       100_000,
+			ContinuationPeriod:      14 * 24 * time.Hour,
+		},
+		TxID:            ids.GenerateTestID(),
+		NodeID:          ids.GenerateTestNodeID(),
+		PublicKey:       blsKey.PublicKey(),
+		SubnetID:        constants.PrimaryNetworkID,
+		Weight:          10,
+		StartTime:       time.Unix(1000, 0),
+		EndTime:         time.Unix(2000, 0),
+		PotentialReward: 100,
+		NextTime:        time.Unix(2000, 0),
+		Priority:        txs.PrimaryNetworkValidatorCurrentPriority,
+	}
+
+	// Add validator to state
+	require.NoError(state.PutCurrentValidator(continuousValidator))
+	require.NoError(state.Commit())
+
+	// Simulate some uptime accumulation
+	accumulatedUptime := 500 * time.Second
+	require.NoError(state.SetUptime(continuousValidator.NodeID, accumulatedUptime, continuousValidator.StartTime))
+	require.NoError(state.Commit())
+
+	// Verify uptime was set
+	upDuration, _, err := state.GetUptime(continuousValidator.NodeID)
+	require.NoError(err)
+	require.Equal(accumulatedUptime, upDuration)
+
+	// Create diff1 on top of state
+	diff1, err := NewDiffOn(state)
+	require.NoError(err)
+
+	// Create diff2 on top of diff1
+	diff2, err := NewDiffOn(diff1)
+	require.NoError(err)
+
+	// Reset the continuous validator cycle on diff2
+	newWeight := continuousValidator.Weight + 10
+	newPotentialReward := continuousValidator.PotentialReward + 15
+	newAccruedRewards := continuousValidator.AccruedRewards + 20
+	newAccruedDelegateeRewards := continuousValidator.AccruedDelegateeRewards + 25
+
+	require.NoError(
+		diff2.ResetContinuousValidatorCycle(
+			continuousValidator,
+			newWeight,
+			newPotentialReward,
+			newAccruedRewards,
+			newAccruedDelegateeRewards,
+		),
+	)
+
+	// Apply diff2 to diff1 - uptime reset should propagate to diff1
+	require.NoError(diff2.Apply(diff1))
+
+	// Apply diff1 to state - uptime reset should be applied
+	require.NoError(diff1.Apply(state))
+
+	// Verify uptime was reset to 0 with the new start time
+	upDuration, lastUpdated, err := state.GetUptime(continuousValidator.NodeID)
+	require.NoError(err)
+	require.Equal(time.Duration(0), upDuration, "uptime should be reset to 0 after cycle restart through diff chain")
+	require.Equal(continuousValidator.EndTime, lastUpdated, "lastUpdated should be set to new cycle start time")
+}


vms/platformvm/block/executor/proposal_block_test.go
```diff
@@ -107,7 +107,6 @@ func TestApricotProposalBlockTimeVerification(t *testing.T) {
 	)
 	onParentAccept.EXPECT().GetTx(addValTx.ID()).Return(addValTx, status.Committed, nil)
 	onParentAccept.EXPECT().GetCurrentSupply(constants.PrimaryNetworkID).Return(uint64(1000), nil).AnyTimes()
-	onParentAccept.EXPECT().GetDelegateeReward(constants.PrimaryNetworkID, utx.NodeID()).Return(uint64(0), nil).AnyTimes()
```
rrazvan1, line 110: > @yacovm: why is this removed?

vms/platformvm/state/staker.go
```diff
@@ -74,24 +103,93 @@ func (s *Staker) Less(than *Staker) bool {
 	return bytes.Compare(s.TxID[:], than.TxID[:]) == -1
 }
 
-func NewCurrentStaker(
+func NewCurrentValidator(
```
rrazvan1, line 106: > @yacovm: why the rename?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Fee this validator charges delegators as a percentage, times 10,000
+	// For example, if this validator has DelegationShares=300,000 then they
+	// take 30% of rewards from delegators
+	DelegationShares uint32 `serialize:"true" json:"shares"`
+
+	// Weight of this validator used when sampling
+	Wght uint64 `serialize:"true" json:"weight"`
+
+	// Percentage of rewards to auto-restake at the end of each cycle, expressed in millionths (percentage * 10,000).
```
rrazvan1, line 66: > @yacovm: For my education - Why do we need such a fine grained accuracy?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+		return fmt.Errorf("%w: weight %d != stake %d", errValidatorWeightMismatch, tx.Wght, totalStakeWeight)
+	}
+
+	// cache that this is valid
+	tx.SyntacticallyVerified = true
+	return nil
+}
+
+func (*AddContinuousValidatorTx) Visit(_ Visitor) error {
+	// TODO ACP-236: implement properly
```
rrazvan1, line 211: > @yacovm: In which PR is this implemented?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,212 @@
...
+	tx.ValidatorRewardsOwner.InitCtx(ctx)
+	tx.DelegatorRewardsOwner.InitCtx(ctx)
+}
+
+func (*AddContinuousValidatorTx) SubnetID() ids.ID {
+	return constants.PrimaryNetworkID
+}
+
+func (tx *AddContinuousValidatorTx) PublicKey() (*bls.PublicKey, bool, error) {
+	if err := tx.Signer.Verify(); err != nil {
```
rrazvan1, line 100: > @yacovm:
> This object seems immutable to me.
> 
> Why do we need to verify this each time?
> 
> Can't we verify it before we create the object or when we create it?

vms/platformvm/state/staker.go
```diff
@@ -120,6 +218,10 @@ func NewPendingStaker(txID ids.ID, staker txs.ScheduledStaker) (*Staker, error)
 }
 
 func (s *Staker) ValidateMutation(ms *Staker) error {
```
rrazvan1, line 220: > @yacovm:
> The most secure way to validate any mutation is to:
> 
> Compute the expected result
> Check if the expected result matches the proposed value.
> Can we therefore take s, copy it into s', and then apply mutations of s' and then compare s' to ms?

vms/platformvm/block/executor/proposal_block_test.go
rrazvan1: `state.GetDelegateeReward` doesn't exist anymore since it has been moved to the staker struct.

So to get the DelegateeRewards you call, `state.GetCurrentValidator` and get `DelegateeReward` field.

vms/platformvm/state/staker.go
rrazvan1: Before, the staker struct contained common fields of validator and delegator.
Now, it has more fields.

Eg.: before, `DelegateeRewards` was available to be fetched directly from state through `state.GetDelegateeRewards(..)`. Now, you need to fetch the validator, and access `DelegateeRewards` field from the staker struct.

Since `DelegateeRewards` is a validator specific field, we don't need to initialize this value for a delegator.

So, we need different constructors for: `Delegator`, `Validator`, `Continuous Validator`.


Note: I am not happy with the implementation, but a proper one which implies using interfaces, will require changing TOO much code. I feel this is good enough; if you have a better middleground solution, I am all ears.

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: #4909 but will be replaced with a cleaner PR soon.

I added this TX in this PR to use it in the tests, because creating a mock TX similar to this one was adding overhead to understand it, just to remove it after.

vms/platformvm/state/staker.go
rrazvan1: I do not understand your proposed approach.
There isn't only one single possible mutation available.

Currently, there are 3 flows: delegator exit (increasing **delegateeReward**), restaking event or updating autorestake config.

Do you suggest creating 3 specific functions these 3 situations? So we are limiting the ability of mutating the staker in infinite ways.

yacovm: I'm not suggesting creating any more functions.

I'm suggesting to identify what fields can change in the flow (as you pointed out) and:

- Copy the existing instance
- Apply the mutations
- Check if the mutated value is equal to the given value.

Let me give you an example. Is there a field (like the BLS public key) that is not expected to be changed? Instead of checking explicitly that you're not changing the key, it will be checked implicitly by comparing the mutated value to the given value.

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: I used the same denominator as existing in `DelegationShares` field for consistency reasons, and for reusing helper functions.

To answer your question: I don't know the reasons behind choosing this denominator 🙃

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: Because how poorly `signer.ProofOfPossession` was implemented.

```
func (p *ProofOfPossession) Key() *bls.PublicKey {
	return p.publicKey
}
```

```
func (p *ProofOfPossession) Verify() error {
	.....
	p.publicKey = publicKey
	return nil
}
```

So `p.publicKey` isn't available, if `Verify` isn't called.

vms/platformvm/state/diff.go
```diff
@@ -633,9 +634,10 @@ func (d *diff) Apply(baseState Chain) error {
 			}
 		}
 	}
-	for subnetID, nodes := range d.modifiedDelegateeRewards {
-		for nodeID, amount := range nodes {
-			if err := baseState.SetDelegateeReward(subnetID, nodeID, amount); err != nil {
+	// Apply uptime resets for continuous validators that restarted their cycle
+	if s, ok := baseState.(uptime.State); ok {
```
rrazvan1: This is not ok, it needs to be changed.

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,213 @@
...
+		return errMissingSigner
+	}
+
+	for _, out := range tx.StakeOuts {
+		if err := out.Verify(); err != nil {
+			return fmt.Errorf("failed to verify output: %w", err)
+		}
+	}
+
+	firstStakeOutput := tx.StakeOuts[0]
```
yacovm, line 182: can we do a length check here?

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,213 @@
...
+	totalStakeWeight := firstStakeOutput.Output().Amount()
+	for _, out := range tx.StakeOuts[1:] {
+		newWeight, err := safemath.Add(totalStakeWeight, out.Output().Amount())
+		if err != nil {
+			return err
+		}
+		totalStakeWeight = newWeight
+
+		assetID := out.AssetID()
+		if assetID != stakedAssetID {
```
yacovm, line 193: 1. Why do we need the asset IDs to be the same?
2. Can we just have the two following functions on the type `[]*avax.TransferableOutput` ? 
      - TotalOutputWeight() (uint64, error)
      - SameAssert() error

And just call them in this function? Makes things more clear.

vms/platformvm/txs/add_continuous_validator_tx.go
```diff
@@ -0,0 +1,213 @@
...
+			return fmt.Errorf("%w: %q and %q", errMultipleStakedAssets, stakedAssetID, assetID)
+		}
+	}
+
+	switch {
+	case !avax.IsSortedTransferableOutputs(tx.StakeOuts, Codec):
```
yacovm, line 199: why is it a problem if they're not sorted?

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: there is one in the first switch:
```
case len(tx.StakeOuts) == 0:
		return errNoStake
```

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: 1. Because you cannot expect multiple types of assets being returned at the end of staking.
2. That code is duplicated in more places, so creating a helper would make sense to be created at the point where we want to "refactor" and not add new code. And a simple answer: consistency and touching less files.

Based on our discussion, you came up with the idea of directly comparing the stake outs asset id with the `(snow.Context).AVAXAssetID`. Will do that!

vms/platformvm/txs/add_continuous_validator_tx.go
rrazvan1: To have one canonical form of the tx.
cc: @StephenButtolph correct me if I'm wrong

PR #4934, opened 2026-02-03
Fix go-mod-tidy and ensure it is checked in CI

maru-ava:
## Why this should be merged

Previous PRs assumed that setting `dir:` in a task command was sufficient to allow `go mod tidy` to operate in the context of a given go module's go.mod. In reality, `GOWORK=off` is required to escape workspace mode and `cd [dir]` is required to ensure that `go mod tidy` operates as expected.

Also, check-go-mod-tidy was previously not being run in CI for some reason. 
 
## How this was tested

CI

N/A


maru-ava:
> I am presuming that go.mod/go.sum changes are correct -- #4933 must be merged and this branch updated for CI to pass.

Good catch, I really should have just included those changes in this PR.

maru-ava:
Rebased

PR #4938, opened 2026-02-03
test: Add test logger

alarso16:
## Why this should be merged

I often find myself adding copious print statements to test logging details when my test fails. However, zap provides a testing helper that only logs if the test fails or if verbose. We can use this, and even add to it if one needs to verify log output.

## How this works

Adds a wrapper for the `logging.Logger` that logs to `t`, and fails if there are any warnings. Also removes `logger.NoWarn`, since this implements the same functionality essentially, but also reports what the failure is.

## How this was tested

Added a unit test, as well as used in many places.

No


alarso16:
Just learned about https://github.com/ava-labs/avalanchego/blob/master/tests/log.go, which is likely sufficient for most use cases

PR #4940, opened 2026-02-03
[ci] Configure dir exclusion from linting via .golangci.yml

maru-ava:
## Why this should be merged

The need to exclude new directories from linting for Goland and soon Bazel suggested a simple way of maintaining a list of such directories.

## How this was tested

CI

N/A

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -67,7 +67,7 @@ linters:
     - govet
     - importas
     - ineffassign
-    - mirror
+    - mirror 
     - misspell
     - nakedret
     - nilerr


scripts/lint.sh
```diff
@@ -2,8 +2,25 @@
...
+EXCLUDE_DIRS=()
+while IFS= read -r dir; do
+  [[ -n "$dir" ]] && EXCLUDE_DIRS+=("$dir")
+done < <(
+  sed -n '/^  skip-dirs:/,/^  [a-z]/p' .golangci.yml \
+    | grep '^\s*-' \
+    | sed 's/^\s*-\s*//' \
+    | sed 's/\s*#.*//'
+)
+
```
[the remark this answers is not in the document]
maru-ava: What do you mean silently continue? Doesn't `set -e` prevent that?

.golangci.yml
```diff
@@ -19,6 +19,13 @@ run:
   build-tags:
     - test
 
+  # Directories to skip (in addition to defaults like vendor, third_party, etc.)
+  skip-dirs:
+    - graft    # Grafted modules have their own lint config
```
alarso16: Silent strange behavior: if we ever added `graft/coreth/graft`, then it wouldn't be linted at all. So I guess we just shouldn't do that...

scripts/lint.sh
```diff
@@ -102,8 +131,14 @@ function test_interface_compliance_nil {
 }
 
 function test_import_testing_only_in_tests {
-  ROOT=$( git rev-parse --show-toplevel )
-  NON_TEST_GO_FILES=$( find "${ROOT}" -iname '*.go' ! -iname '*_test.go' ! -path "${ROOT}/tests/*" ! -path "${ROOT}/${IGNORE_PATH}/*" );
+  local root
+  root=$(git rev-parse --show-toplevel)
```
alarso16:
```suggestion
  local root=$(git rev-parse --show-toplevel)
```
Any reason not to do this? (my understanding of bash syntax is not great)

alarso16 (approved):
I'm not proficient with sed, but the logic seems reasonable and clean

scripts/lint.sh
maru-ava: https://www.shellcheck.net/wiki/SC2155

.golangci.yml
maru-ava: Good catch. I've reverted to using a list defined in the script and switched to always using rooted paths with find (the grep search was prone to this same error).

scripts/lint.sh
maru-ava: Reverted

.golangci.yml
maru-ava: So, reverted.

Elvis339 (approved):

PR #4942, opened 2026-02-03
[bazel] Generate BUILD.bazel files

maru-ava:
## Why this should be merged

Includes all the generated files required by #4544. This PR should only be merged once it's parent is otherwise ready to merge to simplify the review process.

## How this was tested

CI

N/A

## TODO

 - [ ] Wait for approval of #4544


PlatCore (approved):
LGTM

Assuming:

- The entire PR simply generated Bazel metadata using the `gazelle` tool.
- We assume that the output from `gazelle` is correct.

maru-ava:
> LGTM
> 
> Assuming:
> 
> * The entire PR simply generated Bazel metadata using the `gazelle` tool.
> * We assume that the output from `gazelle` is correct.

It should be possible to verify these assumptions by running `task bazel-gazelle-generate` and seeing no change in the working tree and inspecting the task definition. Also, the new bazel ci jobs pass once gazelle has done its work.

PR #4944, opened 2026-02-04
Add continuous staking transactions

rrazvan1:
## Why this should be merged
Enables continuous staking through new introduced transactions.

## How this works
Introduces new P-Chain transactions, reward calculation, wallet builder/signer support, and tx execution/verification logic.

## How this was tested
Unit tests.

Yes.

### DIFF
--- a/vms/platformvm/block/executor/options.go
+++ b/vms/platformvm/block/executor/options.go
@@ -9,6 +9,7 @@ import (
 
 	"go.uber.org/zap"
 
+	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/uptime"
 	"github.com/ava-labs/avalanchego/utils/constants"
@@ -141,12 +142,17 @@ func (*options) ApricotAtomicBlock(*block.ApricotAtomicBlock) error {
 }
 
 func (o *options) prefersCommit(tx *txs.Tx) (bool, error) {
-	unsignedTx, ok := tx.Unsigned.(*txs.RewardValidatorTx)
-	if !ok {
+	var txID ids.ID
+	switch unsignedTx := tx.Unsigned.(type) {
+	case *txs.RewardValidatorTx:
+		txID = unsignedTx.TxID
+	case *txs.RewardContinuousValidatorTx:
+		txID = unsignedTx.TxID
+	default:
 		return false, fmt.Errorf("%w: %T", errUnexpectedProposalTxType, tx.Unsigned)
 	}
 
-	stakerTx, _, err := o.state.GetTx(unsignedTx.TxID)
+	stakerTx, _, err := o.state.GetTx(txID)
 	if err != nil {
 		return false, fmt.Errorf("%w: %w", errFailedFetchingStakerTx, err)
 	}
--- a/vms/platformvm/api/validator.go
+++ b/vms/platformvm/api/validator.go
@@ -86,6 +86,9 @@ type PermissionlessValidator struct {
 	Staked                 []UTXO                    `json:"staked,omitempty"`
 	Signer                 *signer.ProofOfPossession `json:"signer,omitempty"`
 
+	// ACP-236 fields
+	ConfigOwner *Owner `json:"configOwner,omitempty"`
+
 	// The delegators delegating to this validator
 	DelegatorCount  *json.Uint64        `json:"delegatorCount,omitempty"`
 	DelegatorWeight *json.Uint64        `json:"delegatorWeight,omitempty"`
--- a/vms/platformvm/client.go
+++ b/vms/platformvm/client.go
@@ -653,12 +653,37 @@ func GetDeactivationOwners(
 	return deactivationOwners, nil
 }
 
+func GetContinuousValidatorsOwners(
+	c *Client,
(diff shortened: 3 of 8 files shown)


vms/platformvm/reward/calculator.go
```diff
@@ -81,3 +82,20 @@ func Split(totalAmount uint64, shares uint32) (uint64, uint64) {
...
+// ProportionalAmount computes (amount * numerator) / denominator with overflow safety.
+// The result is floored (rounded down) due to integer division. When splitting a
+// value into parts, consider whether rounding errors should favor one side:
+//   - To ensure you don't over-allocate: compute the smaller portion with this
+//     function, then subtract from total to get the remainder.
+func ProportionalAmount(amount, numerator, denominator uint64) (uint64, error) {
+	if denominator == 0 {
+		return 0, errors.New("denominator is zero")
+	}
+
```
yacovm, line 95: I'm confused. If the math.Mul overflows then we just ignore it and return the result?

vms/platformvm/txs/executor/proposal_tx_executor.go
```diff
@@ -414,6 +429,104 @@ func (e *proposalTxExecutor) RewardValidatorTx(tx *txs.RewardValidatorTx) error
 	return nil
 }
 
+func (e *proposalTxExecutor) RewardContinuousValidatorTx(tx *txs.RewardContinuousValidatorTx) error {
+	currentChainTime := e.onCommitState.GetTimestamp()
```
yacovm, line 433: can we add a check that:

```
	if tx.Timestamp > math2.MaxInt64 {
		return fmt.Errof("timestamp too big")
	}
```

?

vms/platformvm/txs/executor/proposal_tx_executor.go
```diff
@@ -414,6 +429,104 @@ func (e *proposalTxExecutor) RewardValidatorTx(tx *txs.RewardValidatorTx) error
...
+		return ErrShouldBePermissionlessStaker
+	}
+
+	continuousStaker, ok := stakerTx.Unsigned.(txs.ContinuousStaker)
+	if !ok {
+		return ErrShouldBeContinuousStaker
+	}
+
+	// in [onAbortState] the staker should be removed despite the continuation state.
+	e.onAbortState.DeleteCurrentValidator(stakerToReward)
```
yacovm, line 485: why do we delete the validator if we abort? Can't it get rewards in future cycles?

yacovm: Is this a "as long as you behave nice, we will let you validate continuously" kind of staking?

vms/platformvm/txs/executor/proposal_tx_executor.go
```diff
@@ -636,3 +749,487 @@ func (e *proposalTxExecutor) rewardDelegatorTx(uDelegatorTx txs.DelegatorTx, del
...
+// setOnCommitStateContinuousValidatorRestake processes rewards for a running
+// continuous validator based on their AutoRestakeShares configuration.
+//
+// The function:
+//  1. Splits rewards (validation + delegatee) into restaking and withdrawing portions
+//  2. Creates UTXOs for the withdrawn portion
+//  3. Increases validator weight and accrued rewards by the restaking portion
+//  4. If restaking would exceed MaxValidatorStake, the excess is withdrawn
+//  5. Updates the validator state
+func (e *proposalTxExecutor) setOnCommitStateContinuousValidatorRestake(validatorTx txs.ValidatorTx, validator *state.Staker) error {
```
yacovm, line 836: can we make this function a bit smaller?

vms/platformvm/block/builder/builder.go
```diff
@@ -323,7 +323,12 @@ func buildBlock(
 		return nil, fmt.Errorf("could not find next staker to reward: %w", err)
 	}
 	if shouldReward {
-		rewardValidatorTx, err := NewRewardValidatorTx(builder.txExecutorBackend.Ctx, stakerTxID)
+		stakerTx, _, err := parentState.GetTx(stakerTxID)
```
yacovm, line 326: `getNextStakerToReward` already fetches the staker, but only returns its txID and here we again fetch the staker by txID... can we make it more efficient?

vms/platformvm/block/executor/options.go
```diff
@@ -141,12 +142,17 @@ func (*options) ApricotAtomicBlock(*block.ApricotAtomicBlock) error {
 }
 
 func (o *options) prefersCommit(tx *txs.Tx) (bool, error) {
```
yacovm: This function is invoked by the Banff block, shouldn't we need here a new block type?

vms/platformvm/reward/calculator.go
```diff
@@ -81,3 +82,20 @@ func Split(totalAmount uint64, shares uint32) (uint64, uint64) {
...
+	result := amount / denominator * numerator
```
DracoLi, line 96: is there a bug here? integer division truncates to 0 when amount < denominator. Then if `amount * numerator` overflows, we just return 0.
This test case fails.
```
		{
			name:        "overflow with amount less than denominator",
			amount:      300 * units.Avax,
			numerator:   200 * units.Avax,
			denominator: 400 * units.Avax,
			expected:    150 * units.Avax,
		}
```

PR #4945, opened 2026-02-04
refactor: Remove required values from x/sync config

alarso16:
## Why this should be merged

Config values should be optional, especially when most of these values have reasonable defaults.

## How this works

Moves non-optional values to the constructor, while all optional values can be provided via config, with seemingly normal default values.

## How this was tested

All existing integration tests, plus some unit tests ensuring the errors returned are expected.

No


alarso16:
Don' think this will ever get merged, if I ever want to, I'll remake it

PR #4948, opened 2026-02-05
Add logging to merklesync

alarso16:
## Why this should be merged

This adds a ticker that occasionally logs sync progress. The progress is estimated by converting keys to big ints, and some math :)

## How this works

Adds a method to find what percentage of a `workHeap` is taken by a specific node ID. This is used by the syncer to occasionally log this percentage

## How this was tested

I added a test for the percentage calculation, but I don't know how to test the logging. 

No


database/merkle/sync/syncer.go
```diff
@@ -290,6 +293,24 @@ func (s *Syncer[_, _]) workLoop(ctx context.Context) {
...
+	for {
+		select {
+		case <-s.doneChan:
+			return
+		case <-ticker.C:
+			s.workLock.Lock()
+			root := s.getTargetRoot()
+			percentage := s.unprocessedWork.Status(root)
+			s.workLock.Unlock()
+			s.config.Log.Info("syncing progress", zap.Float64("percent complete", percentage), zap.Stringer("target root", root))
```
powerslider: `unprocessedWork.Status(root)` returns the percentage of keyspace covered by unprocessed work items, i.e. the work that still needs to be done. At the start of sync the entire range is in `unprocessedWork` (100%), and as ranges complete they get removed. So the raw value is "percent remaining", not "percent complete".

Also nit: `getTargetRoot()` acquires its own `syncTargetLock`, so there's no need to hold `workLock` while calling it. Move it outside the `workLock` critical section to reduce contention.

```suggestion
			root := s.getTargetRoot()
			s.workLock.Lock()
			remaining := s.unprocessedWork.Status(root)
			s.workLock.Unlock()
			s.config.Log.Info("syncing progress", zap.Float64("percent complete", 100-remaining), zap.Stringer("target root", root))
```

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+func maybeToBig(b maybe.Maybe[[]byte], maxLength int) *big.Int {
+	if !b.HasValue() {
+		return nil
+	}
+	s := b.Value()
+	if len(b.Value()) > maxLength {
+		s = s[:maxLength]
+	}
+
+	// Pad with zeros to the left.
```
powerslider: Isn't this actually right padding? - `copy(padded, s)` copies to the start of the destination slice - this is right-padding (trailing zeros), not left-padding (leading zeros).
```suggestion
    // Right-pad with zeros so that short keys occupy the most significant bytes.
```

alarso16: Yes it is - very incorrect comment

database/merkle/sync/syncer.go
alarso16: This was actually a bug - I meant to check the processed work, which I think is the more accurate fix anyway

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+		// Determine the start value (0x0 if no value)
+		start := maybeToBig(item.start, statusBytes)
+		if start == nil {
+			start = big.NewInt(0)
+		}
+
+		// Determine the end value (max if no value)
+		end := maybeToBig(item.end, statusBytes)
+		if end == nil {
+			end = new(big.Int).Lsh(big.NewInt(1), uint(statusBytes*8)) // 2^(maxSize*8)
```
powerslider: This duplicates the package-level `totalSpace` variable. Reuse it:

```suggestion
                    end = new(big.Int).Set(totalSpace)
```

database/merkle/sync/syncer.go
```diff
@@ -289,6 +292,24 @@ func (s *Syncer[_, _]) workLoop(ctx context.Context) {
...
+	for {
+		select {
+		case <-s.doneChan:
+			return
+		case <-ticker.C:
+			root := s.getTargetRoot()
+			s.workLock.Lock()
+			percentage := s.processedWork.Status(root)
+			s.workLock.Unlock()
+			s.config.Log.Info("syncing progress", zap.Float64("percent complete", percentage), zap.Stringer("target root", root))
```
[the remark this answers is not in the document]
powerslider: yes, this is correct

powerslider (approved):

database/merkle/sync/syncer.go
```diff
@@ -289,6 +292,26 @@ func (s *Syncer[_, _]) workLoop(ctx context.Context) {
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-s.doneChan:
+			return
+		case <-ticker.C:
+			root := s.getTargetRoot()
+			s.workLock.Lock()
+			percentage := s.processedWork.Status(root)
+			s.workLock.Unlock()
```
joshua-kim: nit: I generally find these confusing even when used in simple scopes -- I generally try to refactor these into a helper function so i can unlock w/ defer.

database/merkle/sync/syncer.go
```diff
@@ -289,6 +292,26 @@ func (s *Syncer[_, _]) workLoop(ctx context.Context) {
...
+			s.config.Log.Info("syncing progress", zap.Float64("percent_complete", percentage), zap.Stringer("target root", root))
```
joshua-kim: nit: inconsistency in snake case/spacing. Rest of logs use spacing so might make sense to be consistent here.
nit: consider sig figs in percentage if we want to truncate or not.

database/merkle/sync/workheap_test.go
```diff
@@ -308,3 +308,53 @@ func TestWorkHeapMergeInsertRandom(t *testing.T) {
 		}
 	}
 }
+
+func TestWorkHeapStatus(t *testing.T) {
+	require := require.New(t)
```
joshua-kim: nit: makes no difference here but we generally avoid this to avoid subtesting footguns

database/merkle/sync/workheap_test.go
```diff
@@ -308,3 +308,53 @@ func TestWorkHeapMergeInsertRandom(t *testing.T) {
...
+		start:       maybe.Some([]byte{128}),
+		end:         maybe.Some([]byte{192}),
+		priority:    highPriority,
+		localRootID: otherID,
+	})
+	pct = h.Status(id)
+	require.InDelta(75.0, pct, 0.0001)
+	pct = h.Status(otherID)
+	require.InDelta(25.0, pct, 0.0001)
+	_ = h.GetWork() // remove to not corrupt next calculation
```
joshua-kim: This feels like we're stuffing multiple tests into a single test -- can we use subtests or a test table here?

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
 	return wh.innerHeap.Len()
 }
+
+// statusBytes is the number of bytes of the keyspace to consider
+// when calculating the progress percentage.
+// Keys larger than this will be truncated.
+const statusBytes = 64
+
+var (
+	totalSpace      = new(big.Int).Lsh(big.NewInt(1), statusBytes*8)
```
joshua-kim: 1. Similar comment on the rename
2. nit: Why is this Lsh so that the result is `1 0 0 0 ...`? Isn't the maximum key going to be `1 1 1 1 ...` ? We don't care about precision so I guess it doesn't matter...

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+	totalSpaceFloat = new(big.Float).SetInt(totalSpace)
+)
+
+// Status returns the approximate percentage of work in the heap
+// for a given [ids.ID] root, relative to the entire keyspace.
+// If there are many keys larger than 64 bytes, the returned percentage
+// may be inaccurate.
+func (wh *workHeap) Status(root ids.ID) float64 {
```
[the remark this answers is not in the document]
joshua-kim: +1 Status suggests that we're returning some bool/enum/etc.

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
 func (wh *workHeap) Len() int {
 	return wh.innerHeap.Len()
 }
+
+// statusBytes is the number of bytes of the keyspace to consider
+// when calculating the progress percentage.
+// Keys larger than this will be truncated.
+const statusBytes = 64
```
joshua-kim: consider renaming -- this seems to represent maximum/upper bound for the key space to consider

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+	progressFloat := new(big.Float).SetInt(progress)
+	progressFloat.Quo(progressFloat, totalSpaceFloat)
+	progressFloat.Mul(progressFloat, big.NewFloat(100))
+
+	pct, _ := progressFloat.Float64()
+
+	return pct
+}
+
+func maybeToBig(b maybe.Maybe[[]byte], maxLength int) *big.Int {
```
joshua-kim: This isn't really a conversion so the naming is misleading... maybe call this `truncateBig`?

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+	if !b.HasValue() {
+		return nil
+	}
+	s := b.Value()
```
joshua-kim: nit: `v` for value? What is `s` supposed to mean? I could understand `k` for key as well but it's not clear to me what `s` is. Slice?

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+	if len(b.Value()) > maxLength {
+		s = s[:maxLength]
+	}
+
+	// Right-pad with zeros so that short keys occupy the most significant bytes.
+	padded := make([]byte, maxLength)
+	copy(padded, s)
```
joshua-kim: `copy` will only copy the min of dst/src, so I think we can simplify by doing:
```
result := make([]byte, maxLength) // 64 bytes
copy(result, s) 
```

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+// may be inaccurate.
+func (wh *workHeap) Status(root ids.ID) float64 {
+	progress := new(big.Int)
+	wh.sortedItems.Ascend(func(item *workItem) bool {
+		if item.localRootID != root {
+			return true
+		}
+
+		// Determine the start value (0x0 if no value)
+		start := maybeToBig(item.start, statusBytes)
```
joshua-kim: This doesn't need to be a parameter since its usage always uses the same global

database/merkle/sync/syncer.go
alarso16: Easy enough

database/merkle/sync/syncer.go
alarso16: sig figs seems reasonable

database/merkle/sync/workheap_test.go
alarso16: That is an option, but I don't think the second part is even necessary, so I'll just delete it to avoid this janky state change

database/merkle/sync/workheap.go
alarso16: ummm probably. i'll name it something clearer

database/merkle/sync/workheap.go
alarso16: Yeah I think you're right. i was worried about the additional empty key, but it doesn't matter. I ended up changing the whole thing to uint64 to make it simpler though

database/merkle/sync/workheap.go
```diff
@@ -157,3 +159,71 @@ func (wh *workHeap) remove(item *workItem) {
...
+	}
+	s := b.Value()
+	if len(b.Value()) > maxLength {
+		s = s[:maxLength]
+	}
+
+	// Right-pad with zeros so that short keys occupy the most significant bytes.
+	padded := make([]byte, maxLength)
+	copy(padded, s)
+
```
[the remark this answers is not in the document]
alarso16: Started using a format, it's now just uint64s

joshua-kim (approved):

PR #4952, opened 2026-02-08
version: prevent panic on nil peer in Compatible and add test

Ocheretovich:
Prevent panic when Compatible is called with a nil peer. Adds a defensive nil check and a unit test to ensure nil peer is treated as incompatible and does not cause a panic.

### DIFF
--- a/version/compatibility.go
+++ b/version/compatibility.go
@@ -28,6 +28,10 @@ type Compatibility struct {
 // This means that the version is connectable and that consensus messages can be
 // made to the peer.
 func (c *Compatibility) Compatible(peer *Application) bool {
+	if peer == nil {
+		return false
+	}
+
 	if c.Current.Major < peer.Major {
 		return false // If we are on an older major version, we are incompatible.
 	}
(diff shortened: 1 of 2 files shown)


maru-ava (commented):
Given that this change is superficially suggested but incapable of occurring at runtime, I'm assuming this PR was prompted by an AI-assisted search? 

The only caller is [peer.shouldDisconnect()](https://github.com/ava-labs/avalanchego/blob/master/network/peer/peer.go#L717), which passes `p.version`. That field is set [during the handshake](https://github.com/ava-labs/avalanchego/blob/master/network/peer/peer.go#L898) (peer.go:898) before `shouldDisconnect` is ever called — first in the handshake handler itself, then periodically via `Ping`. There is no code path where `p.version` is nil when `Compatible` is called.

This is an internal function with a single call site. Adding a nil guard here masks potential bugs rather than surfacing them — if `p.version` were somehow `nil`, a panic with a clear stack trace is far more useful for debugging than silently returning false and causing mysterious peer disconnections.

Ocheretovich:
Makes sense, thanks for taking the time to explain. Agreed this shouldn't happen at runtime and it's better not to mask a potential bug

PR #4955, opened 2026-02-10
Allow fallback to let every node propose a block in a non primary subnet

yacovm:
## Why this should be merged

The motivation behind the feature is to be able to recover networks in which the entire validator set has failed, but there are non validating nodes that have replicated the chain.
    these non-validating nodes, when made validators, would normally refuse proposing a new block, because the subnet chain's proposerVM latest P-chain height points to a validator set that does not contain these validators.
    
This change makes these nodes propose a block anyway, if the parent timestamp of the block is more than the specified wait time in the past.
    
Once the chain proposes a new block, it will point to the updated P-chain height and the aforementioned non-validating nodes will then consider themselves as validators, and the system operators can then safely revert the config change and remove the emergency flags.
    

## How this works

This commit introduces two emergency config flags:
    
-  proposervm-fallback-nonvalidator-can-propose: Toggles the feature on/off (off by default)
- proposervm-fallback-nonvalidator-wait-max-time: Specifies the maximum wait time proposerVM will wait to propose a block
    
When both config flags are enabled (can-propose is true, wait-time is > 0), the proposerVM will ask the VM to build a block, even if the node is not a validator in the validator set that the P-chain height of the subnet chain points to.
    

## How this was tested

Added a unit test as well as an e2e test (which is removed in the second commit), can be accessed by navigating to the first commit.


vms/proposervm/block.go
```diff
@@ -416,6 +416,14 @@ func (p *postForkCommonComponents) verifyPostDurangoBlockDelay(
 		parentPChainHeight,
 		currentSlot,
 	)
+
+	if p.vm.shouldFallbackToEmergencyProposer(parentTimestamp) && expectedProposerID != proposerID {
+		p.vm.ctx.Log.Warn("fallback to emergency proposer due to block built too long ago, setting expected proposer",
+			zap.String("expectedProposer", expectedProposerID.String()),
+			zap.String("currentProposer", proposerID.String()))
+		expectedProposerID = proposerID
+	}
```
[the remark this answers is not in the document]
yacovm: This is a risk we are willing to take

vms/proposervm/vm.go
```diff
@@ -536,6 +546,44 @@ func (vm *VM) timeToBuild(ctx context.Context) (time.Time, bool, error) {
 	return nextStartTime, true, nil
 }
 
+func (vm *VM) shouldFallbackToEmergencyProposer(parentTimestamp time.Time) bool {
+	// If we're a VM in the primary network, we cannot fall back to an emergency block proposer
+	if vm.ctx.SubnetID == constants.PrimaryNetworkID {
+		return false
+	}
+
+	vm.ctx.Log.Debug("checking if we should fallback to emergency block proposer",
```
DracoLi: should this log be after the setting enabled check? we probably don't care about this if the node did not enable this.

config/config.go
```diff
@@ -1352,6 +1352,14 @@ func GetNodeConfig(v *viper.Viper) (node.Config, error) {
 	nodeConfig.ProposerMinBlockDelay = v.GetDuration(ProposerVMMinBlockDelayKey)
 	nodeConfig.SubnetConfigs = subnetConfigs
 
+	nodeConfig.FallbackNonValidatorCanPropose = v.GetBool(FallbackNonValidatorCanProposeKey)
+	nodeConfig.FallbackProposerMaxWaitTime = v.GetDuration(FallbackNonValidatorProposerMaxWaitTimeKey)
+
+	if nodeConfig.FallbackNonValidatorCanPropose && (nodeConfig.FallbackProposerMaxWaitTime < time.Second ||
+		nodeConfig.FallbackProposerMaxWaitTime > time.Hour) {
+		return node.Config{}, fmt.Errorf("%s must be between 1 second and 1 hour", FallbackNonValidatorProposerMaxWaitTimeKey)
```
DracoLi, line 1360: is 1s too short? when would a value of 1s be appropriate?

vms/proposervm/block.go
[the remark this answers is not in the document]
DracoLi: what are the risks here? can a node propose a block before the wait period by changing the system time?

DracoLi (commented):
why was the e2e test removed?

yacovm:
> why was the e2e test removed?

There is an unofficial habit/agreement within the contributors to avalanchego that we should keep e2e tests to the necessary minimum and not make a test for every possible scenario. 

I personally think that having lots of e2e tests is worthwhile, provided that we have an easy framework to write them.

I am therefore on the fence here regarding this niche corner case that I'm fixing.

config/config.go
yacovm: It's just to prevent us from busy-wait-ing in that loop in the proposerVM. 

So if someone puts "1ms" by accident, we won't hog the CPU too much.

I put 1 second as an arbitrary lower bound that makes sense.

vms/proposervm/vm.go
yacovm: fair, I can move it after the check.

vms/proposervm/block.go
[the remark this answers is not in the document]
yacovm: A node can do anything it wants, since we are in a permissionless byzantine setting. 

Anyone can run its own fork of avalanchego and do whatever he/she pleases.

Copilot was concerned that relying on clock time in validation is non deterministic, which is true.

However, it's a binary choice - either the block is accepted entirely, or it is not. Therefore, we are not endangering here consistency, but only liveness. 

But that's fine, because eventually the clocks of all nodes will advance far enough to deem this block as valid.

DracoLi (approved):

yacovm:
Closing due to lack of interest

PR #4957, opened 2026-02-10
feat: include `.rpm`  files with release

JonathanOppenheimer:
## Why this should be merged

A request by @martineckardt. Adds RPM package generation for avalanchego and subnet-evm using https://nfpm.goreleaser.com/ as suggested by @demosdemon. 

Do note that  the `amd64/arm64` workflow files are nearly identical -- this mirrors how the existing Debian workflows are structured, although they appear to be written years ago. I think if we want to consolidate with a matrix strategy, (I would want to do so) that should be done across all workflows (deb, tgz, rpm) in a follow-up PR.

## How this works

It works how it looks 

## How this was tested

- Generated RPMs locally with nfpm -- both avalanchego and subnet-evm produce valid RPM v3.0 packages

Maybe?

### DIFF
--- a/scripts/actionlint.sh
+++ b/scripts/actionlint.sh
@@ -3,7 +3,8 @@
 set -euo pipefail
 
 AVALANCHE_PATH="$(cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )"
-"${AVALANCHE_PATH}"/scripts/run_tool.sh actionlint "${@}"
+# Only lint top-level workflow files (excludes non-workflow YAML like nfpm configs in subdirectories)
+"${AVALANCHE_PATH}"/scripts/run_tool.sh actionlint "${AVALANCHE_PATH}"/.github/workflows/*.yml "${AVALANCHE_PATH}"/.github/workflows/*.yaml "${@}"
 
 echo "Checking use of scripts/* in GitHub Actions workflows..."
 SCRIPT_USAGE=


.github/workflows/build-rpm-amd64-release.yml
```diff
@@ -0,0 +1,79 @@
+name: build-amd64-rpm-packages
```
JonathanOppenheimer, line 1: This file is practically identical to https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/build-ubuntu-amd64-release.yml

.github/workflows/build-rpm-arm64-release.yml
```diff
@@ -0,0 +1,79 @@
+name: build-arm64-rpm-packages
```
JonathanOppenheimer, line 1: This file is practically identical (fr differs by 8 lines) to `.github/workflows/build-rpm-amd64-release.yml`, and is also nearly equivalent to https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/build-ubuntu-arm64-release.yml

.github/workflows/build-rpm-pkg.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
JonathanOppenheimer, line 1: See https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/build-deb-pkg.sh

JonathanOppenheimer:
As noted by @maru-ava 

> the binaries produced by [this] PR won't be compatible with whatever host the rpm would be deployed to. The workflow you copied packages a dynamically-linked binary on a debian-flavored distro for other debian-flavored distros. We almost certainly need to produce a statically-linked binary so as to not depend on anything in the target userspace.

PR #4959, opened 2026-02-11
[ci] Add RPM release workflow

maru-ava:
## Why this should be merged

Add nfpm-based RPM packaging that builds inside a Rocky Linux 9 container (glibc 2.34) to target RHEL 9.x. Both packages are GPG signed and validated end-to-end (signature verification, installation, smoke test) before output.

## How this was tested

- [x] New CI workflow validating rpm creation
- [x] [Validate upload to s3](https://github.com/ava-labs/avalanchego/actions/runs/21955322630) 

N/A

### DIFF
--- a/.github/packaging/Taskfile.yml
+++ b/.github/packaging/Taskfile.yml
@@ -6,6 +6,10 @@
 version: '3'
 
 vars:
+  # Repo root, computed from this Taskfile's location so that the
+  # packaging Taskfile works both as an include and standalone.
+  REPO_ROOT:
+    sh: cd "{{.TASKFILE_DIR}}/../.." && pwd
   # Go version from the project's go.mod.
   PACKAGING_GO_VERSION:
     sh: go list -m -f '{{ "{{.GoVersion}}" }}' | head -1
@@ -25,7 +29,7 @@ vars:
   PACKAGING_TAG:
     sh: echo "${PACKAGING_TAG:-v0.0.0}"
   PACKAGING_DOCKER_IMAGE: avalanchego-rpm-builder
-  PACKAGING_OUTPUT_DIR: '{{.ROOT_DIR}}/build/rpm'
+  PACKAGING_OUTPUT_DIR: '{{.REPO_ROOT}}/build/rpm'
 
 tasks:
   default:
@@ -45,9 +49,9 @@ tasks:
     env:
       GO_VERSION: '{{.PACKAGING_GO_VERSION}}'
       DOCKER_IMAGE: '{{.PACKAGING_DOCKER_IMAGE}}'
-      CONTEXT_DIR: '{{.ROOT_DIR}}/.github/packaging'
+      CONTEXT_DIR: '{{.REPO_ROOT}}/.github/packaging'
     cmds:
-      - cmd: '{{.ROOT_DIR}}/.github/packaging/scripts/build-builder-image.sh'
+      - cmd: '{{.REPO_ROOT}}/.github/packaging/scripts/build-builder-image.sh'
 
   build-avalanchego-rpm:
     desc: Builds RPM for avalanchego
@@ -59,7 +63,7 @@ tasks:
       - cmd: mkdir -p {{.PACKAGING_OUTPUT_DIR}}
       - cmd: >-
           docker run --rm
-          -v {{.ROOT_DIR}}:/build
+          -v {{.REPO_ROOT}}:/build
           -v {{.PACKAGING_OUTPUT_DIR}}:/output
           {{if .RPM_GPG_KEY_FILE}}-v {{.RPM_GPG_KEY_FILE}}:{{.RPM_GPG_KEY_FILE}}:ro{{end}}
           -e PACKAGE=avalanchego
@@ -83,7 +87,7 @@ tasks:
       - cmd: mkdir -p {{.PACKAGING_OUTPUT_DIR}}
       - cmd: >-
           docker run --rm
-          -v {{.ROOT_DIR}}:/build
+          -v {{.REPO_ROOT}}:/build
           -v {{.PACKAGING_OUTPUT_DIR}}:/output
           {{if .RPM_GPG_KEY_FILE}}-v {{.RPM_GPG_KEY_FILE}}:{{.RPM_GPG_KEY_FILE}}:ro{{end}}
           -e PACKAGE=subnet-evm
@@ -109,4 +113,4 @@ tasks:
       TAG: '{{.PACKAGING_TAG}}'
       GIT_COMMIT: '{{.PACKAGING_GIT_COMMIT}}'
     cmds:
-      - cmd: '{{.ROOT_DIR}}/.github/packaging/scripts/validate-rpm.sh'
+      - cmd: '{{.REPO_ROOT}}/.github/packaging/scripts/validate-rpm.sh'
--- a/.github/workflows/build-rpm-release.yml
+++ b/.github/workflows/build-rpm-release.yml
@@ -31,6 +31,24 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.tag || github.ref }}
+
+      # When building an older tag via workflow_dispatch, the tag's tree
+      # may not contain the packaging scripts. Overlay them from the
+      # workflow branch so that packaging:test-build-rpms is available.
+      - name: Overlay packaging scripts from workflow branch
+        if: github.event.inputs.tag
+        uses: actions/checkout@v4
+        with:
+          path: .packaging-overlay
+          sparse-checkout: .github/packaging
+      - name: Apply packaging overlay
+        if: github.event.inputs.tag
+        run: |
+          cp -r .packaging-overlay/.github/packaging .github/packaging
+          rm -rf .packaging-overlay
+        shell: bash
 
       - uses: ./.github/actions/setup-go-for-project
 
@@ -56,7 +74,7 @@ jobs:
         shell: bash
 
       - name: Build and validate RPMs
-        run: ./scripts/run_task.sh packaging:test-build-rpms
+        run: ./scripts/run_task.sh --taskfile .github/packaging/Taskfile.yml test-build-rpms
         env:
           PACKAGING_TAG: ${{ env.TAG }}
           RPM_GPG_KEY_FILE: ${{ env.GPG_KEY_FILE }}


docs/design/rpm-packaging.md
```diff
@@ -0,0 +1,167 @@
...
+  platform constraints ([rules_rust#2726][4])
+- Building `tikv-jemalloc-sys` under Bazel requires a
+  `rules_foreign_cc` workaround since it uses autotools
+
+There is a residual risk: a binary built and tested against an older
+glibc but run against a newer one in production. glibc's forward
+compatibility is strong in practice, but newer versions can change
+behavior in subtle ways (bug fixes that code inadvertently depended
+on, performance characteristics of allocator or threading primitives,
+new default security hardening). This risk is low but non-zero, and is
```
maru-ava, line 171: This is the problem containers largely solve since we to ship an exact userspace (including glibc) with our binary. The kernel is still a wild-card, but most distros ship good kernels.

.github/packaging/Taskfile.yml
```diff
@@ -0,0 +1,118 @@
+# RPM packaging tasks for avalanchego and subnet-evm.
+#
+# Builds RPMs inside a Rocky Linux 9 container (glibc 2.34) with GPG signing.
+# All tasks require TAG to be set to a git tag (e.g., TAG=v1.14.1).
+
+version: '3'
+
+vars:
+  # Go version from the project's go.mod.
```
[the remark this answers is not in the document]
maru-ava: This is necessary escaping for go-task and the alternative would be worse.

.github/packaging/scripts/build-rpm.sh
```diff
@@ -0,0 +1,185 @@
...
+        echo "avalanchego --version: ${output}"
+
+        if [[ "${output}" != avalanchego/* ]]; then
+            echo "ERROR: --version output does not start with 'avalanchego/'" >&2
+            exit 1
+        fi
+
+        # Verify git commit hash is present in output
+        # shellcheck disable=SC2154
+        short_commit="${git_commit::8}"
```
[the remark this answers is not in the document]
maru-ava: Portable for what? This works on our target platforms, we don't care about POSIX.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,121 @@
...
+        run: |
+          aws s3 cp "build/rpm/avalanchego-${TAG}-${{ matrix.rpm_arch }}.rpm" \
+            "s3://${BUCKET}/linux/rpms/${{ matrix.rpm_arch }}/"
+          aws s3 cp "build/rpm/subnet-evm-${TAG}-${{ matrix.rpm_arch }}.rpm" \
+            "s3://${BUCKET}/linux/rpms/${{ matrix.rpm_arch }}/"
+          aws s3 cp "build/rpm/RPM-GPG-KEY-avalanchego" \
+            "s3://${BUCKET}/linux/rpms/RPM-GPG-KEY-avalanchego"
+        env:
+          BUCKET: ${{ secrets.BUCKET }}
+        shell: bash
```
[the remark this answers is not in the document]
maru-ava: Done.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,121 @@
...
+        with:
+          name: rpms-${{ matrix.rpm_arch }}
+          path: |
+            build/rpm/avalanchego-${{ env.TAG }}-${{ matrix.rpm_arch }}.rpm
+            build/rpm/subnet-evm-${{ env.TAG }}-${{ matrix.rpm_arch }}.rpm
+            build/rpm/RPM-GPG-KEY-avalanchego
+
+      - name: Cleanup
+        run: |
+          rm -f "${GPG_KEY_FILE}"
```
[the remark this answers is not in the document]
maru-ava: Done

.github/packaging/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+#
+# Usage:
+#   docker build --build-arg GO_VERSION=1.24.12 -t avalanchego-rpm-builder .github/packaging/
+#   docker run --rm -v .:/build -v ./build/rpm:/output avalanchego-rpm-builder ...
+
+ARG GO_VERSION=INVALID
+
+FROM rockylinux:9
+
+ARG GO_VERSION
```
JuanLeon2: Subsumed (better) by line 10?

.github/packaging/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+ARG TARGETARCH
```
JuanLeon2, line 14: Ths is not specified in the Usage: above, and if left empty I don't think line 28 would succeed.

.github/packaging/nfpm/avalanchego.yml
```diff
@@ -0,0 +1,21 @@
+name: avalanchego
+arch: "${RPM_ARCH}"
+version: "${VERSION}"
+maintainer: "Ava Labs <security@avalabs.org>"
```
JonathanOppenheimer, line 4: Should we change the debian packages to not have Stephen's email. Or is that change out of scope? I think they should be consistent.

.github/packaging/scripts/extract-changelog.sh
```diff
@@ -0,0 +1,92 @@
...
+# Captures all bullet points (including indented sub-items), skips
+# bare labels like "Added:", "Deprecated:", strips markdown backticks,
+# and omits the "What's Changed" subsection (usually just a link).
+notes=$(echo "${section}" | awk '
+    /^### What'\''s Changed/ { skip = 1; next }
+    /^### / { skip = 0; next }
+    skip { next }
+    /^\s*- / {
+        line = $0
+        sub(/^\s*- /, "", line)
```
JonathanOppenheimer: This does not work (at least for me!), when running. I believe `\s` is the actual s character (escaped). 

You can test this with:

```bash
 printf '%s\n' '- top-level bullet' '  - indented bullet' \
  | awk '/^\s*- / {print "MATCH:", $0}'
```

vs. 

```bash
printf '%s\n' '- top-level bullet' '  - indented bullet' \
  | awk '/^[[:space:]]*- / {print "MATCH:", $0}'
```

Image

.github/packaging/scripts/build-rpm.sh
```diff
@@ -0,0 +1,185 @@
...
+
+# shellcheck disable=SC2154
+echo "Git commit: ${git_commit}"
+
+# Disable Go's automatic VCS stamping — the bind-mounted .git is owned by
+# the host user, causing git to fail inside the container. The commit hash
+# is passed explicitly via AVALANCHEGO_COMMIT and -ldflags instead.
+export GOFLAGS="${GOFLAGS:-} -buildvcs=false"
+
+case "${PACKAGE}" in
```
JonathanOppenheimer, line 49: Do we expect to build more .rpms in the future? As in other cases?

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,130 @@
...
+          aws s3 cp "build/rpm/avalanchego-${TAG}-${{ matrix.rpm_arch }}.rpm" \
+            "s3://${BUCKET}/linux/rpms/${{ matrix.rpm_arch }}/"
+          aws s3 cp "build/rpm/subnet-evm-${TAG}-${{ matrix.rpm_arch }}.rpm" \
+            "s3://${BUCKET}/linux/rpms/${{ matrix.rpm_arch }}/"
+        env:
+          BUCKET: ${{ secrets.BUCKET }}
+        shell: bash
+
+      - name: Upload GPG public key to S3
+        if: github.event_name != 'pull_request' && matrix.arch == 'amd64'
```
JonathanOppenheimer: Is this if statement to ensure the GPG key is only uploaded once, (`matrix.arch == amd64`)? 

As each build job runs in parallel ([according to the matrices for GitHub actions](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations)), it seems possible that the `arm64` job could 'get ahead' of the `amd64` job in execution, and then wouldn't have the GPG key available to it yet, and then things would break. According to the actions docs, the runner for the `amd64` _will_ start first, but I am reasonably confident that it's possible the second to start could still finish faster. 

Anyway, I think the key upload should be a separate job that happens before all of this. It also just seems conceptually weird to tie the single upload to one of the matrix entries. Make this job depend on a prior job that does the key upload?

.github/workflows/build-rpm-release.yml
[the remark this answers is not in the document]
JonathanOppenheimer: I think Copilot led you astray here. See my comment here: https://github.com/ava-labs/avalanchego/pull/4959/changes#r2800778620. There are better solutions to this race.

graft/subnet-evm/scripts/constants.sh
```diff
@@ -25,6 +25,7 @@ source "$AVALANCHE_PATH"/scripts/image_tag.sh
 GOPATH="$(go env GOPATH)"
 DEFAULT_PLUGIN_DIR="${HOME}/.avalanchego/plugins"
 DEFAULT_VM_NAME="subnet-evm"
+# NOTE: This value is sourced by RPM packaging scripts (.github/packaging/).
```
JonathanOppenheimer: Was there a shellcheck on this?

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,130 @@
...
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/setup-go-for-project
+
+      - name: Get tag
+        id: get_tag
```
JonathanOppenheimer: This ID never seems to be used.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,130 @@
...
+            echo "TAG=${GITHUB_REF/refs\/tags\//}" >> "$GITHUB_ENV"
+          else
+            # PR builds: synthetic tag for smoke testing the build pipeline
+            echo "TAG=v0.0.0-pr.${GITHUB_SHA::8}" >> "$GITHUB_ENV"
+          fi
+        shell: bash
+
+      - name: Import GPG key
+        if: github.event_name != 'pull_request'
+        id: gpg_key
```
JonathanOppenheimer: This ID never seems to be used

.github/packaging/scripts/extract-changelog_test.sh
```diff
@@ -0,0 +1,143 @@
...
+
+WORK_DIR=$(mktemp -d)
+trap 'rm -rf "${WORK_DIR}"' EXIT
+
+pass=0
+fail=0
+
+assert_contains() {
+    local file="$1" pattern="$2" msg="$3"
+    if grep -q "${pattern}" "${file}"; then
```
JonathanOppenheimer: (no action necessary) `grep -Eq` instead?

.github/packaging/scripts/extract-changelog_test.sh
```diff
@@ -0,0 +1,143 @@
...
+        echo "    expected pattern: ${pattern}"
+        echo "    file contents:"
+        sed 's/^/    /' "${file}"
+        fail=$((fail + 1))
+    fi
+}
+
+assert_not_contains() {
+    local file="$1" pattern="$2" msg="$3"
+    if ! grep -q "${pattern}" "${file}"; then
```
JonathanOppenheimer: (no action necessary) `grep -Eq` instead?

.github/packaging/scripts/extract-changelog.sh
```diff
@@ -0,0 +1,92 @@
...
+if [[ -z "${section}" ]]; then
+    fallback "version ${version} not found in ${releases_file}"
+    exit 0
+fi
+
+# Convert markdown section to nfpm changelog YAML.
+# Captures all bullet points (including indented sub-items), skips
+# bare labels like "Added:", "Deprecated:", strips markdown backticks,
+# and omits the "What's Changed" subsection (usually just a link).
+notes=$(echo "${section}" | awk '
```
JonathanOppenheimer: (no action necessary) `awk ... <<< "${section}"` could avoid the pipe. _MASSIVE_ performance boosts I'm sure by avoiding one additional process 😂. (shoutout claude simplifier agent for this comment)

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,130 @@
...
+          elif [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            echo "TAG=${GITHUB_REF/refs\/tags\//}" >> "$GITHUB_ENV"
+          else
+            # PR builds: synthetic tag for smoke testing the build pipeline
+            echo "TAG=v0.0.0-pr.${GITHUB_SHA::8}" >> "$GITHUB_ENV"
+          fi
+        shell: bash
+
+      - name: Import GPG key
+        if: github.event_name != 'pull_request'
```
JonathanOppenheimer, line 68: (no action necessary) There's 7 of these ` if: github.event_name != 'pull_request'` statements, which is sort of awkward in my opinion. Perhaps all the s3 logic could be moved into a follow up or separate job?

JonathanOppenheimer (commented):
The key upload technique to S3 is the only big potential issue I see with this PR.

.github/packaging/scripts/extract-changelog.sh
JonathanOppenheimer: This is probably irrelevant according to your other comment: https://github.com/ava-labs/avalanchego/pull/4959#discussion_r2796154420, as it's a similar comment. resolve if we don't care

.github/workflows/build-rpm-release.yml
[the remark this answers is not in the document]
JonathanOppenheimer: Un-resolving this

.github/packaging/nfpm/avalanchego.yml
maru-ava: I'm waiting on clarification on this, I don't know that consistency matters so much as having a sane value.

.github/packaging/scripts/build-rpm.sh
maru-ava: I'm not sure, and I'm also not sure what you're asking. Maybe clarify the question with more detail as to what you're wanting to know?

.github/packaging/nfpm/avalanchego.yml
maru-ava: Confirmed, we're using this email.

.github/workflows/build-rpm-release.yml
maru-ava: As per internal discussion, the s3 upload is to a private bucket and doesn't provide value over retention via github artifacts. Removed.

.github/workflows/build-rpm-release.yml
maru-ava: s3 removed.

.github/workflows/build-rpm-release.yml
maru-ava: Removed

.github/workflows/build-rpm-release.yml
maru-ava: Removed

graft/subnet-evm/scripts/constants.sh
maru-ava: Yes. Obsolete comment, removed.

.github/workflows/build-rpm-release.yml
JuanLeon2: If so change the .md line 7 that refers to S3.

.github/packaging/scripts/extract-changelog_test.sh
maru-ava: Removed this script in favor of linking to the github release page

.github/packaging/scripts/extract-changelog_test.sh
maru-ava: Removed this script in favor of linking to the github release page

.github/packaging/scripts/extract-changelog.sh
maru-ava: Removed this script in favor of linking to the github release page

.github/packaging/Dockerfile
maru-ava: Removed L10, not required when we're not using it in the FROM line like other Dockerfiles in this repo.

.github/packaging/Dockerfile
maru-ava: This is populated by docker automatically.

.github/workflows/build-rpm-release.yml
maru-ava: All references to s3 have been removed.

.github/packaging/scripts/extract-changelog.sh
maru-ava: Removed this script in favor of linking to the github release page

.github/workflows/build-rpm-release.yml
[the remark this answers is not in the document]
maru-ava: S3 upload removed.

.github/workflows/build-rpm-release.yml
maru-ava: These comments are pinned to the earlier state as per the 'Outdated' tag at the top of it. You'll need to look at the PRs changes to get the most recent version.

.github/packaging/Taskfile.yml
```diff
@@ -0,0 +1,112 @@
...
+  PACKAGING_GO_VERSION:
+    sh: go list -m -f '{{ "{{.GoVersion}}" }}' | head -1
+  # Map uname -m to RPM arch names (arm64 -> aarch64).
+  PACKAGING_HOST_ARCH:
+    sh: |
+      arch=$(uname -m)
+      case "${arch}" in
+        x86_64) echo "x86_64" ;;
+        arm64)  echo "aarch64" ;;
+        *)      echo "${arch}" ;;
```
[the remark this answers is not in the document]
maru-ava: No, this is intentional. When the arch doesn't need to be rewritten it falls through to *.

.github/packaging/scripts/validate-rpm.sh
```diff
@@ -0,0 +1,93 @@
...
+
+: "${TAG:?TAG must be set}"
+: "${GIT_COMMIT:?GIT_COMMIT must be set}"
+
+if [[ -z "${RPM_ARCH:-}" ]]; then
+    arch=$(uname -m)
+    case "${arch}" in
+        x86_64) RPM_ARCH="x86_64" ;;
+        arm64)  RPM_ARCH="aarch64" ;;
+        *)      RPM_ARCH="${arch}" ;;
```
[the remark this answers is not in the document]
maru-ava: No, this is intentional. When the arch doesn't need to be rewritten it falls through to *.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          GPG_KEY_FILE="$(mktemp)"
+          echo "${{ secrets.RPM_GPG_PRIVATE_KEY }}" > "${GPG_KEY_FILE}"
+          echo "GPG_KEY_FILE=${GPG_KEY_FILE}" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
maru-ava: Done

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,79 @@
...
+            build/rpm/subnet-evm-${{ env.TAG }}-${{ matrix.rpm_arch }}.rpm
+            build/rpm/RPM-GPG-KEY-avalanchego
+
+      - name: Cleanup
+        run: |
+          if [[ -n "${GPG_KEY_FILE:-}" ]]; then
+            rm -f "${GPG_KEY_FILE}"
+          fi
+          rm -rf build/rpm
+        shell: bash
```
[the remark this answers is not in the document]
maru-ava: Done

.github/packaging/scripts/build-rpm.sh
JonathanOppenheimer: Oh just like I thought it could be an if else. It doesn't really matter.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,81 @@
...
+            rpm_arch: x86_64
+          - arch: arm64
+            runner: custom-arm64-jammy
+            rpm_arch: aarch64
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
```
JonathanOppenheimer, line 33:
```suggestion
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.tag || github.ref }}
```

JonathanOppenheimer: (low importance, not sure if manual runs are even a concern). according to https://github.com/actions/checkout#usage, if `ref` is not set, checkout uses the dispatch event ref (`github.ref`), not `github.event.inputs.tag`, which is used below. If ran manually, this could build from one ref while labeling artifacts with another tag input, which would be misleading.

.github/workflows/build-rpm-release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - uses: ./.github/actions/setup-go-for-project
+
+      - name: Get tag
+        run: |
+          if [[ -n "${{ github.event.inputs.tag }}" ]]; then
+            echo "TAG=${{ github.event.inputs.tag }}" >> "$GITHUB_ENV"
```
JonathanOppenheimer, line 58: Here's `github.event.inputs.tag`

maru-ava:
I need to enable manual triggering against a legacy tag before merge.

Edit: Done, it should now be possible to build rpms for 1.14.1 once this merges.

PR #4960, opened 2026-02-11
chore: Increase firewood defaults

rkuris:
## Why this should be merged

The defaults were fine for development, but we need these larger for production.

Free list changed from 80k to 1000k

## How this was tested

We should see some performance improvements in production

No


alarso16 (changes requested):
We can change the FreeListCacheEntries field, but everything here is only used in testing. You should actually update `core/blockchain.go` with new FreeListCacheEntries. For now, I think it would be better not to change the default values, as that's a breaking change and wouldn't work well with our testing. However, the remaining values you altered can be set via config, so in any deployment used, we can specify a value

alarso16:
Use #4963 instead

PR #4962, opened 2026-02-11
chore: upgrade lint to 2.9.0

JonathanOppenheimer:
## Why this should be merged

With the update to go 1.25.7, we should also our now quite out of date linter. This PR updates the linter, fixes all new lint errors for pre-existing linters, and adds four new commented out linters (after consultation with @joshua-kim and @alarso16)

- `embeddedstructfieldcheck`
- `godoc-lint`
- `iotamixing`
- `modernize`

These will be added one by one in subsequent PRs. 

## How this was tested
CI 

No


tests/antithesis/init_db.go
```diff
@@ -49,9 +48,7 @@ func initBootstrapDB(network *tmpnet.Network, destPath string) error {
 	if err := os.MkdirAll(destPath, perms.ReadWriteExecute); err != nil {
 		return fmt.Errorf("failed to create db path %q: %w", destPath, err)
 	}
-	// TODO(marun) Replace with os.CopyFS once we upgrade to Go 1.23
-	cmd := exec.Command("cp", "-r", sourcePath, destPath)
-	if err := cmd.Run(); err != nil {
+	if err := os.CopyFS(destPath, os.DirFS(sourcePath)); err != nil {
 		return fmt.Errorf("failed to copy bootstrap db from %q to %q: %w", sourcePath, destPath, err)
```
JonathanOppenheimer, line 55: Rather than fix the new lint error, we can just implement the todo now.

node/node.go
```diff
@@ -437,7 +437,7 @@ func (n *Node) initNetworking(reg prometheus.Registerer) error {
 	// 1: https://apple.stackexchange.com/questions/393715/do-you-want-the-application-main-to-accept-incoming-network-connections-pop
 	// 2: https://github.com/golang/go/issues/56998
 	listenAddress := net.JoinHostPort(n.Config.ListenHost, strconv.FormatUint(uint64(n.Config.ListenPort), 10))
-	listener, err := net.Listen(constants.NetworkType, listenAddress)
```
joshua-kim, line 440: `context.Background()` is the root context -- generally this is defined in main and is passed around/children are made off of it. We should always use a derived context from `context.Background`, or `context.TODO` when it is unclear which context we should be using. Using the wrong context will cause cancellation signals to not be propagated correctly.

.golangci.yml
```diff
@@ -56,19 +56,23 @@ linters:
     - copyloopvar
     - depguard
     - durationcheck
+    # - embeddedstructfieldcheck
```
joshua-kim, line 59: What's the reason behind commenting out these linters?

node/node.go
joshua-kim: This PR is stale, but you can refer to it for some additional context: https://github.com/ava-labs/avalanchego/pull/4163

.golangci.yml
JonathanOppenheimer: Trying to add them all in this PR would make this PR massive and hard to review. I wanted to gain consensus on what linters we planned to implement by adding them commented out, and then implement them one by one in following PRs. This is the procedure we followed when adding the AvalancheGo linters to the EVM repositories.

node/node.go
JonathanOppenheimer: Ahh. Would you like me to bring in changes from that PR here, or would you like to make that PR 'active' (I can work on it if desired as well if you want to pass it off)

JonathanOppenheimer: I've separated out the `noctx` linter, so it can be addressed separately.

graft/evm/scripts/lint_setup.sh
```diff
@@ -105,7 +105,7 @@ function setup_lint {
   cp ../../.golangci.yml "$AVALANCHE_LINT_FILE"
 
   # Exclude all upstream files dynamically
-  echo "    paths-except:" >> "$AVALANCHE_LINT_FILE"
+  echo "    paths:" >> "$AVALANCHE_LINT_FILE"
```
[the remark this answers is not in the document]
JonathanOppenheimer: No copilot, this is the correct arguement. 

see `exclusions: fix path-expectIn the change log for v2.2.0`

.golangci.yml
JonathanOppenheimer: resolving this feel free to re-open if you still have concerns.

PR #4964, opened 2026-02-12
Benchlist response handling

aaronbuchwald:
## Why this should be merged

This PR should be merged to observe the success/failed query data from peers even when they are benched.

## How this works

When a request is sent to a benched peer, the previous behavior was to immediately send the internal failed message to the chain router and mark the request as failed to the benchlist.

This change continues to immediately send the internal failed message to the chain router to prefer failing fast, but modifies how it interacts with the benchlist. Rather than immediately observing a failed request to the benched peer, it sends the request anyways and uses the response or timeout to notify the benchlist of the actual observed behavior. This allows the benchlist to gather the success/failed query data from a peer even when it's benched.

This change is split out from https://github.com/ava-labs/avalanchego/tree/stephen-dev-node in addition to fixing the tests.

## How this was tested

Existing tests.

### DIFF
--- a/snow/networking/router/chain_router_test.go
+++ b/snow/networking/router/chain_router_test.go
@@ -1436,66 +1436,6 @@ func TestAppRequest(t *testing.T) {
 	}
 }
 
-func TestAppRequestExternalFailureClearsTimeout(t *testing.T) {
-	require := require.New(t)
-
-	const requestID uint32 = 1
-	appErr := &common.AppError{
-		Code:    1234,
-		Message: "custom error",
-	}
-
-	chainRouter, engine := newChainRouterTest(t)
-
-	var (
-		lock      sync.Mutex
-		callCount int
-		called    = make(chan struct{})
-	)
-	engine.AppRequestFailedF = func(_ context.Context, nodeID ids.NodeID, gotRequestID uint32, gotAppErr *common.AppError) error {
-		lock.Lock()
-		defer lock.Unlock()
-
-		callCount++
-		require.Equal(ids.EmptyNodeID, nodeID)
-		require.Equal(requestID, gotRequestID)
-		require.Equal(appErr.Code, gotAppErr.Code)
-		require.Equal(appErr.Message, gotAppErr.Message)
-		if callCount == 1 {
-			close(called)
-		}
-		return nil
-	}
-
-	ctx := t.Context()
-	chainRouter.RegisterRequest(
-		ctx,
-		ids.EmptyNodeID,
-		ids.Empty,
-		requestID,
-		message.AppResponseOp,
-		message.InboundAppError(ids.EmptyNodeID, ids.Empty, requestID, common.ErrTimeout.Code, common.ErrTimeout.Message),
-		engineType,
-	)
-
-	chainRouter.HandleInbound(
-		ctx,
-		message.InboundAppError(ids.EmptyNodeID, ids.Empty, requestID, appErr.Code, appErr.Message),
-	)
-
-	<-called
-
-	chainRouter.lock.Lock()
-	require.Zero(chainRouter.timedRequests.Len())
-	chainRouter.lock.Unlock()
-
-	time.Sleep(2 * defaultTestTimeout)
-
-	lock.Lock()
-	defer lock.Unlock()
-	require.Equal(1, callCount)
-}
-
 func TestBenchedPeerEarlyFailureThenTimeoutOrResponse(t *testing.T) {
 	var (
 		nodeID  = ids.GenerateTestNodeID()


snow/networking/router/chain_router.go
```diff
@@ -313,17 +322,35 @@ func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMe
...
+			msg.OnFinishedHandling()
+			return
+		}
+
+		// If the failure message comes from a timeout, clear it. Otherwise, this indicates an early,
+		// internal failure and we should wait for either a successful response or a timeout to clear
+		// the request.
+		if timeout || !internal {
+			cr.timedRequests.Delete(uniqueRequestID)
+			cr.metrics.outstandingRequests.Set(float64(cr.timedRequests.Len()))
```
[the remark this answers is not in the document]
aaronbuchwald: Peers never send `*Failed` messages, they are only sent internally. @copilot can you re-review with this understanding?

snow/networking/router/chain_router.go
```diff
@@ -198,28 +205,30 @@ func (cr *ChainRouter) RegisterRequest(
...
-	go cr.handleMessage(ctx, msg, true)
+	go cr.handleMessage(ctx, msg, true, false)
 }
 
 // handleMessage routes a message to the specified chain. Messages may be
 // unrequested, responses, or timeouts. The internal flag indicates whether the
 // message is being sent from an internal component, such as due to a timeout,
 // or if the message originated from a remote peer.
-func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMessage, internal bool) {
+func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMessage, internal, timeout bool) {
```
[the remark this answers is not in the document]
aaronbuchwald: I agree this is non-trivial. I'm migrating the code from another branch, so keeping this unless it's preferred to go with an enum. I don't think an enum actually simplifies things much. The complexity is reasoning across multiple packages that interact with each other very non-obviously.

snow/networking/router/chain_router.go
```diff
@@ -313,17 +322,35 @@ func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMe
...
+		if !exists {
+			// This was a duplicated message
+			msg.OnFinishedHandling()
+			return
+		}
+
+		// If the failure message comes from a timeout, clear it. Otherwise, this indicates an early,
+		// internal failure and we should wait for either a successful response or a timeout to clear
+		// the request.
+		if timeout || !internal {
```
yacovm: How can we time out on an internal message? We are always reachable inside our process, aren't we?

snow/networking/router/chain_router.go
```diff
@@ -313,17 +322,38 @@ func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMe
...
+		}
+
+		// Clear the request if it has definitively failed due to:
+		// 1. Timeout
+		// 2. Peer sent message failure (app-error is the only peer sent, non-internal message type indicating failure)
+		//
+		// For an early internal failure message due to benching, disconnect, or a request to self, we do not clear
+		// the timeout and instead wait for the actual response or timeout.
+		if timeout || !internal {
+			cr.timedRequests.Delete(uniqueRequestID)
```
yacovm, line 355: can't we use `clearRequest` here?

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,362 @@ func TestAppRequest(t *testing.T) {
...
+}
+
 func newChainRouterTest(t *testing.T) (*ChainRouter, *enginetest.Engine) {
 	// Create a timeout manager
 	tm, err := timeout.NewManager(
 		&timer.AdaptiveTimeoutConfig{
-			InitialTimeout:     3 * time.Second,
-			MinimumTimeout:     3 * time.Second,
-			MaximumTimeout:     5 * time.Minute,
+			InitialTimeout:     500 * time.Millisecond,
```
yacovm: why this change?

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,362 @@ func TestAppRequest(t *testing.T) {
...
+				engine.AppResponseF = func(context.Context, ids.NodeID, uint32, []byte) error {
+					failedMessages <- struct{}{}
+					return nil
+				}
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
yacovm: it's a nit so i don't really mind, but - do we really need to do these things now? I thought Golang fixed its closure behavior?

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,362 @@ func TestAppRequest(t *testing.T) {
...
+	type testCase struct {
+		name               string
+		responseOp         message.Op
+		timeoutMsg         func(ids.NodeID, ids.ID, uint32, p2ppb.EngineType) *message.InboundMessage
+		responseMsg        func(ids.NodeID, ids.ID, uint32) *message.InboundMessage
+		setFailureHandler  func(*testing.T, *enginetest.Engine, ids.NodeID, uint32, chan<- struct{}, *sync.WaitGroup)
+		setResponseHandler func(*testing.T, *enginetest.Engine, chan<- struct{})
+	}
+
+	tests := []testCase{
```
yacovm: why do we need a test of each kind of message type if the handling code in the router is independent of message type?

aaronbuchwald: I was just following a pattern in other tests although that may have been in cases where there were distinct code paths. The fact they follow the same code path is an internal detail, so there's an argument to ensure coverage of each message type.

If preferred, happy to collapse this into one instead of covering each message type.

snow/networking/router/chain_router_test.go
aaronbuchwald: You're right, I can remove it since this is no longer an issue.

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,362 @@ func TestAppRequest(t *testing.T) {
...
+			chainRouter.lock.Lock()
+			require.Equal(1, chainRouter.timedRequests.Len())
+			chainRouter.lock.Unlock()
+
+			// Wait for the timeout manager to fire and clear the request.
+			// Note: relying on the timeout manager to fire results in longer running tests.
+			// The alternative not chosen here is to send a mock timeout messag early rather than
+			// relying on the production code, which tests both the timeout manager and its integration
+			// with the chain router.
+			require.Eventually(
```
aaronbuchwald: This is the reason to reduce the when constructing the test chain router. I opted to test the actual code instead of mocking a timeout message, so each instance of this test waits for the timeout manager's timeout to fire.

Reducing the timeout doesn't impact any of the other tests, so if we want to keep this behavior (testing the integration with the actual timeout manager instead of mocking) it's preferred to reduce the timeout values.

snow/networking/router/chain_router_test.go
aaronbuchwald: ref https://github.com/ava-labs/avalanchego/pull/4964/changes#r2805909564

snow/networking/router/chain_router.go
aaronbuchwald: We could, we'd just have to use the fields on the requestID we've already created.

With this change, we no longer just want to clear the request, we want to fetch it, check some conditional logic, and then decide whether or not to clear it here, which is why it's no longer a one to one match.

Keeping this as is to follow Stephen's style from the original branch.

snow/networking/router/chain_router.go
aaronbuchwald: One case is `SendGetAncestors` allows the timeout to fire rather than send an internal early failure so that if it's not connected to anyone. Further, all messages are queued and processed one by one, so if we are blocked processing one message for >2s, then we could see a timeout fire on an internal message. Most likely, we'd still process the internal message before the timeout.

yacovm: yeah makes sense, haven't considered that we might be blocked for too long because an earlier message is processing 👍  makes sense

aaronbuchwald: The reason for `!internal` is to cover the case of receiving `AppError` which is included in the set of failed messages even though it is not internal. This is the only failure message that comes from a peer and should clear the request since there's no longer a need for either the timeout to fire or to wait for a response from a peer (`AppError` is the message)

yacovm:
Marking approved but I think we should do some testing before merging this

aaronbuchwald:
> Marking approved but I think we should do some testing before merging this

I'll start a dev node. Is there anything else you're specifically suggesting?

snow/networking/router/chain_router.go
```diff
@@ -313,17 +322,38 @@ func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMe
...
+			cr.metrics.outstandingRequests.Set(float64(cr.timedRequests.Len()))
+		}
```
StephenButtolph: Are we clearing the timeout here when someone sends us an `AppError`? Do we need to call `cr.timeoutManager.RemoveRequest(uniqueRequestID)` in this block?

I don't think it actually modifies the code behavior, because handling of another `AppError` will hit the above `!exists` check and then return... But I feel like we should normally avoid sending a duplicate message.

snow/networking/router/chain_router.go
```diff
@@ -313,17 +322,38 @@ func (cr *ChainRouter) handleMessage(ctx context.Context, msg *message.InboundMe
...
+
+		if req.handled {
+			// This was a duplicated message
 			msg.OnFinishedHandling()
 			return
 		}
 
-		// Tell the timeout manager we are no longer expecting a response
-		cr.timeoutManager.RemoveRequest(uniqueRequestID)
```
StephenButtolph, line 326: This was the only usage of RemoveRequest, should we delete this function in its entirety?

Looking at it, we may want to add this usage back in for the case we are handling an external `AppError`.

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,359 @@ func TestAppRequest(t *testing.T) {
...
 }
 
+func TestBenchedPeerEarlyFailureThenTimeoutOrResponse(t *testing.T) {
+	type testCase struct {
+		name               string
+		responseOp         message.Op
+		timeoutMsg         func(ids.NodeID, ids.ID, uint32, p2ppb.EngineType) *message.InboundMessage
+		responseMsg        func(ids.NodeID, ids.ID, uint32) *message.InboundMessage
+		setFailureHandler  func(*testing.T, *enginetest.Engine, ids.NodeID, uint32, chan<- struct{}, *sync.WaitGroup)
+		setResponseHandler func(*testing.T, *enginetest.Engine, chan<- struct{})
```
StephenButtolph: Why does `setResponseHandler` take in a `*testing.T`? Are these functions supposed to assert the provided `nodeID` and `requestID` like `setFailureHandler` does? Currently, `setResponseHandler` never asserts anything.

snow/networking/router/chain_router_test.go
```diff
@@ -1433,13 +1435,359 @@ func TestAppRequest(t *testing.T) {
...
+					calls <- struct{}{}
+					return nil
+				}
+			},
+			setResponseHandler: func(_ *testing.T, engine *enginetest.Engine, failedMessages chan<- struct{}) {
+				engine.StateSummaryFrontierF = func(context.Context, ids.NodeID, uint32, []byte) error {
+					failedMessages <- struct{}{}
+					return nil
+				}
+			},
```
StephenButtolph: Why is the channel called `failedMessages` here? isn't `setResponseHandler` the successful path? (Same comment for the rest of the tests).

snow/networking/router/chain_router.go
aaronbuchwald: Agreed. External `AppError` should also clear the timeout-manager entry. I restored `RemoveRequest` for that path and added a regression test to verify we do not later deliver a duplicate timeout failure.

snow/networking/router/chain_router.go
aaronbuchwald: Keeping `RemoveRequest`; it is now used again for externally-originated `AppError` handling in the router failure path.

PR #4966, opened 2026-02-12
style: enable `embeddedstructfieldcheck` linter

JonathanOppenheimer:
## Why this should be merged

This keeps embedded struct field layout consistent across the codebase and makes serialization-sensitive structs easier to review, while avoiding accidental reliance on implicit field order.

Enable `embeddedstructfieldcheck` linter, which enforces two rules:
- Embedded fields should be listed before regular fields
- A blank line must separate embedded fields from regular fields

I used `golangci-lint run --fix` for the blank line separation between embedded and regular fields, and then manually moved embedded fields above regular fields in non-serialized structs. I no-linted codec-serialized structs where field order is part of the wire/storage format and reordering would be breaking.

## How this works
There are no intended semantic changes in this PR. Some JSON byte output may reorder object members, but this should not be significant.

## How this was tested
CI, including golangci-lint with `embeddedstructfieldcheck` enabled.

No


config/node/config.go
```diff
@@ -72,13 +73,14 @@ type IPConfig struct {
 
 type StakingConfig struct {
 	genesis.StakingConfig
+	StakingSignerConfig `json:"stakingSingerConfig"`
```
alarso16, line 76: q: does this change the json encoding?

graft/subnet-evm/params/extras/config.go
```diff
@@ -107,10 +107,11 @@ type ChainConfig struct {
 
 	AvalancheContext `json:"-"` // Avalanche specific context set during VM initialization. Not serialized.
 
+	UpgradeConfig `json:"-"` // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Not serialized.
```
alarso16, line 110: Same here

network/config.go
```diff
@@ -105,7 +105,8 @@ type Config struct {
 	PeerListGossipConfig `json:"peerListGossipConfig"`
 	TimeoutConfig        `json:"timeoutConfigs"`
 	DelayConfig          `json:"delayConfig"`
-	ThrottlerConfig      ThrottlerConfig `json:"throttlerConfig"`
+
+	ThrottlerConfig ThrottlerConfig `json:"throttlerConfig"`
```
alarso16, line 109: Is there any functional difference between this being embedded or not? Does it change encoding or something/

vms/platformvm/api/validator.go
```diff
@@ -47,10 +48,11 @@ type UTXO struct {
 
 // APIL1Validator is the representation of a L1 validator sent over APIs.
 type APIL1Validator struct {
+	BaseL1Validator
```
alarso16, line 51: Does encoding order matter for this?

config/node/config.go
JonathanOppenheimer: Yes but I think per the JSON spec this does not matter: 

Per [RFC 8259 Section 4](https://www.rfc-editor.org/rfc/rfc8259#section-4), JSON object member ordering doesn't matter: 
> "Implementations whose behavior does not depend on member ordering will be interoperable."

As you alluded to though, the JSON bytes will change as the keys appear in a different order in the serialized output.

graft/subnet-evm/params/extras/config.go
JonathanOppenheimer: No this has no impact because it's not serialized. It's not in the JSON.

network/config.go
JonathanOppenheimer: This is just a whitespace change, but yes it matters. 

From the [docs](https://pkg.go.dev/encoding/json#Marshal    ): 

> Embedded struct fields are usually marshaled as if their inner exported fields were fields in the outer struct, subject to the usual Go visibility rules... An anonymous struct field with a name given in its JSON tag is treated as having that name, rather  than being anonymous.

vms/platformvm/api/validator.go
JonathanOppenheimer: See my prior [comment](https://github.com/ava-labs/avalanchego/pull/4966#discussion_r3080878895). This really should not matter.  however, If any downstream consumers are doing byte-level comparison of JSON responses, this would be a visible change. But they really should not be doing that...

vms/nftfx/transfer_output.go
```diff
@@ -29,7 +29,7 @@ type TransferOutput struct {
 
 	GroupID                  uint32              `serialize:"true" json:"groupID"`
 	Payload                  types.JSONByteSlice `serialize:"true" json:"payload"`
-	secp256k1fx.OutputOwners `serialize:"true"`
```
joshua-kim, line 32: We should document why we're ignoring the linter here because it will not be obvious to future mantainers without context of how serialization works.

vms/saevm/sae/recovery_test.go
```diff
@@ -111,12 +111,13 @@ func TestRecoverFromDatabase(t *testing.T) {
 				})
 
 				for _, sys := range []struct {
+					*SUT
+
 					name string
 					ctx  context.Context //nolint:containedctx // Ephemeral so not in contravention of https://go.dev/blog/context-and-structs
-					*SUT
 				}{
-					{"source", srcCtx, src},
```
joshua-kim, line 118: I'm surprised we don't already have a linter for keys specified in structs -- it's fine in this test but it can lead to bugs when doing refactors across fields that share a type

JonathanOppenheimer: Can you expand upon that? What kind of bugs could be introduced (I'm reading this comment as you being against keys specified in structs?)

joshua-kim: If you initialize a struct without its keys specified like `foo{1, 2}` you can break if `foo` is refactored from:
```
type foo struct {
    a int
    b int
}
```

to
```
type foo struct {
    b int
    a int
}
```

because the previous declaration will not break at compile time even though the meaning of `foo{1, 2}` is now different.

JonathanOppenheimer: Ah yeah sorry I misunderstood your message then -- that's the thing here. Should we have a linter that requires that the keys are specified? That would be a very easy PR.

joshua-kim: Yeah, the linter we're looking for is `govet` ([ref](https://golangci-lint.run/docs/linters/configuration/#govet))

JonathanOppenheimer: I don't think govet's `composites` works for this -- it doesn't flag this and is actually already enabled by default. I think `exhaustruct` might be what we're looking for?

JonathanOppenheimer: Actually `exhuaststruct` doesn't work because it also forces all struct members to be named, which we obviously don't want and there's no level of configuration.

joshua-kim: I'm going to resolve this comment because this is low priority

ARR4N:
> ## Why this should be merged
> Enable `embeddedstructfieldcheck `linter, which enforces two rules:

This is what, not why. "Why" is the benefit of the linter, which is...?

JonathanOppenheimer:
> > ## Why this should be merged
> 
> > Enable `embeddedstructfieldcheck `linter, which enforces two rules:
> 
> 
> 
> This is what, not why. "Why" is the benefit of the linter, which is...?

Primarily consistency, especially when creating new structs where serialization matters - the same rationale for import ordering. (It also prevents the class of bugs shown in this PR where the ordering is implicit and the keys aren't named. If the order changes, unexpected things can happen!) I can update the PR description.


Don't bother reviewing this right now though - it's fairly low priority and something I opened a few months ago when we were updating Go. I'll leave it open if we ever wish to return to it.

PR #4971, opened 2026-02-13
ACP-236 E2E tests

rrazvan1:

### DIFF
--- a/tests/e2e/p/acp236/continuous_validator.go
+++ b/tests/e2e/p/acp236/continuous_validator.go
@@ -1,0 +1,357 @@
+// Copyright (C) 2019, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package acp236
+
+import (
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/avalanchego/api/admin"
+	"github.com/ava-labs/avalanchego/api/info"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/tests/e2e/p"
+	"github.com/ava-labs/avalanchego/tests/fixture/e2e"
+	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/crypto/secp256k1"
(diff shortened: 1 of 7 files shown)


tests/e2e/p/acp236/continuous_validator.go
```diff
@@ -0,0 +1,357 @@
...
+
+		tc.By("adding a delegator for half the staking period", func() {
+			delegatorKeychain := secp256k1fx.NewKeychain(delegatorFundingKey)
+			delegatorPWallet := e2e.NewWallet(tc, delegatorKeychain, walletNodeURI).P()
+
+			delegationEndTime := time.Now().Add(delegationPeriod)
+
+			_, err := delegatorPWallet.IssueAddPermissionlessDelegatorTx(
+				&txs.SubnetValidator{
+					Validator: txs.Validator{
```
yacovm, line 153: what about start time? why don't we need it?

tests/e2e/p/acp236/continuous_validator.go
```diff
@@ -0,0 +1,357 @@
...
+			require.NoError(err)
+		})
+
+		var actualDelegationPeriod time.Duration
+		tc.By("verifying the validator is in the current validator set and retrieving delegation period", func() {
+			tc.Eventually(func() bool {
+				validators, err := pvmClient.GetCurrentValidators(tc.DefaultContext(), constants.PrimaryNetworkID, []ids.NodeID{nodeID})
+				require.NoError(err)
+
+				return len(validators) == 1 && len(validators[0].Delegators) == 1
```
yacovm, line 176: shouldn't we also compare the nodeID to the expected nodeID?

tests/e2e/p/acp236/continuous_validator.go
```diff
@@ -0,0 +1,357 @@
...
+
+		tc.By("retrieving supply for third cycle")
+		supplyAtCycle3Start, _, err := pvmClient.GetCurrentSupply(tc.DefaultContext(), constants.PrimaryNetworkID)
+		require.NoError(err)
+
+		tc.By("retrieving wallet balance before exiting")
+		fundedKeyBalancesBeforeExit, err := pWallet.Builder().GetBalance()
+		require.NoError(err)
+
+		tc.By("waiting for the third staking cycle to complete", func() {
```
yacovm, line 315: I think we should also check the elapsed time to make sure it takes the time we expect it to take

rrazvan1:
in favor of #5106

PR #4978, opened 2026-02-13
ci: change PR fuzz tests timing and bump fuzz test timeout 

JonathanOppenheimer:
## Why this should be merged

GitHub seems to have reduced the compute available for these jobs, and we observed "FAIL, ... context deadline exceeded" repeatedly. This PR changes the test-fuzz task to a 1 second fuzz time to reduce the time spent in fuzz tests, and the timeout is increased. 

## How this works

To increase the timeout for the go tests (beyond the 10m default), I added an `explicit -timeout` flag to go test in `build_fuzz.sh`. 

## How this was tested
CI

No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,14 @@ jobs:
         run: ./scripts/run_task.sh test-unit
         env:
           TIMEOUT: ${{ env.TIMEOUT }}
+  Fuzz:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-go-for-project
+      - name: test-fuzz
+        shell: bash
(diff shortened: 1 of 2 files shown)


yacovm:
> Changes the test-fuzz task to a 1 second fuzz time as it flaking CI repeatedly.

Shouldn't we just silence / skip the flaky fuzz test instead of just changing the fuzz time for all fuzz tests?

JonathanOppenheimer:
> > Changes the test-fuzz task to a 1 second fuzz time as it flaking CI repeatedly.
> 
> Shouldn't we just silence / skip the flaky fuzz test instead of just changing the fuzz time for all fuzz tests?

This doesn't change the fuzz time for all the tests, just the singular `test-fuzz`

yacovm (approved):

PR #4980, opened 2026-02-16
chore: upgrade toolchain to Go 1.26

Olexandr88:
Bump minimum Go version to 1.26 across all modules and workspace to align with upcoming release requirements


JonathanOppenheimer:
We purposely do not use the latest major version of Go, but stay one behind.

PR #4981, opened 2026-02-16
[tmpnet] Annotate pods for collection to grafana cloud

maru-ava:
## Why this should be merged

For nodes deployed to kube by tmpnet, a local kind cluster will continue to be collected by the same prometheus and promtail agents, and for managed infrastructure, collection will be to grafana cloud. The latter case requires additional annotations to trigger collection.

## How this was tested

CI for regression, validation of collection will be done post-merge.

N/A

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,10 @@
 
 ## Pending (v1.14.2)
 
+### Breaking Changes
+
+- Removed `graft/subnet-evm/compatibility.json` in favor of using a single source of truth in `version/compatibility.json`. External tools that relied on this file should now use `version/compatibility.json` from the main AvalancheGo repository.
+
 ### Config
 
 - Removed `pull-gossip-poll-size` from the X-chain and P-chain configs.
--- a/RELEASING_README.md
+++ b/RELEASING_README.md
@@ -1,453 +1,404 @@
-# THIS README IS WHOLLY OUT OF DATE AND WILL BE UPDATED! 
+# AvalancheGo Release Guide
 
-# Releasing
+This document covers the complete release process for AvalancheGo and its integrated components (Coreth and Subnet-EVM).
 
-## When to release
+## Overview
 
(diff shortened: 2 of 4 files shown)


tests/fixture/tmpnet/kube.go
```diff
@@ -63,10 +63,17 @@ func NewNodeStatefulSet(
...
 
+	// The following sets of annotations can coexist safely since each
+	// collection agent only reads the annotations it recognizes.
 	podAnnotations := map[string]string{
+		// Configure collection by prometheus and promtail
 		"prometheus.io/scrape": "true",
 		"prometheus.io/path":   "/ext/metrics",
 		"promtail/collect":     "true",
+		// Configure collection by the grafana cloud agent
+		"k8s.grafana.com/metrics_path": "/ext/metrics",
```
yacovm, line 74: I don't understand how this works. Why does specifying the endpoint of Grafana help here? Grafana cloud is the one supposed to scrape the managed infra nodes, so how does specifying its own endpoint help?

tests/fixture/tmpnet/kube.go
```diff
@@ -63,10 +63,17 @@ func NewNodeStatefulSet(
...
+		"k8s.grafana.com/port":         "9650",
```
Elvis339, line 75: The Grafana agent looks for `__meta_kubernetes_pod_annotation_k8s_grafana_com_metrics_port`, so the annotation must include metrics in the key. (if you have verified this just close the comment)

`k8s.grafana.com/port -> k8s.grafana.com/metrics.port`

tests/fixture/tmpnet/kube.go
maru-ava: These annotations are required to configure collection for a given pod by the in-cluster grafana cloud agent.

tests/fixture/tmpnet/kube.go
maru-ava: @stylianosrigas Is this correct?

maru-ava: `/metrics` is correct

stylianosrigas: Yeap it is correct because this is how we configure it centrally.

stylianosrigas (approved):

PR #4982, opened 2026-02-17
Add back EWMA benching logic

aaronbuchwald:
## Why this should be merged

Switches from failure streak + minimum time failing + timeout based benching to EWMA based benching with a configurable timeout.

## How this works

This PR utilizes the success/failure data from consensus requests to track an EWMA for the failure probability of sending a request to a node. When the failure probability exceeds a certain threshold, it is added to the benchlist. When that failure probability falls below a certain threshold, it gets removed. In addition, benched validators are added with a configurable timeout so that they stay on the benchlist indefinitely. When the timeout fires, they are removed from the benchlist AND their EWMA is reset. This timeout must be set to a sufficiently low value during e2e testing so that a node coming online unbenches its peers in a reasonable amount of time for the e2e tests.

## How this was tested

CI.

Yes.

### DIFF
--- a/vms/platformvm/block/builder/builder.go
+++ b/vms/platformvm/block/builder/builder.go
@@ -363,7 +363,10 @@ func packDurangoBlockTxs(
 	pChainHeight uint64,
 	remainingSize int,
 ) ([]*txs.Tx, error) {
-	stateDiff, err := state.NewDiffOn(parentState)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		backend.Config.UpgradeConfig.IsHeliconActivated(timestamp),
+	)
+	stateDiff, err := state.NewDiffOn(parentState, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return nil, err
 	}
@@ -424,7 +427,10 @@ func packEtnaBlockTxs(
 	pChainHeight uint64,
 	minCapacity gas.Gas,
 ) ([]*txs.Tx, error) {
-	stateDiff, err := state.NewDiffOn(parentState)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		backend.Config.UpgradeConfig.IsHeliconActivated(timestamp),
+	)
+	stateDiff, err := state.NewDiffOn(parentState, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return nil, err
 	}
@@ -547,7 +553,10 @@ func executeTx(
 		return false, nil
 	}
 
-	txDiff, err := state.NewDiffOn(stateDiff)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		backend.Config.UpgradeConfig.IsHeliconActivated(stateDiff.GetTimestamp()),
+	)
+	txDiff, err := state.NewDiffOn(stateDiff, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return false, err
 	}
--- a/vms/platformvm/block/builder/helpers_test.go
+++ b/vms/platformvm/block/builder/helpers_test.go
@@ -252,7 +252,7 @@ func addSubnet(t *testing.T, env *environment) {
 	require.NoError(err)
 
 	genesisID := env.state.GetLastAccepted()
-	stateDiff, err := state.NewDiff(genesisID, env.blkManager)
+	stateDiff, err := state.NewDiff(genesisID, env.blkManager, state.StakerAdditionAfterDeletionForbidden)
 	require.NoError(err)
 
 	feeCalculator := state.PickFeeCalculator(env.config, stateDiff)
--- a/vms/platformvm/block/executor/helpers_test.go
+++ b/vms/platformvm/block/executor/helpers_test.go
@@ -252,7 +252,7 @@ func addSubnet(t testing.TB, env *environment) {
 	require.NoError(err)
 
 	genesisID := env.state.GetLastAccepted()
-	stateDiff, err := state.NewDiff(genesisID, env.blkManager)
+	stateDiff, err := state.NewDiff(genesisID, env.blkManager, state.StakerAdditionAfterDeletionForbidden)
 	require.NoError(err)
 
 	feeCalculator := state.PickFeeCalculator(env.config, stateDiff)
--- a/vms/platformvm/block/executor/manager.go
+++ b/vms/platformvm/block/executor/manager.go
@@ -159,7 +159,10 @@ func (m *manager) VerifyTx(tx *txs.Tx) error {
 		return fmt.Errorf("failed verifying warp messages: %w", err)
 	}
 
-	stateDiff, err := state.NewDiff(m.preferred, m)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		m.txExecutorBackend.Config.UpgradeConfig.IsHeliconActivated(m.txExecutorBackend.Clk.Time()),
+	)
+	stateDiff, err := state.NewDiff(m.preferred, m, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return fmt.Errorf("failed creating state diff: %w", err)
 	}
--- a/vms/platformvm/block/executor/verifier.go
+++ b/vms/platformvm/block/executor/verifier.go
@@ -62,7 +62,10 @@ func (v *verifier) BanffProposalBlock(b *block.BanffProposalBlock) error {
 	}
 
 	parentID := b.Parent()
-	onDecisionState, err := state.NewDiff(parentID, v.backend)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		v.txExecutorBackend.Config.UpgradeConfig.IsHeliconActivated(b.Timestamp()),
+	)
+	onDecisionState, err := state.NewDiff(parentID, v.backend, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return err
 	}
@@ -84,12 +87,12 @@ func (v *verifier) BanffProposalBlock(b *block.BanffProposalBlock) error {
 		return err
 	}
 
-	onCommitState, err := state.NewDiffOn(onDecisionState)
+	onCommitState, err := state.NewDiffOn(onDecisionState, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return err
 	}
 
-	onAbortState, err := state.NewDiffOn(onDecisionState)
+	onAbortState, err := state.NewDiffOn(onDecisionState, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return err
 	}
@@ -114,7 +117,10 @@ func (v *verifier) BanffStandardBlock(b *block.BanffStandardBlock) error {
 	}
 
 	parentID := b.Parent()
-	onAcceptState, err := state.NewDiff(parentID, v.backend)
+	isAddingStakerAfterDeletionAllowed := state.StakerAdditionAfterDeletionLegality(
+		v.txExecutorBackend.Config.UpgradeConfig.IsHeliconActivated(b.Timestamp()),
+	)
+	onAcceptState, err := state.NewDiff(parentID, v.backend, isAddingStakerAfterDeletionAllowed)
 	if err != nil {
 		return err
 	}
@@ -159,11 +165,11 @@ func (v *verifier) ApricotProposalBlock(b *block.ApricotProposalBlock) error {
 	}
 
 	parentID := b.Parent()
-	onCommitState, err := state.NewDiff(parentID, v.backend)
+	onCommitState, err := state.NewDiff(parentID, v.backend, state.StakerAdditionAfterDeletionForbidden)
 	if err != nil {
 		return err
 	}
-	onAbortState, err := state.NewDiff(parentID, v.backend)
+	onAbortState, err := state.NewDiff(parentID, v.backend, state.StakerAdditionAfterDeletionForbidden)
 	if err != nil {
 		return err
 	}
@@ -189,7 +195,7 @@ func (v *verifier) ApricotStandardBlock(b *block.ApricotStandardBlock) error {
 	}
 
 	parentID := b.Parent()
-	onAcceptState, err := state.NewDiff(parentID, v)
+	onAcceptState, err := state.NewDiff(parentID, v, state.StakerAdditionAfterDeletionForbidden)
 	if err != nil {
 		return err
 	}
--- a/vms/platformvm/block/executor/verifier_test.go
+++ b/vms/platformvm/block/executor/verifier_test.go
@@ -741,11 +741,11 @@ func TestBanffAbortBlockTimestampChecks(t *testing.T) {
 			s.EXPECT().GetAccruedFees().Return(uint64(0)).Times(3)
 			s.EXPECT().NumActiveL1Validators().Return(0).Times(3)
 
-			onDecisionState, err := state.NewDiff(parentID, backend)
+			onDecisionState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
-			onCommitState, err := state.NewDiff(parentID, backend)
+			onCommitState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
-			onAbortState, err := state.NewDiff(parentID, backend)
+			onAbortState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
 			backend.blkIDToState[parentID] = &blockState{
 				timestamp:      test.parentTime,
@@ -848,11 +848,11 @@ func TestBanffCommitBlockTimestampChecks(t *testing.T) {
 			s.EXPECT().GetAccruedFees().Return(uint64(0)).Times(3)
 			s.EXPECT().NumActiveL1Validators().Return(0).Times(3)
 
-			onDecisionState, err := state.NewDiff(parentID, backend)
+			onDecisionState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
-			onCommitState, err := state.NewDiff(parentID, backend)
+			onCommitState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
-			onAbortState, err := state.NewDiff(parentID, backend)
+			onAbortState, err := state.NewDiff(parentID, backend, state.StakerAdditionAfterDeletionForbidden)
 			require.NoError(err)
 			backend.blkIDToState[parentID] = &blockState{
 				timestamp:      test.parentTime,
@@ -1350,7 +1350,7 @@ func TestDeactivateLowBalanceL1Validators(t *testing.T) {
 				require.NoError(s.PutL1Validator(l1Validator))
 			}
 
-			diff, err := state.NewDiffOn(s)
+			diff, err := state.NewDiffOn(s, state.StakerAdditionAfterDeletionAllowed)
 			require.NoError(err)
 
 			config := validatorfee.Config{
--- a/vms/platformvm/state/diff.go
+++ b/vms/platformvm/state/diff.go
@@ -73,12 +73,19 @@ type diff struct {
 func NewDiff(
 	parentID ids.ID,
 	stateVersions Versions,
+	allowAddingStakerAfterDeletion StakerAdditionAfterDeletionLegality,
 ) (Diff, error) {
 	parentState, ok := stateVersions.GetState(parentID)
 	if !ok {
 		return nil, fmt.Errorf("%w: %s", ErrMissingParentState, parentID)
 	}
 	return &diff{
+		currentStakerDiffs: diffStakers{
+			isAdditionAfterDeletionAllowed: allowAddingStakerAfterDeletion,
+		},
+		pendingStakerDiffs: diffStakers{
+			isAdditionAfterDeletionAllowed: allowAddingStakerAfterDeletion,
+		},
 		parentID:                    parentID,
 		stateVersions:               stateVersions,
 		timestamp:                   parentState.GetTimestamp(),
@@ -101,10 +108,10 @@ func (s stateGetter) GetState(ids.ID) (Chain, bool) {
 	return s.state, true
 }
 
-func NewDiffOn(parentState Chain) (Diff, error) {
+func NewDiffOn(parentState Chain, allowAddingStakerAfterDeletion StakerAdditionAfterDeletionLegality) (Diff, error) {
 	return NewDiff(ids.Empty, stateGetter{
 		state: parentState,
-	})
+	}, allowAddingStakerAfterDeletion)
 }
 
 func (d *diff) GetTimestamp() time.Time {
@@ -593,13 +600,15 @@ func (d *diff) Apply(baseState Chain) error {
 	}
 	for _, subnetValidatorDiffs := range d.currentStakerDiffs.validatorDiffs {
 		for _, validatorDiff := range subnetValidatorDiffs {
-			switch validatorDiff.validatorStatus {
-			case added:
-				if err := baseState.PutCurrentValidator(validatorDiff.validator); err != nil {
+			// We might have removed the validator and then added it in the same diff.
+			// We therefore first delete and then only after add it.
+			if validatorDiff.removed != nil {
+				baseState.DeleteCurrentValidator(validatorDiff.removed)
+			}
+			if validatorDiff.added != nil {
+				if err := baseState.PutCurrentValidator(validatorDiff.added); err != nil {
 					return err
 				}
-			case deleted:
-				baseState.DeleteCurrentValidator(validatorDiff.validator)
 			}
 
 			addedDelegatorIterator := iterator.FromTree(validatorDiff.addedDelegators)
@@ -622,13 +631,15 @@ func (d *diff) Apply(baseState Chain) error {
 	}
 	for _, subnetValidatorDiffs := range d.pendingStakerDiffs.validatorDiffs {
 		for _, validatorDiff := range subnetValidatorDiffs {
-			switch validatorDiff.validatorStatus {
-			case added:
-				if err := baseState.PutPendingValidator(validatorDiff.validator); err != nil {
+			// We might have removed the validator and then added it in the same diff.
+			// We therefore first delete and then only after add it.
+			if validatorDiff.removed != nil {
+				baseState.DeletePendingValidator(validatorDiff.removed)
+			}
+			if validatorDiff.added != nil {
+				if err := baseState.PutPendingValidator(validatorDiff.added); err != nil {
 					return err
 				}
-			case deleted:
-				baseState.DeletePendingValidator(validatorDiff.validator)
 			}
 
 			addedDelegatorIterator := iterator.FromTree(validatorDiff.addedDelegators)
--- a/vms/platformvm/state/diff_test.go
+++ b/vms/platformvm/state/diff_test.go
@@ -33,7 +33,7 @@ func (nilStateGetter) GetState(ids.ID) (Chain, bool) {
 
 func TestDiffMissingState(t *testing.T) {
 	parentID := ids.GenerateTestID()
-	_, err := NewDiff(parentID, nilStateGetter{})
+	_, err := NewDiff(parentID, nilStateGetter{}, StakerAdditionAfterDeletionAllowed)
 	require.ErrorIs(t, err, ErrMissingParentState)
 }
 
@@ -42,7 +42,7 @@ func TestNewDiffOn(t *testing.T) {
 
 	state := newTestState(t, memdb.New())
 
-	d, err := NewDiffOn(state)
+	d, err := NewDiffOn(state, StakerAdditionAfterDeletionAllowed)
 	require.NoError(err)
 
 	assertChainsEqual(t, state, d)
@@ -53,7 +53,7 @@ func TestDiffFeeState(t *testing.T) {
 
 	state := newTestState(t, memdb.New())
 
-	d, err := NewDiffOn(state)
+	d, err := NewDiffOn(state, StakerAdditionAfterDeletionAllowed)
 	require.NoError(err)
 
 	initialFeeState := state.GetFeeState()
@@ -74,7 +74,7 @@ func TestDiffL1ValidatorExcess(t *testing.T) {
 
 	state := newTestState(t, memdb.New())
 
-	d, err := NewDiffOn(state)
+	d, err := NewDiffOn(state, StakerAdditionAfterDeletionAllowed)
 	require.NoError(err)
 
 	initialExcess := state.GetL1ValidatorExcess()
@@ -92,7 +92,7 @@ func TestDiffAccruedFees(t *testing.T) {
 
 	state := newTestState(t, memdb.New())
 
-	d, err := NewDiffOn(state)
+	d, err := NewDiffOn(state, StakerAdditionAfterDeletionAllowed)
 	require.NoError(err)
 
 	initialAccruedFees := state.GetAccruedFees()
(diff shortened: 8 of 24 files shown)


snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +109,185 @@ func NewBenchlist(
...
+				if !ok || deadline.After(now) {
+					break
+				}
+				timeoutHeap.Pop()
+				benched.Remove(nodeID)
+
+				b.ctx.Log.Debug("unbenching node due to timeout",
+					zap.Stringer("nodeID", nodeID),
+				)
+				b.benchable.Unbenched(b.ctx.ChainID, nodeID)
```
aaronbuchwald: if `Unbenched` can block on something external to the benchlist and that thing could also be blocked waiting for backpressure on the channel to clear this could deadlock.

aaronbuchwald: @copilot take a stab at a refactor to prevent any potential deadlock here

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +109,185 @@ func NewBenchlist(
...
-// IsBenched returns true if messages to [nodeID] should not be sent over the
-// network and should immediately fail.
+// IsBenched returns true if messages to nodeID should immediately fail.
 func (b *benchlist) IsBenched(nodeID ids.NodeID) bool {
 	b.lock.RLock()
 	defer b.lock.RUnlock()
 
-	return b.benchlistSet.Contains(nodeID)
+	n, ok := b.nodes[nodeID]
+	return ok && n.isBenched && !b.benchExpired(n)
```
aaronbuchwald: Checking `benchExpired` here means that we check whether the timeout should have fired, not whether it actually has and updated the benchlist and sent `Bench` / `Unbench` notifications. I am not sure if there is a path where this inconsistency could cause issues, but it is probably better to avoid such a divergence in this error prone section of the code.

This may make testing the timing of this more difficult, but that seems preferred over having much harder to reason about inconsistency between `IsBenched` and notifications. @copilot draft a PR to fix this.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
+				if n, exists := b.nodes[nodeID]; exists {
+					n.isBenched = false
+					n.failureProbability = math.NewUninitializedAverager(b.halflife)
+				}
+				b.lock.Unlock()
+
+				b.ctx.Log.Debug("unbenching node due to timeout",
+					zap.Stringer("nodeID", nodeID),
+				)
+				b.benchable.Unbenched(b.ctx.ChainID, nodeID)
```
aaronbuchwald: Keeping these in sync is probably not really necessary. We explicitly cannot hold the lock while firing the notification anyways, so it's just a question of ordering and shouldn't matter.

snow/networking/benchlist/benchlist.go
```diff
@@ -16,100 +14,89 @@ import (
...
+// Config defines the configuration for a benchlist
+type Config struct {
+	Halflife           time.Duration `json:"halflife"`
+	UnbenchProbability float64       `json:"unbenchProbability"`
+	BenchProbability   float64       `json:"benchProbability"`
+	BenchDuration      time.Duration `json:"benchDuration"`
+}
 
-	// notified when a node is benched or unbenched
+// If a node does not respond to a request, the node waits for a timeout. This
```
yacovm: Can we be consistent with the terms? we have both node and peer.

yacovm: > // If a node does not respond to a request, the node waits for a timeout. This  
// can cause elevated latencies if a peer is frequently sent requests.


If a remote node does not respond to a request, the sender node waits for a timeout. This can cause elevated latencies of the sender node frequently sends requests to the remote node.

snow/networking/benchlist/benchlist.go
```diff
@@ -16,100 +14,89 @@ import (
...
+// can cause elevated latencies if a peer is frequently sent requests.
+//
+// Therefore, we attempt to project whether or not a peer is likely to respond
+// to a query. If a node is projected to fail, it is "benched". While it is
+// benched, queries to that node fail immediately to avoid waiting up to the
+// full network timeout.
+//
+// If a node remains benched for longer than [benchDuration], it is
```
yacovm, line 72: Should probably be called `maxBenchDuration`

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
-	b.numBenched.Set(float64(b.benchedHeap.Len()))
-	benchedStake, err := b.vdrs.SubsetWeight(b.ctx.SubnetID, b.benchlistSet)
-	if err != nil {
-		b.ctx.Log.Error("error calculating benched stake",
-			zap.Stringer("subnetID", b.ctx.SubnetID),
-			zap.Error(err),
-		)
+	n, ok := b.nodes[nodeID]
+	if weight := b.vdrs.GetWeight(b.ctx.SubnetID, nodeID); weight == 0 && (!ok || !n.isBenched) {
+		// Don't track non-validators unless they're already benched to avoid
```
yacovm: > 		// Don't track non-validators unless they're already benched to avoid
		// excess memory pressure.

Why would we want to keep tracking a non validator that is benched? Isn't the point of this benchlist to speedup consensus? We don't query non validators, so even if we time out on a non validator, it's not in the critical path (consensus) so it has no effect on performance.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
+				timeoutHeap.Pop()
+				benched.Remove(nodeID)
+				b.lock.Lock()
+				// Timeout unbenching originates in this consumer goroutine
+				// (not in observe), so we must synchronize b.nodes here to
+				// keep IsBenched consistent with callback-visible state.
+				// Timeout-based unbench gives the node a clean slate.
+				if n, exists := b.nodes[nodeID]; exists {
+					n.isBenched = false
+					n.failureProbability = math.NewUninitializedAverager(b.halflife)
```
yacovm: Are we sure there is no ill effect in initializing the averager in this way? 

It will initialize it with a time of zero which is many years ago. 

Wouldn't it be better to initialize it like this?

```
n.failureProbability = math.NewAverager(0, b.halflife, now)
```

It makes the accounting much more accurate, I think.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
 }
 
-// RegisterResponse notes that we received a response from [nodeID]
-func (b *benchlist) RegisterResponse(nodeID ids.NodeID) {
-	b.streaklock.Lock()
-	defer b.streaklock.Unlock()
-
-	delete(b.failureStreaks, nodeID)
-}
+// TODO: Close goroutine within run during shutdown.
```
yacovm: Any reason we don't do it in this PR?

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
+func (b *benchlist) run() {
+	var (
+		benched     set.Set[ids.NodeID]
```
yacovm: this `benched` is a local variable and is controlled by both the input from the jobs queue *and* the expiration timer. 

The expiration timer can remove a node from being benched **in parallel** to a job that benches the node is still waiting in the queue. If that happens, the node in `n.isBenched` will be false, so it won't be un-benched by the observations. However, since the job will be run after the timer acted, the router will be notified to disconnect from the node. So, we can get into a situation where the last thing the router was told to do, was disconnect from the node, but the `b.nodes[nodeID].isBenched` is false, which is inconsistent. 

We therefore need to double check `b.nodes[nodeID].isBenched` before we propagate anything to the router.

snow/networking/benchlist/benchlist.go
```diff
@@ -16,100 +14,89 @@ import (
...
+	jobs     *buffer.UnboundedBlockingDeque[job]
+	jobReady chan struct{}
 
-	// A validator will be benched if [threshold] messages in a row
-	// to them time out and the first of those messages was more than
-	// [minimumFailingDuration] ago
-	threshold              int
-	minimumFailingDuration time.Duration
+	lock  sync.RWMutex
+	nodes map[ids.NodeID]*node
```
yacovm: I don't think we ever delete from this map. Isn't that a problem? 

I think we should delete if the node is not a validator anymore.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
+	shouldBench := !n.isBenched && p > b.benchProbability
+	shouldUnbench := n.isBenched && p < b.unbenchProbability
+	if shouldBench || shouldUnbench {
+		n.isBenched = !n.isBenched
+		_ = b.jobs.PushRight(job{
+			nodeID: nodeID,
+			bench:  n.isBenched,
+		})
+		select {
+		case b.jobReady <- struct{}{}:
```
yacovm: I don't usually comment on code style and patterns, but too long functions and concurrency patterns are among my fixations. 

1. Can we please make the functions smaller? we can extract each case into a function and make the for loop more friendly to read.
2. This is more of a recommendation, but I think we can simplify the code if we extend the job to do either a cleanup triggered via the timer, or the bench/unbench, and then we don't need this jobReady channel. We can then just wait for the queue to have something inside of it, as it is already blocking and thread safe. So, we can have a goroutine that will check the deadlines, and if it is due to run an expiration, we just enqueue it into the jobs queue. And this way, we only do updates through a single goroutine. The only shared structure can be the heap, and the jobs goroutine will add to it and the expiration goroutine will remove from it.

yacovm (commented):
Not finished yet but it's getting late here. Posting what i have in the meantime.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +105,193 @@ func NewBenchlist(
...
-	benchedStake, err := b.vdrs.SubsetWeight(b.ctx.SubnetID, b.benchlistSet)
-	if err != nil {
-		b.ctx.Log.Error("error calculating benched stake",
-			zap.Stringer("subnetID", b.ctx.SubnetID),
-			zap.Error(err),
-		)
+	n, ok := b.nodes[nodeID]
+	if weight := b.vdrs.GetWeight(b.ctx.SubnetID, nodeID); weight == 0 && (!ok || !n.isBenched) {
+		// Don't track non-validators unless they're already benched to avoid
+		// excess memory pressure.
```
yacovm: Thinking some more, is there a reason we shouldn't just move all handling of `b.nodes` in this function into the job dispatcher goroutine? 

We can just pass the observation into the jobs queue, and let the job dispatcher figure out the required changes where it has access to both the information that the expiration timer touches.

Thoughts?

snow/networking/benchlist/benchlist.go
aaronbuchwald: Good calll

aaronbuchwald: Will re-phrase

snow/networking/benchlist/benchlist.go
aaronbuchwald: Good callout - I just used this assuming it was the appropriate way to initialize a new averager.

Reading the comment on the constructor for the uninitialized averager, it seems that initializing it with time of 0 is intentional to solve a problem we'd likely run into here - overweighting the first observation https://github.com/ava-labs/avalanchego/blob/09a9530c3145e3990aac1c920d7d72fea1077873/utils/math/continuous_averager.go#L22.

If we initialized it to now instead of 0 as in the uninitialized constructor, then a single initial failure would result in an observed ~100% failure probability and defeat the purpose of resetting it.

snow/networking/benchlist/benchlist.go
aaronbuchwald: There's no longer randomness in this change (perhaps it should be a requirement to add it back), but with the current code `benchDuration` seems correct to me.

snow/networking/benchlist/benchlist.go
aaronbuchwald: Wrote this up to show the behavior (we use 1 to indicate failure and 0 as success ie. it is the average probability of failure)

```golang
func TestAveragerInit(t *testing.T) {
	require := require.New(t)

	now := time.Now()

	newAvgInitialFail := NewAverager(0, time.Second, now)
	newAvgInitialFail.Observe(1, now)
	require.Equal(0.5, newAvgInitialFail.Read())

	newAvgInitialSuccess := NewAverager(0, time.Second, now)
	newAvgInitialSuccess.Observe(0, now)
	require.Equal(0.0, newAvgInitialSuccess.Read())

	uninitAvgInitialFail := NewUninitializedAverager(time.Second)
	uninitAvgInitialFail.Observe(1, now)
	require.Equal(1.0, uninitAvgInitialFail.Read())

	uninitAvgInitialSuccess := NewUninitializedAverager(time.Second)
	uninitAvgInitialSuccess.Observe(0, now)
	require.Equal(0.0, uninitAvgInitialSuccess.Read())
}
```

snow/networking/benchlist/benchlist.go
aaronbuchwald: Yup, that's correct

snow/networking/benchlist/benchlist.go
aaronbuchwald: Agree this would be a better structure.

snow/networking/benchlist/benchlist.go
aaronbuchwald: Updated so everything is sent to one goroutine to handle both observations, benching/unbenching, and timeouts and broke it into smaller functions.

snow/networking/benchlist/benchlist.go
aaronbuchwald: Adding

snow/networking/benchlist/benchlist.go
aaronbuchwald: refactored as per https://github.com/ava-labs/avalanchego/pull/4982#discussion_r2849593432

snow/networking/benchlist/benchlist.go
aaronbuchwald: Switching to `NewAverager(0, halflife, now)` as you suggested. I tried out `NewMaturedAverager` as well, but this needs to be accounted for in the tests in a way that decreases readability. Since it is a relatively soft improvement if at all, I decided that it wasn't worth it.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
-		}
-		if now.Before(next) {
-			break
+// processObservation updates a node's EWMA and transitions bench state if the
+// failure probability crosses a threshold.
+func (b *benchlist) processObservation(ev event) {
+	nodeID := ev.nodeID
+
+	n, ok := b.nodes[nodeID]
+	if b.vdrs.GetWeight(b.ctx.SubnetID, nodeID) == 0 {
```
yacovm, line 249: I am wondering whether it makes sense to move this to the `enqueue` to not waste resources on enqueuing something we will end up ignoring.

I guess we have to keep this here because the validator might be unstaked/exit or whatever we call it and we need to remove it from being monitored which we only do in this goroutine, is that it?

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
+	switch {
+	case !n.isBenched && p > b.benchProbability:
+		if !b.canBench(nodeID) {
+			return
+		}
 
-	return b.benchlistSet.Contains(nodeID)
-}
+		n.isBenched = true
+		b.timeoutHeap.Push(nodeID, time.Now().Add(b.benchDuration))
```
yacovm, line 274: can we use the clock time here? To be better with testing and more rigorous.

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
-		// This is the first consecutive failure
-		failureStreak.firstFailure = now
-	}
-	b.failureStreaks[nodeID] = failureStreak
-	b.streaklock.Unlock()
+// processTimeouts unbenches any nodes whose bench duration has expired.
+// Timeout-based unbench gives the node a clean EWMA slate so that a single
+// failure after unbenching doesn't immediately re-bench it.
+func (b *benchlist) processTimeouts() {
+	now := time.Now()
```
yacovm, line 305: can we use clock time here?

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
+	defer b.lock.RUnlock()
+	return b.benched.Contains(nodeID)
 }
 
-// TODO: Close this goroutine during node shutdown
+// --- Consumer goroutine (single owner of all mutable node state) ---
+
+// run is the consumer goroutine. It owns the nodes map, timeout heap, and is
+// the only goroutine that calls Benched/Unbenched on the benchable.
 func (b *benchlist) run() {
```
yacovm, line 202: since we have a shutdown method, can we add a waitgroup and Done() it when we return from `run()` ? 

Otherwise we just YOLO the shutdown (invoke it and hope for the best).

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
+	p := n.failureProbability.Read()
 
-// IsBenched returns true if messages to [nodeID] should not be sent over the
-// network and should immediately fail.
-func (b *benchlist) IsBenched(nodeID ids.NodeID) bool {
-	b.lock.RLock()
-	defer b.lock.RUnlock()
+	switch {
+	case !n.isBenched && p > b.benchProbability:
+		if !b.canBench(nodeID) {
```
yacovm, line 269: can we add a debug log here?

snow/networking/benchlist/benchlist.go
```diff
@@ -118,189 +136,203 @@ func NewBenchlist(
...
-func (b *benchlist) RegisterResponse(nodeID ids.NodeID) {
-	b.streaklock.Lock()
-	defer b.streaklock.Unlock()
+		b.lock.Lock()
+		b.benched.Add(nodeID)
+		b.lock.Unlock()
 
-	delete(b.failureStreaks, nodeID)
-}
+		b.ctx.Log.Debug("benching node", zap.Stringer("nodeID", nodeID))
```
yacovm: can we add here and in `b.ctx.Log.Debug("unbenching node",` the `p`?

snow/networking/benchlist/benchlist.go
```diff
@@ -309,52 +341,76 @@ func (b *benchlist) bench(nodeID ids.NodeID) {
...
-
-	b.ctx.Log.Debug("benching validator after consecutive failed queries",
-		zap.Stringer("nodeID", nodeID),
-		zap.Duration("benchDuration", benchedUntil.Sub(now)),
-		zap.Int("numFailedQueries", b.threshold),
-	)
+// benchedStake returns the total stake weight of currently benched validators.
+func (b *benchlist) benchedStake() (uint64, error) {
+	benchedNodeIDs := set.NewSet[ids.NodeID](len(b.nodes))
+	for id, n := range b.nodes {
```
yacovm: why do we loop here if we have `b.benched` ?

snow/networking/benchlist/benchlist_test.go
```diff
@@ -4,364 +4,344 @@
...
-	b.lock.Lock()
-	b.clock.Set(now)
-
-	benched := false
-	benchable.BenchedF = func(ids.ID, ids.NodeID) {
-		benched = true
+	requireBenched := func() {
+		t.Helper()
+		require.True(b.IsBenched(vdrID))
+		require.False(b.IsBenched(nodeID))
```
yacovm, line 110: Is this a typo and supposed to be true?

yacovm: oh I guess one is a non validator but I cannot understand from the test why that is the case, can you explain? seems like both have weight of 1.

snow/networking/benchlist/benchlist_test.go
```diff
@@ -4,364 +4,344 @@
...
 	)
+	require.NoError(err)
+	defer b.shutdown()
+
+	// Bench the node and wait for the consumer to block in Benched.
+	b.RegisterFailure(vdrID)
+	b.RegisterFailure(vdrID)
+	select {
+	case <-benchable.benchedCalled:
+	case <-time.After(time.Second):
```
yacovm: 1 second seems to me like a bit borderline and flaky, lets make it more like 10 sec?

snow/networking/benchlist/benchlist_test.go
```diff
@@ -4,364 +4,344 @@
...
-	require.Eventually(
-		func() bool {
-			return !b.IsBenched(vdrID2)
-		},
-		duration+time.Second,
-		100*time.Millisecond,
-	)
+	select {
+	case <-done:
+	case <-time.After(time.Second):
```
yacovm: 1 second seems to me like a bit borderline and flaky, lets make it more like 10 sec?

snow/networking/benchlist/benchlist.go
aaronbuchwald: I'll add a `shutdownDone` chan to make sure that shutdown blocks until the goroutine is cleaned up.

snow/networking/benchlist/benchlist.go
aaronbuchwald: There are debug or error logs at all of the appropriate sites within `canBench` with more detail including the reason, so I'd prefer to keep it at the more specific locations rather than add it here.

snow/networking/benchlist/benchlist.go
aaronbuchwald: Happy to add in the failure probability

snow/networking/benchlist/benchlist_test.go
aaronbuchwald: Updating to use `defaultTestTimeout` of 10s

snow/networking/benchlist/benchlist_test.go
aaronbuchwald: https://github.com/ava-labs/avalanchego/pull/4982#discussion_r2861042410

snow/networking/benchlist/benchlist.go
aaronbuchwald: Good catch, updating

snow/networking/benchlist/benchlist.go
aaronbuchwald: I prefer to keep it here so that we fetch the node stake once. In addition, the benchlist only gets invoked when a response/timeout gets called for a request we actually sent.

We do send requests to non-validators in some p2p protocols, but it's mostly (might be all) during bootstrapping, so should not have a large impact.

snow/networking/benchlist/benchlist_test.go
aaronbuchwald: `nodeID` is not registered as a validator, so it gets ignored. There are two different nodeIDs added as validators in the test, but one of them is a throwaway that we don't store anywhere just so that we end up with one validator having the stake % we want. `nodeID` is a non-validator.

ref https://github.com/ava-labs/avalanchego/blob/a026dc393f766b22de18ba9c78eeb903b75f9234/snow/networking/benchlist/benchlist_test.go#L77

snow/networking/benchlist/benchlist.go
aaronbuchwald: Updating

snow/networking/benchlist/benchlist.go
aaronbuchwald: Updating

yacovm (approved):

config/config.go
```diff
@@ -554,24 +554,16 @@ func getBenchlistConfig(v *viper.Viper, snowballParameters *snowball.Parameters)
 	if snowballParameters == nil {
 		return benchlist.Config{}, errors.New("snowball parameters must be non-nil")
 	}
-	// AlphaConfidence is used here to ensure that benching can't cause a
-	// liveness failure. If AlphaPreference were used, the benchlist may grow to
-	// a point that committing would be extremely unlikely to happen.
```
StephenButtolph, line 559: This comment still seems relevant no?

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
+// run is the consumer goroutine. It owns the nodes map, timeout heap, and is
+// the only goroutine that calls Benched/Unbenched on the benchable.
 func (b *benchlist) run() {
+	defer close(b.shutdownDone)
+
 	timer := time.NewTimer(0)
 	defer timer.Stop()
 
+	// Drain the initial timer fire since nothing is benched yet.
+	<-timer.C
```
StephenButtolph: We have a helper for setting up a stopped timer
```suggestion
	timer := timer.StoppedTimer()
	defer timer.Stop()
```

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
-	// Update the timer to account for the newly benched node.
-	select {
-	case b.resetTimer <- struct{}{}:
-	default:
+	if _, deadline, ok := b.timeoutHeap.Peek(); ok {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			remaining = 0
+		}
+		timer.Reset(remaining)
```
StephenButtolph: `Reset` handles negative numbers just fine.
```suggestion
		timer.Reset(time.Until(deadline))
```

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
-		zap.Int("numFailedQueries", b.threshold),
-	)
-
-	// Add to benchlist times with randomized delay
-	b.benchlistSet.Add(nodeID)
-	b.benchable.Benched(b.ctx.ChainID, nodeID)
+// benchedStake returns the total stake weight of currently benched validators.
+func (b *benchlist) benchedStake() (uint64, error) {
+	b.lock.RLock()
+	benchedNodeIDs := set.Of(b.benched.List()...)
```
StephenButtolph: Nit: We don't need to allocate the slice here.
```suggestion
	var benchedNodeIDs set.Set[ids.NodeID]
	b.lock.RLock()
	benchedNodeIDs.Union(b.benched)
```

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
-	b.benchedHeap.Push(nodeID, benchedUntil)
+// resetTimer stops the timer and resets it to fire at the earliest deadline in
+// the timeout heap. If the heap is empty, the timer remains stopped and will be
+// reset when the next event arrives.
+func (b *benchlist) resetTimer(timer *time.Timer) {
+	if !timer.Stop() {
+		select {
+		case <-timer.C:
+		default:
+		}
```
StephenButtolph, line 412: Isn't this racy? I don't think `Stop()` returning `false` means that `timer.C` is populated. I think it means that it will _eventually_ be populated.

The race is probably fine... But wanted to know if this was required for some reason.

snow/networking/benchlist/benchlist.go
```diff
@@ -16,100 +15,120 @@ import (
...
+	clock              mockable.Clock
+	halflife           time.Duration
+	unbenchProbability float64
+	benchProbability   float64
+	benchDuration      time.Duration
+	maxPortion         float64
+
+	// Event queue: producers push observations; the single consumer goroutine
+	// pops and processes them. The queue is unbounded so producers never block.
+	events     *buffer.UnboundedBlockingDeque[event]
```
StephenButtolph: Presumably we are using a blocking deque here rather than the normal deque because it is thread safe? It feels a bit weird to have a blocking deque but to never expect to block on the deque.

I feel like ideally we would just have the `events` queue, and then remove `eventReady` and `shutdownChan`. Since both of those are really already handled by the blocking deque... But It seems like that might be harder to incorporate with the timer... I guess it's fine...

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
-	if failureStreak.firstFailure.IsZero() {
-		// This is the first consecutive failure
-		failureStreak.firstFailure = now
-	}
-	b.failureStreaks[nodeID] = failureStreak
-	b.streaklock.Unlock()
+// newFailureProbabilityAverager creates a failure probability averager with an
+// optimistic prior to slightly favor newly tracked nodes.
+func (b *benchlist) newFailureProbabilityAverager(now time.Time) math.Averager {
+	return math.NewAverager(0, b.halflife, now)
```
StephenButtolph:
```suggestion
	return math.NewAverager(success, b.halflife, now)
```

snow/networking/benchlist/benchlist.go
```diff
@@ -118,243 +137,292 @@ func NewBenchlist(
...
+		b.benched.Remove(nodeID)
+		b.lock.Unlock()
 
-	if b.benchlistSet.Contains(nodeID) {
-		// This validator is benched. Ignore failures until they're not.
-		return
+		b.ctx.Log.Debug("unbenching node due to timeout",
+			zap.Stringer("nodeID", nodeID),
+			zap.Float64("oldFailureProbability", oldFailureProbability),
+			zap.Float64("resetFailureProbability", resetFailureProbability),
```
StephenButtolph: Why do we log `resetFailureProbability`? It is always 0.

snow/networking/benchlist/benchlist.go
```diff
@@ -16,100 +15,120 @@ import (
...
-	duration time.Duration
-
-	// The maximum percentage of total network stake that may be benched
-	// Must be in [0,1)
-	maxPortion float64
+	vdrs          validators.Manager
+	numBenched    prometheus.Gauge
+	weightBenched prometheus.Gauge
+
+	clock              mockable.Clock
```
StephenButtolph: If we don't actually mock the clock in the tests, it just adds complexity.
```suggestion
```

snow/networking/benchlist/benchlist.go
aaronbuchwald: Fixed in 76dccfb5e0df0c176dbc7d212fc116d2e30054ff

snow/networking/benchlist/benchlist.go
aaronbuchwald: Removed in b1e71ccaed372bde555c6caf2d8dcdbf5d13ccd4

snow/networking/benchlist/benchlist.go
aaronbuchwald: Fixed in 35dd2e3aeb24645b773ed081739ae768bd335f6e

snow/networking/benchlist/benchlist.go
aaronbuchwald: Fixed in 4daddeee90baef61627b02f12afe78092408facd

snow/networking/benchlist/benchlist.go
aaronbuchwald: Accepted the suggestion and cleaned up remaining references to `b.clock.Time()` in 836b8bf787a9a9724a5a163caf13a3faae7a8010

config/config.go
aaronbuchwald: Restored in 3e595fb37a7f0c3317482672684f6420cadc3f05

snow/networking/benchlist/benchlist.go
aaronbuchwald: Switched to a plain `UnboundedDeque` with an explicit `sync.Mutex` in e103298295485a483cc0d7a7903c632d5f854824. The blocking deque's blocking semantics were unused since the consumer goroutine uses a `select` loop to wait on the timer and shutdown channels alongside the event signal — `sync.Cond` can't participate in `select`. The explicit lock makes it clear we only need thread-safe push/pop, not blocking.

snow/networking/benchlist/benchlist.go
aaronbuchwald: The `default` case is required because the `run` loop may have already consumed the timer value via `case <-timer.C` — a blocking drain would deadlock in that case. Added a comment explaining this and the behavior when the timer hasn't delivered yet in db74defd3b0debd8ef2df48903ea37b1851fb4ea.

PR #4983, opened 2026-02-17
Remove Validator State Interface

samliok:
## Why this should be merged

This PR removes an unnecessary level of indirection since the validatorState interface was only implemented once.

## How this works

Replaces the `validatorState` interface with a `validatorState` struct

No

### DIFF
--- a/vms/platformvm/state/metadata_validator.go
+++ b/vms/platformvm/state/metadata_validator.go
@@ -20,8 +20,6 @@ import (
 // CodecVersionLen + UpDurationLen + LastUpdatedLen + PotentialRewardLen
 const preDelegateeRewardSize = codec.VersionSize + 3*wrappers.LongLen
 
-var _ validatorState = (*metadata)(nil)
-
 type preDelegateeRewardMetadata struct {
 	UpDuration      time.Duration `v0:"true"`
 	LastUpdated     uint64        `v0:"true"` // Unix time in seconds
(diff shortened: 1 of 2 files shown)


vms/platformvm/state/metadata_validator.go
```diff
@@ -77,79 +75,20 @@ func parseValidatorMetadata(bytes []byte, metadata *validatorMetadata) error {
...
-	// SetUptime or SetDelegateeReward.
-	WriteValidatorMetadata(
-		dbPrimary database.KeyValueWriter,
-		dbSubnet database.KeyValueWriter,
-		codecVersion uint16,
-	) error
-}
-
-type metadata struct {
+type validatorState struct {
```
DracoLi: would a name like `validatorMetadataState` be better here? `validatorState` might create some confusion with the `validators.State` interface from `manager.go`

vms/platformvm/state/metadata_validator.go
```diff
@@ -190,7 +129,7 @@ func (m *metadata) SetUptime(
 	return nil
 }
 
-func (m *metadata) GetDelegateeReward(
+func (m *validatorState) GetDelegateeReward(
```
DracoLi: Should we copy over the doc from the interface to these methods?
```
	// GetDelegateeReward returns the current rewards accrued to [vdrID] on
	// [subnetID].
```

samliok: great suggestion, updated!

vms/platformvm/state/metadata_validator.go
```diff
@@ -77,79 +75,24 @@ func parseValidatorMetadata(bytes []byte, metadata *validatorMetadata) error {
...
 		updatedMetadata: make(map[ids.NodeID]set.Set[ids.ID]),
 	}
 }
 
-func (m *metadata) LoadValidatorMetadata(
+// LoadValidatorMetadata sets the [metadata] of [vdrID] on [subnetID].
+// GetUptime and SetUptime will return an error if the [vdrID] and
+// [subnetID] hasn't been loaded. This call will not result in a write to
+// disk.
+func (m *validatorState) LoadValidatorMetadata(
```
joshua-kim: The Go convention for receivers is generally to name it the first letter of the type, or some abbreviation. For `validatorState` `v` for "validatorState" or `s` for "state" or `vs` for "validatorState" would be appropriate ([ref](https://go.dev/wiki/CodeReviewComments#receiver-names)).

vms/platformvm/state/metadata_validator.go
```diff
@@ -173,7 +118,10 @@ func (m *metadata) GetUptime(
 	return metadata.UpDuration, metadata.lastUpdated, nil
 }
 
-func (m *metadata) SetUptime(
+// SetUptime updates the uptime measurements of [vdrID] on [subnetID].
+// Unless these measurements are deleted first, the next call to
+// WriteUptimes will write this update to disk.
```
joshua-kim: Not blocking on this as it's unimportant if we incorporate the feedback, but this is not correct usage of the godoc syntax. Generally:
* Square brackets are used for exported code references, like `[Foo]` or `[Type.Method]`
* The rest is Markdown -- so if you want to refer to a parameter you might want something like backticks to designate a `parameter`.

This feedback is unimportant right now because this is an unexported type, but generally you only need to follow this syntax for exported code that godoc will render. We also can't put this on `state` where it belongs unfortunately, because that's also not exported...

vms/platformvm/state/metadata_validator.go
```diff
@@ -201,7 +151,10 @@ func (m *metadata) GetDelegateeReward(
 	return metadata.PotentialDelegateeReward, nil
 }
 
-func (m *metadata) SetDelegateeReward(
+// SetDelegateeReward updates the rewards accrued to [vdrID] on [subnetID].
+// Unless these measurements are deleted first, the next call to
+// WriteUptimes will write this update to disk.
```
DracoLi: I think this comment is a bit outdated. there is actually no `WriteUptimes`. We should rename these to `WriteValidatorMetadata` instead.

